PR #11, opened 2018-04-17
Reformat the code base

pierremoreau:
Update the whole code base to follow the LLVM coding style guidelines:
* Update the .clang-tidy to the latest version from LLVM;
* Always use UNIX-style line endings;
* Update all the formatting;
* Update all variables/functions names;
* Add missing “override” keywords;
* Add missing virtual destructor to “TypeVisitor”.

LLVM was compiled with “-Wshadow” and the resulting warnings were analysed to ensure that no undesired shadowing was introduced.


lib/SPIRV/Mangler/ParameterType.h
```diff
@@ -420,6 +420,7 @@ struct UserDefinedType : public ParamType {
 struct TypeVisitor {
   SPIRversion spirVer;
   TypeVisitor(SPIRversion ver) : spirVer(ver){};
+  virtual ~TypeVisitor() {}
```
bader: How about compiler generated destructor? 

`virtual ~TypeVisitor() = default;`

lib/SPIRV/Mangler/FunctionDescriptor.h
```diff
@@ -10,8 +10,8 @@
  * Contributed by: Intel Corporation.
  */
 
-#ifndef __FUNCTION_DESCRIPTOR_H__
-#define __FUNCTION_DESCRIPTOR_H__
+#ifndef LLVM_TOOLS_SPIRV_LLVM_TRANSLATOR_LIB_SPIRV_MANGLER_FUNCTIONDESCRIPTOR_H
```
bader: I'm not sure about this change.
LLVM_TOOLS_SPIRV_LLVM_TRANSLATOR_LIB_SPIRV_MANGLER_FUNCTIONDESCRIPTOR_H doesn't seem to be a good name for the guard. Mangler library is a separate library not related to SPIR-V translator, so I don't think we should use notion of SPIR-V translator in this library. Original sources of Mangler library are hosted here: https://github.com/KhronosGroup/SPIR-Tools/tree/master/spir-tools/spir_name_mangler.
Will it help if we move this library out of SPIR-V translator sources?

lib/SPIRV/OCLTypeToSPIRV.h
```diff
@@ -1,3 +1,6 @@
+#ifndef LLVM_TOOLS_SPIRV_LLVM_TRANSLATOR_LIB_SPIRV_OCLTYPETOSPIRV_H
```
bader: 1. I would prefer move the guards after the copyright header.
2. Should we prefix it with "LLVM_TOOLS_"?

lib/SPIRV/OCLTypeToSPIRV.h
```diff
@@ -84,3 +87,5 @@ class OCLTypeToSPIRV : public ModulePass {
 };
 
 } // namespace SPIRV
+
+#endif
```
bader: Shouldn't it be 
`#endif // LLVM_TOOLS_SPIRV_LLVM_TRANSLATOR_LIB_SPIRV_OCLTYPETOSPIRV_H`
?

bader:
It might useful to add LLVM coding style guidelines checks to CI, so they are not violated in the future.

pierremoreau:
> It might useful to add LLVM coding style guidelines checks to CI, so they are not violated in the future.

Definitely, I’ll have a look at how some of the checks are run for SPIRV-Tools.

lib/SPIRV/Mangler/FunctionDescriptor.h
pierremoreau: Ah, I didn’t know about that. I’m not sure how we should handle that; we can always manually revert those changes and ensure that the Mangler will be ignored by clang-format and clang-tidy.

lib/SPIRV/OCLTypeToSPIRV.h
pierremoreau: I have no idea why clang-tidy didn’t automatically add the comment. I’ll manually add them.

lib/SPIRV/OCLTypeToSPIRV.h
pierremoreau: > 1. I would prefer move the guards after the copyright header.

I’ll manually edit that.

> 2. Should we prefix it with "LLVM_TOOLS_"?

It was automatically done as I was building in tree, so you get that “LLVM_TOOLS_” prefix. I guess if I was building out of tree, we wouldn’t get it. Looking at the clang headers, they only have an “LLVM_” prefix.

lib/SPIRV/Mangler/ParameterType.h
pierremoreau: Right, I forgot about `= default`, that would be better.

lib/SPIRV/Mangler/FunctionDescriptor.h
AlexeySotkin: I see that all headers have got such long header guards. All of them start with `LLVM_TOOLS_SPIRV_LLVM_TRANSLATOR_LIB_SPIRV_` prefix.
I think we should shorten it to `SPIRV_` and disable `llvm-header-guard` check/fix in .clang-tidy

lib/SPIRV/OCLUtil.h
```diff
@@ -135,77 +132,80 @@ struct OCLBuiltinTransInfo {
 //
 ///////////////////////////////////////////////////////////////////////////////
 namespace kOCLBuiltinName {
```
AlexeySotkin: [minor] To me, alignment used before looks cleaner.

lib/SPIRV/SPIRVInternal.h
```diff
@@ -1,934 +1,896 @@
...
+  CLK_ADDRESS_NONE = 0x0000,
+  CLK_ADDRESS_CLAMP = 0x0004,
+  CLK_ADDRESS_CLAMP_TO_EDGE = 0x0002,
+  CLK_ADDRESS_REPEAT = 0x0006,
+  CLK_ADDRESS_MIRRORED_REPEAT = 0x0008,
+  CLK_NORMALIZED_COORDS_FALSE = 0x0000,
+  CLK_NORMALIZED_COORDS_TRUE = 0x0001,
+  CLK_FILTER_NEAREST = 0x0010,
+  CLK_FILTER_LINEAR = 0x0020,
+};
```
AlexeySotkin: [minor] To me, alignment used before looks cleaner.

lib/SPIRV/libSPIRV/SPIRVType.h
```diff
@@ -1,764 +1,757 @@
...
+  _SPIRV_OP(image1d_array_t, Dim1D, 0, 1, 0, 0, 0)
+  _SPIRV_OP(image2d_t, Dim2D, 0, 0, 0, 0, 0)
+  _SPIRV_OP(image2d_array_t, Dim2D, 0, 1, 0, 0, 0)
+  _SPIRV_OP(image2d_depth_t, Dim2D, 1, 0, 0, 0, 0)
+  _SPIRV_OP(image2d_array_depth_t, Dim2D, 1, 1, 0, 0, 0)
+  _SPIRV_OP(image2d_msaa_t, Dim2D, 0, 0, 1, 0, 0)
+  _SPIRV_OP(image2d_array_msaa_t, Dim2D, 0, 1, 1, 0, 0)
+  _SPIRV_OP(image2d_msaa_depth_t, Dim2D, 1, 0, 1, 0, 0)
+  _SPIRV_OP(image2d_array_msaa_depth_t, Dim2D, 1, 1, 1, 0, 0)
+  _SPIRV_OP(image3d_t, Dim3D, 0, 0, 0, 0, 0)
```
AlexeySotkin: [minor] To me, alignment used before looks cleaner

pierremoreau:
I was also thinking of bumping the version to 0.2, as the entrypoint names were modified.

lib/SPIRV/Mangler/FunctionDescriptor.h
pierremoreau: I’ll do that. And it will be easy to change back to that long prefix if asked to do so when moving in the LLVM umbrella.

lib/SPIRV/OCLUtil.h
pierremoreau: I can disable clang-format for that part, but at least manually fix the alignment to be consistent for all of them, as it is currently not the case.

pierremoreau:
I disabled the llvm-header-guard check, changed the prefix from “LLVM_TOOLS_SPIRV_LLVM_TRANSLATOR_LIB_SPIRV_” down to “SPIRV_” as suggested by @AlexeySotkin, moved the header guards after the copyright notice and add missing comments on the header guard `#endif`, as asked by @bader.

Now, what do we want to do exactly regarding the Mangler library? Moving it outside of the SPIR-V sources would make it easier to 1) exclude from clang-tidy and clang-format (not 100% for clang-format), and 2) realise it’s a separate library. I can just change the header guards back to what they were, or change the whole files back.

Regarding @AlexeySotkin comments about the formatting: I agree that it looked nicer before, but that means 1) surrounding each block by a `// clang-format off`/`// clang-format on` pair (not too hard), and 2) manually handling the formatting, and given how the code was a mixed state of no-alignment and almost aligned, I’m not holding my breath on keeping the formatting up-to-date.

AlexeySotkin:
LGTM, Thanks!
For the mangler I think we should create a separate issue to decide what to do with it. 
As for my comments on formatting - they are minor and I don't mind to merge as is.

bader:
LGTM.
@pierremoreau, you are going to rebase the patch set before merging it, right?

P.S. sorry I accidentally missed the right button and closed the request for a few seconds.

pierremoreau:
@bader Yes, I’ll do a rebase shortly, and you can merge it afterwards.

pierremoreau:
I rebased the pull request and bumped the project version in the last patch, as it is the one modifying the entrypoints (only changing their capitalisation, but still).
I’ll open issues for 1) what to do with the mangler library, 2) add tests to test clang-format and clang-tidy on branches and pull requests.

pierremoreau:
Thanks for the reviews! I’ll try to address #14 ASAP.

PR #18, opened 2018-04-26
Run clang-format and clang-tidy on pull requests

pierremoreau:
Modifies the Travis configuration file to run clang-format and clang-tidy on pull requests.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -24,6 +24,7 @@ addons:
     - llvm-7-tools
     - llvm-7-dev
     - clang-format-7
+    - clang-tidy-7
 
 compiler:
   - gcc
@@ -33,6 +34,7 @@ env:
   global:
     - MAKEFLAGS="-j2"
     - CHECK_FORMAT=0
+    - CHECK_TIDY=0
   matrix:
     - BUILD_TYPE=Release BUILD_EXTERNAL=1 SHARED_LIBS=ON  MAKE_TARGETS=""           MAKE_TEST_TARGET="test"
     - BUILD_TYPE=Debug   BUILD_EXTERNAL=1 SHARED_LIBS=ON  MAKE_TARGETS=""           MAKE_TEST_TARGET="test"
@@ -55,6 +57,8 @@ matrix:
       osx_image: xcode9.3
 
     - env: BUILD_EXTERNAL=1 CHECK_FORMAT=1
+
+    - env: BUILD_EXTERNAL=1 CHECK_TIDY=1
   allow_failures:
     - compiler: clang
     - os: osx
@@ -79,6 +83,7 @@ script:
           -DLLVM_EXTERNAL_LIT="/usr/lib/llvm-7/build/utils/lit/lit.py" \
           -DLLVM_INCLUDE_TESTS=ON \
           -DCMAKE_INSTALL_PREFIX=../install/ \
+          -DCMAKE_EXPORT_COMPILE_COMMANDS=ON \
           -G "Unix Makefiles"
       fi
     else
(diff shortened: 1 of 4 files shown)


pierremoreau:
I have the following warning, but I haven’t been able to spot the issue and it still seems to be working fine, so if anyone has an idea why, I”d be interested:

```bash
$ if [ $CHECK_FORMAT == "ON" ]; then cd ..; ./utils/check_code_format.sh; elif [ $CHECK_TIDY == "ON" ]; then cd ..; ln -s build/compile_commands.json; ./utils/check_code_tidyness.sh; else make $MAKE_TARGETS && make $MAKE_TEST_TARGET && if [ $BUILD_EXTERNAL == "1" ]; then make install; fi fi
/home/travis/.travis/job_stages: line 57: [: ==: unary operator expected
```

.travis.yml
```diff
@@ -43,11 +47,20 @@ matrix:
     - os: osx
       env: BUILD_TYPE=Debug BUILD_EXTERNAL=0 MAKE_TARGETS="llvm-spirv" MAKE_TEST_TARGET="check-llvm-spirv"
       osx_image: xcode9.3
+
+    - env: BUILD_EXTERNAL=1 CHECK_FORMAT=ON
```
AlexeySotkin: I think it is better to use uniform syntax like `BUILD_EXTERNAL=1 CHECK_FORMAT=1` or `BUILD_EXTERNAL=ON CHECK_FORMAT=ON`

pierremoreau: Changed. I had kept the notation used by SPIRV-Tools

pierremoreau:
> I don't really like that we pull in files which might change or are just fetched from somewhere externally. I would prefer having a package we can install where those tools are or fetch some git repository or something.

And fetching from “some git repository” is not fetching “from somewhere externally”? :stuck_out_tongue: It was fetching from the official LLVM SVN repository for clang. I changed it to use the script installed by the “clang” package (at least on Arch Linux); I need to do some additional testing to check that it is also included in the Travis images, and in the same location.

pierremoreau:
Please have another look: I rebased the series and fixed the remaining bugs.

You can see the results when a patch fails either the clang-tidy or clang-format tests, [here](https://travis-ci.org/pierremoreau/SPIRV-LLVM-Translator/builds/380045234?utm_source=github_status&utm_medium=notification).

utils/clang-tidy-diff.py
```diff
@@ -0,0 +1,149 @@
...
+#===- clang-tidy-diff.py - ClangTidy Diff Checker ------------*- python -*--===#
+#
+#                     The LLVM Compiler Infrastructure
+#
+# This file is distributed under the University of Illinois Open Source
+# License. See LICENSE.TXT for details.
+#
+#===------------------------------------------------------------------------===#
+
+# Changes made to the initial version from
```
AlexeySotkin: Why do we need to modify the original script? I think we can use it as is.

pierremoreau: clang-tidy does not “run” on the header files (as they aren’t part of the .compile_commands.json file), but it will emit warning for headers 1) if the header file matches the `-header-filter` regex option and 2) a C/C++ source file including that header file is being compiled.

The problem with the original script, is that it runs only on the modified files, so if only a header file is modified, the original script will never tell you there could be an issue with your modifications. Hence why I modified it to have it run on all files, but only report warnings/errors on the files that were modified.

pierremoreau:
Friendly ping

AlexeySotkin:
Sorry for the late reply.

pierremoreau:
Thanks! Time to send in some more pull requests now. :smile:

PR #25, opened 2018-05-25
Rename DEBUG macro to LLVM_DEBUG.

AlexeySotkin:


karolherbst:
you were faster.. I was just creating the same commit.

Issue #9, opened 2018-04-15
Remove _SPIRV_LLVM_API


pierremoreau:
When setting `SPIRV_USE_LLVM_API` to `OFF`, the code does not compile due to some parts of the code not using that macro and always using the LLVM API (see further down). The code could be fixed to ensure that the entrypoints properly convert the LLVM streams to the std ones, if needed, but is this option really that useful/used?

```
/home/pmoreau/projects/llvm/tools/SPIRV-LLVM-Translator/lib/SPIRV/SPIRVWriter.cpp:1858:6: error: invalid operands to binary expression ('llvm::raw_ostream' and 'SPIRV::SPIRVModule')
  OS << *BM;
  ~~ ^  ~~~
/home/pmoreau/projects/llvm/include/llvm/Support/raw_ostream.h:201:16: note: candidate function not viable: no known conversion from 'SPIRV::SPIRVModule' to 'const void *' for 1st argument; take the address of the argument with &
  raw_ostream &operator<<(const void *P);
               ^
/home/pmoreau/projects/llvm/tools/SPIRV-LLVM-Translator/lib/SPIRV/libSPIRV/SPIRVModule.h:325:24: note: candidate function not viable: no known conversion from 'llvm::raw_ostream' to 'std::ostream &' (aka 'basic_ostream<char> &') for 1st argument
  friend spv_ostream & operator<<(spv_ostream &O, SPIRVModule& M);
                       ^
/home/pmoreau/projects/llvm/tools/SPIRV-LLVM-Translator/lib/SPIRV/libSPIRV/SPIRVStream.h:150:1: note: candidate function not viable: no known conversion from 'llvm::raw_ostream' to 'const SPIRV::SPIRVEncoder' for 1st argument
operator<<(const SPIRVEncoder& O, T V) {
^
/home/pmoreau/projects/llvm/include/llvm/ADT/APInt.h:2018:21: note: candidate function not viable: no known conversion from 'SPIRV::SPIRVModule' to 'const llvm::APInt' for 2nd argument
inline raw_ostream &operator<<(raw_ostream &OS, const APInt &I) {
                    ^
/home/pmoreau/projects/llvm/include/llvm/IR/Value.h:670:21: note: candidate function not viable: no known conversion from 'SPIRV::SPIRVModule' to 'const llvm::Value' for 2nd argument
inline raw_ostream &operator<<(raw_ostream &OS, const Value &V) {
                    ^
[...]

PR #33, opened 2018-06-09
Remove the _SPIRV_LLVM_API switch

pierremoreau:
The default remains the same, i.e. using the LLVM API, but removes the
option of using the std API, as it was untested and failed to compile.

Fixes https://github.com/KhronosGroup/SPIRV-LLVM-Translator/issues/9

### DIFF
--- a/include/LLVMSPIRVLib.h
+++ b/include/LLVMSPIRVLib.h
@@ -76,7 +76,7 @@ bool isSpirvBinary(std::string &Img);
 /// This function is not thread safe and should not be used in multi-thread
 /// applications unless guarded by a critical section.
 /// \returns true if succeeds.
-bool convertSpirv(std::istream &IS, llvm::raw_ostream &OS, std::string &ErrMsg,
+bool convertSpirv(std::istream &IS, std::ostream &OS, std::string &ErrMsg,
                   bool FromText, bool ToText);
 
 /// \brief Convert SPIR-V between binary and internel text formats.
@@ -95,7 +95,7 @@ namespace llvm {
 
 /// \brief Translate LLVM module to SPIRV and write to ostream.
 /// \returns true if succeeds.
-bool writeSpirv(Module *M, raw_ostream &OS, std::string &ErrMsg);
+bool writeSpirv(Module *M, std::ostream &OS, std::string &ErrMsg);
 
 /// \brief Load SPIRV from istream and translate to LLVM module.
 /// \returns true if succeeds.
@@ -157,7 +157,7 @@ ModulePass *createTransOCLMD();
 
 /// Create and return a pass that writes the module to the specified
 /// ostream.
-ModulePass *createSPIRVWriterPass(raw_ostream &Str);
+ModulePass *createSPIRVWriterPass(std::ostream &Str);
 
 } // namespace llvm
 
--- a/lib/SPIRV/CMakeLists.txt
+++ b/lib/SPIRV/CMakeLists.txt
@@ -1,8 +1,3 @@
-option(SPIRV_USE_LLVM_API "Enable usage of LLVM API for libSPIRV." ON)
-if(SPIRV_USE_LLVM_API)
-  add_definitions(-D_SPIRV_LLVM_API)
-endif(SPIRV_USE_LLVM_API)
-
 add_llvm_library(LLVMSPIRVLib
   LLVMToSPIRVDbgTran.cpp
   Mangler/FunctionDescriptor.cpp
--- a/lib/SPIRV/OCL20To12.cpp
+++ b/lib/SPIRV/OCL20To12.cpp
@@ -48,7 +48,6 @@
 #include "llvm/PassSupport.h"
 #include "llvm/Support/Debug.h"
 #include "llvm/Support/ErrorHandling.h"
-#include "llvm/Support/raw_ostream.h"
 
 using namespace llvm;
 using namespace SPIRV;
--- a/lib/SPIRV/OCL20ToSPIRV.cpp
+++ b/lib/SPIRV/OCL20ToSPIRV.cpp
@@ -51,7 +51,6 @@
 #include "llvm/Pass.h"
 #include "llvm/PassSupport.h"
 #include "llvm/Support/Debug.h"
-#include "llvm/Support/raw_ostream.h"
 
 #include <algorithm>
 #include <set>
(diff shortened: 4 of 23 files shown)


pierremoreau:
Should I replace all occurences of `spv_ostream` by `llvm::raw_ostream`, to get read of the `spv_otream` macro as it is now unecessary to have such a macro (unless we are still interested in easily authorising other stream APIs).

bader:
Pierre, sorry, I should have comment about this in #9 earlier.
If we are going to remove one of output stream API, I would remove LLVM one.
The idea behind libSPIRV is to isolate SPIR-V in-memory representation in a separate library, which is not supposed to have dependencies to LLVM. I think, this is the only link with LLVM and I see any reason to have it. What are the benefits of LLVM API vs standard stream?
We are also considering re-using SPIR-V parser and in-memory representation from [SPIRV-Tools](https://github.com/KhronosGroup/SPIRV-Tools), so isolating in-memory representation library from the code related to LLVM IR translation should make it easier.

pierremoreau:
Okay, writing the opposite patch should be easy :smiley:, though I’ll need to fix the interface of the library (writeSpirv & similar), as they explicitely use the LLVM API, or introduce a conversion from one to the other.

Regarding spv_ostream, should I keep the current macro definition, or get rid of the macro and replace every usage by the correct value?

bader:
@pierremoreau, I think if we leave two options we should test them both, otherwise it untested path will be eventually broken.
@AlexeySotkin, @yxsamliu, what do you think about this?

AlexeySotkin:
I agree with @bader , we should use `std::ostream` instead of `llvm::raw_ostream`.
For `spv_ostream` macro I'm ok both ways. If you decide to keep it I'd prefer to replace it by a `typedef`.

neildhickey:
This change is still using llvm::raw_ostream. It needs to be inverted before we can merge from the review comments above.

pierremoreau:
Yes, I didn’t have the opportunity to change it since the last comments. I’ll try to do that during the week, and send out some other fixes as well.

neildhickey:
Wasn't meaning to nag :) I just wanted to capture that we'd discussed merging during our discussions this week and this can't happen currently.

pierremoreau:
I finally got around to updating that patch. I switched back to WIP as I haven’t tested it yet (I’ll do so tomorrow).

I switched all the `llvm::raw_ostream` to `std::ostream`, even for the API (e.g. `llvm::writeSpirv()`. Is that something you agree with or should we keep the `llvm::ostream` there and convert it internally to `std::ostream`?
If we do change the API, we should bump the version of the library to reflect that.

pierremoreau:
Rebased on master and fixed the naming error from `clang-tidy`.

pierremoreau:
I rebased on top of the latest _master_; please review.

Also, feedback would be very welcome on the following:
> I switched all the `llvm::raw_ostream` to `std::ostream`, even for the API (e.g. `llvm::writeSpirv()`. Is that something you agree with or should we keep the `llvm::ostream` there and convert it internally to `std::ostream`?
> If we do change the API, we should bump the version of the library to reflect that.

pierremoreau:
Thanks @svenvh for the approval.
Any inputs regarding the version bump? Since we are changing the API of SPIRV-LLVM-Translator, the version should probably be changed to reflect that.

AlexeySotkin:
> 
> 
> Thanks @svenvh for the approval.
> Any inputs regarding the version bump? Since we are changing the API of SPIRV-LLVM-Translator, the version should probably be changed to reflect that.

What version do you mean ? Is it the magic number mentioned [here ](https://www.khronos.org/registry/spir-v/specs/unified1/SPIRV.html#_a_id_physicallayout_a_physical_layout_of_a_spir_v_module_and_instruction) , or output from `llvm-spirv --version` ?

pierremoreau:
> > Thanks @svenvh for the approval.
> > Any inputs regarding the version bump? Since we are changing the API of SPIRV-LLVM-Translator, the version should probably be changed to reflect that.
> 
> What version do you mean ? Is it the magic number mentioned [here ](https://www.khronos.org/registry/spir-v/specs/unified1/SPIRV.html#_a_id_physicallayout_a_physical_layout_of_a_spir_v_module_and_instruction) , or output from `llvm-spirv --version` , anything else ?

I was thinking of the version exposed via pkg-config, that other projects will interact with when they import this project using `find_package()` in CMake or similar. If the interface changes, you want a way in your application to either support both the old and new interfaces, or require a version of this project that is known to have the old (or new) interface.

AlexeySotkin:
> 
> 
> > > Thanks @svenvh for the approval.
> > > Any inputs regarding the version bump? Since we are changing the API of SPIRV-LLVM-Translator, the version should probably be changed to reflect that.
> > 
> > 
> > What version do you mean ? Is it the magic number mentioned [here ](https://www.khronos.org/registry/spir-v/specs/unified1/SPIRV.html#_a_id_physicallayout_a_physical_layout_of_a_spir_v_module_and_instruction) , or output from `llvm-spirv --version` , anything else ?
> 
> I was thinking of the version exposed via pkg-config, that other projects will interact with when they import this project using `find_package()` in CMake or similar. If the interface changes, you want a way in your application to either support both the old and new interfaces, or require a version of this project that is known to have the old (or new) interface.
> 
> Edit: Otherwise, you have no way to ensure that your project that uses SPIRV-LLVM-Translator will be able to compile.
> Edit x2:I am talking about [this version](https://github.com/KhronosGroup/SPIRV-LLVM-Translator/blob/master/CMakeLists.txt#L3), which should probably be updated to match the version used in the release tags.

I see. Basically I don't mind to change this version. Currently it is 0.2.1.0. To match to the version in the release tags it should be bumped to 9.0.0.1 on the master branch and to 8.0.0.1 on the llvm_release_80 branch, right ?

pierremoreau:
> I see. Basically I don't mind to change this version. Currently it is 0.2.1.0. To match to the version in the release tags it should be bumped to 9.0.0.1 on the master branch and to 8.0.0.1 on the llvm_release_80 branch, right ?

Yes, I think so; and that could be done after this series is merged, so if someone uses the project, they know that if they ask for 9.0.0.1, they will get the new interface.

pierremoreau:
Do you think this should be back-ported to the 8.0 and 7.0 branches as well?

AlexeySotkin:
No. I think we should do that for not yet released versions only (master and 8.0 branches). On branches, for which we already have "release" for some time, such change may be unexpected by the users.

pierremoreau:
Sounds good.

AlexeySotkin:
@pierremoreau would you like to bump the version with this commit, or we can do that later ?

pierremoreau:
I don’t mind leaving the bump to a later commit, as long as we do it soon (i.e. within a week or two). By the way, I opened #131 yesterday to discuss on how the different version numbers might/should be unified.

pierremoreau:
Thanks @AlexeySotkin! :slightly_smiling_face:

PR #37, opened 2018-07-02
Add support of OpDecorationGroup, OpGroupDecorate, OpGroupMemberDecorate

linqunzju:
This change is port from AMDVLK
1. Decode GroupDecorate and GroupMemberDecorate separately
2. Add function hasMemberDecorate in SPIRVEntry, it will be used in the future.
3. Add const modifier back in the base class of SPIRVDecorateSet, and modify related functions
4. Remove incorrect assert in addDecorationGroup


linqunzju:
close it, since some tests are fail.

PR #38, opened 2018-07-05
Remove transFlags in SPIRVToLLVM

linqunzju:
transFlags sets the "exact" flag in LLVM-IR for AShr and LShr
instructions. This is in fact incorrect as this means that LLVM will do some
optimisations that are potentially illegal.

Setting the exact flag says that there are no set bits in the bits being shifted
off the end during the shift - and thus shift combining is safe to do.

In a situation like this ((X >> 5) << 2) it will be transformed into (X >> 3)
which is incorrect as the 2 LSBs may contain set bits.

This change is ported from AMDVLK


AlexeySotkin:
@linqunzju, please add a LIT test.

AlexeySotkin:
LGTM, thanks!

PR #40, opened 2018-08-06
Changes in upstream LLVM changed the behaviour of out of tree builds linking dynamically

neildhickey:
Trying to get travis working again

### DIFF
--- a/tools/llvm-spirv/CMakeLists.txt
+++ b/tools/llvm-spirv/CMakeLists.txt
@@ -14,6 +14,8 @@ add_llvm_tool(llvm-spirv
   NO_INSTALL_RPATH
 )
 
+target_link_libraries(llvm-spirv PRIVATE LLVMSPIRVLib)
+
 target_include_directories(llvm-spirv
   PRIVATE
     ${LLVM_INCLUDE_DIR}


tools/llvm-spirv/CMakeLists.txt
```diff
@@ -14,6 +14,8 @@ add_llvm_tool(llvm-spirv
   NO_INSTALL_RPATH
 )
 
+target_link_libraries(llvm-spirv PRIVATE LLVMSPIRVLib)
```
bader: Unfortunately this fix breaks the build within LLVM source tree if we link with LLVM libraries dynamically (cmake -G "Unix Makefiles" -DLLVM_LINK_LLVM_DYLIB=ON).
The build itself will pass, but resulting llvm-spirv tool is broken.
The problem is that LLVMSPIRVLib library is linked with llvm-spirv twice:
 - statically here
 - in addition to that symbols from this library are loaded dynamically from into libLLVM.so (which is linked with LLVMSPIRLib because it uses add_llvm_library cmake function).
When I try to run llvm-spirv tool, I following error message:

> : CommandLine Error: Option 'spirv-expand-step' registered more than once!
> LLVM ERROR: inconsistency in registered CommandLine options
> 

@neildhickey, any suggestions how to fix this case?

neildhickey: Sorry @bader , didn't see this, will take a look

PR #41, opened 2018-08-07
Start passing shader type to SPIR-V read function

neildhickey:
This is the first patch from the Arm-Imperial project to extend the translator with Vulkan support.

### DIFF
--- a/include/LLVMSPIRVLib.h
+++ b/include/LLVMSPIRVLib.h
@@ -44,6 +44,8 @@
 #include <iostream>
 #include <string>
 
+#include "../lib/SPIRV/libSPIRV/spirv.hpp"
+
 namespace llvm {
 // Pass initialization functions need to be declared before inclusion of
 // PassSupport.h.
@@ -99,7 +101,7 @@ bool writeSpirv(llvm::Module *M, llvm::raw_ostream &OS, std::string &ErrMsg);
 /// \brief Load SPIRV from istream and translate to LLVM module.
 /// \returns true if succeeds.
 bool readSpirv(llvm::LLVMContext &C, std::istream &IS, llvm::Module *&M,
-               std::string &ErrMsg);
+               std::string &ErrMsg, spv::ExecutionModel ExecModel, std::string &EntryPoint);
 
 /// \brief Regularize LLVM module by removing entities not representable by
 /// SPIRV.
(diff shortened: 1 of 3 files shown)


karolherbst:
General question: if you want llvm-spirv to accept glsl shader files, why not using glslang to produce SPIR-V instead?

karolherbst:
Or do you want to pass in spirv with different extensions indicating their execution model? The SPIR-V should already contain all the information and we should be able to tell alone from that already. Normally you use those file extensions for the text glsl sources and it might become quite confusing as people would need to start doing things like some_shader.spv.geom or something like that.

pierremoreau:
Agreed with Karol's general question. I thought at first you wanted to still consume OpenCL C/C++ but generate Vulkan SPIR-V instead.

neildhickey:
This project is to extend the SPIR-V translator to accept Vukan SPIR-V and convert to LLVM and vice versa.

neildhickey:
I agree with you though that we should have a single file extension and extract the executionmode from the spirv directly

pierremoreau:
Also, you could end up with the whole pipeline within a single SPIR-V binary (like I like to do), in which case you really need to extract it from the SPIR-V.

trenouf:
(I am an AMD employee.)

It is true that the AMD Vulkan version of the spir-v reader (in the https://github.com/GPUOpen-Drivers/llpc component of the AMDVLK open source driver) also passes the required entry name and execution model into readSpirv(). It then goes on to find that entry point and mark it as dllexport so the rest of the compiler can find it.

It is arguable that that approach (and equally this patch) is wrong. The SPIRVRepresentationInLLVM.rst document quotes a goal of spir-v to be representable in other IRs, especially LLVM. That possibly means that a goal of the spir-v reader is to translate the whole spir-v to llvm IR and then let someone else process the llvm IR itself, including finding entry points.

So, instead (the argument goes), we should represent the name and execution model of each entry point, and then the user of the spir-v reader can use that to find the entry point it wants. Thus this api change to readSpirv() is unnecessary.

As @pierremoreau says, you could put all your pipeline shaders in one spir-v module. What's more, an entry point can have a different name to its function, and two entry points can have the same name as long as their execution models differ.

neildhickey:
Thanks @trenouf I agree, I plan to rework/ abandon the patch to allow all execution models to be parsed in a single file.

PR #47, opened 2018-08-13
Update translation of pipe types and built-ins

AlexeySotkin:
Starting from version 3.8 clang generates unmangled names for pipe built-ins.
Plus reflect changes in LLVM IR introduced by https://reviews.llvm.org/D46015.
There are two separate types for read only and write only pipes:
opencl.pipe_ro_t and opencl.pipe_wo_t instead of opencl.pipe_t.

Changes:
* Update SPIRVWriter to consume LLVM IR with non-mangled pipe built-ins.

* Align SPIRV Translator to consume and produce new pipe types. Old ones are no
longer supported.

* Change translation flow to not to convert new pipe types to
spirv-friendly format (like spirv.Pipe._0 or spirv.Pipe._1) before translation.
Spirv-friendly format for pipes is still supported for consuming or producing.


lib/SPIRV/Mangler/ManglingUtils.cpp
```diff
@@ -138,6 +140,7 @@ static const SPIRversion PrimitiveSupportedVersions[PRIMITIVE_NUM] = {
     SPIR12, // PRIMITIVE_IMAGE_2D_ARRAY_DEPTH_T
     SPIR12, // PRIMITIVE_EVENT_T
     SPIR20, // PRIMITIVE_PIPE_T
```
neildhickey: Shouldn't these be RO and WO respectively

neildhickey: In the comment I mean

AlexeySotkin: Fixed

PR #48, opened 2018-08-14
Generate OpControlBarrier for Subgroup scope for OpenCL 1.2/2.0

AlexeySotkin:
'cl_intel_subgroup' extension allows to use sub_group_barrier() built-in
with OpenCL 1.2/2.0. It requires to translate this built-in to
reqular (core) SPIR-V instruction OpControlBarrier.

Note: currently clag generates nothing representing vendor extensions
like 'cl_intel_subgroup'. So, SPIR-V translator doesn't check that the
extensions was enabled. It assumes that if a built-in or a type which
belong to one of supported extension used in the IR, this extensions is
enabled.

### DIFF
--- a/test/transcoding/OpControlBarrier_cl20_subgroup.ll
+++ b/test/transcoding/OpControlBarrier_cl20_subgroup.ll
@@ -1,0 +1,122 @@
+; RUN: llvm-as %s -o %t.bc
+; RUN: llvm-spirv %t.bc -spirv-text -o %t.txt
+; RUN: FileCheck < %t.txt %s --check-prefix=CHECK-SPIRV
+; RUN: llvm-spirv %t.bc -o %t.spv
+; RUN: llvm-spirv -r %t.spv -o %t.rev.bc
+; RUN: llvm-dis < %t.rev.bc | FileCheck %s --check-prefix=CHECK-LLVM
+
+; CHECK-LLVM: call spir_func void @_Z17sub_group_barrierji(i32 2, i32 1) #{{[0-9]+}}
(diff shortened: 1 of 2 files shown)


test/transcoding/OpControlBarrier_cl20_subgroup.ll
```diff
@@ -0,0 +1,122 @@
...
+
+declare spir_func void @_Z17sub_group_barrierj(i32) #1
+declare spir_func void @_Z17sub_group_barrierji(i32, i32) #1
+
+attributes #0 = { nounwind "less-precise-fpmad"="false" "no-frame-pointer-elim"="false" "no-infs-fp-math"="false" "no-nans-fp-math"="false" "no-realign-stack" "stack-protector-buffer-size"="8" "unsafe-fp-math"="false" "use-soft-float"="false" }
+attributes #1 = { "less-precise-fpmad"="false" "no-frame-pointer-elim"="false" "no-infs-fp-math"="false" "no-nans-fp-math"="false" "no-realign-stack" "stack-protector-buffer-size"="8" "unsafe-fp-math"="false" "use-soft-float"="false" }
+attributes #2 = { nounwind }
+
+!opencl.kernels = !{!0}
+!rpencl.enable.FP_CONTRACT = !{}
```
neildhickey: Why is this piece of metadata called rpencl? Does this come from clang?

neildhickey: clang generates opencl so I'm guessing this is simply a cut and paste error. I'll change it.

PR #49, opened 2018-08-14
Update translation of sampler type

AlexeySotkin:
Fix translation of sampler initialization function
Update test for sampled image with the current version of LLVM IR


AlexeySotkin:
@neildhickey , I think we should keep the history linear whenever possible. So, could you [merge PRs](https://help.github.com/articles/merging-a-pull-request/#merging-a-pull-request-on-github) using "rebase and merge" option instead of "create a merge commit".  Thanks!

PR #55, opened 2018-08-15
Make generator's magic number configurable at build time

AlexeySotkin:
To set high order 16 bits of generator magic number one should use
-DSPIRVGEN_ID=<ID> cmake option. And to set low order 16 bits use
-DSPIRVGEN_VER=<VER> cmake option.
Values for SPIRVGEN_ID are enumerated at
https://github.com/KhronosGroup/SPIRV-Headers/blob/master/include/spirv/spir-v.xml

By default: SPIRVGEN_ID=6("Khronos LLVM/SPIR-V Translator"), SPIRVGEN_VER=14

### DIFF
--- a/lib/SPIRV/libSPIRV/SPIRVEnum.h
+++ b/lib/SPIRV/libSPIRV/SPIRVEnum.h
@@ -71,6 +71,18 @@ enum SPIRVGeneratorKind {
   SPIRVGEN_KhronosSPIRVAssembler = 7,
 };
 
+#ifndef SPIRVGEN_ID
+#define SPIRVGEN_ID SPIRVGEN_KhronosLLVMSPIRVTranslator
+#endif
+
+/// The LLVM/SPIR-V translator version used to fill the lower 16 bits of the
+/// generator's magic number in the generated SPIR-V module.
+/// This number should be bumped up whenever format of the generated SPIR-V
+/// is changing (significantly).
+#ifndef SPIRVGEN_VER
+#define SPIRVGEN_VER 14
+#endif
+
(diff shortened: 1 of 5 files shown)


neildhickey:
Is there a way we could construct a unit test for this feature?

AlexeySotkin:
No. I don't think there is an easy way to do that.

bader:
> No. I don't think there is an easy way to do that.

Can we dump the text form of the SPIR-V generated with existing tool and check that it provides the default versions?

AlexeySachkov:
@AlexeySotkin, is there any updates? Do you still plan to finish this PR?

lib/SPIRV/libSPIRV/SPIRVEnum.h
```diff
@@ -71,6 +71,18 @@ enum SPIRVGeneratorKind {
...
 };
 
+#ifndef SPIRVGEN_ID
+#define SPIRVGEN_ID SPIRVGEN_KhronosLLVMSPIRVTranslator
+#endif
+
+/// The LLVM/SPIR-V translator version used to fill the lower 16 bits of the
+/// generator's magic number in the generated SPIR-V module.
+/// This number should be bumped up whenever format of the generated SPIR-V
+/// is changing (significantly).
```
AlexeySachkov, line 81: Having in mind that the SPIR-V output of the translator must correspond to the spec, I wonder what can be _significantly_ changed there that we need to update magic number?

AlexeySachkov:
This PR seems to stuck without any updates for almost 2 years already. I will close it for now. @AlexeySotkin, please update it and re-open if this is still needed

PR #61, opened 2018-09-21
Disabling generation of invalid atomic instructions

vmaksimo:
Some atomic instructions allow only integer result type

### DIFF
--- a/lib/SPIRV/OCL20ToSPIRV.cpp
+++ b/lib/SPIRV/OCL20ToSPIRV.cpp
@@ -400,7 +400,7 @@ void OCL20ToSPIRV::visitCallInst(CallInst &CI) {
   if (DemangledName.find(kOCLBuiltinName::AtomicPrefix) == 0 ||
       DemangledName.find(kOCLBuiltinName::AtomPrefix) == 0) {
 
-    // Compute atomic bulitins do not support floating types.
+    // Compute atomic builtins do not support floating types.
     if (CI.getType()->isFloatingPointTy() &&
         isComputeAtomicOCLBuiltin(DemangledName))
       return;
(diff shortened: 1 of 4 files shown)


lib/SPIRV/OCL20ToSPIRV.cpp
```diff
@@ -399,6 +399,12 @@ void OCL20ToSPIRV::visitCallInst(CallInst &CI) {
   }
   if (DemangledName.find(kOCLBuiltinName::AtomicPrefix) == 0 ||
       DemangledName.find(kOCLBuiltinName::AtomPrefix) == 0) {
+
+    // Compute atomic bulitins do not support floating types.
```
Hopetech: Small nit:
s/bulitins/builtins/g

Hopetech:
Should we need to add the same security in OCL21ToSPIRV too?

lib/SPIRV/OCLUtil.cpp
```diff
@@ -107,6 +107,32 @@ size_t getAtomicBuiltinNumMemoryOrderArgs(StringRef Name) {
...
 }
 
+bool isComputeAtomicOCLBuiltin(StringRef DemangledName) {
+  if (!DemangledName.startswith(kOCLBuiltinName::AtomicPrefix) &&
+      !DemangledName.startswith(kOCLBuiltinName::AtomPrefix))
+    return false;
+
+  return llvm::StringSwitch<bool>(DemangledName)
+      .EndsWith("add", true)
+      .EndsWith("sub", true)
```
Hopetech: It seems that we missed
`.EndsWith("cmpxchg", true)`

Hopetech:
> > Should we need to add the same security in OCL21ToSPIRV too?
> 
> I don't so. OCL21ToSPIRV processes OpenCL 21 specific built-in functions and AFAIK OpenCL 21 does not add new atomic built-in functions, so it should enough to handle the builtins processed in OCL20ToSPIRV.

Ack. Sorry for the noise.

lib/SPIRV/OCLUtil.cpp
AlexeySotkin: xchg allows other then int result types. i.e. float. 
Also I think xchg doesn't do any computation, just access memory. So we don't need it here.

PR #69, opened 2018-10-12
Translation of loop unrolling metadata

vmaksimo:
loop-simplify pass is added to simplify search of Continue Target operand of
OpLoopMerge instruction


test/SPIRV/transcoding/LoopUnroll.ll
```diff
@@ -0,0 +1,204 @@
+; Source:
```
AlexeySotkin: Please move this file to test/transcoding/LoopUnroll.ll

PR #71, opened 2018-10-18
Fix lifetime intrinsics translation.

bader:
LLVM lifetime intrinsics accept only i8* pointers + size as object
descriptors, so typical pattern for local variables is
alloca+bitcast+lifetime instructions.
SPIR-V translator tries store original type into lifetime instruction
looking though bitcast. It also removes bitcast from the SPIR-V file and
tries to recover it during SPIR-V conversion.
This patch simplifies the translation by keeping bitcast instruction and
fixes assertion of lifetime intrinsics for 8-bit objects.

### DIFF
--- a/lib/SPIRV/SPIRVWriter.cpp
+++ b/lib/SPIRV/SPIRVWriter.cpp
@@ -261,8 +261,6 @@ private:
 
   SPIRVInstruction *transBinaryInst(BinaryOperator *B, SPIRVBasicBlock *BB);
   SPIRVInstruction *transCmpInst(CmpInst *Cmp, SPIRVBasicBlock *BB);
-  SPIRVInstruction *transLifetimeIntrinsicInst(Op OC, IntrinsicInst *Intrinsic,
-                                               SPIRVBasicBlock *BB);
 
   void dumpUsers(Value *V);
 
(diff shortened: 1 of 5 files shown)


test/lifetime.ll
```diff
@@ -5,13 +5,12 @@
...
-; CHECK-SPIRV-NOT: Bitcast
-; CHECK-SPIRV: 3 LifetimeStart [[tmp:[0-9]+]] 0
-; CHECK-SPIRV: 3 LifetimeStop [[tmp]] 0
+; CHECK-SPIRV: 3 LifetimeStart [[tmp:[0-9]+]] 4294967295
+; CHECK-SPIRV: 3 LifetimeStop [[tmp]] 4294967295
 
 ; CHECK-LLVM: %[[tmp1:[0-9]+]] = bitcast i32* %{{[0-9]+}} to i8*
-; CHECK-LLVM: call void @llvm.lifetime.start.p0i8(i64 -1, i8* %[[tmp1]])
-; CHECK-LLVM: call void @llvm.lifetime.end.p0i8(i64 -1, i8* %[[tmp1]])
+; CHECK-LLVM: call void @llvm.lifetime.start.p0i8(i64 4294967295, i8* %[[tmp1]])
```
AlexeySotkin: Why 4294967295 instead of -1 ?

lib/SPIRV/libSPIRV/SPIRVInstruction.h
```diff
@@ -1861,11 +1861,13 @@ template <Op OC> class SPIRVLifetime : public SPIRVInstruction {
...
+    assert(ObjType->isTypePointer() && "Objects type must be a pointer");
+    assert(static_cast<SPIRVTypePointer *>(ObjType)->getStorageClass() ==
+               StorageClassFunction &&
            "Invalid storage class");
-    assert(Obj->getType()->isTypePointer() && "Objects type must be a pointer");
-    if (!Obj->getType()->getPointerElementType()->isTypeVoid() ||
+    if (!(ObjType->getPointerElementType()->isTypeVoid() ||
+          ObjType->getPointerElementType()->isTypeInt(8)) ||
         !Module->hasCapability(CapabilityAddresses))
       assert(Size == 0 && "Size must be 0");
```
AlexeySotkin: I think it's worth quoting the spec to explain these conditions and assert.

test/lifetime.ll
bader: Added comments which should resolve both questions.

test/lifetime.ll
```diff
@@ -5,13 +5,12 @@
 ; RUN: llvm-spirv -r %t.spv -o %t.spv.bc
 ; RUN: llvm-dis < %t.spv.bc | FileCheck %s --check-prefix=CHECK-LLVM
 
-; CHECK-SPIRV-NOT: Bitcast
-; CHECK-SPIRV: 3 LifetimeStart [[tmp:[0-9]+]] 0
-; CHECK-SPIRV: 3 LifetimeStop [[tmp]] 0
+; CHECK-SPIRV: 3 LifetimeStart [[tmp:[0-9]+]] 4294967295
```
AlexeySotkin: I think we should map size=-1 from LLVM to size=0 in SPIR-V.

test/lifetime.ll
AlexeySotkin: When we create the intrinsic during translation from SPIR-V to LLVM and the size is zero I think we can omit size parameter, i.e use default parameter value of CreateLifetimeStart function.

lib/SPIRV/SPIRVWriter.cpp
```diff
@@ -1299,9 +1259,13 @@ SPIRVValue *LLVMToSPIRV::transIntrinsicInst(IntrinsicInst *II,
         transValue(II->getOperand(2), BB),
         GetMemoryAccess(cast<MemIntrinsic>(II)), BB);
   case Intrinsic::lifetime_start:
-    return transLifetimeIntrinsicInst(OpLifetimeStart, II, BB);
+    return BM->addLifetimeInst(
+        OpLifetimeStart, transValue(II->getOperand(1), BB),
+        dyn_cast<ConstantInt>(II->getOperand(0))->getSExtValue(), BB);
```
AlexeySotkin: Potential problem might be here because LLVM reserves 64 bit for the size while in SPIR-V it only 32 bit.

bader: Another problem is that LLVM IR value is *signed*, whereas SPIR-V is *unsigned*.

PR #74, opened 2018-10-23
Initial commit to prepare SPIRV-LLVM Translator to accept Vulkan

neildhickey:
Initial refactoring of the code to accept vulkan shaders.

### DIFF
--- a/lib/SPIRV/libSPIRV/spirv.hpp
+++ b/lib/SPIRV/libSPIRV/spirv.hpp
@@ -46,7 +46,7 @@ namespace spv {
 
 typedef unsigned int Id;
 
-#define SPV_VERSION 0x10100
+#define SPV_VERSION 0x10300
 #define SPV_REVISION 7
 
 static const unsigned int MagicNumber = 0x07230203;
--- a/lib/SPIRV/SPIRVReader.cpp
+++ b/lib/SPIRV/SPIRVReader.cpp
@@ -2899,7 +2899,11 @@ Instruction *SPIRVToLLVM::transOCLBarrierFence(SPIRVInstruction *MB,
 bool SPIRVToLLVM::transSourceLanguage() {
   SPIRVWord Ver = 0;
   SourceLanguage Lang = BM->getSourceLanguage(&Ver);
-  assert((Lang == SourceLanguageOpenCL_C || Lang == SourceLanguageOpenCL_CPP) &&
+  assert((Lang == SourceLanguageUnknown ||
+          Lang == SourceLanguageOpenCL_C ||
+          Lang == SourceLanguageOpenCL_CPP ||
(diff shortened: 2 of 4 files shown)


lib/SPIRV/libSPIRV/spirv.hpp
```diff
@@ -46,7 +46,7 @@ namespace spv {
 
 typedef unsigned int Id;
 
-#define SPV_VERSION 0x10100
+#define SPV_VERSION 0x10300
 #define SPV_REVISION 7
```
AlexeySotkin: I think we don't use SPV_REVISION at all. Can we remove it ?

AlexeySotkin:
It would be great if the PR was checked by Travis CI

Naghasan:
FYI, you can't reliably know what is the source language from a SPIR-V module (see https://www.khronos.org/registry/spir-v/specs/unified1/SPIRV.html#OpSource)

> This has no semantic impact and can safely be removed from a module.

As there is some code path that rely on that instruction, this is not really SPIR-V legal.

thewilsonator:
Yeah, I have a frontend using this library and I currently fake as being OpenCL_C, I would love to also be able to target Vulkan as well. I suppose I could fake being GLSL or ESSL (btw this is missing HLSL) in a different module but I think this should be user controllable at the level of invocation of the pubic API interface.

AlexeySachkov:
@neildhickey, do you have any updates for this PR? Do you still plan to merge this change?

VarLad:
Amy update on this?

fknfilewalker:
Why is this project called SPIRV-LLVM if all the shader(vulkan) related issues and pull requests are ignored?

AlexeySachkov:
Hi @fknfilewalker,

> Why is this project called SPIRV-LLVM

Because this project provides a bi-directional SPIR-V - LLVM translator

> if all the shader(vulkan) related issues and pull requests are ignored?

SPIR-V != Vulcan. SPIR-V also supports compute flavor, which is actively used in OpenCL and SYCL. It just happens that currently active project maintainers are only (or mostly) interested in compute flavor of SPIR-V and therefore do not work on supporting Vulcan shaders here. Contributions are always welcome, but so far I haven't seen a PR whose author would be interested enough to get any Vulcan-related changes merged (I mean look at this PR: it hasn't been updated for years - that is not something maintainers are accountable for)

keryell:
@AnastasiaStulova perhaps something to investigate again during the next F2F for example to have the big picture behind the scene with the Vulkan committee?

largeword:
> llvm-dis

Hi @thewilsonator, did you succeed in using such pipeline? Have you met any difficulties?

thewilsonator:
I mean it worked, but I've not touched it for a very long time.
I was also quite a pain to actually use, as I never made a build system rule for it.

PR #75, opened 2018-10-23
This patch introduces the agreed mapping from SPIRV StorageClasses to…

neildhickey:
… llvm address spaces

### DIFF
--- a/lib/SPIRV/SPIRVInternal.h
+++ b/lib/SPIRV/SPIRVInternal.h
@@ -178,7 +178,13 @@ enum SPIRAddressSpace {
   SPIRAS_Constant,
   SPIRAS_Local,
   SPIRAS_Generic,
-  SPIRAS_Count,
+  SPIRAS_Input,
+  SPIRAS_Output,
+  SPIRAS_PushConstant,
+  SPIRAS_AtomicCounter,
+  SPIRAS_Image,
+  SPIRAS_StorageBuffer,
+  SPIRAS_Uniform
 };
 
 template <> inline void SPIRVMap<SPIRAddressSpace, std::string>::init() {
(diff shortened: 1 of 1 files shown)


lib/SPIRV/SPIRVInternal.h
```diff
@@ -178,7 +178,13 @@ enum SPIRAddressSpace {
...
   SPIRAS_Generic,
-  SPIRAS_Count,
+  SPIRAS_Input,
+  SPIRAS_Output,
+  SPIRAS_PushConstant,
+  SPIRAS_AtomicCounter,
+  SPIRAS_Image,
+  SPIRAS_StorageBuffer,
+  SPIRAS_Uniform
 };
```
AlexeySotkin: Please do not remove SPIRAS_Count

lib/SPIRV/SPIRVInternal.h
```diff
@@ -194,10 +200,19 @@ typedef SPIRVMap<SPIRAddressSpace, SPIRVStorageClassKind>
 template <>
 inline void SPIRVMap<SPIRAddressSpace, SPIRVStorageClassKind>::init() {
   add(SPIRAS_Private, StorageClassFunction);
+  add(SPIRAS_Private, StorageClassPrivate);
```
AlexeySotkin: This is not clear to me. What SPIRAS_Private supposed to be mapped to StorageClassFunction or StorageClassPrivate ? Technically the later add() call overwrites former ones.
I don't think OpenCL private address space can be mapped to StorageClassPrivate in SPIR-V, they are not the same. Maybe we could add some other kind of private in the SPIRAddressSpace enum above, like "SPIRAS_PrivateGlobal" or "SPIRAS_PrivateShader", or whatever you think is suitable.

AlexeySachkov:
@neildhickey, do you have any updates for this PR? Do you still plan to merge this change?

PR #76, opened 2018-10-24
Enable publishing of built artifacts via github releases.

AlexeySotkin:
This should upload the latest version of the translator to the github repo.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -108,3 +108,30 @@ script:
     else
       make $MAKE_TARGETS && make $MAKE_TEST_TARGET && if [ $BUILD_EXTERNAL == "1" ]; then make install; fi
     fi
+
+after_success:
+  # Create tarball for deployment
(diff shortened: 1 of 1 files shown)


thewilsonator:
Is anything holding this up? This would be very useful.

AlexeySotkin:
The release archive to be published is over 400MB. It includes the whole build directory. We need to update the patch or travis config.

AlexeySotkin:
I think, ideally, it should be a .deb package. Though, this doesn't mean we can't have zip-packages too.

pierremoreau:
Should it really create a new tag (and release) every time there is a new commit (if I understand properly the changes from this PR)?
Maybe there could be a single release/tag whose artifacts get overwritten every time a new commit is pushed, and separate releases/tags for each actual version of the project.

AlexeySotkin:
You are right. I have fixed that.

PR #79, opened 2018-10-30
Make it possible to keep typedef'ed type names in SPIRV

AlexeySotkin:
In LLVM IR information about typedef'ed type name stored only in
kernel_arg_type metadata, but it is missed while translating in SPIRV.
There is no possibility to store typedef's in SPIRV.

As a workaround we store information about type names from
kernel_arg_type metadata in an OpString instruction and restore it
while translating from SPIRV to LLVM.

### DIFF
--- a/lib/SPIRV/SPIRVReader.cpp
+++ b/lib/SPIRV/SPIRVReader.cpp
@@ -2207,6 +2207,35 @@ bool SPIRVToLLVM::transNonTemporalMetadata(Instruction *I) {
   return true;
 }
 
+// Information of types of kernel arguments may be additionally stored in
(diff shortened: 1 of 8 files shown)


AlexeySotkin:
I would like to commit this change by the end of the week, unless there are objections.
Note that this is rather a workaround, and possibly will be reverted in the future.

AlexeySotkin:
As it was discussed on the call, we might use debug info instruction to carry this information in SPIRV. But that would bring in a significant amount of debug info instructions and increase size of the module. Also I think implementation of debug info approach will be complex, because translator is designed to generate debug info from corresponding debug info metadata, which are not always available.
So, since there is no other comments, I'm going to merge it now. We can revise this change later if  needed.

PR #81, opened 2018-11-19
Some fixes I didn't tried to upstream yet

karolherbst:
some minor fixes.


AlexeySotkin:
LGTM.
I think the failures in travis-ci are caused by outdated (Nov 11) llvm packages used for "external" builds.

karolherbst:
yeah, the packages are still at revision 346579 for whatever reason. Maybe the package task doesn't build as nightly as the name suggests?

karolherbst:
nvm, that is just trusty which is outdated *sigh*

PR #86, opened 2018-11-22
travis: move to xenial

karolherbst:
With xenial we get more up to date nightly packages.

UPDATE: seems like it takes some minutes until we get workers, but seems good enough for now


karolherbst:
seems like on xenial the external builds with clang are passing now as well. Probably this was caused by some stupid runtime issue.

PR #91, opened 2018-11-28
Fix unresolved temporary MDNode processing function declarations

sergey-semenov:


lib/SPIRV/SPIRVToLLVMDbgTran.cpp
```diff
@@ -528,9 +528,12 @@ DINode *SPIRVToLLVMDbgTran::transFunctionDecl(const SPIRVExtInst *DebugInst) {
   if (isa<DICompositeType>(Scope) || isa<DINamespace>(Scope))
     DIS = Builder.createMethod(Scope, Name, LinkageName, File, LineNo, Ty, 0, 0,
                                nullptr, Flags, SPFlags);
-  else
-    DIS = Builder.createFunction(Scope, Name, LinkageName, File, LineNo, Ty, 0,
-                                 Flags, SPFlags);
+  else {
```
AlexeySotkin: @sergey-semenov, could you add a comment explaining why we need to go through a temporary node.

PR #93, opened 2018-12-04
Do not include image access qualifier in kernel_arg metadata

svenvh:
According to OpenCL 2.0 Table 5.21:

  The type name returned will be the argument type name as it was
  declared with any whitespace removed. [...] The argument type
  name returned does not include any type qualifiers.


AlexeySotkin:
LGTM, Thanks!

PR #94, opened 2018-12-07
Fix image type name recognition

svenvh:
Image type names may end up getting a .<n> suffix, e.g. when multiple
LLVM IR modules defining the same type name are loaded into an
LLVMContext.  The translator already has some basic handling for that,
but the included test case triggers an assert.

Improve the type name recognition handling and factor it out into a
helper function.


AlexeySotkin:
LGTM, Thanks!

PR #106, opened 2019-01-02
Fix error path in transValueWithoutDecoration

svenvh:
The error path in transValueWithoutDecoration was inside the default
case of a switch statement.  Move it to the correct place, i.e., after
the switch statement.

Reported-by: Pedro Olsen Ferreira


svenvh:
Fair enough, updated to drop the dead code.

PR #109, opened 2019-01-07
Set noduplicate attribute for barriers

svenvh:
Without this attribute, LLVM passes such as SimplifyCFG may duplicate
barrier calls, which would cause incorrect behaviour of the compiled
program.


turinevgeny:
@svenvh, could you please provide an example?
AFAIK noduplicate prevents inlining into several callsites and seems more restrictive.
clang produces convergent attribute at this moment. What are the cases when convergent is not sufficient?

svenvh:
> @svenvh, could you please provide an example?
> AFAIK noduplicate prevents inlining into several callsites and seems more restrictive.
> clang produces convergent attribute at this moment. What are the cases when convergent is not sufficient?

Sorry for the delay in getting back to you.  I think you are correct that `noduplicate` is too restrictive and that it should be `convergent` instead.  I have opened #286 to fix this, thanks for reporting!

PR #113, opened 2019-01-14
Update tests

svenvh:
Update test files after LLVM commits:

 - 9597cfd242c ("Rename DIFlagFixedEnum to DIFlagEnumClass. NFC", 2019-01-08).
 - 08223c34d49 ("Replace "no-frame-pointer-*" function attributes with "frame-pointer"", 2019-01-14).


svenvh:
Seems I'm a bit too early with the second patch (-disable-fp-elim).  It was merged to LLVM only today so we first need to wait for Travis to grab the next LLVM nightly drop.

PR #116, opened 2019-01-15
Do not drop undef initializer for global variables

Fznamznon:


svenvh:
Just noticed your change is failing all the "allowed failures" in your check: https://travis-ci.org/KhronosGroup/SPIRV-LLVM-Translator/builds/480258067 (but maybe you already saw that).

pierremoreau:
As far as I understand the SPIR-V specification, the current behaviour is valid; from the SPIR-V 1.3 specification about `OpVariable`

> Initializer must be an <id> from a constant instruction or a global (module scope) OpVariable instruction.

The result of `OpUndef` belongs to neither of those, so it is not valid to initialise a variable with an undef.

PR #127, opened 2019-02-06
Fixed SPIRVReader/transKernelArgTypeMedataFromString

Djilka:
Fixed SPIRVReader/transKernelArgTypeMedataFromString to track  brackets and to not split argument when comma is inside brackets.


Djilka:
I added a test. Please review it again.

Djilka:
> It would be good if you can include a test case that demonstrates the issue (that is, fails without your change and now passes with your change).

Hi, could you review?

PR #130, opened 2019-02-14
Fixing preservation of unsignedness for read_imageui functions

neildhickey:
Test for this was added by svenvh. I can add to this pull request though if desired.

### DIFF
--- a/lib/SPIRV/SPIRVReader.cpp
+++ b/lib/SPIRV/SPIRVReader.cpp
@@ -800,16 +800,31 @@ bool SPIRVToLLVM::postProcessOCLBuiltinWithArrayArguments(
   return true;
 }
 
-static char getTypeSuffix(Type *T) {
-  char Suffix;
+static std::string getTypeSuffix(Type *T, SPIRVType *SPIRVT) {
+  std::string Suffix;
 
   Type *ST = T->getScalarType();
   if (ST->isHalfTy())
-    Suffix = 'h';
+    Suffix = "h";
   else if (ST->isFloatTy())
-    Suffix = 'f';
-  else
-    Suffix = 'i';
+    Suffix = "f";
+  else {
+    SPIRVTypeInt *STI = nullptr;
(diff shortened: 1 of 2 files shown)


AlexeySotkin:
@neildhickey, could you fix clang-format and clang-tidy issues.

lib/SPIRV/SPIRVReader.cpp
```diff
@@ -800,16 +800,27 @@ bool SPIRVToLLVM::postProcessOCLBuiltinWithArrayArguments(
   return true;
 }
 
-static char getTypeSuffix(Type *T) {
-  char Suffix;
+static std::string getTypeSuffix(Type *T, SPIRVType *SpirT) {
```
AlexeySotkin: Can we please do not mix SPIR and SPIR-V. It might be confusing.

lib/SPIRV/SPIRVReader.cpp
```diff
@@ -800,16 +800,27 @@ bool SPIRVToLLVM::postProcessOCLBuiltinWithArrayArguments(
...
   if (ST->isHalfTy())
-    Suffix = 'h';
+    Suffix = "h";
   else if (ST->isFloatTy())
-    Suffix = 'f';
-  else
-    Suffix = 'i';
+    Suffix = "f";
+  else {
+    SPIRVTypeInt *SIT = nullptr;
```
AlexeySotkin: What does "SIT" means? Is it "SPIRV Integer Type"?

lib/SPIRV/SPIRVReader.cpp
neildhickey: Agreed sorry, will fix

lib/SPIRV/SPIRVReader.cpp
neildhickey: Yes, I'll rename it to STI, though for some reason that seemed to be a wrong thing to say :)

PR #142, opened 2019-03-08
Fix typo in help text for -spirv-lower-const-expr

kpet:
Signed-off-by: Kevin Petit <kevin.petit@arm.com>


Hopetech:
`Reviewed-by: Elie Tournier <elie.tournier@collabora.com>`

PR #143, opened 2019-03-08
Fix OpSwitch miscalculation of Pair sizes

svenvh:
The attached test case asserted in SPIRVSwitch::validate().

Fix by Neil Hickey.  Test case by Sven van Haastregt.

### DIFF
--- a/lib/SPIRV/libSPIRV/SPIRVInstruction.h
+++ b/lib/SPIRV/libSPIRV/SPIRVInstruction.h
@@ -1084,10 +1084,12 @@ public:
   SPIRVBasicBlock *getDefault() const {
     return static_cast<SPIRVBasicBlock *>(getValue(Default));
   }
-  size_t getLiteralsCount() const {
-    return getSelect()->getType()->getBitWidth() / (sizeof(SPIRVWord) * 8);
+  size_t getLiteralSize() const {
+    unsigned ByteWidth = getSelect()->getType()->getBitWidth() / 8;
+    unsigned Remainder = (ByteWidth % sizeof(SPIRVWord)) != 0;
+    return (ByteWidth / sizeof(SPIRVWord)) + Remainder;
   }
-  size_t getPairSize() const { return getLiteralsCount() + 1; }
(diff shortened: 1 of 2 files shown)


lib/SPIRV/libSPIRV/SPIRVInstruction.h
```diff
@@ -1085,7 +1085,8 @@ class SPIRVSwitch : public SPIRVInstruction {
     return static_cast<SPIRVBasicBlock *>(getValue(Default));
   }
   size_t getLiteralsCount() const {
-    return getSelect()->getType()->getBitWidth() / (sizeof(SPIRVWord) * 8);
+    unsigned ByteWidth = getSelect()->getType()->getBitWidth() / 8;
+    return (ByteWidth / sizeof(SPIRVWord)) + !!(ByteWidth % sizeof(SPIRVWord));
```
AlexeySachkov: I think it would be more readable if rewrite this `!!(ByteWidth % sizeof(SPIRVWord))` into `((ByteWitdh % sizeof(SPIRVWord)) != 0)`

svenvh: Good point, perhaps it's even better to factor the modulo expression out into a separate variable.

lib/SPIRV/libSPIRV/SPIRVInstruction.h
```diff
@@ -1085,7 +1085,9 @@ class SPIRVSwitch : public SPIRVInstruction {
     return static_cast<SPIRVBasicBlock *>(getValue(Default));
   }
   size_t getLiteralsCount() const {
```
AlexeySotkin: The function name is unfortunate. It doesn't return number of literals, but it returns number of words required for each literal. @svenvh could you rename it to something like `getLiteralSize` ?
Other than that LGTM. Thanks!

svenvh: Thanks!  Renamed to `getLiteralSize`

PR #144, opened 2019-03-10
Add the input storage class and use it for builtins instead of Uniform

karolherbst:
builtins are generally not uniformly constant, so using the UniformConstant
storage class is missleading and might confuse the consumer of the spir-v

Sadly there is no perfect fit, but Input is what comes the closest

### DIFF
--- a/lib/SPIRV/OCL20ToSPIRV.cpp
+++ b/lib/SPIRV/OCL20ToSPIRV.cpp
@@ -1229,9 +1229,9 @@ void OCL20ToSPIRV::transWorkItemBuiltinsToVariables() {
     bool IsVec = I.getFunctionType()->getNumParams() > 0;
     Type *GVType =
         IsVec ? VectorType::get(I.getReturnType(), 3) : I.getReturnType();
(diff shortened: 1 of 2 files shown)


karolherbst:
@AlexeySotkin any opinions on that change?

PawelJurek:
Hi @karolherbst, we found that this pull request has negative functional impact on opencl-clang project and intel-graphics-compiler. We are currently investigating the issue and will get back to you soon.

karolherbst:
@PawelJurek yeah.. I kind of expected some fallout with this change, but that's one of the things which should be changed sooner than later.

Sadly there is no storage class which fits perfectly and where mainly wondering how we want to deal with that. UniformConstant kind of implies that the value is the same in each thread, which is not true for most builtins. and I could imagine that this might confuse some compilers requiring those to add workarounds for that as well, especially now that hardware comes out with registers accessible by the entire local group :/

PawelJurek:
@AlexeySotkin, @karolherbst - this PR can be merged as far as Intel Graphics Compiler is concerned.

PR #146, opened 2019-03-13
Sync with LLVM

AlexeySotkin:
Update lifetime.ll to align with changes introduced by:
commit 6e8fb99b69ceb9a2b0fe5bbd4754e1204128181f
Author: Matt Arsenault <Matthew.Arsenault@amd.com>
Date:   Tue Mar 12 21:02:54 2019 +0000

    IR: Add immarg attribute


AlexeySotkin:
It seems like version of llvm packages from apt.llvm.org git stuck at Feb 27 (20190227030930)
To check the status look here https://llvm-jenkins.debian.net/view/Ubuntu%20Xenial/job/llvm-toolchain-xenial-binaries/

svenvh:
> It seems like version of llvm packages from https://apt.llvm.org got stuck at Feb 27 (20190227030930)
> To check the status look here https://llvm-jenkins.debian.net/view/Ubuntu%20Xenial/job/llvm-toolchain-xenial-binaries/

Right, not sure if there's much we can do then...

AlexeySotkin:
> Any idea if we can do something about that?

I hope new version of the packages will be available soon. If the problem will persist for long time, I think we can relax the test and drop checks for declarations.

PR #147, opened 2019-03-13
Fix warning in release build

vmaksimo:


lib/SPIRV/SPIRVWriter.cpp
```diff
@@ -1113,7 +1113,8 @@ parseAnnotations(StringRef AnnotatedCode) {
 
   size_t OpenBracketNum = AnnotatedCode.count('{');
   size_t CloseBracketNum = AnnotatedCode.count('}');
-  assert(OpenBracketNum == CloseBracketNum);
+  if (OpenBracketNum != CloseBracketNum)
```
svenvh: Or you could keep the assert and do
```(void)CloseBracketNum;```

vmaksimo: Assert was used here by mistake, if-condition is a more correct for this check.

svenvh: Fair enough, if you just want to ignore any annotations with unmatched braces then this change looks good to me.

Issue #54, opened 2018-08-14
To consider enabling tests which check OpenCL C to SPIR-V translation


AlexeySotkin:
This kind of end-to-end testing should help us to keep compatibility with Clang on LLVM IR level.

AlexeySachkov:
Here is an example of end-to-end test: https://github.com/AlexeySachkov/SPIRV-LLVM-Translator/pull/1

Looks like everything is working out of the box - we need just to add some tests

AnastasiaStulova:
May be we can just setup whatever is easier to start from and then incrementally improve this with time.

svenvh:
Implemented by https://github.com/KhronosGroup/SPIRV-LLVM-Translator/pull/169

PR #169, opened 2019-04-11
Enable OpenCL C to SPIR-V tests and convert some tests

svenvh:
Add support for OpenCL C tests, such that we can test the entire
OpenCL C to SPIR-V flow.  This enables writing tests that are immune
to LLVM IR changes, improving robustness of the test suite.

Manually convert some tests from .ll files to .cl files.  I selected
these particular tests for the following reasons:
 - they contained the original source,
 - they contained a full clang invocation line,
 - they do not test specific LLVM IR module patterns,
 - they span different domains (e.g. debug info, atomics, images).

The OpenCL C tests seem to run a bit slower than LLVM IR tests, at
least for debug builds of clang, so we probably do not want to convert
the entire test suite to OpenCL C.

Implements https://github.com/KhronosGroup/SPIRV-LLVM-Translator/issues/54


svenvh:
`The job exceeded the maximum time limit for jobs, and has been terminated.`

Seems like we do not have enough CPU time to build clang on travis.  I've already cut the clang build config as much as possible (i.e., no targets, no static analyzer), but that did not help.  So the only remaining solution is to disable the .cl tests for `BUILD_EXTERNAL=0` builds.

AlexeySachkov:
> So the only remaining solution is to disable the .cl tests for `BUILD_EXTERNAL=0` builds.

We could try to install pre-built clang just to perform such end-to-end testing. Not sure that it can be easily achieved (clang package most likely requires llvm package and this pre-built llvm might conflict with one which was built from sources during Travis CI job)

svenvh:
> We could try to install pre-built clang just to perform such end-to-end testing. Not sure that it can be easily achieved (clang package most likely requires llvm package and this pre-built llvm might conflict with one which was built from sources during Travis CI job)

That is a great suggestion, it looks like we already install the clang and llvm packages for all build types.  So in theory: if I don't specify that the tests depend on clang, then it will not build clang and pick the pre-built clang instead (unless lit refuses to pick up a different clang).  Let me see if that works, I'll try to make the patch a bit less obscure if so...

AlexeySotkin:
What do you think of the following idea? What if we split the tests(except debug info maybe) as the following:
1. We strip CHECK-SPIRV part from the .cl files and put them to clang repo under /test/CodeGenOpenCL/ directory.
2. Configure via a CMake option location of /test/CodeGenOpenCL/ directory
3. Add tests with the CHECK-SPIRV part to this repo. These tests will run pre-installed clang and use files from /test/CodeGenOpenCL/ directory as input.

svenvh:
Not sure if it is ideal to put tests intended for llvm-spirv in the clang repo...  Imagine the situation where you want to add a .cl test for an llvm-spirv regression.  Then you'll first have to make a clang commit and then a translator commit.  It also means that tests between clang and the translator will have to be kept in sync.

I like the idea of reusing clang's CodeGenOpenCL tests for testing llvm-spirv though, as it may give us extra coverage.  But I think we should support translator-specific .cl tests in the translator repo too.

Applying the prebuilt clang trick seems to work for the .cl tests, but as my latest travis build shows we shouldn't rely on any other prebuilt tools as it makes the DebugInfo tests more fragile.  I'll rework my changes to only use prebuilt clang and leave the DebugInfo tests untouched.

svenvh:
I have added a CMake option `SPIRV_SKIP_CLANG_BUILD` such that we can avoid building clang on travis and use a prebuilt clang instead, and that seems to work.  All travis jobs are passing now, so this is ready for landing, unless there are any concerns?

AlexeySotkin:
> What do you think of the following idea? What if we split the tests ...

Yeah, probably that was not a good idea :)

PR #170, opened 2019-04-15
Fix debug info for __local qualified variables

AlexeySotkin:
Current mechanism for searching of global variables via name lookup is not
reliable. With this patch we look through all global variables untill we find
one which has attached !dbg metadata, which refers via
DIGlobalVariableExpression to DIGlobalVariable which is being translated.

### DIFF
--- a/lib/SPIRV/LLVMToSPIRVDbgTran.cpp
+++ b/lib/SPIRV/LLVMToSPIRVDbgTran.cpp
@@ -352,6 +352,18 @@ SPIRVEntry *LLVMToSPIRVDbgTran::getScope(DIScope *S) {
   }
 }
 
+SPIRVEntry *LLVMToSPIRVDbgTran::getGlobalVariable(const DIGlobalVariable *GV) {
+  for (GlobalVariable &V : M->globals()) {
+    SmallVector<DIGlobalVariableExpression *, 4> GVs;
+    V.getDebugInfo(GVs);
+    for (DIGlobalVariableExpression *GVE : GVs) {
+      if (GVE->getVariable() == GV)
+        return SPIRVWriter->transValue(&V, nullptr);
+    }
+  }
+  return getDebugInfoNone();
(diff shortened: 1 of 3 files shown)


lib/SPIRV/LLVMToSPIRVDbgTran.cpp
```diff
@@ -352,6 +352,18 @@ SPIRVEntry *LLVMToSPIRVDbgTran::getScope(DIScope *S) {
   }
 }
 
+SPIRVEntry *LLVMToSPIRVDbgTran::getGlobalVariable(const DIGlobalVariable *GV) {
+  for (llvm::GlobalVariable &V : M->globals()) {
```
svenvh: You shouldn't need `llvm::` here.

lib/SPIRV/LLVMToSPIRVDbgTran.cpp
```diff
@@ -352,6 +352,18 @@ SPIRVEntry *LLVMToSPIRVDbgTran::getScope(DIScope *S) {
...
+    SmallVector<DIGlobalVariableExpression *, 4> GVs;
+    V.getDebugInfo(GVs);
+    for (DIGlobalVariableExpression *GVE : GVs) {
+      if (GVE->getVariable() == GV)
+        return  SPIRVWriter->transValue(&V, nullptr);
```
svenvh: nitpick: double space after `return`.

lib/SPIRV/LLVMToSPIRVDbgTran.cpp
AlexeySotkin: Fixed

lib/SPIRV/LLVMToSPIRVDbgTran.cpp
AlexeySotkin: Fixed

PR #174, opened 2019-04-17
Translate Intel FPGA singlepump/doublepump memory attributes

AGindinson:

### DIFF
--- a/test/IntelFPGAMemoryAttributes.ll
+++ b/test/IntelFPGAMemoryAttributes.ll
@@ -11,6 +11,8 @@
 ; CHECK-SPIRV: Decorate {{[0-9]+}} NumbanksINTEL 4
 ; CHECK-SPIRV: Decorate {{[0-9]+}} BankwidthINTEL 8
 ; CHECK-SPIRV: Decorate {{[0-9]+}} MaxconcurrencyINTEL 4
+; CHECK-SPIRV: Decorate {{[0-9]+}} SinglepumpINTEL 1
+; CHECK-SPIRV: Decorate {{[0-9]+}} DoublepumpINTEL 1
 
 target datalayout = "e-i64:64-v16:16-v24:32-v32:32-v48:64-v96:128-v192:256-v256:256-v512:512-v1024:1024"
 target triple = "spir64-unknown-linux"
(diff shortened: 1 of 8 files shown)


AGindinson:
Public spec for these attributes is awaiting merge: https://github.com/KhronosGroup/SPIRV-Registry/pull/22

AGindinson:
@svenvh Could you take a look, please?

AlexeySachkov:
> Public spec for these attributes is awaiting merge: [KhronosGroup/SPIRV-Registry#22](https://github.com/KhronosGroup/SPIRV-Registry/pull/22)
> UPD: The above-mentioned PR was merged

Please put a link to the extension spec into commit message

svenvh:
> @svenvh Could you take a look, please?

Looks good to me.  As this is rather Intel-specific, I'll wait for someone from Intel to approve it before merging it.

test/IntelFPGAMemoryAttributes.ll
```diff
@@ -11,6 +11,8 @@
 ; CHECK-SPIRV: Decorate {{[0-9]+}} NumbanksINTEL 4
 ; CHECK-SPIRV: Decorate {{[0-9]+}} BankwidthINTEL 8
 ; CHECK-SPIRV: Decorate {{[0-9]+}} MaxconcurrencyINTEL 4
+; CHECK-SPIRV: Decorate {{[0-9]+}} SinglepumpINTEL 1
```
AlexeySachkov: Strictly speaking, this test violates the [spec](https://github.com/KhronosGroup/SPIRV-Registry/blob/master/extensions/INTEL/SPV_INTEL_fpga_memory_attributes.asciidoc)

As I can see, `RegisterINTEL`, `SinglepumpINTEL` and `DoublepumpINTEL` have **no** extra operands and you inserting one [integer literal](https://www.khronos.org/registry/spir-v/specs/unified1/SPIRV.html#_terms) here.

I guess it happened because you re-used existing infrastructure for these attributes introduced in d51037eb

This may cause problems if someone will use another SPIR-V consumer (which corresponds to the spec) instead of this translator.
Probably it is not critical, but I would like to see it fixed either in the spec or here in translator

AGindinson: Good spot, thank you! This will be discussed with the spec authors, and the necessary adjustments will be made within May.

Issue #120, opened 2019-01-24
NoSignedWrap decoration is emitted without declaring SPV_KHR_no_integer_wrap_decoration


bader:
(re-posting from [SYCL compiler issue #2](https://github.com/intel/llvm/issues/2))
In cases when the translator emits NoSignedWrap decoration the SPV_KHR_no_integer_wrap_decoration isn't declared with an OpExtension.

airlied:
@Fznamznon looks like your commit needs some fixes.

AlexeySotkin:
These decorations are part of the core spec now. Why do we need to declare the extension ? Maybe spirv-val needs another update.

AlexeySotkin:
I think the translator doesn't emit OpExtension properly in general, for any extension, not just for this one.

kpet:
Is anyone looking at this? This issue makes it impossible to get useful feedback from the validator on many modules.

AlexeySotkin:
I have it in my backlog.

kpet:
Happy to hear that :). Thanks!

AlexeySachkov:
Hi, I'm working on extensions support in SPIRV-Translator right now, feel free to review and leave your feedback in #175

PR #175, opened 2019-04-18
Add infrastructure to emit OpExtension when it is required

AlexeySachkov:
This PR introduces basic infrastructure for emitting `OpExtension`s opcodes for SPIR-V extensions similar to existing infrastructure around `OpCapability`.
It also adds generation of corresponding `OpExtension` for `SPV_INTEL_device_side_avc_motion_estimation` and `SPV_KHR_no_integer_wrap_decoration`.

We already have some infrastructure for emitting `OpExtension`s, but looks like it is related to some clang generated metadata for OpenCL extensions and SPIR. I don't know is it still actual or not, but I re-used some part of it.


kpet:
I've pulled this locally. It does seem to solve #120. Thanks :).

PR #177, opened 2019-04-19
Addition of lit test to check private array initialization.

garimagu:


garimagu:
Please review. @AlexeySotkin

garimagu:
> LGTM, just wondering if we should make this a transcoding test perhaps and move it into `test/transcoding`?

Sure. I changed the test to be in the trancoding folder.

svenvh:
> > LGTM, just wondering if we should make this a transcoding test perhaps and move it into `test/transcoding`?
> 
> Sure. I changed the test to be in the trancoding folder.

To be a real transcoding test, I would suggest adding some FileCheck patterns to check that `llvm-spirv -r` gives you back the initializer again.

Issue #176, opened 2019-04-18
Interface variables aren't declared in OpEntryPoint 


kpet:
Interface variables in the `Input` and `Output` storage classes are required to be declared in `OpEntryPoint` as per section 2.6 of the SPIR-V spec.

The following module allows to reproduce the issue:

```
; ModuleID = 'simple.cl'
source_filename = "simple.cl"
target datalayout = "e-i64:64-v16:16-v24:32-v32:32-v48:64-v96:128-v192:256-v256:256-v512:512-v1024:1024"
target triple = "spir64-unknown-unknown"

; Function Attrs: convergent nounwind writeonly
define spir_kernel void @test(i32 addrspace(1)* nocapture %out) local_unnamed_addr #0 !kernel_arg_addr_space !4 !kernel_arg_access_qual !5 !kernel_arg_type !6 !kernel_arg_base_type !6 !kernel_arg_type_qual !7 {
entry:
  %call = tail call spir_func i64 @_Z13get_global_idj(i32 0) #2
  %conv = trunc i64 %call to i32
  %idxprom = and i64 %call, 4294967295
  %arrayidx = getelementptr inbounds i32, i32 addrspace(1)* %out, i64 %idxprom
  store i32 %conv, i32 addrspace(1)* %arrayidx, align 4, !tbaa !8
  ret void
}

; Function Attrs: convergent nounwind readnone
declare spir_func i64 @_Z13get_global_idj(i32) local_unnamed_addr #1

attributes #0 = { convergent nounwind writeonly "correctly-rounded-divide-sqrt-fp-math"="false" "denorms-are-zero"="false" "disable-tail-calls"="false" "less-precise-fpmad"="false" "min-legal-vector-width"="0" "no-frame-pointer-elim"="false" "no-infs-fp-math"="false" "no-jump-tables"="false" "no-nans-fp-math"="false" "no-signed-zeros-fp-math"="false" "no-trapping-math"="false" "stack-protector-buffer-size"="8" "uniform-work-group-size"="true" "unsafe-fp-math"="false" "use-soft-float"="false" }
[...]

PR #178, opened 2019-04-22
Encode and decode Input and Output Variables used in entry points

karolherbst:
spirv-val reported an error as an OpEntryPoint has to list all used input and output variables. Sorry for missing this.

With this patch all input and output variables get added to each OpEntryPointer declaration, but figuring out which entry point is used isn't exactly trivially solveable.

Add the SPIRV output storage class as well, so we don't forget to update the code if it gets added later.

Fixes: 2745e2a6ae71dba4f8948ec5cc7ee07e8f9a4a97
Fixes: #176 

We probably want to do a quick analysis to figure out which GlobalVariable is used from which entry point, but for that I didn't really came up with a acceptable way of doing that. Maybe we could do something with the LazyCallGraph pass here?

### DIFF
--- a/lib/SPIRV/SPIRVWriter.cpp
+++ b/lib/SPIRV/SPIRVWriter.cpp
@@ -128,6 +128,7 @@ LLVMToSPIRV::LLVMToSPIRV(SPIRVModule *SMod)
 
 bool LLVMToSPIRV::runOnModule(Module &Mod) {
   M = &Mod;
+  CG = make_unique<CallGraph>(Mod);
   Ctx = &M->getContext();
   DbgTran->setModule(M);
   assert(BM && "SPIR-V module not initialized");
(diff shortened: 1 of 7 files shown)


lib/SPIRV/SPIRVWriter.cpp
```diff
@@ -1409,6 +1409,16 @@ bool LLVMToSPIRV::transGlobalVariables() {
   return true;
 }
 
+void LLVMToSPIRV::collectInputOutputVariables(SPIRVFunction *F) {
+  for (auto &GV : M->globals()) {
+    const auto AS = GV.getAddressSpace();
+    if (AS != SPIRAS_Input && AS != SPIRAS_Output)
+      continue;
+    // TODO: only add used variables
```
svenvh: You mentioned about a possible solution in https://github.com/KhronosGroup/SPIRV-LLVM-Translator/issues/176 .  Are you planning to address that in this pull request?

karolherbst: I will try to get something working. Just have to get more familiar with how to run llvm passes which aren't compatible with that "AU.addRequired<PASS>();" approach inside getAnalysisUsage. I'd like to use the LazyCallGraph or CallGraph Analysis passes to get the tree of called functions and just check if the node with the variable usage are reachable or just iterate through the tree starting from a function node...
Just have to figure out how all that works.

karolherbst: done

lib/SPIRV/SPIRVWriter.cpp
```diff
@@ -1409,6 +1410,59 @@ bool LLVMToSPIRV::transGlobalVariables() {
...
+      if (Done.find(NNF) == Done.end()) {
+        ToDo.insert(NNF);
+      }
+    }
+  }
+
+  return false;
+}
+
+void LLVMToSPIRV::collectInputOutputVariables(SPIRVFunction *F, Function *I) {
```
svenvh: `I` is a bit of an unfortunate name for a Function (`I` is typically used for iterators or instructions).

lib/SPIRV/SPIRVWriter.cpp
```diff
@@ -1409,6 +1410,59 @@ bool LLVMToSPIRV::transGlobalVariables() {
...
+  for (auto &GV : M->globals()) {
+    const auto AS = GV.getAddressSpace();
+    if (AS != SPIRAS_Input && AS != SPIRAS_Output)
+      continue;
+
+    for (const auto &U : GV.uses()) {
+      const Instruction *Inst = dyn_cast<Instruction>(U.getUser());
+      if (!Inst)
+        continue;
```
svenvh: Wondering if we should silently continue or assert here.  Does it mean we have hit a construct we cannot translate?

lib/SPIRV/libSPIRV/SPIRVFunction.h
```diff
@@ -105,6 +105,15 @@ class SPIRVFunction : public SPIRVValue, public SPIRVComponentExecutionModes {
     return getFunctionType()->getNumParameters();
   }
   SPIRVId getArgumentId(size_t I) const { return Parameters[I]->getId(); }
+  const std::vector<SPIRVId> getVariables() const {
+    std::vector<SPIRVId> Ids;
+    for (auto Variable : Variables)
+      Ids.push_back(Variable->getId());
+    return Ids;
+  };
```
svenvh: Spurious `;`

lib/SPIRV/SPIRVWriter.cpp
```diff
@@ -1409,6 +1410,59 @@ bool LLVMToSPIRV::transGlobalVariables() {
   return true;
 }
 
+bool LLVMToSPIRV::isFunctionReachableFromFunction(const Function *FS,
+                                                  const Function *FG) {
```
svenvh: Could this be a `const` method?

lib/SPIRV/SPIRVWriter.cpp
karolherbst: I am not quite sure, but there seems to be some kind of end Node which loops around the tree.

karolherbst:
mhh, I kind of came up with a better algorithm. We should collect all Functions where a GlobalVariable is loaded and check if we can reach any of them instead of checking each until we get a hit.

karolherbst:
> Thanks!
> 
> Could you also add a test please? (or update an existing test to check that the variables are emitted alongside OpEntryPoint)

I don't think there is an easy way to test that inside the tests we have. I could hard code ids, but that's rather annoying than helpful. I think our best bet would be to just push all generated spirvs through spirv-val as well and verify that this comes back without errors.

karolherbst:
@svenvh I've created #180 so that we are able to use spirv-val inside tests

svenvh:
> @svenvh I've created #180 so that we are able to use spirv-val inside tests

Fair enough, that will do as well!

PR #180, opened 2019-04-26
test: run spirv-val on generated spirvs

karolherbst:
this mostly adds the required infrastructure in order to run spirv-val on top of generated spirvs inside tests.

Because there are no PPAs available with a recent enough spirv-tools package, we have to build it inside travis-ci which increases the time of each build by around 8-9 minutes. I think most of the time is spend on building spirv-opt. As there is no official way to disable certain binaries, we might either want to add it to spirv-tools or just use some sed commands to decrease the compilation time of spirv-tools.

Because there are quite some validation issues we can't just enable it everywhere, so I ended up enabling it only for a few tests. We should get to a point where we can add it to every test though and fix all the upcoming issues.

There were also some invalid spt files which I fixed as it is pointless to verify we translating invalid spirvs works.

Any thoughts on that?

### DIFF
--- a/README.md
+++ b/README.md
@@ -64,11 +64,12 @@ make llvm-spirv -j`nproc`
 
 ## Test instructions
 
-All tests related to the translator are placed in the [test](test) directory.
+All tests related to the translator are placed in the [test](test) directory. Optionally the tests can make use of spirv-val (part of SPIRV-Tools) in order to validate the generated SPIR-V against the official SPIR-V specification.
+In case tests are failing due to SPIRV-Tools not supporting certain SPIR-V features, please get an updated package. The `PKG_CONFIG_PATH` environmental variable can be used to let cmake point to a custom installation.
 
-Execute the following command to run translator tests:
+Execute the following command inside the build directory to run translator tests:
 ```
-llvm-lit test
+make test
 ```
 This requires that the `-DLLVM_INCLUDE_TESTS=ON` argument was passed to CMake during the build step.
 
--- a/test/CMakeLists.txt
+++ b/test/CMakeLists.txt
@@ -1,9 +1,27 @@
+include(FindPkgConfig)
+
 llvm_canonicalize_cmake_booleans(SPIRV_SKIP_CLANG_BUILD)
 llvm_canonicalize_cmake_booleans(SPIRV_SKIP_DEBUG_INFO_TESTS)
 
 # required by lit.site.cfg.py.in
 get_target_property(LLVM_SPIRV_DIR llvm-spirv BINARY_DIR)
 set(LLVM_SPIRV_TEST_SOURCE_DIR ${CMAKE_CURRENT_SOURCE_DIR})
+
+# find spirv-val
+pkg_search_module(SPIRV_TOOLS SPIRV-Tools)
+if(SPIRV_TOOLS_FOUND)
+  find_program(SPIRV_TOOLS_SPIRV_VAL NAMES spirv-val PATHS ${SPIRV_TOOLS_PREFIX}/bin)
+  if(SPIRV_TOOLS_SPIRV_VAL)
+    get_filename_component(SPIRV_TOOLS_BINDIR "${SPIRV_TOOLS_SPIRV_VAL}" DIRECTORY)
+    set(SPIRV_TOOLS_SPIRV_VAL_FOUND True)
+  endif()
+endif()
+
+if(NOT SPIRV_TOOLS_SPIRV_VAL)
+  message(WARNING "spirv-val not found! SPIR-V generated for test suite will not be validated.")
+  set(SPIRV_TOOLS_SPIRV_VAL_FOUND False)
+endif()
+
 configure_lit_site_cfg(
   ${CMAKE_CURRENT_SOURCE_DIR}/lit.site.cfg.py.in
   ${CMAKE_CURRENT_BINARY_DIR}/lit.site.cfg.py
--- a/.travis.yml
+++ b/.travis.yml
@@ -19,11 +19,14 @@ addons:
     sources:
     - sourceline: 'deb http://apt.llvm.org/xenial/ llvm-toolchain-xenial main'
       key_url: 'https://apt.llvm.org/llvm-snapshot.gpg.key'
+    - sourceline: 'deb https://packages.lunarg.com/vulkan xenial main'
+      key_url: 'http://packages.lunarg.com/lunarg-signing-key-pub.asc'
     packages:
     - llvm-9-tools
     - llvm-9-dev
     - clang-format-9
     - clang-tidy-9
+    - spirv-tools
 
 compiler:
   - gcc
@@ -71,10 +74,17 @@ script:
       git clone https://git.llvm.org/git/llvm.git/ --depth 1
       mv llvm-spirv llvm/tools/llvm-spirv
     fi
+  - |
+    if [ $TRAVIS_OS_NAME == "osx" ]; then
+      wget -r --accept="*.tgz" --no-directories https://storage.googleapis.com/spirv-tools/badges/build_link_macos_clang_release.html
+      tar -xf install.tgz
+      sed -i '' -e 's|^prefix=\(.*\)|prefix='$PWD'/install|g' install/lib/pkgconfig/SPIRV-Tools.pc
+      export PKG_CONFIG_PATH=$PWD/install/lib/pkgconfig
+    fi
   - mkdir build && cd build
   - |
     if [ $BUILD_EXTERNAL == "1" ]; then
-      if [ $CHECK_FORMAT != "1" ]; then
+      if [ $CHECK_FORMAT != "1" ] && [ $CHECK_TIDY != "1" ]; then
         cmake .. \
           -DCMAKE_BUILD_TYPE=${BUILD_TYPE} \
           -DBUILD_SHARED_LIBS=${SHARED_LIBS} \
--- a/test/AtomicCompareExchange_cl12.ll
+++ b/test/AtomicCompareExchange_cl12.ll
@@ -1,4 +1,8 @@
-; RUN: llvm-as < %s | llvm-spirv -spirv-text | FileCheck %s
+; RUN: llvm-as %s -o %t.bc
+; RUN: llvm-spirv %t.bc -spirv-text -o %t
+; RUN: FileCheck < %t %s
+; RUN: llvm-spirv %t.bc -o %t.spv
+; RUN: spirv-val %t.spv
 
 target datalayout = "e-p:32:32-i64:64-v16:16-v24:32-v32:32-v48:64-v96:128-v192:256-v256:256-v512:512-v1024:1024"
 target triple = "spir-unknown-unknown"
--- a/test/AtomicCompareExchange_cl20.ll
+++ b/test/AtomicCompareExchange_cl20.ll
@@ -1,4 +1,7 @@
-; RUN: llvm-as < %s | llvm-spirv -spirv-text | FileCheck %s
+; RUN: llvm-as %s -o %t.bc
+; RUN: llvm-spirv %t.bc -spirv-text -o - | FileCheck %s
+; RUN: llvm-spirv %t.bc -o %t.spv
+; RUN: spirv-val %t.spv
 
 target datalayout = "e-p:32:32-i64:64-v16:16-v24:32-v32:32-v48:64-v96:128-v192:256-v256:256-v512:512-v1024:1024"
 target triple = "spir-unknown-unknown"
--- a/test/CheckCapKernelWithoutKernel.ll
+++ b/test/CheckCapKernelWithoutKernel.ll
@@ -1,6 +1,8 @@
-; RUN: llvm-as < %s | llvm-spirv -spirv-text -o %t
+; RUN: llvm-as %s -o %t.bc
+; RUN: llvm-spirv %t.bc -spirv-text -o %t
 ; RUN: FileCheck < %t %s
-
+; RUN: llvm-spirv %t.bc -o %t.spv
+; RUN: spirv-val %t.spv
 
 ; ModuleID = '../rel/CheckCapKernelWithoutKernel.bc'
 target datalayout = "e-i64:64-v16:16-v24:32-v32:32-v48:64-v96:128-v192:256-v256:256-v512:512-v1024:1024-n8:16:32:64"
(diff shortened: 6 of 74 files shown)


karolherbst:
it seems the MacOS build fails because there is no python3 :/

svenvh:
Thanks for working on this!  This seems like a good way to check that the generated SPIR-V is valid.

The build time increase is a valid concern though.  Perhaps the build system exposes a target like `install-spirv-val` or something like that? (sorry, can't check myself at the moment)

Alternatively there are releases at https://github.com/KhronosGroup/SPIRV-Tools/releases but they seem a bit outdated.  The master-tot release sounds like what we need, but apparently there is an issue: https://github.com/KhronosGroup/SPIRV-Tools/issues/2517

It sounds reasonable to only fix a few tests for now, we can enable spirv-val for more tests incrementally.

karolherbst:
@svenvh yeah.. I just added two spirv-val for cl based tests using the builtins so that we have this covered already.

karolherbst:
wrote a local patch to exclude the opt library and this cut the compile time locally by around 70%

But I kind of prefer having a package we could rely on or some pre built binaries for that. Maybe they get that CI stuff working again and we could just use that.

test/CMakeLists.txt
```diff
@@ -1,9 +1,17 @@
...
+
 llvm_canonicalize_cmake_booleans(SPIRV_SKIP_CLANG_BUILD)
 llvm_canonicalize_cmake_booleans(SPIRV_SKIP_DEBUG_INFO_TESTS)
 
 # required by lit.site.cfg.py.in
 get_target_property(LLVM_SPIRV_DIR llvm-spirv BINARY_DIR)
 set(LLVM_SPIRV_TEST_SOURCE_DIR ${CMAKE_CURRENT_SOURCE_DIR})
+
+# find spirv-val
+pkg_search_module(SPIRV_TOOLS REQUIRED SPIRV-Tools)
```
AlexeySachkov: Can we make it optional?

Downstream users might experience problems if SPIRV-Tools is not available in the environment

karolherbst: mhh, I think it's fine to require it for tests only. This file is only included if LLVM_INCLUDE_TESTS is set as well.

karolherbst: anyway, if we make SPIRV-tools optional we also have to figure out how to make the spirv-val invocations optional inside the tests as well

AlexeySachkov: > mhh, I think it's fine to require it for tests only

I guess anyone might launch tests on cloned repo. We probably need to document this dependency in the README

> anyway, if we make SPIRV-tools optional we also have to figure out how to make the spirv-val invocations optional inside the tests as well

We could try to replace `spirv-val` in LIT tests with some substitution which is defined as actual spirv-val binary if it available and as some noop command otherwise

karolherbst: > > mhh, I think it's fine to require it for tests only
> 
> I guess anyone might launch tests on cloned repo. We probably need to document this dependency in the README
> 
> > anyway, if we make SPIRV-tools optional we also have to figure out how to make the spirv-val invocations optional inside the tests as well
> 
> We could try to replace `spirv-val` in LIT tests with some substitution which is defined as actual spirv-val binary if it available and as some noop command otherwise

I think that's something we might want to do anyway. Most of the commands executed are the same in a lot of files and only differ in trivial changes (pipe vs temp files).

I think we should rework how we execute tests and have one wrapper for each type of file (spt, ll, cl) and then handle something like that there as well?

karolherbst: but then the lit substitution stuff doesn't work anymore. Maybe we should let cmake generate those test headers and then we can make it a little easier to add new tests.

karolherbst: @AlexeySachkov how strong is your opinion on making spirv-val optional? Personally I'd say that skipping that if somebody goes through the trouble running those tests then they also should have spirv-val somewhere as it's also not that difficult to setup (especially if we wait until the auto built binaries are available again)

AlexeySachkov: I, personally, think that having spirv-val in translator testing is a pretty good idea and this is important. So, I don't want to block this commit

I hope that downstream users will be able to workaround/fulfill this spirv-val requirement - it shouldn't be a very hard to do so

karolherbst: I think the most annoying issue will be that a distribution ships an outdated spirv-tools and a deveoper has to grab a newer one and get cmake to use this. I think I will add some documentation on this as most of the changes inside the test/lit.* files I made are due to this reason.

svenvh: > especially if we wait until the auto built binaries are available again

There are binaries available (e.g. https://github.com/KhronosGroup/SPIRV-Tools/releases/download/master-tot/SPIRV-Tools-master-linux-RelWithDebInfo.zip).  It's just that they're not up to date, this one is from July 2018.  I think there is a fair chance that they will actually work for us, even though they are outdated.

karolherbst: I am sure they won't work. We use some fairly recent features. I know that I run with my distribution provided spirv-tools into issues where spirv-val wasn't able to parse the spirv due to unknown capabilities and the package version is 2018.4.

The use of SPV_KHR_no_integer_wrap_decoration is one of the reason.

karolherbst: not saying it is a good idea, but something like that seems to kind of work: https://github.com/karolherbst/SPIRV-LLVM-Translator/commit/d900d6511c4d29a927644104377f77849d8567e7

I just don't like the idea of adding packages built against a newer version of ubuntu, especially if this includes C++ binaries.

karolherbst: And I also don't want to rely on a PPA providing backports...

karolherbst: @AlexeySachkov @svenvh maybe we should just download the vulkan sdk and extract the relevant binaries from there? I think this is something we could actually rely on. There are even ubuntu packages.

I will look into this and update the PR if I end up with something reliable and clean.

karolherbst: Will resolve this conversation as I think with the vulkan SDK apt repository that's a good enough solution.

karolherbst:
No idea what to do about macos though. We could download a vulkan SDK with a hardcoded version string or we could skip tests there? Or maybe we could ask for a "latest" symlink for downloading always the most recent SDK for macos and go with hardcoded until then.

karolherbst:
heh.. that google storage api link does some stupid meta refresh redirect which neither wget nor curl can follow. Using a HTTP redirect would be too simple I guess...

maybe I am able to figure something out which doesn't involve having to use sed to extract the target URL :/

karolherbst:
@svenvh @AlexeySachkov 
I think I came up with a more or less relible way of fetching the recent binaries on mac os x. I kept using the vulkan SDK apt repository for the linux build, because that's usually less painful and I think using the latest release of the tools should be good enough?

Don't know if we actually want to depend on master code or not as this makes it a bit troublesome for consumers as well. Any thoughts on that?

svenvh:
Thanks for looking at this!  I think we can keep it as you suggest for now, if it turns out to be problematic in some way we can switch to different binaries.

@karolherbst, do you think this PR is ready for merging now?  Perhaps you want to fix the typos in your commit messages first (unsued -> unused, "a ll" -> all).

test/CMakeLists.txt
AlexeySotkin: Sorry for the late feedback. But I have a strong opinion on having dependency on spirv-tools. Actually I think users must have an option to run tests without having to install spirv-tools package. 
Also will Vulkan SDK solution work on Windows OS?

AlexeySotkin: Also dependency on spirv-tools will break existing builds in 3rd party build/test systems.

karolherbst:
@svenvh updated the commit messages. I think this should be good to go

svenvh:
> @svenvh updated the commit messages. I think this should be good to go

@AlexeySotkin raised some concerns about the spirv-tools dependency, see the open conversation somewhere above.  So perhaps we should make validation optional, by defining a configurable lit substitution for spirv-val, or by scraping all .spv files after testing and running them through spirv-val for example.

test/CMakeLists.txt
karolherbst: > Sorry for the late feedback. But I have a strong opinion on having dependency on spirv-tools. Actually I think users must have an option to run tests without having to install spirv-tools package.
> Also will Vulkan SDK solution work on Windows OS?

there is a vulkan sdk package for windows. But I don't see the point of running tests if you don't run the full tests in the first place, because you don't get the full benefit of running them.

> Also dependency on spirv-tools will break existing builds in 3rd party build/test systems.

yeah.. that's actually a valid concern. Maybe we could make it optional, but print a big warning telling that whatever result the user is getting might not actually be reliable, because there could be potential spirv validation issue, which, personally, I think is kind of the big point here anyway. Although arguably somebody could deal with invalid spirv inside their platform, but they can't expect that to work on others and I think we shouldn't create incentives to be able to just ignore that, which we shouldn't because we don't want patches leading to invalid spirv, right?

So yeah, optional, but printing an easy to catch warning about it?

AlexeySotkin: Yes, I'm fine with the warning.
I agree that we don't want patches leading to invalid spirv. Probably we should establish a policy to add spirv-val check for all new tests producing spirv.

AlexeySotkin:
I think in case validation is disabled we can substitute spirv-val by `:` (colon).

karolherbst:
was playing around on how to do it properly and the best I could come up with was this. Any better suggestions, ideas?

```
diff --git a/test/CMakeLists.txt b/test/CMakeLists.txt
index 26aa9ee..e13ecbd 100644
--- a/test/CMakeLists.txt
+++ b/test/CMakeLists.txt
@@ -8,9 +8,17 @@ get_target_property(LLVM_SPIRV_DIR llvm-spirv BINARY_DIR)
 set(LLVM_SPIRV_TEST_SOURCE_DIR ${CMAKE_CURRENT_SOURCE_DIR})
 
 # find spirv-val
-pkg_search_module(SPIRV_TOOLS REQUIRED SPIRV-Tools)
-find_program(SPIRV_TOOLS_SPIRV_VAL NAMES spirv-val PATHS ${SPIRV_TOOLS_PREFIX}/bin)
-get_filename_component(SPIRV_TOOLS_BINDIR "${SPIRV_TOOLS_SPIRV_VAL}" DIRECTORY)
+pkg_search_module(SPIRV_TOOLS SPIRV-Tools)
+if(SPIRV_TOOLS_FOUND)
+  find_program(SPIRV_TOOLS_SPIRV_VAL NAMES spirv-val PATHS ${SPIRV_TOOLS_PREFIX}/bin)
+  if(SPIRV_TOOLS_SPIRV_VAL)
+    get_filename_component(SPIRV_TOOLS_BINDIR "${SPIRV_TOOLS_SPIRV_VAL}" DIRECTORY)
+  endif()
+endif()
+
+if(NOT SPIRV_TOOLS_SPIRV_VAL)
+  message(WARNING "spirv-val not found! Correctness of spirv can't be verified by provided tests")
+endif()
 
 configure_lit_site_cfg(
   ${CMAKE_CURRENT_SOURCE_DIR}/lit.site.cfg.py.in
diff --git a/test/lit.cfg.py b/test/lit.cfg.py
index 1e3f300..b7fde32 100644
--- a/test/lit.cfg.py
+++ b/test/lit.cfg.py
@@ -54,4 +54,5 @@ if not config.spirv_skip_debug_info_tests:
     tools.extend(['llc', 'llvm-dwarfdump', 'llvm-objdump', 'llvm-readelf', 'llvm-readobj'])
 
 llvm_config.add_tool_substitutions(tools, tool_dirs)
-llvm_config.add_tool_substitutions(['spirv-val'], [config.spirv_tools_bin_dir])
+llvm_config.add_tool_substitutions([ToolSubst('%spirv-val', command=FindTool('spirv-val'), unresolved='ignore')], [config.spirv_tools_bin_dir])
+config.substitutions.append(('%spirv-val', ':'))
diff --git a/test/transcoding/pipe_builtins.cl b/test/transcoding/pipe_builtins.cl
index 4c95520..4262454 100644
--- a/test/transcoding/pipe_builtins.cl
+++ b/test/transcoding/pipe_builtins.cl
@@ -4,7 +4,7 @@
 // RUN: llvm-spirv %t.bc -spirv-text -o %t.spt
 // RUN: FileCheck < %t.spt %s --check-prefix=CHECK-SPIRV
 // RUN: llvm-spirv %t.bc -o %t.spv
-// RUN: spirv-val %t.spv
+// RUN: %spirv-val %t.spv
 // RUN: llvm-spirv -r %t.spv -o %t.bc
 // RUN: llvm-dis < %t.bc | FileCheck %s --check-prefix=CHECK-LLVM
 

```

svenvh:
> was playing around on how to do it properly and the best I could come up with was this. Any better suggestions, ideas?

Looks reasonable enough.  Would it make sense for the lit substitution to be conditional on what cmake discovers?

karolherbst:
> > was playing around on how to do it properly and the best I could come up with was this. Any better suggestions, ideas?
> 
> Looks reasonable enough. Would it make sense for the lit substitution to be conditional on what cmake discovers?

yeah.. I was thinking about that. I think this would allow us to omit the added "%".. should be trivial enough I hope.

README.md
```diff
@@ -64,11 +64,12 @@ make llvm-spirv -j`nproc`
 
 ## Test instructions
 
-All tests related to the translator are placed in the [test](test) directory.
+All tests related to the translator are placed in the [test](test) directory. The tests require a recent enough SPIRV-Tools installation.
```
svenvh: This text needs to be updated to say that SPIRV-Tools are optional, not required.

test/CMakeLists.txt
```diff
@@ -1,9 +1,27 @@
...
+if(SPIRV_TOOLS_FOUND)
+  find_program(SPIRV_TOOLS_SPIRV_VAL NAMES spirv-val PATHS ${SPIRV_TOOLS_PREFIX}/bin)
+  if(SPIRV_TOOLS_SPIRV_VAL)
+    get_filename_component(SPIRV_TOOLS_BINDIR "${SPIRV_TOOLS_SPIRV_VAL}" DIRECTORY)
+    set(SPIRV_TOOLS_SPIRV_VAL_FOUND True)
+  endif()
+endif()
+
+if(NOT SPIRV_TOOLS_SPIRV_VAL)
+  message(WARNING "spirv-val not found! Correctness of spirv can't be verified by provided tests")
```
svenvh: spirv-val does not check that the translation was actually correct, only that the result conforms to the SPIR-V spec, so I'd rephrase the second sentence as:
"SPIR-V generated for test suite will not be validated."

PR #181, opened 2019-04-29
Fix translation of some Intel FPGA memory attributes

AGindinson:
This commit aligns the attributes' translation with the original specification

Extension specification can be found at:

### DIFF
--- a/lib/SPIRV/SPIRVWriter.cpp
+++ b/lib/SPIRV/SPIRVWriter.cpp
@@ -1135,11 +1135,11 @@ parseAnnotations(StringRef AnnotatedCode) {
       Dec = llvm::StringSwitch<Decoration>(S)
                 .Case("1", DecorationSinglepumpINTEL)
                 .Case("2", DecorationDoublepumpINTEL);
-      Value = "1";
+    } else if (F == "register") {
+      Dec = DecorationRegisterINTEL;
     } else {
       Dec = llvm::StringSwitch<Decoration>(F)
                 .Case("memory", DecorationMemoryINTEL)
-                .Case("register", DecorationRegisterINTEL)
                 .Case("numbanks", DecorationNumbanksINTEL)
                 .Case("bankwidth", DecorationBankwidthINTEL)
                 .Case("max_concurrency", DecorationMaxconcurrencyINTEL);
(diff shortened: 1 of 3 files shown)


lib/SPIRV/SPIRVWriter.cpp
```diff
@@ -1158,7 +1158,11 @@ void addIntelFPGADecorations(
   for (const auto &I : Decorations) {
     if (I.first == DecorationMemoryINTEL)
```
asavonic: How about a `switch` statement here?

lib/SPIRV/SPIRVWriter.cpp
```diff
@@ -1158,7 +1158,11 @@ void addIntelFPGADecorations(
...
       E->addDecorate(new SPIRVDecorateMemoryINTELAttr(E, I.second));
-    else {
+    else if (I.first == DecorationRegisterINTEL ||
+             I.first == DecorationSinglepumpINTEL ||
+             I.first == DecorationDoublepumpINTEL) {
+      E->addDecorate(I.first);
```
AlexeySachkov: You can also add an assertion that `I.second` is an empty `StringRef`

lib/SPIRV/SPIRVWriter.cpp
AGindinson: Done

lib/SPIRV/SPIRVWriter.cpp
AGindinson: Done

lib/SPIRV/SPIRVWriter.cpp
```diff
@@ -1156,12 +1156,21 @@ void addIntelFPGADecorations(
...
       E->addDecorate(new SPIRVDecorateMemoryINTELAttr(E, I.second));
-    else {
+      break;
+    case DecorationRegisterINTEL:
+    case DecorationSinglepumpINTEL:
+    case DecorationDoublepumpINTEL:
+      assert(I.second.empty());
+      E->addDecorate(I.first);
+      break;
+    default:
```
AGindinson: I guess 'default' could also be replaced by a full listing of remaining attributes (clarification for future conributors). @asavonic @AlexeySachkov please reply if this is necessary

AlexeySachkov: I'm not sure that it is necessary. Probably, just a comment like `The rest of Intel FPGA-specific decorations (DecorationA, DecorationB, etc.)` could be helpful

AGindinson: Thanks, I'll add the comment then

PR #183, opened 2019-05-01
Open SPIR-V output file as a binary file

bashbaug:
I'm using the SPIRV-LLVM-Translator on Windows.  On Windows, "text" files replace a newline (character '\n') with two characters representing a carriage return and line feed ("\r\n").  Unfortunately, if the newline actually represents a SPIR-V ID, the replacement corrupts the SPIR-V file.

Fix is to use a "binary" file instead, which only writes the single byte '\n'.  With this change I am able to use the SPIRV-LLVM-Translator on Windows.

### DIFF
--- a/tools/llvm-spirv/llvm-spirv.cpp
+++ b/tools/llvm-spirv/llvm-spirv.cpp
@@ -139,7 +139,7 @@ static int convertLLVMToSPIRV() {
   std::string Err;
   bool Success = false;
   if (OutputFile != "-") {
-    std::ofstream OutFile(OutputFile);
+    std::ofstream OutFile(OutputFile, std::ios::binary);
     Success = writeSpirv(M.get(), OutFile, Err);
   } else {
     Success = writeSpirv(M.get(), std::cout, Err);


tools/llvm-spirv/llvm-spirv.cpp
```diff
@@ -139,7 +139,7 @@ static int convertLLVMToSPIRV() {
   std::string Err;
   bool Success = false;
   if (OutputFile != "-") {
-    std::ofstream OutFile(OutputFile);
+    std::ofstream OutFile(OutputFile, std::ios::binary);
```
svenvh: Whether we provide `std::ios::binary` should perhaps be conditional on `SPIRV::SPIRVUseTextFormat`, see line 136?
(although I am not entirely sure when `SPIRVUseTextFormat` would be true in normal usage of llvm-spirv)

bashbaug: I'm able to generate readable text output using `std::ios::binary`, but it'd be easy to chose the file mode selectively if desired.  Something like:

````c++
    std::ofstream OutFile;
    if (SPIRV::SPIRVUseTextFormat) {
      OutFile.open(OutputFile);
    } else {
      OutFile.open(OutputFile, std::ios::binary);
    }
````

Let me know what you'd prefer, thanks!

svenvh: > I'm able to generate readable text output using `std::ios::binary`

Fair enough, then your patch should be fine, thanks!

PR #185, opened 2019-05-06
Initial support of Intel FPGA loop attributes/pragmas

MrSidims:
Following attributes/pragmas are supported:
 1. ivdep
 2. ivdep safelen (N)
 3. ii(N)
 4. max_concurrency_loop(N)

LLVM IR ->SPIRV translation (appropriatly the set order):
 1. !"llvm.loop.ivdep.enable ->
    4 LoopMerge {{[0-9]+}} {{[0-9]+}} 4
 2. !"llvm.loop.ivdep.safelen", i32 N ->
    5 LoopMerge {{[0-9]+}} {{[0-9]+}} 8 N
 3. !"llvm.loop.ii.count", i32 N ->
    6 LoopMerge {{[0-9]+}} {{[0-9]+}} 0x80000000 5889 N
 4. !"llvm.loop.max_concurrency.count", i32 N ->
    6 LoopMerge {{[0-9]+}} {{[0-9]+}} 0x80000000 5890 N

SPEC: https://github.com/KhronosGroup/SPIRV-Registry/pull/28

PR #186, opened 2019-05-07
Support partial unrolling

MrSidims:
Possible use cases:
1. unroll [N]
   going to be translated into:
   SPIR-V: 5 LoopMerge [X] [Y] 256 [N]
   LLVM-IR: !{!"llvm.loop.unroll.count", i32 [N]}

2. unroll 1
   in this case loop unrolling should be disabled,
   going to be translated into:
   LLVM-IR:  !{!"llvm.loop.unroll.disable"}

unroll default (without hint) behaviour was changed as well, see:
LLVM-IR: !{!"llvm.loop.unroll.full"} -> !{!"llvm.loop.unroll.enable"}

### DIFF
--- a/lib/SPIRV/SPIRVReader.cpp
+++ b/lib/SPIRV/SPIRVReader.cpp
@@ -547,6 +547,18 @@ void SPIRVToLLVM::setName(llvm::Value *V, SPIRVValue *BV) {
     V->setName(Name);
 }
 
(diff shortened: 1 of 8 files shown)


MrSidims:
I will commit https://github.com/KhronosGroup/SPIRV-LLVM-Translator/pull/185 first since it will bring multiple LoopControls attached to a single loop handling.

Issue #20, opened 2018-05-08
Separate SPIR-V parsing and translation


neildhickey:
Currently the translator parses and translates SPIR-V to LLVM in one step from the API perspective. It would be useful to separate out the parsing and translation into two separate API calls. This would allow the correct implementation of clCreateProgramWithIL as this API function needs to be able to return CL_INVALID_VALUE if the passed IL is not well-formed. Currently passing malformed IL causes a crash in the parser or translation step. This also allows a separate call to clCompileProgram/clBuildProgram to perform the translation at that stage, giving an appropriate error, CL_INVALID_OPERATION, if the program has not been created from source or IL and also if the target device doesn't support a capability requested in the SPIR-V binary, see https://www.khronos.org/registry/OpenCL/sdk/2.1/docs/man/xhtml/clCompileProgram.html.
clCompileProgram/clBuildProgram also supports the passing of compiler options, which we need to think about how to support at this stage, and we'd want to be able to extract the build log using clGetProgramBuildInfo which should be populated by the last call to clCompileProgram or clBuildProgram.

The additional benefit of doing this separation would be to allow us to reuse the common SPIR-V parsing and in-memory representation found in the SPIRV-Tools repository, cutting down on the amount of duplication and effort required to support newer versions of SPIR-V.

svenvh:
Implemented by #188

PR #188, opened 2019-05-10
Add separate functions for reading and translating SPIR-V

StuartDBrady:
The OpenCL runtime API requires that IL be loaded and compiled with
separate function calls:

 1. clCreateProgramWithIL(), which must check that the IL is well-formed
    and return CL_INVALID_VALUE if it is not.

 2. clBuildProgram() or clCompileProgram(), which allow build options to
    be specified.

To facilitate the above, this change adds two new API functions:

 1. readSpirvModule(), which reads an istream to produce a SPIRVModule.

 2. convertSpirvToLLVM(), which converts a SPIRVModule to LLVM IR.

The existing readSpirv() function is refactored in terms of these calls,
but is kept in the API for the convenience of tools such as llvm-spirv.

Fixes issue #20.

### DIFF
--- a/include/LLVMSPIRVLib.h
+++ b/include/LLVMSPIRVLib.h
@@ -89,6 +89,11 @@ bool convertSpirv(std::string &Input, std::string &Out, std::string &ErrMsg,
 bool isSpirvText(std::string &Img);
 #endif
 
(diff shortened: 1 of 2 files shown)


AlexeySotkin:
LGTM! Thanks.

PR #189, opened 2019-05-13
Translation of llvm.dbg.declare in case the local variable is optimiz…

AlexeySotkin:
…ed out

### DIFF
--- a/lib/SPIRV/SPIRVToLLVMDbgTran.cpp
+++ b/lib/SPIRV/SPIRVToLLVMDbgTran.cpp
@@ -893,6 +893,21 @@ SPIRVToLLVMDbgTran::transDebugIntrinsic(const SPIRVExtInst *DebugInst,
   case SPIRVDebug::Declare: {
     using namespace SPIRVDebug::Operand::DebugDeclare;
     auto LocalVar = GetLocalVar(Ops[DebugLocalVarIdx]);
+    if (getDbgInst<SPIRVDebug::DebugInfoNone>(Ops[VariableIdx])) {
+      // If we don't have the variable(e.g. alloca might be promoted by mem2reg)
+      // we should generate the following IR:
+      // call void @llvm.dbg.declare(metadata !4, metadata !14, metadata !5)
+      // !4 = !{}
+      // DIBuilder::insertDeclare doesn't allow to pass nullptr for the Storage
+      // parameter. To work around this limitation we create a dummy temp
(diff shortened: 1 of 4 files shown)


lib/SPIRV/SPIRVToLLVMDbgTran.cpp
```diff
@@ -893,6 +893,21 @@ SPIRVToLLVMDbgTran::transDebugIntrinsic(const SPIRVExtInst *DebugInst,
   case SPIRVDebug::Declare: {
     using namespace SPIRVDebug::Operand::DebugDeclare;
     auto LocalVar = GetLocalVar(Ops[DebugLocalVarIdx]);
+    if (getDbgInst<SPIRVDebug::DebugInfoNone>(Ops[VariableIdx])) {
+      // If we don't have the variable(e.g. alloca might be promoted by mem2reg)
+      // we should generate the following IR:
+      // call void @llvm.dbg.declare(metadata !4, metadata !14, metadata !5)
+      // !4 = !{}
+      // DIBuilder::insertDeclare doesn't allow to pass nullptr for the Storage
+      // paramter. To work around this limitation we create a dummy temp alloca,
```
svenvh: paramter -> parameter

lib/SPIRV/SPIRVToLLVMDbgTran.cpp
```diff
@@ -893,6 +893,21 @@ SPIRVToLLVMDbgTran::transDebugIntrinsic(const SPIRVExtInst *DebugInst,
...
+      // use it to create llvm.dbg.declare, and then remove the alloca.
```
svenvh: My DebugInfo knowledge is limited, but would it not be better to create an `llvm.dbg.value` call instead, as that seems to be what llvm's mem2reg is doing?

lib/SPIRV/SPIRVToLLVMDbgTran.cpp
AlexeySotkin: Fixed

lib/SPIRV/SPIRVToLLVMDbgTran.cpp
AlexeySotkin: mem2reg creates `llvm.dbg.value` for initialized variable. For uninitialized variables mem2reg just removes alloca and leaves `llvm.deb.declare`

PR #190, opened 2019-05-13
Translate annotate attribute

vmaksimo:
This patch implements translation of LLVM IR var/ptr.annotation intrinsic to SPIR-V UserSemantic decoration.
Example of code with annotate attribute:
`[[clang::annotate("{FOO}")]] int a;`

### DIFF
--- a/lib/SPIRV/SPIRVReader.cpp
+++ b/lib/SPIRV/SPIRVReader.cpp
@@ -2247,6 +2247,8 @@ void generateIntelFPGAAnnotation(const SPIRVEntry *E,
     Out << "{pump:1}";
   if (E->hasDecorate(DecorationDoublepumpINTEL))
     Out << "{pump:2}";
+  if (E->hasDecorate(DecorationUserSemantic))
+    Out << E->getDecorationStringLiteral(DecorationUserSemantic);
 }
 
 void generateIntelFPGAAnnotationForStructMember(
@@ -2273,6 +2275,9 @@ void generateIntelFPGAAnnotationForStructMember(
     Out << "{pump:1}";
   if (E->hasMemberDecorate(DecorationDoublepumpINTEL, 0, MemberNumber))
     Out << "{pump:2}";
+  if (E->hasMemberDecorate(DecorationUserSemantic, 0, MemberNumber))
+    Out << E->getMemberDecorationStringLiteral(DecorationUserSemantic,
+                                               MemberNumber);
 }
 
 void SPIRVToLLVM::transIntelFPGADecorations(SPIRVValue *BV, Value *V) {
--- a/lib/SPIRV/SPIRVWriter.cpp
+++ b/lib/SPIRV/SPIRVWriter.cpp
@@ -1142,8 +1142,12 @@ parseAnnotations(StringRef AnnotatedCode) {
                 .Case("memory", DecorationMemoryINTEL)
                 .Case("numbanks", DecorationNumbanksINTEL)
                 .Case("bankwidth", DecorationBankwidthINTEL)
-                .Case("max_concurrency", DecorationMaxconcurrencyINTEL);
-      Value = S;
+                .Case("max_concurrency", DecorationMaxconcurrencyINTEL)
+                .Default(DecorationUserSemantic);
+      if (Dec == DecorationUserSemantic)
+        Value = AnnotatedCode.substr(From, To + 1);
+      else
+        Value = S;
     }
 
     Decorates.push_back({Dec, Value});
@@ -1159,6 +1163,9 @@ void addIntelFPGADecorations(
     switch (I.first) {
     case DecorationMemoryINTEL:
       E->addDecorate(new SPIRVDecorateMemoryINTELAttr(E, I.second));
+      break;
+    case DecorationUserSemantic:
+      E->addDecorate(new SPIRVDecorateUserSemanticAttr(E, I.second));
       break;
     case DecorationRegisterINTEL:
     case DecorationSinglepumpINTEL:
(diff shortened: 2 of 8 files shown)


lib/SPIRV/SPIRVReader.cpp
```diff
@@ -2206,6 +2206,8 @@ void generateIntelFPGAAnnotation(const SPIRVEntry *E,
     Out << "{pump:1}";
   if (E->hasDecorate(DecorationDoublepumpINTEL))
     Out << "{pump:2}";
+  if (E->hasDecorate(DecorationUserSemantic))
```
AlexeySachkov: This decoration is a core spec feature and it is not related to intelfpga annotations. So, it looks a little bit strange to handle core functionality in this function

lib/SPIRV/SPIRVWriter.cpp
```diff
@@ -1313,8 +1324,12 @@ SPIRVValue *LLVMToSPIRV::transIntrinsicInst(IntrinsicInst *II,
 
     std::vector<std::pair<Decoration, std::string>> Decorations =
         parseAnnotations(AnnotationString);
-
-    addIntelFPGADecorations(SV, Decorations);
+    if (Decorations.empty()) {
```
AlexeySachkov: `DecorateUserSemantic` decoration is handled inside `parseAnnotations` function. So, as long as `AnnotationString` is not empty, `Decorations` will always contain something. What is the point of this `if` then?

lib/SPIRV/libSPIRV/SPIRVDecorate.cpp
```diff
@@ -91,6 +91,8 @@ void SPIRVDecorate::encode(spv_ostream &O) const {
     SPIRVDecorateLinkageAttr::encodeLiterals(Encoder, Literals);
   else if (Dec == DecorationMemoryINTEL)
     SPIRVDecorateMemoryINTELAttr::encodeLiterals(Encoder, Literals);
+  else if (Dec == DecorationUserSemantic)
```
AlexeySachkov: I think it is better to refactor this if-else chain into swtich...case statement

test/annotate_attribute.ll
```diff
@@ -0,0 +1,141 @@
...
+; Function Attrs: nounwind
+define spir_func void @_Z3foov() #3 {
+entry:
+  %var_one = alloca i32, align 4
+  %var_two = alloca i32, align 4
+  %var_three = alloca i8, align 1
+  %0 = bitcast i32* %var_one to i8*
+  call void @llvm.lifetime.start.p0i8(i64 4, i8* %0) #4
+  %var_one1 = bitcast i32* %var_one to i8*
+  ; CHECK-LLVM: call void @llvm.var.annotation(i8* %[[VAR1:[a-zA-Z0-9_]+]], i8* getelementptr inbounds ([3 x i8], [3 x i8]* [[STR1]], i32 0, i32 0), i8* undef, i32 undef)
```
AlexeySachkov: `STR1` is equal to `42`, while  `@.str.1` is equal to `annotate_attribute.cpp\00`, how does this test work?

lib/SPIRV/SPIRVWriter.cpp
vmaksimo: The only case for `DecorateUserSemantic` that is handled in `parseAnnotations` function is when `AnnotationString` contains arbitrary string in `{}` brackets. 
I can remove support of this case from `parseAnnotations`, so Decorations will be empty if `AnnotationString` contains any random string except the case with `{SOME_STRING}` - this situation causes assertion from StringSwitch in `parseAnnotations` function as it can't found correct match.

test/annotate_attribute.ll
vmaksimo: Didn't get your concern. The test is similar to existing tests on attributes translation, e.g. `IntelFPGAMemoryAttributes.ll`.

lib/SPIRV/SPIRVReader.cpp
vmaksimo: Agree. But now I have no ideas how to add this feature and not to duplicate whole `transIntelFPGADecorations` function, which functionality is fully used for UserSemantic translation to LLVM-IR.
Thinking about removing `IntelFPGA` from related function names, but it's cool solution too.

test/annotate_attribute.ll
AlexeySachkov: I'm just looking at llvm ir:

```
call void @llvm.var.annotation( ... @.str.1, ...)
; where @.str.1 is "annotate_attribute.cpp\00"
```
And your check:
```
call void @llvm.var.annotation( ... [[STR1]], ...)
; where STR1 is "42"
```

I wonder how does it work if you are checking for different annotation string

vmaksimo: The test checks only first two arguments of var/ptr.annotation intrinsic. 

The first argument is a pointer to a value, the second is a pointer to a global string with annotation.
The third argument `@.str.1` with source file name and the fourth argument with line number are not translated from SPIR-V to LLVM-IR, so they are not checked.

Numeration of `STR1, STR2, ...` is related only to global annotation strings, they are `'42', '{FOO}', ..`.

test/annotate_attribute.ll
```diff
@@ -0,0 +1,141 @@
...
+; CHECK-SPIRV: MemberDecorate {{[0-9]+}} 1 UserSemantic "128"
+; CHECK-SPIRV: MemberDecorate {{[0-9]+}} 0 UserSemantic "{baz}"
+
+target datalayout = "e-i64:64-v16:16-v24:32-v32:32-v48:64-v96:128-v192:256-v256:256-v512:512-v1024:1024"
+target triple = "spir64-unknown-linux"
+
+%class.anon = type { i8 }
+%struct.bar = type { i32, i8 }
+
+; CHECK-LLVM: [[STR1:@[a-zA-Z0-9_.]+]] = {{.*}}42
```
AlexeySotkin: I think it will be easier to understand the test if you rename this variable to STR, in order to match its name with IR counterpart.

AlexeySotkin:
I don't think that this is a correct usage if this decoration. Note that ID 5635 is reserved for `UserSemantic` and for `HlslSemanticGOOGLE`. And originally intention of this decoration was to "express user-defined and system semantic values for **shaders originally written in HLSL**". Moreover the spec requires that the variable being decorated must be in `Input` or `Output` storage class.

MrSidims:
@AlexeySotkin do you approve the PR after answer in https://github.com/KhronosGroup/SPIRV-Headers/issues/110 given?

PR #192, opened 2019-05-15
Set atomic_compare_exchange argument address space to generic

svenvh:
Ensure that the `expected` argument of atomic_compare_exchange_* lives
in the generic address space in accordance with the OpenCL 2.0
specificationSection 6.13.11, "The atomic_compare_exchange
functions".

### DIFF
--- a/lib/SPIRV/SPIRVToOCL20.cpp
+++ b/lib/SPIRV/SPIRVToOCL20.cpp
@@ -382,7 +382,11 @@ void SPIRVToOCL20::visitCallSPIRVAtomicBuiltin(CallInst *CI, Op OC) {
                                     .getFirstInsertionPt()));
           PExpected->setAlignment(CI->getType()->getScalarSizeInBits() / 8);
           new StoreInst(Args[1], PExpected, PInsertBefore);
-          Args[1] = PExpected;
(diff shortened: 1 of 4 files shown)


AlexeySotkin:
@svenvh, please add to commit message reference to the requirement in the spec

lib/SPIRV/SPIRVToOCL20.cpp
```diff
@@ -382,7 +382,11 @@ void SPIRVToOCL20::visitCallSPIRVAtomicBuiltin(CallInst *CI, Op OC) {
                                     .getFirstInsertionPt()));
           PExpected->setAlignment(CI->getType()->getScalarSizeInBits() / 8);
           new StoreInst(Args[1], PExpected, PInsertBefore);
-          Args[1] = PExpected;
+          unsigned AddrSpc = SPIRAS_Generic;
```
AlexeySotkin: What if use this address space directly in Alloca's constructor above?

svenvh: The LLVM IR Verifier rejects alloca's with an address space different from the DataLayout's alloca address space.
(so the assertion "Allocation instruction pointer not in the stack address space!" will fire if you do).

PR #195, opened 2019-05-21
Replace hardcoded address space for clk_event_t

svenvh:
Set the address space through `getOCLOpaqueTypeAddrSpace`, which is
build-time configurable through the `SPIRV_CLK_EVENT_T_ADDR_SPACE`
preprocessor macro.

There is no easy way to test this functionality using the llvm-lit test suite currently.

### DIFF
--- a/lib/SPIRV/SPIRVReader.cpp
+++ b/lib/SPIRV/SPIRVReader.cpp
@@ -1881,7 +1881,8 @@ Instruction *SPIRVToLLVM::transEnqueueKernelBI(SPIRVInstruction *BI,
   Function *F = M->getFunction(FName);
   if (!F) {
     Type *EventTy = PointerType::get(
-        getOrCreateOpaquePtrType(M, SPIR_TYPE_NAME_CLK_EVENT_T, SPIRAS_Private),
+        getOrCreateOpaquePtrType(M, SPIR_TYPE_NAME_CLK_EVENT_T,
+                                 getOCLOpaqueTypeAddrSpace(OpTypeDeviceEvent)),
         SPIRAS_Generic);
 
     SmallVector<Type *, 8> Tys = {


AlexeySachkov:
> There is no easy way to test this functionality using the llvm-lit test suite currently.

@svenvh, we can update [lit.site.cfg.py.in](test/lit.site.cfg.py.in) in order to pass this `SPIRV_CLK_EVENT_T_ADDR_SPACE` define down to [lit.site.cfg.py](test/lit.site.cfg.py) where we can add some substitution in order to use this value in LIT tests.

I'm not sure how, but I imagine something like:
```
; RUN: llvm-spirv -r ... | FileCheck %s --check-prefix=ADDRSPACE%clk_event_t_addr_space
; ADDRSPACE0: check private
; ADDRSPACE1: check another address space
```

svenvh:
Thanks @AlexeySachkov , yeah we could make that work, but it means we need to let cmake substitute the FileCheck prefixes or patterns inside the test.  That feels a bit obscure to me to be honest.  I am not sure it is worth adding this obscurity to the test suite; what do you think?

AlexeySachkov:
> but it means we need to let cmake substitute the FileCheck prefixes or patterns inside the test. That feels a bit obscure to me to be honest.

@svenvh, I agree that it is probably too tricky/hacky approach.

I haven't checked, but I assume that there are no regressions directly related to this change. It means that either:
  * we don't have tests for clk_even_t address space
  * we have such tests and they check for the same address space as default value of `SPIRV_CLK_EVENT_T_ADDR_SPACE`

I'm ok with this patch as-is, but there are some thoughts about these compile-time switches in general:
* What happens if someone change one of defines and launch LIT tests? How many tests will fail? I.e. how friendly our testing infrastructure for such changes? 
* Do we need or have tests for default values of these `SPIRV_*_ADDR_SPACE` defines?
  * What if someone change one of defines, but all tests will pass anyway?
    * Does that mean that we don't have corresponding tests? Or they are not needed at all since value of the define is not critical?

I understand that all of these questions might look like some overkill, especially if we don't have users of these defines.

svenvh:
> * we have such tests and they check for the same address space as default value of `SPIRV_CLK_EVENT_T_ADDR_SPACE`

This is indeed the case.  If I add `-DSPIRV_CLK_EVENT_T_ADDR_SPACE=SPIRAS_Local` to `CMAKE_CXX_FLAGS`, then two tests (transcoding/enqueue_kernel.cl and transcoding/spirv-types.ll) fail as the FileCheck patterns expect a different address space.

> I'm ok with this patch as-is

Thanks!

> but there are some thoughts about these compile-time switches in general:
> 
> * What happens if someone change one of defines and launch LIT tests? How many tests will fail? I.e. how friendly our testing infrastructure for such changes?

We can expect various tests to fail like the ones I mentioned above.  The problem is that these defines are set at build time, so to test that they work as expected, we need different build configurations.  I don't believe they have to be build time constants actually, so I can have a look at making these dynamically configurable as a followup to this.

> I understand that all of these questions might look like some overkill, especially if we don't have users of these defines.

We find them quite handy, so it's definitely something I'd like to keep/improve.

AlexeySotkin:
@svenvh, could you rebase your branch on top of master to re-run travis with updated config.

PR #198, opened 2019-05-24
travis: Fix CHECK_TIDY config

AlexeySotkin:


AlexeySachkov:
From [here](https://travis-ci.org/KhronosGroup/SPIRV-LLVM-Translator/jobs/536825715#L420):
```
No source code to check for tidying.
```

@AlexeySotkin, is it expected?

AlexeySotkin:
No. Good catch!
My guess is that output of echo command [here](https://github.com/KhronosGroup/SPIRV-LLVM-Translator/blob/ece29378764c7b16c273276e62a61efad801c633/utils/check_code_tidyness.sh#L36) is suppressed by travis somehow.

Issue #30, opened 2018-05-29
Support Vulkan? 


AlexeySotkin:
To continue tracking issue https://github.com/KhronosGroup/SPIRV-LLVM/issues/202 from the old repository.

karolherbst:
For this we have to structurize the CFG. We have something work in progress for this here: https://github.com/karolherbst/SPIRV-LLVM-Translator/commit/93c051c1d4a35d0ef178c34e6432f5f5ba094b5e

But this generates invalid structured CFG in a few cases (especially like nested loops and so on)

vron:
I have interest in this as well (in order to add a backend for gioui.org) and just wanted to check if there are any news / anyone working in this - or if this is something we should not expect to happen?

PR #200, opened 2019-05-24
Update LLVM IR metadata preprocessing

AlexeySachkov:
Main purpose of this PR is to allow SPIRV-LLVM-Translator to handle some OpenCL-specific metadata (like `reqd_work_group_size`, `intel_reqd_sub_group_size`) even if LLVM IR wasn't generated from OpenCL C (`opencl.ocl.version` metadata is missed).

Summary of changes:
* renamed `TransOCLMD` pass into `PreprocessMetadata`
* update the pass to handle some metadata unconditionally 

The first user of this PR can be found in intel/llvm#164: I've re-used existing OpenCL attribute `intel_reqd_sub_group_size` in SYCL.

I'm not particularly sure about the last commit in this PR: _Remove generating of spirv.EntryPoint metadata_ - I've removed generation of `spriv.EntryPoint` metadata since there are no uses of it. Probably it is a wrong decision. **Please speak up if you think that it is better to leave this metadata** and I will remove the commit.

### DIFF
--- a/lib/SPIRV/PreprocessMetadata.cpp
+++ b/lib/SPIRV/PreprocessMetadata.cpp
@@ -143,8 +143,7 @@ void PreprocessMetadata::visit(Module *M) {
   // The rest of metadata might come not only from OpenCL
 
   // Create metadata representing (empty so far) list
-  // of OpEntryPoint and OpExecutionMode instructions
-  auto EP = B.addNamedMD(kSPIRVMD::EntryPoint);    // !spirv.EntryPoint = {}
+  // of OpExecutionMode instructions
   auto EM = B.addNamedMD(kSPIRVMD::ExecutionMode); // !spirv.ExecutionMode = {}
 
   // Add execution modes for kernels. We take it from metadata attached to
@@ -152,16 +151,6 @@ void PreprocessMetadata::visit(Module *M) {
   for (Function &Kernel : *M) {
     if (Kernel.getCallingConv() != CallingConv::SPIR_KERNEL)
       continue;
-
-    // Add EntryPoint(which actually is adding its operands) to the list of
-    // entry points:
-    // !{i32 6, void (i32 addrspace(1)*)* @kernel, !"kernel" }
-    MDNode *EPNode;
-    EP.addOp()
-        .add(spv::ExecutionModelKernel)
-        .add(&Kernel)
-        .add(Kernel.getName())
-        .done(&EPNode);
 
     // Specifing execution modes for the Kernel and adding it to the list
     // of ExecutionMode instructions.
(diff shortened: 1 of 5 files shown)


lib/SPIRV/PreprocessMetadata.cpp
```diff
@@ -98,50 +95,55 @@ bool TransOCLMD::runOnModule(Module &Module) {
...
+        .add(CLVer)
+        .done();
+    if (EraseOCLMD)
+      B.eraseNamedMD(kSPIR2MD::OCLVer).eraseNamedMD(kSPIR2MD::SPIRVer);
+
+    // !spirv.MemoryModel = !{!x}
+    // !{x} = !{i32 1, i32 2}
+    Triple TT(M->getTargetTriple());
+    assert(isSupportedTriple(TT) && "Invalid triple");
+    B.addNamedMD(kSPIRVMD::MemoryModel)
```
bader: > Main purpose of this PR is to allow SPIRV-LLVM-Translator to handle some OpenCL-specific metadata (like reqd_work_group_size, intel_reqd_sub_group_size) even if LLVM IR wasn't generated from OpenCL C (opencl.ocl.version metadata is missed).


Can we also untie setting of addressing model from OpenCL version metadata presence?

Just out of the curiosity: do you know why we set AddressingModel based on the triple and not data layout?

AlexeySachkov: Not sure that it is a good idea. SPIR-V spec defines [two more AddressingModels](https://www.khronos.org/registry/spir-v/specs/unified1/SPIRV.html#_a_id_addressing_model_a_addressing_model) and I understand  we need to have some default value, but we already have such:
* [Dummy initial value in SPIRVModuleImpl](https://github.com/KhronosGroup/SPIRV-LLVM-Translator/blob/b46b910eeaee4ac4a8e2360a9240d68015dc3a59/lib/SPIRV/libSPIRV/SPIRVModule.cpp#L68)
* [transAddressingMode](https://github.com/KhronosGroup/SPIRV-LLVM-Translator/blob/1d9349cf53a47065ade67d86d687a78e4c5c29a4/lib/SPIRV/SPIRVWriter.cpp#L1408) which is used [here](https://github.com/KhronosGroup/SPIRV-LLVM-Translator/blob/1d9349cf53a47065ade67d86d687a78e4c5c29a4/lib/SPIRV/SPIRVWriter.cpp#L1589)

I just don't want to add another one default value which needs to be changed later if we decide to support different flavors of SPIR-V (#30).

Moreover, looks like this MemoryModel spirv-friendly metadata is also unused. BTW, AddressingModel is stored together with MemoryModel in spirv-friendly metadata, so I cannot just translate one of them without additional changes.

PR #209, opened 2019-06-03
Fix mangling of enqueue_marker

svenvh:
Minor refactoring to aid debugging of mangler substitution issues.

Fix a mangling issue for `enqueue_marker`.

### DIFF
--- a/lib/SPIRV/Mangler/Mangler.cpp
+++ b/lib/SPIRV/Mangler/Mangler.cpp
@@ -100,7 +100,7 @@ public:
       if (!mangleSubstitution(T, mangledPrimitiveString(T->getPrimitive()))) {
         size_t Index = Stream.str().size();
         Stream << mangledPrimitiveString(T->getPrimitive());
-        Substitutions[Stream.str().substr(Index)] = SeqId++;
+        recordSubstitution(Stream.str().substr(Index));
       }
     } else {
       Stream << MangledPrimitive;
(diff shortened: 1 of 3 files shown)


lib/SPIRV/Mangler/Mangler.cpp
```diff
@@ -92,15 +92,15 @@ class MangleVisitor : public TypeVisitor {
...
+    // Builtin primitives such as int are not substitution candidates, but
+    // all other primitives are.  Even though most of these do not appear
+    // repeatedly in builtin function signatures, we need to track them in
+    // the substitution map.
+    if (T->getPrimitive() >= PRIMITIVE_STRUCT_FIRST) {
       if (!mangleSubstitution(T, mangledPrimitiveString(T->getPrimitive()))) {
         size_t Index = Stream.str().size();
         Stream << mangledPrimitiveString(T->getPrimitive());
-        Substitutions[Stream.str().substr(Index)] = SeqId++;
+        recordSubstitution(Stream.str().substr(Index), SeqId++);
```
AlexeySachkov: `recordSubstitution` is a pretty simple function which just replaces call to `operator[]` - would it make sense to move usages of `SeqId` into it too? As I can see it is not used anywhere else

lib/SPIRV/Mangler/Mangler.cpp
```diff
@@ -187,6 +187,10 @@ class MangleVisitor : public TypeVisitor {
   }
 
 private:
+  void recordSubstitution(std::string Str, unsigned Id) {
```
AlexeySachkov: Why not `const std::string &Str`?

lib/SPIRV/Mangler/Mangler.cpp
svenvh: Good point, done!

lib/SPIRV/Mangler/Mangler.cpp
svenvh: Yeah why not, done! :)

PR #212, opened 2019-06-07
Cherrypick: 'Do not convert sampler to i32' and 'Align image type mangling to clang'

aratajew:
Preserve the sampler_t type for samplers instead of converting it to
i32.  This aligns the SPIR-V translator's output to clang's output,
with the aim of unifying subsequent compilation after importing SPIR-V
into the LLVM domain.

Transform OpConstantSampler operands back into
__translate_sampler_initializer calls.

test/SampledImage.ll currently loses the addrspace(2) qualifier on
sampler types after transcoding, as the sampler address space is fixed
at SPIR-V translator build time.  I believe this should not be a
concern, as other opaque types such as events also have their address
spaces fixed at translator build time.


aratajew:
This is a cherrypick commit from: https://github.com/KhronosGroup/SPIRV-LLVM-Translator/commit/6bb9231efe7677fcb1f3388897987c82629d78fe

aratajew:
This is the next cherrypick commit, from: https://github.com/KhronosGroup/SPIRV-LLVM-Translator/commit/665d001df8e4b2636fdca313c6b93658dca05799
This one is not fully cherrypicked, because the following two files:
test/transcoding/image_builtins.ll
test/transcoding/image_with_suffix.ll
are not present on release70 branch. It's not trivial to cherry pick the commit that add those files, because they depend on other ones and so on.

PR #223, opened 2019-06-14
Support addrspacecast on SPIR-V built-in variables

asavonic:
Before this patch, SPIR-V Translator assumed that a builtin global
variable can only be used by a load instruction (since all globals are
pointers). This patch adds support for addrspacecast instruction that
is followed by a load instruction.


asavonic:
@svenvh, can you merge the patch?

PR #226, opened 2019-06-17
Fix mangling for work_group_barrier and sub_group_barrier functions

aratajew:


lib/SPIRV/OCLUtil.cpp
```diff
@@ -453,10 +453,11 @@ class OCLBuiltinFuncMangleInfo : public SPIRV::BuiltinFuncMangleInfo {
         setArgAttr(0, SPIR::ATTR_CONST);
         addVoidPtrArg(0);
       }
-    } else if (UnmangledName.find("barrier") == 0 ||
-               UnmangledName.find("work_group_barrier") == 0 ||
-               UnmangledName.find("sub_group_barrier") == 0) {
+    } else if (UnmangledName.find("barrier") != std::string::npos) {
       addUnsignedArg(0);
+      if (UnmangledName.find("work_group_barrier") != std::string::npos ||
```
svenvh: Would it be safer to check for `== 0` here, to avoid triggering on user defined names like `my_custom_work_group_barrier`?  Or should we perhaps even check for an exact name match?

aratajew: Nice catch. We definitely can check for an exact name match here.

PR #227, opened 2019-06-19
Fix atomic functions pointers SPIRV -> LLVM

aratajew:
OpenCL 1.2 allows to specify local and global address spaces for pointer
arguments in atomic function. OpenCL 2.0 uses generic address space for
all of them.

### DIFF
--- a/lib/SPIRV/SPIRVToOCL20.cpp
+++ b/lib/SPIRV/SPIRVToOCL20.cpp
@@ -382,14 +382,28 @@ void SPIRVToOCL20::visitCallSPIRVAtomicBuiltin(CallInst *CI, Op OC) {
                                     .getFirstInsertionPt()));
           PExpected->setAlignment(CI->getType()->getScalarSizeInBits() / 8);
           new StoreInst(Args[1], PExpected, PInsertBefore);
-          unsigned AddrSpc = SPIRAS_Generic;
-          Type *PtrTyAS =
-              PExpected->getType()->getElementType()->getPointerTo(AddrSpc);
-          Args[1] = CastInst::CreatePointerBitCastOrAddrSpaceCast(
-              PExpected, PtrTyAS, PExpected->getName() + ".as", PInsertBefore);
+          Args[1] = PExpected;
           std::swap(Args[3], Args[4]);
           std::swap(Args[2], Args[3]);
           RetTy = Type::getInt1Ty(*Ctx);
+        }
+        // Fix atomic functions pointers:
+        // OpenCL 1.2 allows to specify local and global address spaces for
+        // pointer arguments in atomic functions. OpenCL 2.0 uses generic
+        // address space for all of them.
+        for (size_t I = 0; I < Args.size(); ++I) {
+          Value *PtrArg = Args[I];
+          if (PtrArg->getType()->isPointerTy()) {
+            // If atomic function is translated from OpenCL 2.0 version, then
+            // address space is already correct
+            if (PtrArg->getType()->getPointerAddressSpace() != 4) {
+              Type *FixedPtr =
+                  PtrArg->getType()->getPointerElementType()->getPointerTo(
+                      SPIRAS_Generic);
+              Args[I] = CastInst::CreatePointerBitCastOrAddrSpaceCast(
+                  PtrArg, FixedPtr, PtrArg->getName() + ".as", PInsertBefore);
+            }
+          }
         }
         return Name;
       },
--- a/test/transcoding/AtomicCompareExchange_cl12.ll
+++ b/test/transcoding/AtomicCompareExchange_cl12.ll
@@ -15,8 +15,9 @@ target triple = "spir-unknown-unknown"
 ; CHECK-LABEL:   entry
 ; CHECK:         [[PTR:%expected[0-9]*]] = alloca i32, align 4
 ; CHECK:         store i32 {{.*}}, i32* [[PTR]]
+; CHECK:         %object.as = addrspacecast i32 addrspace(1)* %object to i32 addrspace(4)*
 ; CHECK:         [[PTR]].as = addrspacecast i32* [[PTR]] to i32 addrspace(4)*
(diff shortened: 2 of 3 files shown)


AnastasiaStulova:
Does this mean that if someone compiles OpenCL C v1.2 source they will end up with generic AS in SPIR-V? That shouldn't happen because OpenCL v1.2 devices won't be able to import such SPIR-
V file.

aratajew:
No, it doesn't. SPIRVToOCL20 pass is used while translating from SPIRV to LLVM, so we are considering llvm output here.
I'll describe current implementation to explain the motivation:
Let's image we translate kernel from llvm -> spirv -> back to llvm.
First llvm contains 1.2 version of builtin (but kernel is in v2.0 in general) - 
int atomic_xchg ( volatile __global int *p, int val)
Then spirv module contains:
OpAtomicExchange
Then we translate back to llvm and we get:
C atomic_exchange ( volatile **__global** A  *object, int  desired)  <-- wrong AS

The builtin above is in version 2.0, because source kernel was in version 2.0 in general, but its pointer argument has __global address space which is inconsistent with specification, it should be generic (https://www.khronos.org/registry/OpenCL/sdk/2.0/docs/man/xhtml/atomic_exchange.html).

You probably will ask why we produce v2.0 builtin, since opencl c input kernel was using v1.2 builtin. That's because the whole kernel was specified as a kernel in v2.0. If it would be 1.2 version in general, then OCL20To12 pass should resolve the builtin back to v1.2.

svenvh:
> ...
> You probably will ask why we produce v2.0 builtin, since opencl c input kernel was using v1.2 builtin. That's because the whole kernel was specified as a kernel in v2.0. If it would be 1.2 version in general, then OCL20To12 pass should resolve the builtin back to v1.2.

Not sure if I follow this.  Assuming we have a kernel compiled with clang `-cl-std=CL1.2`: how/where would the OCL20To12 pass resolve the builtins back to their v1.2 version?

aratajew:
It won't currently, because there is no support for that in OCL20To12 pass, but it should work like that. If I understand SPIRVReader correctly (https://github.com/KhronosGroup/SPIRV-LLVM-Translator/blob/master/lib/SPIRV/SPIRVReader.cpp#L3038), we run SPIRVToOCL20 pass disregarding the version of OCL kernel, because there is no check there. Then we run OCL20To12 pass which checks the version of OCL (https://github.com/KhronosGroup/SPIRV-LLVM-Translator/blob/master/lib/SPIRV/OCL20To12.cpp#L81) and if it's older than 2.0, we translate some builtins back.

My change affects both OCL kernels versions:

- -cl-std=CL2.0: still produces 2.0 version of builtin, but with fixed address space - consistent with OpenCL C spec

- -cl-std=CL1.2: still produces 2.0 version of builtin (but fixed), because there is no back-translation implementation in OCL20To12 - I think we should open seperate issue tracking for that.

Please correct my if my understanding is wrong.

svenvh:
> -cl-std=CL1.2: still produces 2.0 version of builtin (but fixed), because there is no back-translation implementation in OCL20To12 - I think we should open seperate issue tracking for that.

Maybe we should address that before merging this PR?  We must avoid introducing generic address spaces in CL1.2 tool flows as @AnastasiaStulova pointed out.

There is another problem with the SPIR-V -> OCL20 -> OCL12 approach at the moment: the OCL20To12 pass is relying on `OpSource` to decide what to do, which is wrong according to the SPIR-V spec ("This has no semantic impact and can safely be removed from a module."). So I am a bit reluctant to add more steps to this staged lowering for OpenCL 1.2 until we have a reliable and valid method to select the resulting CL version. Or maybe we cannot and we need to remove the OCL20To12 pass altogether?

aratajew:
I think that we are not even able to implement SPIR-V -> OCL20 -> OCL12 approach for this case correctly, because after translation to OCL20, pointer address space will be generic and then we will not be able to decide what builtin version was used while translating to OCL12.

I think that the best approach here should be:
SPIRV -> OCL
if(pointer address space is __local or __global)
  generate v1.2 builtin
else if (pointer address space is __generic)
  generate v2.0 builtin

This is the most safe approach here, I think. What do you think about that? I could prepare another PR.

svenvh:
> I think that the best approach here should be:
> SPIRV -> OCL
> if(pointer address space is __local or __global)
> generate v1.2 builtin
> else if (pointer address space is __generic)
> generate v2.0 builtin
> 
> This is the most safe approach here, I think. What do you think about that? I could prepare another PR.

That sounds reasonable indeed, worth trying if that solves the issue for you.

AlexeySotkin:
It seems there is an ambiguity in translation of SPIRV instruction to opencl builtins. Does it make sense to control this behavior via an option? So one might choose to which builtin versions (opencl 1.2 or opencl 2.0) they want to translate.

aratajew:
Great idea. What about SPIR-V -> OCL20 -> OCL12 approach. The problem discussed here showed that it's not much reasonable approach. Some data required to obtain correct OCL12 builtin may be lost while translating to OCL20. On the other hand we probably don't have enough functional coverage in OCL12 pass to run it alone, like: SPIR-V -> OCL12.

svenvh:
Agreed, it doesn't sound trivial to recover the specific address spaces again after converting them to generic.  @AlexeySotkin's approach to have an option that selects which builtins to generate feels like the best way forward to me.

In the longer term I think we should get rid of the entire OCL20To12 pass, as I am not convinced there is a reliable way to select between CL1.2 or CL2.0 builtins from the SPIR-V file alone.  Having an option is a first step towards that direction.

aratajew:
I think we can easily merge OCL20To12 pass into SPIRVToOCL20 pass and rename it to SPIRVToOCL. Then some mutations in SPIRVToOCL pass will depend on the option you mentioned. I'll prepare PR with two commits: 
-new option to choose what version of builtins to generate 
-merge OCL20To12 to SPIRVToOCL20
Are you ok with that?

svenvh:
Sounds reasonable to me.

Issue #105, opened 2018-12-29
Enable publishing of WINDOWS built artifacts via github releases?


oscarbg:
now that we have Linux artifacts..
thanks..

AlexeySotkin:
The artifacts are posted from travis.ci, which doesn't support windows.

thewilsonator:
Could you add OSX ones as well?

oscarbg:
sorry for reviving this thread.. :-)
MacOS binaries would be nice to have..  (and not much effort to add?)
also regarding Windows seems there is some Travis.ci support already
https://blog.travis-ci.com/2018-10-11-windows-early-release
https://travis-ci.community/t/c-support-on-windows/282
@AlexeySotkin can you share if, in light of this, there is interest to try to add Win Travis support now?
thanks..

AlexeySotkin:
Hi @oscarbg, it seems interesting, but I do not plan to work on it right now. Though contributions are welcome :)

PR #230, opened 2019-06-20
Making LIT tests passing on Windows

AlexeySotkin:


AlexeySachkov:
@AlexeySotkin, don't you think that we need CI on Windows (#105) to check such patches? :)

AlexeySotkin:
I do think so :) I'm going to add it to travis a little bit later.

tremmelg:
The tests that use `spirv-val` will currently fail on Windows when `spirv-val` is not found, as `:` is not a valid command (https://github.com/KhronosGroup/SPIRV-LLVM-Translator/blob/master/test/lit.cfg.py#L61)

vladimirlaz:
> The tests that use `spirv-val` will currently fail on Windows when `spirv-val` is not found, as `:` is not a valid command (https://github.com/KhronosGroup/SPIRV-LLVM-Translator/blob/master/test/lit.cfg.py#L61)

can we just replace `:` with `echo`. It should work both on Windows and Linux.

tremmelg:
> > The tests that use `spirv-val` will currently fail on Windows when `spirv-val` is not found, as `:` is not a valid command (https://github.com/KhronosGroup/SPIRV-LLVM-Translator/blob/master/test/lit.cfg.py#L61)
> 
> can we just replace `:` with `echo`. It should work both on Windows and Linux.

Yeah, that would. work

vladimirlaz:
Do we have any chance to submit this PR? It passed testing and has approval.

PR #234, opened 2019-06-21
Fix translation of undef initializers of LLVM structures

Fznamznon:
Now if internal global constant has undef initializer SPIRV translator
will drop this initializer. It produces errors in translation back to
LLVM IR because internal global variables must have initializers.
Undef initializer for LLVM structure be can translated to OpConstantComposite
with OpUndef constituents.

### DIFF
--- a/lib/SPIRV/SPIRVWriter.cpp
+++ b/lib/SPIRV/SPIRVWriter.cpp
@@ -746,7 +746,7 @@ SPIRVValue *LLVMToSPIRV::transValueWithoutDecoration(Value *V,
       // Undef initializer for LLVM structure be can translated to
       // OpConstantComposite with OpUndef constituents.
       auto I = ValueMap.find(Init);
-      if (ValueMap.find(Init) == ValueMap.end()) {
+      if (I == ValueMap.end()) {
         std::vector<SPIRVValue *> Elements;
         for (Type *E : ST->elements())
           Elements.push_back(transValue(UndefValue::get(E), nullptr));
--- a/lib/SPIRV/SPIRVWriter.cpp
+++ b/lib/SPIRV/SPIRVWriter.cpp
@@ -728,6 +728,7 @@ SPIRVValue *LLVMToSPIRV::transValueWithoutDecoration(Value *V,
     llvm::Value *Init = GV->hasInitializer() && !GV->hasCommonLinkage()
                             ? GV->getInitializer()
                             : nullptr;
+    SPIRVValue *BVarInit = nullptr;
     StructType *ST = Init ? dyn_cast<StructType>(Init->getType()) : nullptr;
     if (ST && ST->hasName() && isSPIRVConstantName(ST->getName())) {
       auto BV = transConstant(Init);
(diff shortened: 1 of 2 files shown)


lib/SPIRV/SPIRVWriter.cpp
```diff
@@ -740,10 +741,22 @@ SPIRVValue *LLVMToSPIRV::transValueWithoutDecoration(Value *V,
         Ty = static_cast<PointerType *>(Init->getType());
       }
       Inst->dropAllReferences();
+      BVarInit = transValue(Init, nullptr);
+    } else if (ST && isa<UndefValue>(Init)) {
+      // Undef initializer for LLVM structure be can translated to
+      // OpConstantComposite with OpUndef constituents.
+      auto I = ValueMap.find(Init);
+      if (ValueMap.find(Init) == ValueMap.end()) {
```
```suggestion
      if (I == ValueMap.end()) {
```

Fznamznon: Done, thanks!

AlexeySachkov:
@AlexeySotkin, please not that Travis CI build failed for this PR.
I haven't checked, but probably due to known issues in DebugInfo tests

Also, CHECK_TIDY failed: looks like it tried to handle LIT test file: https://travis-ci.org/KhronosGroup/SPIRV-LLVM-Translator/jobs/549624445. Can we skip `test/` directory during check-tidy step?

svenvh:
> Also, CHECK_TIDY failed: looks like it tried to handle LIT test file: https://travis-ci.org/KhronosGroup/SPIRV-LLVM-Translator/jobs/549624445. Can we skip test/ directory during check-tidy step?

If the test had the correct `.ll` extension (instead of `.cpp`) then there wouldn't be a check-tidy/format problem I suppose?

AlexeySachkov:
> If the test had the correct .ll extension (instead of .cpp) then there wouldn't be a check-tidy/format problem I suppose?

Ouch, it has `.cpp` extension. I've missed that fact.
This, actually, means that test is not launched at all, according to [this line of code](https://github.com/KhronosGroup/SPIRV-LLVM-Translator/blob/master/test/lit.cfg.py#L19) LIT looks only for `.cl`, `.ll` and `.spt` files.

@Fznamznon, could you please re-commit your test with the correct extension?

Fznamznon:
> > If the test had the correct .ll extension (instead of .cpp) then there wouldn't be a check-tidy/format problem I suppose?
> 
> Ouch, it has `.cpp` extension. I've missed that fact.
> This, actually, means that test is not launched at all, according to [this line of code](https://github.com/KhronosGroup/SPIRV-LLVM-Translator/blob/master/test/lit.cfg.py#L19) LIT looks only for `.cl`, `.ll` and `.spt` files.
> 
> @Fznamznon, could you please re-commit your test with the correct extension?

Oh, of course, sorry for that.

AlexeySotkin:
The SPIRV module produced in the test doesn't pass validation. This is weird diagnostics from spirv-val, but on the other hand this is kind of a corner case. The module doesn't contain kernels or anything visible from the outside. What is the real life purpose of such module?
@Fznamznon Please add a global variable with export linkage type or a kernel to the test, so the translator will add the linkage capability.

PR #238, opened 2019-06-25
Re-commit undef_initalizer test with correct extension

Fznamznon:


Fznamznon:
> @Fznamznon Please add a global variable with export linkage type or a kernel to the test, so the translator will add the linkage capability.

Done.

PR #241, opened 2019-06-28
Reimplement the way of ocl builtins mutations

aratajew:
- Reimplement the way of ocl builtins mutations
SPIR-V -> OCL20 -> OCL12 approach needed to be reimplemented due to
important data looses while tranlating to v2.0. For some cases, it was
impossible to properly translate builtin to v1.2 if it was already
translated to v2.0.
The new approach gathers common (for v1.2 and v2.0) translation
implementation into SPIRVToOCL class. Then SPIRVToOCL12 and SPIRVToOCL20
inherits from it. It leads to the following possible translations:
1. SPIR-V -> OCL12
2. SPIR-V -> OCL20

Target version of ocl builtins is dependent on the new option "spirv-ocl-builtins-version".
- Fix ocl v2.0 atomic functions arguments address spaces


lib/SPIRV/SPIRVToOCL.cpp
```diff
@@ -0,0 +1,470 @@
...
+    else
+      assert("Invalid spirv-ocl-builtins-version value");
+  }
+  // Below part of code is here just temporarily (to not broke existing
+  // projects based on translator), because ocl builtins versions shouldn't has a dependency
+  // on OpSource spirv opcode. 
+  // OpSource spec: "This has no semantic impact and can safely be removed from a module."
+  // After some time it can be removed, then only factor impacting version of ocl builtins
+  // will be spirv-ocl-builtins-version command option.
+  unsigned OCLVersion = getOCLVersion(&M);
```
aratajew: Please, pay attention on this part of code.
I've added an option to specify target ocl builtins version, but I left the logic based on OpSource opcode as it was working before to avoid crashing the projects that are dependent on the Translator, but don't have an option specified yet. Is it a good approach, or we should remove that logic permanently in this commit?

svenvh: It sounds reasonable to have a transition period.  I wonder if we should emit a warning that this behaviour is now deprecated?

lib/SPIRV/SPIRVToOCL.h
```diff
@@ -0,0 +1,130 @@
...
+  ///       atomic_work_item_fence(flag(sema), order(sema), map(scope))
+  /// - OCL1.2: mem_fence
+  virtual void visitCallSPIRVMemoryBarrier(CallInst *CI) = 0;
+
+  static char ID;
+protected:
+  Module * M;
+  LLVMContext *Ctx;
+};
+}
```
svenvh: Github seems to think that there is a newline missing at the end of this file (and other files too), please check and fix.

lib/SPIRV/SPIRVToOCL.h
```diff
@@ -0,0 +1,130 @@
...
+/// \file
+///
+/// This file contains declaration of SPIRVToOCL class which implements
+/// common transform of SPIR-V builtins to OCL builtins.
+///
+//===----------------------------------------------------------------------===//
+
+#include "OCLUtil.h"
+#include "SPIRVInternal.h"
+#include "llvm/ADT/StringSwitch.h"
```
AlexeySachkov: Please move unused includes down to `.cpp` file:
* StringSwitch
* IRBuilder
* CommandLine
* Debug
* something else?

lib/SPIRV/SPIRVToOCL.cpp
```diff
@@ -0,0 +1,466 @@
...
+// Some of the visit functions are translations to OCL2.0 builtins, but they
+// are currently used also for OCL1.2, so theirs implementations are placed
+// in this pass as a common functionality for both versions.
+//
+//===----------------------------------------------------------------------===//
+#define DEBUG_TYPE "spvtocl"
+
+#include "SPIRVToOCL.h"
+
+namespace SPIRV{
```
AlexeySachkov: Please apply clang-format

lib/SPIRV/SPIRVToOCL.cpp
```diff
@@ -0,0 +1,470 @@
...
+
+static cl::opt<std::string>
+MangledAtomicTypeNamePrefix("spirv-atomic-prefix",
+  cl::desc("Mangled atomic type name prefix"),
+  cl::init("U7_Atomic"));
+
+static cl::opt<unsigned int>
+OCLBuiltinsVersion("spirv-ocl-builtins-version",
+  cl::desc("Specify version of OCL builtins to translate to (ex. 12, 20, 21)"));
```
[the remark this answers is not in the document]
aratajew: Currently we depend on ocl version specified in llvm metadata opencl.ocl.version. If it's not specified, then getOCLVersion(&M) returns 0 and this leads to applying OCL1.2 builtins translation, because there is the following logic in OCL20to12.cpp:
`if (getOCLVersion(M) >= kOCLVer::CL20)
    return false;`

So, should we set the default value for this option to CL1.2?

lib/SPIRV/SPIRVToOCL.cpp
```diff
@@ -0,0 +1,466 @@
...
+  if (isPipeOpCode(OC)) {
+    visitCallSPIRVPipeBuiltin(&CI, OC);
+    return;
+  }
+  if (OCLSPIRVBuiltinMap::rfind(OC))
+    visitCallSPIRVBuiltin(&CI, OC);
+}
+
+void SPIRVToOCL::visitCastInst(CastInst &Cast) {
+  if (!isa<ZExtInst>(Cast) && !isa<SExtInst>(Cast) && !isa<TruncInst>(Cast) &&
```
AlexeySachkov: Looks like it is easier to enumerate allowed casts then disallowed ones:

```c++
  if (isa<AddrSpacecastInst>(Cast) || isa<BitCastInst>(Cast) || isa<PtrToIntCast>(Cast) || isa<IntToPtrCast>(Cast))
    return;
```

lib/SPIRV/SPIRVToOCL.cpp
```diff
@@ -0,0 +1,466 @@
...
+    !isa<FPTruncInst>(Cast) && !isa<FPExtInst>(Cast) &&
+    !isa<FPToUIInst>(Cast) && !isa<FPToSIInst>(Cast) &&
+    !isa<UIToFPInst>(Cast) && !isa<SIToFPInst>(Cast))
+    return;
+
+  Type const *SrcTy = Cast.getSrcTy();
```
AlexeySachkov:
```suggestion
  const Type *SrcTy = Cast.getSrcTy();
```

lib/SPIRV/SPIRVToOCL.cpp
aratajew: Right, do we have any specific mechanism that could let us to generate such warning?

lib/SPIRV/SPIRVToOCL.cpp
```diff
@@ -0,0 +1,466 @@
...
+  else if (ImgTyName.startswith("opencl.image3d")) {
+    ImgDim = 3;
+  }
+  assert(ImgDim != 0 && "unexpected image dimensionality");
+
+  if (ImgTyName.count("_array_") != 0) {
+    ImgArray = true;
+  }
+
+  AttributeList Attributes = CI->getCalledFunction()->getAttributes();
```
AlexeySachkov:
```suggestion
  AttributeList Attributes = Func->getAttributes();
```

include/LLVMSPIRVLib.h
```diff
@@ -59,6 +59,7 @@ void initializeSPIRVLowerSPIRBlocksPass(PassRegistry &);
 void initializeSPIRVLowerOCLBlocksPass(PassRegistry &);
```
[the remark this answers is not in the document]
aratajew: Done

lib/SPIRV/SPIRVToOCL.h
aratajew: Done

lib/SPIRV/SPIRVToOCL.cpp
```diff
@@ -0,0 +1,465 @@
...
+    // The width of integer type returning by OpImageQuerySize[Lod] may
+    // differ from i32
+    if (CI->getType()->getScalarType() != Int32Ty) {
+      GetImageSize = CastInst::CreateIntegerCast(GetImageSize,
+        CI->getType()->getScalarType(),
+        false, CI->getName(), CI);
+    }
+  }
+  else {
+    assert((ImgDim == 2 || ImgDim == 3) && "invalid image type");
```
AlexeySachkov: I don't see any value in this assertion. `ImgDim` might be zero, but the fact that it is not zero has been asserted already

lib/SPIRV/SPIRVToOCL.cpp
```diff
@@ -0,0 +1,465 @@
...
+
+  bool HasGroupOperation = hasGroupOperation(OC);
+  if (!HasGroupOperation) {
+    DemangledName = Prefix + DemangledName;
+  }
+  else {
+    auto GO = getArgAs<spv::GroupOperation>(CI, 1);
+    StringRef Op = DemangledName;
+    Op = Op.drop_front(strlen(kSPIRVName::GroupPrefix));
+    bool Unsigned = Op.front() == 'u';
```
AlexeySachkov:
```suggestion
    bool Unsigned = Op.startswith("u");
```

lib/SPIRV/SPIRVToOCL.cpp
AlexeySachkov: We have mechanism to emit errors, but I'm afraid that we don't have anything for warnings.

At least, we can document this change somewhere in README or in the special file which describes translation from SPIR-V to LLVM IR for OpenCL

lib/SPIRV/SPIRVToOCL.cpp
aratajew: All of the code in this file is just copied from SPIRVToOCL20.cpp file, so it's not something that I wrote. Should we fix all of that things in this pull request?

AlexeySachkov: I don't have strong opinion on that. If you have time to improve things, it would be nice. Otherwise, you can skip my comments if code wasn't written by you right now.

But please apply clang-format: it is now a part of Travis CI checks

lib/SPIRV/SPIRVToOCL.cpp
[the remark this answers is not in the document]
svenvh: Okay, then CL1.2 as a default makes sense.

lib/SPIRV/SPIRVToOCL.cpp
svenvh: Only if you want to I'd say... :-)  But please do that in a separate commit.

svenvh: Also applying the reformatting would be better in a separate commit as part of this pull request.

aratajew: Reformatting shouldn't be in a seperate comment I think. It's incorrect because of my IDE formatting settings different than clang-format, that not because original SPIRVToOCL20.cpp file was formatted incorrectly.

svenvh: Oh right, if the original formatting was fine then you should fix the formatting in the same commit indeed.

aratajew:
- Formatting fixed

- Seperated changes reponsible for additional option '-spirv-ocl-builtins-version'

I'm going to open a seperate PR for small changes suggested by Alexey Sotkin, because I want to have changes related to builtins approach reimplementation tested separately.

lib/SPIRV/SPIRVToOCL.h
```diff
@@ -0,0 +1,131 @@
...
+#include "llvm/IR/Instructions.h"
+#include "llvm/IR/Verifier.h"
+#include "llvm/Pass.h"
+#include "llvm/PassSupport.h"
+#include "llvm/Support/CommandLine.h"
+#include "llvm/Support/Debug.h"
+
+#include <cstring>
+
+using namespace llvm;
```
Naghasan: Please don't use `using namespace` in a header files

lib/SPIRV/SPIRVToOCL.h
Naghasan: `Instructions` and `Verifier` ?

Types coming from Instructions can be forward declared

aratajew: Done

lib/SPIRV/SPIRVToOCL.cpp
aratajew: Added short info about new option to README. Don't know if it's enough.

lib/SPIRV/SPIRVToOCL.h
aratajew: Fixed

README.md
```diff
@@ -89,6 +89,8 @@ To translate between LLVM IR and SPIR-V:
     ```
     llvm-spirv -r input.spv
     ```
+    Recommendations:
+    * `-spirv-ocl-builtins-version` - to specify target version of OCL builtins to translate to (default CL1.2)
```
svenvh: Recommendations -> Recommended options
of OCL builtins -> of OpenCL builtins

lib/SPIRV/SPIRVToOCL20.cpp
```diff
@@ -211,350 +97,32 @@ void SPIRVToOCL20::visitCallSPIRVMemoryBarrier(CallInst *CI) {
...
-        assert(isa<PointerType>(T));
-        auto ET = T->getPointerElementType();
-        if (!ET->isIntegerTy(8) ||
-            T->getPointerAddressSpace() != SPIRAS_Generic) {
-          auto NewTy = PointerType::getInt8PtrTy(*Ctx, SPIRAS_Generic);
-          P = CastInst::CreatePointerBitCastOrAddrSpaceCast(P, NewTy, "", CI);
+        for (size_t I = 0; I < Args.size(); ++I) {
+          Value *PtrArg = Args[I];
+          if (PtrArg->getType()->isPointerTy()) {
+            if (PtrArg->getType()->getPointerAddressSpace() != 4) {
```
svenvh: `4` -> `SPIRAS_Generic`

lib/SPIRV/SPIRVToOCL20.cpp
```diff
@@ -211,350 +97,32 @@ void SPIRVToOCL20::visitCallSPIRVMemoryBarrier(CallInst *CI) {
...
-        auto T = P->getType();
-        assert(isa<PointerType>(T));
-        auto ET = T->getPointerElementType();
-        if (!ET->isIntegerTy(8) ||
-            T->getPointerAddressSpace() != SPIRAS_Generic) {
-          auto NewTy = PointerType::getInt8PtrTy(*Ctx, SPIRAS_Generic);
-          P = CastInst::CreatePointerBitCastOrAddrSpaceCast(P, NewTy, "", CI);
+        for (size_t I = 0; I < Args.size(); ++I) {
+          Value *PtrArg = Args[I];
+          if (PtrArg->getType()->isPointerTy()) {
```
svenvh: You're using `PtrArg->getType()` a few times, so the code might be more readable if you factor it out into a new variable.

lib/SPIRV/SPIRVToOCL.cpp
```diff
@@ -0,0 +1,471 @@
...
+
+static cl::opt<std::string>
+    MangledAtomicTypeNamePrefix("spirv-atomic-prefix",
+                                cl::desc("Mangled atomic type name prefix"),
+                                cl::init("U7_Atomic"));
+
+static cl::opt<std::string>
+    OCLBuiltinsVersion("spirv-ocl-builtins-version",
+                       cl::desc("Specify version of OCL builtins to translate "
+                                "to (ex. CL1.2, CL2.0, CL2.1)"));
```
svenvh: Remove `ex.`, as these are now the only accepted values.

lib/SPIRV/SPIRVToOCL20.cpp
```diff
@@ -211,350 +97,32 @@ void SPIRVToOCL20::visitCallSPIRVMemoryBarrier(CallInst *CI) {
...
+  return SPIRVToOCL::visitCallSPIRVAtomicBuiltin(CiFixedAs, OC);
 }
 
 } // namespace SPIRV
 
 INITIALIZE_PASS(SPIRVToOCL20, "spvtoocl20",
                 "Translate SPIR-V builtins to OCL 2.0 builtins", false, false)
 
-ModulePass *llvm::createSPIRVToOCL20() { return new SPIRVToOCL20(); }
+ModulePass *llvm::createSPIRVToOCL20() { return new SPIRVToOCL20(); }
```
svenvh: GitHub is still showing that files are not ending with a newline, please fix before merging.

svenvh:
The last commit in this PR should also add a test for the `spirv-ocl-builtins-version` option values `CL2.0` and `CL2.1` (or update an existing test).

aratajew:
> The last commit in this PR should also add a test for the `spirv-ocl-builtins-version` option values `CL2.0` and `CL2.1` (or update an existing test).

You meant middle commit (908ace9), not last, right? I can add some test that translates spv to llvm with different `spirv-ocl-builtins-version` values. By different I mean `OCL1.2` and `OCL2.0`, because those are currently the versions that differ.

lib/SPIRV/SPIRVToOCL20.cpp
aratajew: Fixed.

lib/SPIRV/SPIRVToOCL.cpp
aratajew: Removed.

lib/SPIRV/SPIRVToOCL20.cpp
aratajew: Refactored.

README.md
aratajew: Done

lib/SPIRV/SPIRVToOCL20.cpp
aratajew: Done

svenvh:
> > The last commit in this PR should also add a test for the `spirv-ocl-builtins-version` option values `CL2.0` and `CL2.1` (or update an existing test).
> 
> You meant middle commit ([908ace9](https://github.com/KhronosGroup/SPIRV-LLVM-Translator/commit/908ace9bf4a364dc656b18e89030037f27f66f35)), not last, right?

I meant the last commit, because that commit is changing the behavior of the translator in case it is invoked with `-spirv-ocl-builtins-version=CL2.0`; so the new behavior should be tested.  The commits before don't really change the behavior (besides adding a new option).

> I can add some test that translates spv to llvm with different `spirv-ocl-builtins-version` values. By different I mean `OCL1.2` and `OCL2.0`, because those are currently the versions that differ.

Actually that's a good idea.

aratajew:
I've written a test for `spirv-ocl-builtins-version` and output llvm wasn't different for `OCL1.2` and `OCL2.0`. I started debugging it, and it turned out that there are other part of translator that are dependent on OpSource, not only SPIRVToOCL20 and OCL20ToOCL12 passes.
Here is the example: https://github.com/KhronosGroup/SPIRV-LLVM-Translator/blob/master/lib/SPIRV/SPIRVReader.cpp#L2803
I think we should provide two options: SourceLanguage and SourceLanguageVersion, not only option specific for OCL.
SPIRV Spec:
"The source language is for debug purposes only, with no semantics that affect the meaning of other parts of the module.
Used by OpSource"

What do you think about adding the following two options:

- SPIRVSourceLang, possible values: "ESSL", "GLSL", "OpenCL_C", "OpenCL_CPP" and "HLSL" as it is in spirv specification 3.2

- SPIRVSourceLangVersion, don't exactly know what possible values should be here, for APIs different than OCL

aratajew:
I considered that and I think we should get rid of all source-version dependent translations from SPIRVReader. It should all be resolved in SPIRVToOCL20 and SPIRVToOCL12 passes. I'll try to refactor that, so `spirv-ocl-builtins-version` can stay as it is.

svenvh:
> I considered that and I think we should get rid of all source-version dependent translations from SPIRVReader. It should all be resolved in SPIRVToOCL20 and SPIRVToOCL12 passes. I'll try to refactor that, so `spirv-ocl-builtins-version` can stay as it is.

That sounds good to me, thanks for looking into that!

aratajew:
- Moved translation logic dependent on ocl version from SPIRVReader to version-dependent passes (SPIRVToOCL12 and SPIRVToOCL20).
- Added test for `spirv-ocl-builtins-version` option

lib/SPIRV/SPIRVToOCL.cpp
```diff
@@ -440,6 +445,23 @@ Instruction *SPIRVToOCL::visitCallSPIRVAtomicBuiltin(CallInst *CI, Op OC) {
 } // namespace SPIRV
 
 ModulePass *llvm::createSPIRVToOCL(Module &M) {
+  if (OCLBuiltinsVersion.getNumOccurrences() > 0) {
+    if (OCLBuiltinsVersion.getValue() == "CL1.2")
+      return createSPIRVToOCL12();
+    else if (OCLBuiltinsVersion.getValue() == "CL2.0" ||
+             OCLBuiltinsVersion.getValue() == "CL2.1")
```
AlexeySachkov: Note: there is no such thing as OpenCL C 2.1

lib/SPIRV/SPIRVToOCL12.cpp
```diff
@@ -91,6 +96,24 @@ void SPIRVToOCL12::visitCallSPIRVMemoryBarrier(CallInst *CI) {
       &Attrs);
 }
 
+void SPIRVToOCL12::visitCallSPIRVControlBarrier(CallInst *CI) {
+  AttributeList Attrs = CI->getCalledFunction()->getAttributes();
+  Attrs = Attrs.addAttribute(CI->getContext(), AttributeList::FunctionIndex,
+                             Attribute::NoDuplicate);
+  CallInst *NewCI = mutateCallInstOCL(
```
aratajew: Will remove variable assignment from here, as NewCI is unused.

lib/SPIRV/SPIRVToOCL.cpp
aratajew: Right. Fixed.

aratajew:
It would be great to commit those changes as fast as it's possible, because I need to manually merge all the changes that goes into SPIRVToOCL20.cpp file, as those are moved into SPIRVToOCL.cpp file.
@svenvh could you please have a look?

lib/SPIRV/SPIRVToOCL12.cpp
svenvh: Are you still going to remove this variable assignment?

aratajew:
Variable assignment removed, thanks for being perceptive.

PR #245, opened 2019-07-05
Map memory semantics and memory order enums in runtime

AlexeySotkin:
The "memory_order" and "memory_semantics" arguments of atomic OpenCL builtins
are not always compile/translation time known constants. In this case can map
OpenCL enums to SPIR-V enums in runtime.
The same is applicable for translation from SPIR-V to LLVM IR. Scope<id> and
MemorySemantics<id> operands are 32-bit integer scalars. They are guaranteed
to be OpConstants only for Shader capabilites, not for Kernel.


AlexeySachkov:
I will take care of this patch instead of @AlexeySotkin. New PR with fixes: #247

PR #246, opened 2019-07-09
Reimplement the way of ocl builtins mutations llvm_release_80

aratajew:
- Reimplement the way of ocl builtins mutations

- Select target ocl builtins version based on new spirv-ocl-builtins-version option [SPIRV->LLVM]

- Fix ocl v2.0 atomic functions arguments address spaces

- Move OpControlBarrier and OpMemoryBarrier translation into SPIRVToOCL passes


aratajew:
@svenvh Could you please take a look, it's an analogical implementation to the changes that are already committed to master. Thanks.

PR #247, opened 2019-07-09
Map memory semantics and memory order enums in runtime

AlexeySachkov:
The "memory_order" and "memory_semantics" arguments of atomic OpenCL builtins
are not always compile/translation time known constants. In this case can map
OpenCL enums to SPIR-V enums in runtime.
The same is applicable for translation from SPIR-V to LLVM IR. Scope<id> and
MemorySemantics<id> operands are 32-bit integer scalars. They are guaranteed
to be OpConstants only for Shader capabilites, not for Kernel.

This is a re-upload of #245 with fixes for LIT test, clang-format and clang-tidy + applied comments from myself

@svenvh, please review


AlexeySachkov:
@svenvh, I'm sorry for committing this without waiting for your review.

Please report any issues that you will find or any comments that you have. Me or @AlexeySotkin will apply them as follow-up PRs

svenvh:
No problem!  Sorry I was out of office but LGTM, thanks!

PR #249, opened 2019-07-11
Fix mangling for cl_intel_subgroups functions when reading SPIRV

turinevgeny:
Mangle uint parameters according to 
[cl_intel_subgroups](https://www.khronos.org/registry/OpenCL/extensions/intel/cl_intel_subgroups.html) spec.


turinevgeny:
Fixed clang-format issues.

PR #254, opened 2019-07-15
Translate __fpga_reg builtin to FPGARegINTEL instruction

AGindinson:
LLVM IR representation produced by SYCL Clang boils down to llvm.annotation.* and llvm.ptr.annotation.* intrinsic calls accessing "__builtin_intel_fpga_reg" global constant string literal.

Extension specification can be found at:
https://github.com/KhronosGroup/SPIRVRegistry/blob/master/extensions/INTEL/SPV_INTEL_fpga_reg.asciidoc                                                                                                                                                                                                                   

Original patch author: Viktoria Maksimova <viktoria.maksimova@intel.com>


AGindinson:
> @AGindinson please add a test.

@AlexeySotkin Please see test/IntelFPGAReg.ll - it may not have loaded in the review due to the large file size

lib/SPIRV/SPIRVReader.cpp
```diff
@@ -1760,6 +1760,40 @@ Value *SPIRVToLLVM::transValueWithoutDecoration(SPIRVValue *BV, Function *F,
...
+    Value *UndefInt8Ptr = UndefValue::get(Int8PtrTyPrivate);
+    Value *UndefInt32 = UndefValue::get(Int32Ty);
+
+    StringRef AnnotStr("__builtin_intel_fpga_reg");
+    Constant *GS = BB->getModule()->getNamedGlobal("__reg");
+    if (!GS)
+      GS = Builder.CreateGlobalStringPtr(AnnotStr, "__reg");
+
+    Function *AnnotationFn;
+    if (auto RetIntTy = dyn_cast<IntegerType>(transType(BC->getType()))) {
```
AlexeySotkin: Please move transType call to a separate line, just not to call it twice in case the 'if' condition will fail.

lib/SPIRV/SPIRVReader.cpp
```diff
@@ -1760,6 +1760,40 @@ Value *SPIRVToLLVM::transValueWithoutDecoration(SPIRVValue *BV, Function *F,
...
+
+    Value *UndefInt8Ptr = UndefValue::get(Int8PtrTyPrivate);
+    Value *UndefInt32 = UndefValue::get(Int32Ty);
+
+    StringRef AnnotStr("__builtin_intel_fpga_reg");
+    Constant *GS = BB->getModule()->getNamedGlobal("__reg");
+    if (!GS)
+      GS = Builder.CreateGlobalStringPtr(AnnotStr, "__reg");
+
+    Function *AnnotationFn;
```
AlexeySotkin: Please initialize by nullptr.

lib/SPIRV/SPIRVReader.cpp
```diff
@@ -1760,6 +1760,40 @@ Value *SPIRVToLLVM::transValueWithoutDecoration(SPIRVValue *BV, Function *F,
...
+      AnnotationFn =
+          llvm::Intrinsic::getDeclaration(M, Intrinsic::annotation, RetIntTy);
+    } else if (auto RetPtrTy =
+                   dyn_cast<PointerType>(transType(BC->getType()))) {
+      AnnotationFn = llvm::Intrinsic::getDeclaration(
+          M, Intrinsic::ptr_annotation, RetPtrTy);
+    }
+
+    llvm::Value *Args[] = {transValue(BC->getOperand(0), F, BB),
+                           Builder.CreateBitCast(GS, Int8PtrTyPrivate),
```
AlexeySotkin: Are you sure you need this bitcast? CreateGlobalStringPtr returns i8*.

lib/SPIRV/SPIRVReader.cpp
```diff
@@ -1760,6 +1760,40 @@ Value *SPIRVToLLVM::transValueWithoutDecoration(SPIRVValue *BV, Function *F,
...
+                           UndefInt8Ptr, UndefInt32};
+    auto Call = Builder.CreateCall(AnnotationFn, Args);
```
AlexeySotkin: Also you could use `Builder.CreateIntrinsic`

lib/SPIRV/SPIRVWriter.cpp
```diff
@@ -1322,6 +1322,19 @@ SPIRVValue *LLVMToSPIRV::transIntrinsicInst(IntrinsicInst *II,
     return DbgTran->createDebugDeclarePlaceholder(cast<DbgDeclareInst>(II), BB);
   case Intrinsic::dbg_value:
     return DbgTran->createDebugValuePlaceholder(cast<DbgValueInst>(II), BB);
+  case Intrinsic::annotation: {
+    SPIRVType *Ty = transType(II->getType());
+
+    GetElementPtrInst *GEP = cast<GetElementPtrInst>(II->getArgOperand(1));
```
AlexeySotkin: Are you sure it is *always* GEP ?

lib/SPIRV/SPIRVReader.cpp
```diff
@@ -1760,6 +1760,40 @@ Value *SPIRVToLLVM::transValueWithoutDecoration(SPIRVValue *BV, Function *F,
...
+    SPIRVFPGARegINTELInstBase *BC =
+        static_cast<SPIRVFPGARegINTELInstBase *>(BV);
+
+    Type *Int8PtrTyPrivate = Type::getInt8PtrTy(*Context, SPIRAS_Private);
+    IntegerType *Int32Ty = Type::getInt32Ty(*Context);
+
+    Value *UndefInt8Ptr = UndefValue::get(Int8PtrTyPrivate);
+    Value *UndefInt32 = UndefValue::get(Int32Ty);
+
+    StringRef AnnotStr("__builtin_intel_fpga_reg");
```
AlexeySotkin: Please move the bulitin name string to OCLUtil.h

lib/SPIRV/SPIRVReader.cpp
```diff
@@ -1760,6 +1760,40 @@ Value *SPIRVToLLVM::transValueWithoutDecoration(SPIRVValue *BV, Function *F,
...
+    Constant *GS = BB->getModule()->getNamedGlobal("__reg");
```
AlexeySotkin: Why "__reg" ?

lib/SPIRV/SPIRVWriter.cpp
```diff
@@ -1352,28 +1365,34 @@ SPIRVValue *LLVMToSPIRV::transIntrinsicInst(IntrinsicInst *II,
...
-    std::vector<std::pair<Decoration, std::string>> Decorations =
-        parseAnnotations(AnnotationString);
-
-    if (Decorations.empty()) {
-      Ty->addMemberDecorate(new SPIRVMemberDecorateUserSemanticAttr(
-          Ty, MemberNumber,
-          AnnotationString.substr(0, AnnotationString.size() - 1)));
+        dyn_cast<ConstantDataArray>(C->getOperand(0))->getAsCString();
+
+    SPIRVType *Ty;
```
AlexeySotkin: Please initialize with nullptr or move in scope of if-else statement.

lib/SPIRV/SPIRVReader.cpp
AGindinson: As a submitter, I could speculate that this is to avoid possible collisions with the already-initialized named globals. @asavonic, could you please confirm/correct me here?

lib/SPIRV/SPIRVWriter.cpp
AGindinson: This is a very good catch. I should fix this to dyn_cast & add a check

lib/SPIRV/SPIRVReader.cpp
AGindinson: This has to do with the address space coherence. `CreateGlobalStringPtr` returns the global pointer, so `CreateBitCast` casts the pointer to `Int8PtrPrivate`, which is in turn declared with `SPIRAS_Private` argument. Additionally, please see `Int8PtrTyPrivate` declaration at line 1772.

lib/SPIRV/SPIRVReader.cpp
AGindinson: I'm confused as to where I should put it there. Do you mean the `kOCLBuiltinName` namespace?

lib/SPIRV/SPIRVReader.cpp
asavonic: Not sure I understand either. 
What if we already happen to have a global named "__reg" with a different content? This code is going to use this totally unrelated global. If that is an attempt to cache string constants, then IRBuilder.CreateGlobalStringPtr should do this.

lib/SPIRV/SPIRVReader.cpp
asavonic: If you want to cast between address spaces, then you need IRBuilder.CreateAddrSpaceCast.
Anyway, I don't think you should do this in the first place. Global and private address spaces do not overlap, and an addrspacecast between them would be is invalid. Can the annotate intrinsic take a pointer in global address space?

AGindinson: After consideration, I don't see why the intrinsic should not accept a global AS pointer. Will fix despite the natural "it works though!" :)

asavonic: According to Intrinsics.td, annotation intrinsic only supports pointers in (0) address space, so this cast is required.

AlexeySotkin: >  `CreateGlobalStringPtr`  returns the global pointer

Not necessarily.

Actually `CreateGlobalStringPtr` takes address space as a parameter. See here https://github.com/llvm/llvm-project/blob/52c39396151978ca946e2a80d9118c8672bace14/llvm/include/llvm/IR/IRBuilder.h#L1861-L1862
Moreover its default value is 0.

lib/SPIRV/SPIRVReader.cpp
AGindinson: I had reworked it in a resemblant way, but unfortunately, we still have to perform the cast of `Ty`. Because of this, going with the original flow ends up simpler.

lib/SPIRV/SPIRVReader.cpp
AGindinson: This was a very dangerous workaround that would produce ill code should the variable already exist

lib/SPIRV/SPIRVWriter.cpp
```diff
@@ -1322,6 +1322,22 @@ SPIRVValue *LLVMToSPIRV::transIntrinsicInst(IntrinsicInst *II,
...
+
+    GetElementPtrInst *GEP = dyn_cast<GetElementPtrInst>(
+                                 II->getArgOperand(1));
+    if (!GEP)
+      return nullptr;
+    Constant *C = cast<Constant>(GEP->getOperand(0));
+    StringRef AnnotationString =
+        cast<ConstantDataArray>(C->getOperand(0))->getAsCString();
+
+    if (AnnotationString == "__builtin_intel_fpga_reg")
```
AlexeySotkin:
```suggestion
    if (AnnotationString == kOCLBuiltinName::FPGARegIntel)
```

lib/SPIRV/SPIRVWriter.cpp
```diff
@@ -1352,28 +1368,33 @@ SPIRVValue *LLVMToSPIRV::transIntrinsicInst(IntrinsicInst *II,
...
+            Ty, MemberNumber, AnnotationString));
+      } else {
+        addIntelFPGADecorationsForStructMember(Ty, MemberNumber, Decorations);
+      }
+      II->replaceAllUsesWith(II->getOperand(0));
     } else {
-      addIntelFPGADecorationsForStructMember(Ty, MemberNumber, Decorations);
+      auto *Ty = transType(II->getType());
+      auto *BI = dyn_cast<BitCastInst>(II->getOperand(0));
+      if (AnnotationString == "__builtin_intel_fpga_reg")
```
AlexeySotkin:
```suggestion
      if (AnnotationString == kOCLBuiltinName::FPGARegIntel)
```

lib/SPIRV/SPIRVReader.cpp
```diff
@@ -1760,6 +1760,38 @@ Value *SPIRVToLLVM::transValueWithoutDecoration(SPIRVValue *BV, Function *F,
...
+
+    CallInst *IntrinsicCall = nullptr;
+    if (auto *RetIntTy = dyn_cast<IntegerType>(Ty))
+      IntrinsicCall =
+          Builder.CreateIntrinsic(Intrinsic::annotation, RetIntTy, Args);
+    else if (auto *RetPtrTy = dyn_cast<PointerType>(Ty))
+      if (isa<IntegerType>(RetPtrTy->getElementType()))
+        IntrinsicCall =
+            Builder.CreateIntrinsic(Intrinsic::ptr_annotation, RetPtrTy, Args);
+    assert(IntrinsicCall &&
```
AlexeySotkin: According to the extension spec, OpFPGARegINTEL can have operands of any type. This implementation is limited to integer and pointer to integer types.
Should we insert a bitcast for other types here (to make it consumable by llvm intrinsics) ?

AGindinson: For the current scenarios, no other type is going to arrive to SPIR-V, as Translator creates FPGARegIntel from the same LLVM intrinsics anyway. However, I agree I should bitcast here to lessen dependency on LLVM->SPIR-V translation module and achieve spec compliance.

lib/SPIRV/SPIRVWriter.cpp
```diff
@@ -1322,6 +1322,22 @@ SPIRVValue *LLVMToSPIRV::transIntrinsicInst(IntrinsicInst *II,
     return DbgTran->createDebugDeclarePlaceholder(cast<DbgDeclareInst>(II), BB);
   case Intrinsic::dbg_value:
     return DbgTran->createDebugValuePlaceholder(cast<DbgValueInst>(II), BB);
+  case Intrinsic::annotation: {
+    SPIRVType *Ty = transType(II->getType());
+
+    GetElementPtrInst *GEP = dyn_cast<GetElementPtrInst>(
+                                 II->getArgOperand(1));
```
AlexeySotkin: Is it possible to use getConstantStringInfo() to get AnnotationString from the operand.

AGindinson: Personally, I would also prefer it done through `getConstantStringInfo()`. However, the original approach is consistent with handling of `var_annotation` and `ptr_annotation`. I believe refactoring all three had better be done separately

AlexeySotkin: Please add a TODO comment as a reminder to do re-factoring and possible using of `getConstantStringInfo()`

PR #258, opened 2019-07-16
Fix build warnings

svenvh:
Fix -Wimplicit-fallthrough warning

 * `unannotated fall-through between switch labels [-Wimplicit-fallthrough]`

Fix -Wcovered-switch-default warning

 * `default label in switch which covers all enumeration values [-Wcovered-switch-default]`

### DIFF
--- a/lib/SPIRV/libSPIRV/SPIRVInstruction.h
+++ b/lib/SPIRV/libSPIRV/SPIRVInstruction.h
@@ -1522,7 +1522,7 @@ protected:
     case OpTypeStruct:
       break;
     default:
-      static_assert("Invalid type", "");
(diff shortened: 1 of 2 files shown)


lib/SPIRV/libSPIRV/SPIRVInstruction.h
```diff
@@ -1517,6 +1517,7 @@ class SPIRVCompositeConstruct : public SPIRVInstruction {
     case OpTypeVector:
       assert(getConstituents().size() > 1 &&
              "There must be at least two Constituent operands in vector");
+      // Fallthrough
```
StuartDBrady: This would probably be better as `break;` instead of the annotation.

PR #259, opened 2019-07-16
Adding bitreverse/bitcount translation and improving test coverage

svenvh:
This improves coverage of various SPIR-V operations and translator code:

* Validate function control field
* Translate OpBitReverse
* Translate OpBitCount

Add or extend tests to improve testing coverage of the following.
The translator already supports these, but there were no tests
yet.

 - OpConstantFalse,
 - OpConvertPtrToU,
 - OpSatConvertUToS and OpSatConvertSToU


test/transcoding/bit_ops.cl
```diff
@@ -1,4 +1,4 @@
-// RUN: %clang_cc1 -triple spir-unknown-unknown -O1 -cl-std=CL2.0 -emit-llvm-bc %s -o %t.bc
+// RUN: %clang_cc1 -triple spir-unknown-unknown -O1 -cl-std=CL2.0 -finclude-default-header -emit-llvm-bc %s -o %t.bc
```
AlexeySotkin: Including the default headers slow down compilation and increase test execution time significantly. Could you just copy/paste the declaration for popcount from the header?

svenvh: I could, though longer term I want to switch all .cl tests to use the fast tablegen-based builtin lookup (which is currently under development in clang).  So the test execution time increase is only temporary; do you still want me to change it?

AlexeySotkin: No, I think we can leave it as is.

PR #260, opened 2019-07-17
Refactor the way of atomics builtins mutations

aratajew:
This PR contains the following changes:

- 5a8e4e9 : Our target is to seperate the code responsible for ocl-specific mutations into SPIRVToOCL-specific passes. With this change it's easier to introduce different logic for OpenCL 2.0 and OpenCL1.2 mutations.
- 6c91d56 : Fix translation of OpAtomicIIncrement and OpAtomicIDecrement opcodes for OpenCL 1.2. Those builtins were translated into OpenCL 2.0 version (even if target version was 1.2), but with wrong pointer address space (not generic). So the output for 1.2 kernel was incorrect 2.0 version of builtin.
This patch introduces translation to atomic_inc and atomic_dec builtins for OpenCL1.2.

### DIFF
--- a/test/transcoding/atomics_1.2.ll
+++ b/test/transcoding/atomics_1.2.ll
@@ -16,10 +16,10 @@ target triple = "spir64-unknown-unknown"
 ; Function Attrs: nounwind
 define spir_kernel void @test_atomic_global(i32 addrspace(1)* %dst) #0 !kernel_arg_addr_space !1 !kernel_arg_access_qual !2 !kernel_arg_type !3 !kernel_arg_base_type !5 !kernel_arg_type_qual !4 {
   ; atomic_inc
-  %inc_ig = tail call spir_func i32 @_Z10atomic_incPVU3AS1i(i32 addrspace(1)* %dst) #0
-  ; CHECK: _Z25atomic_fetch_add_explicitPU3AS1VU7_Atomicii12memory_order12memory_scope(i32 addrspace(1)* {{.*}}, i32 1
-  %dec_jg = tail call spir_func i32 @_Z10atomic_decPVU3AS1j(i32 addrspace(1)* %dst) #0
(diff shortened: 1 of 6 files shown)


test/transcoding/atomics_1.2.ll
```diff
@@ -80,12 +80,12 @@ define spir_kernel void @test_atomic_global(i32 addrspace(1)* %dst) #0 !kernel_a
...
+  ; CHECK: _Z10atomic_incPU3AS3Vi(i32 addrspace(3)* %dst) #0
 
   ; atomic dec
-  %dec_jl = tail call spir_func i32 @_Z10atomic_decPVU3AS3j(i32 addrspace(3)* %dst) #0
-  ; CHECK: _Z25atomic_fetch_sub_explicitPU3AS3VU7_Atomicii12memory_order12memory_scope(i32 addrspace(3)* {{.*}}, i32 1
+  %dec_jl = tail call spir_func i32 @_Z10atomic_decPU3AS3Vj(i32 addrspace(3)* %dst) #0
+  ; CHECK: _Z10atomic_decPU3AS3Vi(i32 addrspace(3)* %dst) #0
 
   ; atomic_max
   %max_il = tail call spir_func i32 @_Z10atomic_maxPVU3AS3ii(i32 addrspace(3)* %dst, i32 0) #0
```
aratajew: As a target, we also need to fix the translation of those builtins. This is a test for opencl 1.2 and we expect 2.0 version of builtin:
_Z25atomic_fetch_max_explicitPU3AS3VU7_Atomicii12memory_order12memory_scope
Additionally, this builtin is incorrect, because pointer address space has to be generic (4 not 3) to be consistent with clang.

I'll handle all the builtins below in the next PR, because it requires additional map specific for OpenCL 1.2.

aratajew:
I've close this pull request, because it can be implemented as a generic logic for all of the builtins. I'll create another PR for the new implementation

PR #262, opened 2019-07-19
Disabling mem2reg by default

AlexeySotkin:
Translator is not supposed to optimize code.
Also mem2reg complicates debugger's work.


AlexeySotkin:
@PawelJurek, please take a look.

PR #263, opened 2019-07-23
Add support for OpenCL 1.2 atomics builtins translation from SPIRV to LLVM

aratajew:
This change implements support for translating atomic builtin functions into OpenCL 1.2 builtins if source language version is 1.2 (or spirv-ocl-builtins-version=CL1.2). Without this change, atomics are translated to OpenCL 2.0 builtins with incorrect address space (not generic), not matching to what clang produces.

OpenCL2.0 path stays unchanged.


lib/SPIRV/SPIRVToOCL12.cpp
```diff
@@ -114,6 +138,91 @@ void SPIRVToOCL12::visitCallSPIRVControlBarrier(CallInst *CI) {
...
+      &Attrs);
+}
+
+Instruction *SPIRVToOCL12::visitCallSPIRVAtomicBuiltin(CallInst *CI, Op OC) {
+  CallInst *CIG = mutateCommonAtomicArguments(CI, OC);
+
+  Instruction *NewCI = nullptr;
+  switch (OC) {
+  case OpAtomicUMin:
+  case OpAtomicUMax:
```
AlexeySotkin: What about OpAtomic**S**Min and OpAtomic**S**Max ?

lib/SPIRV/SPIRVToOCL12.cpp
```diff
@@ -114,6 +138,91 @@ void SPIRVToOCL12::visitCallSPIRVControlBarrier(CallInst *CI) {
...
+        return OCL12SPIRVBuiltinMap::rmap(OpAtomicCompareExchange);
+      },
+      &Attrs);
+}
+
+Instruction *SPIRVToOCL12::visitCallSPIRVAtomicBuiltin(CallInst *CI, Op OC) {
+  CallInst *CIG = mutateCommonAtomicArguments(CI, OC);
+
+  Instruction *NewCI = nullptr;
+  switch (OC) {
```
AlexeySotkin: For OpAtomicLoad, OpAtomicStore, OpAtomicFlagTestAndSet, OpAtomicFlagClear see my https://github.com/KhronosGroup/SPIRV-LLVM-Translator/issues/266#issuecomment-514263083 and https://github.com/KhronosGroup/SPIRV-LLVM-Translator/issues/266#issuecomment-514579228

lib/SPIRV/SPIRVToOCL12.cpp
aratajew: Those opcodes are handled directly using map here: https://github.com/KhronosGroup/SPIRV-LLVM-Translator/pull/263/files#diff-ac841da0a76051bb8d965e112c5b3836R708

OpAtomicUMin and OpAtomicUMax are handle here, because those are translated to the same builtins in llvm and couldn't be added to the map.

lib/SPIRV/SPIRVToOCL12.cpp
aratajew: Support for those four instructions added. Thank you for your comments in #266. I've modified just implementation for OpAtomicFlagTestAndSet, as we don't need to issue icmp here. The only thing we need to do is truncate the i32 result from atomic_xchg to i1, as atomic_xchg returns the old value.

Let's assume old value is 1, then 1 is casted to boolaen true value which means the same as the result of OpAtomicFlagTestAndSet (The instruction’s result is true if the flag was in the set state).
Analogical for the false value.
It's safer than issuing icmp.

aratajew:
Good point, test  with a spirv file with all atomic instructions being translated to llvm ir with -spirv-ocl-builtins-version=CL1.2 and -spirv-ocl-builtins-version=CL2.0 added.

test/transcoding/atomics.spt
```diff
@@ -0,0 +1,95 @@
...
+; CHECK-LLVM-12: %.tmp = call spir_func i32 @_Z11atomic_xchgPU3AS1Vii(i32 addrspace(1)* %object, i32 1) [[attr]]
+; CHECK-LLVM-12: trunc i32 %.tmp to i1
+; CHECK-LLVM-12: call spir_func i32 @_Z11atomic_xchgPU3AS1Vii(i32 addrspace(1)* %object, i32 0) [[attr]]
+
+; RUN: llvm-spirv %s -to-binary -o %t2.spv
+; RUN: spirv-val %t2.spv
+; RUN: llvm-spirv -r %t2.spv -o %t2.bc -spirv-ocl-builtins-version="CL2.0"
+; RUN: llvm-dis < %t2.bc | FileCheck %s --check-prefix=CHECK-LLVM-20
+
+; CHECK-LLVM-20: call spir_func i32 @_Z25atomic_fetch_add_explicitPU3AS4VU7_Atomicii12memory_order12memory_scope(i32 addrspace(4)* %dst.as, i32 1, i32 2, i32 5) [[attr:#[0-9]+]]
```
AlexeySotkin: This strange that the last two arguments are 2, 5, while for other calls they are 5, 2. Please check.

test/transcoding/atomics.spt
```diff
@@ -0,0 +1,95 @@
...
+; CHECK-LLVM-12: call spir_func i32 @_Z14atomic_cmpxchgPU3AS1Viii(i32 addrspace(1)* %dst, i32 0, i32 1) [[attr]]
+; CHECK-LLVM-12: call spir_func i32 @_Z11atomic_xchgPU3AS1Vii(i32 addrspace(1)* %dst, i32 1) [[attr]]
+; CHECK-LLVM-12: call spir_func i32 @_Z10atomic_addPU3AS1Vii(i32 addrspace(1)* %object, i32 0) [[attr]]
+; CHECK-LLVM-12: call spir_func i32 @_Z11atomic_xchgPU3AS1Vii(i32 addrspace(1)* %object, i32 %desired) [[attr]]
+; CHECK-LLVM-12: %.tmp = call spir_func i32 @_Z11atomic_xchgPU3AS1Vii(i32 addrspace(1)* %object, i32 1) [[attr]]
+; CHECK-LLVM-12: trunc i32 %.tmp to i1
+; CHECK-LLVM-12: call spir_func i32 @_Z11atomic_xchgPU3AS1Vii(i32 addrspace(1)* %object, i32 0) [[attr]]
+
+; RUN: llvm-spirv %s -to-binary -o %t2.spv
+; RUN: spirv-val %t2.spv
```
AlexeySotkin: It is not necessary to create the same spv file twice and run spirv-val twice for the same file.

test/transcoding/atomics.spt
```diff
@@ -0,0 +1,95 @@
...
+
+
+5 Function 2 6 0 5 
+3 FunctionParameter 4 7 
+3 FunctionParameter 4 8 
+3 FunctionParameter 4 9 
+3 FunctionParameter 3 10 
+
+2 Label 11 
+6 AtomicIIncrement 3 12 7 13 14
```
AlexeySotkin: AtomicIDecrement is missing.

lib/SPIRV/SPIRVToOCL20.cpp
```diff
@@ -128,9 +142,63 @@ void SPIRVToOCL20::visitCallSPIRVControlBarrier(CallInst *CI) {
...
+        // atomic_fetch_sub_explicit OpenCL 2.0 builtins with "operand" argument
+        // = 1.
+        auto Name = OCLSPIRVBuiltinMap::rmap(
+            OC == OpAtomicIIncrement ? OpAtomicIAdd : OpAtomicISub);
+        auto Ptr = findFirstPtr(Args);
+        Type *ValueTy =
+            cast<PointerType>(Args[Ptr]->getType())->getElementType();
+        assert(ValueTy->isIntegerTy());
+        Args.push_back(llvm::ConstantInt::get(ValueTy, 1));
+        std::swap(Args[Ptr + 1], Args.back());
```
AlexeySotkin: I guess the test is wrong because of this swap

aratajew: Yes, that was it. Thanks :)

test/transcoding/atomics.spt
aratajew: Added.

test/transcoding/atomics.spt
aratajew: Sure, thanks.

test/transcoding/atomics.spt
aratajew: Fixed.

PR #264, opened 2019-07-23
Initialize mem fence passing to barrier in runtime if necessary

MrSidims:
If a flag passed to ControlBarrier or MemoryBarrier is being
initialized in runtime - don't lose it's initialization.

### DIFF
--- a/lib/SPIRV/OCLUtil.h
+++ b/lib/SPIRV/OCLUtil.h
@@ -66,6 +66,18 @@ enum OCLMemFenceKind {
   OCLMF_Image = 4,
 };
 
+// This enum declares extra constants for OpenCL mem_fence flag. It includes
+// combinations of local/global/image flags.
+enum OCLMemFenceExtendedKind {
+  OCLMFEx_Local = OCLMF_Local,
+  OCLMFEx_Global = OCLMF_Global,
+  OCLMFEx_Local_Global = OCLMF_Global | OCLMF_Local,
+  OCLMFEx_Image = OCLMF_Image,
+  OCLMFEx_Image_Local = OCLMF_Image | OCLMF_Local,
+  OCLMFEx_Image_Global = OCLMF_Image | OCLMF_Global,
+  OCLMFEx_Image_Local_Global = OCLMF_Image | OCLMF_Global | OCLMF_Local,
+};
+
 enum OCLScopeKind {
   OCLMS_work_item,
   OCLMS_work_group,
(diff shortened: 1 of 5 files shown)


lib/SPIRV/SPIRVToOCL12.cpp
```diff
@@ -85,12 +85,25 @@ void SPIRVToOCL12::visitCallSPIRVMemoryBarrier(CallInst *CI) {
...
-        auto GetArg = [=](unsigned I) {
-          return cast<ConstantInt>(Args[I])->getZExtValue();
-        };
-        auto Sema = mapSPIRVMemSemanticToOCL(GetArg(1));
-        Args.resize(1);
-        Args[0] = getInt32(M, Sema.first);
+        if (isa<ConstantInt>(Args[1])) {
+          auto GetArg = [=](unsigned I) {
+            return cast<ConstantInt>(Args[I])->getZExtValue();
+          };
```
AlexeySotkin: Why lambda? Can we just cast ?

lib/SPIRV/SPIRVToOCL12.cpp
```diff
@@ -85,12 +85,25 @@ void SPIRVToOCL12::visitCallSPIRVMemoryBarrier(CallInst *CI) {
...
+          auto Sema = mapSPIRVMemSemanticToOCL(GetArg(1));
+          Args.resize(1);
+          Args[0] = getInt32(M, Sema.first);
+        } else {
+          CallInst *TransCall = dyn_cast<CallInst>(Args[1]);
+          Function *F = TransCall ? TransCall->getCalledFunction() : nullptr;
+          if (F && F->getName().equals(kSPIRVName::TranslateOCLMemScope)) {
+            Args[0] = TransCall->getArgOperand(0);
+          } else {
+            Args[0] = getOrCreateSwitchFunc(kSPIRVName::TranslateSPIRVMemScope,
```
AlexeySotkin: This is not the same as `mapSPIRVMemSemanticToOCL`.

lib/SPIRV/SPIRVToOCL12.cpp
MrSidims: Yes. Done.

lib/SPIRV/SPIRVToOCL12.cpp
MrSidims: Fixed, thanks!

lib/SPIRV/OCLUtil.h
```diff
@@ -66,6 +66,17 @@ enum OCLMemFenceKind {
   OCLMF_Image = 4,
 };
 
+// This enum declares extra constants for OpenCL mem_fence flag. It includes
+// combinations of local/global/image flags, except global-image combination.
```
AlexeySotkin: I'm not sure we should exclude this combination.

MrSidims: Indeed, fixed

MrSidims:
Changed layout of OCLMemFenceExtendedKind

PR #265, opened 2019-07-23
travis: Switch to LLVM GitHub monorepo

svenvh:
The community is now advising that all workflows and CI systems
migrate to the GitHub monorepo, see [1] and [2].

[1] http://lists.llvm.org/pipermail/llvm-dev/2019-July/133840.html
[2] http://llvm.org/GitHubMigrationStatus.html

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -75,7 +75,7 @@ script:
       mkdir llvm-spirv
       mv * llvm-spirv
       git clone https://github.com/llvm/llvm-project --depth 1
-      mv llvm-spirv llvm-project/llvm/tools/llvm-spirv
+      mv llvm-spirv llvm-project/llvm-spirv
     fi
   - |
     if [ $TRAVIS_OS_NAME == "osx" ]; then
@@ -99,6 +99,7 @@ script:
           -G "Unix Makefiles"
       fi
     else
+      BUILDDIR=$(pwd)
       cmake ../llvm-project/llvm/ \
         -DCMAKE_BUILD_TYPE=${BUILD_TYPE} \
         -DBUILD_SHARED_LIBS=${SHARED_LIBS} \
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -74,8 +74,8 @@ script:
     if [ $BUILD_EXTERNAL == "0" ]; then
       mkdir llvm-spirv
       mv * llvm-spirv
-      git clone https://git.llvm.org/git/llvm.git/ --depth 1
-      mv llvm-spirv llvm/tools/llvm-spirv
+      git clone https://github.com/llvm/llvm-project --depth 1
+      mv llvm-spirv llvm-project/llvm/tools/llvm-spirv
```
AlexeySotkin: I think we better have it at `llvm-project/llvm-spirv`, like other llvm projects.

svenvh: I doubt that's going to work for `BUILD_EXTERNAL=0` builds, because cmake will not pick up the llvm-spirv project.  This is normally done by `add_llvm_implicit_projects()` in `llvm-project/llvm/tools/CMakeLists.txt`.

But let me push your suggestion as a separate commit to see if it works or not.

svenvh: Failed as expected:
```make: *** No rule to make target 'llvm-spirv'.  Stop.```
So I'll change it back to `llvm-project/llvm/tools/llvm-spirv`

Comment (AlexeySotkin) [reply .travis.yml]: According to this page https://llvm.org/docs/CMake.html I think we should specify the following cmake options:
```
LLVM_EXTERNAL_LLVM_SPIRV_SOURCE_DIR=../llvm-spirv
LLVM_EXTERNAL_PROJECTS="llvm-spirv"
```

bader: AFAIK, in addition to that you need also "enable" this external project.

-DLLVM_ENABLE_PROJECTS="llvm-spirv"

You can take a look how we build `llvm-spirv` with monorepo in SYCL project here:
https://github.com/intel/llvm/blob/sycl/sycl/doc/GetStartedWithSYCLCompiler.md#build-the-sycl-compiler-and-runtime

svenvh: Got it working now, thanks for your suggestions.

PR #267, opened 2019-07-23
Handle translation of llvm.invariant.* intrinsics

AlexeySotkin:


test/llvm.invariant.ll
```diff
@@ -0,0 +1,32 @@
+; Make sure the trasnlator doesn't crash if the input LLVM IR contains llvm.invariant.* intrinsics
```
svenvh: trasnlator -> translator

AlexeySotkin: Good catch, thanks :)

PR #269, opened 2019-07-24
Fix for OpNop instruction

matborzyszkowski:
Until now the kernel that contains OpNop instruction didn't compile.

### DIFF
--- a/lib/SPIRV/libSPIRV/SPIRVFunction.cpp
+++ b/lib/SPIRV/libSPIRV/SPIRVFunction.cpp
@@ -136,6 +136,10 @@ bool SPIRVFunction::decodeBB(SPIRVDecoder &Decoder) {
       break;
     }
 
+    if (Decoder.OpCode == OpNop) {
+      continue;
+    }
+
     if (Decoder.OpCode == OpLine) {
       Module->add(Decoder.getEntry());
       continue;
--- a/test/OpNop.spt
+++ b/test/OpNop.spt
@@ -1,0 +1,121 @@
+119734787 65536 393230 54 0 
+2 Capability Addresses 
+2 Capability Linkage 
+2 Capability Kernel 
+2 Capability Int64 
+2 Capability Int8 
+10 Extension "SPV_KHR_no_integer_wrap_decoration"
+5 ExtInstImport 1 "OpenCL.std"
+3 MemoryModel 2 2 
+5 EntryPoint 6 10 "Add" 
+12 String 51 "kernel_arg_type.Add.int*,int*,int*,int,"
+3 Source 3 102000 
+11 Name 5 "__spirv_BuiltInGlobalInvocationId"
+3 Name 11 "a"
+3 Name 12 "b"
+3 Name 13 "c"
+4 Name 14 "size"
+4 Name 15 "entry"
+4 Name 16 "if.then"
+4 Name 17 "if.end"
+4 Name 19 "a.addr"
+4 Name 20 "b.addr"
+4 Name 21 "c.addr"
+5 Name 23 "size.addr"
+3 Name 24 "n"
+4 Name 29 "call"
+4 Name 30 "conv"
+3 Name 34 "cmp"
+4 Name 37 "idxprom"
+5 Name 38 "arrayidx"
+5 Name 42 "idxprom2"
+5 Name 43 "arrayidx3"
+3 Name 45 "add"
+5 Name 48 "idxprom4"
+5 Name 49 "arrayidx5"
+4 Decorate 52 Alignment 4 
+2 DecorationGroup 52 
+4 Decorate 53 Alignment 8 
+2 DecorationGroup 53 
+4 Decorate 5 BuiltIn 28 
+3 Decorate 5 Constant 
+13 Decorate 5 LinkageAttributes "__spirv_BuiltInGlobalInvocationId" Import 
+3 Decorate 45 NoSignedWrap 
+4 GroupDecorate 52 23 24 
+5 GroupDecorate 53 19 20 21 
+4 TypeInt 2 64 0 
+4 TypeInt 7 32 0 
+4 TypeInt 25 8 0 
+4 TypeVector 3 2 3 
+4 TypePointer 4 1 3 
+2 TypeVoid 6 
+4 TypePointer 8 5 7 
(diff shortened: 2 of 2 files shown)


lib/SPIRV/libSPIRV/SPIRVFunction.cpp
```diff
@@ -136,6 +136,10 @@ bool SPIRVFunction::decodeBB(SPIRVDecoder &Decoder) {
       break;
     }
 
+    if (Decoder.OpCode == OpNop) {
+      continue;
```
AlexeySotkin: BTW, there is OpNop in SPIRV. Why we should not translate it?

StuartDBrady: There is no `nop` instruction in LLVM IR, so there's nothing to translate it _to_, although there may be some benefit to decoding it.

With the decoding change above, `llvm-spirv -to-text` strips out `OpNop`, and `llvm-spirv -to-binary` still fails on `assert(Found && "Invalid key")` in `SPIRVMap::rmap()`.

The changes required for this in SPIRVReader are not obvious to me, but it seems to me that there is at least no need to support `OpNop` outside of function bodies, as in this regard, it is just an ordinary instruction.

AlexeySotkin: Sorry, I thought that we are translating from llvm ir to spirv here. Please disregard my comment.
I think a better solution might be if we add a case for OpNop here

https://github.com/KhronosGroup/SPIRV-LLVM-Translator/blob/0905afb6c3388c629ff1050bbea31bb7afd1ad93/lib/SPIRV/SPIRVReader.cpp#L1506-L1511

StuartDBrady: > I think a better solution might be if we add a case for OpNop here

There is currently no definition for `Nop` in `lib/SPIRV/libSPIRV/SPIRVInstruction.h`, so presumably that needs to be added. @AlexeySotkin, should this go under "Other instructions", i.e. should it be templated over `SPIRVInstTemplateBase` along with `SpecConstOp`, `GenericPtrMemSemantics` and `GenericCastToPtrExplicit`? These seem not to require any special capabilities / extensions and don't have any special validation.

StuartDBrady:
FWIW, here's one of many articles on the net with (IMO) fairly good advice on writing commit messages: https://chris.beams.io/posts/git-commit/

StuartDBrady:
Note also that GitHub pull request titles and commit message titles are two entirely separate things. The GitHub UI doesn't make this entirely clear at times.

matborzyszkowski:
Okey. Thank you for your tips.

lib/SPIRV/libSPIRV/SPIRVFunction.cpp
AlexeySotkin: We have `OpNop` value here 
https://github.com/KhronosGroup/SPIRV-LLVM-Translator/blob/2f667b52e89201cd4e1ae84ccf9e03da52946b39/lib/SPIRV/libSPIRV/spirv.hpp#L690
I think this should be enough for the purposes of this patch, right ?

StuartDBrady: Currently the translator fails at the `Inst->getOpCode()` call in `SPIRVFunction::decodeBB()`, due to the `SPIRVDecoder::getEntry()` returning null on `OpCode == OpNop`. Once that is fixed, I expect `SPIRVEntry::create()` will need to create a `SPIRVNop` object.

matborzyszkowski:
> Could you explain the function in OpNop.spt, please? I would have expected a much shorter test that this, with a function consisting of a Return and nothing else.
> 
> Also, could you please squash fixes to prior commits in your PR?

The function in OpNop.spt is adding two vectors and store result to third vector.

StuartDBrady:
> The function in OpNop.spt is adding two vectors and store result to third vector.

A minimal test test focused only on OpNop would be more appropriate, here, as there would be less test code to maintain.

There may be other specific cases worth testing, but for each test, we should have a clear idea of what that test achieves that is not covered by the minimal test of a function with only an OpNop and an OpReturn.

If the aim was to test OpNop together with other instructions, then this is a worthy goal, but there are a few points to consider:
- Given that OpNop has no semantics, it is unlikely to interact badly with other instructions. There may be interactions between other instructions that are more likely to pose a problem. If so, we should add tests for those instead.
- The SPIRV-LLVM-Translator generally does a fairly direct translation from SPIR-V to LLVM, without any of the sorts of optimisation that would cause valid combinations of instructions to fail. That is to say, if translation of an operation works in a minimal test, it is very likely to work in any expanded test too.

We should aim to keep the tests as simple as possible, but no simpler. As far as in-depth testing of interactions between instructions is concerned, fuzz testing might be more appropriate. SPIR-V Tools has a fuzzer which could be extended to inject NOPs randomly into existing SPIR-V files, which would provide much better coverage than can be achieved through tests that are written and maintained by hand. The fuzzer is still being developed, and there are many things including NOPs which it does not currently inject.

If we say that OpNop is unlike other instructions, then it may be worth testing OpNop after one other instruction, to ensure that the other instruction is still translated correctly, but it would be better to pick the simplest test possible. For example, there is no need to perform a vector operation here: a scalar operation would be fine. Loads and stores are probably best avoided. I can imagine that we might want to check that OpNop does not cause interference if placed at the very start or as the penultimate instruction of a basic block, but we can test that with an OpNop in an otherwise empty function. If we want to test OpNop together with flow control, then it is important to know what the relevant test cases are.

StuartDBrady:
BTW, if you prefer, it may be better to concentrate on the implementation for now. I am planning to submit a bunch of extra tests soon, and OpNop is one of the instructions that is covered by them. Having said that, I'd be more than happy to discuss with you ways in which my own test could be improved, and will post a link to the pull request here once I have submitted it.

StuartDBrady:
As I promised, PR #285 adds a minimal test for OpNop. I am, of course, happy to discuss any additional testing that we might need.

StuartDBrady:
@matborzyszkowski, the test for OpNop has now been added. Wondering if you're still looking at this, and if so, whether we can offer any help?

aratajew:
@matborzyszkowski I think you can close this, as it has been already implemented: 5a56513

PR #244, opened 2019-07-03
Add infrastructure for handling SPIR-V versions and extensions

AlexeySachkov:
Prototype of #110 

Sorry for such a huge PR, but I've tried to split into a smaller patches to simplify the review.

There are still some FIXMEs/TODOs in the code: I've spotted a few places which can be improved, a few issues with intel-specific extensions, etc. I don't want to fix all of these in this PR since it is huge enough already, but I will do so (or delegate to someone) as follow-up PRs.

For anyone who is not such interested in the implementation, please review at least the last patch `
Add documentation for version and extensions handling`

I would like also to write some kind of `Extending LLVM-SPIRV-Translator` document which will describe how to add support for new extensions to the translator: where to look in the code, which guards/checks for new instructions are required to be inserted, which testing coverage is expected, etc. Not sure that I will prepare this document anytime soon, but I will do it later as a follow-up PR.

Any feedback/comments/suggesstions is welcome and appreciated

PR #270, opened 2019-07-24
 Handle loop metadatas attached to unstructured loops

MrSidims:
Implemented SPV_INTEL_unstructured_loop_controls extension,
that adds a new capability: UnstructuredLoopControlsINTEL
and a new instruction: LoopControlINTEL.

This instruction can be used in place of an OpLoopMerge on
unstructured loops.

Specification:
https://github.com/KhronosGroup/SPIRV-Registry/pull/38


lib/SPIRV/SPIRVWriter.cpp
```diff
@@ -886,6 +886,11 @@ SPIRVValue *LLVMToSPIRV::transValueWithoutDecoration(Value *V,
           BM->addLoopMergeInst(Br->getFalseLabel()->getId(), // Merge Block
                                BB->getId(),                  // Continue Target
                                LoopControl, Parameters, SuccessorTrue);
+        } else {
+          // For unstructured loop we add a special loop control instruction.
+          // Simple example of unstructured loop is an infinite loop, that has
+          // no terminate instruction.
+          BM->addLoopControlINTELInst(LoopControl, Parameters, SuccessorTrue);
```
Naghasan: This depends on #244 as this is an extension

not sure how should this handled, i.e. wait for #244 to land or follow up once it landed

MrSidims: I think I will commit it first and then rebase on #244, preparing a patch that Alexey can include in his PR.

PR #271, opened 2019-07-24
Fix OpCopyMemorySized conversion to memset

mlychkov:
Don't use optimization of OpCopyMemorySized operation to memset intrinsic
if element size of zero-initialized array differs from 1 byte.
Otherwise, OpCopyMemorySized conversion to memset instruction will fail.

### DIFF
--- a/test/sycl_array_zero_init.ll
+++ b/test/sycl_array_zero_init.ll
@@ -1,0 +1,56 @@
+; RUN: llvm-as %s -o %t.bc
+; RUN: llvm-spirv %t.bc -o %t.spv
+; RUN: llvm-spirv -r %t.spv -o %t.rev.bc
+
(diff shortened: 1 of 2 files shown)


mlychkov:
> LGTM, thanks!
> 
> Could you please add a test that used to crash before?

Added test file on which translation was failed.

test/sycl_array_zero_init.ll
```diff
@@ -0,0 +1,56 @@
+; RUN: llvm-as %s -o %t.bc
+; RUN: llvm-spirv %t.bc -o %t.spv
+; RUN: llvm-spirv -r %t.spv -o %t.rev.bc
+
```
AlexeySotkin: Please add a check to make sure that after round-trip translation we still have the same llvm.memcpy intrinsic, not llvm.memset

PR #275, opened 2019-07-26
Changing deploy mechanism

AlexeySotkin:
Currently every commit on a release branch can overwrite published release
on github. New approach in this patch presumes that releases will be done
manually via web interface on github. This will trigger another build in
travis, which in turn will deploy build aritfacts on github.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -13,6 +13,11 @@ git:
 branches:
   only:
     - llvm_release_80
+# From https://docs.travis-ci.com/user/customizing-the-build/#safelisting-or-blocklisting-branches
+# "Note that safelisting also prevents tagged commits from being built.
+#  If you consistently tag your builds in the format v1.3 you can safelist them
+#  all with regular expressions, for example /^v\d+\.\d+(\.\d+)?(-\S*)?$/."
+    - /^v\d+\.\d+(\.\d+)?(-\S*)?$/
 
 addons:
   apt:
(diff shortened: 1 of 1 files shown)


AlexeySotkin:
I consider retagging a bad idea in general. Every tag is a release in this repo. I don't think that particular release binaries are going to change over time. We rather should create another release. The scheme for tags is the following: first 3 digits match with version of llvm the release is based on, the fourth digit specify version of the translator itself compatible with version of llvm specified in the first 3 digits.
So v8.0.1-1 should be bumped to v8.0.1-2 because only translator has been changed, llvm version stays the same.
> Also, whether you wish to rename 7.0.1-1→v7.0.1-1 and 7.1.0-1→v7.1.0-1?

I'd prefer to create another tags/releases for them.

StuartDBrady:
> I consider retagging a bad idea in general.

Agreed, but if we've already been retagging, then perhaps the damage is already done... The question is really whether the older tags would be of any use, and if you do want to recreate those tags, what would you call them? It might be better simply to delete the tags that have been unintentionally rolling entirely (assuming my understanding of this is correct), but I don't really know. I don't mean to push for a particular course of action, just asking about the available options for this.

AlexeySotkin:
I think I get your point. Tags for v8.0.1-1 and 7.1.0-1 was moved unintentionally :( We need to decide if we want to move them back, leave as is, delete, or anything else.
I don't have the answer right now. Need to think it through, probably discuss it at the Khronos call.

AlexeySotkin:
Specifically,
tag v8.0.0-1 moved from 83298e3c9b124486c16d0fde54c764a6c5a2b554 to 30dacfcd6350395abd7c6735c3cf3c62de4a204c
tag v8.0.1-1 moved from c141308f72065e12510b1b76083e2d574e04202b to 5abe20ab9f2d5985cbfba1d003054aae60ff18b1
My suggestion is to move the tags back, because the original hashes are closer to the ones taken by Linux distributions. For example Ubuntu 19.04 has translator from 2019-03-14 https://launchpad.net/ubuntu/+source/spirv-llvm-translator/8.0.0+git20190314-0ubuntu2/+build/16639927

Issue #211, opened 2019-06-04
Remove support for the internal SPIR-V text format


kpet:
The translator currently supports its own textual format for SPIR-V assembly. This format is incompatible with the official format handled by the SPIRV-Tools. It is furthermore poorly tested, probably doesn't support all the corner cases and is generally a liability to the translator code base. 

The translator currently produces or consumes this format via three different options:
- `--spirv-text` for a translation from IR to the SPIR-V text format
- `--to-binary` to assemble a SPIR-V text file to a SPIR-V binary
- `--to-text` to disassemble a SPIR-V binary to a SPIR-V text file

I'm tempted to suggest that none of this functionality belongs in the translator and that we should get rid of it entirely, using the relevant SPIR-V Tools for testing.

A possible staged approach could be to:
- [ ] Remove support for `--to-text`. There are 11 uses in tests.
- [ ] Remove support for `--to-binary`. There are 17 uses in tests.
- [ ] Require SPIR-V Tools as a hard dependency and remove the optionality of SPIR-V validation (correctness shouldn't be an option :)) in the testing.
- [ ] Remove support for `--spirv-text` completely (125 uses in tests) and use `spirv-dis` in testing. The various name map data structures could be kept for in-debugger use (I know @neildhickey has been thinking about this).

These changes would make it easier to reuse code from SPIR-V tools as suggested in #32.

jigsawecho:
I did it a couple of years ago in a fork. Is there anyone working on it? The current support to text-format spirv makes the spt test cases difficult to read and write. It would be great to lean against either KhronosGroup/SPIRV-Tools or KhronosGroup/glslang.

svenvh:
You can write `.spvasm` tests as of 9977928 ("Add SPIR-V assembly tests", 2019-08-01).

jigsawecho:
@svenvh Thanks. I see, in this way both text formats are supported.

bwlodarcz:
@svenvh @MrSidims It's an old topic but it's relevant againt. The current approach is: We should go into using spirv-tools as our default test format and SPT usage should be minimized. In addition to that binary and text formats should be split in parser implementation to improve performance and readability of our main hot path.
In the long run SPT should be removed completely and tests should only use spirv-tools. 
Unfortunately there are a few Intel extensions in preview stage which are supported by translator but aren't in spirv-tools.
The possible discussed approaches to extensions problem:
[...]

PR #285, opened 2019-08-01
Add SPIR-V assembly tests and fix translation of OpLogicalNot

StuartDBrady:
This change adds SPIR-V assembly tests for OpFNegate, OpSNegate, OpFRem, OpNot, OpLogicalAnd, OpLogicalOr, OpLogicalEqual, OpLogicalNotEqual, OpLogicalNot and OpNop, and fixes translation of the OpLogicalNot instruction to LLVM IR.

As the OpNop instruction cannot currently be decoded, the test for this is an XFAIL test.

This adds a dependency on SPIR-V Tools, which provides the `spirv-as` program needed to assemble the tests. These tests are added in the SPIR-V assembly format with the intent that other tests will be converted away from the internal SPIR-V text format in future, as per issue #211.

### DIFF
--- a/test/CMakeLists.txt
+++ b/test/CMakeLists.txt
@@ -10,11 +10,22 @@ set(LLVM_SPIRV_TEST_SOURCE_DIR ${CMAKE_CURRENT_SOURCE_DIR})
 # find spirv-val
 pkg_search_module(SPIRV_TOOLS SPIRV-Tools)
 if(SPIRV_TOOLS_FOUND)
+  find_program(SPIRV_TOOLS_SPIRV_AS NAMES spirv-as PATHS ${SPIRV_TOOLS_PREFIX}/bin)
+  if(SPIRV_TOOLS_SPIRV_AS)
+    set(SPIRV_TOOLS_SPIRV_AS_FOUND True)
+  endif()
+
   find_program(SPIRV_TOOLS_SPIRV_VAL NAMES spirv-val PATHS ${SPIRV_TOOLS_PREFIX}/bin)
   if(SPIRV_TOOLS_SPIRV_VAL)
-    get_filename_component(SPIRV_TOOLS_BINDIR "${SPIRV_TOOLS_SPIRV_VAL}" DIRECTORY)
     set(SPIRV_TOOLS_SPIRV_VAL_FOUND True)
   endif()
+
+  set(SPIRV_TOOLS_BINDIR "${SPIRV_TOOLS_PREFIX}/bin")
+endif()
+
+if(NOT SPIRV_TOOLS_SPIRV_AS)
+  message(WARNING "spirv-as not found! SPIR-V assembly tests will not be run.")
+  set(SPIRV_TOOLS_SPIRV_AS_FOUND False)
 endif()
 
 if(NOT SPIRV_TOOLS_SPIRV_VAL)
--- a/README.md
+++ b/README.md
@@ -64,7 +64,8 @@ make llvm-spirv -j`nproc`
 
 ## Test instructions
 
-All tests related to the translator are placed in the [test](test) directory. Optionally the tests can make use of spirv-val (part of SPIRV-Tools) in order to validate the generated SPIR-V against the official SPIR-V specification.
+All tests related to the translator are placed in the [test](test) directory. A number of the tests require spirv-as (part of SPIR-V Tools) to run, but the remainder of the tests can still be run without this. Optionally the tests can make use of spirv-val (part of SPIRV-Tools) in order to validate the generated SPIR-V against the official SPIR-V specification.
+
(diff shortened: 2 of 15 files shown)


StuartDBrady:
The previous build failed due to Travis infrastructure failures. There are no changes in the newly pushed branch since the previous push, besides the refreshing of commit dates to force a rebuild.

test/CMakeLists.txt
```diff
@@ -15,13 +15,19 @@ if(SPIRV_TOOLS_FOUND)
...
+  find_program(SPIRV_TOOLS_SPIRV_AS NAMES spirv-as PATHS ${SPIRV_TOOLS_PREFIX}/bin)
 endif()
 
 if(NOT SPIRV_TOOLS_SPIRV_VAL)
   message(WARNING "spirv-val not found! SPIR-V generated for test suite will not be validated.")
   set(SPIRV_TOOLS_SPIRV_VAL_FOUND False)
 endif()
 
+if(NOT SPIRV_TOOLS_SPIRV_AS)
+  message(FATAL_ERROR "spirv-as not found! Needed for SPIR-V assembly tests.")
```
AlexeySachkov: I think this has been discussed already [in this thread](https://github.com/KhronosGroup/SPIRV-LLVM-Translator/pull/180#discussion_r283801824) - having hard dependency on spirv-tools at least have to be explicitly communicated to downstream users.

I think that it is okay to have it once we really use SPIR-V text representations from the spirv-tools, but for now it looks quite strange/restrictive that this dependency is needed for nothing more than a few tests.

Tagging @AlexeySotkin and @svenvh for more opinions

StuartDBrady: I could alter the dependency so that a warning is produced instead when `spirv-as` cannot be found. `config.suffixes` in `test/lit.cfg.py` would only be updated to include `'.spvasm'` if `spirv-as` was found. `README.md` would need to be updated to mention this in the _Test instructions_ section alongside the existing explanation of `spirv-val`. Would this be okay?

AlexeySachkov: Not sure that having some tests that are intended to test core features, but disabled due to missed optional dependency is a good idea.

Probably if all our tests used spirv-tools we could claim that `spirv-as`/`spirv-dis` tools are required to perform testing and make hard dependency on the spirv-tools. But I think that we need to discuss this with the rest of the translator community.

@AnastasiaStulova, @AlexeySotkin, @svenvh, we have faced this discussion second time already (previously [here](https://github.com/KhronosGroup/SPIRV-LLVM-Translator/pull/180#discussion_r283801824)), would it worth adding this topic into agenda for tooling call?

svenvh: > @AnastasiaStulova, @AlexeySotkin, @svenvh, we have faced this discussion second time already (previously [here](https://github.com/KhronosGroup/SPIRV-LLVM-Translator/pull/180#discussion_r283801824)), would it worth adding this topic into agenda for tooling call?

I agree that we should discuss this in the tooling call before merging.

AlexeySotkin: Yes, a warning is definitely better than a fatal error.
But also I'd prefer to add something like `; REQUIRE: spirv-as` to the test and update the  test config to enable them if spirv-as is found. This way the user will explicitly see which tests are skipped as UNSUPPORTED.

Looking forward, I think text representation of SPIR-V should be consistent across all tests for the translator. It is worth discussing if we want to take direction to switching SPIR-V text representation to the spirv-tools format for all tests.

StuartDBrady: Okay. I have pushed a new version that adds `spirv-as` as an optional dependency, and adds `; REQUIRES: spirv-as` to each of the SPIR-V assembly tests. I have confirmed that they are listed as UNSUPPORTED when running without `spirv-as`.

AlexeySachkov: @StuartDBrady, I, personally, okay to merge this since these tests will be launched by Travis CI and downstream users are responsible to configure environment correctly if they want to perform full testing of the translator. Of course, it still looks a little bit strange - tests for core feature, but they are effectively optional.

It would be really nice to see a corresponding issue to continue this discussion if this patch gets merged

StuartDBrady: @AlexeySachkov, we do have an issue for switching to the SPIR-V assembly format, which is issue #211. As more translator tests get added in SPIR-V assembly format (and converted from the internal SPIR-V text format), I imagine we might want to make running of the test suite conditional upon having `spirv-as` and `spirv-val`. This would be done without adding a dependency for users building the translator without wanting to run the tests.

If this revision is acceptable, I suggest we continue discussion on issue #211.

StuartDBrady:
@AlexeySotkin, @AlexeySachkov, wondering what your thoughts are on the latest version of this PR?

PR #286, opened 2019-08-01
Set convergent attribute for barriers

svenvh:
The `noduplicate` attribute is unnecessarily restrictive.  Align with
clang by setting the `convergent` attribute instead of `noduplicate`.

Followup from https://github.com/KhronosGroup/SPIRV-LLVM-Translator/pull/109


AlexeySachkov:
@turinevgeny: FYI

turinevgeny:
Thanks, appreciate it!

PR #291, opened 2019-08-09
Fix invalid .spv->.bc translation of Intel FPGA merge attribute

vmaksimo:
This patch introduces fix for the problem related to translation of SPIR-V decoration with two (or more) Literal Strings to LLVM IR. The processing of getting Literal Strings from the related decoration was limited to the first Literal String (first '\0' symbol).
In particular, the merge attribute has two Literal String according to the specification.

### DIFF
--- a/lib/SPIRV/libSPIRV/SPIRVEntry.cpp
+++ b/lib/SPIRV/libSPIRV/SPIRVEntry.cpp
@@ -335,57 +335,23 @@ bool SPIRVEntry::hasMemberDecorate(Decoration Kind, size_t Index,
   return true;
 }
 
-std::string SPIRVEntry::getDecorationStringLiteral(Decoration Kind) const {
-  std::vector<SPIRVWord> Literals;
-  auto Loc = Decorates.find(Kind);
-  if (Loc == Decorates.end())
-    return std::string();
-
-  for (SPIRVWord I = 0; I < Loc->second->getLiteralCount(); ++I)
-    Literals.push_back(Loc->second->getLiteral(I));
-
-  return getString(Literals);
-}
-
 std::vector<std::string>
-SPIRVEntry::getDecorationStringLiterals(Decoration Kind) const {
-  std::vector<SPIRVWord> Literals;
+SPIRVEntry::getDecorationStringLiteral(Decoration Kind) const {
   auto Loc = Decorates.find(Kind);
   if (Loc == Decorates.end())
     return {};
 
-  for (SPIRVWord I = 0; I < Loc->second->getLiteralCount(); ++I)
-    Literals.push_back(Loc->second->getLiteral(I));
-
-  return getStrings(Literals);
+  return getVecString(Loc->second->getVecLiteral());
 }
 
-std::string
+std::vector<std::string>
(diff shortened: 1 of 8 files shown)


lib/SPIRV/libSPIRV/SPIRVEntry.cpp
```diff
@@ -344,7 +344,20 @@ std::string SPIRVEntry::getDecorationStringLiteral(Decoration Kind) const {
   for (SPIRVWord I = 0; I < Loc->second->getLiteralCount(); ++I)
     Literals.push_back(Loc->second->getLiteral(I));
 
-  return getString(Literals.cbegin(), Literals.cend());
+  return getString(Literals);
+}
+
+std::vector<std::string>
+SPIRVEntry::getDecorationStringLiterals(Decoration Kind) const {
```
AlexeySachkov: This function looks absolutely the same as the previous one. I've spent several minutes to find differences and spot these two additional `s` symbols across ~8 lines of code.

I've searched for usages of `getDecorationStringLiteral` and looks like it is used only 3 times. Would it be better to completely replace `getDecorationStringLiteral` by the new `getDecorationStringLiterals`?

UPD: looks like I've found the reason for duplication: new `getDecorationStringLiterals` has to use new `getStrings` method under the hood, while the original `getString` method (used by `getDecorationStringLiteral`) is heavily used in `LLVMToSPIRVDbgTran.cpp` and some other places. So, it will be a huge change. Not asking then to refactor this right now - will leave this comment for everyone who is also wondering why is this code looks pretty the same as the existing.

MrSidims:
@AlexeySotkin please take a look

lib/SPIRV/libSPIRV/SPIRVEntry.cpp
```diff
@@ -358,7 +371,21 @@ SPIRVEntry::getMemberDecorationStringLiteral(Decoration Kind,
...
+
+std::vector<std::string>
+SPIRVEntry::getMemberDecorationStringLiterals(Decoration Kind,
+                                              SPIRVWord MemberNumber) const {
+  std::vector<SPIRVWord> Literals;
+  auto Loc = MemberDecorates.find({MemberNumber, Kind});
+  if (Loc == MemberDecorates.end())
+    return {};
+
+  for (SPIRVWord I = 0; I < Loc->second->getLiteralCount(); ++I)
```
AlexeySotkin: I think we can avoid copying of a vector if we move this loop to `SPIRVDecorateGeneric` class.

lib/SPIRV/libSPIRV/SPIRVEntry.cpp
AlexeySotkin: > while the original getString method (used by getDecorationStringLiteral) is heavily used in LLVMToSPIRVDbgTran.cpp and some other places

In LLVMToSPIRVDbgTran.cpp we heavily use `SPIRVModuleImpl::getString()` method not a free function `getString()` from SPIRVUtil.h

AlexeySachkov: Then, it is probably better to update existing `getDecorationStringLiteral` function (or replace it with the new one, more generic) instead of creating an almost full copy.

PR #294, opened 2019-08-14
Fix atomic functions conversion to OpenCL 1.2

mlychkov:
Some SPIR-V atomic functions are converted to OpenCL 1.2 builtins with
integral return value or arguments. For SPIR-V atomic functions with
float arguments it leads to generation of builtins with unsupported
signature.
Correct conversion of atomic functions arguments types and return type
from SPIR-V to IR with OpenCL 1.2 builtins.


mlychkov:
> Please add a regression test.

Done,

PR #295, opened 2019-08-14
Implementation of SPV_INTEL_function_pointers extension

AlexeySotkin:
The extension specification is published here
https://github.com/intel/llvm/blob/sycl/sycl/doc/extensions/spirv/SPV_INTEL_function_pointers.asciidoc

Overview:
This extensions brings two "levels" of function pointers support added under
corresponding capabilities:

Two new instructions are added under FunctionPointersINTEL capability:
 - OpFunctionPointerINTEL to support "address of" operator for functions,
 - OpFunctionPointerCallINTEL to do indirect function calls.

ReferencedIndirectlyINTEL decoration is added under IndirectReferencesINTEL
capability. This decoration can be attached to functions which are not
referenced directly in the module. These function must not be optimized out
based on call graph/reachability analysis.

Signed-off-by: Alexey Sachkov <alexey.sachkov@intel.com>


AlexeySotkin:
@AlexeySachkov, please take a look.

lib/SPIRV/libSPIRV/SPIRVOpCodeEnum.h
```diff
@@ -429,7 +431,7 @@ _SPIRV_OP(SubgroupAvcSicGetIpeChromaModeINTEL, 5813)
 _SPIRV_OP(SubgroupAvcSicGetPackedSkcLumaCountThresholdINTEL, 5814)
 _SPIRV_OP(SubgroupAvcSicGetPackedSkcLumaSumThresholdINTEL, 5815)
 _SPIRV_OP(SubgroupAvcSicGetInterRawSadsINTEL, 5816)
+_SPIRV_OP(LoopControlINTEL, 5887)
```
AlexeySachkov: I think it is better to remove this unrelated change

lib/SPIRV/libSPIRV/spirv.hpp
```diff
@@ -1127,10 +1132,10 @@ enum Op {
   OpSubgroupAvcSicGetPackedSkcLumaCountThresholdINTEL = 5814,
   OpSubgroupAvcSicGetPackedSkcLumaSumThresholdINTEL = 5815,
   OpSubgroupAvcSicGetInterRawSadsINTEL = 5816,
+  OpLoopControlINTEL = 5887,
```
AlexeySachkov: Another one unrelated change

lib/SPIRV/SPIRVWriter.cpp
```diff
@@ -776,9 +791,19 @@ SPIRVValue *LLVMToSPIRV::transValueWithoutDecoration(Value *V,
       MemoryAccess[0] |= MemoryAccessNontemporalMask;
     if (MemoryAccess.front() == 0)
       MemoryAccess.clear();
+
+    SPIRVValue *BSV = nullptr;
+    if(Function *F = dyn_cast<Function>(ST->getValueOperand())) {
```
AlexeySachkov: I think it worth checking that `SPV_INTEL_function_pointers` extension is supported before inserting `FunctionPointerINTEL` instruction

lib/SPIRV/SPIRVWriter.cpp
```diff
@@ -1455,6 +1489,12 @@ SPIRVValue *LLVMToSPIRV::transIntrinsicInst(IntrinsicInst *II,
 }
 
 SPIRVValue *LLVMToSPIRV::transCallInst(CallInst *CI, SPIRVBasicBlock *BB) {
+  if (CI->isIndirectCall())
```
AlexeySachkov: I think we need to insert an 'if' here to check if `SPV_INTEL_function_pointers` was enabled

lib/SPIRV/libSPIRV/SPIRVOpCodeEnum.h
AlexeySotkin: I just fixed the order of instructions. Do you want to make it in a separate commit?

lib/SPIRV/SPIRVWriter.cpp
AlexeySotkin: Maybe we should check it in addFunctionPointerINTELInst() then ?

PR #296, opened 2019-08-15
Fix OpFMod, implement OpSMod and add SPIR-V assembly tests

StuartDBrady:
**OpFMod**(_a_, _b_) was implemented as **copysign**(**frem**(_a_, _b_), _b_), which is not a remainder as described in _§2.2.3 Computation_ of the SPIR-V specification.

For example, **OpFMod**(-1.0, 100.0) resulted in 1.0, but the correct result is 99.0.  Likewise, **OpFMod**(1.0, -100.0) should result in -99.0, not -1.0.

Fix this by adding the divisor (_b_) if the signs of operands differ and the result of **frem**(_a_, _b_) is non-zero.

As **frem**(_a_, _b_) returns a result of the same sign as the dividend (_a_), the sign can be compared by taking the result of **frem**(_a_, _b_), copying the sign of the divisor to the result, and performing a test for equality to determine whether copying the sign modifies the value.

If the divisor is infinity, then the result of the **OpFMod** operation will also be infinity.  If either operand is a NaN, then the result will also be a NaN.  If the result is infinity, a NaN, or zero, then it will have the correct sign (i.e. the sign of the dividend).

This also implements **OpSMod**(_a_, _b_) in terms of **srem**(_a_, _b_).  In this case, the sign is compared by comparing (_a_ ^ _b_) < 0, which is **true** if and only if the sign bits differ.


lib/SPIRV/SPIRVReader.cpp
```diff
@@ -1739,36 +1739,48 @@ Value *SPIRVToLLVM::transValueWithoutDecoration(SPIRVValue *BV, Function *F,
...
+    auto Cmp = Builder.CreateAnd(CmpSign, CmpFRem, "cmp.res");
+
+    auto Select = Builder.CreateSelect(Cmp, FAdd, CopySign);
+    return mapValue(BV, Select);
+  }
+
+  case OpSMod: {
+    // translate OpSMod(a, b) to:
+    //   r = srem(a, b)
+    //   needs_fixing = a < 0 != b < 0 && r != 0
```
AlexeySachkov: Please put some `(`, `)` brackets here, it is a little bit hard to parse this expression

StuartDBrady: Okay, although I'm unsure as to where you want the brackets. Is this okay? :

```
needs_fixing = ((a < 0) != (b < 0) && r != 0)
```

StuartDBrady:
> `copysign(frem(a, b))` -> `copysign(frem(a, b), b)`
> 
> `copying the of the divisor to the result` -> `copying the *sign* of the divisor to the result` (3rd paragraph)

Thanks, will fix.

lib/SPIRV/SPIRVReader.cpp
AlexeySachkov: I was mostly confused by `a < 0 != b < 0` and spend some time trying to understand this expression :)

StuartDBrady:
> @StuartDBrady, am I right that new implementation is basically the same as the old one, except `needs_fixing` case where the result is `frem(a, b) + b` instead of `copysign(frem(a, b), b)`?

Yes, although the `needs_fixing` case is the main case where **OpFMod**(_a_, _b_) !== **OpFRem**(_a_, _b_). The only cases where we actually rely on **copysign**() for adjusting the result are corner cases to do with zeroes, NaNs and infinities.

> Could you please provide some test-cases (I mean probably some certain numbers) which show that current implementation is invalid? This would simplify review.

Sure.

For cases where both operands are normals or subnormals of the same sign, **OpFMod** should provide the same result as **OpFRem**, which is what both the new and old implementations do. Test cases with a positive or zero dividend and a positive divisor:
* OpFMod(±0, 100) = +0
* OpFMod(1, 100) = 1
* OpFMod(99, 100) = 99
* OpFMod(100, 100) = +0
* OpFMod(101, 100) = 1
* OpFMod(199, 100) = 99
* OpFMod(200, 100) = +0
* OpFMod(201, 100) = 1

 Test cases with a negative or zero dividend and a negative divisor:
* OpFMod(±0, -100) = -0
* OpFMod(-1, -100) = -1
* OpFMod(-99, -100) = -99
* OpFMod(-100, -100) = -0
* OpFMod(-101, -100) = -1
* OpFMod(-199, -100) = -99
* OpFMod(-200, -100) = -0
* OpFMod(-201, -100) = -1

In cases where the sign of the operands differ, the _old implementation gave the wrong value_. It would evaluate **OpFMod**(-1, 100) as 1, when the correct answer is 99, and it would evaluate **OpFMod**(1, -100) as -1, when the correct answer is -99. The relevant section of the SPIR-V specification is _§2.2.3 Computation_. The new implementation adheres to the specification.

Test cases with a negative dividend a positive divisor:
* OpFMod(-1, 100) = 99
* OpFMod(-99, 100) = 1
* OpFMod(-100, 100) = +0
* OpFMod(-101, 100) = 99
* OpFMod(-199, 100) = 1
* OpFMod(-200, 100) = +0
* OpFMod(-201, 100) = 99

Test cases with a positive dividend a negative divisor:
* OpFMod(1, -100) = -99
* OpFMod(99, -100) = -1
* OpFMod(100, -100) = -0
* OpFMod(101, -100) = -99
* OpFMod(199, -100) = -1
* OpFMod(200, -100) = -0
* OpFMod(201, -100) = -99

As for normals and subnormals, both the old implementation and the new one give the correct answer when the operands are of the same sign and the divisor is infinity. For a positive dividend with positive infinity:
* OpFMod(±0, +∞) = +0
* OpFMod(1, +∞) = 1
* OpFMod(100, +∞) = 100

For a negative dividend with negative infinity:
* OpFMod(±0, -∞) = -0
* OpFMod(-1, -∞) = -1
* OpFMod(-100, -i∞) = -100

However, if the signs of the dividend and divisor differ, and the divisor is infinity, the old implementation gives incorrect results, i.e. **OpFMod**(-1, +∞) would give 1, and **OpFMod**(1, -∞) would give 1, whereas the only mathematically justifiable result for **OpFMod**(-1, +∞) is +∞, and the only mathematically justifiable result for **OpFMod**(1, -∞) is -∞.

In all of the remaining cases, **OpFRem**(_a_, _b_) results in a NaN or infinity, as would the old implementation of **OpFMod**. However, it is correct to return a NaN or infinity of the same sign as the divisor, so this part of the implementation of **OpFMod** has been left unchanged.

lib/SPIRV/SPIRVReader.cpp
StuartDBrady: Ah, yes, `(a ^ b) < 0` is equivalent, but I felt it wouldn't have been helpful to put that in the comment. I'd be happy to reword both of the translation comments if you have any suggestions on how I could improve upon them.

StuartDBrady:
> > `copysign(frem(a, b))` -> `copysign(frem(a, b), b)`
> > `copying the of the divisor to the result` -> `copying the *sign* of the divisor to the result` (3rd paragraph)
> 
> Thanks, will fix.

Actually, this was on the description, and was not in the commit message. Have fixed the description.

StuartDBrady:
@AlexeySachkov, I have pushed version with the changes you have asked for, and also updated the `needs_fixing` comment for `OpFMod` so it nows says `!isnan(r) && !isnan(c)` instead of `isordered(r, b)`, which I think better reflects what the code actually does and should be more readily understood.

StuartDBrady:
@AlexSachkov, I have just pushed another version that simplifies the logic. Previously, the code tested _r_ != **copysign**(_r_, _b_) && _r_ != 0, but if _r_ is 0, then **copysign**(0, _b_) is also 0, and so the check for _r_ != 0 was unnecessary.

StuartDBrady:
> LGTM.
> 
> @StuartDBrady, I think it would be good to insert one or two particular examples into commit message, so anyone exploring git history can get more information.
> 
> BTW, looks like this patch requires rebasing

Thanks. Have rebased, and added an example for **OpFMod**(-1.0, 100.0) and one for **OpFMod**(1.0, -100.0). For brevity, I've not given separate examples for **OpSMod** as I think it should be clear enough that the operation is analogous.

I have also made a minor change to the code comment: instead `!isnan(r) && !isnan(c) && r != c`, I now state `islessgreater(r, c)`, for brevity, as the semantics are the same. AFAICT from this, `OpLessOrGreater` is therefore equivalent to `OpFOrdNotEqual`, although I will seek to confirm this with the SPIR WG.

PR #299, opened 2019-08-19
Fix IntelFPGA memory attrs on struct that has c'tor

MrSidims:
Strip all bitcast and addrspace casts from the pointer argument:
llvm annotation intrinsic only takes i8*, so the original pointer
probably had to loose its addrspace and its original type.

### DIFF
--- a/lib/SPIRV/SPIRVWriter.cpp
+++ b/lib/SPIRV/SPIRVWriter.cpp
@@ -1198,6 +1198,11 @@ void addIntelFPGADecorations(
     return;
 
   for (const auto &I : Decorations) {
+    // Such decoration already exists on a type, skip it
+    if (E->hasDecorate(I.first, /*Index=*/0, /*Result=*/nullptr)) {
+      continue;
+    }
+
     switch (I.first) {
     case DecorationUserSemantic:
       E->addDecorate(new SPIRVDecorateUserSemanticAttr(E, I.second));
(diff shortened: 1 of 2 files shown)


MrSidims:
@AlexeySachkov @bader can you take a look?

lib/SPIRV/SPIRVWriter.cpp
```diff
@@ -1240,6 +1240,12 @@ void addIntelFPGADecorationsForStructMember(
     return;
 
   for (const auto &I : Decorations) {
+    // Such decoration already exists on a type, skip it
```
AlexeySachkov: What about the same check in `addIntelFPGADecorations` ?

lib/SPIRV/SPIRVWriter.cpp
```diff
@@ -1426,20 +1432,21 @@ SPIRVValue *LLVMToSPIRV::transIntrinsicInst(IntrinsicInst *II,
...
     StringRef AnnotationString =
         dyn_cast<ConstantDataArray>(C->getOperand(0))->getAsCString();
 
-    if (GI) {
+    // Strip all bitcast and addrspace casts from the pointer argument:
+    //   llvm annotation intrinsic only takes i8*, so the original pointer
+    //   probably had to loose its addrspace and its original type.
+    Value *AnnotSubj = II->getArgOperand(0);
+    while (isa<BitCastInst>(AnnotSubj) || isa<AddrSpaceCastInst>(AnnotSubj)) {
+      AnnotSubj = cast<CastInst>(AnnotSubj)->getOperand(0);
```
AlexeySachkov: What if we have cast of cast or cast of addrspacecast, i.e. more than one cast?

MrSidims: It's not possible to happen. We will always have bitcast or addrspacecast.

lib/SPIRV/SPIRVWriter.cpp
MrSidims: Good idea! Done.

lib/SPIRV/SPIRVWriter.cpp
AlexeySachkov: Oh, you actually have a `while` loop here which will unwrap nested casts if any. Please, disregard my comment then

PR #301, opened 2019-08-20
Fix enqueue_kernel builtins translation - SPIRV to LLVM

aratajew:
__enqueue_kernel_basic and __enqueue_kernel_varargs bultins were not produced by SPIRVReader, instead of them __enqueue_kernel_basic_events and __enqueue_kernel_events_varargs with null event arguments were produced. This approach wasn't clang-consistent.

### DIFF
--- a/lib/SPIRV/SPIRVReader.cpp
+++ b/lib/SPIRV/SPIRVReader.cpp
@@ -2085,26 +2085,45 @@ Instruction *SPIRVToLLVM::transEnqueueKernelBI(SPIRVInstruction *BI,
   // Find or create enqueue kernel BI declaration
   auto Ops = BI->getOperands();
   bool HasVaargs = Ops.size() > 10;
+  bool HasEvents = true;
+  SPIRVValue *NumEvents = Ops[3];
+  if (NumEvents->getOpCode() == OpConstant) {
+    SPIRVConstant *NE = static_cast<SPIRVConstant *>(NumEvents);
+    HasEvents = NE->getZExtIntValue() != 0;
+  }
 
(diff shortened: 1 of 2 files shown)


lib/SPIRV/SPIRVReader.cpp
```diff
@@ -2085,26 +2085,45 @@ Instruction *SPIRVToLLVM::transEnqueueKernelBI(SPIRVInstruction *BI,
...
-                                : "__enqueue_kernel_basic_events";
   Function *F = M->getFunction(FName);
   if (!F) {
-    Type *EventTy = PointerType::get(
-        getOrCreateOpaquePtrType(M, SPIR_TYPE_NAME_CLK_EVENT_T,
-                                 getOCLOpaqueTypeAddrSpace(OpTypeDeviceEvent)),
-        SPIRAS_Generic);
-
-    SmallVector<Type *, 8> Tys = {
+    SmallVector<Type *, 3> Tys = {
```
svenvh: I don't think you should change the size to 3, as this SmallVector will always contain more than 3 elements (which will cause a heap allocation).

lib/SPIRV/SPIRVReader.cpp
```diff
@@ -2119,17 +2138,21 @@ Instruction *SPIRVToLLVM::transEnqueueKernelBI(SPIRVInstruction *BI,
   }
 
   // Create call to enqueue kernel BI
-  SmallVector<Value *, 8> Args = {
+  SmallVector<Value *, 3> Args = {
```
svenvh: Same here, I think this can be left at 8.

aratajew: Good point. Fixed.

PR #307, opened 2019-08-21
Fix translation of OpBitReverse and extend testing

StuartDBrady:


test/OpBitReverse_v2i16.ll
```diff
@@ -0,0 +1,41 @@
...
+; CHECK-SPIRV: 4 TypeVector [[short2:[0-9]+]] [[short]] 2
+; CHECK-SPIRV: 4 BitReverse [[short2]]
+
+; CHECK-LLVM: call <2 x i16> @llvm.bitreverse.v2i16
+
+target datalayout = "e-p:32:32-i64:64-v16:16-v24:32-v32:32-v48:64-v96:128-v192:256-v256:256-v512:512-v1024:1024"
+target triple = "spir-unknown-unknown"
+
+; Function Attrs: convergent nounwind writeonly
+define spir_kernel void @testBitRev(<2 x i16> %a, <2 x i16> %b, <2 x i16> %c, <2 x i16> addrspace(1)* nocapture %res) local_unnamed_addr #0 !kernel_arg_addr_space !3 !kernel_arg_access_qual !4 !kernel_arg_type !5 !kernel_arg_base_type !5 !kernel_arg_type_qual !6 {
```
AlexeySachkov: Would it be better to merge this test into a existing one? They are not so complex

StuartDBrady: I'm never quite sure whether it's better to merge tests:

- Combining tests means we have fewer files, which makes for less clutter and means the tests run a bit more quickly.
- Keeping them apart means they can be diffed more easily, which eases maintenance, especially for larger tests (e.g. `tests/transcoding/relationals_half.ll` and `tests/transcoding/relationals_half.ll`).
- Keeping them apart provides greater distinction between the tests, i.e. it's a much more obvious bug if a v2i16 test has lines that test the i32 version of an op, than if they're all mixed together.
- Keeping them apart provides separate test results, which provides useful information immediately if anything starts failing.
- Keeping them apart simplifies the tests, in that it avoids the need to use `CHECK-DAG:` where types are defined which could appear in the assembly output in any arbitrary order.

I suppose having lots of small files is generally pretty cheap these days, so that's the direction I think I'd _prefer_ to go in, but execution time of the tests is a valid concern.

If there are large set-up costs with starting clang, LLVM tools, llvm-spirv, and/or the SPIR-V tools, then we should look into what can be done to reduce those costs. E.g. currently, a lot of our tests translate twice: once to produce `.spt` output for running FileCheck, and then again to produce a '.spv' file for running `spirv-val` and `llvm-spirv -r`.

AlexeySachkov: Thanks for such detailed response! This is definitely makes sense. 

> I suppose having lots of small files is generally pretty cheap these days, so that's the direction I think I'd prefer to go in, but execution time of the tests is still a valid concern.

I don't have strong opinion about this topic and I'm fine with both ways (separate and unified).

For these `.ll` tests I'm sure that they are fast enough so we can add a few more without any issues.

svenvh: Besides execution time and clutter, another concern is maintenance of the `RUN` lines, and with many test files, this can become a bit tedious (as there can be subtle differences between `RUN` lines across different files).  So I prefer combining tests into the same file when they are related.  Also, the LLVM Testing Guide [1] mentions:

> Put related tests into a single file rather than having a separate file per
test. Check if there are files already covering your feature and consider
adding your code there instead of creating a new file.

(though no motivation is given)

I think in this case the tests are related enough so personally I would have merged them into a single file.

[1] http://llvm.org/docs/TestingGuide.html

StuartDBrady: Thanks, Sven.

Regarding the RUN lines, I appreciate that occasionally a test does need to do something special, but should we factor the majority of these out into scripts? I.e. one for `.cl` transcoding tests, one for `.ll` transcoding tests, one for `.spt' tests and one for `.spvasm` tests. There are quite a few needless differences (including cases where we don't run `spirv-val`) which it would be good to avoid.

The tests that currently have the most common sets of `RUN:` lines are:
* 33 `.ll` transcoding tests, three which do not reside in the `transcoding` directory, as presumably they should, i.e. `CreatePipeFromPipeStorage.ll`, `llvm.memcpy.align.ll` and `PipeStorage.ll`
* 17 `.ll` tests that check `-spirv-text` output and run `spirv-val`.
* All 14 `.spvasm` tests (with this change included).
* 9 `.spt` tests that run `spirv-val` and 6 tests that don't (including `OpLoopMergeDontUnrollHint1.spt` and `OpLoopMergePartialUnroll.spt`) and another 6 tests that do run `spirv-val` but use `CHECK:` instead of `CHECK-LLVM:`. This also includes 4 `.spt` tests in the `transcoding` directory that do not appear to be transcoding tests at all.

I note that there are in the order of 29000 tests for LLVM itself and around 11000 for Clang. It seems reasonable to imagine that the number of tests for the LLVM/SPIR-V translator will increase quite a bit over time, but perhaps it's reasonable to split them up a little less than for LLVM?

svenvh: > Regarding the RUN lines, I appreciate that occasionally a test does need to do something special, but should we factor the majority of these out into scripts?

The amount of scripting payload in each test is currently quite high indeed and I have been thinking about factoring it out into scripts in the past.  At the same time, I am wary of putting too much magic into the test suite and hiding the commands that are being run from lit itself.

> I note that there are in the order of 29000 tests for LLVM itself and around 11000 for Clang. It seems reasonable to imagine that the number of tests for the LLVM/SPIR-V translator will increase quite a bit over time, but perhaps it's reasonable to split them up a little less than for LLVM?

Not sure, how would you define "a little less"?  Unless there is a strong reason to deviate from the LLVM Testing Guide, I think I'd rather follow their advice of combining where relevant.

PR #313, opened 2019-08-28
Retain llvm.loop metadata for LoopControlINTEL instruction

MrSidims:
In case when debug information presents in the module, the
metadata was missing with translation back of LoopControlINTEL
instruction due to incorrect instruction ID.

### DIFF
--- a/test/DebugInfo/DebugUnstructuredControlFlow.cl
+++ b/test/DebugInfo/DebugUnstructuredControlFlow.cl
@@ -1,0 +1,25 @@
+// RUN: %clang_cc1 -triple spir64-unknown-unknown -cl-std=CL2.0 -O0 -debug-info-kind=standalone -emit-llvm-bc %s -o %t.bc
+// RUN: llvm-spirv %t.bc --spirv-ext=+SPV_INTEL_unstructured_loop_controls -o %t.spv
+// RUN: llvm-spirv %t.spv --to-text -o %t.spt
+// RUN: FileCheck < %t.spt %s --check-prefix=CHECK-SPIRV
+// RUN: llvm-spirv -r %t.spv -o %t.bc
+// RUN: llvm-dis < %t.bc | FileCheck %s --check-prefix=CHECK-LLVM
+
+// Test that no debug info instruction is inserted between LoopControlINTEL and
+// Branch instructions. Otherwise, debug info interferes with SPIRVToLLVM
+// translation of structured flow control. Yet Line DebugInfo instruction
+// is still presenting between LoopControlINTEL and Branch instructions.
+
+kernel
+void sample() {
+  #pragma clang loop unroll(full)
+  for(;;);
+}
+
+// CHECK-SPIRV: 2 LoopControlINTEL 1
+// CHECK-SPIRV-NOT: ExtInst
+// CHECK-SPIRV: {{[0-9]+}} Line {{[0-9]+}} {{[0-9]+}} {{[0-9]+}}
+// CHECK-SPIRV: Branch
+// CHECK-LLVM: br label %{{.*}}, !dbg !{{[0-9]+}}, !llvm.loop ![[MD:[0-9]+]]
+// CHECK-LLVM: ![[MD]] = distinct !{![[MD]], ![[MD_unroll:[0-9]+]]}
+// CHECK-LLVM: ![[MD_unroll]] = !{!"llvm.loop.unroll.enable"}
--- a/lib/SPIRV/libSPIRV/SPIRVBasicBlock.cpp
+++ b/lib/SPIRV/libSPIRV/SPIRVBasicBlock.cpp
@@ -70,7 +70,8 @@ SPIRVBasicBlock::addInstruction(SPIRVInstruction *I,
     // If insertion of a new instruction before the one passed to the function
     // is illegal, insertion before the returned instruction is guaranteed
     // to retain correct instruction order in a block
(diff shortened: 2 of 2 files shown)


AlexeySachkov:
@AGindinson, please take a look

test/OpLoopControlINTELGDB.spt
```diff
@@ -0,0 +1,100 @@
...
+
+2 Label 23
+6 ExtInst 3 57 2 DebugScope 48
+2 LoopControlINTEL 1
+4 Line 28 19 5
+2 Branch 23
+
+1 FunctionEnd
+
+; RUN: llvm-spirv %s -to-binary -o %t.spv
```
AGindinson: Why is there no check for LLVM --> SPIR-V translation?

test/OpLoopControlINTELGDB.spt
```diff
@@ -0,0 +1,100 @@
...
+6 ExtInst 3 56 2 DebugScope 48
+4 Line 28 0 0
+8 ExtInst 3 26 2 DebugDeclare 51 25 52
+4 Line 28 19 5
+2 Branch 23
+
+2 Label 23
+6 ExtInst 3 57 2 DebugScope 48
+2 LoopControlINTEL 1
+4 Line 28 19 5
```
AGindinson: As far as I recall, this is the exact broken instruction order we're trying to avoid. The code inserts the Line instruction before the LoopControl, but the test suggests otherwise.

test/OpLoopControlINTELGDB.spt
MrSidims: Because the patch fixes a bug in LLVM IR consumption. I'm sure, that LLVM --> SPIR-V translation with debug information is good covered in DebugInfo/DebugControlFlow.cl test.

test/OpLoopControlINTELGDB.spt
AGindinson: What I should've also recalled is that OpLine doesn't interfere with the translation flow.

test/OpLoopControlINTELGDB.spt
MrSidims: Reworked, thanks!

test/DebugInfo/DebugUnstructuredControlFlow.cl
```diff
@@ -0,0 +1,29 @@
...
+
+kernel
+void sample() {
+  #pragma clang loop unroll(full)
+  for(;;);
+}
+
+// CHECK-SPIRV: 2 LoopControlINTEL 1
+// CHECK-SPIRV-NOT: ExtInst
+// CHECK-SPIRV: {{[0-9]+}} Line {{[0-9]+}} {{[0-9]+}} {{[0-9]+}}
```
AGindinson: To clarify: is it guaranteed that OpLine will land here? If there's no certainty as to its placement, we could simply omit the check.

MrSidims: I think it should present before Branch instruction.

AGindinson:
LGTM

test/DebugInfo/DebugUnstructuredControlFlow.cl
```diff
@@ -0,0 +1,29 @@
+// RUN: %clang_cc1 -triple spir64-unknown-unknown -cl-std=CL2.0 -O0 -debug-info-kind=standalone -emit-llvm %s -o %t.ll
```
AlexeySachkov: You can directly emit byte-code via `-emit-llvm-bc` and avoid calling `llvm-as`

test/DebugInfo/DebugUnstructuredControlFlow.cl
```diff
@@ -0,0 +1,29 @@
...
+// RUN: llvm-as %t.ll -o %t.bc
+// RUN: llvm-spirv %t.bc --spirv-ext=+SPV_INTEL_unstructured_loop_controls -spirv-text -o %t.spt
```
AlexeySachkov: Also you can use `-to-text` option to generate `.spt` file from `.spv` file - in this case you don't need to pass `--spriv-ext` option again

test/DebugInfo/DebugUnstructuredControlFlow.cl
```diff
@@ -0,0 +1,29 @@
...
+// RUN: llvm-spirv %t.bc --spirv-ext=+SPV_INTEL_unstructured_loop_controls -o %t.spv
+// RUN: llvm-spirv -r %t.spv -o %t.bc
+// RUN: llvm-dis < %t.bc | FileCheck %s --check-prefix=CHECK-LLVM
+
+// Test that no debug info instruction is inserted between LoopControlINTEL and
+// Branch instructions. Otherwise, debug info interferes with SPIRVToLLVM
+// translation of structured flow control.
+//
+// Currently, Line DebugInfo instructions are still present between
+// LoopControlINTEL and Branch instructions. This does not affect SPIRVToLLVM
```
AlexeySachkov: Are you going to fix it later? What do you think about leaving FIXME/TODO here?

test/DebugInfo/DebugUnstructuredControlFlow.cl
MrSidims: Thanks, done!

test/DebugInfo/DebugUnstructuredControlFlow.cl
MrSidims: Well, yeah, the line was confusing. Nothing to fix here.

PR #314, opened 2019-08-28
[NFC] Fix misleading comment in the test for debug info/control flow

AGindinson:

### DIFF
--- a/test/DebugInfo/DebugControlFlow.cl
+++ b/test/DebugInfo/DebugControlFlow.cl
@@ -10,11 +10,6 @@
 // between LoopMerge and Branch/BranchConditional instructions.
 // Otherwise, debug info interferes with SPIRVToLLVM translation
 // of structured flow control
-//
-// Currently, Line DebugInfo instructions are still present
-// between LoopMerge and Branch/BranchConditional instructions.
-// This does not affect SPIRVToLLVM translation, however
-// should be fixed separately
 
 kernel
 void sample() {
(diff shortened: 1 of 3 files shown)


AGindinson:
Tagging @AlexeySachkov @MrSidims

MrSidims:
LGTM

test/DebugInfo/DebugControlFlow.cl
```diff
@@ -11,10 +11,9 @@
 // Otherwise, debug info interferes with SPIRVToLLVM translation
 // of structured flow control
 //
-// Currently, Line DebugInfo instructions are still present
+// Line DebugInfo instructions are still present
 // between LoopMerge and Branch/BranchConditional instructions.
```
AlexeySachkov: Is it valid according to the spec?

I'm just curious if everything is valid, why is it so important to mention this fact? Otherwise, we should probably put some TODO here to fix what is incorrect

AGindinson: The spec clearly states that `OpLoopMerge` must be second to last in its basic block, `OpBranch` being its immediate/only successor. Moreover, taking another glance at the spec:

"- if a branch merge instruction is used, the last `OpLine` in the block must be before its merge instruction"

I think I should remove the comment here and in the test for LoopContol brought by #313. I should add the TODO in the code that generates `OpLine`s.

AGindinson: It remains to be determined why does incorrect `OpLine` placement NOT interfere with backwards translation.

AlexeySachkov:
@MrSidims, please take a look at the updated version

MrSidims:
Yet LGTM

PR #315, opened 2019-08-28
[SYCL][FPGA] Skip *.annotation call debug info

mlychkov:
Currently there is no analogue of *.annotation intrinsic in SPIR-V.
During IR translation this call is skipped or stored as a decoration.
Position of this intrinsic in a basic block of IR after reverse
translation may change or even lost. However, debug scope of the call
may be translated into SPIR-V in inappropriate place and result in
corruption of debug metadata for other instructions in that basic
block. Even if debug info of the call doesn't corrupt debug info of
other instructions in that basic block it isn't restored for the call
during reverse translation.
Skip debug scope info of *.annotation intrinsic call as it is useless
for now.

### DIFF
--- a/test/annotation_dbg_info_drop.ll
+++ b/test/annotation_dbg_info_drop.ll
@@ -1,0 +1,321 @@
+; RUN: llvm-as %s -o %t.bc
+; RUN: llvm-spirv %t.bc -o %t.spv
+; RUN: llvm-spirv %t.spv --to-text -o %t.spt
+; RUN: FileCheck < %t.spt %s --check-prefixes=CHECK,CHECK-SPV
+
+; RUN: llvm-spirv %t.bc --spirv-ext=+SPV_INTEL_fpga_reg -o %t.fpga_reg.spv
+; RUN: llvm-spirv %t.fpga_reg.spv --to-text -o %t.fpga_reg.spt
+; RUN: FileCheck < %t.fpga_reg.spt %s --check-prefixes=CHECK,CHECK-SPV-FPGA_REG
+
+; -- Check that reverse translation is not failed.
+; RUN: llvm-spirv -r %t.spv -o %t.rev.bc
+; RUN: llvm-spirv -r %t.fpga_reg.spv -o %t.rev.fpga_reg.bc
+
+; ModuleID = 'annotation_dbg_info_drop.cpp'
+source_filename = "annotation_dbg_info_drop.cpp"
+target datalayout = "e-i64:64-v16:16-v24:32-v32:32-v48:64-v96:128-v192:256-v256:256-v512:512-v1024:1024"
+target triple = "spir64-unknown-linux-sycldevice"
+
+%"class._ZTSZZ4mainENK3$_0clERN2cl4sycl7handlerEEUlvE_.anon" = type { i64 }
+%"struct._ZTSZZZ4mainENK3$_0clERN2cl4sycl7handlerEENKUlvE_clEvE3s_d.s_d" = type { [8 x i32] }
+%"struct._ZTSZZZ4mainENK3$_0clERN2cl4sycl7handlerEENKUlvE_clEvE3s_e.s_e" = type { [4 x i32] }
+
+$"_ZTSZZ4mainENK3$_0clERN2cl4sycl7handlerEE11test_kernel" = comdat any
+
+@.str = private unnamed_addr constant [25 x i8] c"{memory:DEFAULT}{pump:1}\00", section "llvm.metadata"
+@.str.1 = private unnamed_addr constant [29 x i8] c"annotation_dbg_info_drop.cpp\00", section "llvm.metadata"
+@.str.2 = private unnamed_addr constant [25 x i8] c"__builtin_intel_fpga_reg\00", section "llvm.metadata"
+@.str.3 = private unnamed_addr constant [29 x i8] c"{memory:DEFAULT}{numbanks:2}\00", section "llvm.metadata"
+
+; CHECK: {{[0-9]+}} Name [[ANNO_PUMP_ID:[0-9]+]] ".str"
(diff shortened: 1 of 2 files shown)


mlychkov:
@AlexeySachkov please review a fix

mlychkov:
@AGindinson please review a fix

mlychkov:
@svenvh could you please review a fix?

lib/SPIRV/SPIRVWriter.cpp
```diff
@@ -1392,6 +1392,11 @@ SPIRVValue *LLVMToSPIRV::transIntrinsicInst(IntrinsicInst *II,
         return transValue(II->getOperand(0), BB);
     }
 
+    // Skip debug info for this call because this call is dropped in SPIR-V.
+    // TODO: remove skipping of debug info when *.annotation call will be
+    //       handled properly during IR <=> SPIR-V translation.
```
AlexeySachkov: Since there is no equivalent for this intrinsic in SPIR-V, not sure that there is a proper way to translate it.

```suggestion
    //       handled in a better way during IR <=> SPIR-V translation.
```

lib/SPIRV/SPIRVWriter.cpp
```diff
@@ -1423,6 +1428,12 @@ SPIRVValue *LLVMToSPIRV::transIntrinsicInst(IntrinsicInst *II,
     } else {
       addIntelFPGADecorations(SV, Decorations);
     }
+
+    // Skip debug info for this call because this call is dropped in SPIR-V.
+    // TODO: remove skipping of debug info when *.annotation call will be
+    //       handled properly during IR <=> SPIR-V translation.
```
```suggestion
    //       handled in a better way during IR <=> SPIR-V translation.
```

lib/SPIRV/SPIRVWriter.cpp
```diff
@@ -1470,6 +1481,12 @@ SPIRVValue *LLVMToSPIRV::transIntrinsicInst(IntrinsicInst *II,
           return transValue(BI, BB);
       }
     }
+
+    // Skip debug info for this call because this call is dropped in SPIR-V.
+    // TODO: remove skipping of debug info when *.annotation call will be
+    //       handled properly during IR <=> SPIR-V translation.
```
```suggestion
    //       handled in a better way during IR <=> SPIR-V translation.
```

lib/SPIRV/SPIRVWriter.cpp
```diff
@@ -1392,6 +1392,11 @@ SPIRVValue *LLVMToSPIRV::transIntrinsicInst(IntrinsicInst *II,
         return transValue(II->getOperand(0), BB);
     }
 
+    // Skip debug info for this call because this call is dropped in SPIR-V.
+    // TODO: remove skipping of debug info when *.annotation call will be
+    //       handled in a better way during IR <=> SPIR-V translation.
+    II->setDebugLoc(DebugLoc());
```
AlexeySotkin: I'm not sure this is the right place to insert this code. What if we return on line 1389 or 1392, the debug info will not be removed, right ?

mlychkov: Debug info for annotation intrinsic line is dropped only in case if the whole call is dropped from SPIR-V.
On lines this call is translated into call of intel intrinsic fpga_reg or replaced by assignment of first argument of annotation method. So, in these cases debug info can be linked to a particular line in SPIR-V representation.

AlexeySotkin: Ok

test/annotation_dbg_info_drop.ll
```diff
@@ -0,0 +1,149 @@
...
+  %this.addr = alloca %"class._ZTSZZ4mainENK3$_0clERN2cl4sycl7handlerEEUlvE_.anon" addrspace(4)*, align 8
+  %Buf = alloca [1 x i8], align 1
+  store %"class._ZTSZZ4mainENK3$_0clERN2cl4sycl7handlerEEUlvE_.anon" addrspace(4)* %this, %"class._ZTSZZ4mainENK3$_0clERN2cl4sycl7handlerEEUlvE_.anon" addrspace(4)** %this.addr, align 8, !tbaa !34
+  call void @llvm.dbg.declare(metadata %"class._ZTSZZ4mainENK3$_0clERN2cl4sycl7handlerEEUlvE_.anon" addrspace(4)** %this.addr, metadata !31, metadata !DIExpression()), !dbg !38
+  %this1 = load %"class._ZTSZZ4mainENK3$_0clERN2cl4sycl7handlerEEUlvE_.anon" addrspace(4)*, %"class._ZTSZZ4mainENK3$_0clERN2cl4sycl7handlerEEUlvE_.anon" addrspace(4)** %this.addr, align 8
+  %0 = bitcast [1 x i8]* %Buf to i8*, !dbg !39
+  call void @llvm.lifetime.start.p0i8(i64 1, i8* %0) #5, !dbg !39
+  call void @llvm.dbg.declare(metadata [1 x i8]* %Buf, metadata !33, metadata !DIExpression()), !dbg !40
+  %Buf2 = bitcast [1 x i8]* %Buf to i8*, !dbg !39
+  call void @llvm.var.annotation(i8* %Buf2, i8* getelementptr inbounds ([25 x i8], [25 x i8]* @.str, i32 0, i32 0), i8* getelementptr inbounds ([15 x i8], [15 x i8]* @.str.1, i32 0, i32 0), i32 13), !dbg !39
```
AlexeySotkin: Please add checks for `@llvm.ptr.annotation` and `@llvm.annotation`.

lib/SPIRV/SPIRVWriter.cpp
mlychkov: This fix is changed on more appropriate solution.

test/annotation_dbg_info_drop.ll
mlychkov: Updated the test.

Issue #303, opened 2019-08-20
InvalidFunctionCall: Unexpected llvm intrinsic: llvm.ctlz.i32


davidrohr:
The attached test cails fails during spir-v translation with:
```
InvalidFunctionCall: Unexpected llvm intrinsic: llvm.ctlz.i32 [Src: /var/tmp/portage/dev-util/spirv-llvm-translator-9999/work/spirv-llvm-translator-9999/lib/SPIRV/SPIRVWriter.cpp:1485  ]
Stack dump:
0.      Program arguments: llvm-spirv test.bc 
1.      Running pass 'LLVMToSPIRV' on module 'test.bc'.
 #0 0x00007f7765b4af6a llvm::sys::PrintStackTrace(llvm::raw_ostream&) (/usr/lib/llvm/10/lib64/libLLVMSupport.so.10svn+0x188f6a)
 #1 0x00007f7765b48d64 llvm::sys::RunSignalHandlers() (/usr/lib/llvm/10/lib64/libLLVMSupport.so.10svn+0x186d64)
 #2 0x00007f7765b48ee5 (/usr/lib/llvm/10/lib64/libLLVMSupport.so.10svn+0x186ee5)
 #3 0x00007f77653350d0 (/lib64/libc.so.6+0x390d0)
 #4 0x00007f776533500b raise (/lib64/libc.so.6+0x3900b)
 #5 0x00007f776531e535 abort (/lib64/libc.so.6+0x22535)
 #6 0x00005606619d0d45 SPIRV::SPIRVErrorLog::checkError(bool, SPIRV::SPIRVErrorCode, std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&, char const*, char const*, unsigned int) (/usr/bin/llvm-spirv+0x1e3d45)
 #7 0x00005606619c4208 SPIRV::LLVMToSPIRV::transIntrinsicInst(llvm::IntrinsicInst*, SPIRV::SPIRVBasicBlock*) (/usr/bin/llvm-spirv+0x1d7208)
 #8 0x00005606619be37c SPIRV::LLVMToSPIRV::transValueWithoutDecoration(llvm::Value*, SPIRV::SPIRVBasicBlock*, bool) (/usr/bin/llvm-spirv+0x1d137c)
 #9 0x00005606619bfd9d (/usr/bin/llvm-spirv+0x1d2d9d)
[...]

PR #317, opened 2019-08-29
Translate ctlz/cttz intrinsics to OpExtInst

svenvh:
LLVM transformation passes may introduce calls to these intrinsics and
supporting them in the translator also allows using GCC/Clang builtins
`__builtin_clz` and `__builtin_ctz`.

Fixes https://github.com/KhronosGroup/SPIRV-LLVM-Translator/issues/303


test/count-zero-bits.ll
```diff
@@ -0,0 +1,62 @@
...
+
+; Function Attrs: nounwind readnone
+define spir_func <4 x i32> @TestCtzVec(<4 x i32> %x) local_unnamed_addr #0 {
+entry:
+  %0 = tail call <4 x i32> @llvm.cttz.v4i32(<4 x i32> %x, i1 true)
+  ret <4 x i32> %0
+}
+
+; Function Attrs: nounwind readnone speculatable willreturn
+declare i32 @llvm.ctlz.i32(i32, i1 immarg) #1
```
StuartDBrady: LGTM, but I wonder if we should handle `immarg` as `{{(immarg)?}}` to reduce the diff between branches, or whether it would be better not.

svenvh: It's not that trivial to make `immarg` optional in the LLVM IR input file: I guess we'd have to run some preprocessing step.  I'm inclined to say that avoiding a small diff between previous release branches doesn't justify adding such clutter on master.

StuartDBrady: Good point!  If we had a _very_ lightweight and conventional way of doing the preprocessing, that might be a different matter, but I think anything we do will be too intrusive and also too surprising to be justifiable.

svenvh:
> My only concern is that we probably need to document that IR->SPIRV->IR translation is lossy: I'm afraid that transformation of`@llvm.ctlz.*` intrinsic to call of some external function during such translation might be unexpected for some users

I've added a paragraph to the README to mention this.

AlexeySachkov:
> I've added a paragraph to the README to mention this.

Thanks!

PR #318, opened 2019-08-29
Update Travis and README for release branch

svenvh:


svenvh:
There seems to be a problem with the lldb-9 package (which we need because of the new lit-cpuid dependency):

```
The following packages have unmet dependencies:

 lldb-9 : Breaks: llvm-9-tools (< 1:9~+rc1-1~exp2) but 1:9~svn369744-1~exp1~20190823103928.28 is to be installed

E: Unable to correct problems, you have held broken packages.
```

svenvh:
Superseded by https://github.com/KhronosGroup/SPIRV-LLVM-Translator/pull/323

PR #321, opened 2019-09-02
Translate Intel FPGA attributes for static variables

vmaksimo:
Attributes applied to static variables are generated with
@llvm.global.annotations, not with intrinsics as for "usual" variables.
This patch introduces converting of @llvm.global.annotations to correct SPIR-V
decorations and back.

### DIFF
--- a/lib/SPIRV/SPIRVReader.cpp
+++ b/lib/SPIRV/SPIRVReader.cpp
@@ -2403,16 +2403,8 @@ bool SPIRVToLLVM::translate() {
     if (BV->getStorageClass() != StorageClassFunction)
       transValue(BV, nullptr, nullptr);
   }
-  if (!GlobalAnnotations.empty()) {
-    Constant *Array =
-        ConstantArray::get(ArrayType::get(GlobalAnnotations[0]->getType(),
-                                          GlobalAnnotations.size()),
-                           GlobalAnnotations);
-    auto *GV = new GlobalVariable(*M, Array->getType(), /*IsConstant*/ false,
-                                  GlobalValue::AppendingLinkage, Array,
-                                  "llvm.global.annotations");
-    GV->setSection("llvm.metadata");
-  }
+  transGlobalAnnotations();
+
   // Compile unit might be needed during translation of debug intrinsics.
   for (SPIRVExtInst *EI : BM->getDebugInstVec()) {
     // Translate Compile Unit first.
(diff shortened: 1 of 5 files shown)


lib/SPIRV/SPIRVWriter.cpp
```diff
@@ -1585,9 +1585,44 @@ SPIRVWord LLVMToSPIRV::transFunctionControlMask(Function *F) {
...
+void LLVMToSPIRV::transGlobalAnnotation(GlobalVariable *V) {
+  SPIRVDBG(dbgs() << "[transGlobalAnnotation] " << *V << '\n');
+
+  // @llvm.global.annotations is an array that contains structs with 4 fields.
+  // Get the array of structs with metadata
+  ConstantArray *CA = cast<ConstantArray>(V->getOperand(0));
+  for (Value *Op : CA->operands()) {
+    ConstantStruct *CS = cast<ConstantStruct>(Op);
+    // The first field of the struct contains a pointer to annotated variable
+    Value *AnnotatedVar = CS->getOperand(0)->getOperand(0);
```
AlexeySotkin: I guess you assume `CS->getOperand(0)` is a bitcast instruction. Will it work if we annotate a char ?

lib/SPIRV/SPIRVWriter.cpp
```diff
@@ -1585,9 +1585,44 @@ SPIRVWord LLVMToSPIRV::transFunctionControlMask(Function *F) {
...
+    SPIRVValue *SV = transValue(AnnotatedVar, nullptr);
+
+    // The second field contains a pointer to a global annotation string
+    GlobalVariable *GV = cast<GlobalVariable>(CS->getOperand(1)->getOperand(0));
```
AlexeySotkin: Can we use stripPointerCasts() here?

lib/SPIRV/SPIRVWriter.cpp
```diff
@@ -1585,9 +1585,44 @@ SPIRVWord LLVMToSPIRV::transFunctionControlMask(Function *F) {
...
+    // TODO: Refactor to use getConstantStringInfo()
```
AlexeySotkin: Should we use getConstantStringInfo() now ?

lib/SPIRV/SPIRVWriter.cpp
```diff
@@ -1585,9 +1585,44 @@ SPIRVWord LLVMToSPIRV::transFunctionControlMask(Function *F) {
...
+    } else {
+      addIntelFPGADecorations(SV, Decorations);
+    }
+  }
+}
+
 bool LLVMToSPIRV::transGlobalVariables() {
   for (auto I = M->global_begin(), E = M->global_end(); I != E; ++I) {
-    if (!transValue(&(*I), nullptr))
+    if ((&(*I))->getName() == "llvm.global.annotations")
```
AlexeySotkin: Please move this `if` to `LLVMToSPIRV::transValueWithoutDecoration`

lib/SPIRV/SPIRVReader.cpp
```diff
@@ -2403,6 +2403,16 @@ bool SPIRVToLLVM::translate() {
     if (BV->getStorageClass() != StorageClassFunction)
       transValue(BV, nullptr, nullptr);
   }
+  if (!GlobalAnnotations.empty()) {
```
AlexeySotkin: Please factor it out into a function or method

lib/SPIRV/SPIRVWriter.cpp
vmaksimo: I think that it's better to do this refactoring in a separate commit as there are another three similar uses.

lib/SPIRV/SPIRVWriter.cpp
vmaksimo: Thank you for noticing, this code will not work properly with char variables :)

But I don't understand what is the advantage of using ->stripPointerCasts() instead just ->getOperand(0). Does it improve readability?

PR #324, opened 2019-09-07
Add support for MatrixTimesScalar

jigsawecho:
Support for Matrix type, OpTypeMatrix, is added with support
to OpMatrixTimesScalar. New test case is added accordingly.

Signed-off-by: Qinglai Xiao <q.xiao@think-silicon.com>


jigsawecho:
Tidy check fails because SPIRVType.h is in the patch. But I suppose SPIRVType.h shall not be fixed in this patch, right?

svenvh:
Thanks for your patch!

You are only adding support for **reading** `OpMatrixTimesScalar`, so it would be good if you could mention that in your commit message.

> Tidy check fails because SPIRVType.h is in the patch. But I suppose SPIRVType.h shall not be fixed in this patch, right?

Is the problem simply that you are not adding `override` to the overridden methods that you are adding?

jigsawecho:
Hi @svenvh

Yeah, the patch only translates matrix mult scalar to llvm IR, but not the other way around. I will update the commit message accordingly. 

The SPIRVType.h has a dozen of tidy problems due to missing override. So it seems to me not proper to fix all of those function declarations in this patch. Please advise.

svenvh:
> The SPIRVType.h has a dozen of tidy problems due to missing override.

Can you point out which clang-tidy problems you think are not caused by your patch?  The first few problems in the list seem to come from your patch, for example:

```
/home/travis/build/KhronosGroup/SPIRV-LLVM-Translator/lib/SPIRV/libSPIRV/SPIRVType.h:333:15: warning: 'getRequiredCapability' overrides a member function but is not marked 'override'

/home/travis/build/KhronosGroup/SPIRV-LLVM-Translator/lib/SPIRV/libSPIRV/SPIRVType.h:340:37: warning: 'getNonLiteralOperands' overrides a member function but is not marked 'override'

/home/travis/build/KhronosGroup/SPIRV-LLVM-Translator/lib/SPIRV/libSPIRV/SPIRVType.h:344:8: warning: 'validate' overrides a member function but is not marked 'override'
```

jigsawecho:
Hi @svenvh, 

Oh my bad. Yes you are right. I will push another patch later today. Thanks for your reply.

jigsawecho:
Hi @svenvh ,

This time the tidy check went fine. But 2 other tests failed due to apt-get failure. Could you please re-trigger the test? Thanks in advance.

AlexeySachkov:
Hi @jigsawecho, I've re-started Travis jobs. FYI: you can re-trigget travis by force pushing to your branch (you could use `git commit --amend --no-edit` to re-calculate hash of the last commit to be sure).

lib/SPIRV/libSPIRV/SPIRVInstruction.h
```diff
@@ -1248,6 +1248,56 @@ class SPIRVVectorTimesScalar : public SPIRVInstruction {
...
+  void validate() const override {
+    SPIRVInstruction::validate();
+    if (getValue(Matrix)->isForward() || getValue(Scalar)->isForward())
+      return;
+
+    SPIRVType *Ty = getType()->getScalarType();
+    SPIRVType *MTy = getValueType(Matrix)->getScalarType();
+    SPIRVType *STy = getValueType(Scalar);
+
+    assert(Ty == MTy && Ty == STy && "Float type mismatch");
```
svenvh: `"Float type mismatch"` is not the most helpful message.  In fact the code is not checking for floats at all as far as I can see?

Perhaps you can break up the assert into two asserts, one for the matrix operand and one for the scalar operand, with appropriate messages.

test/matrix_times_scalar.spt
```diff
@@ -0,0 +1,57 @@
...
+4 Load 6 17 14
+5 MatrixTimesScalar 8 18 16 17
+3 Store 12 18
+1 Return
+1 FunctionEnd
+
+; FIXME: LIT comments/commands are moved at the end because llvm-spirv stops
+; reading the file after first ';' symbol
+
+; RUN: llvm-spirv %s -to-binary -o %t.spv
```
AlexeySachkov: I think it worth launching `spirv-val` to ensure that this source file is "valid" in terms of SPIR-V validation rules

jigsawecho: Yes added in next patch. It really captured some error. Thanks.

lib/SPIRV/libSPIRV/SPIRVInstruction.h
jigsawecho: Added in next patch. Pls advise if the assert message is appropriate. Thanks.

PR #328, opened 2019-09-12
[SYCL][FPGA] Fix loop metadata layout in reverse translation

mlychkov:
In accordance with the SPIR-V spec loop metadata should be attached to the header block of the loop (via OpLoopMerge or OpLoopControlINTEL) whilst in LLVM IR it should be attached to latch block of the loop (via llvm.loop metadata).
Fix attachment place of loop metadata in reverse translation.

### DIFF
--- a/lib/SPIRV/SPIRVReader.cpp
+++ b/lib/SPIRV/SPIRVReader.cpp
@@ -51,6 +51,7 @@
 #include "SPIRVValue.h"
 
 #include "llvm/ADT/DenseMap.h"
+#include "llvm/Analysis/CFG.h"
 #include "llvm/BinaryFormat/Dwarf.h"
 #include "llvm/IR/Constants.h"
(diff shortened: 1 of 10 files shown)


lib/SPIRV/SPIRVReader.cpp
```diff
@@ -694,6 +695,108 @@ void SPIRVToLLVM::setLLVMLoopMetadata(LoopInstType *LM, BranchInst *BI) {
...
+    return false;
+  }
+
+  default:
+    return false;
+  }
+}
+
+static inline bool isBBReachableFrom(const SPIRVValue *CB, SPIRVId TBId) {
+  std::unordered_set<SPIRVId> VBBIds{};
```
keryell: `std::unordered_set<SPIRVId> VBBIds;` is enough

mlychkov: Thank you. Fixed.

Issue #331, opened 2019-09-16
enqueue_kernel builtins translation - SPIRV to LLVM fails


hewj03:
OpenCL CTS Device_execution21_offline/test_enqueue_block failure is caused by the following commit:
commit d68d83bed67861af79999dbf1edcbd480778ec51
Author: Andrzej Ratajewski <andrzej.ratajewski@intel.com>
Date:   Mon Aug 19 12:48:43 2019 +0200
    Fix enqueue_kernel builtins translation - SPIRV to LLVM

The following kernel is translated to __enqueue_kernel_basic instead of __enqueue_kernel_basic_events:
    int enq_res = enqueue_kernel(def_q, CLK_ENQUEUE_FLAGS_NO_WAIT, ndrange, 0, NULL, &block_evt1, kernelBlock);

AlexeySachkov:
@aratajew: FYI

PR #332, opened 2019-09-16
Fix enqueue_kernel translation

hewj03:
Set HasEvents to true if num_events_in_wait_list is NULL and event_ret
is not NULL. See #331

### DIFF
--- a/lib/SPIRV/SPIRVReader.cpp
+++ b/lib/SPIRV/SPIRVReader.cpp
@@ -2109,10 +2109,14 @@ Instruction *SPIRVToLLVM::transEnqueueKernelBI(SPIRVInstruction *BI,
   auto Ops = BI->getOperands();
   bool HasVaargs = Ops.size() > 10;
   bool HasEvents = true;
-  SPIRVValue *NumEvents = Ops[3];
-  if (NumEvents->getOpCode() == OpConstant) {
-    SPIRVConstant *NE = static_cast<SPIRVConstant *>(NumEvents);
-    HasEvents = NE->getZExtIntValue() != 0;
+  SPIRVValue *EventRet = Ops[5];
+  if (EventRet->getOpCode() == OpConstantNull) {
+    SPIRVValue *NumEvents = Ops[3];
+    if (NumEvents->getOpCode() == OpConstant) {
+      SPIRVConstant *NE = static_cast<SPIRVConstant *>(NumEvents);
+      HasEvents = NE->getZExtIntValue() != 0;
+    } else if (NumEvents->getOpCode() == OpConstantNull)
+      HasEvents = false;
   }
 
   std::string FName = "";
--- a/test/transcoding/enqueue_kernel.cl
+++ b/test/transcoding/enqueue_kernel.cl
@@ -10,6 +10,7 @@
 // CHECK-SPIRV: EntryPoint {{[0-9]+}} [[BlockKer2:[0-9]+]] "__device_side_enqueue_block_invoke_2_kernel"
 // CHECK-SPIRV: EntryPoint {{[0-9]+}} [[BlockKer3:[0-9]+]] "__device_side_enqueue_block_invoke_3_kernel"
 // CHECK-SPIRV: EntryPoint {{[0-9]+}} [[BlockKer4:[0-9]+]] "__device_side_enqueue_block_invoke_4_kernel"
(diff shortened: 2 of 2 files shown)


lib/SPIRV/SPIRVReader.cpp
```diff
@@ -2114,6 +2114,9 @@ Instruction *SPIRVToLLVM::transEnqueueKernelBI(SPIRVInstruction *BI,
     SPIRVConstant *NE = static_cast<SPIRVConstant *>(NumEvents);
     HasEvents = NE->getZExtIntValue() != 0;
   }
+  SPIRVValue *EventRet = Ops[5];
+  if (EventRet->getOpCode() == OpPtrCastToGeneric)
```
AlexeySotkin: Why `OpPtrCastToGeneric` ?

aratajew: I'd prefer: `if (EventRet->getOpCode() != OpConstantNull)`. It's more readable.

lib/SPIRV/SPIRVReader.cpp
```diff
@@ -2114,6 +2114,9 @@ Instruction *SPIRVToLLVM::transEnqueueKernelBI(SPIRVInstruction *BI,
     SPIRVConstant *NE = static_cast<SPIRVConstant *>(NumEvents);
```
aratajew: I think we could move below condition here, like:
`if (EventRet->getOpCode() == OpConstantNull && NumEvents->getOpCode() == OpConstant)`
As it doesn't make sense to set HasEvents to false and then back to true in the next condition.

aratajew: This comment refers to line 2113.

hewj03: done

aratajew: LGTM, thanks!

test/transcoding/enqueue_kernel.cl
```diff
@@ -40,8 +41,6 @@
 // CHECK-LLVM: @__block_literal_global = internal addrspace(1) constant [[BlockTy1]] { i32 12, i32 4, i8 addrspace(4)* addrspacecast (i8* null to i8 addrspace(4)*) }, align 4
 // CHECK-LLVM: @__block_literal_global.1 = internal addrspace(1) constant [[BlockTy1]] { i32 12, i32 4, i8 addrspace(4)* addrspacecast (i8* null to i8 addrspace(4)*) }, align 4
 
-typedef struct {int a;} ndrange_t;
```
AlexeySachkov: It would be nice if it is possible to leave this as-is: `opencl-c.h` which is included by `-finclude-default-header` is pretty big and will slow down the test execution

hewj03: NULL is defined in opencl-c.h
So adding a NULL define here?

AlexeySachkov: Yep, there is no problem with copying definition of NULL here

hewj03: done

lib/SPIRV/SPIRVReader.cpp
```diff
@@ -2110,7 +2110,9 @@ Instruction *SPIRVToLLVM::transEnqueueKernelBI(SPIRVInstruction *BI,
   bool HasVaargs = Ops.size() > 10;
   bool HasEvents = true;
   SPIRVValue *NumEvents = Ops[3];
-  if (NumEvents->getOpCode() == OpConstant) {
+  SPIRVValue *EventRet = Ops[5];
+  if (EventRet->getOpCode() == OpConstantNull &&
+      NumEvents->getOpCode() == OpConstant) {
```
AlexeySotkin: I guess `NumEvents` as well can be an `OpConstantNull`. Should we take it into account ?

hewj03: I don't know if NumEvents can be OpConstantNull. @AlexeySachkov @aratajew do you have any comments?

aratajew: Current support for enqueue_kernel translation from LLVM to SPIRV does not produce OpConstantNull. It produces OpConstant with 0 value. But OpConstantNull as an NumEvents argument is also correct in SPIRV, so we should take it into account, indeed.

hewj03: I've updated the code.

PR #333, opened 2019-09-16
[Do not merge] Make SPIRV-Tools a hard dependency for tests

svenvh:
Including a throwaway commit to test internal builds.  Will remove that before merging.


svenvh:
@StuartDBrady, internal builds are indeed failing with
```
CMake Error at /home/travis/build/KhronosGroup/SPIRV-LLVM-Translator/llvm-project/llvm-spirv/test/CMakeLists.txt:27 (message):

  spirv-as not found!
```
when not setting any value for `LLVM_INCLUDE_TESTS`, as its default is `ON` for internal builds.

test/CMakeLists.txt
```diff
@@ -24,12 +24,12 @@ if(SPIRV_TOOLS_FOUND)
 endif()
 
 if(NOT SPIRV_TOOLS_SPIRV_AS)
-  message(WARNING "spirv-as not found! SPIR-V assembly tests will not be run.")
+  message(FATAL_ERROR "spirv-as not found!")
   set(SPIRV_TOOLS_SPIRV_AS_FOUND False)
```
StuartDBrady: Should we also remove `SPIRV_TOOLS_SPIRV_AS_FOUND`?

If so, `config.spirv_tools_have_spirv_as` will need to be removed from `lit.site.cfg.py.in`, too.

test/CMakeLists.txt
```diff
@@ -24,12 +24,12 @@ if(SPIRV_TOOLS_FOUND)
...
 endif()
 
 if(NOT SPIRV_TOOLS_SPIRV_VAL)
-  message(WARNING "spirv-val not found! SPIR-V generated for test suite will not be validated.")
+  message(FATAL_ERROR "spirv-val not found!")
   set(SPIRV_TOOLS_SPIRV_VAL_FOUND False)
```
StuartDBrady: Ditto for `SPIRV_TOOLS_SPIRV_VAL_FOUND` and `config.spirv_tools_have_spirv_val`.

test/lit.cfg.py
```diff
@@ -57,14 +57,11 @@
 
 if config.spirv_tools_have_spirv_as:
```
StuartDBrady: Should we remove the if statement here, and unindent the following lines?

test/lit.cfg.py
```diff
@@ -57,14 +57,11 @@
...
     llvm_config.add_tool_substitutions(['spirv-as'], [config.spirv_tools_bin_dir])
-    config.available_features.add('spirv-as')
     using_spirv_tools = True
 
 if config.spirv_tools_have_spirv_val:
```
StuartDBrady: Ditto.

test/lit.cfg.py
```diff
@@ -57,14 +57,11 @@
...
     llvm_config.add_tool_substitutions(['spirv-val'], [config.spirv_tools_bin_dir])
     using_spirv_tools = True
-else:
-    config.substitutions.append(('spirv-val', ':'))
 
 if using_spirv_tools:
```
StuartDBrady: Should also we remove `using_spirv_tools` and the lines above that set it, and unindent the following lines?

test/lit.cfg.py
```diff
@@ -57,14 +57,11 @@
 
 if config.spirv_tools_have_spirv_as:
     llvm_config.add_tool_substitutions(['spirv-as'], [config.spirv_tools_bin_dir])
-    config.available_features.add('spirv-as')
     using_spirv_tools = True
 
 if config.spirv_tools_have_spirv_val:
     llvm_config.add_tool_substitutions(['spirv-val'], [config.spirv_tools_bin_dir])
```
StuartDBrady: Should we add `spirv-as` and `spirv-val'` with a single call to `add_tool_substitutions()`, as per `llvm-as`, `llvm-dis`, etc.?

Should we also rename `tools` to `llvm_tools`?

test/CMakeLists.txt
```diff
@@ -11,26 +11,18 @@ set(LLVM_SPIRV_TEST_SOURCE_DIR ${CMAKE_CURRENT_SOURCE_DIR})
...
-    set(SPIRV_TOOLS_SPIRV_VAL_FOUND True)
-  endif()
 
   set(SPIRV_TOOLS_BINDIR "${SPIRV_TOOLS_PREFIX}/bin")
 endif()
 
 if(NOT SPIRV_TOOLS_SPIRV_AS)
-  message(WARNING "spirv-as not found! SPIR-V assembly tests will not be run.")
-  set(SPIRV_TOOLS_SPIRV_AS_FOUND False)
+  message(FATAL_ERROR "spirv-as not found!")
```
AlexeySachkov: I think that if we a bit afraid that downstream users are not ready to this changes then we could design them in a way that they could be easily reverted by the users: I mean, let's by default have strong dependency on SPIRV-Tools, but also let's leave an option which allows to make this dependency optional and partially disable some tests.

In this case we can commit this patch sooner and remove this option later (let's say in a month or so from PR merge)

```
$ cmake -DLLVM_ENABLE_TESTING=ON ..
...
FATAL_ERROR: spirv-tools are required to test the translator.
For transition period you can specify 'SPIRV_TOOLS_FORCE_DISABLE=ON' to skip these tests.
Note: this option is only intended for transition period and will be removed in the upcoming revisions of the translator.
cmake failed
$ cmake -DLLVM_ENABLE_TESTING=ON -DSPIRV_TOOLS_FORCE_DISABLE=ON ..
...
WARNING: You are using 'SPIRV_TOOLS_FORCE_DISABLE' which will be removed in the upcoming revisions of the translator.
Please configure your environment to include SPIRV-Tools as well to test the translator.
Some tests will be disabled for now.
cmake passed
```

I understand that such change is basically almost a "revert" of this patch, but we can look at this as an another way of communicating such important upcoming change for our downstream users.

@svenvh, @StuartDBrady, @AlexeySotkin, @bader, what do you think?

bader: Sounds good to me.

svenvh: Sounds reasonable, I'll prepare an alternative patch.

AlexeySachkov: @svenvh, One more idea from @AlexeySotkin: right now this patch relies on `LLVM_INCLUDE_TESTS` and basically, there is no way for user to disable tests only for the translator.

If we add something like `LLVM_SPIRV_INCLUDE_TESTS` to add possibility for users to disable testing specifically for the translator, then we probably can go with this patch as-is (almost).

Summarizing, two options:
* rely on `LLVM_INCLUDE_TESTS`, but add possibility to temporary disable new hard requirement for SPIRV-Tools
* rely on separate `LLVM_SPRIV_INCLUDE_TESTS` - this patch can be left almost as-is then

AlexeySotkin: Looks like it is a common practice in the LLVM project to have an option to enable/disable tests for subprojects. For example, there are [CLANG_INCLUDE_TESTS](https://github.com/llvm/llvm-project/blob/559ae14c9b6cf203e0e7495892080e52e89c1ea0/clang-tools-extra/CMakeLists.txt#L17), 
[LIBCXX_INCLUDE_TESTS](https://github.com/llvm/llvm-project/blob/80ef2f05d878fd6b00bbfb2ed1c04e2657d241cc/libcxx/CMakeLists.txt#L82) and [LLDB_INCLUDE_TESTS](https://github.com/llvm/llvm-project/blob/ff5982aa91d1ca0c11e00b3d2b22d32c66bb8590/lldb/CMakeLists.txt#L72)

svenvh: > rely on separate `LLVM_SPRIV_INCLUDE_TESTS` - this patch can be left almost as-is then

Introducing `LLVM_SPIRV_INCLUDE_TESTS` and this patch does not allow partially disabling translator tests.  But it will allow disabling the entire translator test suite while still allowing to run Clang/LLVM tests.  Do you still want to have a transition period before making SPIRV-Tools a hard dependency (to allow partially running the test suite in the mean time)?

I agree that `LLVM_SPIRV_INCLUDE_TESTS` is a good thing to have, so I've opened https://github.com/KhronosGroup/SPIRV-LLVM-Translator/pull/335

AlexeySachkov: @svenvh, personally, I think that if we have a separate option to enable/disable testing of the translator, then we don't need to invent any workarounds or transition periods.

So, from my point of view, I would go with #335 first and then this PR can be committed as-is: any user who wants to test the translator must get SPIRV-Tools in his environment and at the same time he can disable the translator testing at all for the transition period if such is required.

Comparing to that, `SPIRV_TOOLS_FORCE_DISABLE`, looks like a hack :)

svenvh: > @svenvh, personally, I think that if we have a separate option to enable/disable testing of the translator, then we don't need to invent any workarounds or transition periods.
> 

Fair enough, I have updated the fatal error message to mention how llvm-spirv tests can be disabled.

Keeping open for a bit longer such that any other objections can be raised before merging.

test/lit.cfg.py
```diff
@@ -53,19 +53,7 @@
...
-if config.spirv_tools_have_spirv_val:
-    llvm_config.add_tool_substitutions(['spirv-val'], [config.spirv_tools_bin_dir])
-    using_spirv_tools = True
-else:
-    config.substitutions.append(('spirv-val', ':'))
-
-if using_spirv_tools:
-    new_ld_library_path = os.path.pathsep.join((config.spirv_tools_lib_dir, config.environment['LD_LIBRARY_PATH']))
-    config.environment['LD_LIBRARY_PATH'] = new_ld_library_path
+llvm_config.add_tool_substitutions(['spirv-as', 'spirv-val'], [config.spirv_tools_bin_dir])
```
AlexeySachkov: Minor: we could `spirv-as` and `spirv-val` into existing `tools` list to simplify the code. `config.spirv_tools_bin_dir` can be included into existing `tools_dir` list in this case

README.md
```diff
@@ -64,7 +64,7 @@ make llvm-spirv -j`nproc`
 
 ## Test instructions
 
-All tests related to the translator are placed in the [test](test) directory. A number of the tests require spirv-as (part of SPIR-V Tools) to run, but the remainder of the tests can still be run without this. Optionally the tests can make use of spirv-val (part of SPIRV-Tools) in order to validate the generated SPIR-V against the official SPIR-V specification.
+All translator tests are placed in the [test](test) directory. Some of the tests require spirv-as (part of SPIR-V Tools). The tests make use of spirv-val (part of SPIRV-Tools) to validate the generated SPIR-V against the official SPIR-V specification.
```
AlexeySachkov:
```suggestion
All translator tests are placed in the [test](test) directory.
Some of the tests require [`spirv-as`](https://github.com/KhronosGroup/SPIRV-Tools#assembler-tool) (part of SPIR-V Tools).
The tests make use of [`spirv-val`](https://github.com/KhronosGroup/SPIRV-Tools#validator-tool) (part of SPIRV-Tools) to validate the generated SPIR-V against the official SPIR-V specification.
```

test/lit.cfg.py
```diff
@@ -53,19 +53,7 @@
...
+
+new_ld_library_path = os.path.pathsep.join((config.spirv_tools_lib_dir, config.environment['LD_LIBRARY_PATH']))
+config.environment['LD_LIBRARY_PATH'] = new_ld_library_path
```
AlexeySotkin: Minor. Probably we can get rid of `new_ld_library_path`.
```suggestion
config.environment['LD_LIBRARY_PATH'] = os.path.pathsep.join((
    config.spirv_tools_lib_dir, config.environment['LD_LIBRARY_PATH']))
```

svenvh:
Closing after today's OpenCL Tooling TSG teleconference: we currently include SPIR-V Tools in the CI jobs and as such there is no strict need for this right now. We have accepted many contributions since this PR was opened without any issues. Also there were concerns that this might be a nuisance for developers without adding much value.

PR #335, opened 2019-09-20
cmake: Add LLVM_SPIRV_INCLUDE_TESTS option

svenvh:
Make it possible to disable the llvm-spirv test suite while still
allowing LLVM_INCLUDE_TESTS to be set to true.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -2,6 +2,10 @@ cmake_minimum_required(VERSION 3.3)
 
 set(LLVM_SPIRV_VERSION 0.2.1.0)
 
+option(LLVM_SPIRV_INCLUDE_TESTS
+  "Generate build targets for the llvm-spirv lit tests."
+  ${LLVM_INCLUDE_TESTS})
+
 # check if we build inside llvm or not
 if(CMAKE_SOURCE_DIR STREQUAL CMAKE_CURRENT_SOURCE_DIR)
   set(BUILD_EXTERNAL YES)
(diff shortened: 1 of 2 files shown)


CMakeLists.txt
```diff
@@ -2,6 +2,9 @@ cmake_minimum_required(VERSION 3.3)
 
 set(LLVM_SPIRV_VERSION 0.2.1.0)
 
+option(LLVM_SPIRV_INCLUDE_TESTS
```
AlexeySachkov: I think it would be perfect if this option (unless explicitly specified) followed `LLVM_INCLUDE_TESTS`.

I.e. tests for the translator are enabled or disabled by default based on LLVM settings, but this can be specifically configured if user don't want to follow LLVM settings.

@svenvh, if this is not so easy to do, or you don't have enough time, I'm good with this patch as-is, thanks!

svenvh: I can change the default initialization to `LLVM_INCLUDE_TESTS` instead of `ON`, if that is preferable?

AlexeySachkov: > I can change the default initialization to `LLVM_INCLUDE_TESTS` instead of `ON`, if that is preferable?

If it works it would be great, because it basically will preserve backward compatibility: right now translator follows `LLVM_INCLUDE_TESTS` and this patch breaks this behavior

svenvh: Defaulting to `LLVM_INCLUDE_TESTS` seems to be what `CLANG_INCLUDE_TESTS` and `LLDB_INCLUDE_TESTS` do as well, so it makes sense. Thanks for the suggestion!

PR #341, opened 2019-09-27
Added creation of a deb package

KOLANICH:
The description is taken from Debian package.


AlexeySachkov:
@dvrogozh, @tjaalton: FYI

KOLANICH:
BTW, I am not going to sign any CLAs. But my contribution can be considered as licensed under [Unlicense](https://unlicense.org/).

dvrogozh:
That's my private opinion... To me that's a wrong way since distros in any case don't use this path and create packages on its own. This just becomes a death code in build system which needs to be supported. If packages are needed for some other OS it is always possible to borrow scripts from existing one and modify accordingly or write from scratch, that's not that difficult actually. Another argument is that per my experience cpack has a lot of issues and I always stepped into something which is either not fixed or requires absolutely new most recent version which is totally annoying. And if you use older version of cpack you just can't make what you want to make.

KOLANICH:
>To me that's a wrong way since distros in any case don't use this path and create packages on its own.

It is mainly not for distros (bit I personally think that distros should use CMake instead of debhelper gecause debhelper is a very glitchy and fragile software, especially when combined with CMake). It is for the ones needing bleeding edge packages. `make install` or `ninja install` are problematic and unrecommended way to install software. They both cause problems and conflicts and also have no automated way to remove. (And manual way is also problematic, for example yesterday I have accidentially nuked /usr/lib/x86_64-gnu-linux).

>This just becomes a death code in build system which needs to be supported. 

No, it won't. The ones needing bleeding edge either have to use checkinstall and input the same info multiple times manually, or somehow build a package or just use `ninja install` making mess.

>If packages are needed for some other OS it is always possible to borrow scripts from existing one and modify accordingly or write from scratch

Dealing with debhelper is extremily painful. IMHO: debhelper must be retired as soon as possible and a better tool must be written in python.

>Another argument is that per my experience cpack has a lot of issues and I always stepped into something which is either not fixed

Yeah, it is a pain when something is implemented in CMake incorrectly or too in a too restricted way (IMHO CMake is itself terrible programming language, it would be better to take some extremily lightweight embeddable scripting engine like luajit or QuickJS, reimplement some CMake API in it and transpile the existing scripts to it): one has to send a PR or leave an issue to their GitLab, but their GitLab is plagued with reCAPTCHA and also they have issues ignored for years.

>or requires absolutely new most recent version

IMHO it is completely OK, as long as a deb repo is available. I have created a one with cmake packages deb created from official binary releases. They are not suitable for desktop, I cannot expect security from a shared CI (in practice it is impossible to sandbox Turing-complete language), but for using these packages for CI on the same GitLab - why not?. The ones needing the freshest CMake packages can just clone the repo, audit its code (it is small enough) and build the package locally.

KOLANICH:
Is it going to be merged?

AlexeySotkin:
Sorry for the late reply.
This patch breaks the tests in travis-ci. @KOLANICH could you fix it? Thanks.

AlexeySotkin:
Also, regarding the CLA. Sorry, but I think we can't accept your contribution without signing the CLA.

KOLANICH:
>Sorry, but I think we can't accept your contribution without signing the CLA.

Are you absolutely sure that you cannot read my patches and recreate them from your own account?

PR #342, opened 2019-09-30
Fix detection of recursive data types in presence of function pointers

AlexeySachkov:


lib/SPIRV/SPIRVWriter.cpp
```diff
@@ -224,8 +224,18 @@ static bool recursiveType(const StructType *ST, const Type *Ty) {
...
 
-    if (auto *PtrTy = dyn_cast<PointerType>(Ty))
-      return Run(PtrTy->getPointerElementType());
+    if (auto *PtrTy = dyn_cast<PointerType>(Ty)) {
+      auto *ElTy = PtrTy->getPointerElementType();
+      if (auto *FTy = dyn_cast<FunctionType>(ElTy)) {
+        // If we have a function pointer, then argument types and return type of
+        // the referenced function also need to be checked
+        return Run(FTy->getReturnType()) ||
+               find_if(FTy->param_begin(), FTy->param_end(), Run) !=
```
AlexeySotkin: Minor. Can we use `llvm::any_of` or `std::any_of` here ?

test/transcoding/SPV_INTEL_function_pointers/fp-in-recusive-type.ll
```diff
@@ -0,0 +1,84 @@
...
+; CHECK-SPIRV: TypeStruct [[TYPETY:[0-9]+]] [[INTTY]] [[FPTRTY:[0-9]+]]
+; CHECK-SPIRV: TypeStruct [[DESCTY:[0-9]+]] [[INTTY]] [[TYPEPTRTY:[0-9]+]]
+; CHECK-SPIRV: TypePointer [[TYPEPTRTY]] {{[0-9]+}} [[TYPETY]]
+; CHECK-SPIRV: TypePointer [[DESCPTRTY:[0-9]+]] {{[0-9]+}} [[DESCTY]]
+; CHECK-SPIRV: TypeFunction [[FTY:[0-9]+]] [[VOIDTY]] [[DESCPTRTY]]
+; CHECK-SPIRV: TypePointer [[FPTRTY]] {{[0-9]+}} [[FTY]]
+; CHECK-SPIRV: TypePointer [[DESCPTR1TY:[0-9]+]] {{[0-9]+}} [[TYPETY]]
+; CHECK-SPIRV: Variable [[DESCPTR1TY]]
+;
+; CHECK-LLVM: %[[TYPE:.*]].Type = type { i32, void (%[[DESC:.*]].Desc addrspace(4)*) addrspace(4)* }
```
AlexeySotkin: Did you want to capture just `struct` or `struct.Type` ?
The same for question for `DESC`.

AlexeySachkov: Oh, fair point, there is no sense in capturing `struct` and I don't really need any filecheck variable here. I think I will remove `%[[TYPE]]` and `%[[DESC]]` at all and replace it with `struct.Type` and `struct.Desc`

lib/SPIRV/SPIRVWriter.cpp
AlexeySachkov: Done

AlexeySachkov:
@svenvh, do you have any objections against merging this PR (or any comments)?

PR #351, opened 2019-10-17
[Backport] Translate Intel FPGA attributes for static variables

AlexeySachkov:
Attributes applied to static variables are generated with
@llvm.global.annotations, not with intrinsics as for "usual" variables.
This patch introduces converting of @llvm.global.annotations to correct SPIR-V
decorations and back.


AlexeySachkov:
@PawelJurek: FYI

PR #352, opened 2019-10-17
Add nullptr checks to avoid potential nullptr dereferencing

AGindinson:


lib/SPIRV/LLVMToSPIRVDbgTran.cpp
```diff
@@ -208,6 +208,8 @@ void LLVMToSPIRVDbgTran::transLocationInfo() {
           // According to the spec, OpLine for an OpBranch/OpBranchConditional
           // must precede the merge instruction and not the branch instruction
           auto *VPrev = static_cast<SPIRVInstruction *>(V)->getPrevious();
+          if (!VPrev)
```
AlexeySotkin: Even if `VPref` is null, we still need to call `BM->addLine()`

AGindinson: True. At first I cringed that regression tests didn't catch this.

AGindinson: Done

PR #353, opened 2019-10-18
Reenable optimization of OpCopyMemorySized operation to memset intrinsic

aratajew:
Use optimization of OpCopyMemorySized operation to memset intrinsic if element size of zero-initialized array differs from 1 byte.

Change created by: Alex Paige <alexander.paige@intel.com>

The following change disabled the optimization if element size of array differs from 1 byte:
https://github.com/KhronosGroup/SPIRV-LLVM-Translator/commit/60047ee612adcf25ad6228d0f64e244018d8f111
It can be reenabled by inserting pointer cast instruction.


AlexeySachkov:
@mlychkov: FYI

lib/SPIRV/SPIRVReader.cpp
```diff
@@ -1513,12 +1513,16 @@ Value *SPIRVToLLVM::transValueWithoutDecoration(SPIRVValue *BV, Function *F,
...
+            Type *Int8Ty = Type::getInt8Ty(Dst->getContext());
+            llvm::Value *Src = ConstantInt::get(Int8Ty, 0);
+            llvm::Value *NewDst = Dst;
+            if (!Dst->getType()->getPointerElementType()->isIntegerTy(8)) {
+              Type *Int8PointerTy = Type::getInt8PtrTy(
+                  Dst->getContext(), Dst->getType()->getPointerAddressSpace());
+              NewDst = llvm::BitCastInst::CreatePointerCast(Dst, Int8PointerTy,
+                                                            "", BB);
             }
+            CI = Builder.CreateMemSet(NewDst, Src, Size, Align, IsVolatile);
```
mlychkov: Is the Size always equal to number of bytes that the copied memory utilizes?

aratajew: Yes, it comes from OpCopyMemorySized Size operand:
"Size is the number of bytes to copy.". It is independent of the type of pointer. CreateMemSet also takes number of bytes to set as a Size argument.

PR #354, opened 2019-10-18
Fix translation of llvm.global.annotation

vmaksimo:
This patch fixes the problem that if LLVM IR contains llvm.global.annotation,
translator drops all the rest code except this variable.

### DIFF
--- a/lib/SPIRV/SPIRVWriter.cpp
+++ b/lib/SPIRV/SPIRVWriter.cpp
@@ -725,11 +725,6 @@ SPIRVValue *LLVMToSPIRV::transValueWithoutDecoration(Value *V,
     return transFunctionDecl(F);
 
   if (auto GV = dyn_cast<GlobalVariable>(V)) {
-    if (GV->getName() == "llvm.global.annotations") {
-      transGlobalAnnotation(GV);
-      return nullptr;
-    }
-
     llvm::PointerType *Ty = GV->getType();
     // Though variables with common linkage type are initialized by 0,
     // they can be represented in SPIR-V as uninitialized variables with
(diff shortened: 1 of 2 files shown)


lib/SPIRV/SPIRVWriter.cpp
```diff
@@ -727,7 +727,6 @@ SPIRVValue *LLVMToSPIRV::transValueWithoutDecoration(Value *V,
   if (auto GV = dyn_cast<GlobalVariable>(V)) {
     if (GV->getName() == "llvm.global.annotations") {
       transGlobalAnnotation(GV);
-      return nullptr;
```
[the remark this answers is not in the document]
maximova136: Actually, it's not translated twice, as I see. `transGlobalAnnotation` only translates the operands that are referenced in `llvm.global.annotations`.

lib/SPIRV/SPIRVWriter.cpp
```diff
@@ -1663,7 +1658,9 @@ void LLVMToSPIRV::transGlobalAnnotation(GlobalVariable *V) {
 
 bool LLVMToSPIRV::transGlobalVariables() {
   for (auto I = M->global_begin(), E = M->global_end(); I != E; ++I) {
-    if (!transValue(&(*I), nullptr))
+    if ((&(*I))->getName() == "llvm.global.annotations")
```
AGindinson: Would this work as well?
```suggestion
    if (((*I).getName() == "llvm.global.annotations")
```

vmaksimo: As it turns out, yes

PR #358, opened 2019-10-21
Fix translator crash when provided empty file

imashkov:
It's checking file for emptiness before translating

Signed-off-by: Ilya Mashkov <ilya.mashkov@intel.com>

### DIFF
--- a/tools/llvm-spirv/llvm-spirv.cpp
+++ b/tools/llvm-spirv/llvm-spirv.cpp
@@ -177,6 +177,11 @@ static int convertLLVMToSPIRV(const SPIRV::TranslatorOpts &Opts) {
   return 0;
 }
 
+static bool isFileEmpty(const std::string &FileName) {
+  std::ifstream File(FileName);
+  return File && File.peek() == EOF;
+}
+
 static int convertSPIRVToLLVM(const SPIRV::TranslatorOpts &Opts) {
   LLVMContext Context;
   std::ifstream IFS(InputFile, std::ios::binary);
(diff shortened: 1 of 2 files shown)


tools/llvm-spirv/llvm-spirv.cpp
```diff
@@ -177,6 +177,15 @@ static int convertLLVMToSPIRV(const SPIRV::TranslatorOpts &Opts) {
   return 0;
 }
 
+static bool isFileEmpty(const std::string &FileName) {
+  std::ifstream file(FileName);
+  bool Result = false;
+  if (file && file.peek() == EOF)
+     Result = true;
+  file.close();
```
AlexeySachkov: Is it safe to call `close` if file wasn't opened at all? Will it be done automatically in `ifstream` destructor?

imashkov: I deleted this call, thanks.

tools/llvm-spirv/llvm-spirv.cpp
```diff
@@ -177,6 +177,14 @@ static int convertLLVMToSPIRV(const SPIRV::TranslatorOpts &Opts) {
   return 0;
 }
 
+static bool isFileEmpty(const std::string &FileName) {
+  std::ifstream File(FileName);
+  bool Result = false;
+  if (File && File.peek() == EOF)
+    Result = true;
+  return Result;
```
AGindinson: How about:
```
return File && File.peek == EOF;
```

PR #360, opened 2019-10-21
Add support for 'fneg' when translate from LLVM to SPIR-V

### LINKED ISSUE #359: Missing support for 'fneg' LLVM operation
llvm-spirv crashes on the following input:
```
; ModuleID = 'neg.bc'
source_filename = "neg.cl"
target datalayout = "e-i64:64-v16:16-v24:32-v32:32-v48:64-v96:128-v192:256-v256:256-v512:512-v1024:1024"
target triple = "spir64"

; Function Attrs: convergent noinline nounwind optnone
define spir_kernel void @foo(float %a, float addrspace(1)* %b) #0 !kernel_arg_addr_space !4 !kernel_arg_access_qual !5 !kernel_arg_type !6 !kernel_arg_base_type !6 !kernel_arg_type_qual !7 {
entry:
  %a.addr = alloca float, align 4
  %b.addr = alloca float addrspace(1)*, align 8
  store float %a, float* %a.addr, align 4
  store float addrspace(1)* %b, float addrspace(1)** %b.addr, align 8
  %0 = load float, float* %a.addr, align 4
  %neg = fneg float %0
  %1 = load float addrspace(1)*, float addrspace(1)** %b.addr, align 8
  store float %neg, float addrspace(1)* %1, align 4
  ret void
}

attributes #0 = { convergent noinline nounwind optnone "correctly-rounded-divide-sqrt-fp-math"="false" "denorms-are-zero"="false" "disable-tail-calls"="false" "frame-pointer"="none" "less-precise-fpmad"="false" "min-legal-vector-width"="0" "no-infs-fp-math"="false" "no-jump-tables"="false" "no-nans-fp-math"="false" "no-signed-zeros-fp-math"="false" "no-trapping-math"="false" "stack-protector-buffer-size"="8" "uniform-work-group-size"="true" "unsafe-fp-math"="false" "use-soft-float"="false" }

!llvm.module.flags = !{!0}
!opencl.ocl.version = !{!1}
!opencl.spir.version = !{!2}
!llvm.ident = !{!3}

!0 = !{i32 1, !"wchar_size", i32 4}
!1 = !{i32 1, i32 0}
!2 = !{i32 1, i32 2}
!3 = !{!"clang version 10.0.0-svn369720-1~exp1+0~20190823011559.2434~1.gbpf9fff7 (trunk)"}
!4 = !{i32 0, i32 1}
!5 = !{!"none", !"none"}
!6 = !{!"float", !"float*"}
!7 = !{!"", !""}
```

Backtrace:
```
#0  0x00007ffff33fd428 in __GI_raise (sig=sig@entry=6) at ../sysdeps/unix/sysv/linux/raise.c:54
#1  0x00007ffff33ff02a in __GI_abort () at abort.c:89
#2  0x00007ffff33f5bd7 in __assert_fail_base (fmt=<optimized out>, assertion=assertion@entry=0xab57bb "Found && \"Invalid key\"",
    file=file@entry=0xab2300 "SPIRV-LLVM-Translator/lib/SPIRV/libSPIRV/SPIRVUtil.h", line=line@entry=90,
    function=function@entry=0xabb4a0 <SPIRV::SPIRVMap<unsigned int, spv::Op, void>::map(unsigned int)::__PRETTY_FUNCTION__> "static Ty2SPIRV::SPIRVMap<Ty1, Ty2, Identifier>::map(Ty1) [with Ty1 = unsigned int; Ty2 = spv::Op; Identifier = void]") at assert.c:92
#3  0x00007ffff33f5c82 in __GI___assert_fail (assertion=0xab57bb "Found && \"Invalid key\"",
    file=0xab2300 "SPIRV-LLVM-Translator/lib/SPIRV/libSPIRV/SPIRVUtil.h", line=90,
    function=0xabb4a0 <SPIRV::SPIRVMap<unsigned int, spv::Op, void>::map(unsigned int)::__PRETTY_FUNCTION__> "static Ty2 SPIRV::SPIRVMap<Ty1, Ty2, Identifier>::map(Ty1) [with Ty1 = unsigned int; Ty2 = spv::Op; Identifier = void]") at assert.c:101
#4  0x0000000000967420 in SPIRV::SPIRVMap<unsigned int, spv::Op, void>::map (Key=12)
    at SPIRV-LLVM-Translator/lib/SPIRV/libSPIRV/SPIRVUtil.h:90
#5  0x0000000000951fcd in SPIRV::LLVMToSPIRV::transUnaryInst (this=0xda9180, U=0xda4d48, BB=0xdb2eb0)
    at SPIRV-LLVM-Translator/lib/SPIRV/SPIRVWriter.cpp:680
#6  0x0000000000954129 in SPIRV::LLVMToSPIRV::transValueWithoutDecoration (this=0xda9180, V=0xda4d48, BB=0xdb2eb0, CreateForward=false)
    at SPIRV-LLVM-Translator/lib/SPIRV/SPIRVWriter.cpp:971
#7  0x0000000000951b97 in SPIRV::LLVMToSPIRV::transValue (this=0xda9180, V=0xda4d48, BB=0xdb2eb0, CreateForward=false)
    at SPIRV-LLVM-Translator/lib/SPIRV/SPIRVWriter.cpp:636
#8  0x0000000000958afa in SPIRV::LLVMToSPIRV::transFunction (this=0xda9180, I=0xd9c748)
    at SPIRV-LLVM-Translator/lib/SPIRV/SPIRVWriter.cpp:1681
#9  0x000000000095920b in SPIRV::LLVMToSPIRV::translate (this=0xda9180)
    at SPIRV-LLVM-Translator/lib/SPIRV/SPIRVWriter.cpp:1733
#10 0x000000000094d806 in SPIRV::LLVMToSPIRV::runOnModule (this=0xda9180, Mod=...)
    at SPIRV-LLVM-Translator/lib/SPIRV/SPIRVWriter.cpp:115
#11 0x00007ffff491fefa in llvm::legacy::PassManagerImpl::run(llvm::Module&) () from /usr/lib/llvm-9/lib/libLLVM-9.so.1
#12 0x000000000095ba9b in llvm::writeSpirv (M=0xd9a010, Opts=..., OS=..., ErrMsg="")
    at SPIRV-LLVM-Translator/lib/SPIRV/SPIRVWriter.cpp:2087
#13 0x00000000008686a0 in convertLLVMToSPIRV (Opts=...) at SPIRV-LLVM-Translator/tools/llvm-spirv/llvm-spirv.cpp:168
#14 0x0000000000869bce in main (Ac=2, Av=0x7fffffffe8d8) at SPIRV-LLVM-Translator/tools/llvm-spirv/llvm-spirv.cpp:380
```

imashkov:
Fix #359


test/TransFNeg.ll
```diff
@@ -0,0 +1,41 @@
...
+!5 = !{!"none", !"none"}
+!6 = !{!"float", !"float*"}
+!7 = !{!"", !""}
+
+; RUN: llvm-as %s -o %t.bc
+; RUN: llvm-spirv %t.bc -o %t.spv
+; RUN: spirv-val %t.spv
+; RUN: llvm-spirv %t.spv --to-text -o - | FileCheck %s --check-prefix=CHECK-SPIRV
+
+; CHECK-SPIRV: FNegate
```
AlexeySachkov: Could you please move this to the beginning of the file?

imashkov: Done

test/TransFNeg.ll
```diff
@@ -0,0 +1,41 @@
+; RUN: llvm-as %s -o %t.bc
```
StuartDBrady: For LLVM 8.0.0 and onwards, this should be a transcoding test, and `test/OpFNegate.spvasm` should be removed. Recent releases of Clang generate **fneg** for CL source, so this should be implemented using OpenCL C as input, unless there is a good reason not to: I would recommend `DivRem.cl` as an example.

(If this test _does_ have to continue to use LLVM IR as input for any reason, then the IR should be simplified, e.g. there should be no need to use **alloca**. If needed, it is possible that compiling `neg.cl` with `-O1` or `-O2` might provide a better starting point.)

The test should cover a few different types. For example, we should have one test with a single precision float, another with a double precision float, and a third with a vector of half precision floats. (I admit that there is an aspect of "do as I say, not as I do" to this: the SPIR-V assembly tests that I added only covered scalars, with the exception of the `OpFMod` and `OpSMod` tests that were added later. However, any new tests that are added should cover both, unless there is a strong reason not to.)

If LLVM's **fneg** instruction guarantees that **fneg**(-0.0) = +0.0 and **fneg**(+0.0) = -0.0, then this test otherwise looks okay in principle. It is not clear to me whether this change alters the handling of `OpFNegate` in `SPIRVReader`. Could you confirm this, please?

We should ensure we use `UnaryOperator::CreateFNeg()` instead of `BinaryOperator::CreateFNeg()`, either now, or in future versions of the translator. As noted above, recent versions of Clang do generate **fneg**, although there is a comment within LLVM which suggests that the use of LLVM's **fneg** instruction as output may result in optimizations being missed. It is not clear whether this comment is still accurate, but I would tend to assume not, if Clang now generates the instruction.

Note that LLVM 7.0.x does not provide an **fneg** instruction. It would be helpful to mention that it was introduced in LLVM 8.0.0 in the commit message and in the pull request comment. I believe that the logic of subtracting from -0.0 when using LLVM 7.0.x matches the semantics of `OpFNegate`, i.e. for any floating point value _f_, **signbit**(**fneg**(_f_)) ≠ **fneg**(_f_). Can you confirm this, please?

imashkov:
> This should be a transcoding test, written using OpenCL C as input. Testing of vector types and of all three floating point types should be included. The old `OpFNegate` SPIR-V assembly test should be removed. The handling of `OpFNegate` within `SPIRVReader` should be updated to use the unary **fneg** operation instead a "binary" **fneg** (i.e. a subtraction from -0). The commit and pull request messages should provide some clarification as to what is changing, and why, e.g. it is not clear how this affects translation of SPIR-V to LLVM IR.

I  removed old test and changed attached.

test/TransFNeg.ll
StuartDBrady: > Thanks for the suggestion, I've generated reproducer for #359 from OpenCL C using `clang-10`, but I had to manually edit the IR to get `fneg` there.

Are you using a recent version of the Clang development tree? The version I'm using generates **fneg** with no manual modifications needed to the IR. If I use Clang 9, then I find that an **fsub** is generated instead.

> Looking at [lang ref](https://llvm.org/docs/LangRef.html#fneg-instruction) I cannot find exact comments about this, it says the following:
> 
> > The value produced is a copy of the operand with its sign bit flipped.
> 
> So, I guess fneg(-0.0) = +0.0 is true with fneg.

Yes, I came to the same conclusion. It seems intended to correspond to the **negate** operation in IEEE 754-2008/2019 § 5.5.1 _Sign bit operations_, which is defined as follows:

> "**negate**(_x_) copies a floating-point operand _x_ to a destination in the same format, reversing the sign bit. **negate**(_x_) is not the same as **subtraction**(0, _x_)

My understanding is that **negate**(_x_) _is_ the same as **subtraction**(−0, _x_) for rounding modes except **roundTowardNegative**.

> According to SPIR-V spec, `OpFNegate`:
> 
> > Floating-point subtract of Operand from zero.

There have been cases in the past where a SPIR-V operation was intended to align with an LLVM instruction and/or IEEE 754 semantics, and we have sought clarification. It is worth considering whether `OpFNegate` is really intended to be different to the LLVM **fneg** operation (which is the same as the IEEE 754-2008/2019 **negate** operation).

> So, I think that our current implementation with fsub is correct. We might want to update it to use fneg instead but I'm not sure that this is something critical that should be resolved immediately.

If `OpFNegate` and **fneg** are intended to be the same, then it would be appropriate to translate `OpFNegate` to LLVM's **fneg** operation, and to translate LLVM's **fneg** operation to `OpFNegate`. However, it would not be appropriate to translate `OpFNegate` to **fsub** for the **roundTowardNegative** rounding mode.

If `OpFNegate` and **fneg** are intended to be different, then it would not be appropriate to translate **fneg** (with precise IEEE 754-2008 sematics) into `OpFNegate` (with possibly looser semantics).

> Would it be okay with you if we submit an issue to this repo to consider generating `fneg` from `OpFNegate`?

I'm absolutely okay with this being done in stages. It seems completely reasonable to me to add support for translation of **fneg** to `OpFNegate`, as a first step.

If the amount of extra work needed is minimal (as I believe it is) then I think it may be preferable to make all of the necessary changes in this pull request, assuming that the question of the exact semantics of `OpFNegate` is not a barrier.

However, if **fneg** is poorly optimized for older versions of LLVM, and it would simplify backporting to the {{llvm_release_90}} and/or {{llvm_release_80}} branches to separate out any changes relating to translation from SPIR-V to LLVM IR (allowing the second part to be dropped when backporting), then these changes *should* be made separately.

> According to SPIR-V spec, `OpFNegate`:
> 
> > Floating-point subtract of Operand from zero.
> 
> If I understand correctly, this is exact match to LLVM 7.0.x IR

LLVM's `BinaryOperator::CreateFNeg()` creates a subtraction from _negative_ zero, which is correct for all rounding modes except **roundTowardNegative**. Although it would be possible to construct something that works for **roundTowardNegative** with LLVM 7.0.x, it would not be worthwhile, especially as it would result in poorer performing code, and so I think the existing **fsub** would have to suffice.

imashkov:
> Looks okay to me in principle, but the commit message should be updated, e.g.:
> 
> ```
> It adds support for the fneg operation (added in LLVM 8.0.0) when
> translating from LLVM IR to SPIR-V, using the SPIR-V FNegate operation.
> ```

@StuartDBrady  thanks for suggestion, I changed it

PR #363, opened 2019-10-24
Fix loss of loop metadata during reverse translation

mlychkov:
A loop's attribute information may be lost during reverse translation
from SPIR-V in case if a function, which contains the loop, also have
a call to another function that is defined after the function with
the loop.
Fix this loss of loop's metadata.

### DIFF
--- a/lib/SPIRV/SPIRVReader.cpp
+++ b/lib/SPIRV/SPIRVReader.cpp
@@ -725,10 +725,9 @@ void SPIRVToLLVM::transLLVMLoopMetadata(const Function *F) {
         const auto *LCI = static_cast<const SPIRVLoopControlINTEL *>(LMD);
         setLLVMLoopMetadata<SPIRVLoopControlINTEL>(LCI, BI);
       }
-    }
 
-    // Loop metadata map should be re-filled during each function translation.
-    FuncLoopMetadataMap.clear();
+      FuncLoopMetadataMap.erase(LMDItr);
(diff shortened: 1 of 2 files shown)


mlychkov:
@AlexeySachkov Could you please review?

lib/SPIRV/SPIRVReader.h
```diff
@@ -185,7 +185,8 @@ class SPIRVToLLVM {
   // A SPIRV value may be translated to a load instruction of a placeholder
   // global variable. This map records load instruction of these placeholders
   // which are supposed to be replaced by the real values later.
-  typedef std::map<SPIRVValue *, LoadInst *> SPIRVToLLVMPlaceholderMap;
+  typedef std::unordered_map<SPIRVValue *, LoadInst *>
```
AGindinson: IMO, it may be worthwhile to first measure if this really improves the performance.

As far as I understand, this depends on how many load instructions of a global variable we are likely to have during an "average" translation - the hash map may actually be slower at low elements numbers, linear complexity outweighing only at higher elements number. @AlexeySachkov, what are your thoughts on this?

mlychkov: Firstly, I accidentally changed definition for another type. I wanted to change SPIRVToLLVMLoopMetadataMap definition.
Secondly, I thought and agreed with you, better to leave it as is until performance improvement is required.

PR #367, opened 2019-11-01
Fix setting of nsw flag when translating OpSNegate and add translation of FPFastMathMode decorations

StuartDBrady:
This includes tests for translation of FPFastMathMode decorations for OpFAdd, OpFSub, OpFMul, OpFDiv, OpFRem and OpFNegate.

This also adds tests for translation of NoSignedWrap and NoUnsignedWrap decorations for OpIAdd, OpIMul, OpISub and OpShiftLeftLogical, in addition to testing the translation of NoSignedWrap for OpSNegate.

### DIFF
--- a/lib/SPIRV/SPIRVReader.cpp
+++ b/lib/SPIRV/SPIRVReader.cpp
@@ -816,6 +816,17 @@ Value *SPIRVToLLVM::transConvertInst(SPIRVValue *BV, Function *F,
   return ConstantExpr::getCast(CO, dyn_cast<Constant>(Src), Dst);
 }
 
+static void applyNoIntegerWrapDecorations(const SPIRVValue *BV,
+                                          Instruction *Inst) {
+  if (BV->hasDecorate(DecorationNoSignedWrap)) {
+    Inst->setHasNoSignedWrap(true);
+  }
+
+  if (BV->hasDecorate(DecorationNoUnsignedWrap)) {
+    Inst->setHasNoUnsignedWrap(true);
+  }
+}
+
 BinaryOperator *SPIRVToLLVM::transShiftLogicalBitwiseInst(SPIRVValue *BV,
                                                           BasicBlock *BB,
                                                           Function *F) {
@@ -829,15 +840,7 @@ BinaryOperator *SPIRVToLLVM::transShiftLogicalBitwiseInst(SPIRVValue *BV,
   auto Inst = BinaryOperator::Create(BO, transValue(BBN->getOperand(0), F, BB),
                                      transValue(BBN->getOperand(1), F, BB),
                                      BV->getName(), BB);
-
-  if (BV->hasDecorate(DecorationNoSignedWrap)) {
-    Inst->setHasNoSignedWrap(true);
-  }
-
-  if (BV->hasDecorate(DecorationNoUnsignedWrap)) {
-    Inst->setHasNoUnsignedWrap(true);
-  }
-
+  applyNoIntegerWrapDecorations(BV, Inst);
   return Inst;
 }
 
@@ -1869,9 +1872,10 @@ Value *SPIRVToLLVM::transValueWithoutDecoration(SPIRVValue *BV, Function *F,
 
   case OpSNegate: {
     SPIRVUnary *BC = static_cast<SPIRVUnary *>(BV);
-    return mapValue(
-        BV, BinaryOperator::CreateNSWNeg(transValue(BC->getOperand(0), F, BB),
-                                         BV->getName(), BB));
+    auto Neg = BinaryOperator::CreateNeg(transValue(BC->getOperand(0), F, BB),
+                                         BV->getName(), BB);
+    applyNoIntegerWrapDecorations(BV, Neg);
+    return mapValue(BV, Neg);
   }
 
   case OpFMod: {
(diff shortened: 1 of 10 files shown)


lib/SPIRV/SPIRVReader.cpp
```diff
@@ -829,15 +861,8 @@ BinaryOperator *SPIRVToLLVM::transShiftLogicalBitwiseInst(SPIRVValue *BV,
...
-  if (BV->hasDecorate(DecorationNoSignedWrap)) {
-    Inst->setHasNoSignedWrap(true);
-  }
-
-  if (BV->hasDecorate(DecorationNoUnsignedWrap)) {
-    Inst->setHasNoUnsignedWrap(true);
-  }
-
+  applyNoIntegerWrapDecorations(BV, Inst);
+  applyFPFastMathModeDecorations(BV, Inst);
```
AlexeySachkov: I guess this line can be removed, since we cannot perform logical bitwise shift on floating-point types, can we?

StuartDBrady: No, as this is where translation of binary floating-point arithmetic instructions is handled.

Perhaps the functions should be renamed, and perhaps isBinaryShiftLogicalBitwiseOpCode and isLogicalOpCode should be rewritten to simply list all of the applicable opcodes, in place of the confusing ranges that they currently use.

It would also make sense to split transValueWithoutDecoration() up.  It's 644 lines long.  Something 10-20 lines long that calls helper functions would be better.

AlexeySachkov: Ok, thanks for explanation

lib/SPIRV/SPIRVReader.cpp
```diff
@@ -816,6 +816,36 @@ Value *SPIRVToLLVM::transConvertInst(SPIRVValue *BV, Function *F,
...
+  if (BV->hasDecorate(DecorationFPFastMathMode, 0, &V)) {
+    if (V & FPFastMathModeNotNaNMask)
+      FMF.setNoNaNs();
+    if (V & FPFastMathModeNotInfMask)
+      FMF.setNoInfs();
+    if (V & FPFastMathModeNSZMask)
+      FMF.setNoSignedZeros();
+    if (V & FPFastMathModeAllowRecipMask)
+      FMF.setAllowReciprocal();
+    if (V & FPFastMathModeFastMask)
```
vzakhari, line 843: Can someone please explain why `SPIR-V` `Fast` implies `LLVM` `fast`?  `SPIR-V` `Fast` [means](https://www.khronos.org/registry/spir-v/specs/unified1/SPIRV.html#FP_Fast_Math_Mode) `NotNaN|NotInf|NSZ|AllowRecip`, while LLVM 'fast' [means](https://llvm.org/docs/LangRef.html#fast-math-flags) `nnan|ninf|nsz|arcp|`**contract|afn|reassoc**.

AlexeySachkov: Tagging @StuartDBrady here for visibility

StuartDBrady: > I would assume that somewhere in SPIR-V Writer we would map LLVM `fast` into SPIR-V `fast`, so, in order to preserve LLVM IR -> SPIR-V -> LLVM IR roundtrip we use this particular mapping here.

Yes, the round-trip mapping is the reason.

For SPIR-V which _specifically_ requires the `NotNaN`, `NotInf`, `NSZ` and `AllowRecip` flags and no others, `NotNaN|NotInf|NSZ|AllowRecip` should be used. The entire point of having a `Fast` flag is to allow extra flags to be added to it.

vzakhari: Then I am concerned about the documentation of `SPIR-V` `Fast`, which says:

> Allow algebraic transformations according to real-number associative and distributive algebra. This flag implies all the others.

Are you saying that `all the others` means `any possible` FP optimization flags existing in the particular `SPIR-V` consumer implementation?  It does not sound very robust for the spec.  I also do not see any comments explaining that `SPIR-V` producers complying with the spec must not use `Fast` as a replacement of `NotNaN|NotInf|NSZ|AllowRecip`.  Did I miss it?

grey-eminence: My understading is:
`all the others` == `NotNaN|NotInf|NSZ|AllowRecip` 
`associative and distributive algebra` == `contract|afn|reassoc`

vzakhari: I agree that "associative ... algebra" includes `reassoc`, but I think `contract` and `afn` are not covered by this.  What is the right way to get an explanation of the SPIR-V specification, if I have doubts?

AlexeySachkov: >  What is the right way to get an explanation of the SPIR-V specification, if I have doubts?

You could ask for the clarification: I'm not aware of the open-source repo or bug tracker related to core SPIR-V spec, but you can try creating issues in https://github.com/KhronosGroup/SPIRV-Headers or https://github.com/KhronosGroup/SPIRV-Registry. And for sure the best way is probably to discuss this in internal KhronosGroup repos corresponding to SPIR-V spec

PR #370, opened 2019-11-05
Enable loop simplify pass for all llvm.loop attributes

vmaksimo:
It is done to avoid duplicate OpLoopMerge instruction in case of continue
operand in the loop.


lib/SPIRV/SPIRVInternal.h
```diff
@@ -930,7 +930,7 @@ template <> inline void SPIRVMap<std::string, Op, SPIRVOpaqueType>::init() {
 }
 
 // Check if the module contains llvm.loop.unroll.* metadata
```
svenvh:
```suggestion
// Check if the module contains llvm.loop.* metadata
```

lib/SPIRV/SPIRVUtil.cpp
```diff
@@ -1501,16 +1501,13 @@ llvm::Constant *getOCLNullClkEventPtr(Module *M) {
...
   for (const Function &F : *M)
     for (const BasicBlock &BB : F) {
       const Instruction *Term = BB.getTerminator();
       if (!Term)
         continue;
       if (const MDNode *MD = Term->getMetadata("llvm.loop"))
-        for (const MDOperand &MDOp : MD->operands())
-          if (getMDOperandAsString(dyn_cast<MDNode>(MDOp), 0)
-                  .find("llvm.loop.unroll.") == 0)
             return true;
```
mlychkov: Please align indent.

test/transcoding/FPGAInfiniteLoop.ll
```diff
@@ -0,0 +1,178 @@
+; RUN: llvm-as %s -o %t.bc
```
AGindinson: I'd suggest commenting out the source code and providing the link to the intel/llvm project, which apparently was used to produce the LLVM IR (the same way it was done for test/IntelFPGAReg.ll)

vmaksimo:
> `while (true)`

Where is the difference? `while (i < m)` will always be true

mlychkov:
> > `while (true)`
> 
> Where is the difference? `while (i < m)` will always be true

Looks like compiler doesn't detect that this is infinite loop.

lib/SPIRV/SPIRVWriter.cpp
```diff
@@ -2154,7 +2154,7 @@ bool llvm::writeSpirv(Module *M, const SPIRV::TranslatorOpts &Opts,
 
   legacy::PassManager PassMgr;
   addPassesForSPIRV(PassMgr);
-  if (hasLoopUnrollMetadata(M))
+  if (hasLoopMetadata(M))
```
mlychkov: [Optional] It might be good to have a comment why we need loopsimplify pass here.

AlexeySotkin: Indeed. @vmaksimo please add a comment explaining why we need to simplify loops.

PR #375, opened 2019-11-09
Add support for reading OpTranspose

jigsawecho:
Signed-off-by: Qinglai Xiao <q.xiao@think-silicon.com>


jigsawecho:
@AlexeySachkov 3x3 is slowpath test case. I am adding another test case 3x4 which is also slowpath.

PR #381, opened 2019-11-20
Translation of SPV_INTEL_media_block_io from SPIRV to LLVM

aratajew:


aratajew:
@AlexeySotkin could you please take a look? Corresponding changes has already been committed to llvm_release_90 branch: https://github.com/KhronosGroup/SPIRV-LLVM-Translator/commit/27a33e74d3e43f14fd7227f57c4b3a8670b1ec36

Here is the same for llvm_release_80 branch: https://github.com/KhronosGroup/SPIRV-LLVM-Translator/pull/386

Thanks

lib/SPIRV/SPIRVReader.cpp
```diff
@@ -571,6 +571,19 @@ bool SPIRVToLLVM::isSPIRVCmpInstTransToLLVMInst(SPIRVInstruction *BI) const {
...
+  // Not every spirv opcode which is placed in OCLSPIRVBuiltinMap is
+  // translated directly to OCL builtin. Some of them are translated
+  // to LLVM representation without any modifications (SPIRV format of
+  // instruction is represented in LLVM) and then its translated to
+  // clang-consistent format in SPIRVToOCL pass.
+  return (OCLSPIRVBuiltinMap::rfind(OpCode, nullptr) &&
+          !isAtomicOpCode(OpCode) && !isGroupOpCode(OpCode) &&
+          !isPipeOpCode(OpCode) && !isMediaBlockINTELOpcode(OpCode)) ||
+         isSubgroupAvcINTELInstructionOpCode(OpCode) ||
+         isIntelSubgroupOpCode(OpCode);
```
AlexeySachkov: Suggest to simplify this a bit:

```c++
  if (isSubGroupAVCINTELInstructionOpCode(OpCode) || isIntelSubGroupOpCpde(OpCode))
    return true;
  if (OCLSPIRVBuiltinMap::rfind(OpCode, nullptr) {
    // everything except atomics, groups, pipes and media_block_io_intel is directly translated
    return !(isAtomicOpCode(OpCode) || isGroupOpCode(OpCode) || isPipeOpCode(OpCode) || isMediaBlockINTELOpCode(OpCode));
  }
  return false;
```

lib/SPIRV/SPIRVToOCL.cpp
```diff
@@ -319,6 +323,41 @@ void SPIRVToOCL::visitCallSPIRVPipeBuiltin(CallInst *CI, Op OC) {
...
 }
 
+void SPIRVToOCL::visitCallSPIRVImageMediaBlockBuiltin(CallInst *CI, Op OC) {
+  AttributeList Attrs = CI->getCalledFunction()->getAttributes();
+  mutateCallInstOCL(
+      M, CI,
+      [=](CallInst *, std::vector<Value *> &Args) {
+        // Moving the first argument to the end.
+        std::rotate(Args.rbegin(), Args.rend() - 1, Args.rend());
+        Type *FuncType = CI->getType();
```
AlexeySachkov:
```suggestion
        Type *RetType = CI->getType();
```

`FuncType` confuses me with `FunctionType` which is a bit different thing. I got quite a strange first impression trying to understand why are you expecting `FunctionType` to be equal to `VectorType`

lib/SPIRV/libSPIRV/SPIRVOpCode.h
```diff
@@ -135,6 +135,12 @@ inline bool isGroupOpCode(Op OpCode) {
   return OpGroupAll <= OC && OC <= OpGroupSMax;
 }
 
+inline bool isMediaBlockINTELOpcode(Op OpCode) {
```
AlexeySachkov:
```suggestion
inline bool isMediaBlockINTELOpcode(Op OC) {
```

test/SPV_INTEL_media_block_io.ll
```diff
@@ -11,18 +11,22 @@
 ;}
 
 ; RUN: llvm-as < %s -o %t.bc
-; RUN: llvm-spirv %t.bc -spirv-text -o - | FileCheck %s
+; RUN: llvm-spirv %t.bc -spirv-text -o %t.txt
+; RUN: FileCheck < %t.txt %s --check-prefix=CHECK-SPIRV
+; RUN: llvm-spirv %t.bc -o %t.spv
+; RUN: llvm-spirv -r %t.spv -o %t.rev.bc
```
AlexeySachkov: Since this test checks translation in both ways, it should be moved into `transcoding` sub-dir

lib/SPIRV/SPIRVToOCL.cpp
```diff
@@ -319,6 +323,41 @@ void SPIRVToOCL::visitCallSPIRVPipeBuiltin(CallInst *CI, Op OC) {
...
+        if (OC == OpSubgroupImageMediaBlockWriteINTEL) {
+          assert(Args.size() >= 4 && "Wrong media block write signature");
+          FuncType = Args.at(3)->getType(); // texel type
+        }
+        unsigned int BitWidth = FuncType->getScalarSizeInBits();
+        std::string FuncPostfix;
+        if (BitWidth == 8)
+          FuncPostfix = "_uc";
+        else if (BitWidth == 16)
+          FuncPostfix = "_us";
```
AlexeySachkov: This suffix is untested

lib/SPIRV/SPIRVToOCL.cpp
```diff
@@ -319,6 +323,41 @@ void SPIRVToOCL::visitCallSPIRVPipeBuiltin(CallInst *CI, Op OC) {
...
+        else if (BitWidth == 32)
+          FuncPostfix = "_ui";
+        else
+          assert(0 && "Unsupported texel type!");
+
+        if (FuncType->isVectorTy()) {
+          unsigned int NumEl = FuncType->getVectorNumElements();
```
AlexeySachkov: Please add a test for this piece of code: right now you don't have a test returning vector

aratajew:
Thanks for the great comments @AlexeySachkov , but unfortunately this commit has already been committed to llvm_release_90 branch (27a33e74). Do we want to introduce history inconsistency between those branches? Or should I prepare a seperate commit with fixes and cherry-pick it to llvm_release_90 branch?

AlexeySachkov:
> but unfortunately this commit has already been committed to llvm_release_90 branch (27a33e7). Do we want to introduce history inconsistency between those branches? 

I'm not sure that history consistency is very important here (however, it might simplify investigating diff between branches)

For example, [UserSemantic](https://github.com/KhronosGroup/SPIRV-LLVM-Translator/commit/d9f33c9f6cff84d0dc885233b2ea75f410d4ee6d) decoration on `llvm_release_80` branch is implemented via different amount/order of commits comparing to `master` branch and I don't see any problems there.

> Or should I prepare a seperate commit with fixes and cherry-pick it to llvm_release_90 branch?

I, personally, don't think that it is necessary to commit changes related to my comments to `llvm_release_90`, since all of them are not about functionality, but about code style.

svenvh:
> > but unfortunately this commit has already been committed to llvm_release_90 branch ([27a33e7](https://github.com/KhronosGroup/SPIRV-LLVM-Translator/commit/27a33e74d3e43f14fd7227f57c4b3a8670b1ec36)). Do we want to introduce history inconsistency between those branches?
> 
> I'm not sure that history consistency is very important here (however, it might simplify investigating diff between branches)

Yeah exactly for that reason I have a slight preference to keep branches in sync as much as possible.

AlexeySotkin:
@aratajew do you plan to address comments from @AlexeySachkov?
Regarding inconsistency between branches I think it is ok have a difference in this case. If we want to simplify investigations, the comments can be addressed with another patch, so we can filter out the difference.

aratajew:
Sure, I do, but I have a lot of higher priority work to do now. I should find some time to fix this next week.

lib/SPIRV/SPIRVReader.cpp
aratajew: Done

lib/SPIRV/SPIRVToOCL.cpp
aratajew: Done

lib/SPIRV/libSPIRV/SPIRVOpCode.h
aratajew: I've left "OpCode" parameter name (to keep consistent with entire file naming convention), but removed assigning it to the local variable OC.

test/SPV_INTEL_media_block_io.ll
aratajew: I've added two-way translation tests for all cases from specification: https://www.khronos.org/registry/OpenCL/extensions/intel/cl_intel_media_block_io.txt

lib/SPIRV/SPIRVToOCL.cpp
aratajew: Test added.

lib/SPIRV/SPIRVToOCL.cpp
aratajew: Done

test/transcoding/SPV_INTEL_media_block_io.cl
```diff
@@ -0,0 +1,214 @@
+// RUN: %clang_cc1 -triple spir-unknown-unknown -O1 -cl-std=CL2.0 -finclude-default-header -emit-llvm-bc %s -o %t.bc
+// RUN: llvm-spirv %t.bc --spirv-ext=+SPV_INTEL_media_block_io -spirv-text -o - | FileCheck %s --check-prefix=CHECK-SPIRV
```
AlexeySachkov: I suggest generate text representation from the binary one - in this case, you don't need to pass `--spirv-ext` option since the translator doesn't check for allowed/disallowed extensions while translating to/from text representation.

```
// RUN: llvm-spirv %t.spv -to-text -o - | FileCheck %s --check-prefix=CHECK-SPIRV
```

test/transcoding/SPV_INTEL_media_block_io.cl
```diff
@@ -0,0 +1,214 @@
...
+// RUN: llvm-spirv %t.bc --spirv-ext=+SPV_INTEL_media_block_io -o %t.spv
+// RUN: spirv-val %t.spv
+// RUN: llvm-spirv -r --spirv-ext=+SPV_INTEL_media_block_io %t.spv -o %t.rev.bc
```
AlexeySachkov:
```suggestion
// RUN: llvm-spirv -r %t.spv -o %t.rev.bc
```

All extensions are considered as "allowed" during consumption step

test/transcoding/SPV_INTEL_media_block_io.cl
aratajew: Done

test/transcoding/SPV_INTEL_media_block_io.cl
aratajew: Done. Thanks for the suggesiton.

PR #382, opened 2019-11-20
Translation of SPV_INTEL_media_block_io from SPIRV to LLVM

aratajew:


lib/SPIRV/SPIRVReader.cpp
```diff
@@ -1855,7 +1855,8 @@ Value *SPIRVToLLVM::transValueWithoutDecoration(SPIRVValue *BV, Function *F,
     } else if ((OCLSPIRVBuiltinMap::rfind(OC, nullptr) ||
                 isSubgroupAvcINTELInstructionOpCode(OC) ||
                 isIntelSubgroupOpCode(OC)) &&
-               !isAtomicOpCode(OC) && !isGroupOpCode(OC) && !isPipeOpCode(OC)) {
+               !isAtomicOpCode(OC) && !isGroupOpCode(OC) && !isPipeOpCode(OC) &&
+               !isMediaBlockINTELOpcode(OC)) {
```
AlexeySotkin: Should we factor it out to some helper function?

aratajew: Sure, I'll try to factor it out later this day.

aratajew: Done

AlexeySotkin: Please do the same in #381

PR #383, opened 2019-11-22
Support using an external LLVM tree when embedding in another project

kpet:
When embedding the translator in another CMake project the existing
automatic detection for so-called external builds no longer works.

This change adds an LLVM_SPIRV_BUILD_EXTERNAL variable that users
can set to force the decision. If the variable isn't set, the
behaviour is detected as previously.

Signed-off-by: Kevin Petit <kevin.petit@arm.com>


kpet:
I've just rebased this on top of the build fixes, now passing the tests. Any takers to review 30 lines of CMake/README changes?

PR #384, opened 2019-11-22
Translate specialization constants from SPIRV

AlexeySotkin:
Add support for OpSpecConstantTrue, OpSpecConstantFalse, OpSpecConstant and
OpSpecConstantComposite instructions.

Allow to provide values to be used instead of the default values of
specialization constants while translating SPIRV to LLVM.


AlexeySotkin:
@Naghasan,  @aratajew, FIY.

include/LLVMSPIRVOpts.h
```diff
@@ -99,12 +102,63 @@ class TranslatorOpts {
...
+  // CL_INVALID_VALUE erros?
+  bool addSpecConst(unsigned SpecId, size_t Size /*in bytes*/,
+                    const void *Value) {
+    if (std::find(ExternalSpecializationIds.begin(),
+                  ExternalSpecializationIds.end(),
+                  SpecId) != ExternalSpecializationIds.end()) {
+      return true;
+    }
+    ExternalSpecializationIds.push_back(SpecId);
+    ExternalSpecializationValues.emplace_back(static_cast<const char *>(Value),
```
AlexeySachkov: In `checkSpecializationConstant` you assume that `Size` is always 8, 16, 32 or 64, but you don't have any checks if this is not the case. I suggest to at least put some assertions here or into `switch (Buffer.size()) {` from `checkSpecializationConstant`

AlexeySachkov: Found the assert within SPIRVReader.cpp, but I still think that it is needed somewhere in this header as well, because users might initialize spec constants via `TranslatorOpts` in their apps

lib/SPIRV/libSPIRV/SPIRVStream.cpp
```diff
@@ -256,6 +256,12 @@ SPIRVEntry *SPIRVDecoder::getEntry() {
     }
   }
 
+  if (!M.getErrorLog().checkError(Entry->isImplemented(),
+                                  SPIRVEC_UnimplementedOpCode,
+                                  std::to_string(Entry->getOpCode()))) {
+    M.setInvalid();
+  }
+
```
AlexeySachkov: Do you have tests for this code?

tools/llvm-spirv/llvm-spirv.cpp
```diff
@@ -371,6 +458,10 @@ int main(int Ac, char **Av) {
   SPIRV::TranslatorOpts Opts(MaxSPIRVVersion, ExtensionsStatus,
                              SPIRVGenKernelArgNameMD);
 
+  if (parseSpecConstOpt(SpecConst, Opts)) {
+    return -1;
+  }
+
```
AlexeySachkov: Please add a few negative tests for this function/command line option as well

include/LLVMSPIRVOpts.h
```diff
@@ -99,12 +102,63 @@ class TranslatorOpts {
 
   void enableGenArgNameMD() { GenKernelArgNameMD = true; }
 
+  // Stores values for specialization constants. Potentially, can be used for
+  // implementation of clSetProgramSpecializationConstant API.
```
aratajew: clSetProgramSpecializationConstant API is seperate interface from clBuildProgram. 
Lets imagine potential controlflow and decide whether such function is fully reasonable in this case.
User calls clSetProgramSpecializationConstant from host code, then Runtime needs to verify whether specified specialization constant is valid with specialization constant within the kernel (lets remember that clBuildProgram does not have to be called yet, so we have no kernel compiled). At this moment Runtime/Compiler needs to parse and only parse the kernel(it does not have to translate it) to validate whether specialization constant (passed as an arguments to API call) is correct. 
In my opinion, we should implement the function parsing SPIRV module and returning specialization constants within it. It would be more helpful for OpenCL Runtimes as they could just call this function and use the result to validate user-specified specialization constant.

Now, lets imagine that OpenCL Runtime calls clBuildProgram. In my opinion your function (addSpecConst) is reasonable in this case only, as Runtime/Compiler needs to update previously gathered (during clSetProgramSpecializationConstant) specialization constants.

Please correct me if my way of thinking is unreasonable. Thanks.

AlexeySotkin: I think your reasoning makes sense and it is related to my TODO a couple lines below.
It seems we do need an API to check spec constants passed to clSetProgramSpecializationConstant against the SPIR-V module. I'll update the patch.

include/LLVMSPIRVOpts.h
```diff
@@ -99,12 +102,63 @@ class TranslatorOpts {
...
+    ExternalSpecializationIds.push_back(SpecId);
+    ExternalSpecializationValues.emplace_back(static_cast<const char *>(Value),
+                                              Size);
+    return false;
+  }
+
+  // Checks if a specialization constant with \p SpecId has been externally
+  // specialized. If so, writes externally provided value to \p Value.
+  // Returns true on success.
+  bool checkSpecializationConstant(uint32_t SpecId, uint64_t &Value) const {
```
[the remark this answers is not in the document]
AlexeySotkin: Sure

include/LLVMSPIRVOpts.h
```diff
@@ -99,12 +102,63 @@ class TranslatorOpts {
...
+    case sizeof(uint16_t):
+      Value = *reinterpret_cast<const uint16_t *>(Buffer.data());
+      break;
+    case sizeof(uint32_t):
+      Value = *reinterpret_cast<const uint32_t *>(Buffer.data());
+      break;
+    case sizeof(uint64_t):
+      Value = *reinterpret_cast<const uint64_t *>(Buffer.data());
+      break;
+    }
```
AlexeySotkin: default case is missing

tools/llvm-spirv/llvm-spirv.cpp
```diff
@@ -349,6 +354,88 @@ static int parseSPVExtOption(
...
 }
 
+bool parseSpecConstOpt(const std::string &SpecConstStr,
+                       SPIRV::TranslatorOpts &Opts) {
+  llvm::StringRef SpecConst(SpecConstStr);
+  SmallVector<StringRef, 8> Split;
+  SpecConst.split(Split, ' ', -1, false);
+  for (StringRef Option : Split) {
+    SmallVector<StringRef, 4> Params;
+    Option.split(Params, ':', 3);
```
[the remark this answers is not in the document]
AlexeySotkin: Agree

lib/SPIRV/SPIRVReader.cpp
```diff
@@ -1327,15 +1327,25 @@ Value *SPIRVToLLVM::transValueWithoutDecoration(SPIRVValue *BV, Function *F,
...
+  case OpConstant:
+  case OpSpecConstant: {
     SPIRVConstant *BConst = static_cast<SPIRVConstant *>(BV);
     SPIRVType *BT = BV->getType();
     Type *LT = transType(BT);
+    uint64_t ConstValue = BConst->getZExtIntValue();
+    SPIRVWord SpecId = 0;
+    if (OC == OpSpecConstant && BV->hasDecorate(DecorationSpecId, 0, &SpecId)) {
+      BM->getSpecializationConstant(SpecId, ConstValue);
+      assert(
```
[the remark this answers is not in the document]
AlexeySotkin: No, `ConstValue` can be updated by `getSpecializationConstant` call.
But I'd rather should put this assert under `if`.

lib/SPIRV/SPIRVReader.cpp
```diff
@@ -3622,3 +3645,48 @@ bool llvm::readSpirv(LLVMContext &C, const SPIRV::TranslatorOpts &Opts,
...
+  // all constant instructions must appear before function declarations.
+  while (D.OpCode != OpFunction && D.getWordCountAndOpCode()) {
+    switch (D.OpCode) {
+    case OpDecorate:
+      // The decoration is added to the module in scope of SPIRVDecorate::decode
+      D.getEntry();
+      break;
+    case OpTypeBool:
+    case OpTypeInt:
+    case OpTypeFloat:
```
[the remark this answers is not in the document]
AlexeySotkin: The `SpecId` decoration can be applied only to literals. The user can't specialize composite constants directly. Instead the user should/can specialize some or all elements of a constant composite type.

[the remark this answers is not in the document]
AlexeySotkin: Note: `OpSpecConstantOp` is out of scope for this PR.

Issue #281, opened 2019-07-30
Incorrect translation of Function Control


mantognini:
The mapping between function attributes in LLVM IR to SPIR-V Function Control seems to be incorrect.

Specifically, currently `readonly` is mapped to `Const` and `readnone` is mapped to `Pure`.

For reference, here are the relevant [SPIR-V definitions](https://www.khronos.org/registry/spir-v/specs/unified1/SPIRV.html#Function_Control):

> ### `Pure`
>
> Compiler can assume this function has no side effect, but might read global memory or read through dereferenced function parameters. Always computes the same result for the same argument values.
>
> ### `Const`
> Compiler can assume this function has no side effects, and will not access global memory or dereference function parameters. Always computes the same result for the same argument values.

And the [LLVM IR definitions](https://llvm.org/docs/LangRef.html#function-attributes):

> ### `readonly`
>
> On a function, this attribute indicates that the function does not write through any pointer arguments [...] or otherwise modify any state [...] visible to caller functions. It may dereference pointer arguments and read state that may be set in the caller. A readonly function always returns the same value[...] when called with the same set of arguments and global state. [...]
>
> ### `readnone`
>
> On a function, this attribute indicates that the function computes its result [...] based strictly on its arguments, without dereferencing any pointer arguments or otherwise accessing any mutable state [...] visible to caller functions. It does not write through any pointer arguments [...] and never changes any state visible to callers. [...]

I suppose the correct mapping would be:
+ `readonly` <=> `Pure`
+ `readnone` <=> `Const`

AlexeySotkin:
@mantognini, thank you for reporting the issue. It should be easy to fix here.
https://github.com/KhronosGroup/SPIRV-LLVM-Translator/blob/2f667b52e89201cd4e1ae84ccf9e03da52946b39/lib/SPIRV/SPIRVInternal.h#L235-L242
Also a regression test is needed.

Hopetech:
@mantognini Out of curiosity, how did you catch that issue? By reading the code?

mantognini:
@Hopetech I was writing tests for https://github.com/KhronosGroup/LLVM-SPIRV-Backend and was using the translator as some kind of reference implementation and a test showed a disagreement between the two implementations about this.

MrSidims:
Fixed by https://github.com/KhronosGroup/SPIRV-LLVM-Translator/pull/1697

PR #387, opened 2019-12-01
fixed issue 281

hexamage:
https://github.com/KhronosGroup/SPIRV-LLVM-Translator/issues/281


StuartDBrady:
> Could you please rewrite commit message in accordance with some common rules (i.e. https://llvm.org/docs/DeveloperPolicy.html#commit-messages)?

Could the same also be done for the PR title and description, please?

(In general, when a PR consists of a single commit, my view is that it is helpful if the title and description of the PR match those of the commit as closely as possible. Although this duplicates effort on the part of the owner of the PR, it reduces effort for those searching through the history of PRs, which is usually the more important consideration.)

StuartDBrady:
CLA was not accepted.

PR #389, opened 2019-12-05
Use correct intrinsic overloading for Intel FPGA decorations applied to fields

vmaksimo:
llvm.ptr.annotation intrinsic is overloaded so it can be used with the integer of any width (https://llvm.org/docs/LangRef.html#llvm-ptr-annotation-intrinsic). However, the previous approach used bitcasting to int8 for any integer type.
Now the correct intrinsic overload is used for integer types.


test/annotate_attribute.ll
```diff
@@ -95,26 +95,26 @@ entry:
   %0 = bitcast %struct.bar* %s1 to i8*
   call void @llvm.lifetime.start.p0i8(i64 8, i8* %0) #4
   ; CHECK-LLVM: %[[FIELD1:.*]] = getelementptr inbounds %struct.bar, %struct.bar* %{{[a-zA-Z0-9]+}}, i32 0, i32 0
-  ; CHECK-LLVM: %[[CAST1:.*]] = bitcast{{.*}}%[[FIELD1]]
-  ; CHECK-LLVM: call i8* @llvm.ptr.annotation.p0i8{{.*}}%[[CAST1]]{{.*}}[[STR4]]
```
AlexeySachkov: Could you please add one more test case into this file which requires `bitcast` insertion?

PR #391, opened 2019-12-09
Fix OpAtomicUMin and OpAtomicUMax translation

aratajew:


aratajew:
@AlexeySachkov, @AlexeySotkin guys, could we merge those changes please? Would be great to also merge it to llvm_release_80 branch: https://github.com/KhronosGroup/SPIRV-LLVM-Translator/pull/392

PR #393, opened 2019-12-09
Fix unused variable warning

svenvh:
Fix unused variable warning after bc9fa54 ("Enable loop simplify pass
for all llvm.loop attributes", 2019-11-05).


lib/SPIRV/SPIRVUtil.cpp
```diff
@@ -1507,7 +1507,7 @@ bool hasLoopMetadata(const Module *M) {
       const Instruction *Term = BB.getTerminator();
       if (!Term)
         continue;
-      if (const MDNode *MD = Term->getMetadata("llvm.loop"))
+      if (Term->getMetadata("llvm.loop"))
```
mlychkov: Maybe to rewrite this as:
`
const Instruction *Term = BB.getTerminator();
if (Term && Term->getMetadata("llvm.loop"))
  return true;
`

PR #397, opened 2019-12-18
Implement SPV_INTEL_fpga_function_attributes extension

MrSidims:
Spec can be found in: https://github.com/KhronosGroup/SPIRV-Registry/pull/55

The extension introduces 4 more execution modes:
  - MaxWorkgroupSizeINTEL:
  Indicates the maximum possible work-group size in the x, y, and z
  dimensions. Only valid with the Kernel Execution Model.
  - MaxWorkDimINTEL:
  Indicates the maximum number of work dimensions. Legal values range
  from 0 to 3. A maximum dimensionality of 0 indicates that the
  kernel can only be launched with a single work-item. Only valid
  with the Kernel Execution Model.
  - NoGlobalOffsetINTEL (out of scope of these patch):
  Indicates that the global offset is always (0, 0, 0). Only valid
  with the Kernel Execution Model.
  - NumSIMDWorkitemsINTEL:
  Indicates that work-items should be vectorized with the provided
  vector width. Only valid with the Kernel Execution Model.

  The execution modes are needed to handle the appropriate SYCL kernel
  attributes.


MrSidims:
@AlexeySotkin could you please take a look?

test/IntelFPGAFunctionAttributes.ll
```diff
@@ -0,0 +1,100 @@
...
+; RUN: llvm-as %s -o %t.bc
+; RUN: llvm-spirv %t.bc --spirv-ext=+SPV_INTEL_kernel_attributes -o %t.spv
+; RUN: llvm-spirv %t.spv -to-text -o %t.spt
+; RUN: FileCheck < %t.spt %s --check-prefix=CHECK-SPIRV
+
+; RUN: llvm-spirv -r %t.spv -o %t.rev.bc
+; RUN: llvm-dis < %t.rev.bc | FileCheck %s --check-prefix=CHECK-LLVM
+
+; RUN: llvm-spirv -spirv-text -r %t.spt -o %t.rev.bc
+; RUN: llvm-dis < %t.rev.bc | FileCheck %s --check-prefix=CHECK-LLVM
```
AlexeySachkov: Why do you need this?

Text representation should only be used for debug purposes, so, I don't see huge value in testing that we can correctly convert it back to LLVM IR

MrSidims: +1 to Ronan's comment.

Recently I was experimenting with AP integers handling in the translator (BTW, why do they arbitrary precision integers and not arbitrary width integers?). So in theory the translator shall work with with their translation, but in practice I failed to emit both SPIR-V and LLVM IR texts for my synthetic code.

MrSidims:
> For the changes in spirv.hpp, is there any plan to upstream these into the official khronos header? Further diverging the downstream header just makes building compatible products more challenging.
> It might be better in any case to produce a SPIRV-LLVM Translator extension file that contains these additional enumerations, until they can be properly added to the spec in either case as having a copy of the standard header that we have heavily modified just adds confusion and compatibility issues.

I'm not sure if there are any plans, I'm totally down with the upstreaming.

lib/SPIRV/SPIRVReader.cpp
```diff
@@ -3147,6 +3147,21 @@ bool SPIRVToLLVM::transKernelMetadata() {
       auto SizeMD = ConstantAsMetadata::get(getUInt32(M, EM->getLiterals()[0]));
       F->setMetadata(kSPIR2MD::SubgroupSize, MDNode::get(*Context, SizeMD));
     }
+    // Generate metadata for intel_max_work_group_size
```
AlexeySotkin: These metadata don't have `intel_` prefix in the test

MrSidims: No, fixed.

PR #399, opened 2019-12-19
Translate Intel FPGA bank_bits memory attribute

vmaksimo:

### DIFF
--- a/lib/SPIRV/SPIRVReader.cpp
+++ b/lib/SPIRV/SPIRVReader.cpp
@@ -2808,6 +2808,14 @@ void generateIntelFPGAAnnotation(const SPIRVEntry *E,
       Out << ":" << Str;
     Out << '}';
   }
+  if (E->hasDecorate(DecorationBankBitsINTEL)) {
+    Out << "{bank_bits:";
+    auto Literals = E->getDecorationLiterals(DecorationBankBitsINTEL);
+    for (size_t I = 0; I < Literals.size() - 1; ++I) {
+      Out << Literals[I] << ",";
+    }
+    Out << Literals[Literals.size() - 1] << '}';
+  }
   if (E->hasDecorate(DecorationUserSemantic))
     Out << E->getDecorationStringLiteral(DecorationUserSemantic).front();
 }
@@ -2848,6 +2856,15 @@ void generateIntelFPGAAnnotationForStructMember(
                                                         MemberNumber))
       Out << ":" << Str;
     Out << '}';
+  }
+  if (E->hasMemberDecorate(DecorationBankBitsINTEL, 0, MemberNumber)) {
+    Out << "{bank_bits:";
+    auto Literals =
+        E->getMemberDecorationLiterals(DecorationBankBitsINTEL, MemberNumber);
+    for (size_t I = 0; I < Literals.size() - 1; ++I) {
+      Out << Literals[I] << ",";
+    }
+    Out << Literals[Literals.size() - 1] << '}';
   }
 
   if (E->hasMemberDecorate(DecorationUserSemantic, 0, MemberNumber))
--- a/lib/SPIRV/SPIRVReader.cpp
+++ b/lib/SPIRV/SPIRVReader.cpp
@@ -2811,9 +2811,8 @@ void generateIntelFPGAAnnotation(const SPIRVEntry *E,
   if (E->hasDecorate(DecorationBankBitsINTEL)) {
     Out << "{bank_bits:";
     auto Literals = E->getDecorationLiterals(DecorationBankBitsINTEL);
-    for (size_t I = 0; I < Literals.size() - 1; ++I) {
+    for (size_t I = 0; I < Literals.size() - 1; ++I)
       Out << Literals[I] << ",";
-    }
     Out << Literals[Literals.size() - 1] << '}';
   }
   if (E->hasDecorate(DecorationUserSemantic))
@@ -2861,9 +2860,8 @@ void generateIntelFPGAAnnotationForStructMember(
     Out << "{bank_bits:";
     auto Literals =
         E->getMemberDecorationLiterals(DecorationBankBitsINTEL, MemberNumber);
-    for (size_t I = 0; I < Literals.size() - 1; ++I) {
+    for (size_t I = 0; I < Literals.size() - 1; ++I)
       Out << Literals[I] << ",";
-    }
     Out << Literals[Literals.size() - 1] << '}';
   }
 
--- a/lib/SPIRV/SPIRVReader.cpp
+++ b/lib/SPIRV/SPIRVReader.cpp
@@ -2813,7 +2813,7 @@ void generateIntelFPGAAnnotation(const SPIRVEntry *E,
     auto Literals = E->getDecorationLiterals(DecorationBankBitsINTEL);
     for (size_t I = 0; I < Literals.size() - 1; ++I)
       Out << Literals[I] << ",";
-    Out << Literals[Literals.size() - 1] << '}';
+    Out << Literals.back() << '}';
   }
   if (E->hasDecorate(DecorationUserSemantic))
     Out << E->getDecorationStringLiteral(DecorationUserSemantic).front();
@@ -2862,7 +2862,7 @@ void generateIntelFPGAAnnotationForStructMember(
         E->getMemberDecorationLiterals(DecorationBankBitsINTEL, MemberNumber);
     for (size_t I = 0; I < Literals.size() - 1; ++I)
       Out << Literals[I] << ",";
-    Out << Literals[Literals.size() - 1] << '}';
+    Out << Literals.back() << '}';
   }
 
   if (E->hasMemberDecorate(DecorationUserSemantic, 0, MemberNumber))
--- a/lib/SPIRV/SPIRVWriter.cpp
+++ b/lib/SPIRV/SPIRVWriter.cpp
@@ -1196,6 +1196,7 @@ tryParseIntelFPGAAnnotationString(StringRef AnnotatedCode) {
                 .Case("bankwidth", DecorationBankwidthINTEL)
                 .Case("max_private_copies", DecorationMaxPrivateCopiesINTEL)
                 .Case("max_replicates", DecorationMaxReplicatesINTEL)
+                .Case("bank_bits", DecorationBankBitsINTEL)
                 .Case("merge", DecorationMergeINTEL)
                 .Default(DecorationUserSemantic);
       if (Dec == DecorationUserSemantic)
@@ -1235,6 +1236,18 @@ void addIntelFPGADecorations(
       StringRef Direction = StringRef(I.second).split(':').second;
       E->addDecorate(
           new SPIRVDecorateMergeINTELAttr(E, Name.str(), Direction.str()));
+    } break;
+    case DecorationBankBitsINTEL: {
+      SmallVector<StringRef, 1> A;
+      StringRef(I.second).split(A, ',');
+      std::vector<SPIRVWord> Bits(A.size());
+      SPIRVWord Result;
+      for (size_t i = 0; i < A.size(); ++i) {
+        Result = 0;
+        A[i].getAsInteger(10, Result);
+        Bits[i] = Result;
+      }
+      E->addDecorate(new SPIRVDecorateBankBitsINTELAttr(E, Bits));
     } break;
     case DecorationRegisterINTEL:
     case DecorationSinglepumpINTEL:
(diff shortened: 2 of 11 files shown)


lib/SPIRV/SPIRVReader.cpp
```diff
@@ -2808,6 +2808,14 @@ void generateIntelFPGAAnnotation(const SPIRVEntry *E,
       Out << ":" << Str;
     Out << '}';
   }
+  if (E->hasDecorate(DecorationBankBitsINTEL)) {
+    Out << "{bank_bits:";
+    auto Literals = E->getDecorationLiterals(DecorationBankBitsINTEL);
+    for (size_t I = 0; I < Literals.size() - 1; ++I) {
```
AGindinson:
```suggestion
    for (auto &Decoration L : Literals) {
```

lib/SPIRV/SPIRVReader.cpp
```diff
@@ -2849,6 +2857,15 @@ void generateIntelFPGAAnnotationForStructMember(
       Out << ":" << Str;
     Out << '}';
   }
+  if (E->hasMemberDecorate(DecorationBankBitsINTEL, 0, MemberNumber)) {
+    Out << "{bank_bits:";
+    auto Literals =
+        E->getMemberDecorationLiterals(DecorationBankBitsINTEL, MemberNumber);
+    for (size_t I = 0; I < Literals.size() - 1; ++I) {
```
AGindinson: Same as with the previous comment - suggest using a range-based loop.

lib/SPIRV/SPIRVReader.cpp
```diff
@@ -2808,6 +2808,14 @@ void generateIntelFPGAAnnotation(const SPIRVEntry *E,
       Out << ":" << Str;
     Out << '}';
   }
+  if (E->hasDecorate(DecorationBankBitsINTEL)) {
+    Out << "{bank_bits:";
+    auto Literals = E->getDecorationLiterals(DecorationBankBitsINTEL);
+    for (size_t I = 0; I < Literals.size() - 1; ++I) {
+      Out << Literals[I] << ",";
+    }
+    Out << Literals[Literals.size() - 1] << '}';
```
AGindinson: Would it make sense to factor this routine out into a separate function and re-use in `generateIntelFPGAAnnotationForStructMember()`? (Just asking)

test/IntelFPGAMemoryAttributes.ll
```diff
@@ -146,25 +150,32 @@ entry:
...
   call void @llvm.var.annotation(i8* %var_ten10, i8* getelementptr inbounds ([37 x i8], [37 x i8]* @.str.10, i32 0, i32 0), i8* getelementptr inbounds ([13 x i8], [13 x i8]* @.str.1, i32 0, i32 0), i32 11)
-  %9 = bitcast i32* %var_ten to i8*
-  call void @llvm.lifetime.end.p0i8(i64 4, i8* %9) #4
-  %10 = bitcast i32* %var_nine to i8*
+  %9 = bitcast i32* %var_eleven to i8*
+  call void @llvm.lifetime.start.p0i8(i64 4, i8* %9) #4
+  %var_eleven11 = bitcast i32* %var_eleven to i8*
+  ; CHECK-LLVM: call void @llvm.var.annotation(i8* [[VAR11:%[a-zA-Z0-9_]+]], i8* getelementptr inbounds ([46 x i8], [46 x i8]* [[STR11]], i32 0, i32 0), i8* undef, i32 undef)
+  call void @llvm.var.annotation(i8* %var_eleven11, i8* getelementptr inbounds ([46 x i8], [46 x i8]* @.str.11, i32 0, i32 0), i8* getelementptr inbounds ([13 x i8], [13 x i8]* @.str.1, i32 0, i32 0), i32 12)
+  %10 = bitcast i32* %var_eleven to i8*
```
AGindinson: (it may sound a bit daunting) Please, consider rewriting the test to use a separate function for each attribute? This would make any subsequent changes far easier. A `TODO` and addressing this separately would be fine.

lib/SPIRV/SPIRVReader.cpp
vmaksimo: It would make sense if we need to iterate through all the literals. The problem is in a comma symbol, that should be printed after all literals except the last one. That's why there is no range loop here (maybe there is a better solution?)

lib/SPIRV/SPIRVReader.cpp
vmaksimo: Maybe not, it is a story about just 3 common lines of code 🤔

AGindinson: Ah, of course. My bad.

lib/SPIRV/SPIRVReader.cpp
AGindinson: Ah, of course. My bad.

test/IntelFPGAMemoryAttributes.ll
vmaksimo: Actually, it is a good idea to rewrite test this way, should be easier to support. I will left a `TODO` note

lib/SPIRV/SPIRVReader.cpp
MrSidims: nit: no need for brackets here and after for single-string loops, but if the rest code in the file is also written in the same way, I'm OK with leaving it as is.

mlychkov:
@AlexeySotkin please review for merge speedup.

lib/SPIRV/SPIRVReader.cpp
```diff
@@ -2808,6 +2808,13 @@ void generateIntelFPGAAnnotation(const SPIRVEntry *E,
       Out << ":" << Str;
     Out << '}';
   }
+  if (E->hasDecorate(DecorationBankBitsINTEL)) {
+    Out << "{bank_bits:";
+    auto Literals = E->getDecorationLiterals(DecorationBankBitsINTEL);
+    for (size_t I = 0; I < Literals.size() - 1; ++I)
+      Out << Literals[I] << ",";
+    Out << Literals[Literals.size() - 1] << '}';
```
AlexeySotkin:
```suggestion
    Out << Literals.back() << '}';
```

lib/SPIRV/SPIRVReader.cpp
```diff
@@ -2849,6 +2856,14 @@ void generateIntelFPGAAnnotationForStructMember(
       Out << ":" << Str;
     Out << '}';
   }
+  if (E->hasMemberDecorate(DecorationBankBitsINTEL, 0, MemberNumber)) {
+    Out << "{bank_bits:";
+    auto Literals =
+        E->getMemberDecorationLiterals(DecorationBankBitsINTEL, MemberNumber);
+    for (size_t I = 0; I < Literals.size() - 1; ++I)
+      Out << Literals[I] << ",";
+    Out << Literals[Literals.size() - 1] << '}';
```
AlexeySotkin:
```suggestion
    Out << Literals.back() << '}';
```

lib/SPIRV/SPIRVWriter.cpp
```diff
@@ -1236,6 +1237,18 @@ void addIntelFPGADecorations(
       E->addDecorate(
           new SPIRVDecorateMergeINTELAttr(E, Name.str(), Direction.str()));
     } break;
+    case DecorationBankBitsINTEL: {
+      SmallVector<StringRef, 1> A;
+      StringRef(I.second).split(A, ',');
+      std::vector<SPIRVWord> Bits(A.size());
+      SPIRVWord Result;
+      for (size_t J = 0; J < A.size(); ++J) {
```
AlexeySotkin: for (auto& Bit: Bits) {

lib/SPIRV/SPIRVWriter.cpp
```diff
@@ -1236,6 +1237,18 @@ void addIntelFPGADecorations(
       E->addDecorate(
           new SPIRVDecorateMergeINTELAttr(E, Name.str(), Direction.str()));
     } break;
+    case DecorationBankBitsINTEL: {
+      SmallVector<StringRef, 1> A;
```
AlexeySotkin: I'm not sure 1 is a good choice for this parameter.
Can we have a more meaningful name then just A? Something like `Bits` maybe.

lib/SPIRV/SPIRVWriter.cpp
```diff
@@ -1236,6 +1237,18 @@ void addIntelFPGADecorations(
...
+      StringRef(I.second).split(A, ',');
+      std::vector<SPIRVWord> Bits(A.size());
+      SPIRVWord Result;
+      for (size_t J = 0; J < A.size(); ++J) {
+        Result = 0;
+        A[J].getAsInteger(10, Result);
```
AlexeySotkin: Will it work?
```suggestion
        A[J].getAsInteger(10, Bits[J]);
```

AlexeySotkin: We need to check if `getAsInteger()` call was successful.

lib/SPIRV/SPIRVWriter.cpp
```diff
@@ -1286,6 +1299,19 @@ void addIntelFPGADecorationsForStructMember(
...
+        Bits[J] = Result;
+      }
```
AlexeySotkin: This piece of code looks identical to the non-member variant of this decoration. Can we factor it out?

lib/SPIRV/SPIRVWriter.cpp
vmaksimo: Yes, it will work if we don't use range-based loop

lib/SPIRV/SPIRVWriter.cpp
vmaksimo: Sure

lib/SPIRV/libSPIRV/SPIRVDecorate.h
```diff
@@ -500,6 +515,20 @@ class SPIRVMemberDecorateMergeINTELAttr : public SPIRVMemberDecorate {
...
 };
 
+class SPIRVMemberDecorateBankBitsINTELAttr : public SPIRVMemberDecorate {
+public:
+  // Complete constructor for BankBitsINTEL decoration
+  SPIRVMemberDecorateBankBitsINTELAttr(SPIRVEntry *TheTarget,
+                                       SPIRVWord MemberNumber,
+                                       const std::vector<SPIRVWord> &TheBits)
+      : SPIRVMemberDecorate(DecorationBankBitsINTEL, MemberNumber, TheTarget) {
+    for (auto &I : TheBits) {
```
mlychkov: Just a nit:
Literals.insert(Literals.end(), TheBits.begin(), TheBits.end());

lib/SPIRV/SPIRVWriter.cpp
```diff
@@ -1210,6 +1211,18 @@ tryParseIntelFPGAAnnotationString(StringRef AnnotatedCode) {
   return Decorates;
 }
 
+std::vector<SPIRVWord> getBankBitsFromString(StringRef S) {
```
AlexeySotkin: Probably in the future we should make this function more generic(add a separator arg) and move it to OCLUtil.

lib/SPIRV/libSPIRV/SPIRVDecorate.h
AlexeySotkin: Why we can't just call the copy constructor?
```suggestion
    Literals(TheBits) {
```

vmaksimo: Copy constructor here can be called as `Literals = TheBits`. Seems, that it's the simplest way to add elements to Literals.

AlexeySotkin:
@vmaksimo please fix lit tests.

lib/SPIRV/libSPIRV/SPIRVDecorate.h
```diff
@@ -448,6 +450,17 @@ class SPIRVDecorateMergeINTELAttr : public SPIRVDecorate {
   }
 };
 
+class SPIRVDecorateBankBitsINTELAttr : public SPIRVDecorate {
+public:
+  // Complete constructor for BankBitsINTEL decoration
+  SPIRVDecorateBankBitsINTELAttr(SPIRVEntry *TheTarget,
+                                 const std::vector<SPIRVWord> &TheBits)
+      : SPIRVDecorate(DecorationBankBitsINTEL, TheTarget) {
+    Literals = TheBits;
```
mlychkov: Why not to move it even in creation block:
`
    : SPIRVDecorate(DecorationBankBitsINTEL, TheTarget),
      Literals(TheBits) {
      ....
`

AlexeySotkin: @vmaksimo Please apply comment from @mlychkov  and squash the commits into one. Thanks.

vmaksimo: It's not possible to access this field this way, because it's a field in a parent class. We can only initialize fields of the current class and call the constructors of parent classes, but they don't contain Literals vector initialization.

PR #401, opened 2019-12-19
Gracefully reject invalid integer bit widths

svenvh:
Instead of asserting (which may not be available in release builds),
reject invalid/unsupported integer bit widths.


keryell:
It is pretty common on FPGA to have 512 bit integers just to have efficient memory transfers...

svenvh:
> It is pretty common on FPGA to have 512 bit integers just to have efficient memory transfers...

Sure, if you have a use case to relax the 64-bit limitation I'd recommend raising an issue or opening a PR.  This PR is only to catch a current limitation of the translator in a more user-friendly way.

MrSidims:
I had a feeling, that SPIR-V supports integers of any width, or it this feeling contradicts the spec? Honestly, I can't find it in the spec.

svenvh:
> I had a feeling, that SPIR-V supports integers of any width, or it this feeling contradicts the spec? Honestly, I can't find it in the spec.

I don't think there is a restriction from the SPIR-V side on integer bit width, but the OpenCL SPIR-V Environment Specification does limit the bit width to 64 and that's where the translator's restriction comes from I suppose.

AlexeySotkin:
#### 2.16.1. Universal Validation Rules
* Data rules 
> Scalar integer types can be parameterized only as 32 bit, plus any additional sizes enabled by capabilities.

lib/SPIRV/SPIRVWriter.cpp
```diff
@@ -257,8 +257,14 @@ SPIRVType *LLVMToSPIRV::transType(Type *T) {
   if (T->isIntegerTy(1))
     return mapType(T, BM->addBoolType());
 
-  if (T->isIntegerTy())
-    return mapType(T, BM->addIntegerType(T->getIntegerBitWidth()));
+  if (T->isIntegerTy()) {
+    unsigned BitWidth = T->getIntegerBitWidth();
+    if (BM->getErrorLog().checkError(BitWidth > 1 && BitWidth <= 64,
```
AlexeySotkin: Does it make sense to check that bitwidth is precisely 8,16,32 or 64 bits?

svenvh: I wasn't aware that bit width restrictions are actually dictated by the SPIR-V spec itself, thanks for pointing that out!  Yes, in this case it makes sense to have an explicit check for those sizes.  I will update the patch and add a spec reference.

PR #402, opened 2019-12-24
[SYCL] Fix incorrect debug info translation of recursive template type

mlychkov:
When translating debug info for composite type from IR to SPIR-V
its template parameters are processed after deletion of temporary
forward declaration of composite type's metadata node. However,
pointer on this node is still present in metadata nodes cache map.
If some of template parameters or their dependences has reference
on this composite type's node then it leads to accessing to freed
memory.


mlychkov:
@AlexeySotkin @AGindinson @MrSidims Could you please review it?

PR #405, opened 2020-01-09
Align FPGA Loop Controls translation with the spec (rev C)

AGindinson:
The specification itself can be found at
https://github.com/KhronosGroup/SPIRV-Registry/blob/master/extensions/INTEL/SPV_INTEL_fpga_loop_controls.asciidoc

As per revision B -> revision C transition in the spec, the patch
removes the "extended loop controls" field, storing the FPGA loop
controls as bit fields in the "common" LoopControl mask.

### DIFF
--- a/lib/SPIRV/SPIRVReader.cpp
+++ b/lib/SPIRV/SPIRVReader.cpp
@@ -628,6 +628,7 @@ void SPIRVToLLVM::setLLVMLoopMetadata(const LoopInstType *LM,
           getMetadataFromNameAndParameter("llvm.loop.ivdep.safelen",
                                           LoopControlParameters[NumParam])));
       ++NumParam;
+      // TODO: Fix the increment/assertion logic in all of the conditions
       assert(NumParam <= LoopControlParameters.size() &&
              "Missing loop control parameter!");
     }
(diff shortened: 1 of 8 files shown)


AGindinson:
Tagging @AlexeySotkin @mlychkov @MrSidims @vmaksimo

lib/SPIRV/SPIRVReader.cpp
```diff
@@ -666,117 +666,106 @@ void SPIRVToLLVM::setLLVMLoopMetadata(const LoopInstType *LM,
...
-      }
-      case MaxConcurrencyINTEL: {
-        Metadata.push_back(llvm::MDNode::get(
-            *Context, getMetadataFromNameAndParameter(
-                          "llvm.loop.max_concurrency.count",
-                          LoopControlParameters[++NumParam])));
-        break;
+  if (LC & InitiationIntervalINTEL) {
+    // To generate a correct integer part of metadata we skip a parameter
+    // that encodes name of the metadata and take the next one
```
AlexeySachkov: Is it still true? You have changed access pattern around `NumParam` from pre-increment to post-increment

lib/SPIRV/SPIRVReader.cpp
```diff
@@ -666,117 +666,106 @@ void SPIRVToLLVM::setLLVMLoopMetadata(const LoopInstType *LM,
...
+    Metadata.push_back(llvm::MDNode::get(
+        *Context, getMetadataFromNameAndParameter(
+                      "llvm.loop.ii.count", LoopControlParameters[NumParam])));
+    ++NumParam;
+    assert(NumParam <= LoopControlParameters.size() &&
```
AlexeySachkov: I believe relation operator should be `<`, because you can only access elements in range `[0, LoopControlParameters.size())`. Can be fixed in a separate NFC patch in all places at once

lib/SPIRV/SPIRVReader.cpp
```diff
@@ -666,117 +666,106 @@ void SPIRVToLLVM::setLLVMLoopMetadata(const LoopInstType *LM,
...
+        // that GEP instructions for some of the arrays were already
+        // marked by the algorithm when it went over the outer level loops.
+        // In order to retain the IVDep information for each "loop
+        // dimension", we will mark such GEP's into a separate joined node
+        // that will refer to the previous levels' index groups AND to the
+        // index group specific to the current loop.
+        std::vector<llvm::Metadata *> CurrentDepthOperands;
+        for (auto &Op : PreviousIdxGroup->operands())
+          CurrentDepthOperands.push_back(Op);
+        if (CurrentDepthOperands.size() == 0)
```
AlexeySachkov:
```suggestion
        if (CurrentDepthOperands.empty())
```

lib/SPIRV/SPIRVReader.cpp
```diff
@@ -666,117 +666,106 @@ void SPIRVToLLVM::setLLVMLoopMetadata(const LoopInstType *LM,
...
-        llvm_unreachable(
-            "Unexpected token in LoopControlExtendedConstrolsMask");
+        // If we're dealing with an embedded loop, it may be the case
+        // that GEP instructions for some of the arrays were already
+        // marked by the algorithm when it went over the outer level loops.
+        // In order to retain the IVDep information for each "loop
+        // dimension", we will mark such GEP's into a separate joined node
+        // that will refer to the previous levels' index groups AND to the
+        // index group specific to the current loop.
+        std::vector<llvm::Metadata *> CurrentDepthOperands;
```
AlexeySachkov:
```suggestion
        std::vector<llvm::Metadata *> CurrentDepthOperands(PreviousIdxGroup->op_begin(), PreviousIdxGroup->op_end());
```

lib/SPIRV/SPIRVReader.cpp
```diff
@@ -666,117 +666,106 @@ void SPIRVToLLVM::setLLVMLoopMetadata(const LoopInstType *LM,
...
+      SPIRVId ArraySPIRVId = LoopControlParameters[++NumParam];
+      Value *ArrayVar = ValueMap[M->getValue(ArraySPIRVId)];
+      unsigned Safelen = LoopControlParameters[++NumParam];
+      ArraySflnMap.emplace(ArrayVar, Safelen);
+    }
+
+    // A single run over the loop to retrieve all GetElementPtr instructions
+    // that access relevant array variables
+    std::map<Value *, std::vector<GetElementPtrInst *>> ArrayGEPMap;
+    for (auto &BB : LoopObj->blocks()) {
```
AlexeySachkov: const?

lib/SPIRV/SPIRVReader.cpp
AGindinson: Nice catch. I'd actually prefer fixing this within this PR!

AGindinson: Actually, there's more to it: the assertions should be moved to stand before the accesses to `LoopControlParameters`' elements.

AGindinson: Or **actually**, I wouldn't. I'll add a `TODO` here though.

lib/SPIRV/SPIRVReader.cpp
AGindinson: Fixed this and other code-style leftovers with a separate commit

lib/SPIRV/SPIRVReader.cpp
AGindinson: Thanks, removed the comment

test/transcoding/FPGALoopAttr.ll
```diff
@@ -23,7 +23,7 @@ entry:
   store i32 0, i32* %i, align 4
   br label %for.cond
 ; CHECK-SPIRV: 4 LoopMerge {{[0-9]+}} {{[0-9]+}} 4
```
AlexeySotkin: Please add a comment explaining which metadata was used in translation to get the LoopMerge argument(s).

AGindinson: Would adding a comment with the LoopControlMask field(s) name(s) suffice? Akin to what I've done for `FPGAIVDepLoopAttr.ll`.

AGindinson: I've updated the files with more comments - please decide if these are good enough.

PR #408, opened 2020-01-14
Fix translation of read_image* built-ins to SPIR-V

AlexeySachkov:
There are several overloads of read_image function defined by OpenCL C
spec, but all of them use the same SPIR-V instruction, so, we need to
add one more optional postfix to differentiate instructions with the
same argument types, but different return types.

Example:
- int4 read_imagei(image2d_t, sampler_t, int2)
- uint4 read_imageio(image2d_t, sampler_t, int2)

Both functions above are represented by the same SPIR-V instruction and
we need to distinguish them in SPIR-V friendly LLVM IR.

### DIFF
--- a/lib/SPIRV/OCL20ToSPIRV.cpp
+++ b/lib/SPIRV/OCL20ToSPIRV.cpp
@@ -1027,9 +1027,23 @@ void OCL20ToSPIRV::transBuiltin(CallInst *CI, OCLBuiltinTransInfo &Info) {
   unsigned ExtOp = ~0U;
   if (StringRef(Info.UniqName).startswith(kSPIRVName::Prefix))
     return;
-  if (OCLSPIRVBuiltinMap::find(Info.UniqName, &OC))
-    Info.UniqName = getSPIRVFuncName(OC);
-  else if ((ExtOp = getExtOp(Info.MangledName, Info.UniqName)) != ~0U)
+  if (OCLSPIRVBuiltinMap::find(Info.UniqName, &OC)) {
+    if (OC == OpImageRead) {
+      // There are several read_image* functions defined by OpenCL C spec, but
+      // all of them use the same SPIR-V Instruction - some of them might only
+      // differ by return type, so, we need to include return type into the
+      // mangling scheme to get them differentiated.
+      //
+      // Example: int4 read_imagei(image2d_t, sampler_t, int2)
+      //          uint4 read_imageui(image2d_t, sampler_t, int2)
+      // Both functions above are represented by the same SPIR-V
+      // instruction: argument types are the same, only return type is
+      // different
+      Info.UniqName = getSPIRVFuncName(OC, CI->getType());
+    } else {
+      Info.UniqName = getSPIRVFuncName(OC);
+    }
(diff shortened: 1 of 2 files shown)


lib/SPIRV/OCL20ToSPIRV.cpp
```diff
@@ -1027,9 +1027,23 @@ void OCL20ToSPIRV::transBuiltin(CallInst *CI, OCLBuiltinTransInfo &Info) {
...
   if (StringRef(Info.UniqName).startswith(kSPIRVName::Prefix))
     return;
-  if (OCLSPIRVBuiltinMap::find(Info.UniqName, &OC))
-    Info.UniqName = getSPIRVFuncName(OC);
-  else if ((ExtOp = getExtOp(Info.MangledName, Info.UniqName)) != ~0U)
+  if (OCLSPIRVBuiltinMap::find(Info.UniqName, &OC)) {
+    if (OC == OpImageRead) {
+      // There are several read_image* functions defined by OpenCL C spec, but
+      // all of them use the same SPIR-V Instruction - some of them might only
+      // differ by return type, so, we need to include return type into the
```
svenvh: Including the return type in the mangled name does not sound like the correct way to handle this issue.

SPIR-V 1.4 adds the SignExtend and ZeroExtend image operands.  It would be better to make use of those somehow.

We have a downstream patch that uses a single bit in the image operands to disambiguate.  I could try to upstream that instead if you are interested, but it will require a bit of rework and it will not be conformant to any pre-1.4 spec.

AlexeySotkin: We also should disambiguate between the following 2 function calls:
```
uint4 c = read_imageui(input, (int4)(0, 0, 0, 0));
float4 f = read_imagef(input, (int4)(0, 0, 0, 0));
```
I don't think the image operands can help to represent these calls in SPIR-V-friendly LLVM IR.

svenvh: Ah sorry, I didn't realize this was only about the SPIR-V representation in LLVM IR.

LGTM then!

test/read_image.cl
```diff
@@ -0,0 +1,16 @@
+// RUN: %clang_cc1 -triple spir64 -finclude-default-header -O0 -cl-std=CL2.0 -emit-llvm-bc %s -o %t.bc
+// RUN: llvm-spirv %t.bc -o %t.spv
+// RUN: spirv-val %t.spv
+// RUN: llvm-spirv -s %t.bc -o %t1.bc
+// RUN: llvm-dis %t1.bc -o - | FileCheck %s
+
+// CHECK: call spir_func <4 x i32> @_Z24__spirv_ImageRead_Ruint414ocl_image3d_roDv4_i
+// CHECK: call spir_func <4 x float> @_Z25__spirv_ImageRead_Rfloat414ocl_image3d_roDv4_i
+
```
AlexeySotkin: Please add a check that it was translated into 2 OpImageRead instructions with correct return types.

AlexeySachkov: Done

PR #410, opened 2020-01-17
Enable NoGlobalOffsetINTEL execution mode translation

vmaksimo:
This patch implements the rest of SPV_INTEL_kernel_attributes extension.
NoGlobalOffsetINTEL execution mode indicates that the global offset is
always (0, 0, 0). Only valid with the Kernel Execution Model.

Specification can be found there: https://github.com/KhronosGroup/SPIRV-Registry/pull/55

### DIFF
--- a/lib/SPIRV/PreprocessMetadata.cpp
+++ b/lib/SPIRV/PreprocessMetadata.cpp
@@ -214,6 +214,12 @@ void PreprocessMetadata::visit(Module *M) {
           .done();
     }
 
+    // !{void (i32 addrspace(1)*)* @kernel, i32 no_global_work_offset}
+    if (MDNode *NoGlobalOffsetINTEL =
+            Kernel.getMetadata(kSPIR2MD::NoGlobalOffset)) {
+      EM.addOp().add(&Kernel).add(spv::ExecutionModeNoGlobalOffsetINTEL).done();
+    }
+
     // !{void (i32 addrspace(1)*)* @kernel, i32 max_global_work_dim, i32 dim}
     if (MDNode *MaxWorkDimINTEL = Kernel.getMetadata(kSPIR2MD::MaxWGDim)) {
       EM.addOp()
--- a/lib/SPIRV/SPIRVInternal.h
+++ b/lib/SPIRV/SPIRVInternal.h
@@ -376,6 +376,7 @@ const static char WGSize[] = "reqd_work_group_size";
 const static char WGSizeHint[] = "work_group_size_hint";
 const static char SubgroupSize[] = "intel_reqd_sub_group_size";
 const static char MaxWGSize[] = "max_work_group_size";
+const static char NoGlobalOffset[] = "no_global_work_offset";
 const static char MaxWGDim[] = "max_global_work_dim";
 const static char NumSIMD[] = "num_simd_work_items";
 } // namespace kSPIR2MD
--- a/test/IntelFPGAFunctionAttributes.ll
+++ b/test/IntelFPGAFunctionAttributes.ll
@@ -1,7 +1,8 @@
 ;; Can be compiled using https://github.com/intel/llvm SYCL compiler from:
 ;; class Foo {
 ;; public:
-;;   [[intelfpga::max_global_work_dim(1),
+;;   [[intelfpga::no_global_work_offset,
+;;     intelfpga::max_global_work_dim(1),
 ;;     intelfpga::max_work_group_size(1,1,1),
 ;;     intelfpga::num_simd_work_items(8)]] void operator()() {}
 ;; };
@@ -14,6 +15,7 @@
 ;; void bar() {
 ;;   Foo boo;
 ;;   kernel<class kernel_name>(boo);
+;;   kernel<class kernel_name2>([]() [[intelfpga::no_global_work_offset(0)]]{});
 ;; }
 
 ; RUN: llvm-as %s -o %t.bc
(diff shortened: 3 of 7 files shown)


lib/SPIRV/PreprocessMetadata.cpp
```diff
@@ -214,6 +214,15 @@ void PreprocessMetadata::visit(Module *M) {
           .done();
     }
 
+    // !{void (i32 addrspace(1)*)* @kernel, i32 uses_global_work_offset}
+    if (MDNode *NoGlobalOffsetINTEL =
+            Kernel.getMetadata(kSPIR2MD::NoGlobOffset)) {
```
AlexeySachkov:
```suggestion
            Kernel.getMetadata(kSPIR2MD::NoGlobalOffset)) {
```

test/IntelFPGAFunctionAttributes.ll
```diff
@@ -94,7 +97,8 @@ attributes #3 = { nounwind }
 !5 = !{i32 8}
 !6 = !{i32 1, i32 1, i32 1}
 !7 = !{i32 1}
-!8 = !{!9, !9, i64 0}
-!9 = !{!"any pointer", !10, i64 0}
-!10 = !{!"omnipotent char", !11, i64 0}
-!11 = !{!"Simple C++ TBAA"}
+!8 = !{i32 0}
```
AlexeySachkov: Is it possible to get LLVM IR with `uses_global_work_offset` metadata equal to `1`?

vmaksimo: No, this SPIR-V attribute is mapped only to `uses_global_work_offset` equal to `0`

test/IntelFPGAFunctionAttributes.ll
```diff
@@ -1,7 +1,8 @@
 ;; Can be compiled using https://github.com/intel/llvm SYCL compiler from:
 ;; class Foo {
 ;; public:
-;;   [[intelfpga::max_global_work_dim(1),
+;;   [[intelfpga::uses_global_work_offset(0),
```
mendell27: Can you also please test that uses_global_work_offset(0) **does not** set the Execution Mode/metadata?

test/IntelFPGAFunctionAttributes.ll
AlexeySachkov: I mean what if you get input LLVM IR (which needs to be translated into SPIR-V), where `uses_global_work_offset` metadata is equal to `1` or something else?

From the updated test I see that FE compiler you are using doesn't generate this metadata at all if non-zero argument was specified, but this might not be the case in the future or with other FE compiler.

lib/SPIRV/SPIRVInternal.h
```diff
@@ -376,6 +376,7 @@ const static char WGSize[] = "reqd_work_group_size";
 const static char WGSizeHint[] = "work_group_size_hint";
 const static char SubgroupSize[] = "intel_reqd_sub_group_size";
 const static char MaxWGSize[] = "max_work_group_size";
+const static char NoGlobOffset[] = "uses_global_work_offset";
```
AlexeySotkin: I'm confused with the naming. Should the metadata name be something like `no_global_work_offset` ?

AlexeySotkin:
According to KhronosGroup/SPIRV-Registry#55 the name of the extension is "SPV_INTEL_kernel_attributes", not "SPV_INTEL_fpga_function_attributes"

lib/SPIRV/SPIRVInternal.h
vmaksimo: Agree, I will rename this metadata variable name to `UsesGlobalOffset` to avoid this confusing difference

test/IntelFPGAFunctionAttributes.ll
vmaksimo: I see, will add the check for not `0` value

vmaksimo:
> According to [KhronosGroup/SPIRV-Registry#55](https://github.com/KhronosGroup/SPIRV-Registry/pull/55) the name of the extension is "SPV_INTEL_kernel_attributes", not "SPV_INTEL_fpga_function_attributes"
I will fix the commit message while squashing all the commits from this PR (once patch is approved)

lib/SPIRV/SPIRVInternal.h
mendell27: The name of the metadata that the FPGA compiler expecting is **uses_global_work_offset**.  Are you proposing to change that?

AlexeySotkin: The way you are proposing it is confusing even more. Why the variable should be called `UsesGlobalOffset` while the execution mode in the spec is called `NoGlobalOffsetINTEL`?
Just looking at `UsesGlobalOffset` name makes me think that we **do** use global offset in some meaningful way, but description of `NoGlobalOffsetINTEL` tells me that the offset is 0, which is effectively means we **do not** use offset.

vmaksimo: Here the variable name should correlate with LLVM IR metadata naming, so it is called `UsesGlobalOffset` (see the last changes in PR).

I see that all the rest can be confusing, will try to describe it in details:
`uses_global_work_offset` metadata instructs the compiler **to use** or **not to use** the global work offset values. Its valid values are `0` and `1`, where `0` means that we **do not** use the global offset. And this case is mapped/translated to `NoGlobalOffsetINTEL` execution mode in SPIR-V. This logic is implemented in `PreprocessMetadata.cpp`

AlexeySotkin: > The name of the metadata that the FPGA compiler expecting is **uses_global_work_offset**. Are you proposing to change that?

Yes indeed. Is it possible, @mendell27?

mendell27: Please do **not** change the name.

vmaksimo:
Aligned changes with the planned renaming of metadata https://github.com/intel/llvm/pull/1090

test/IntelFPGAFunctionAttributes.ll
vmaksimo: As far as metadata doesn't have arguments with the last compiler changes, I mark this conversation as resolved

PR #418, opened 2020-01-28
Generate specialization constant instructions

AlexeySotkin:
Translate SPIRV-friendly LLVM IR builtins represnting specialization constants `__spirv_SpecConstant` to corresponding SPIR-V instructions.

This approach works only for scalar constants, so support for `OpSpecConstantComposite` and `OpSpecConstantOp` is not implemented in this patch.

This PR also includes a patch to improve error handling for `llvm::getSpecConstInfo` API: Change return value from void to bool. Return `false` to indicate a failure. Return `true` on success.

### DIFF
--- a/lib/SPIRV/SPIRVReader.cpp
+++ b/lib/SPIRV/SPIRVReader.cpp
@@ -3650,7 +3650,7 @@ bool llvm::readSpirv(LLVMContext &C, const SPIRV::TranslatorOpts &Opts,
   return true;
 }
 
-void llvm::getSpecConstInfo(std::istream &IS,
+bool llvm::getSpecConstInfo(std::istream &IS,
                             std::vector<SpecConstInfoTy> &SpecConstInfo) {
   std::unique_ptr<SPIRVModule> BM(SPIRVModule::createSPIRVModule());
   BM->setAutoAddExtensions(false);
@@ -3659,7 +3659,7 @@ void llvm::getSpecConstInfo(std::istream &IS,
   D >> Magic;
   if (!BM->getErrorLog().checkError(Magic == MagicNumber, SPIRVEC_InvalidModule,
                                     "invalid magic number")) {
-    return;
+    return false;
   }
   // Skip the rest of the header
   D.ignore(4);
@@ -3693,4 +3693,5 @@ void llvm::getSpecConstInfo(std::istream &IS,
       D.ignoreInstruction();
     }
   }
+  return !IS.fail();
 }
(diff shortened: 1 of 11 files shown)


lib/SPIRV/SPIRVWriter.cpp
```diff
@@ -2039,6 +2042,28 @@ void LLVMToSPIRV::oclGetMutatedArgumentTypesByBuiltin(
...
 
+SPIRVValue *
+LLVMToSPIRV::transBuiltinToConstant(const std::string &DemangledName,
+                                    CallInst *CI) {
+  Op OC = getSPIRVFuncOC(DemangledName);
+  if(!isSpecConstantOpCode(OC))
+    return nullptr;
+  Type *Ty = CI->getArgOperand(1)->getType();
+  assert(Ty == CI->getType() && "Type mismatch!");
+  Value *V = CI->getArgOperand(1);
```
```c++
  Value *V = CI->getArgOperand(1);
  Type *Ty = V->getType();
  assert(Ty == CI->getType() && "Type mismatch!");
```

lib/SPIRV/SPIRVWriter.cpp
```diff
@@ -2039,6 +2042,28 @@ void LLVMToSPIRV::oclGetMutatedArgumentTypesByBuiltin(
...
+  uint64_t Val = 0;
+  if(Ty->isIntegerTy())
```
```suggestion
  if (Ty->isIntegerTy())
```

The same at line 2057

lib/SPIRV/libSPIRV/SPIRVType.h
```diff
@@ -191,7 +191,7 @@ class SPIRVTypeFloat : public SPIRVType {
     SPIRVCapVec CV;
     if (isTypeFloat(16)) {
       CV.push_back(CapabilityFloat16Buffer);
-      auto Extensions = getModule()->getExtension();
+      auto Extensions = getModule()->getSourceExtension();
```
AlexeySachkov: Seems like unrelated change to me, is it possible to move it into a separate PR?

AlexeySotkin:
@aratajew Please take a look.

aratajew:
> @aratajew Please take a look.
Regarding to the second commit (c2e6c6c9): LGTM, thanks for implementing this.

lib/SPIRV/libSPIRV/SPIRVType.h
AlexeySotkin: What about moving it to a separate commit?

lib/SPIRV/SPIRVWriter.cpp
AlexeySotkin: Fixed, thanks

lib/SPIRV/SPIRVWriter.cpp
AlexeySotkin: Fixed, thanks.

lib/SPIRV/SPIRVReader.cpp
```diff
@@ -3693,4 +3693,7 @@ void llvm::getSpecConstInfo(std::istream &IS,
       D.ignoreInstruction();
     }
   }
+  if (IS.fail())
+    return false;
+  return true;
```
AlexeySachkov:
```c++
return !IS.fail();
```

Issue #357, opened 2019-10-21
Segmentation fault during running llvm-spirv


davidrohr:
The attached test case segfaults for me while running llvm-spirv (Using current master llvm-spirv and latest clang from git commit c78da41cdae532adb51d10ed73482642881346dd + the fixes from https://bugs.llvm.org/show_bug.cgi?id=43145).
[test.tar.gz](https://github.com/KhronosGroup/SPIRV-LLVM-Translator/files/3750470/test.tar.gz)

Commands to reproduce:
```
clang++ -cl-std=clc++ -x cl -emit-llvm --target=spir64-unknown-unknown -cl-denorms-are-zero -cl-mad-enable -cl-no-signed-zeros -ferror-limit=1000 -Xclang -finclude-default-header -Dcl_clang_storage_class_specifiers -Wno-invalid-constexpr -c test.cl -o test.bc
llvm-spirv test.bc
```

I am getting the following segfault:
```
Stack dump:
0.      Program arguments: llvm-spirv test.bc 
1.      Running pass 'LLVMToSPIRV' on module 'test.bc'.
 #0 0x00007f1a027c995a llvm::sys::PrintStackTrace(llvm::raw_ostream&) (/usr/lib/llvm/10/lib64/libLLVM-10svn.so+0xa3e95a)
 #1 0x00007f1a027c7924 llvm::sys::RunSignalHandlers() (/usr/lib/llvm/10/lib64/libLLVM-10svn.so+0xa3c924)
 #2 0x00007f1a027c7aa5 (/usr/lib/llvm/10/lib64/libLLVM-10svn.so+0xa3caa5)
 #3 0x00007f1a018f10d0 (/lib64/libc.so.6+0x390d0)
 #4 0x000055cfd6eb9c86 SPIRV::SPIRVModuleImpl::addUnaryInst(spv::Op, SPIRV::SPIRVType*, SPIRV::SPIRVValue*, SPIRV::SPIRVBasicBlock*) (/usr/bin/llvm-spirv+0x23bc86)
 #5 0x000055cfd6e54446 SPIRV::LLVMToSPIRV::transValueWithoutDecoration(llvm::Value*, SPIRV::SPIRVBasicBlock*, bool) (/usr/bin/llvm-spirv+0x1d6446)
 #6 0x000055cfd6e5533d (/usr/bin/llvm-spirv+0x1d733d)
[...]

PR #429, opened 2020-02-06
Handle non-zero values for llvm.memset

svenvh:
Relax a limitation of the SPIR-V writer.  The memset intrinsic is mapped
to `OpCopyMemorySized`, causing it to be converted to memcpy when
converting back from SPIR-V to LLVM.

Fixes https://github.com/KhronosGroup/SPIRV-LLVM-Translator/issues/357

### DIFF
--- a/test/transcoding/llvm.memset.ll
+++ b/test/transcoding/llvm.memset.ll
@@ -1,49 +1,57 @@
-;; struct S1
-;; {
-;;   int x;
-;;   int y;
-;;   int z;
-;; };
-;; S1 foo11()
-;; {
-;;   return S1();
-;; }
-
 ; RUN: llvm-as %s -o %t.bc
 ; RUN: llvm-spirv %t.bc -spirv-text -o %t.spt
 ; RUN: FileCheck < %t.spt %s --check-prefix=CHECK-SPIRV
 ; RUN: llvm-spirv %t.bc -o %t.spv
 ; RUN: llvm-spirv -r %t.spv -o %t.rev.bc
+; RUN: spirv-val %t.spv
 ; RUN: llvm-dis < %t.rev.bc | FileCheck %s --check-prefix=CHECK-LLVM
 
 ; CHECK-SPIRV: TypeInt [[Int8:[0-9]+]] 8 0
-; CHECK-SPIRV: Constant {{[0-9]+}} [[Len:[0-9]+]] 12
+; CHECK-SPIRV: Constant {{[0-9]+}} [[Lenmemset21:[0-9]+]] 4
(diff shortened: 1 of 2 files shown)


test/transcoding/llvm.memset.ll
```diff
@@ -17,34 +6,46 @@
 ; RUN: llvm-dis < %t.rev.bc | FileCheck %s --check-prefix=CHECK-LLVM
 
 ; CHECK-SPIRV: TypeInt [[Int8:[0-9]+]] 8 0
-; CHECK-SPIRV: Constant {{[0-9]+}} [[Len:[0-9]+]] 12
+; CHECK-SPIRV: Constant {{[0-9]+}} [[Lenmemset21:[0-9]+]] 4
```
AlexeySachkov: This is probably out of scope of this PR, but I wonder why do we have `OpConstant` here instead of `OpConstantComposite`?

AlexeySotkin: If I remember correctly, if the length is large enough(like 128), we have to declare a structure with large number of members, while for the array we just have to specify the number of elements.

test/transcoding/llvm.memset.ll
```diff
@@ -17,34 +6,46 @@
...
 ; CHECK-SPIRV: TypePointer [[Int8PtrConst:[0-9]+]] 0 [[Int8]]
 ; CHECK-SPIRV: ConstantNull [[Int8x12]] [[Init:[0-9]+]]
 ; CHECK-SPIRV: Variable {{[0-9]+}} [[Val:[0-9]+]] 0 [[Init]]
 
 ; CHECK-SPIRV: Bitcast [[Int8Ptr]] [[Target:[0-9]+]] {{[0-9]+}}
 ; CHECK-SPIRV: Bitcast [[Int8PtrConst]] [[Source:[0-9]+]] [[Val]]
-; CHECK-SPIRV: CopyMemorySized [[Target]] [[Source]] [[Len]] 2 4
+; CHECK-SPIRV: CopyMemorySized [[Target]] [[Source]] [[Lenmemset0]] 2 4
 
+; CHECK-SPIRV: CopyMemorySized {{[0-9]+}} {{[0-9]+}} [[Lenmemset21]] 2 4
```
AlexeySotkin: @svenvh could you reflect in the test that the source operand in this instruction is a constant composite and that members of the composite are correct.

svenvh: Done.

test/transcoding/llvm.memset.ll
svenvh: This is the "length" argument which is a scalar, i.e. an `OpConstant`.

PR #443, opened 2020-02-25
Allow Intel Loop Controls only with SPV_INTEL_fpga_loop_controls

MrSidims:
Added the appropriate checks if the extension is allowed.

### DIFF
--- a/lib/SPIRV/SPIRVWriter.cpp
+++ b/lib/SPIRV/SPIRVWriter.cpp
@@ -779,9 +779,9 @@ private:
 /// Go through the operands !llvm.loop metadata attached to the branch
 /// instruction, fill the Loop Control mask and possible parameters for its
 /// fields.
-static spv::LoopControlMask
-getLoopControl(const BranchInst *Branch, std::vector<SPIRVWord> &Parameters,
-               LLVMToSPIRV::LLVMToSPIRVMetadataMap &IndexGroupArrayMap) {
+spv::LoopControlMask
+LLVMToSPIRV::getLoopControl(const BranchInst *Branch,
+                            std::vector<SPIRVWord> &Parameters) {
   if (!Branch)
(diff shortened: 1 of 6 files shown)


test/transcoding/FPGAIVDepLoopAttr.ll
```diff
@@ -52,7 +52,7 @@
 ; }
 
 ; RUN: llvm-as < %s > %t.bc
-; RUN: llvm-spirv %t.bc -o %t.spv
+; RUN: llvm-spirv %t.bc --spirv-ext=+SPV_INTEL_fpga_loop_controls -o %t.spv
```
mlychkov: May we add test without specifying this extension to check that intel loop controls won't be translated and Extension/Capability won't be added in SPIR-V?

MrSidims: That make sense, thanks!

lib/SPIRV/SPIRVWriter.cpp
```diff
@@ -855,6 +863,8 @@ getLoopControl(const BranchInst *Branch, std::vector<SPIRVWord> &Parameters,
       Parameters.push_back(ArraySflnPair.first);
       Parameters.push_back(ArraySflnPair.second);
     }
+    BM->addExtension(ExtensionID::SPV_INTEL_fpga_loop_controls);
```
AlexeySachkov: @MrSidims, you are adding `SPV_INTEL_fpga_loop_controls`  capability here if `DependencyArrayParameters` is not empty. But this might be the case if `"llvm.loop.unroll.disable"` metadata was encountered (for example), which is not part of `SPV_INTEL_fpga_loop_controls`

MrSidims: All ok

PR #447, opened 2020-02-28
Fix build after ddd11273d

MrSidims:
BinaryOperator::CreateFNeg -> UnaryOperator::CreateFNeg


AlexeySachkov:
@StuartDBrady: FYI. This is related to #360.

The change itself looks good to me, since semantics of `OpFnegate` and `fneg` seem to match between each other (and IEEE 754-2008/2019 `negate` operation as well):

> OpFNegate. Inverts the sign bit of Operand.

> `fneg`. The value produced is a copy of the operand with its sign bit flipped.

> "negate(x) copies a floating-point operand x to a destination in the same format, reversing the sign bit. negate(x) is not the same as subtraction(0, x)

PR #448, opened 2020-02-28
travis: Restore macOS builds

svenvh:
macOS builds seem to have been broken since 3027e3a ("Update version
of LLVM from 10 to 11", 2020-01-28) due to apt commands being moved
into the `before_install` step.  Restore macOS testing by only running
apt commands for Linux builds.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -20,17 +20,20 @@ branches:
     - /^v\d+\.\d+(\.\d+)?(-\S*)?$/
 
 before_install:
-  - curl -L "https://apt.llvm.org/llvm-snapshot.gpg.key" | sudo apt-key add -
-  - curl -L "http://packages.lunarg.com/lunarg-signing-key-pub.asc" | sudo apt-key add -
-  - echo "deb http://apt.llvm.org/bionic/ llvm-toolchain-bionic main" | sudo tee -a ${TRAVIS_ROOT}/etc/apt/sources.list
-  - echo "deb https://packages.lunarg.com/vulkan bionic main" | sudo tee -a ${TRAVIS_ROOT}/etc/apt/sources.list
-  - sudo apt-get update
-  - sudo apt-get -yq --no-install-suggests --no-install-recommends install
-      llvm-${LLVM_VERSION}-dev
-      clang-${LLVM_VERSION}
-      clang-format-${LLVM_VERSION}
-      clang-tidy-${LLVM_VERSION}
-      spirv-tools
+  - |
+    if [ $TRAVIS_OS_NAME == "linux" ]; then
+      curl -L "https://apt.llvm.org/llvm-snapshot.gpg.key" | sudo apt-key add -
+      curl -L "http://packages.lunarg.com/lunarg-signing-key-pub.asc" | sudo apt-key add -
+      echo "deb http://apt.llvm.org/bionic/ llvm-toolchain-bionic main" | sudo tee -a ${TRAVIS_ROOT}/etc/apt/sources.list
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -20,17 +20,19 @@ branches:
...
-  - echo "deb http://apt.llvm.org/bionic/ llvm-toolchain-bionic main" | sudo tee -a ${TRAVIS_ROOT}/etc/apt/sources.list
-  - echo "deb https://packages.lunarg.com/vulkan bionic main" | sudo tee -a ${TRAVIS_ROOT}/etc/apt/sources.list
-  - sudo apt-get update
-  - sudo apt-get -yq --no-install-suggests --no-install-recommends install
-      llvm-${LLVM_VERSION}-dev
-      clang-${LLVM_VERSION}
-      clang-format-${LLVM_VERSION}
-      clang-tidy-${LLVM_VERSION}
-      spirv-tools
+  - if [ $TRAVIS_OS_NAME == "linux" ]; then
```
StuartDBrady: The newlines are getting swallowed by YAML, here.

Would probably recommend a literal block scalar as we use elsewhere:
```
  - |
    if [ $TRAVIS_OS_NAME == "line" ]; then
```
(... or you could add semicolons, instead.)

.travis.yml
```diff
@@ -20,17 +20,19 @@ branches:
...
+      curl -L "https://apt.llvm.org/llvm-snapshot.gpg.key" | sudo apt-key add -
+      curl -L "http://packages.lunarg.com/lunarg-signing-key-pub.asc" | sudo apt-key add -
+      echo "deb http://apt.llvm.org/bionic/ llvm-toolchain-bionic main" | sudo tee -a ${TRAVIS_ROOT}/etc/apt/sources.list
+      echo "deb https://packages.lunarg.com/vulkan bionic main" | sudo tee -a ${TRAVIS_ROOT}/etc/apt/sources.list
+      sudo apt-get update
+      sudo apt-get -yq --no-install-suggests --no-install-recommends install
+        llvm-${LLVM_VERSION}-dev
+        clang-${LLVM_VERSION}
+        clang-format-${LLVM_VERSION}
+        clang-tidy-${LLVM_VERSION}
```
StuartDBrady: Note: a literal block scalar would require line continuation characters (`\`) here.

.travis.yml
```diff
@@ -20,17 +20,19 @@ branches:
...
-  - echo "deb https://packages.lunarg.com/vulkan bionic main" | sudo tee -a ${TRAVIS_ROOT}/etc/apt/sources.list
-  - sudo apt-get update
-  - sudo apt-get -yq --no-install-suggests --no-install-recommends install
-      llvm-${LLVM_VERSION}-dev
-      clang-${LLVM_VERSION}
-      clang-format-${LLVM_VERSION}
-      clang-tidy-${LLVM_VERSION}
-      spirv-tools
+  - if [ $TRAVIS_OS_NAME == "linux" ]; then
+      curl -L "https://apt.llvm.org/llvm-snapshot.gpg.key" | sudo apt-key add -
```
StuartDBrady: Could the LLVM parts of this be guarded with a check for `$BUILD_EXTERNAL == 1`, and the `clang-format`/`clang-tidy` parts guarded with a check for `$CHECK_FORMAT == 1`? (We still need the SPIR-V tools, of course.)

.travis.yml
svenvh: Good catch!

.travis.yml
svenvh: I'm not sure if we can guard LLVM parts as we might depend on tools that are not part of the build (because they take too long to build).  This feels out of scope of this change anyway.

PR #449, opened 2020-02-28
Transform fabs and ceil llvm intrinsics into SPIRV extended instructions

asudarsa:

### DIFF
--- a/lib/SPIRV/SPIRVInternal.h
+++ b/lib/SPIRV/SPIRVInternal.h
@@ -57,6 +57,10 @@
 using namespace SPIRV;
 using namespace llvm;
 
+namespace llvm {
+class IntrinsicInst;
+}
+
 namespace SPIRV {
 
 /// The LLVM/SPIR-V translator version used to fill the lower 16 bits of the
@@ -936,6 +940,8 @@ template <> inline void SPIRVMap<std::string, Op, SPIRVOpaqueType>::init() {
 // Check if the module contains llvm.loop.* metadata
 bool hasLoopMetadata(const Module *M);
 
+// check LLVM Intrinsics type(s) for validity
+bool checkTypeForSPIRVExtendedInstLowering(IntrinsicInst *II, SPIRVModule *BM);
 } // namespace SPIRV
 
 #endif // SPIRV_SPIRVINTERNAL_H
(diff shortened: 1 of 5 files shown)


lib/SPIRV/SPIRVInternal.h
```diff
@@ -50,6 +50,7 @@
 #include "llvm/IR/Attributes.h"
 #include "llvm/IR/Constants.h"
 #include "llvm/IR/Instructions.h"
+#include "llvm/IR/IntrinsicInst.h"
```
svenvh: We can avoid this include in the .h file; please use a forward declaration here and include the header only in .cpp files.

test/ceil.ll
```diff
@@ -0,0 +1,66 @@
+; RUN: llvm-as %s -o %t.bc
```
svenvh: Maybe rename this file to `test/llvm.ceil.ll` (to make it clearer this is testing an llvm intrinsic)?  Same for the other test.

lib/SPIRV/SPIRVWriter.cpp
```diff
@@ -1573,6 +1573,23 @@ SPIRVValue *LLVMToSPIRV::transIntrinsicInst(IntrinsicInst *II,
     return BM->addExtInst(transType(II->getType()),
                           BM->getExtInstSetId(SPIRVEIS_OpenCL), OpenCLLIB::Sqrt,
                           {transValue(II->getOperand(0), BB)}, BB);
```
AlexeySotkin: `}` closing the sqrt case is missing.

test/ceil.ll
asudarsa: This has already been checked into xmain branch. Name change might cause a bit of headache during pull down later on.

AlexeySotkin:
@asudarsa Please fix clang-format and clang-tidy issues.

test/ceil.ll
svenvh: > This has already been checked into xmain branch. Name change might cause a bit of headache during pull down later on.

I don't know what the "xmain branch" is, but that is not relevant to this project.  For this project, it would be good to keep some structure in the naming of test files, which is why I made this suggestion.

lib/SPIRV/SPIRVWriter.cpp
asudarsa: AH. That must have snuck through during the merge. Thanks for the catch

test/ceil.ll
asudarsa: I agree. I will change this.

lib/SPIRV/SPIRVInternal.h
asudarsa: Done

test/ceil.ll
asudarsa: Done

lib/SPIRV/SPIRVWriter.cpp
asudarsa: Fixed

PR #456, opened 2020-03-13
Fix build after changes in the llvm project

AlexeySotkin:
The build was broken by llvm/llvm-project@118abf2
"[SVE] Update API ConstantVector::getSplat() to use ElementCount."


svenvh:
Verified locally that this fixes the build.  apt.llvm.org still needs to catch up, but I'll merge this now.

PR #457, opened 2020-03-13
Translate new set of Intel FPGA Loop Controls

vmaksimo:
Specification can be found here:
https://github.com/KhronosGroup/SPIRV-Registry/pull/62

As per revision C -> revision E transition in the spec, patch introduces
translation of the following loop controls:
* PipelineDisableINTEL
* LoopCoalesceINTEL
* MaxInterleavingINTEL
* SpeculatedIterationsINTEL

### DIFF
--- a/lib/SPIRV/SPIRVReader.cpp
+++ b/lib/SPIRV/SPIRVReader.cpp
@@ -786,6 +786,47 @@ void SPIRVToLLVM::setLLVMLoopMetadata(const LoopInstType *LM,
       Metadata.push_back(llvm::MDNode::get(*Context, Parameters));
     }
   }
+  if (LC & LoopControlPipelineEnableINTEL) {
+    Metadata.push_back(llvm::MDNode::get(
+        *Context,
+        getMetadataFromNameAndParameter("llvm.loop.intel.pipelining.enable",
+                                        LoopControlParameters[NumParam])));
+    ++NumParam;
+    assert(NumParam <= LoopControlParameters.size() &&
+           "Missing loop control parameter!");
+  }
+  if (LC & LoopControlLoopCoalesceINTEL) {
+    if (LoopControlParameters.size()) {
+      Metadata.push_back(llvm::MDNode::get(
+          *Context,
+          getMetadataFromNameAndParameter("llvm.loop.coalesce.count",
+                                          LoopControlParameters[NumParam])));
+      ++NumParam;
+    } else { // If LoopCoalesce has no parameters
+      Metadata.push_back(llvm::MDNode::get(
+          *Context, getMetadataFromName("llvm.loop.coalesce.enable")));
+    }
+    assert(NumParam <= LoopControlParameters.size() &&
+           "Missing loop control parameter!");
+  }
+  if (LC & LoopControlMaxInterleavingINTEL) {
+    Metadata.push_back(llvm::MDNode::get(
+        *Context,
+        getMetadataFromNameAndParameter("llvm.loop.max_interleaving.count",
+                                        LoopControlParameters[NumParam])));
+    ++NumParam;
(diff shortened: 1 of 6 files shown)


lib/SPIRV/SPIRVReader.cpp
```diff
@@ -784,6 +784,47 @@ void SPIRVToLLVM::setLLVMLoopMetadata(const LoopInstType *LM,
...
+    Metadata.push_back(llvm::MDNode::get(
+        *Context,
+        getMetadataFromNameAndParameter("llvm.loop.intel.pipelining.enable",
+                                        LoopControlParameters[NumParam])));
+    ++NumParam;
+    assert(NumParam <= LoopControlParameters.size() &&
+           "Missing loop control parameter!");
+  }
+  if (LC & LoopControlLoopCoalesceINTEL) {
+    if (LoopControlParameters.size()) {
```
mlychkov: Should we check params size with NumParam value?

test/transcoding/FPGALoopAttr.ll
```diff
@@ -168,6 +168,175 @@ for.end36:                                        ; preds = %for.cond29
   ret void
 }
 
+; Function Attrs: noinline nounwind optnone
```
AGindinson: This test and `FPGALoopMergeInst.ll` are practically identical. I'd suggest moving these new checks to the better-documented `FPGALoopMergeInst.ll` (adding the source code to the comment section as well).

lib/SPIRV/SPIRVReader.cpp
```diff
@@ -784,6 +784,47 @@ void SPIRVToLLVM::setLLVMLoopMetadata(const LoopInstType *LM,
       Metadata.push_back(llvm::MDNode::get(*Context, Parameters));
     }
   }
+  if (LC & LoopControlPipelineEnableINTEL) {
+    Metadata.push_back(llvm::MDNode::get(
+        *Context,
+        getMetadataFromNameAndParameter("llvm.loop.intel.pipelining.enable",
+                                        LoopControlParameters[NumParam])));
+    ++NumParam;
```
AGindinson:
```suggestion
                                        LoopControlParameters[NumParam++])));
```
..., which could apply to several sections below.

test/transcoding/FPGALoopAttr.ll
vmaksimo: Are you suggesting only moving new checks or duplicating is also fine (although it's obviously not)?

AGindinson: My suggestion is to move the new checks; and to remove this file altogether (the latter probably deserves a separate PR so as not to confuse ourselves & the project maintainers).

lib/SPIRV/SPIRVReader.cpp
vmaksimo: No, these two are not related

lib/SPIRV/SPIRVReader.cpp
```diff
@@ -786,6 +786,43 @@ void SPIRVToLLVM::setLLVMLoopMetadata(const LoopInstType *LM,
...
+  if (LC & LoopControlPipelineEnableINTEL) {
+    Metadata.push_back(llvm::MDNode::get(
+        *Context,
+        getMetadataFromNameAndParameter("llvm.loop.intel.pipelining.enable",
+                                        LoopControlParameters[NumParam++])));
+    assert(NumParam <= LoopControlParameters.size() &&
+           "Missing loop control parameter!");
+  }
+  if (LC & LoopControlLoopCoalesceINTEL) {
+    if (LoopControlParameters.size()) {
```
AGindinson:
```suggestion
    if (!LoopControlParameters.empty()) {
```

mlychkov:
LGTM.

PR #461, opened 2020-03-18
Fix of test, failed due to upstream llvm-objdump option changes

kvladimi:
One debug info test failed because now llvm-objdump do not understand --section-headers option


StuartDBrady:
It would not be a surprise if `DebugInfo/X86/xray-split-dwarf-interaction.ll` also starts failing at some point in the future, as it uses `llvm-readelf -sections`.

PR #464, opened 2020-03-19
[Backport] Implementation of SPV_INTEL_inline_assembly extension

NikitaRudenkoIntel:
Extension is published as github.com/intel/llvm/pull/1290
Original commit: github.com/KhronosGroup/SPIRV-LLVM-Translator/commit/7a0767f2a


AlexeySachkov:
Do we really need to backport this to `llvm_release_70`?

kvladimi:
> Do we really need to backport this to `llvm_release_70`?

It is not urgent, and probably not immediately required, but it is for free. Why not?

PR #469, opened 2020-03-20
Add support for read/write ocl builtins from cl_intel_subgroups_{char…

aus-intel:
…|long}

Link to PR with extension: https://github.com/KhronosGroup/OpenCL-Registry/pull/79.


lib/SPIRV/SPIRVReader.cpp
```diff
@@ -2774,8 +2774,23 @@ std::string SPIRVToLLVM::getOCLBuiltinName(SPIRVInstruction *BI) {
...
+      case 16:
         Name << "_us";
+        break;
+      case 32:
+        // Intentionally does nothing since _ui variant is only an alias.
+        break;
+      case 64:
+        Name << "_ul";
+        break;
+      default:
```
AlexeySotkin: There are 2 very similar pieces of code. Can we move it to an util function? Will it work if we will return "_ui" in the reader?

aus-intel: I think that it is better to not return "_ui" because cl_intel_subgroups uses version only without suffix. Alias with suffix was added in subsequent extensions (char, short, long). However, I can try to drop "_ui" from "visitSubgroupBlockReadINTEL". As I understand, in this function some kind of demangling is done and postfix is added only to distinguish between different signatures. Without "_ui" all variants will be different too and everying should work fine.
Will it be fine if I move code for postfixes to OCLUtil?

AlexeySotkin: > Will it be fine if I move code for postfixes to OCLUtil?

Yes

aus-intel:
Moved common part to OCLUtil.cpp. @AlexeySotkin, @AlexeySachkov, can you verify changes please?

lib/SPIRV/OCL20ToSPIRV.cpp
```diff
@@ -1594,6 +1594,29 @@ void OCL20ToSPIRV::visitCallKernelQuery(CallInst *CI, StringRef DemangledName) {
...
+static void processSubgroupBlockReadWriteINTEL(CallInst *CI,
+                                               OCLBuiltinTransInfo &Info,
+                                               const Type *DataTy, Module *M) {
+  unsigned VectorNumElements = 1;
+  if (DataTy->isVectorTy())
+    VectorNumElements = DataTy->getVectorNumElements();
+  unsigned ElementBitSize = DataTy->getScalarSizeInBits();
+  Info.Postfix = "_";
+  Info.Postfix +=
+      getIntelSubgroupBlockDataPostfix(ElementBitSize, VectorNumElements);
```
AlexeySotkin: That will generate double underscore like `__us2` in `_Z41__spirv_SubgroupImageBlockWriteINTEL__us214ocl_image2d_woDv2_iDv2_s`.
Is it intentional? If not we can
```suggestion
  Info.Postfix =
      getIntelSubgroupBlockDataPostfix(ElementBitSize, VectorNumElements);
```
Maybe we don't need underscores at all?

aus-intel: Everything that comes after underscore considered to be just postfix for internal functions created between ocl2tospirv pass and writer, and it won't be visible for others. In writer, this postfix is dropped without any analysis.
However, if there is no underscore postfix it will be an error. Extra underscore solves this problem for integer subgroup_block builtins where no underscore will be generated by postfix function and doesn't do wrong things for other versions. In case we have "intel_sub_group_block_write2" without extra "_" there will be generated something like "__spirv_SubgroupImageBlockWriteINTEL2" and then "dePrefixSPIRVName" will not delete "2" from name (since it searches for underscore) and translation will be incorrect. If we want to drop underscores, "dePrefixSPIRVName" should be changed too. But this can complicate analysis of real spirv op name. It will be quite problematic to find spirv name without any markers like underscore.

PR #470, opened 2020-03-23
Implement -spirv-allow-unknown-intrinsics option

imashkov:
Here is implementation of command-line option which will allow
to represent unknown llvm intrinsics as external function call in SPIR-V

Signed-off-by: Ilya Mashkov <ilya.mashkov@intel.com>

### DIFF
--- a/lib/SPIRV/SPIRVWriter.cpp
+++ b/lib/SPIRV/SPIRVWriter.cpp
@@ -91,6 +91,11 @@ namespace SPIRV {
 cl::opt<bool> SPIRVMemToReg("spirv-mem2reg", cl::init(false),
                             cl::desc("LLVM/SPIR-V translation enable mem2reg"));
 
+cl::opt<bool> SPIRVAllowUnknownIntrinsics(
+    "spirv-allow-unknown-intrinsics", cl::init(false),
+    cl::desc("Unknown LLVM intrinsics will be translated as external function "
+             "calls in SPIR-V"));
+
 static void foreachKernelArgMD(
     MDNode *MD, SPIRVFunction *BF,
     std::function<void(const std::string &Str, SPIRVFunctionParameter *BA)>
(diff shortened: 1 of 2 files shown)


lib/SPIRV/SPIRVWriter.cpp
```diff
@@ -1766,11 +1771,18 @@ SPIRVValue *LLVMToSPIRV::transIntrinsicInst(IntrinsicInst *II,
...
-                                 __FILE__, __LINE__);
+    if (SPIRVAllowUnknownIntrinsics)
+      return BM->addCallInst(
+          transFunctionDecl(II->getCalledFunction()),
+          transArguments(II, BB,
+                         SPIRVEntry::createUnique(OpFunctionCall).get()),
+          BB);
+    else
+      // Other LLVM intrinsics shouldn't get to SPIRV, because they
+      // can't be represented in SPIRV or not implemented yet.
```
AGindinson:
```suggestion
      // can't be represented in SPIRV or aren't implemented yet.
```

test/AllowIntrinsics.ll
```diff
@@ -0,0 +1,37 @@
+; The test checks command-line option for the translator
+; which will allow to represent unknown llvm intrinsics as external function call in SPIR-V.
+; RUN: llvm-as %s -o %t.bc
+; RUN: llvm-spirv -spirv-allow-unknown-intrinsics %t.bc -o %t.spv
```
svenvh: Maybe add a `spirv-val` step as well, to ensure the generated SPIR-V is valid?

imashkov: Done, thanks for suggestion

lib/SPIRV/SPIRVWriter.cpp
imashkov: Yes, grammar is important :) . Done.

imashkov:
@AlexeySotkin  It seems like patch is ready to be merged.

PR #471, opened 2020-03-24
Translate Intel FPGA force_pow2_depth memory attribute

mlychkov:

### DIFF
--- a/test/IntelFPGAMemoryAttributes.ll
+++ b/test/IntelFPGAMemoryAttributes.ll
@@ -150,6 +150,25 @@
 ;   s.field = 0;
 ; }
 ;
+; void force_pow2_depth_attr() {
+;   [[intelfpga::force_pow2_depth(0)]] int fp2d_var;
+;
+;   [[intelfpga::force_pow2_depth(1)]] struct fp2d_st {
+;     int field;
+;   } s;
+;   s.field = 0;
+; }
+;
+; template <int A>
+; void templ_force_pow2_depth_attr() {
+;   [[intelfpga::force_pow2_depth(A)]] int templ_fp2d_var;
+;
+;   [[intelfpga::force_pow2_depth(A)]] struct templ_fp2d_st {
+;     int field;
+;   } s;
+;   s.field = 0;
+; }
+;
 ; template <typename name, typename Func>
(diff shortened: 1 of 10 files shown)


mlychkov:
@MrSidims @AGindinson @AlexeySotkin Please review

test/IntelFPGAMemoryAttributes.ll
```diff
@@ -150,6 +150,25 @@
 ;   s.field = 0;
 ; }
 ;
+; void force_pow2_depth_attr() {
+;   [[intelfpga::force_pow2_depth(0)]] int fp2d_var;
+;
+;   [[intelfpga::force_pow2_depth(1)]] struct fp2d_st {
```
AGindinson: Could we have a test-case for `force_pow2_depth(0)` on a `struct`?

test/IntelFPGAMemoryAttributesForStruct.ll
```diff
@@ -556,6 +583,38 @@ entry:
   ret void
 }
 
+; Function Attrs: norecurse nounwind
+define spir_func void @_Z27field_force_pow2_depth_attrv() #3 {
+entry:
+  %s = alloca %struct.force_pow2_depth_st, align 4
+  %0 = bitcast %struct.force_pow2_depth_st* %s to i8*
+  call void @llvm.lifetime.start.p0i8(i64 4, i8* %0) #5
+  ; CHECK-LLVM: %[[FLD_FP2_SCT:.*]] = getelementptr inbounds %struct.force_pow2_depth_st, %struct.force_pow2_depth_st* %{{[a-zA-Z0-9]+}}, i32 0, i32 0
```
AGindinson: IMO, the mapping on a particular global (line 280-281) had best be checked. The globals could be pre-extracted by `CHECK-DAG`s, I believe.

test/IntelFPGAMemoryAttributes.ll
mlychkov: We may have it, but for what reason? From translator's point of view processing of attribute with any argument value is not depend on a type of variable.

AGindinson: Fair enough - just my morning safe-guard mood speaking :)

PR #474, opened 2020-03-25
Add translation of fast math flags for binary arithmetic operators and add fneg/fcmp transcoding tests

StuartDBrady:


test/transcoding/fadd.ll
```diff
@@ -0,0 +1,65 @@
...
+; CHECK-SPIRV: 3 Name [[#r7:]] "r7"
+; CHECK-SPIRV-NOT: 4 Decorate [[#r1]] FPFastMathMode
+; CHECK-SPIRV-DAG: 4 Decorate [[#r2]] FPFastMathMode 1
+; CHECK-SPIRV-DAG: 4 Decorate [[#r3]] FPFastMathMode 2
+; CHECK-SPIRV-DAG: 4 Decorate [[#r4]] FPFastMathMode 4
+; CHECK-SPIRV-DAG: 4 Decorate [[#r5]] FPFastMathMode 8
+; CHECK-SPIRV-DAG: 4 Decorate [[#r6]] FPFastMathMode 16
+; CHECK-SPIRV-DAG: 4 Decorate [[#r7]] FPFastMathMode 3
+; CHECK-SPIRV: 3 TypeFloat [[float:[0-9]+]] 32
+; CHECK-SPIRV: 5 FAdd [[float]]
```
AlexeySotkin:
```suggestion
; CHECK-SPIRV: 5 FAdd [[float]] [[#r1]]
```

test/transcoding/fdiv.ll
```diff
@@ -0,0 +1,65 @@
...
+; CHECK-SPIRV: 3 Name [[#r7:]] "r7"
+; CHECK-SPIRV-NOT: 4 Decorate [[#r1]] FPFastMathMode
+; CHECK-SPIRV-DAG: 4 Decorate [[#r2]] FPFastMathMode 1
+; CHECK-SPIRV-DAG: 4 Decorate [[#r3]] FPFastMathMode 2
+; CHECK-SPIRV-DAG: 4 Decorate [[#r4]] FPFastMathMode 4
+; CHECK-SPIRV-DAG: 4 Decorate [[#r5]] FPFastMathMode 8
+; CHECK-SPIRV-DAG: 4 Decorate [[#r6]] FPFastMathMode 16
+; CHECK-SPIRV-DAG: 4 Decorate [[#r7]] FPFastMathMode 3
+; CHECK-SPIRV: 3 TypeFloat [[float:[0-9]+]] 32
+; CHECK-SPIRV: 5 FDiv [[float]]
```
AlexeySotkin:
```suggestion
; CHECK-SPIRV: 5 FDiv [[float]] [[#r1]]
```

test/transcoding/fmul.ll
```diff
@@ -0,0 +1,65 @@
...
+; CHECK-SPIRV: 3 Name [[#r7:]] "r7"
+; CHECK-SPIRV-NOT: 4 Decorate [[#r1]] FPFastMathMode
+; CHECK-SPIRV-DAG: 4 Decorate [[#r2]] FPFastMathMode 1
+; CHECK-SPIRV-DAG: 4 Decorate [[#r3]] FPFastMathMode 2
+; CHECK-SPIRV-DAG: 4 Decorate [[#r4]] FPFastMathMode 4
+; CHECK-SPIRV-DAG: 4 Decorate [[#r5]] FPFastMathMode 8
+; CHECK-SPIRV-DAG: 4 Decorate [[#r6]] FPFastMathMode 16
+; CHECK-SPIRV-DAG: 4 Decorate [[#r7]] FPFastMathMode 3
+; CHECK-SPIRV: 3 TypeFloat [[float:[0-9]+]] 32
+; CHECK-SPIRV: 5 FMul [[float]]
```
AlexeySotkin:
```suggestion
; CHECK-SPIRV: 5 FMul [[float]] [[#r1]]
```

test/transcoding/frem.ll
```diff
@@ -0,0 +1,65 @@
...
+; CHECK-SPIRV: 3 Name [[#r7:]] "r7"
+; CHECK-SPIRV-NOT: 4 Decorate [[#r1]] FPFastMathMode
+; CHECK-SPIRV-DAG: 4 Decorate [[#r2]] FPFastMathMode 1
+; CHECK-SPIRV-DAG: 4 Decorate [[#r3]] FPFastMathMode 2
+; CHECK-SPIRV-DAG: 4 Decorate [[#r4]] FPFastMathMode 4
+; CHECK-SPIRV-DAG: 4 Decorate [[#r5]] FPFastMathMode 8
+; CHECK-SPIRV-DAG: 4 Decorate [[#r6]] FPFastMathMode 16
+; CHECK-SPIRV-DAG: 4 Decorate [[#r7]] FPFastMathMode 3
+; CHECK-SPIRV: 3 TypeFloat [[float:[0-9]+]] 32
+; CHECK-SPIRV: 5 FRem [[float]]
```
AlexeySotkin:
```suggestion
; CHECK-SPIRV: 5 FRem [[float]] [[#r1]]
```

test/transcoding/fsub.ll
```diff
@@ -0,0 +1,65 @@
...
+; CHECK-SPIRV: 3 Name [[#r7:]] "r7"
+; CHECK-SPIRV-NOT: 4 Decorate [[#r1]] FPFastMathMode
+; CHECK-SPIRV-DAG: 4 Decorate [[#r2]] FPFastMathMode 1
+; CHECK-SPIRV-DAG: 4 Decorate [[#r3]] FPFastMathMode 2
+; CHECK-SPIRV-DAG: 4 Decorate [[#r4]] FPFastMathMode 4
+; CHECK-SPIRV-DAG: 4 Decorate [[#r5]] FPFastMathMode 8
+; CHECK-SPIRV-DAG: 4 Decorate [[#r6]] FPFastMathMode 16
+; CHECK-SPIRV-DAG: 4 Decorate [[#r7]] FPFastMathMode 3
+; CHECK-SPIRV: 3 TypeFloat [[float:[0-9]+]] 32
+; CHECK-SPIRV: 5 FSub [[float]]
```
AlexeySotkin:
```suggestion
; CHECK-SPIRV: 5 FSub [[float]] [[#r1]]
```

lib/SPIRV/SPIRVWriter.cpp
```diff
@@ -1320,6 +1320,32 @@ bool LLVMToSPIRV::transDecoration(Value *V, SPIRVValue *BV) {
     }
   }
 
+  if (auto BVF = dyn_cast_or_null<FPMathOperator>(V)) {
+    auto Opcode = BVF->getOpcode();
+    if (Opcode == Instruction::FAdd ||
+        Opcode == Instruction::FSub ||
+        Opcode == Instruction::FMul ||
+        Opcode == Instruction::FDiv ||
+        Opcode == Instruction::FRem) {
```
AlexeySotkin: According to https://www.khronos.org/registry/spir-v/specs/unified1/SPIRV.html#_a_id_fp_fast_math_mode_a_fp_fast_math_mode, looks like `OpFMod` is missing, isn't it?

test/transcoding/fadd.ll
StuartDBrady: Thanks, will update that.

lib/SPIRV/SPIRVWriter.cpp
StuartDBrady: There is no `FMod` instruction in LLVM IR.

SPIR-V's `OpFRem` is equivalent to LLVM IR's `FRem` operation, which is equivalent to the C99 remainder operator (`%`). (C90 had a modulus operator, instead, which give an implementation-defined result if one or both of the operands is negative. C99 replaced this with a remainder operator which gives fully language-defined results for negative operands.)

It seems that certain graphics language specs define an operation that is equivalent to `OpFMod`, and it has also been used by some non-graphics languages (e.g. Ada), but my understanding is that we are unlikely to see heavy use of this for OpenCL in the foreseeable future, although support for translation _from_ `OpFMod` is mandated by the SPIR-V specification.

In any case, it translates to a complex sequence of instructions in LLVM IR, of which it may be possible to eliminate some when fast math flags are applied. If we did want to apply fast math flags to instructions generated when translating `OpFMod`, we would want to be selective about which instructions we apply those flags to. Given that `OpFMod` is basically unused for OpenCL, my view is that it's worth being compliant with the specification, but not worth optimizing at present.

StuartDBrady: Note that GLSL has a `mod()` function that is described in GLSL 4.60.7 § 12.2.14. _Mapping of other instructions_ as mapping to `OpFMod`.

The `%` operator is described in that same section as mapping to `OpSMod`, but the description of the `%` operator in § 5.9 _Expressions_ states that "results are undefined if one or both operands are negative", so by my reading, `OpSRem` would be valid, too.

I suppose if we wanted the translator to support GLSL shaders that have been translated into SPIR-V which uses `OpFMod` and/or `OpSMod`, and we wanted to translate those `OpFMod` and/or `OpSMod` instructions into something _efficient_, we would be forced to introduce intrinsics (or _perhaps_ an extra instruction or two in LLVM IR).

test/transcoding/fadd.ll
StuartDBrady: Thanks, fixed.

test/transcoding/fmul.ll
StuartDBrady: Thanks, fixed.

test/transcoding/fsub.ll
StuartDBrady: Thanks, fixed.

test/transcoding/frem.ll
StuartDBrady: Thanks, fixed.

test/transcoding/fdiv.ll
StuartDBrady: Thanks, fixed.

PR #475, opened 2020-03-26
Fix translation of SPIR-V OpControlBarier to LLVM IR

AlexeySachkov:
This patch features two things:
- fixed translation of OpControlBarrier into OpenCL 2.0 built-ins:

  Scope of the operation is now respected and we can either get
  work_group_barrier or sub_group_barrier depending on it

- improved the translation so it handles non-constant 'Memory Semantic'
  operand of the instruction

### DIFF
--- a/lib/SPIRV/SPIRVToOCL20.cpp
+++ b/lib/SPIRV/SPIRVToOCL20.cpp
@@ -135,11 +135,23 @@ void SPIRVToOCL20::visitCallSPIRVControlBarrier(CallInst *CI) {
           return cast<ConstantInt>(Args[I])->getZExtValue();
         };
         auto ExecScope = static_cast<Scope>(GetArg(0));
-        auto MScope = static_cast<Scope>(GetArg(1));
-        auto Sema = mapSPIRVMemSemanticToOCL(GetArg(2));
+        auto MemScope = static_cast<Scope>(GetArg(1));
+
+        if (auto Arg = dyn_cast<ConstantInt>(Args[2])) {
+          auto Sema = mapSPIRVMemSemanticToOCL(Arg->getZExtValue());
+          Args[0] = getInt32(M, Sema.first);
+        } else {
+          int ClMemFenceMask = MemorySemanticsWorkgroupMemoryMask |
+                               MemorySemanticsCrossWorkgroupMemoryMask |
+                               MemorySemanticsImageMemoryMask;
+          Args[0] = getOrCreateSwitchFunc(
+              kSPIRVName::TranslateSPIRVMemFence, Args[2],
+              OCLMemFenceExtendedMap::getRMap(), true /*IsReverse*/, None, CI,
+              M, ClMemFenceMask);
+        }
+        Args[1] = getInt32(M, rmap<OCLScopeKind>(MemScope));
         Args.resize(2);
-        Args[0] = getInt32(M, Sema.first);
-        Args[1] = getInt32(M, rmap<OCLScopeKind>(MScope));
+
         return (ExecScope == ScopeWorkgroup) ? kOCLBuiltinName::WorkGroupBarrier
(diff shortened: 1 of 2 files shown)


AlexeySachkov:
@aratajew: FYI

lib/SPIRV/SPIRVToOCL20.cpp
```diff
@@ -135,11 +135,29 @@ void SPIRVToOCL20::visitCallSPIRVControlBarrier(CallInst *CI) {
...
         auto ExecScope = static_cast<Scope>(GetArg(0));
-        auto MScope = static_cast<Scope>(GetArg(1));
-        auto Sema = mapSPIRVMemSemanticToOCL(GetArg(2));
+        auto MemScope = static_cast<Scope>(GetArg(1));
+
         Args.resize(2);
-        Args[0] = getInt32(M, Sema.first);
-        Args[1] = getInt32(M, rmap<OCLScopeKind>(MScope));
+        if (auto Arg = dyn_cast<ConstantInt>(Args[2])) {
+           auto Sema = mapSPIRVMemSemanticToOCL(Arg->getZExtValue());
```
svenvh: Indentation seems off by one here (see also the Travis build).
It would be nice to also have a brief comment that says which cases the `if` is distinguishing.

AlexeySachkov: Fixed

lib/SPIRV/SPIRVToOCL20.cpp
```diff
@@ -135,11 +135,33 @@ void SPIRVToOCL20::visitCallSPIRVControlBarrier(CallInst *CI) {
...
+          auto Sema = mapSPIRVMemSemanticToOCL(Arg->getZExtValue());
+          Args[0] = getInt32(M, Sema.first);
+        } else {
+          CallInst *TransCall = dyn_cast<CallInst>(Args[2]);
+          Function *F = TransCall ? TransCall->getCalledFunction() : nullptr;
+          // In case the SPIR-V module was created from an OpenCL program by
+          // *this* SPIR-V generator, we know that the value passed to
+          // __translate_ocl_memory_fence is what we should pass to the OpenCL
+          // builtin now.
+          if (F && F->getName().equals(kSPIRVName::TranslateOCLMemFence)) {
```
AlexeySachkov: This is not quite true actually: we allow translating of non-constant memory semantic from SPIR-V to LLVM IR, but not vice-versa.

It is probably better to prepare a separate patch to address that, because there are several places similar to that one

AlexeySachkov: Ok, I though that there is at least `TranslateOCLMemFence` function and apparently it doesn't exists. I don't see a lot value then in wrong condition here: remove this kind of optimization completely

lib/SPIRV/SPIRVToOCL20.cpp
```diff
@@ -135,11 +135,23 @@ void SPIRVToOCL20::visitCallSPIRVControlBarrier(CallInst *CI) {
...
         };
         auto ExecScope = static_cast<Scope>(GetArg(0));
-        auto MScope = static_cast<Scope>(GetArg(1));
-        auto Sema = mapSPIRVMemSemanticToOCL(GetArg(2));
+        auto MemScope = static_cast<Scope>(GetArg(1));
+
         Args.resize(2);
-        Args[0] = getInt32(M, Sema.first);
-        Args[1] = getInt32(M, rmap<OCLScopeKind>(MScope));
+        if (auto Arg = dyn_cast<ConstantInt>(Args[2])) {
```
svenvh: This is accessing `Args[2]`, but there's `Args.resize(2)` a few lines above.

AlexeySachkov: Good catch, thanks! Should be fixed now

PR #476, opened 2020-03-26
Fix translation of relational operations

Fznamznon:
Whereas it is not possible to generate vector of bools in FE,
we have to change return type for corresponding instructions in SPIRV
translator to vector of bools. SPIRV translator already did this for
some instructions, this patch extends this behaviour to handle more
instructions.

### DIFF
--- a/lib/SPIRV/SPIRVWriter.cpp
+++ b/lib/SPIRV/SPIRVWriter.cpp
@@ -2417,24 +2417,28 @@ LLVMToSPIRV::transBuiltinToInstWithoutDecoration(Op OC, CallInst *CI,
     if (isCvtOpCode(OC) && OC != OpGenericCastToPtrExplicit) {
       return BM->addUnaryInst(OC, transType(CI->getType()),
                               transValue(CI->getArgOperand(0), BB), BB);
-    } else if (isCmpOpCode(OC)) {
-      assert(CI && CI->getNumArgOperands() == 2 && "Invalid call inst");
+    } else if (isCmpOpCode(OC) || isRelationalOpCode(OC)) {
       auto ResultTy = CI->getType();
       Type *BoolTy = IntegerType::getInt1Ty(M->getContext());
       auto IsVector = ResultTy->isVectorTy();
       if (IsVector)
         BoolTy = VectorType::get(BoolTy, ResultTy->getVectorNumElements());
       auto BBT = transType(BoolTy);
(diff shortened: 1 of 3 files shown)


lib/SPIRV/SPIRVWriter.cpp
```diff
@@ -2417,24 +2417,28 @@ LLVMToSPIRV::transBuiltinToInstWithoutDecoration(Op OC, CallInst *CI,
...
       Type *BoolTy = IntegerType::getInt1Ty(M->getContext());
       auto IsVector = ResultTy->isVectorTy();
       if (IsVector)
         BoolTy = VectorType::get(BoolTy, ResultTy->getVectorNumElements());
       auto BBT = transType(BoolTy);
-      auto Cmp = BM->addCmpInst(OC, BBT, transValue(CI->getArgOperand(0), BB),
-                                transValue(CI->getArgOperand(1), BB), BB);
+      SPIRVInstruction *Res;
+      if (isCmpOpCode(OC))
+        Res = BM->addCmpInst(OC, BBT, transValue(CI->getArgOperand(0), BB),
```
svenvh: You could move the assert that you removed above into this `if` (and also add an assert `CI && CI->getNumArgOperands() == 1` in the `else`.

Fznamznon: Sure, added.

lib/SPIRV/libSPIRV/SPIRVOpCode.h
```diff
@@ -74,6 +74,10 @@ inline bool isLogicalOpCode(Op OpCode) {
   return (unsigned)OpCode >= OpLogicalEqual && (unsigned)OpCode <= OpLogicalNot;
 }
 
+inline bool isRelationalOpCode(Op OpCode) {
```
AlexeySotkin: What do you think about `isUnaryPredicateOpCode` name for the function?

Fznamznon: Hmm, okay.

Fznamznon: Done.

AlexeySotkin:
@Fznamznon, could you squash your commits?

bader:
> @Fznamznon, could you squash your commits?

I/you can squash during merge. Do you have any other comments?

bader:
https://github.com/KhronosGroup/SPIRV-LLVM-Translator/commit/cb3be8626ec224aaf1bdf76a26ac5232b34d43fa - squashed commit.

PR #478, opened 2020-03-31
Do not rely on OpSource while translating into LLVM IR

AlexeySachkov:
- Refactored `--spirv-ocl-builtins-version` option so it can be set via library
  interface as well and renamed it into `--spirv-target-env`
- Removed logic that relied on OpSource while generating LLVM IR for
  OpenCL built-in functions
- Added one more library entry entry point: 'convertSpirvToLLVM' which
  is a copy of existing one, but accept TranslatorOpts argument

### DIFF
--- a/include/LLVMSPIRVLib.h
+++ b/include/LLVMSPIRVLib.h
@@ -140,6 +140,13 @@ bool getSpecConstInfo(std::istream &IS,
 std::unique_ptr<Module>
 convertSpirvToLLVM(LLVMContext &C, SPIRV::SPIRVModule &BM, std::string &ErrMsg);
 
+/// \brief Convert a SPIRVModule into LLVM IR using specified options
+/// \returns null on failure.
+std::unique_ptr<Module> convertSpirvToLLVM(LLVMContext &C,
+                                           SPIRV::SPIRVModule &BM,
+                                           const SPIRV::TranslatorOpts &Opts,
+                                           std::string &ErrMsg);
+
 /// \brief Regularize LLVM module by removing entities not representable by
 /// SPIRV.
 bool regularizeLlvmForSpirv(Module *M, std::string &ErrMsg);
@@ -183,7 +190,7 @@ ModulePass *createSPIRVRegularizeLLVM();
 
 /// Create a pass for translating SPIR-V builtin functions to OCL builtin
 /// functions.
-ModulePass *createSPIRVToOCL(Module &M);
+ModulePass *createSPIRVToOCL(Module &, SPIRV::BIsRepresentation);
 
 /// Create a pass for translating SPIR-V builtin functions to OCL 1.2 builtin
 /// functions.
--- a/include/LLVMSPIRVLib.h
+++ b/include/LLVMSPIRVLib.h
@@ -188,9 +188,9 @@ ModulePass *createSPIRVLowerMemmove();
 /// Create a pass for regularize LLVM module to be translated to SPIR-V.
 ModulePass *createSPIRVRegularizeLLVM();
 
-/// Create a pass for translating SPIR-V builtin functions to OCL builtin
-/// functions.
-ModulePass *createSPIRVToOCL(Module &, SPIRV::BIsRepresentation);
+/// Create a pass for translating SPIR-V Instructions to desired
+/// representation in LLVM IR (OpenCL built-ins, SPIR-V Friendly IR, etc.)
+ModulePass *createSPIRVBIsLoweringPass(Module &, SPIRV::BIsRepresentation);
 
 /// Create a pass for translating SPIR-V builtin functions to OCL 1.2 builtin
 /// functions.
--- a/include/LLVMSPIRVOpts.h
+++ b/include/LLVMSPIRVOpts.h
@@ -65,12 +65,7 @@ enum class ExtensionID : uint32_t {
   Last,
 };
 
-enum class BIsRepresentation : uint32_t {
-  OpenCL12,
-  OpenCL20
-  // TODO: consider targeting SPIR-V friendly IR for some non-OpenCL backends,
-  // if there are any
-};
+enum class BIsRepresentation : uint32_t { OpenCL12, OpenCL20, SPIRVFriendlyIR };
 
 /// \brief Helper class to manage SPIR-V translation
 class TranslatorOpts {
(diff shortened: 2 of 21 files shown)


AlexeySachkov:
@aratajew: FYI

lib/SPIRV/SPIRVToOCL.cpp
```diff
@@ -403,31 +398,15 @@ std::string SPIRVToOCL::getGroupBuiltinPrefix(CallInst *CI) {
...
+    case SPIRV::BIsRepresentation::OpenCL12:
       return createSPIRVToOCL12();
-    else if (OCLBuiltinsVersion.getValue() == "CL2.0" ||
-             OCLBuiltinsVersion.getValue() == "CL2.1")
+    case SPIRV::BIsRepresentation::OpenCL20:
       return createSPIRVToOCL20();
-    else {
-      assert(0 && "Invalid spirv-ocl-builtins-version value");
+    default:
+      assert(false && "Unsupported built-ins representation");
```
Naghasan:
```suggestion
      llvm_unreachable("Unsupported built-ins representation");
```

include/LLVMSPIRVOpts.h
```diff
@@ -65,6 +65,13 @@ enum class ExtensionID : uint32_t {
   Last,
 };
 
+enum class BIsRepresentation : uint32_t {
+  OpenCL12,
+  OpenCL20
+  // TODO: consider targeting SPIR-V friendly IR for some non-OpenCL backends,
```
Naghasan: Wouldn't this also help testing / debug ?

AlexeySachkov: Added one more option for that

tools/llvm-spirv/llvm-spirv.cpp
```diff
@@ -116,6 +116,17 @@ static cl::opt<bool> SPIRVGenKernelArgNameMD(
     cl::desc("Enable generating OpenCL kernel argument name "
              "metadata"));
 
+// TODO: rename this option
+static cl::opt<SPIRV::BIsRepresentation> BIsRepresentation(
+    "spirv-ocl-builtins-version",
+    cl::desc("Specify version of OCL builtins to translate "),
```
AlexeySachkov, line 122: Since the option is now more generic than just OCL builtins version, we should probably rename it. @svenvh, @Naghasan, any suggestions?

svenvh: How about (probably) the shortest and simplest: `--builtins`?

include/LLVMSPIRVLib.h
```diff
@@ -188,9 +188,9 @@ ModulePass *createSPIRVLowerMemmove();
 /// Create a pass for regularize LLVM module to be translated to SPIR-V.
 ModulePass *createSPIRVRegularizeLLVM();
 
-/// Create a pass for translating SPIR-V builtin functions to OCL builtin
-/// functions.
-ModulePass *createSPIRVToOCL(Module &, SPIRV::BIsRepresentation);
+/// Create a pass for translating SPIR-V builtin functions to desired
+/// represenation (OpenCL built-ins, SPIR-V Friendly IR, etc.)
```
svenvh, line 192:
```suggestion
/// representation (OpenCL built-ins, SPIR-V Friendly IR, etc.)
```

tools/llvm-spirv/llvm-spirv.cpp
Naghasan: +1 this keeps it neutral so probably the best. Probably better to have `spirv-builtins` for uniformity.

AlexeySachkov: What about `--builtins-representation`? Or `--bi(s)-representation`?

To me, personally, just `builtins` looks a bit unclear, what is builtins, which builtins?

svenvh: +1 for keeping uniformity so `spirv-builtins` would be better.  Although that could be confusing because it can also be interpreted as "SPIR-V builtins"?

I don't think adding the word "representation" answers the question "what/which builtins".

Another idea: `spirv-builtins-target` or `spirv-target-builtins`?

AlexeySachkov: > I don't think adding the word "representation" answers the question "what/which builtins".

Together with proper description it should be clear that this option is about how to represent "SPIRV builtins".

> Although that could be confusing because it can also be interpreted as "SPIR-V builtins"?

I agree here. What about even more generic `spirv-ops-target`, `spirv-ops-representation`, `spirv-to-ir-representation`, `spirv-in-ir-representation`?

AlexeySachkov: `spirv-target-builtins` with description talking about how to represent _SPIR-V instructions_, rather than _SPIR-V Builtins_ looks good to me as well

svenvh: > I agree here. What about even more generic spirv-ops-target

Building on top of that suggestion: `spirv-target-env`?  It's nice and short, and implies that the output will be targeted towards a particular environment.

AlexeySachkov: > Building on top of that suggestion: spirv-target-env? It's nice and short, and implies that the output will be targeted towards a particular environment.

Looks nice!

We might even consider adding additional validation based on this option, using [OpenCL SPIR-V Environment spec](https://www.khronos.org/registry/OpenCL/specs/2.2/html/OpenCL_Env.html) as reference

lib/SPIRV/SPIRVToOCL.cpp
AlexeySachkov: Done

include/LLVMSPIRVLib.h
AlexeySachkov: Fixed

AlexeySachkov:
@svenvh, @Naghasan, I decided to go with `spirv-target-env` - please let me know if there are any comments.

I probably would like to squash commits prior merging to have one commit describing final diff rather than a few including "apply code-review comments"

tools/llvm-spirv/llvm-spirv.cpp
Naghasan: +1

PR #483, opened 2020-04-01
Fix build after D72467 Remove "mask" operand from shufflevector.

AlexeySachkov:


lib/SPIRV/SPIRVReader.cpp
```diff
@@ -1974,20 +1974,23 @@ Value *SPIRVToLLVM::transValueWithoutDecoration(SPIRVValue *BV, Function *F,
...
 
       case 4: {
         for (unsigned Idx = 0; Idx < 4; ++Idx) {
-          Value *V1 =
-              Builder.CreateShuffleVector(MCache[0], MCache[1], {Idx, Idx + 4});
-          Value *V2 =
-              Builder.CreateShuffleVector(MCache[2], MCache[3], {Idx, Idx + 4});
-          Value *V3 = Builder.CreateShuffleVector(V1, V2, {0, 1, 2, 3});
+          Value *V1 = Builder.CreateShuffleVector(
+              MCache[0], MCache[1], ArrayRef<int>({Idx, Idx + 4}));
```
svenvh, line 1989: I opened this one earlier: https://github.com/KhronosGroup/SPIRV-LLVM-Translator/pull/482 ;-)

The only difference I can see is that I used `ArrayRef<unsigned int>` here to avoid narrowing-conversion warnings.

bader: Does `CreateShuffleVector` has an overload for `ArrayRef<unsigned int>`? I see only for `ArrayRef<uint32_t>`.

AlexeySachkov: @bader, I guess `unsigned int` is equal to `uint32_t` in most cases

@svenvh, let's proceed with your version

AlexeySachkov:
Closing in favor of KhronosGroup/SPIRV-LLVM-Translator#482

lib/SPIRV/SPIRVReader.cpp
bader: > @bader, I guess `unsigned int` is equal to `uint32_t` in most cases

No always, whereas `uint32_t` is always equal to `uint32_t`.
On some systems #482 is not build-able.

svenvh: bq. On some systems #482 is not build-able.

I'm sorry if #482 is causing trouble for you on some systems; could you elaborate on the problem?  Is it that you are getting narrowing conversion warnings when `unsigned int` is larger than 32 bits for example? (which would be false positives, as values are in range `[0, 3+4]`).

bader: @svenvh, #482 is not causing troubles for me yet.
My notes are about unnecessary type conversion.
I think the clean fix should use `int` or `uint32_t` types for mask instead of `unsigned int`.
Here is API translator is using:
https://github.com/llvm/llvm-project/blob/master/llvm/include/llvm/IR/IRBuilder.h#L2625
```c++
        for (int Idx = 0; Idx < 4; ++Idx) {
          Value *V1 = Builder.CreateShuffleVector(
            MCache[0], MCache[1], ArrayRef<int>({Idx, Idx + 4}));
```
or 
```c++
        for (uint32_t Idx = 0; Idx < 4; ++Idx) {
          Value *V1 = Builder.CreateShuffleVector(
            MCache[0], MCache[1], ArrayRef<uint32_t>({Idx, Idx + 4}));
```

svenvh: Fair enough, I have opened https://github.com/KhronosGroup/SPIRV-LLVM-Translator/pull/484 to clean up the fix.

bader: Thanks!

PR #488, opened 2020-04-10
Fix build after D76269

AlexeySachkov:
Several deprecated IR Builder APIs were removed, see
https://reviews.llvm.org/D76269 [opaque pointer types] Remove deprecated
Instruction/IRBuilder APIs.


AlexeySachkov:
Missed the second unformatted fragment of code, CI should pass now

PR #491, opened 2020-04-13
Backport 4f459a8e1c6d9e5d15ad58ce42bc4cee6081ee5f and

aus-intel:
99e0a01715ab79dfd7f5d07c1b302ca0591fa0b1 (squashed):

Move Pass ID out of SPIRVToOCL base class

Having the SPIRVToOCL12 and SPIRVToOCL20 passes share the base class
Pass ID could cause the LLVM PassRegistry to believe it would register
the same pass twice.

Move the ID member down the class hierarchy such that each pass has
its own ID.

Fix SPIRVToOCL pass ID argument to be a reference

The ModulePass constructor takes a reference so we should forward the
reference given to the SPIRVToOCL constructor.

Reported-by: Stuart Brady

### DIFF
--- a/lib/SPIRV/SPIRVToOCL.cpp
+++ b/lib/SPIRV/SPIRVToOCL.cpp
@@ -56,8 +56,6 @@ static cl::opt<std::string>
                        cl::desc("Specify version of OCL builtins to translate "
                                 "to (CL1.2, CL2.0, CL2.1)"));
 
-char SPIRVToOCL::ID = 0;
-
 void SPIRVToOCL::visitCallInst(CallInst &CI) {
   LLVM_DEBUG(dbgs() << "[visistCallInst] " << CI << '\n');
   auto F = CI.getCalledFunction();
--- a/lib/SPIRV/SPIRVToOCL.h
+++ b/lib/SPIRV/SPIRVToOCL.h
@@ -49,7 +49,7 @@
 namespace SPIRV {
 class SPIRVToOCL : public ModulePass, public InstVisitor<SPIRVToOCL> {
 protected:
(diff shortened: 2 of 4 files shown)


AlexeySotkin:
@aus-intel please fix the formatting issues https://travis-ci.org/github/KhronosGroup/SPIRV-LLVM-Translator/jobs/674473079#L548

aus-intel:
There is a strange problem with chech_code_format.sh. As I see from the source, it gets diff between master and current commit.
`FORMAT_DIFF=$(git diff -U0 master -- ${FILES_TO_CHECK} | ./utils/clang-format-diff.py -p1 -style=file)`
This is not correct for release branches -- diff should be taken between release branch and commit to this branch. In this PR this led to the problem, that script found problem in the code that was not modified by this particular PR. This part of code was modified in my previous PR to master branch, and with this PR script thinks that the code was reformatted back and raises an error.
PR 490 is merged, and I will just rebase this change on top of 90 branch. The error should disappear. However, there still will be an error in check_code_format.sh.

AlexeySotkin:
> There is a strange problem with chech_code_format.sh. As I see from the source, it gets diff between master and current commit.
> `FORMAT_DIFF=$(git diff -U0 master -- ${FILES_TO_CHECK} | ./utils/clang-format-diff.py -p1 -style=file)`
> This is not correct for release branches -- diff should be taken between release branch and commit to this branch. In this PR this led to the problem, that script found problem in the code that was not modified by this particular PR. This part of code was modified in my previous PR to master branch, and with this PR script thinks that the code was reformatted back and raises an error.

Thank you for noticing this. I have created an issue #496

PR #492, opened 2020-04-14
[NFC] Refactor for upcoming D77278 Remove asserting getters from base Type

AlexeySachkov:
A few methods is going to be removed from Type class, see
https://reviews.llvm.org/D77278

Applied the following transformation to the code:

```
Ty->getVectorNumElements() -> cast<VectorType>(Ty)->getNumElements()
Ty->getVectorElementCount() -> cast<VectorType>(Ty)->getElementCount()
Ty->getVectorElementType() -> cast<llvm::VectorType>(Ty)->getElementType()
```


svenvh:
Thanks for doing this ahead of time!

PR #502, opened 2020-04-22
Support lowering of constant expressions vector

zuban32:
Since the translator has to lower LLVM constant expressions to be able to represent them in SPIR-V,
we now need to support not only single constexprs but vectors of constexprs as well.

### DIFF
--- a/lib/SPIRV/SPIRVLowerConstExpr.cpp
+++ b/lib/SPIRV/SPIRVLowerConstExpr.cpp
@@ -123,31 +123,51 @@ void SPIRVLowerConstExpr::visit(Module *M) {
     auto FBegin = I.begin();
     while (!WorkList.empty()) {
       auto II = WorkList.front();
+
+      auto LowerOp = [&II, &FBegin, &I](ConstantExpr *CE) {
+        SPIRVDBG(dbgs() << "[lowerConstantExpressions] " << *CE;)
+        auto ReplInst = CE->getAsInstruction();
+        auto InsPoint = II->getParent() == &*FBegin ? II : &FBegin->back();
+        ReplInst->insertBefore(InsPoint);
+        SPIRVDBG(dbgs() << " -> " << *ReplInst << '\n';)
+        std::vector<Instruction *> Users;
+        // Do not replace use during iteration of use. Do it in another loop
+        for (auto U : CE->users()) {
+          SPIRVDBG(dbgs() << "[lowerConstantExpressions] Use: " << *U << '\n';)
+          if (auto InstUser = dyn_cast<Instruction>(U)) {
+            // Only replace users in scope of current function
+            if (InstUser->getParent()->getParent() == &I)
+              Users.push_back(InstUser);
+          }
+        }
+        for (auto &User : Users)
+          User->replaceUsesOfWith(CE, ReplInst);
+        return ReplInst;
+      };
+
       WorkList.pop_front();
       for (unsigned OI = 0, OE = II->getNumOperands(); OI != OE; ++OI) {
         auto Op = II->getOperand(OI);
-
-        if (auto CE = dyn_cast<ConstantExpr>(Op)) {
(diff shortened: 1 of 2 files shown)


test/constexpr_vector.ll
```diff
@@ -0,0 +1,66 @@
+; RUN: llvm-as < %s | llvm-spirv -s | llvm-dis | FileCheck %s --check-prefix=CHECK-LLVM
+
```
AlexeySotkin, line 2: Please add also checks for SPIR-V representation.
Other then that LGTM. Thanks!

lib/SPIRV/SPIRVLowerConstExpr.cpp
```diff
@@ -124,31 +124,51 @@ void SPIRVLowerConstExpr::visit(Module *M) {
...
+          unsigned Idx = 0;
+          for (auto V : ReplList)
+            Repl = InsertElementInst::Create(
+                (Repl ? Repl : UndefValue::get(Vec->getType())), V,
+                ConstantInt::get(Type::getInt32Ty(M->getContext()), Idx++), "",
+                II);
+          II->replaceUsesOfWith(Op, Repl);
+          WorkList.splice(WorkList.begin(), ReplList);
+        } else if (auto CE = dyn_cast<ConstantExpr>(Op))
+          WorkList.push_front(lowerOp(CE));
```
AlexeySachkov, line 171:
```suggestion
          WorkList.push_front(LowerOp(CE));
```

AlexeySotkin:
I have looked into this more thoroughly. It seems that we get invalid SPIR-V module. @zuban32 could you add spirv-val to the test?

zuban32:
> I have looked into this more thoroughly. It seems that we get invalid SPIR-V module. @zuban32 could you add spirv-val to the test?

Done. But locally it seems to be fine

test/constexpr_vector.ll
```diff
@@ -0,0 +1,121 @@
...
+; CHECK-LLVM-NEXT:   %33 = insertelement <16 x i8> %32, i8 %17, i32 13
+; CHECK-LLVM-NEXT:   %34 = insertelement <16 x i8> %33, i8 %18, i32 14
+; CHECK-LLVM-NEXT:   %35 = insertelement <16 x i8> %34, i8 %19, i32 15
+; CHECK-LLVM-NEXT:   store <16 x i8> %35, <16 x i8>* %Funcs, align 16
+; CHECK-LLVM-NEXT:   ret void
+; CHECK-LLVM-NEXT: }
+
+; RUN: llvm-as %s -o %t.bc
+; RUN: llvm-spirv %t.bc -spirv-text -o %t
+; RUN: FileCheck --check-prefix=CHECK-SPIRV < %t %s
```
AlexeySotkin, line 48:
```suggestion
; RUN: llvm-spirv %t.bc -spirv-text -o - | FileCheck %s --check-prefix=CHECK-SPIRV
```

test/constexpr_vector.ll
```diff
@@ -0,0 +1,121 @@
...
+; RUN: llvm-spirv %t.bc -o %t.spv
+; RUN: spirv-val %s.spv
```
AlexeySotkin, line 50:
```suggestion
; RUN: spirv-val %t.spv
```

zuban32:
> The right fix would be to add `OpFunctionPointerINTEL` and use its result as operand for `ConvertPtrToU`. That would bring in the SPV_INTEL_function_pointers extension which `spirv-val` doesn't "know" about. So I think adding `spirv-val` doesn't make much sense right now.
> And fix for missing `OpFunctionPointerINTEL`, IMO, can be done in a separate patch/PR.

Agree. And as it's a different story, I suggest I remove the whole SPIRV check from the test for now, and then put it back in a corresponding FuncPointer fix

AlexeySotkin:
@zuban32 could you rebase your branch on top of the master branch? The issue failing build in travis.ci seems to be fixed by #513

PR #503, opened 2020-04-22
[NFC] Refactor uses of Type::getPointerElementType

AlexeySachkov:
Corresponding method is going to be removed and need to be replaced with
PointerType::getPointerElementType. See https://reviews.llvm.org/D15689

### DIFF
--- a/lib/SPIRV/SPIRVUtil.cpp
+++ b/lib/SPIRV/SPIRVUtil.cpp
@@ -579,10 +579,10 @@ bool containsUnsignedAtomicType(StringRef Name) {
 }
 
 bool isFunctionPointerType(Type *T) {
-  if (isa<PointerType>(T) && isa<FunctionType>(T->getPointerElementType())) {
-    return true;
-  }
-  return false;
+  auto *PtrTy = dyn_cast<PointerType>(T);
+  if (!T)
+    return false;
+  return isa<FunctionType>(PtrTy->getPointerElementType());
 }
 
 bool hasFunctionPointerArg(Function *F, Function::arg_iterator &AI) {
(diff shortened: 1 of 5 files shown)


AlexeySachkov:
Note: this change doesn't refactor all "Type::getPointerElementType" usages, but I guess it replaces at least half of them with new API

svenvh:
> Corresponding method is going to be removed and need to be replaced with
> PointerType::getPointerElementType. See https://reviews.llvm.org/D15689

Not sure if that is the LLVM review you wanted to link? It's an abandoned review from 2015.

lib/SPIRV/SPIRVUtil.cpp
```diff
@@ -579,10 +579,10 @@ bool containsUnsignedAtomicType(StringRef Name) {
...
 
 bool isFunctionPointerType(Type *T) {
-  if (isa<PointerType>(T) && isa<FunctionType>(T->getPointerElementType())) {
-    return true;
-  }
-  return false;
+  auto *PtrTy = dyn_cast<PointerType>(T);
+  if (!T)
+    return false;
+  return isa<FunctionType>(PtrTy->getPointerElementType());
```
mlychkov, line 585:
```suggestion
  if (auto *PtrTy = dyn_cast<PointerType>(T))
    return isa<FunctionType>(PtrTy->getPointerElementType());
  return false;
```

AlexeySachkov:
> > Corresponding method is going to be removed and need to be replaced with
> > PointerType::getPointerElementType. See https://reviews.llvm.org/D15689
> 
> Not sure if that is the LLVM review you wanted to link? It's an abandoned review from 2015.

Honestly, I'm not sure why, but I was under impression that such change is going to happen soon. Apparently I'm not quite right and the whole PR looks a bit silly :) 

There is definitely some ongoing work on opaque pointer types in the community, but it hasn't yet gotten to refactoring of `Type` and `PointerType` classes - I will close this for now as it doesn't make a lot of sense

PR #505, opened 2020-04-22
LLVM to SPIR-V Support for cl_khr_extended_subgroups

bashbaug:
These changes add support for the new "extended subgroup" instructions added to SPIR-V 1.3.

Changes include:

* LLVM to SPIR-V translation for all instructions.
* Adding the correct capabilities and SPIR-V 1.3 support for the new instructions.
* Tests to ensure correct translation from LLVM to SPIR-V.

Still TODO (not part of this PR):

* Clang headers with the new functions.
* SPIR-V to LLVM translation for all instructions.

Many thanks to @aratajew for the help with these changes!


include/LLVMSPIRVOpts.h
```diff
@@ -51,10 +51,12 @@ enum class VersionNumber : uint32_t {
   // Instruction
   SPIRV_1_0 = 0x00010000,
   SPIRV_1_1 = 0x00010100,
-  // TODO: populate this enum with the latest versions (up to 1.4) once
-  // translator get support of correponding features
+  SPIRV_1_2 = 0x00010200,
+  SPIRV_1_3 = 0x00010300,
```
AlexeySachkov, line 55: I'm not sure that this is critical, but there is some inconsistency: we do have SPIR-V 1.2 mentioned here, but we cannot limit maximum allowed SPIR-V version to it:

https://github.com/KhronosGroup/SPIRV-LLVM-Translator/blob/b656125fb2ad0295f910630f1b3efcdab4f88dfe/tools/llvm-spirv/llvm-spirv.cpp#L101-L106

test/transcoding/sub_group_ballot.ll
```diff
@@ -0,0 +1,1194 @@
...
+; DISABLED: llvm-dis < %t.rev.bc | FileCheck %s --check-prefix=CHECK-LLVM
+
+; ModuleID = 'subgroup_ballot.cl'
+source_filename = "subgroup_ballot.cl"
+target datalayout = "e-i64:64-v16:16-v24:32-v32:32-v48:64-v96:128-v192:256-v256:256-v512:512-v1024:1024"
+target triple = "spir64"
+
+; CHECK-SPIRV-DAG: {{[0-9]*}} Capability GroupNonUniformBallot
+
+; CHECK-SPIRV-DAG: Decorate [[eqMask:[0-9]+]] BuiltIn 4416
```
AlexeySachkov, line 174: Minor: there is new FileCheck feature called [Numeric substitution block](https://llvm.org/docs/CommandGuide/FileCheck.html#filecheck-numeric-substitution-blocks): it potentially could simplify the test a bit

aratajew: How would you like to use it in this case? I think we don't have any complex numeric relations here.

AlexeySachkov: @aratajew, ok, "simplify" is probably should be "reduce amount of symbols" :)

I mean, it will allow to remove all of these `[0-9]+` regex since in most cases we are looking for plain numbers here. This is definitely very minor

lib/SPIRV/OCL20ToSPIRV.cpp
```diff
@@ -948,48 +948,81 @@ void OCL20ToSPIRV::visitCallGroupBuiltin(CallInst *CI,
...
+              FuncName.contains("clustered_") ? "non_uniform_" : "";
+          StringRef LogicalOp = FuncName.contains("logical_") ? "logical_" : "";
+          StringRef GroupOp = StringSwitch<StringRef>(FuncName)
+                                  .Case("ballot_bit_count", "add")
+                                  .Case("ballot_inclusive_scan", "add")
+                                  .Case("ballot_exclusive_scan", "add")
+                                  .Default(FuncName.take_back(
+                                      3)); // assumes op is three characters
+          if (GroupOp.startswith("_"))
+            GroupOp = GroupOp.take_back(2); // when op is two characters
```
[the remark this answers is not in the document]
bashbaug: Fixed - thanks!

include/LLVMSPIRVOpts.h
bashbaug: It's not entirely clear to me what the behavior should be if a kernel uses these extended subgroup opcodes, which require SPIR-V 1.3, and the maximum SPIR-V version is set to SPIR-V 1.2 or earlier.

Should this:

* Fail gracefully?
* Emit SPIR-V 1.3 anyhow?
* Assert, because this is an invalid request?

As best I can tell right now, the maximum SPIR-V version is only used when deciding whether to emit some optional features that require a newer SPIR-V version.

AlexeySachkov: I think here the best thing would be "fail gracefully" or "assert, because this is an invalid request", see corresponding [docs](https://github.com/KhronosGroup/SPIRV-LLVM-Translator/blob/master/docs/SPIRVVersionsAndExtensionsHandling.rst#spir-v-generation-step):

> If the translator encounters something that cannot be represented by set of allowed SPIR-V versions (which might contain only one version), it does one of the following things:
>
>- ignores LLVM IR entity in the input file.
>
>   For example, dereferencable LLVM IR attribute can be ignored if it is not allowed to generate SPIR-V 1.1 and higher.
>
>- tries to represent LLVM IR entity with allowed instructions.
>
>   For example, `OpPtrEqual` can be used if SPIR-V 1.4 is not allowed and can be emulated via `OpConvertPtrToU` + `OpIEqual` sequence.
>
>- emits error if LLVM IR entity cannot be ignored and cannot be emulated using available instructions.
>
>   For example, if global constructors/destructors (represented by `@llvm.global_ctors`/`@llvm.global_dtors`) are present in a module then the translator should emit error if it cannot use SPIR-V 1.1 and higher where `Initializer` and `Finalizer` execution modes are described.

For now, I think I'm okay with some `TODO` about enabling proper version checks for new instructions

test/transcoding/sub_group_ballot.ll
aratajew: Could we please treat this as a future opportunity to refactor?  Unfortunately, I don't have a bandwidth now to do this.

AlexeySachkov: @aratajew, this is for sure minor comment: feel free to even ignore it since it won't bring any actual profits except removal of a few symbols from each regexp

aratajew:
> As agreed on the 2020/05/22 OpenCL tooling call, I've had a look at the code. I couldn't find anything obviouly wrong after a quick review. I do, however, have a couple of questions on the tests:
> 
> * Should these be OpenCL C tests instead? They would be more readable and would start failing if the mangling changes in Clang.
> * If we keep them as IR tests, should the IR be shortened a bit (unncessary instructions and metadata notably)?

Clang changes are still in code review. We wanted to make the translator changes independent, that's why the tests are in IR form. Once clang changes gets committed, we're going to convert the tests to OpenCL C form, that's why the source OpenCL C kernel is inlined as a comment to each test. Do you think it's necessary to refactor IR tests, taking into account that it's only a temporary solution?

kpet:
@aratajew Works for me! I don't think refactorig the tests is necessary then.

PR #512, opened 2020-04-27
Add support for SPV_INTEL_optimization_hints extension

MrSidims:
This extension adds representation of llvm.assume and llvm.expect
intrinsics in SPIR-V that can be used as helpers in optimizations.

Initial spec: https://github.com/intel/llvm/pull/1593/files#diff-333c6846d675d58988f3711e40258bda


MrSidims:
Tagging @AlexeySachkov  @AlexeySotkin

lib/SPIRV/SPIRVWriter.cpp
```diff
@@ -1634,6 +1634,19 @@ SPIRVValue *LLVMToSPIRV::transIntrinsicInst(IntrinsicInst *II,
   };
 
   switch (II->getIntrinsicID()) {
+  case Intrinsic::assume: {
+    // llvm.assume translation is currently supported only within
+    // SPV_INTEL_optimization_hints extension, ignore it otherwise, since it's
+    // an optimization hint
+    if (BM->isAllowedToUseExtension(
+          ExtensionID::SPV_INTEL_optimization_hints)) {
+      BM->addCapability(CapabilityOptimizationHintsINTEL);
```
AlexeySachkov, line 1643: No need to manually insert capability, just register it here: https://github.com/KhronosGroup/SPIRV-LLVM-Translator/blob/10cdc73aca7ece07d7428676a84f6e15a06a7c58/lib/SPIRV/libSPIRV/SPIRVEnum.h#L313-L315

test/llvm.assume.ll
```diff
@@ -0,0 +1,110 @@
+; RUN: llvm-as %s -o %t.bc
+; RUN: llvm-spirv %t.bc --spirv-ext=+SPV_INTEL_optimization_hints -spirv-text -o %t.spt
+; RUN: FileCheck < %t.spt %s --check-prefix=CHECK-SPIRV
+; RUN: llvm-spirv %t.bc --spirv-ext=+SPV_INTEL_optimization_hints -o %t.spv
```
AlexeySachkov, line 4: By creating `.spt` file from `.spv`, you can avoid repeating `--spirv-ext` option:

```
; RUN: llvm-spirv %t.bc --spirv-ext=+SPV_INTEL_optimization_hints -o %t.spv
; RUN: llvm-spirv %t.spv -to-text -o %t.spt
; RUN: FileCheck < %t.spt %s --check-prefix=CHECK-SPIRV
```

test/llvm.assume.ll
```diff
@@ -0,0 +1,110 @@
...
+; RUN: llvm-spirv -r %t.spv -o %t.bc
+; RUN: llvm-dis < %t.bc | FileCheck %s --check-prefix=CHECK-LLVM
+
+; RUN: llvm-spirv %t.bc -spirv-text -o %t.spt
+; RUN: FileCheck < %t.spt %s --check-prefix=CHECK-SPIRV-NO-EXT
+; RUN: llvm-spirv %t.bc -o %t.spv
```
AlexeySachkov, line 10:
```suggestion
; RUN: llvm-spirv %t.bc -o %t.spv
; RUN: spirv-val %t.spv
```

Without extension, resulting SPIR-V file should be correct in terms of validation rules, make sense to check that

test/llvm.assume.ll
```diff
@@ -0,0 +1,110 @@
...
+; RUN: llvm-spirv -r %t.spv -o %t.bc
+; RUN: llvm-dis < %t.bc | FileCheck %s --check-prefix=CHECK-LLVM-NO-EXT
+
+; CHECK-SPIRV: 2 Capability OptimizationHintsINTEL
```
AlexeySachkov, line 14:
```suggestion
; CHECK-SPIRV: Capability OptimizationHintsINTEL
```

No need to check number of arguments

test/llvm.assume.ll
```diff
@@ -0,0 +1,110 @@
+; RUN: llvm-as %s -o %t.bc
```
AlexeySachkov, line 1: This file should be in `transcoding` subdirectory, because it performs translation in both ways

test/llvm.expect.ll
```diff
@@ -0,0 +1,195 @@
...
+; RUN: llvm-spirv %t.bc -spirv-text -o %t.spt
+; RUN: FileCheck < %t.spt %s --check-prefix=CHECK-SPIRV-NO-EXT
+; RUN: llvm-spirv %t.bc -o %t.spv
+; RUN: llvm-spirv -r %t.spv -o %t.bc
+; RUN: llvm-dis < %t.bc | FileCheck %s --check-prefix=CHECK-LLVM-NO-EXT
+
+; CHECK-SPIRV: 2 Capability OptimizationHintsINTEL
+; CHECK-SPIRV: 9 Extension "SPV_INTEL_optimization_hints"
+; CHECK-SPIRV: 5 Function
+; CHECK-SPIRV: 5 ExpectINTEL {{[0-9]+}} [[RES1:[0-9]+]] {{[0-9]+}} {{[0-9]+}}
```
AlexeySachkov, line 17: I think you should also check the rest of `ExpectINTEL` arguments, i.e. you should at least check expected value passed to it

StuartDBrady:
I don't see why this extension is Intel-specific. Please can we discuss this before merging? Thanks.

kpet:
@StuartDBrady Have a look at the SPIR mailing list. There was a proposal to standardise this and another INTEL extension. I expressed our interest.

StuartDBrady:
@kpet I saw it, but I didn't want to refer to that publicly. :)

MrSidims:
@StuartDBrady @kpet  we have no objections of making the extension to be a KHR or even a part of the core spec. The only reason, why it's Intel specific is to faster the process of development.
As [spec](https://github.com/intel/llvm/blob/056ae4031e549ff2599f8c4ebbca3de10ef5d86f/sycl/doc/extensions/SPIRV/SPV_INTEL_optimization_hints.asciidoc) states:

> This is a preview extension specification, intended to provide early access to a
> feature for review and community feedback.

kpet:
@MrSidims We completely get that :)

lib/SPIRV/SPIRVReader.cpp
```diff
@@ -2186,6 +2186,18 @@ Value *SPIRVToLLVM::transValueWithoutDecoration(SPIRVValue *BV, Function *F,
...
+  case OpExpectINTEL: {
+    IRBuilder<> Builder(BB);
+    SPIRVExpectINTELInstBase *BC = static_cast<SPIRVExpectINTELInstBase *>(BV);
+    Type *RetTy = transType(BC->getType());
+    Value *Val = transValue(BC->getOperand(0), F, BB);
+    Value *ExpVal = transValue(BC->getOperand(1), F, BB);
+    auto IID = Intrinsic::expect;
+    Value *Args[] = {Val, ExpVal};
+    auto *IntrinsicCall = Builder.CreateIntrinsic(IID, RetTy, Args);
+    return mapValue(BV, IntrinsicCall);
```
AlexeySotkin, line 2198: Will it work this way?
```suggestion
    return mapValue(BV, Builder.CreateIntrinsic(Intrinsic::expect, RetTy, {Val, ExpVal}));
```

lib/SPIRV/libSPIRV/SPIRVOpCodeEnum.h
```diff
@@ -438,3 +438,4 @@ _SPIRV_OP(LoopControlINTEL, 5887)
 _SPIRV_OP(ReadPipeBlockingINTEL, 5946)
 _SPIRV_OP(WritePipeBlockingINTEL, 5947)
 _SPIRV_OP(FPGARegINTEL, 5949)
+_SPIRV_OP(ExpectINTEL, 9997)
```
mlychkov, line 441: What does 9997 means?

AlexeySotkin: I guess, it means 5631 :)

MrSidims: Oh, that an artifact from a "prior values reservation" age. Thanks!

lib/SPIRV/SPIRVReader.cpp
MrSidims: It shall. I just had a desire to have some granularity here (otherwise I would inline Val and ExpVal processing here as well). But yeah, your suggestion looks nicer.

StuartDBrady:
Note that when translating from LLVM IR, to SPIR-V, and back to LLVM IR, the positioning of any call to the **llvm.assume** intrinsic with respect to control flow should be unchanged. This can be achieved by generating an **OpCopyObject** instruction upon which to apply the decoration. However, it may be better to have an instruction for **assume**, too. (It is not clear to me why **expect** uses an instruction but **assume** uses a decoration.)

If the choice is made to generate **OpCopyObject**, then there should be a transcoding test to ensure this works correctly. If the choice is made to add an additional instruction for **assume**, there is no need for an extra (or more complex) test.

lib/SPIRV/SPIRVWriter.cpp
```diff
@@ -1634,6 +1634,18 @@ SPIRVValue *LLVMToSPIRV::transIntrinsicInst(IntrinsicInst *II,
...
 
   switch (II->getIntrinsicID()) {
+  case Intrinsic::assume: {
+    // llvm.assume translation is currently supported only within
+    // SPV_INTEL_optimization_hints extension, ignore it otherwise, since it's
+    // an optimization hint
+    if (BM->isAllowedToUseExtension(
+            ExtensionID::SPV_INTEL_optimization_hints)) {
+      SPIRVValue *SV = transValue(II->getOperand(0), BB);
+      SV->addDecorate(DecorationAssumeTrueINTEL);
```
StuartDBrady, line 1644: This seems to be adding the decoration to the _input_ of the intrinsic, which does not seem correct with respect to control flow.

MrSidims:
> . (It is not clear to me why **expect** uses an instruction but **assume** uses a decoration.)

My original thoughts were to make both of them to be decorations because it’s easy to drop decorations from a SPIR-V module without making the module invalid, which seems about right for the instructions that can be ignored. But then I dropped this idea for llvm.expect because of:
1.	I found type tracking during translation to be not obvious (someone can disagree here with me);
2.	The same decoration shall be attached to several IDs. It’s possible to do with OpGroupDecorate, but it was deprecated.

lib/SPIRV/SPIRVWriter.cpp
MrSidims: Remade it to be an instruction.

lib/SPIRV/SPIRVWriter.cpp
```diff
@@ -1634,6 +1634,17 @@ SPIRVValue *LLVMToSPIRV::transIntrinsicInst(IntrinsicInst *II,
   };
 
   switch (II->getIntrinsicID()) {
+  case Intrinsic::assume: {
+    // llvm.assume translation is currently supported only within
+    // SPV_INTEL_optimization_hints extension, ignore it otherwise, since it's
+    // an optimization hint
+    if (BM->isAllowedToUseExtension(
+            ExtensionID::SPV_INTEL_optimization_hints)) {
+      SPIRVValue *Condition = transValue(II->getArgOperand(0), BB);
```
StuartDBrady, line 1643: Do you need to add the CapabilityOptimizationHintsINTEL capability here, also?

MrSidims: I'd say, it's not required (the appropriate test has passed), since I've already added:
```
  SPIRVCapVec getRequiredCapability() const override {
    return getVec(CapabilityOptimizationHintsINTEL);
  }
```
in the instruction definition. So I'll remove it for llvm.expect and see what will happen (sorry for debugging throught review, my working machine went down).

StuartDBrady:
Thanks, @MrSidims. The recent changes look good IMO, but I'll defer to others to for the remainder of the review.

lib/SPIRV/SPIRVWriter.cpp
StuartDBrady: Okay, LGTM!

(BTW, I sympathize regarding the lockdown: the exact problems are different for everyone, but almost nobody is unaffected by all of this.)

MrSidims:
@StuartDBrady thank you very much for the feedback!

@AlexeySachkov @AlexeySotkin ping

Issue #514, opened 2020-04-29
MangledAtomicTypeNamePrefix option - is this used by anybody?


aratajew:
I'd like to ask if anybody is using `MangledAtomicTypeNamePrefix `translator option?

It is handled in this part of code:
https://github.com/KhronosGroup/SPIRV-LLVM-Translator/blob/master/lib/SPIRV/SPIRVToOCL.cpp#L365

`void SPIRVToOCL::translateMangledAtomicTypeName()` function is logically meaningless and it has a lot of bugs, example:
```
auto Loc = MangledName.find(kOCLBuiltinName::AtomPrefix);
Loc = MangledName.find(kMangledName::AtomicPrefixInternal, Loc);
MangledName.replace(Loc, strlen(kMangledName::AtomicPrefixInternal),
                        MangledAtomicTypeNamePrefix);
```
Loc stored two times. No check if Loc isn't string::npos.

This function makes translator crash on such a simple OpenCL C kernel:
```
void __attribute__((overloadable)) atom_custom_function(){}
void kernel test()
{
	atom_custom_function();
}
```

My questions are:
1. Is anybody using it?
      a)  If not, I'll totaly remove this code.
      b)  If yes, please let me know, I'll just fix this code to avoid crashes.

Thanks,
Andrzej

aratajew:
@yxsamliu I think, you an author of this code. Is it still in use?

aratajew:
@svenvh , @AlexeySachkov , @AlexeySotkin guys, if any of you know some customer of this code, please, let me know.

svenvh:
> @svenvh , @AlexeySachkov , @AlexeySotkin guys, if any of you know some customer of this code, please, let me know.

No objections from my side to remove the option.

yxsamliu:
> @yxsamliu I think, you're an author of this code. Is it still in use?

I am no longer working on this project, so I will let others to decide. Thanks.

AlexeySachkov:
Fixed by #515

PR #515, opened 2020-04-30
Remove deprecated spirv-atomic-prefix option

aratajew:
For details about motivation of removing this code check: https://github.com/KhronosGroup/SPIRV-LLVM-Translator/issues/514

### DIFF
--- a/lib/SPIRV/SPIRVToOCL.cpp
+++ b/lib/SPIRV/SPIRVToOCL.cpp
@@ -47,11 +47,6 @@
 
 namespace SPIRV {
 
-static cl::opt<std::string>
-    MangledAtomicTypeNamePrefix("spirv-atomic-prefix",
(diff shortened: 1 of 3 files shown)


aratajew:
@svenvh could we please merge it? Also for the other branches.

svenvh:
> @svenvh could we please merge it? Also for the other branches.

I was waiting for @AlexeySachkov and @AlexeySotkin to confirm they are happy with the removal too (perhaps I should have tagged them).

bader:
Sorry, for the delay, guys. Russia is on public holidays. I hope Alexey's will be able to review tomorrow.

AlexeySachkov:
We are mostly using the translator via library interface and since this option is not available there - we couldn't rely on it and this PR should make no difference for us.

intel/llvm project is using the translator as binary, but we don't use this option in there.

@AlexeySotkin, please also confirm that removal is safe for us

PR #521, opened 2020-04-30
Enable strict rules to set ContractionOFF

asavonic:
This patch makes several improvements to enable correct setting for
ContractionOFF:

  1) Plain `fadd` or `fsub` instruction trigger ContractionOFF to
     avoid optimizations that may fold them to an FMA.

  2) `contract` flag is recognized on `fadd` and `fsub`, and it allows
     to suppress (1).

  3) Undefined (external) functions, function pointers and inline asm
     calls force ContractionOFF flag.

  4) Contraction information is now propagated up the call stack until
     an entry point.

### DIFF
--- a/lib/SPIRV/OCLUtil.cpp
+++ b/lib/SPIRV/OCLUtil.cpp
@@ -884,23 +884,36 @@ bool isKernelQueryBI(const StringRef MangledName) {
 // // #pragma OPENCL FP_CONTRACT OFF. Otherwise the current version of clang
 // would generate:
 // %0 = call float @llvm.fmuladd.f32(float %a, float %b, float %c)
-// TODO We need a more reliable mechanism to expres the FP_CONTRACT pragma
-// in LLVM IR. Fox example adding the 'contract' attribute to fp operations
-// by default (according the OpenCL spec fp contraction is enabled by default).
-void checkFpContract(BinaryOperator *B, SPIRVBasicBlock *BB) {
+bool isUnfusedMulAdd(BinaryOperator *B) {
   if (B->getOpcode() != Instruction::FAdd &&
       B->getOpcode() != Instruction::FSub)
-    return;
-  // Ok, this is fadd or fsub. Now check its operands.
-  for (auto *Op : B->operand_values()) {
-    if (auto *I = dyn_cast<Instruction>(Op)) {
-      if (I->getOpcode() == Instruction::FMul) {
-        SPIRVFunction *BF = BB->getParent();
-        BF->setUncontractedFMulAddFound();
-        break;
-      }
-    }
+    return false;
+
+  if (B->hasAllowContract()) {
+    // If this fadd or fsub itself has a contract flag, the operation can be
+    // contracted regardless of the operands.
+    return false;
   }
+
+  // Otherwise, we cannot easily tell if the operation can be a candidate for
+  // contraction or not. Consider the following cases:
+  //
+  //   %mul = alloca float
+  //   %t1 = fmul float %a, %b
+  //   store float* %mul, float %t
+  //   %t2 = load %mul
+  //   %r = fadd float %t2, %c
+  //
+  // LLVM IR does not allow %r to be contracted. However, after an optimization
+  // it becomes a candidate for contraction if ContractionOFF is not set in
+  // SPIR-V:
+  //
+  //   %t1 = fmul float %a, %b
+  //   %r = fadd float %t1, %c
+  //
+  // To be on a safe side, we disallow everything that is even remotely similar
+  // to fmul + fadd.
+  return true;
 }
 
 std::string getIntelSubgroupBlockDataPostfix(unsigned ElementBitSize,
--- a/lib/SPIRV/OCLUtil.cpp
+++ b/lib/SPIRV/OCLUtil.cpp
@@ -893,8 +893,8 @@ bool isKernelQueryBI(const StringRef MangledName) {
 //   %mul = fmul contract float %a, %b
 //   %add = fadd contract float %mul, %c
 //
-// Note that optimizations that may form an unfused fmuladd, so this check is
-// quite restrictive (see the comment below).
+// Note that optimizations may form an unfused fmuladd from fadd+load or
+// fadd+call, so this check is quite restrictive (see the comment below).
 //
 bool isUnfusedMulAdd(BinaryOperator *B) {
   if (B->getOpcode() != Instruction::FAdd &&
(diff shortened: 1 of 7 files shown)


lib/SPIRV/SPIRVWriter.cpp
```diff
@@ -2150,6 +2184,67 @@ void LLVMToSPIRV::mutateFuncArgType(
...
+    if (Function *F = dyn_cast<Function>(U)) {
+      if (!joinFPContract(F, FPC)) {
+        // FP contract was not updated - no need to propagate
+        // This also terminates a recursion (if any).
+        if (EnableLogger) {
+          SPIRVDBG(dbgs() << "[fp-contract] already disabled " << F->getName()
+                          << '\n');
+        }
+        continue;
+      }
```
AlexeySachkov, line 2223: If I understand correctly, if there are several calls to the same function within another function, then this message will be logged several times, right?

This is not critical, I guess, but if `joinFPContract` is "heavy", then it might be better to avoid adding the same function into the queue several times

lib/SPIRV/SPIRVWriter.cpp
```diff
@@ -2374,9 +2453,48 @@ bool LLVMToSPIRV::transExecutionMode() {
...
+
   return true;
 }
 
+void LLVMToSPIRV::transFPContract() {
+  for (Function &F : *M) {
+    SPIRVValue *TranslatedF = getTranslatedValue(&F);
+    if (!TranslatedF) {
+      continue;
+    }
```
AlexeySachkov, line 2467: I would assert that all functions were translated, otherwise, how can you be sure that you have gathered all information required for fp-contract handling?

Moreover, this function is called almost at the end of [`translate`](https://github.com/KhronosGroup/SPIRV-LLVM-Translator/pull/521/commits/8504e880a16f5d8794662aedac23ab90175d9a72#diff-3d6ba7d110dbca252605cea8fff24de7R2312)

lib/SPIRV/SPIRVWriter.cpp
```diff
@@ -2585,6 +2703,34 @@ LLVMToSPIRV::transLinkageType(const GlobalValue *GV) {
   return SPIRVLinkageTypeKind::LinkageTypeExport;
 }
 
+LLVMToSPIRV::FPContract LLVMToSPIRV::getFPContract(Function *F) {
+  auto It = FPContractMap.find(F);
+  if (It == FPContractMap.end()) {
+    return FPContract::UNDEF;
+  }
+  return It->second;
```
AlexeySachkov, line 2711: Honestly, I don't know what is default value for enumerations, but theoretically, the whole function body could be replaced by a call to [lookup method](https://llvm.org/doxygen/classllvm_1_1DenseMapBase.html#a0b2ca98dc28c61793ff5c90d23e5f14e)

lib/SPIRV/SPIRVWriter.cpp
asavonic: Intrinsics and SPIR-V built-in functions are not translated to SPIR-V functions, so they are excluded by this check.

> how can you be sure that you have gathered all information required for fp-contract handling?

Everything that gets to `transFunction` is handled. Here in `transFPContract` we only apply fp-contract mode based on the information we gathered in all `transFunction` calls.

lib/SPIRV/SPIRVWriter.cpp
asavonic: Default value should be zero (matches `FPContract::UNDEF`), so `lookup` can be used here.
Current code is more explicit though, so I prefer to keep it unless you suggest to do otherwise.

lib/SPIRV/OCLUtil.cpp
```diff
@@ -884,23 +884,36 @@ bool isKernelQueryBI(const StringRef MangledName) {
 // // #pragma OPENCL FP_CONTRACT OFF. Otherwise the current version of clang
 // would generate:
 // %0 = call float @llvm.fmuladd.f32(float %a, float %b, float %c)
-// TODO We need a more reliable mechanism to expres the FP_CONTRACT pragma
-// in LLVM IR. Fox example adding the 'contract' attribute to fp operations
-// by default (according the OpenCL spec fp contraction is enabled by default).
-void checkFpContract(BinaryOperator *B, SPIRVBasicBlock *BB) {
+bool isUnfusedMulAdd(BinaryOperator *B) {
```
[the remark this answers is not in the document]
asavonic: The comment gives an example of fused and unfused fmuladd. I'll add a note that there are more cases exist, and the check itself is restrictive.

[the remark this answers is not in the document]
asavonic: Done.

test/ContractionOff.ll
```diff
@@ -18,17 +18,41 @@
...
+; CHECK-OFF: ExecutionMode [[K3]] 31
+
+; CHECK-FAST-NOT: ExecutionMode [[K1]] 31
+; CHECK-FAST-NOT: ExecutionMode [[K2]] 31
+; CHECK-FAST-NOT: ExecutionMode [[K3]] 31
+
+; CHECK-ON: ExecutionMode [[K1]] 31
+; CHECK-ON-NOT: ExecutionMode [[K2]] 31
+; CHECK-ON: ExecutionMode [[K3]] 31
+; CHECK-ON-NOT: ExecutionMode [[K2]] 31
```
AlexeySotkin, line 55: Duplicate line

AlexeySotkin:
This implementation looks too strict to me, but I don't think we can get anything better with the current SPIR-V spec.

test/ContractionOff.ll
asavonic: Removed.

lib/SPIRV/SPIRVWriter.cpp
```diff
@@ -2607,6 +2730,34 @@ LLVMToSPIRV::transLinkageType(const GlobalValue *GV) {
...
+  case FPContract::ENABLED:
+    if (C == FPContract::DISABLED) {
+      Existing = C;
+      return true;
+    }
+    return false;
+  case FPContract::DISABLED:
+    return false;
+  }
+}
```
abhinavgaba, line 2759: This is causing a build fail with `-werror` currently with `gcc 7.3`:

    error: control reaches end of non-void function [-Werror=return-type]

asavonic: Thanks. Submitted https://github.com/KhronosGroup/SPIRV-LLVM-Translator/pull/548

PR #523, opened 2020-05-01
[NFC] Improve parsing of FPGA memory decorations

AGindinson:
- Use regex to avoid regular `find`s and substring
  assignments;
- Make variable names more readable.

### DIFF
--- a/lib/SPIRV/SPIRVWriter.cpp
+++ b/lib/SPIRV/SPIRVWriter.cpp
@@ -80,6 +80,7 @@
 #include <iostream>
 #include <memory>
 #include <queue>
+#include <regex>
 #include <set>
 #include <vector>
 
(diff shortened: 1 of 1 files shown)


AGindinson:
> I believe that a situation with missmatching number of '{' and '}' isn't handled with this approach as it was before;

How likely is such a situation?

> With the change readability of the code declines faster, than performance rises. If you believe, that this code isn't going to be touched in the future - this comment can be ignored.

My original reasoning was that the regex could actually make stuff more readable than a hand-crafted parsing algorithm. @vmaksimo, can we defer to you as a regular implementer of memory attributes? I'd be fine with simply leaving the renaming if that's largely preferred.

mlychkov:
> Regex changes are LGTM, but:
> Major:
> I believe that a situation with missmatching number of '{' and '}' isn't handled with this approach as it was before;
I think that the following isn't handled too: '}word:value{'. Validity check is '^({\w+(:\w+(,\w+)*)?})+$'.
> Minor (but that tends to be major):
> With the change readability of the code declines faster, than performance rises. If you believe, that this code isn't going to be touched in the future - this comment can be ignored.
As for me, readability is improved, but I would try to use shorter names for temp variables.

vmaksimo:
Renaming looks good to me.

> Regex changes are LGTM, but:
> Major:
> I believe that a situation with missmatching number of '{' and '}' isn't handled with this approach as it was before;

I believe that correct regex will be `{(\w+):?(\w+,\w+)?}`. If we use capturing groups, it's not needed to split strings by delimeter. But it's not easy-to-read expression, I would comment it properly if decided to put this feature here.

mlychkov:
> I believe that correct regex will be `{(\w+):?(\w+,\w+)?}`.

This regex treats '{abcd,ab}' as valid and '{ab:cd}' as invalid.
And also the following decoration is valid: `{merge:foo:bar}`.
So, the following one regex is correct: `{(\w+)(:\w+((,\w+)+|(:\w+)+)?)?}`.
It is not so readable unfortunately.

AGindinson:
> > I believe that correct regex will be `{(\w+):?(\w+,\w+)?}`.
> 
> This regex treats '{abcd,ab}' as valid and '{ab:cd}' as invalid.
> And also the following decoration is valid: `{merge:foo:bar}`.
> So, the following one regex is correct: `{(\w+)(:\w+((,\w+)+|(:\w+)+)?)?}`.
> It is not so readable unfortunately.

@mlychkov, @vmaksimo, given that LLVM IR is generally correct, would `{[:,\w+]}` be a decent compromise?

The reason I'd like to proceed with this change is to unify the solution for the existing attributes and the LSU controls' support -- using `regex_iterator`s could solve a lot of ugliness that arises otherwise.

lib/SPIRV/SPIRVWriter.cpp
```diff
@@ -1448,30 +1449,31 @@ std::vector<std::pair<Decoration, std::string>>
 tryParseIntelFPGAAnnotationString(StringRef AnnotatedCode) {
   std::vector<std::pair<Decoration, std::string>> Decorates;
 
-  size_t OpenBracketNum = AnnotatedCode.count('{');
```
mlychkov, line 1451: Are you sure that there cannot be other strings with different format except annotation strings with curly brackets?

AGindinson: I mean, I can't envision such a situation that could make it into `tryParseIntelFPGAAnnotationString(...)`. Am I missing some cases with var/ptr annotation strings?

vmaksimo: This function is designed to handle only strings with curly brackets. Incorrect (without curly brackets) string will be handled as UserSemantic decoration

mlychkov: Ok, anyway Alexey already approved these changes.

AlexeySotkin: I'm ready to merge this PR, but there is a conflict. Please resolve it.

AGindinson: @AlexeySotkin, done. I've also updated the patch to match curly-braced sequences (with a subsequent drop of these symbols). Please check if it's still good to go.

lib/SPIRV/SPIRVWriter.cpp
```diff
@@ -1569,30 +1570,32 @@ std::vector<std::pair<Decoration, std::string>>
...
-    StringRef Value;
+  // Intel FPGA decorations are separated into
+  // {word} OR {word:value,value,...} blocks
+  std::regex DecorationRegex("\\{[\\w:,]+\\}");
+  using RegexIterT = std::regex_iterator<StringRef::const_iterator>;
+  RegexIterT DecorationsIt(AnnotatedCode.begin(), AnnotatedCode.end(),
+                           DecorationRegex);
+  RegexIterT DecorationsEnd;
+  for (; DecorationsIt != DecorationsEnd; ++DecorationsIt) {
+    const std::string AnnotatedDecoration = DecorationsIt->str();
```
mlychkov, line 1581: Is this copying necessary?

AGindinson: The reason this assignment is needed at all is that `StringRef` has issues with temporary objects (which is [to be expected](https://llvm.org/doxygen/classllvm_1_1StringRef.html#Detailed_Description)).

AGindinson: I've reworked this to avoid the unnecessary copy of the `regex_iterator`'s underlying buffer.

PR #526, opened 2020-05-08
Refactor translation of memory scopes, orders and flags

AlexeySachkov:
We have a lot of duplicated code for handling memory scope, order and flags arguments of different OpenCL built-ins, like barriers, atomics and fences.

Refactored all common code into helper functions. (This should be an NFC change.)

Improved translation of `OpMemoryBarrier` by adding handling for non-const operands. (This is actually what I want to achieve, but while doing so, I decided to perform a refactoring.)

Note: I'm not going to add support for non-const memory scopes, orders and flags for each and every built-in and SPIR-V instruction in scope of this PR: this is just improvement of `OpMemoryBarrier` + refactoring.

### DIFF
--- a/lib/SPIRV/OCLUtil.h
+++ b/lib/SPIRV/OCLUtil.h
@@ -496,12 +496,13 @@ Instruction *
 getOrCreateSwitchFunc(StringRef MapName, Value *V,
                       const SPIRVMap<KeyTy, ValTy, Identifier> &Map,
                       bool IsReverse, Optional<int> DefaultCase,
-                      Instruction *InsertPoint, Module *M, int KeyMask = 0) {
+                      Instruction *InsertPoint, int KeyMask = 0) {
   static_assert(std::is_convertible<KeyTy, int>::value &&
                     std::is_convertible<ValTy, int>::value,
                 "Can map only integer values");
   Type *Ty = V->getType();
   assert(Ty && Ty->isIntegerTy() && "Can't map non-integer types");
+  Module *M = InsertPoint->getModule();
   Function *F = getOrCreateFunction(M, Ty, Ty, MapName);
   if (!F->empty()) // The switch function already exists. just call it.
     return addCallInst(M, MapName, Ty, V, nullptr, InsertPoint);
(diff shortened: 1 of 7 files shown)


AlexeySachkov:
@svenvh, @AlexeySotkin, I have decided to move tests refactoring into a separate PR.

This one was re-worked to separate code refactoring and functional changes - could you please take a look?

AlexeySachkov:
I think this can be considered as a follow-up (one of several) on [this comment](https://github.com/KhronosGroup/SPIRV-LLVM-Translator/pull/475#discussion_r398523505) from #475

lib/SPIRV/OCLUtil.h
```diff
@@ -543,6 +544,79 @@ getOrCreateSwitchFunc(StringRef MapName, Value *V,
   return addCallInst(M, MapName, Ty, V, nullptr, InsertPoint);
 }
 
+/// Performs conversion from OpenCL memory_scope into SPIR-V Scope.
+///
+/// Supports both constant and non-constant values. To handle the latter case,
+/// function with swtich..case statement will be insterted into module which
+/// \arg InsertBefore belongs to (in order to perform mapping at runtime)
+///
+/// \param [in] MemScope memory_scope value which needs to be translated
```
svenvh, line 553: Haven't tested locally, but isn't it a doxygen error if you refer to an argument name that is omitted from the declaration?

lib/SPIRV/OCLUtil.h
```diff
@@ -543,6 +544,79 @@ getOrCreateSwitchFunc(StringRef MapName, Value *V,
   return addCallInst(M, MapName, Ty, V, nullptr, InsertPoint);
 }
 
+/// Performs conversion from OpenCL memory_scope into SPIR-V Scope.
+///
+/// Supports both constant and non-constant values. To handle the latter case,
+/// function with swtich..case statement will be insterted into module which
```
svenvh, line 550:
```suggestion
/// function with switch..case statement will be inserted into module which
```

Same typos occur further down.

lib/SPIRV/OCLUtil.h
AlexeySachkov: I will check that, haven't though about this

AlexeySachkov: From http://www.doxygen.nl/manual/commands.html#cmdparam:

> The existence of the parameter is checked and a warning is given if the documentation of this (or any other) parameter is missing or not present in the function declaration or definition.

So, it looks like declaration requires parameter names to be specified, will fix it

AlexeySachkov: Should be fixed now

lib/SPIRV/OCLUtil.h
AlexeySachkov: Fixed, thanks!

PR #530, opened 2020-05-12
Refactor some tests for OpenCL built-ins

AlexeySachkov:
The main idea of this PR is to cleanup and improve some tests:
- remove some duplicates
- replace LLVM IR inputs with OpenCL C inputs

This should be considered as work in progress for now, because I have a few concerns, which I would like to discuss first:

IMO, OpenCL C tests are better in terms of end-to-end testing of the translator, however, they are more expensive in terms of execution time. Where do you think lies the border between this is better as OpenCL C test and this is better as LLVM IR test?

In order to try to reduce the time spent in testing, I used `-fdeclare-opencl-builtins` where possible; in other cases I forward-declared desired OpenCL built-ins, macro and types instead of including `opencl-c.h` header file.

Some thoughts on the testing of the translator: this is just to collect more inputs and opinions on the topic, probably we will prepare some testing plan for the translator?

Personally, I divide the translator into several smaller components:
- LLVM IR regularization/legalization/preprocessing: converts "some" LLVM IR into SPIR-V friendly representation
- LLVM IR -> SPIR-V translation, which is done *only* from SPIR-V friendly representation
- SPIR-V -> LLVM IR translation

Based on that, I would probably like to see up to 4 tests per SPIR-V instruction/OpenCL built-in:
a. test that checks that particular OpenCL built-in is correctly translated from LLVM IR generated by clang into SPIR-V friendly representation
b. test that checks that SPIR-V friendly representation is correctly translated into SPIR-V
c. test that checks that SPIR-V is correctly translated back into LLVM IR
d. "end-to-end" test, which performs full-stack tests: starting from OpenCL C down to LLVM IR through SPIR-V

If I understand correctly, (b), (c) and (d) is what we have now. (a) is missed.

Tagging @StuartDBrady, @bashbaug and @AnastasiaStulova


AlexeySachkov:
UPD: I think that (b) is actually also missed in our testing

AlexeySachkov:
@svenvh, @StuartDBrady, @AlexeySotkin, ping

svenvh (approved):
LGTM.  In particular I appreciate the approach you have taken to minimize testing time by declaring things instead of including the full opencl-c.h header.  We should be able to drop those declarations in the future when `-fdeclare-opencl-builtins` reaches parity with the header.

PR #533, opened 2020-05-18
Backport SPV_INTEL_function_pointers + fixes

zuban32:


AlexeySotkin:
@AlexeySachkov, you are right this patch is not needed on branches before release_9.x. For IGC we have this patch [here](https://github.com/intel/opencl-clang/blob/ocl-open-80/patches/spirv/0001-Update-LowerOpenCL-pass-to-handle-new-blocks-represn.patch).

zuban32:
That will require quite significant changes to that pass as it isn't able to process function pointers IR

PR #539, opened 2020-05-19
Move translation of OCL metadata to separate function

NikitaRudenkoIntel:
This refactor is needed for some of the future commits
The idea behind these changes was to abstract from specific types of metadata.
transMetadata() does everything is needed to be done with metadata.
transOCLMetadata(), which is called by transMetadata does everything is needed to be done with OCL metadata.

### DIFF
--- a/lib/SPIRV/SPIRVReader.cpp
+++ b/lib/SPIRV/SPIRVReader.cpp
@@ -2949,7 +2949,7 @@ bool SPIRVToLLVM::translate() {
     transFunction(BM->getFunction(I));
   }
 
-  if (!transKernelMetadata())
+  if (!transMetadata())
     return false;
   if (!transFPContractMetadata())
     return false;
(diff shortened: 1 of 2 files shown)


lib/SPIRV/SPIRVReader.cpp
```diff
@@ -3284,89 +3284,17 @@ static bool transKernelArgTypeMedataFromString(LLVMContext *Ctx,
...
 }
 
-bool SPIRVToLLVM::transKernelMetadata() {
+bool SPIRVToLLVM::transMetadata() {
   for (unsigned I = 0, E = BM->getNumFunctions(); I != E; ++I) {
     SPIRVFunction *BF = BM->getFunction(I);
     Function *F = static_cast<Function *>(getTranslatedValue(BF));
     assert(F && "Invalid translated function");
+
+    transOCLMetadata(BF);
```
AlexeySachkov, line 3293: I suggest to move this call below the check for `spir` calling convention

lib/SPIRV/SPIRVReader.cpp
```diff
@@ -3415,6 +3343,89 @@ bool SPIRVToLLVM::transKernelMetadata() {
   return true;
 }
 
+bool SPIRVToLLVM::transOCLMetadata(SPIRVFunction *BF) {
+  Function *F = static_cast<Function *>(getTranslatedValue(BF));
+  assert(F && "Invalid translated function");
+  if (F->getCallingConv() != CallingConv::SPIR_KERNEL)
```
AlexeySachkov, line 3349: With the suggestion above, you can replace this check with an assertion

lib/SPIRV/SPIRVReader.cpp
NikitaRudenkoIntel: I think it is better to leave this function before the check. A couple of saved instructions don't worth breaking encapsulation

svenvh:
> Overall, LGTM
> 
> @svenvh, any feedback?

LGTM.

PR #540, opened 2020-05-19
Add infrastructure for SPV_KHR_float_controls extension

NikitaRudenkoIntel:


lib/SPIRV/SPIRVWriter.cpp
```diff
@@ -2391,6 +2391,17 @@ bool LLVMToSPIRV::transExecutionMode() {
           BM->addCapability(CapabilityFPGAKernelAttributesINTEL);
         }
       } break;
+
+      case spv::ExecutionModeDenormPreserve:
+      case spv::ExecutionModeDenormFlushToZero:
+      case spv::ExecutionModeSignedZeroInfNanPreserve:
+      case spv::ExecutionModeRoundingModeRTE:
+      case spv::ExecutionModeRoundingModeRTZ:{
+          unsigned TargetWidth;
```
AlexeySachkov, line 2400: I think this should be guarded by an if about whether it is allowed to use corresponding extension or not.

By default, the translator is expected to not use any extensions and with this code, it won't be true. Moreover, with #537, this should trigger an assertion

lib/SPIRV/SPIRVWriter.cpp
```diff
@@ -2391,6 +2391,17 @@ bool LLVMToSPIRV::transExecutionMode() {
...
+          N.get(TargetWidth);
+          BF->addExecutionMode(BM->add(new SPIRVExecutionMode(
```
AlexeySachkov, line 2402: I don't see corresponding changes for the `SPIRVReader`, which means that this execution mode is going to be lost during reverse translation: are there plans to add SPIR-V -> LLVM IR support for this extension?

NikitaRudenkoIntel: The idea behind this commit is to introduce infrastructure, without any Reader/Writer part.
This particular change is also a part of infrastructure - before this piece of code will work, somebody has to add those Exec Modes in PreprocessMetadata.cpp
Until that, reader or writer will not even touch these exec modes: they just do not know from which structure in LLVM IR translate to ExecutionModeRoundingModeRTE, neither what to do with this mode in reader part.

lib/SPIRV/SPIRVWriter.cpp

NikitaRudenkoIntel:
The idea behind this patch is to separate introduction of SPV_KHR_float_controls extension from the particular implementation.
Thus, there is not much to test here: this commit basically extends existing enums, without providing any reader/writer parts.
On the other hand, it is the implementation, which will convert from LLVM IR to SPIRV and vice versa should (and will) introduce the testing

AlexeySachkov:
> Thus, there is not much to test here: this commit basically extends existing enums, without providing any reader/writer parts.

But you have changed the writer part: it is now actually able to insert new decorations into SPIR-V file

NikitaRudenkoIntel:
I added the test, and moved commit with capabilities check here

lib/SPIRV/libSPIRV/SPIRVEntry.h
```diff
@@ -665,19 +665,38 @@ class SPIRVExecutionMode : public SPIRVAnnotation<OpExecutionMode> {
...
 
 public:
   void addExecutionMode(SPIRVExecutionMode *ExecMode) {
-    ExecModes[ExecMode->getExecutionMode()] = ExecMode;
+    SPIRVExecutionModeKind EMK = ExecMode->getExecutionMode();
+    // There should not be more than 1 execution mode kind except the ones
+    // mentioned in SPV_KHR_float_controls.
+#ifndef NDEBUG
+    bool IsValid =
+        (ExecModes.count(EMK) == 0 || EMK == ExecutionModeDenormPreserve ||
```
AlexeySachkov, line 681: From the [extension spec](http://htmlpreview.github.io/?https://github.com/KhronosGroup/SPIRV-Registry/blob/master/extensions/KHR/SPV_KHR_float_controls.html):

> An OpEntryPoint can set at most one of the DenormFlushToZero, DenormPreserve Execution Modes for the given Floating-Point Type Bit Width.
>
> An OpEntryPoint can set at most one of the RoundingModeRTE, RoundingModeRTZ Execution Modes for the given Floating-Point Type Bit Width.

If I understand correctly, there are a few cases missing in this validation: for example, there is no check that `ExecutionModeRoundingModeRTE` is set together with `ExecutionModeRoundingModeRTZ` for the same `Target Width`.

This is probably a not critical comment since this code is anyway under `#ifndef`, but it worth adding a `TODO`, at least

test/exec_mode_float_control_khr.ll
```diff
@@ -0,0 +1,66 @@
+; RUN: llvm-as %s -o %t.bc
+; RUN: llvm-spirv %t.bc -o %t.spv --spirv-ext=+SPV_KHR_float_controls
+; RUN: llvm-spirv %t.spv -o %t.spt --to-text
+; RUN: FileCheck %s --input-file %t.spt -check-prefix=SPV
```
AlexeySachkov, line 4: Please launch validator ([1](https://github.com/KhronosGroup/SPIRV-Tools#validator), [2](https://github.com/KhronosGroup/SPIRV-LLVM-Translator/pull/180)) on the produced SPIR-V: if I understand correctly, right now this test won't produce valid SPIR-V, because of the presence of several execution modes on the same kernel


UPD: if I understand correctly, the validator doesn't catch such errors right now, but it will be fixed at some point, I guess: KhronosGroup/SPIRV-Tools#2150

NikitaRudenkoIntel: Do you say that there could be only one execution mode for kernel?
I cannot find that in spec...

lib/SPIRV/libSPIRV/SPIRVEntry.h
NikitaRudenkoIntel: Fixed

test/exec_mode_float_control_khr.ll
NikitaRudenkoIntel: Fixed

include/LLVMSPIRVExtensions.inc
```diff
@@ -18,3 +18,4 @@ EXT(SPV_INTEL_io_pipes)
 EXT(SPV_INTEL_inline_assembly)
 EXT(SPV_INTEL_arbitrary_precision_integers)
 EXT(SPV_INTEL_optimization_hints)
+EXT(SPV_KHR_float_controls)
```
svenvh, line 21: Should we move this to the top, so that the SPV_KHR extensions are together?

lib/SPIRV/libSPIRV/spirv.hpp
```diff
@@ -91,46 +91,51 @@ enum MemoryModel {
...
+  ExecutionModeSubgroupsPerWorkgroup = 36,
+  ExecutionModeMaxWorkgroupSizeINTEL = 5893,
+  ExecutionModeMaxWorkDimINTEL = 5894,
+  ExecutionModeNoGlobalOffsetINTEL = 5895,
+  ExecutionModeNumSIMDWorkitemsINTEL = 5896,
+  ExecutionModeDenormPreserve = 4459,
+  ExecutionModeDenormFlushToZero = 4460,
+  ExecutionModeSignedZeroInfNanPreserve = 4461,
+  ExecutionModeRoundingModeRTE = 4462,
+  ExecutionModeRoundingModeRTZ = 4463,
```
svenvh, line 137: Maybe better to move this a few lines up, so that the enum values are in their natural order.

NikitaRudenkoIntel: Fixed

include/LLVMSPIRVExtensions.inc
NikitaRudenkoIntel: Fixed

Issue #524, opened 2020-05-01
Assertion failure when debug info, loop metadata and function declaration interact


aus-intel:
Reverse translation of this opencl code
```
float bar(int x);

__kernel void foo(__global float* outPtr, int i) {
#pragma clang loop unroll(enable)
  for (int j = 0; j < i; ++j) {
    outPtr[j] = bar(j);
  }
}
```
fails with assertion failure in debug and segfault in release.
Commands to reproduce:
```
$ clang repro.cl -Xclang -finclude-default-header -c -emit-llvm -O0 -target spir64 -gline-tables-only
$ llvm-spirv repro.bc
$ llvm-spirv -r repro.spv -o repro_rev.bc
```
Result:
```
llvm-spirv: llvm-project/llvm/include/llvm/ADT/ilist_iterator.h:138: llvm::ilist_iterator<OptionsT, IsReverse, IsConst>::reference llvm::ilist_iterator<OptionsT, IsReverse, IsConst>::operator*() const [with OptionsT = llvm::ilist_detail::node_options<llvm::BasicBlock, true, false, void>; bool IsReverse = false; bool IsConst = false; llvm::ilist_iterator<OptionsT, IsReverse, IsConst>::reference = llvm::BasicBlock&]: Assertion `!NodePtr->isKnownSentinel()' failed.
```

aus-intel:
I tried to investigate main source of problem. The issue arises mostly because of this call https://github.com/KhronosGroup/SPIRV-LLVM-Translator/blob/54f072eb9f1eee8eac799cd701e96c82e9b2d403/lib/SPIRV/SPIRVReader.cpp#L2431 in transFunction.

transLLVMLoopMetadata tries to build LoopInfo for translated function. It keeps member map with metadata. When this map is empty, transLLVMLoopMetadata immediately returns without any attempts to build LoopInfo.
https://github.com/KhronosGroup/SPIRV-LLVM-Translator/blob/54f072eb9f1eee8eac799cd701e96c82e9b2d403/lib/SPIRV/SPIRVReader.cpp#L834-L841
This works when translation goes inorder -- declarations are handled before they are referenced in definitions.

However, when debug info is present, it can forward-reference function definitions. In this case, forward reference is immediately translated as it was real definition point. This leads to out of order translation with very strange effects. In this case it looks like this:
1) Debug info translation is started.
2) Forward reference to foo is met so foo translation is started.
3) Metadata in foo is placed in map
[...]

PR #541, opened 2020-05-19
Fix translation of undefined function call when debug info is present

vmaksimo:
This fix resolves issue:
https://github.com/KhronosGroup/SPIRV-LLVM-Translator/issues/524

Function translation invokes translation of LLVM loop metadata as a last step.
SPIRVToLLVM keeps a map with metadata. If function with loop metadata is
forward-referenced, translation starts and metadata is placed in the map. If
this function contains calls of undefined functions inside, it will fail -
metadata map is not empty, but there is no loop info inside the undefined
function.

Example:
```
float bar(int x);

__kernel void foo(__global float* outPtr, int i) {
  for (int j = 0; j < i; ++j) {
    outPtr[j] = bar(j);
  }
}
```

### DIFF
--- a/test/DebugInfo/Generic/undef-func-call.ll
+++ b/test/DebugInfo/Generic/undef-func-call.ll
@@ -1,0 +1,100 @@
+; RUN: llvm-as < %s -o %t.bc
+; RUN: llvm-spirv %t.bc -o %t.spv
+; RUN: llvm-spirv -r %t.spv
+
+; This is a regression test for reported issue https://github.com/KhronosGroup/SPIRV-LLVM-Translator/issues/524.
+; Test checks that reverse translation will not fail with assertion.
+
+; Build from the following source with clang -c -emit-llvm -O0 -target spir64 -gline-tables-only
+; float bar(int x);
+
+; __kernel void foo(__global float* outPtr, int i) {
+; #pragma clang loop unroll(enable)
+;   for (int j = 0; j < i; ++j) {
+;     outPtr[j] = bar(j);
+;   }
+; }
+
+target datalayout = "e-i64:64-v16:16-v24:32-v32:32-v48:64-v96:128-v192:256-v256:256-v512:512-v1024:1024"
+target triple = "spir64"
+
+; Function Attrs: convergent noinline norecurse nounwind optnone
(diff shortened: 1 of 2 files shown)


test/DebugInfo/Generic/undef-func-call.ll
```diff
@@ -0,0 +1,98 @@
+; RUN: llvm-as < %s -o %t.bc
+; RUN: llvm-spirv %t.bc -o %t.spv
+; RUN: llvm-spirv -r %t.spv -o - | llvm-dis -o %t.ll
+; RUN: llc -mtriple=%triple < %t.ll
```
AlexeySotkin, line 4: Why do you run llc?

test/DebugInfo/Generic/undef-func-call.ll
```diff
@@ -0,0 +1,98 @@
...
+
```
AlexeySotkin, line 5: I think the tests should have CHECK lines, or an explicit comment that this is a regression test for crash reported in https://github.com/KhronosGroup/SPIRV-LLVM-Translator/issues/524

vmaksimo: Actually I just wanted to check that translator doesn't crash in this pipeline. Will add a comment (and remove llc run as a redundant run)

test/DebugInfo/Generic/undef-func-call.ll
```diff
@@ -0,0 +1,100 @@
+; RUN: llvm-as < %s -o %t.bc
+; RUN: llvm-spirv %t.bc -o %t.spv
+; RUN: llvm-spirv -r %t.spv -o - | llvm-dis -o %t.ll
+
+; This is a regression test for reported issue https://github.com/KhronosGroup/SPIRV-LLVM-Translator/issues/524.
+; Test checks that reverse translation will not fail.
```
AlexeySotkin, line 6:
```suggestion
; Test checks that reverse translation will not fail with assertion.
```

test/DebugInfo/Generic/undef-func-call.ll
```diff
@@ -0,0 +1,100 @@
+; RUN: llvm-as < %s -o %t.bc
+; RUN: llvm-spirv %t.bc -o %t.spv
+; RUN: llvm-spirv -r %t.spv -o - | llvm-dis -o %t.ll
```
AlexeySotkin, line 3: I think `llvm-dis` is not needed as well.

vmaksimo:
> Please squash the commits.

Can it be "squash and merged" automatically when merging PR?

PR #542, opened 2020-05-20
Function pointers support improvements

zuban32:
- Auto add ReferencedIndirectly decoration when creating a function pointer
- Fix constexprs lowering for const vectors of funcpointers

### DIFF
--- a/lib/SPIRV/SPIRVLowerConstExpr.cpp
+++ b/lib/SPIRV/SPIRVLowerConstExpr.cpp
@@ -113,7 +113,6 @@ bool SPIRVLowerConstExpr::runOnModule(Module &Module) {
 
 void SPIRVLowerConstExpr::visit(Module *M) {
   for (auto &I : M->functions()) {
-    std::map<ConstantExpr *, Instruction *> CMap;
     std::list<Instruction *> WorkList;
     for (auto &BI : I) {
       for (auto &II : BI) {
@@ -124,7 +123,10 @@ void SPIRVLowerConstExpr::visit(Module *M) {
     while (!WorkList.empty()) {
       auto II = WorkList.front();
 
-      auto LowerOp = [&II, &FBegin, &I](ConstantExpr *CE) {
+      auto LowerOp = [&II, &FBegin, &I](Value *V) -> Value * {
+        if (isa<Function>(V))
+          return V;
+        auto *CE = cast<ConstantExpr>(V);
         SPIRVDBG(dbgs() << "[lowerConstantExpressions] " << *CE;)
         auto ReplInst = CE->getAsInstruction();
         auto InsPoint = II->getParent() == &*FBegin ? II : &FBegin->back();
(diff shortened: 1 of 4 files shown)


lib/SPIRV/libSPIRV/SPIRVModule.cpp
```diff
@@ -1232,6 +1232,7 @@ SPIRVInstruction *SPIRVModuleImpl::addIndirectCallInst(
 
 SPIRVInstruction *SPIRVModuleImpl::addFunctionPointerINTELInst(
     SPIRVType *TheType, SPIRVFunction *TheFunction, SPIRVBasicBlock *BB) {
+  TheFunction->addDecorate(DecorationReferencedIndirectlyINTEL);
```
AlexeySachkov, line 1235: Note: this actually changes the semantic:

I guess that sometimes compiler is able to infer which exact function is used and insert direct call instead of an indirect, inline the function and remove its body and declaration from the module. However, with this declaration function must not be removed from the module even if it called directly and inlined.

One more concern is that this decoration is enabled under separate capability, because it is not really needed for function pointers to work, but used to enable different use cases.

Probably the best place for such change is some specific backend and not generic SPIR-V generator, but since the extension is vendor-specific and in state of first public draft, I don't have strong objections against this change right now. Just keep in mind that in longer term this might need to be moved to some specific backend/SPIR-V consumer

lib/SPIRV/SPIRVLowerConstExpr.cpp
```diff
@@ -149,25 +151,30 @@ void SPIRVLowerConstExpr::visit(Module *M) {
...
+          std::transform(Vec->op_begin(), Vec->op_end(),
+                         std::back_inserter(OpList),
+                         [LowerOp](Value *V) { return LowerOp(V); });
           Value *Repl = nullptr;
           unsigned Idx = 0;
-          for (auto V : ReplList)
+          std::list<Instruction *> ReplList;
+          for (auto V : OpList) {
+            if (auto *Inst = dyn_cast<Instruction>(V))
+              ReplList.push_back(Inst);
```
AlexeySachkov, line 168:
```suggestion
            if (isa<Instruction>(V))
              ReplList.push_back(V);
```

No need to perform actual cast, since type of the list changed

lib/SPIRV/SPIRVWriter.cpp
```diff
@@ -1255,13 +1255,19 @@ SPIRVValue *LLVMToSPIRV::transValueWithoutDecoration(Value *V,
 
   if (auto Ins = dyn_cast<InsertElementInst>(V)) {
     auto Index = Ins->getOperand(2);
-    if (auto Const = dyn_cast<ConstantInt>(Index))
+    if (auto Const = dyn_cast<ConstantInt>(Index)) {
+      SPIRVValue *InsVal = nullptr;
+      if (auto *F = dyn_cast<Function>(Ins->getOperand(0)))
+        InsVal = BM->addFunctionPointerINTELInst(
```
AlexeySachkov, line 1261: Should be guarded by a check if corresponding extension is supported

test/constexpr_vector.ll
```diff
@@ -40,8 +39,12 @@
...
 ; CHECK-LLVM-NEXT:   %35 = insertelement <16 x i8> %34, i8 %19, i32 15
 ; CHECK-LLVM-NEXT:   store <16 x i8> %35, <16 x i8>* %Funcs, align 16
-; CHECK-LLVM-NEXT:   ret void
-; CHECK-LLVM-NEXT: }
+; CHECK-LLVM:        %Funcs1 = alloca <2 x i64>, align 16
+; CHECK-LLVM-NEXT:   %36 = ptrtoint i32 (i32)* @_Z2f1u2CMvb32_j to i64
+; CHECK-LLVM-NEXT:   %37 = ptrtoint i32 (i32)* @_Z2f2u2CMvb32_j to i64
+; CHECK-LLVM-NEXT:   %38 = insertelement <2 x i64> undef, i64 %36, i32 0
+; CHECK-LLVM-NEXT:   %39 = insertelement <2 x i64> %38, i64 %37, i32 1
+; CHECK-LLVM-NEXT:   store <2 x i64> %39, <2 x i64>* %Funcs1, align 16
```
AlexeySachkov, line 47: `SPIRVWriter` part was changed, but corresponding check-lines in this test weren't updated

lib/SPIRV/SPIRVLowerConstExpr.cpp
zuban32: No, we can't push Value* to a list<Instruction*> w/o any explicit cast

lib/SPIRV/libSPIRV/SPIRVModule.cpp
zuban32: Fine, I'll drop this

PR #543, opened 2020-05-20
added missing implementation of getPassName for SPIRVLowerOCLBlocks

aparshinIE:
SPIRV Lower OCL blocks pass does not have getPassName implementation. This commit adds one


lib/SPIRV/SPIRVLowerOCLBlocks.cpp
```diff
@@ -75,6 +75,10 @@ class SPIRVLowerOCLBlocks : public ModulePass {
 public:
   SPIRVLowerOCLBlocks() : ModulePass(ID) {}
 
+  StringRef getPassName() const override {
+    return "Lower OCL Blocks For SPIRV";
```
svenvh, line 79:
```suggestion
    return "Lower OpenCL Blocks For SPIR-V";
```

PR #546, opened 2020-05-21
Clang, due to some ABI constraints needs to generate llvm.trap

zahiraam:
intrinsics. According to the LangRef documentation a trap instruction
is lowered to a target's trap instruction or to an abort() function
if the target doesn't have a trap instruction. SPIRV has neither a
trap nor an abort instruction and no current opcode has the semantics
of an abort/trap.
Currently the IR to SPIRV translator is crashing when it finds an
llvm.trap intrinsic. The solution will require some thoughts on the
SPIRV side to decide how to implement an abort instruction.
This patch changes the translator so it doesn't crash. This will be
revised when a decision is taken.

NOTE: clang could eventually not generate an llvm.trap instruction in
the current case (non-base destructor of an abstract class needs to be
emmitted) but keep in mind that clang might generate an llvm.trap
intrinsic some other ways and we might stumble into this issue again.
An alternative to this, could be to add an LLVM pass that will get rid
of the llvm.trap intrisincs in the code before the SPIRV translator.
But having the translator solve this issue is a preferred solution.

### DIFF
--- a/lib/SPIRV/OCL20ToSPIRV.cpp
+++ b/lib/SPIRV/OCL20ToSPIRV.cpp
@@ -810,29 +810,13 @@ void OCL20ToSPIRV::transAtomicBuiltin(CallInst *CI, OCLBuiltinTransInfo &Info) {
         const size_t ArgsCount = Args.size();
         const size_t ScopeIdx = ArgsCount - 1;
         const size_t OrderIdx = ScopeIdx - NumOrder;
-        if (auto ScopeInt = dyn_cast_or_null<ConstantInt>(Args[ScopeIdx])) {
-          Args[ScopeIdx] = mapUInt(M, ScopeInt, [](unsigned I) {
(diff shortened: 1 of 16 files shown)


AlexeySotkin:
@zahiraam please add a title to the commit message and left one empty line between the title and the commit message body.

AlexeySotkin:
@zahiraam please add a test - LLVM IR containing `llvm.trap` intrinsic being translated to SPIR-V.

zahiraam:
> @zahiraam please add a title to the commit message and left one empty line between the title and the commit message body.

Done.

zahiraam:
> @zahiraam please add a test - LLVM IR containing `llvm.trap` intrinsic being translated to SPIR-V.

Done.

zahiraam:
Let's see if this time it will work. Thanks!

AlexeySotkin:
@zahiraam please squash your commits and rebase on top of the master branch. Thanks

PR #556, opened 2020-05-27
Add translation of memory model from module metadata

NikitaRudenkoIntel:


lib/SPIRV/SPIRVWriter.cpp
```diff
@@ -105,6 +105,18 @@ static void foreachKernelArgMD(
   }
 }
 
+static SPIRVMemoryModelKind getMemoryModel(Module &M) {
+  NamedMDNode *MemoryModelMD = M.getNamedMetadata(kSPIRVMD::MemoryModel);
+  if (MemoryModelMD) {
+    auto Model = static_cast<SPIRVMemoryModelKind>(
```
AlexeySotkin, line 111: I think we should either return result of static cast directly, or break down this expression into several simpler ones, possibly with checks that pointers are not null (static analysis tools may find issues here).

NikitaRudenkoIntel: I added some checks

test/memory_model_md.ll
```diff
@@ -0,0 +1,31 @@
...
+; RUN: llvm-dis %t.bc -o %t.ll
+; RUN: FileCheck %s --input-file %t.ll  -check-prefix=LLVM
+; RUN: FileCheck %s --input-file %t.spt -check-prefix=SPV
+
+; ModuleID = 'float_control_empty.bc'
+source_filename = "float_control_empty.cpp"
+target datalayout = "e-p:64:64-i64:64-n8:16:32"
+target triple = "spir"
+
+; LLVM-DAG: !spirv.MemoryModel = !{![[MMMD:[0-9]+]]}
```
svenvh, line 14: The use of `-DAG` feels a bit too strong for this case.

NikitaRudenkoIntel: Fixed

PR #558, opened 2020-05-28
Ignore llvm.trap intrinsic

AlexeySotkin:
instruction.

Clang, due to some ABI constraints needs to generate llvm.trap
intrinsics. According to the LangRef documentation a trap instruction
is lowered to a target's trap instruction or to an abort() function
if the target doesn't have a trap instruction. SPIRV has neither a
trap nor an abort instruction and no current opcode has the semantics
of an abort/trap.
Currently the IR to SPIRV translator is crashing when it finds an
llvm.trap intrinsic. The solution will require some thoughts on the
SPIRV side to decide how to implement an abort instruction.
This patch changes the translator so it doesn't crash. This will be
revised when a decision is taken.

NOTE: clang could eventually not generate an llvm.trap instruction in
the current case (non-base destructor of an abstract class needs to be
emmitted) but keep in mind that clang might generate an llvm.trap
intrinsic some other ways and we might stumble into this issue again.
An alternative to this, could be to add an LLVM pass that will get rid
of the llvm.trap intrisincs in the code before the SPIRV translator.
But having the translator solve this issue is a preferred solution.

### DIFF
--- a/lib/SPIRV/SPIRVWriter.cpp
+++ b/lib/SPIRV/SPIRVWriter.cpp
@@ -1896,6 +1896,9 @@ SPIRVValue *LLVMToSPIRV::transIntrinsicInst(IntrinsicInst *II,
   case Intrinsic::invariant_start:
   case Intrinsic::invariant_end:
   case Intrinsic::dbg_label:
+  case Intrinsic::trap:
+    // llvm.trap intrinsic is not implemented. But for now don't crash. This
+    // change is pending the trap/abort intrisinc implementation.
     return nullptr;
   default:
     if (SPIRVAllowUnknownIntrinsics)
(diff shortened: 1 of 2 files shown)


zahiraam:
@AlexeySotkin Thanks. Looks good. Shouldn't the word "instruction" be in the title ?

AlexeySotkin:
> @AlexeySotkin Thanks. Looks good. Shouldn't the word "instruction" be in the title ?

I have rewritten the title if you don't mind.

zahiraam:
> > @AlexeySotkin Thanks. Looks good. Shouldn't the word "instruction" be in the title ?
> 
> I have rewritten the title if you don't mind.
No I don't. But I think the commit message needs to be fixed. We can add the sentence that you removed from the title inside the commit message before the word "instruction".
Can you do that?

AlexeySotkin:
I think there is a confusion between the PR title and commit message. I guess, you meant the PR title or  comment. Otherwise I don't understand why should I change the commit message to add words saying that commit message has been changed. Only one commit message will be kept in the git history.
Anyway I suggest just merge this PR unless there are other objections.

Issue #496, opened 2020-04-17
Wrong target branch in the check_code_format.sh.


AlexeySotkin:
There is a strange problem with check_code_format.sh. As I see from the source, it gets diff between master and current commit.
`FORMAT_DIFF=$(git diff -U0 master -- ${FILES_TO_CHECK} | ./utils/clang-format-diff.py -p1 -style=file)`
This is not correct for release branches -- diff should be taken between release branch and commit to this branch. In this PR this led to the problem, that script found problem in the code that was not modified by this particular PR. This part of code was modified in my previous PR to master branch, and with this PR script thinks that the code was reformatted back and raises an error.
PR 490 is merged, and I will just rebase this change on top of 90 branch. The error should disappear. However, there still will be an error in check_code_format.sh.

_Originally posted by @aus-intel in https://github.com/KhronosGroup/SPIRV-LLVM-Translator/pull/491#issuecomment-614583164_

AlexeySachkov:
I removed that script in #770 during switch to GitHub Actions, so this is not actual anymore, closing

PR #562, opened 2020-06-01
Fix base branch in code check scripts for llvm 10

aus-intel:
Patch fixes github issue for llvm 10 branch:
https://github.com/KhronosGroup/SPIRV-LLVM-Translator/issues/496

Example failure without this patch: https://travis-ci.org/github/KhronosGroup/SPIRV-LLVM-Translator/jobs/693502720.

### DIFF
--- a/utils/check_code_format.sh
+++ b/utils/check_code_format.sh
@@ -24,7 +24,25 @@
 # * Add a filtering step on the files to check, with the result stored in
 #   FILES_TO_CHECK.
 
-MODIFIED_FILES=$(git diff --name-only master | grep -E ".*\.(cpp|cc|c\+\+|cxx|c|h|hpp)$")
+if [ -n "${TRAVIS_BRANCH+check}" ]; then
+    BASE_BRANCH="travis/${TRAVIS_BRANCH}"
+    git fetch --depth=1 origin ${TRAVIS_BRANCH}:${BASE_BRANCH}
+else
+    BASE_BRANCH=$(git for-each-ref --format='%(upstream:short)' "$(git symbolic-ref -q HEAD)")
+fi
+
+if [ -z ${BASE_BRANCH} ]; then
+    cat <<EOF
+Error: no branch to compare with.
+If you are running script locally, please set upstream branch
+that you want to compare your changes with: "git branch --set-upstream-to=<branch>".
+EOF
+    exit 2
+fi
+
+echo "Checking changes between '${BASE_BRANCH}' and 'HEAD'..."
+
+MODIFIED_FILES=$(git diff --name-only ${BASE_BRANCH} | grep -E ".*\.(cpp|cc|c\+\+|cxx|c|h|hpp)$")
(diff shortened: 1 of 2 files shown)


aus-intel:
Good, I will create corresponding PRs for other relevant branches (master and other releases) when this one will be merged.

utils/check_code_format.sh
```diff
@@ -24,15 +24,17 @@
 # * Add a filtering step on the files to check, with the result stored in
 #   FILES_TO_CHECK.
 
-MODIFIED_FILES=$(git diff --name-only master | grep -E ".*\.(cpp|cc|c\+\+|cxx|c|h|hpp)$")
+BASE_BRANCH=llvm_release_100
```
AlexeySotkin, line 27: Having the `BASE_BRANCH` hard-coded makes it one more thing we have to take care of when we create new branch.
I think we might use `$TRAVIS_BRANCH` (https://docs.travis-ci.com/user/environment-variables/#default-environment-variables) :
1. pass it from `.travis.yaml`, as a command line argument to the script. OR
2. use `$TRAVIS_BRANCH` in the scrip if the env variable exists and use `master` otherwise.

AlexeySotkin: Looks good.

aus-intel:
I see that `$TRAVIS_BRANCH` is set to just 'llvm_release_100' and this branch does not exist in repo during CI testing (but it looks like remote is set to 'origin' and 'origin/llvm_release_100' exists). This leads to failure in diff calculation.

I think I will just prepend 'origin/' to `$TRAVIS_BRANCH` when it is used.

AlexeySachkov:
> Still fails because branch is missing. It happens because `--depth=1` is passed to `git clone`. In this case no branches other than master are fetched (shallow clone). Proposed solution: change `depth: 1` to `depth: false` in `travis.yml` (https://docs.travis-ci.com/user/customizing-the-build#git-clone-depth).

If I understand correctly, this is done to improve performance of checkout operation. We don't have a lot branches in the repo and I don't have data how exactly shallow clone help us, so I cannot say for sure whether it is okay to do so or not.

What about performing additional fetch within clang-format and tidy scripts? If I understand correctly, it should be doable even with shallow clone and it won't affect the rest of jobs which doesn't need other git references rather then PR

utils/check_code_format.sh
```diff
@@ -24,15 +24,32 @@
...
+    cat <<EOF
+Error: no branch to compare with.
+If you are running script locally, please set upstream branch
+that you want to compare your changes with: "git branch --set-upstream-to=<branch>".
+EOF
+    exit 2
+fi
+
+echo "Checking changes between '${BASE_BRANCH}' and 'HEAD'..."
+
```
AlexeySotkin, line 43:
```suggestion

git fetch origin ${BASE_BRANCH}
```

aus-intel:
Updated patch. Now in case of travis ci, last commit of `$TRAVIS_BRANCH` is fetched as local branch `travis/$TRAVIS_BRANCH`. With this, check works as expected.

PR #565, opened 2020-06-02
Implement lowering of llvm.sadd.with.overflow.[i16/i32/i64]

AlexeySotkin:
Since the intrinsic can't be mapped to SPIR-V directly, its semantics has been
implemented as a function in LLVM IR. This patch introduces a pass which links
the implementation function into the IR being translated to SPIR-V.


include/LLVMSaddWithOverflow.inc
```diff
@@ -0,0 +1,56 @@
...
+
+; #include <stdlib.h>
+; const unsigned int abs_pos_max = 2147483647;
+; const unsigned int abs_neg_max = 2147483648;
+;
+; void llvm_sadd_with_overflow_i32(int a, int b, int& c, bool& o) {
+;     bool overflow = false;
+;     bool both_pos = (a>=0 && b>=0);
+;     bool both_neg = (a<0 && b<0);
+;     if (both_pos || both_neg) {
```
AlexeySotkin, line 13:
```suggestion
;     bool both_neg = (a<0 && b<0);
;     if (both_neg && (a == int_min || b == int_min))
        overflow = true;
;     else if (both_pos || both_neg) {
```

lib/SPIRV/SPIRVLowerSaddWithOverflow.cpp
```diff
@@ -0,0 +1,133 @@
...
+      // the loaded module.
+      StringRef FuncName = "llvm_sadd_with_overflow_i32";
+      FunctionCallee F = Mod->getOrInsertFunction(FuncName, I.getFunctionType());
+      //TODO Should we set attributes for the function F?
+      I.setCalledFunction(F);
+
+      // Link in the intrinsic's implementation.
+      Linker L(*Mod);
+      L.linkInModule(std::move(Src), Linker::Flags::LinkOnlyNeeded);
+      //TODO add error handling
```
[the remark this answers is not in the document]
AlexeySotkin: Right

lib/SPIRV/SPIRVLowerSaddWithOverflow.cpp
```diff
@@ -0,0 +1,133 @@
...
+  virtual void visitIntrinsicInst(CallInst &I) {
+    IntrinsicInst *II = dyn_cast<IntrinsicInst>(&I);
+    if (!II || II->getIntrinsicID() != Intrinsic::sadd_with_overflow)
+      return;
+
+    // Read LLVM IR with the intrinsic's implementation
+    SMDiagnostic Diag;
+    auto Buffer = MemoryBuffer::getMemBuffer(Impl);
+    auto& DLStr = Mod->getDataLayoutStr();
+    auto Src = parseIR(*Buffer, Diag, *Context, [&](StringRef){return DLStr;});
```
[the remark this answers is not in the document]
AlexeySotkin: No, we need to parse and link the IR only once. I'll fix it. Thanks!

AlexeySotkin:
> However, I wonder how long this IR will be able to stay here without any modifications, because LLVM is evolving and something might be changed in the future.

I expect it to be about the same as for programmatic/API approach. From my experience(correct me if I'm wrong) when LLVM IR is changing, then API, textual and binary formats are changing at the same time.

AlexeySotkin:
> One more thing: this approach increases size of the translator, because the whole LLVM IR with comments will be included as a plain string into a corresponding section in binary. Probably we can pre-compile it down to LLVM BC during the translator build and include binary form into the pass?

Sounds interesting. I'll give it a try.

AlexeySotkin:
> > One more thing: this approach increases size of the translator, because the whole LLVM IR with comments will be included as a plain string into a corresponding section in binary. Probably we can pre-compile it down to LLVM BC during the translator build and include binary form into the pass?
> 
> Sounds interesting. I'll give it a try.

I could not find a portable solution to include a binary into a C++ source file. But I found solution how to not include comments in the translator binary. I'v made the comments part of C++ code instead of LLVM IR.

lib/SPIRV/SPIRVLowerSaddWithOverflow.cpp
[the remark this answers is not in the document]
AlexeySotkin: `parseIR` returns a `uniqueue_ptr` which is passed to `linkModules` which takes the ownership of the pointer. So next time we need to have another `unique_ptr`.

lib/SPIRV/LLVMSaddWithOverflow.h
```diff
@@ -0,0 +1,244 @@
+//===- LLVMSaddWithOverflow.h - implementation of llvm.sadd.with.overflow -===//
+//
+//                     The LLVM/SPIRV Translator
+//
+// This file is distributed under the University of Illinois Open Source
```
svenvh, line 5: Since you are adding some new files, should we reconsider the copyright header on those? (see https://github.com/KhronosGroup/SPIRV-LLVM-Translator/issues/80)

lib/SPIRV/SPIRVLowerSaddWithOverflow.cpp
```diff
@@ -0,0 +1,156 @@
...
+  virtual void visitIntrinsicInst(CallInst &I) {
+    IntrinsicInst *II = dyn_cast<IntrinsicInst>(&I);
+    if (!II || II->getIntrinsicID() != Intrinsic::sadd_with_overflow)
+      return;
+
+    Function *IntrinsicFunc = I.getCalledFunction();
+    assert(IntrinsicFunc && "Missing function");
+    StringRef IntrinsicName = IntrinsicFunc->getName();
+    std::string FuncName = "llvm_sadd_with_overflow_i";
+    if (IntrinsicName.endswith(".i16"))
```
svenvh, line 80: Minor detail: the commit message says you're only handling i32, but you seem to handle i16/i32/i64.

lib/SPIRV/LLVMSaddWithOverflow.h
AlexeySotkin: I'm not sure that it is okay to have files with different licenses in one project. 
I'd prefer to see clarifications on the Khronos re-licensing process from Neil. I think it will be more consistent if we change all copyright headers in scope of other PR.

svenvh: Fair enough, I'm fine leaving those headers as they are.

lib/SPIRV/LLVMSaddWithOverflow.h
```diff
@@ -0,0 +1,244 @@
...
+// THE SOFTWARE.
+//
+//===----------------------------------------------------------------------===//
+//
+// This file implements lowering of llvm.sadd.with.overflow.* into basic LLVM
+// operations.
+//
+//===----------------------------------------------------------------------===//
+
+// The IR below is slightly manually modified IR which was produced by Clang
```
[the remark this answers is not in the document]
AlexeySotkin: I've added a couple of sentences.

lib/SPIRV/SPIRVLowerSaddWithOverflow.cpp
AlexeySotkin: Fixed.

PR #566, opened 2020-06-03
SPIRV to LLVM translation for cl_khr_subgroup_extensions

aratajew:


lib/SPIRV/OCLUtil.cpp
```diff
@@ -761,6 +761,22 @@ class OCLBuiltinFuncMangleInfo : public SPIRV::BuiltinFuncMangleInfo {
     } else if (UnmangledName.find("intel_sub_group_media_block_write") !=
                std::string::npos) {
       addUnsignedArg(3);
+    } else if (UnmangledName.find(kOCLBuiltinName::SubGroupPrefix) !=
```
AlexeySachkov, line 764: This is not for this PR, but I wonder why do we use `std::string` here? We could use `llvm::StringRef::contains` and significantly shorten all these ifs.

lib/SPIRV/SPIRVReader.cpp
```diff
@@ -587,7 +589,8 @@ bool SPIRVToLLVM::isDirectlyTranslatedToOCL(Op OpCode) const {
     // everything except atomics, groups, pipes and media_block_io_intel is
```
AlexeySachkov, line 589: I suggest to update this comment: right now it tries to explicitly list all exceptions. Probably some more generic phase that "there are some exceptions" would be better since we increase this list

lib/SPIRV/SPIRVToOCL.cpp
```diff
@@ -247,51 +247,183 @@ void SPIRVToOCL::visitCallSPRIVImageQuerySize(CallInst *CI) {
...
+      FuncName = getNonUniformArithmeticBuiltinName(CI, OC);
+    else if (OC == OpGroupNonUniformBallotBitCount)
+      FuncName = getBallotBuiltinName(CI, OC);
+    else
+      llvm_unreachable("Unsupported opcode!");
+  }
+  return FuncName;
+}
+
+bool SPIRVToOCL::extendRetTyToi32(Op OC) {
```
AlexeySachkov, line 384:
```suggestion
bool SPIRVToOCL::needToExtendRetTyToi32(Op OC) {
```

Also, it is probably better to make it just a `static` function instead of class member function

lib/SPIRV/OCLUtil.cpp
svenvh: I've had that cleanup on my TODO list for a long time indeed, but never got around to do it.  But that's something beyond the scope of this PR anyway.

lib/SPIRV/SPIRVToOCL.cpp
aratajew: Done.

lib/SPIRV/SPIRVReader.cpp
aratajew: Done.

lib/SPIRV/OCLUtil.cpp
aratajew: Ok, let's leave it as a future refactoring opportunity.

PR #569, opened 2020-06-03
Implementation of vector compute extension

NikitaRudenkoIntel:


lib/SPIRV/VectorComputeUtil.cpp
```diff
@@ -0,0 +1,116 @@
+//=- VectorComputeUtil.cpp - vector compute utilities implemetation * C++ -*-=//
+//
+//                     The LLVM/SPIR-V Translator
```
AlexeySotkin, line 3: Please add the license header as in https://github.com/KhronosGroup/SPIRV-LLVM-Translator/blob/master/lib/SPIRV/LLVMToSPIRVDbgTran.cpp

lib/SPIRV/VectorComputeUtil.h
```diff
@@ -0,0 +1,131 @@
+//=- VectorComputeUtil.h - vector compute utilities declarations -*- C++ -*-=//
+//
+//                     The LLVM/SPIR-V Translator
```
AlexeySotkin, line 3: Please add the license header as in https://github.com/KhronosGroup/SPIRV-LLVM-Translator/blob/master/lib/SPIRV/LLVMToSPIRVDbgTran.cpp

AlexeySotkin:
In general the PR looks good to me. It is marked as draft. Are there anything else to be done? If not I suggest to push "Ready for review" button.

lib/SPIRV/VectorComputeUtil.h
NikitaRudenkoIntel: Fixed

lib/SPIRV/VectorComputeUtil.cpp
NikitaRudenkoIntel: Fixed

PR #571, opened 2020-06-03
Introduce a workaround for emitting only required capabilities

NikitaRudenkoIntel:
Introduce a workaround for emitting only required capabilities instead of all listed in SPIRVEnum.h

SPIRV specification declares that:
When an instruction, enumerant, or other feature specifies multiple
enabling capabilities, only one such capability needs to be declared to
use the feature. This declaration does not itself imply anything about
the presence of the other enabling capabilities: The execution
environment needs to support only the declared capability.
  
But the translator always emits all listed capabilities. This patch is a
workaround for choosing only actually needed capabilities for specific
cases.

https://www.khronos.org/registry/spir-v/specs/unified1/SPIRV.html#_a_id_capabilities_a_language_capabilities

### DIFF
--- a/lib/SPIRV/libSPIRV/SPIRVDecorate.h
+++ b/lib/SPIRV/libSPIRV/SPIRVDecorate.h
@@ -78,7 +78,9 @@ public:
   void setOwner(SPIRVDecorationGroup *Owner) { this->Owner = Owner; }
 
   SPIRVCapVec getRequiredCapability() const override {
-    return getCapability(Dec);
+    auto Caps = getAllEnablingCapabilities(Dec);
+    Module->chooseBestCapability(Caps);
+    return Caps;
   }
 
   SPIRVWord getRequiredSPIRVVersion() const override {
--- a/lib/SPIRV/libSPIRV/SPIRVEntry.h
+++ b/lib/SPIRV/libSPIRV/SPIRVEntry.h
@@ -648,9 +648,7 @@ public:
   SPIRVExecutionMode() : ExecMode(ExecutionModeInvocations) {}
   SPIRVExecutionModeKind getExecutionMode() const { return ExecMode; }
   const std::vector<SPIRVWord> &getLiterals() const { return WordLiterals; }
-  SPIRVCapVec getRequiredCapability() const override {
-    return getCapability(ExecMode);
-  }
+  SPIRVCapVec getRequiredCapability() const override;
 
   SPIRVWord getRequiredSPIRVVersion() const override {
     switch (ExecMode) {
--- a/lib/SPIRV/libSPIRV/SPIRVType.h
+++ b/lib/SPIRV/libSPIRV/SPIRVType.h
@@ -249,7 +249,8 @@ public:
     auto Cap = getVec(CapabilityAddresses);
     if (getElementType()->isTypeFloat(16))
       Cap.push_back(CapabilityFloat16Buffer);
-    auto C = getCapability(ElemStorageClass);
+    auto C = getAllEnablingCapabilities(ElemStorageClass);
+    Module->chooseBestCapability(C);
     Cap.insert(Cap.end(), C.begin(), C.end());
     return Cap;
   }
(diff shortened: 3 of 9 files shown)


lib/SPIRV/libSPIRV/SPIRVDecorate.h
```diff
@@ -78,7 +78,9 @@ class SPIRVDecorateGeneric : public SPIRVAnnotationGeneric {
   void setOwner(SPIRVDecorationGroup *Owner) { this->Owner = Owner; }
 
   SPIRVCapVec getRequiredCapability() const override {
-    return getCapability(Dec);
+    auto Caps = getAllEnablingCapabilities(Dec);
+    Module->chooseBestCapability(Caps);
```
AlexeySachkov, line 82:
```suggestion
    Module->chooseBestCapability(getAllEnablingCapabilities(Dec));
```

lib/SPIRV/libSPIRV/SPIRVEntry.h
```diff
@@ -642,7 +642,9 @@ class SPIRVExecutionMode : public SPIRVAnnotation<OpExecutionMode> {
   SPIRVExecutionModeKind getExecutionMode() const { return ExecMode; }
   const std::vector<SPIRVWord> &getLiterals() const { return WordLiterals; }
   SPIRVCapVec getRequiredCapability() const override {
-    return getCapability(ExecMode);
+    auto Caps = getAllEnablingCapabilities(ExecMode);
+    Module->chooseBestCapability(Caps);
```
AlexeySachkov, line 646:
```suggestion
    Module->chooseBestCapability(getAllEnablingCapabilities(ExecMode));
```

lib/SPIRV/libSPIRV/SPIRVType.h
```diff
@@ -249,7 +249,8 @@ class SPIRVTypePointer : public SPIRVType {
     auto Cap = getVec(CapabilityAddresses);
     if (getElementType()->isTypeFloat(16))
       Cap.push_back(CapabilityFloat16Buffer);
-    auto C = getCapability(ElemStorageClass);
+    auto C = getAllEnablingCapabilities(ElemStorageClass);
+    Module->chooseBestCapability(C);
```
AlexeySachkov, line 253:
```suggestion
    Module->chooseBestCapability(getAllEnablingCapabilities(ElemStorageClass));
```

lib/SPIRV/libSPIRV/SPIRVDecorate.h
NikitaRudenkoIntel: There is "return Caps" one line below, so we cannot optimize it this way

lib/SPIRV/libSPIRV/SPIRVEntry.h
NikitaRudenkoIntel: There is "return Caps" one line below, so we cannot optimize it this way

lib/SPIRV/libSPIRV/SPIRVType.h
NikitaRudenkoIntel: There is "Cap.insert(Cap.end(), C.begin(), C.end());" one line below, so we cannot optimize it this way

lib/SPIRV/libSPIRV/SPIRVDecorate.h
AlexeySachkov: Oh, completely missed that, sorry

AlexeySotkin:
> Overall, this approach seems fine to me, however, I don't fully understand which exact code we should put into `chooseBestCapabilities`.

I think it is easier to say which capabilities we don't want to be emitted by the translator. If I understand correctly, purpose of the project in this repo is to handle _compute_ flavor of SPIR-V. So, I believe, in this function we should exclude the **Shader** capability and all capabilities that enable it implicitly or indirectly. For example, for **OpPtrAccessChain** we have the following enabling capabilities and their implicitly declared capabilities:
```
Addresses
VariablePointers -> VariablePointersStorageBuffer -> Shader
VariablePointersStorageBuffer -> Shader
PhysicalStorageBufferAddresses -> Shader
```
In this case `chooseBestCapabilities` should return only **Addresses**.

NikitaRudenkoIntel:
That's right. In order to handle compute flavor (like VectorCompute) we want to emit a *minimal* set of capabilities. The definition of a minimal set is rather unclear and could vary from extension to extension.
In terms of VectorCompute *minimal* means always choosing CapabilityVectorCompute when possible.
In the current state of translator it is actually the only example when the special decision is needed. (All other cases mentioned in SPIRVEnum.h are just not implemented in other parts of the translator)

PR #573, opened 2020-06-04
[NFC] Move internal definitions to spirv_internal.hpp

MrSidims:
To align translator's version of spirv.hpp with
https://github.com/KhronosGroup/SPIRV-Headers
we need first to eliminate internal definitions in this repository.

With this PR a new header spirv_internal.hpp introduced.
It contains internal (for SPIR-V to LLVM IR translator) definitions
that are not a part of SPIR-V spec. For not it is LinkageTypeInternal
and OpForward.


MrSidims:
Tagging @svenvh  @AlexeySachkov @AlexeySotkin

AlexeySotkin:
> @AlexeySotkin, do you have any feedback?

Yes. I suggest to use `LinkageTypeMax` instead of `LinkageTypeInternal`. That should allow us not to add `spirv_internal.hpp`.
Other option might be to wrap `spv::LinkageType` into a `shared_ptr<spv::LinkageType>` and use `nullptr` to represent missing linkage type.

AlexeySachkov:
> > @AlexeySotkin, do you have any feedback?
> 
> Yes. I suggest to use `LinkageTypeMax` instead of `LinkageTypeInternal`. That should allow us not to add `spirv_internal.hpp`.

We will probably anyway have something like `spirv_internal.hpp`, because I doubt that every our extension proposal will go into KhronosGroup/SPIRV-Headers from the beginning, but this is a good suggestion anyway, I think

> Other option might be to wrap `spv::LinkageType` into a `shared_ptr<spv::LinkageType>` and use `nullptr` to represent missing linkage type.

This looks weird, IMHO. I would rather introduce new enumeration value via proposed approach from Dmitry or re-use existing one instead of introducing `shared_ptr` to hold an enum value, i.e. plain `int`

MrSidims:
> > @AlexeySotkin, do you have any feedback?
> 
> Yes. I suggest to use `LinkageTypeMax` instead of `LinkageTypeInternal`. That should allow us not to add `spirv_internal.hpp`.

I was thinking about it, but decided to introduce a new header instead (even if the copy-paste of OpCodes for removing OpForward from spirv.hpp is looking monstrous). The idea here is to, as Alexey mentioned, have a header to where we can put definitions from 'preview' extensions (for example inline assembly), and only when an extension masters and stabilizes we can upstream it in to SPIR-V headers repository itself (and therefore to spirv.hpp).

AlexeySotkin:
> > > @AlexeySotkin, do you have any feedback?
> > 
> > 
> > Yes. I suggest to use `LinkageTypeMax` instead of `LinkageTypeInternal`. That should allow us not to add `spirv_internal.hpp`.
> 
> I was thinking about it, but decided to introduce a new header instead (even if the copy-paste of OpCodes for removing OpForward from spirv.hpp is looking monstrous). The idea here is to, as Alexey mentioned, have a header to where we can put definitions from 'preview' extensions (for example inline assembly), and only when an extension masters and stabilizes we can upstream it in to SPIR-V headers repository itself (and therefore to spirv.hpp).

I don't see much value/difference if we still have to copy-paste OpCodes from the original spirv.hpp header.

AlexeySotkin:
How about something like https://godbolt.org/z/qjhXq_ ?

MrSidims:
> How about something like https://godbolt.org/z/qjhXq_ ?

I'll try it out, thanks!

MrSidims:
@svenvh @AlexeySachkov @AlexeySotkin sorry for a long time prior updating. Here is a patch, that seems to be a good resolution of the problem

MrSidims:
There are 2 error reported by clang-format, that I haven't fixed. Shall I? They seems (for me) to be unreasonable.

lib/SPIRV/libSPIRV/SPIRVNameMapEnum.h
```diff
@@ -56,7 +57,7 @@ namespace SPIRV {
 template <> inline void SPIRVMap<LinkageType, std::string>::init() {
   add(LinkageTypeExport, "Export");
   add(LinkageTypeImport, "Import");
-  add(LinkageTypeInternal, "Internal");
+  add(LinkageType(internal::LinkageTypeInternal), "Internal");
```
AlexeySachkov, line 60: To avoid bunch of casts like this you could define `constexpr` variables within `spirv_internal.hpp`, something like:

```c++
enum LinkageTypeInternal {
  LTPrev = LinkageTypeMax - 2,
  LinkageTypeInternal
};

constexpr LinkageType LinkageTypeInternal = static_cast<LinkageType>(LinkagetTypeInternal);
// I hope that static_cast works within constexpr :)
```

BTW, having `LinkageTypeInternal` as both enum name and enum value doesn't seem to be right: it probably works, but if I understand correctly, since this is an unscoped enum, `LinkageTypeInternal` (as value) is also available as global constant, which kind of creates ambiguity with `LinkageTypeInternal` as typename of enum

AlexeySachkov:
> There are 2 error reported by clang-format, that I haven't fixed. Shall I? They seems (for me) to be unreasonable (or at least they go against project's code style).

I think that the suggestion from clang-format is valid, i.e. it tries to save lines on a small enum. However, I'm okay with not applying it - it won't be actual anymore if/when we extend the enum

lib/SPIRV/libSPIRV/SPIRVNameMapEnum.h
MrSidims: Thanks! Done

MrSidims: I think, I'll merged 3 commits when PR is approved.

lib/SPIRV/libSPIRV/SPIRVOpCode.h
```diff
@@ -42,14 +42,19 @@
...
 
 using namespace spv;
 namespace SPIRV {
 
 template <> inline void SPIRVMap<Op, std::string>::init() {
 #define _SPIRV_OP(x, ...) add(Op##x, #x);
+#define _SPIRV_OP_INTERNAL(x, ...) add(Op(internal::Op##x), #x);
 #include "SPIRVOpCodeEnum.h"
+  // Value for OpForward is set as OpMax - 2
+  _SPIRV_OP_INTERNAL(Forward, 2147483646)
```
AlexeySachkov, line 56: Why can't we directly use `OpMax - 2` here instead of this magic number?

AlexeySachkov (approved):
LGTM

> I think, I'll merged 3 commits when PR is approved.

BTW, I can do this myself while merging - we just need to align on the final commit message (for example I can grab the PR description)

svenvh (approved):
LGTM; thanks for finishing this off, @MrSidims !

PR #579, opened 2020-06-09
Add support for SPV_INTEL_usm_storage_classes extension

MrSidims:
With this extension 2 new Storage Classes are introduced:
DeviceOnlyINTEL and HostOnlyINTEL appropriately mapped on
global_device and global_host SYCL/OpenCL address spaces which
are part of SYCL_INTEL_usm_address_spaces extension.

Spec: https://github.com/intel/llvm/pull/1840

Co-authored-by: Viktoria Maksimova <viktoria.maksimova@intel.com>

Signed-off-by: Viktoria Maksimova <viktoria.maksimova@intel.com>


MrSidims:
Tagging @vmaksimo @mlychkov @AlexeySotkin @AlexeySachkov

MrSidims:
The error:

> error: line 138: Expected input to have storage class Generic: GenericCastToPtr
>   %59 = OpGenericCastToPtr %_ptr_CrossWorkgroup_uint %58

which I don't really understand from where it comes. Would appreciate any kind of hint.

vmaksimo:
> The error:
> 
> > error: line 138: Expected input to have storage class Generic: GenericCastToPtr
> > %59 = OpGenericCastToPtr %_ptr_CrossWorkgroup_uint %58
> 
> which I don't really understand from where it comes. Would appreciate any kind of hint.

@MrSidims This error is produced by `spirv-val`. It doesn't fail locally if you don't have it in the PATH variable

MrSidims:
> > The error:
> > > error: line 138: Expected input to have storage class Generic: GenericCastToPtr
> > > %59 = OpGenericCastToPtr %_ptr_CrossWorkgroup_uint %58
> > 
> > 
> > which I don't really understand from where it comes. Would appreciate any kind of hint.
> 
> @MrSidims This error is produced by `spirv-val`. It doesn't fail locally if you don't have it in the PATH variable

Got it, thanks! The reasons is that spirv-val tools checks that GenericCastToPtr has input of Generic storage type, while in the extension we allows an input of CrossWorkgroup storage class in case if it's being casted to Device/Host.

MrSidims:
I have disabled spirv-val for now. But may be we will reconsider the specification.

MrSidims:
@svenvh @AlexeySotkin can I ask your opinion of changing rules for already existing instructions GenericCastToPtr and PtrCastToGeneric (under the extension) and if it's possible to upstream the changes to spirv-val tool? If it's not OK we can just implement new instructions to reflect desired address space casts in SPIR-V.

lib/SPIRV/SPIRVReader.cpp
```diff
@@ -382,11 +382,18 @@ Type *SPIRVToLLVM::transType(SPIRVType *T, bool IsClassMember) {
   case OpTypeArray:
     return mapType(T, ArrayType::get(transType(T->getArrayElementType()),
                                      T->getArrayLength()));
-  case OpTypePointer:
+  case OpTypePointer: {
+    SPIRAddressSpace AS =
+        SPIRSPIRVAddrSpaceMap::rmap(T->getPointerStorageClass());
+    // If generation of SYCL USM address spaces isn't allowed in the reversed
+    // translation - transform them into global address space
+    if (!BM->isUsmAddrspacesEnabled() &&
```
AlexeySachkov, line 390: What is the reason to add this option? I mean, new storage classes enable `USMStorageClassesINTEL` capability in the module and it is responsibility of the particular consumer to either reject SPIR-V with such capability as unsupported or accept new storage classes, which can be done by lowering them to whichever address spaces or to anything else what is desired by the consumer.

So, the generator is responsible to either generate more portable SPIR-V, which doesn't use the extension, or generate less portable SPIR-V, which uses the extension, but cannot be supplied to any consumer (because of lack of the extension support).

If consumer is aware of that option, it means that it is aware of that extension and therefore, it could perform desired lowering by itself and code in the translator would be simpler.

I do understand that some projects are using the translator as SPIR-V consumer as well and this option allows to simplify their code, but it would be good if such customizations are not unique to a single consumer/backend/vendor and we don't end up with bunch of tweaks in the translator for each particular backend.

P.S. This is just my vision and it might be wrong. Probably it is such a problem to have some consumer part customizations not in the actual SPIR-V consumers, but here, in the translator. It would be good to hear feedback from someone else

lib/SPIRV/SPIRVInternal.h
```diff
@@ -191,6 +191,8 @@ enum SPIRAddressSpace {
   SPIRAS_Input,
   SPIRAS_Output,
   SPIRAS_Count,
+  SPIRAS_GlobalDevice = 11,
```
AlexeySachkov, line 194: I have a concern about `SPIRAS_GlobalDevice` value here: I'm not sure that this is reliable until we have this assignment documented/mentioned/used in llvm/llvm-project.

Someone might add new address space into upstream LLVM and use these values for some other purpose

MrSidims: Fair enough

AlexeySachkov:
> @svenvh @AlexeySotkin can I ask your opinion of changing rules for already existing instructions GenericCastToPtr and PtrCastToGeneric (under the extension)

This is probably question not to the translator, but to the SPIR-V core and extension specs maintainers, i.e. what is better for extension: define new instruction or change semantic of an existing one? I guess there is no single answer to that and it should be decided case by case

> and if it's possible to upstream the changes to spirv-val tool?

Again, this is probably question not for the translator, but for KhronosGroup/SPIRV-Tools. AFAIK, if you have your extension published in KhronosGroup/SPIRV-Headers there should be no problem with extending `spirv-val` with support for a new extension. I remember @bashbaug advised me to do so several times

lib/SPIRV/SPIRVReader.cpp
MrSidims: In general I don't see such customization as a big problem, we have here in the translator different flows for OpenCL 2.0 and OpenCL 1.2, right? Yet I see your point and understand it. Would also love to here other opinions.

(The story below isn't valuable for SPIR-V to LLVM IR translator project at all, but I want you to see the logic, why this option was added).
This option came from (surprise-surprise) some implementation details of SYCL compiler's driver and desire to make the feature adoptable not only for FPGA, but for other targets in JIT compilation (aka with triple `spirv-unknown-unknown` for which we add `spirv-ext=+all` option.

MrSidims:
> @svenvh @AlexeySotkin can I ask your opinion of changing rules for already existing instructions GenericCastToPtr and PtrCastToGeneric (under the extension) and if it's possible to upstream the changes to spirv-val tool? If it's not OK we can just implement new instructions to reflect desired address space casts in SPIR-V.

We have changed the extension, adding two more instructions.

lib/SPIRV/SPIRVInternal.h
MrSidims: Created https://reviews.llvm.org/D82174

MrSidims: Taking these numeric values out of scope for now, could please review reset of the patch?

lib/SPIRV/SPIRVWriter.cpp
```diff
@@ -756,14 +764,41 @@ SPIRV::SPIRVInstruction *LLVMToSPIRV::transUnaryInst(UnaryInstruction *U,
...
+      // SPV_INTEL_usm_storage_classes we can also have casts from global_device
+      // and global_host to global addr space and vice versa.
+    } else if (SrcAddrSpace == SPIRAS_GlobalDevice ||
+               SrcAddrSpace == SPIRAS_GlobalHost) {
+      assert(
+          (DestAddrSpace == SPIRAS_Global || DestAddrSpace == SPIRAS_Generic) &&
+          "Casts from global_device/global_host only allowed to \
+             global/generic");
+      if (!BM->isAllowedToUseExtension(
+              ExtensionID::SPV_INTEL_usm_storage_classes))
```
MrSidims, line 783: Here is a bug. We shan't emit any instructions here if we have an address space cast from global_device/host to global address space when the extension is disabled (it would lead to a cast from CrossWorkgroup storage cast pointer to  CrossWorkgroup storage cast pointer).

MrSidims:
@AlexeySotkin @AlexeySachkov ping

lib/SPIRV/SPIRVWriter.cpp
MrSidims: Fixed

lib/SPIRV/SPIRVInternal.h
MrSidims: @karolherbst
Hi, I see that you have added previously SPIRAS_Input and SPIRAS_Output into the map. Are they mapped to the actual address spaces in LLVM IR? If not I would love to shift them in the enum, claiming 5 and 6 to SPIRAS_GlobalDevice and SPIRAS_GlobalHost. The reason I want to do that is https://reviews.llvm.org/D82174 where I'm adding the appropriate address spaces in SPIR address space map. Currently the values assigned are 11 and 12, but it seems to be more logical to make a continues map without a gap from 4 to 11.
Thank you in advance.

karolherbst: I am not sure about the actual llvm mapping there, I essentially just added them to fix some spirv-val issues with generated shaders and I think there needs to be some further changes as 1.4 or 1.5 changed the rules further.

But as long as the generated spirvs are still valid I really don't mind any changes in this area.

MrSidims: Thanks! Looks like swapping has no affect (at least in local testing). Moved it here.

lib/SPIRV/SPIRVWriter.cpp
```diff
@@ -1052,8 +1093,18 @@ SPIRVValue *LLVMToSPIRV::transValueWithoutDecoration(Value *V,
     if (IsVectorCompute)
       StorageClass =
           VectorComputeUtil::getVCGlobalVarStorageClass(AddressSpace);
-    else
+    else {
+      // Lower global_device and global_host address spaces that were added in
+      // SYCL as part of SYCL_INTEL_usm_address_spaces extension to just global
+      // address space if device doesn't support SPV_INTEL_usm_storage_classes
+      // extension
+      if (!BM->isAllowedToUseExtension(
```
AlexeySachkov, line 1101: I think it is better to swap order: check for address space first, and only then for extension. The reason is that extension might be disabled all the time for other users and there is no sense to check for it unless you encountered something that requires it

lib/SPIRV/libSPIRV/SPIRVInstruction.h
```diff
@@ -538,9 +538,13 @@ class SPIRVStore : public SPIRVInstruction, public SPIRVMemoryAccess {
     SPIRVInstruction::validate();
     if (getSrc()->isForward() || getDst()->isForward())
       return;
-    assert(getValueType(PtrId)->getPointerElementType() ==
-               getValueType(ValId) &&
-           "Inconsistent operand types");
+    if (getValueType(PtrId)->getPointerElementType() != getValueType(ValId)) {
```
AlexeySachkov, line 541: No need to wrap assert with an `if`: I mean this might be needed if assertion is not correct under some circumstances, but in that case, please wrap everything into `#ifndef NDEBUG`

lib/SPIRV/libSPIRV/SPIRVModule.h
```diff
@@ -476,6 +476,10 @@ class SPIRVModule {
     return TranslationOpts.getFPContractMode();
   }
 
+  bool isUsmAddrspacesEnabled() {
```
AlexeySachkov, line 479:
```suggestion
  bool isUsmAddrspacesEnabled() const {
```

lib/SPIRV/SPIRVReader.cpp
AlexeySachkov: > we have here in the translator different flows for OpenCL 2.0 and OpenCL 1.2, right?

This is for sure shared between several vendors

> This option came from (surprise-surprise) some implementation details of SYCL compiler's driver and desire to make the feature adoptable not only for FPGA, but for other targets in JIT compilation (aka with triple spirv-unknown-unknown for which we add spirv-ext=+all option.

This is exactly what I'm talking about: device-/vendor-specific consumption rules are being moved into the translator. I think it is okay to outline common parts into the translator, but at the same time I wouldn't like to see bunch of vendor-specific options in this repo.

I don't have strong objections at this point: probably we can add this option as a first example and when we have the second one we can see how this can be generalized, where it goes to and how does it look like

AlexeySotkin: A SPIR-V module declares capabilities, i.e. features which are used in the module. A client API environment (aka target environment) should check capabilities declared in the module against capabilities it supports. If the client API doesn't support a capability declared in the module it's "allowed to reject the module."
The `EnableUsmAddrspaces` option looks like we are trying adapt a SPIR-V module with `USMStorageClassesINTEL` capability for client API(device) which doesn't support that capability. IMO a client API should **either support the capability or reject the module**.
I agree with @AlexeySachkov, we should not move device specific logic into the translator.

MrSidims: Ok. Do you have any comments other than this one?

lib/SPIRV/SPIRVReader.cpp
```diff
@@ -385,11 +385,13 @@ Type *SPIRVToLLVM::transType(SPIRVType *T, bool IsClassMember) {
...
-  case OpTypePointer:
+  case OpTypePointer: {
+    SPIRAddressSpace AS =
+        SPIRSPIRVAddrSpaceMap::rmap(T->getPointerStorageClass());
     return mapType(
         T, PointerType::get(
-               transType(T->getPointerElementType(), IsClassMember),
-               SPIRSPIRVAddrSpaceMap::rmap(T->getPointerStorageClass())));
+               transType(T->getPointerElementType(), IsClassMember), AS));
+  }
```
AlexeySotkin, line 394: I guess this change is not needed anymore

lib/SPIRV/SPIRVReader.cpp
```diff
@@ -931,8 +933,19 @@ Value *SPIRVToLLVM::transConvertInst(SPIRVValue *BV, Function *F,
   switch (BC->getOpCode()) {
   case OpPtrCastToGeneric:
   case OpGenericCastToPtr:
+  case OpPtrCastToCrossWorkgroupINTEL:
+  case OpCrossWorkgroupCastToPtrINTEL: {
+    // If module has pointers with DeviceOnlyINTEL and HostOnlyINTEL storage
+    // classes and if during reversed translation 'enable-usm-addrspaces'
+    // wasn't passed there will be a situation, when global_device/global_host
+    // address space will be lowered to just global address space. If there also
+    // is an addrspacecast - we need to replace it with source pointer.
```
AlexeySotkin, line 942: I think this comment should be updated.

lib/SPIRV/SPIRVReader.cpp
```diff
@@ -3358,15 +3371,15 @@ bool SPIRVToLLVM::transOCLMetadata(SPIRVFunction *BF) {
...
   addOCLKernelArgumentMetadata(
       Context, SPIR_MD_KERNEL_ARG_ADDR_SPACE, BF, F,
       [=](SPIRVFunctionParameter *Arg) {
         SPIRVType *ArgTy = Arg->getType();
         SPIRAddressSpace AS = SPIRAS_Private;
-        if (ArgTy->isTypePointer())
+        if (ArgTy->isTypePointer()) {
           AS = SPIRSPIRVAddrSpaceMap::rmap(ArgTy->getPointerStorageClass());
-        else if (ArgTy->isTypeOCLImage() || ArgTy->isTypePipe())
+        } else if (ArgTy->isTypeOCLImage() || ArgTy->isTypePipe())
```
AlexeySotkin, line 3382: Brace are not needed.

lib/SPIRV/SPIRVWriter.cpp
```diff
@@ -756,14 +764,47 @@ SPIRV::SPIRVInstruction *LLVMToSPIRV::transUnaryInst(UnaryInstruction *U,
...
+          "Casts from global_device/global_host only allowed to \
+             global/generic");
+      if (!BM->isAllowedToUseExtension(
+              ExtensionID::SPV_INTEL_usm_storage_classes)) {
+        if (DestAddrSpace == SPIRAS_Global)
+          return nullptr;
+        BOC = OpPtrCastToGeneric;
+      } else {
+        BOC = OpPtrCastToCrossWorkgroupINTEL;
+      }
```
AlexeySotkin, line 789: Minor: this probably can (but not necessarily) be factored out to a function or lambda.

lib/SPIRV/SPIRVWriter.cpp
```diff
@@ -1311,7 +1362,8 @@ SPIRVValue *LLVMToSPIRV::transValueWithoutDecoration(Value *V,
   if (UnaryInstruction *U = dyn_cast<UnaryInstruction>(V)) {
     if (isSpecialTypeInitializer(U))
       return mapValue(V, transValue(U->getOperand(0), BB));
-    return mapValue(V, transUnaryInst(U, BB));
+    auto UI = transUnaryInst(U, BB);
+    return mapValue(V, UI ? UI : transValue(U->getOperand(0), BB));
```
AlexeySotkin, line 1366: Should `transUnaryInst`  return `transValue(U->getOperand(0), BB)` instead of `nullptr`?

MrSidims: `transUnaryInst` returns `SPIRVInstuction *`, while `transValue` returns `SPIRVValue *`, so it won't work.

AlexeySotkin:
The commit message should be updated. Also it would be great if you squash your commits. Thanks.

lib/SPIRV/SPIRVReader.cpp
```diff
@@ -931,8 +931,18 @@ Value *SPIRVToLLVM::transConvertInst(SPIRVValue *BV, Function *F,
...
+  case OpPtrCastToCrossWorkgroupINTEL:
+  case OpCrossWorkgroupCastToPtrINTEL: {
+    // If module has pointers with DeviceOnlyINTEL and HostOnlyINTEL storage
+    // classes there will be a situation, when global_device/global_host
+    // address space will be lowered to just global address space. If there also
+    // is an addrspacecast - we need to replace it with source pointer.
+    if (Src->getType()->getPointerAddressSpace() ==
+        Dst->getPointerAddressSpace())
+      return Src;
+    CO = Instr
```
AlexeySotkin, line 943: `Instr` ?

MrSidims: Typo, thanks!

MrSidims:
> I'd remove the second paragraph of the commit message.
> Other than that, LGTM. Thanks!

Okay, thank you and @AlexeySachkov for the review!

PR #581, opened 2020-06-09
Emit error if WordCount exceeds 16-bit max value

AlexeySotkin:
According to the SPIR-V specification(Table 2) the 16 high-order bits of every
instruction encode word count(length) of the instruction. This fundamental
limitation doesn't allow to have instruction longer than 65535 words.


test/word_count_limit.ll
```diff
@@ -0,0 +1,22 @@
+; RUN: llvm-as %s -o %t.bc
```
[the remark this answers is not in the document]
AlexeySotkin: Done

lib/SPIRV/SPIRVWriter.cpp
```diff
@@ -993,6 +993,21 @@ SPIRVValue *LLVMToSPIRV::transValueWithoutDecoration(Value *V,
...
+        // 2) Result Type and 3) Result Id.
+        // Max length of SPIRV instruction = 65535 words.
+        const int maxNumElements = 65535 - 3;
+        if (ArrTy->getNumElements() > maxNumElements &&
+            !isa<ConstantAggregateZero>(Init)) {
+          std::stringstream SS;
+          SS << "Global variable has a constant array initializer with number "
+             << "of elements greater than OpConstantComposite can have(65532)."
+             << " Should the array be splitted?\n"
+             << "Original LLVM value:\n" << toString(GV);
```
svenvh, line 1007:
```suggestion
          SS << "Global variable has a constant array initializer with a number "
             << "of elements greater than OpConstantComposite can have (65532)."
             << " Should the array be split?\n"
             << "Original LLVM value:\n" << toString(GV);
```

test/negative/word_count_limit.ll
```diff
@@ -0,0 +1,24 @@
+; RUN: llvm-as %s -o %t.bc
+; RUN: not llvm-spirv %t.bc -o %t.spv 2>&1 | FileCheck %s --check-prefix=CHECK-ERROR
+
+; CHECK-ERROR: InvalidWordCount: Can't encode instruction with word count greater than 65535:
+; CHECK-ERROR-NEXT: Global variable has a constant array initializer with number of elements greater than OpConstantComposite can have(65532). Should the array be splitted?
```
svenvh, line 5:
```suggestion
; CHECK-ERROR-NEXT: Global variable has a constant array initializer with a number of elements greater than OpConstantComposite can have (65532). Should the array be split?
```

lib/SPIRV/SPIRVWriter.cpp
AlexeySotkin: Fixed

test/negative/word_count_limit.ll
AlexeySotkin: Fixed

PR #589, opened 2020-06-15
Handle LLVM function aliases

fadeeval:
In case separate constructor declaration and definition the IR representation has GlobalAlias objects which didn't process properly.

Signed-off-by: Aleksander Fadeev <aleksander.fadeev@intel.com>

### DIFF
--- a/lib/SPIRV/SPIRVWriter.cpp
+++ b/lib/SPIRV/SPIRVWriter.cpp
@@ -1423,8 +1423,13 @@ SPIRVValue *LLVMToSPIRV::transValueWithoutDecoration(Value *V,
     if (BM->isAllowedToUseExtension(ExtensionID::SPV_INTEL_inline_assembly))
       return mapValue(V, transAsmINTEL(IA));
 
-  if (CallInst *CI = dyn_cast<CallInst>(V))
+  if (CallInst *CI = dyn_cast<CallInst>(V)) {
+    if (auto Alias =
+            dyn_cast_or_null<llvm::GlobalAlias>(CI->getCalledOperand())) {
+      CI->setCalledFunction(cast<Function>(Alias->getAliasee()));
+    }
     return mapValue(V, transCallInst(CI, BB));
+  }
 
   llvm_unreachable("Not implemented");
   return nullptr;
--- a/test/custom_class.ll
+++ b/test/custom_class.ll
@@ -1,0 +1,100 @@
+; RUN: llvm-as %s -o %t.bc
+; RUN: llvm-spirv %t.bc
+
+; --- Source code ---
+; Generated with "intel/llvm/clang++ -fsycl-device-only  -emit-llvm -S"
+;
+;#include <CL/sycl.hpp>
+;
+;class CustomClass {
+;public:
+;   CustomClass(int value);
+;private:
+;   int data;
+;}; 
+;
+;CustomClass::CustomClass(int value) : data(value) {}
+;
+;int main() {
+;   cl::sycl::queue queue;
+;   queue.submit( [&]( cl::sycl::handler& handler ) {
+;         handler.single_task< class SYCLCustom >([=]() {
+;               CustomClass dummyClass(1);
+;            } );
+;      } );
+;
+;   return 0;
+;}
+
+; ModuleID = 'custom_class.ll'
+target datalayout = "e-i64:64-v16:16-v24:32-v32:32-v48:64-v96:128-v192:256-v256:256-v512:512-v1024:1024-n8:16:32:64"
+target triple = "spir64-unknown-linux-sycldevice"
+
+%class._ZTS11CustomClass.CustomClass = type { i32 }
+%"class._ZTSZZ4mainENK3$_0clERN2cl4sycl7handlerEEUlvE_.anon" = type { i8 }
+
+$"_ZTSZZ4mainENK3$_0clERN2cl4sycl7handlerEE10SYCLCustom" = comdat any
+
+@_ZN11CustomClassC1Ei = dso_local unnamed_addr alias void (%class._ZTS11CustomClass.CustomClass addrspace(4)*, i32), void (%class._ZTS11CustomClass.CustomClass addrspace(4)*, i32)* @_ZN11CustomClassC2Ei
+
+; Function Attrs: norecurse
+define weak_odr dso_local spir_kernel void @"_ZTSZZ4mainENK3$_0clERN2cl4sycl7handlerEE10SYCLCustom"() #0 comdat !kernel_arg_addr_space !4 !kernel_arg_access_qual !4 !kernel_arg_type !4 !kernel_arg_base_type !4 !kernel_arg_type_qual !4 {
+entry:
+  %0 = alloca %"class._ZTSZZ4mainENK3$_0clERN2cl4sycl7handlerEEUlvE_.anon", align 1
+  %1 = bitcast %"class._ZTSZZ4mainENK3$_0clERN2cl4sycl7handlerEEUlvE_.anon"* %0 to i8*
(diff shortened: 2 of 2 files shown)


test/custom_class.ll
```diff
@@ -0,0 +1,512 @@
+; RUN: llvm-spirv %s -to-binary -o %t.spv
```
mlychkov, line 1: Can this IR be reduced to the smallest subset of instructions which still reproduces the crash? There are a lot of SYCL library methods which are confusing.

fadeeval: The problem is reproducible only with -mem2reg optimization.

test/custom_class.ll
```diff
@@ -0,0 +1,75 @@
+; RUN: llvm-as %s -o %t.bc
+; RUN: llvm-spirv %t.bc
+; ModuleID = 'custom_class.ll'
+source_filename = "/localdisk2/icl/fadeeval/sycl_workspace/repo2/custom_class.cpp"
```
mlychkov, line 4: Please, remove this line because it is useless for test and discloses unnecessary information.
Could you please include cpp test file contents in comments and command that you used for creation of this IR?
You can see an example in https://github.com/KhronosGroup/SPIRV-LLVM-Translator/blob/master/test/DebugInfo/X86/dwarf-aranges-no-dwarf-labels.ll#L14

test/custom_class.ll
```diff
@@ -0,0 +1,75 @@
...
+
+; Function Attrs: norecurse nounwind
+define dso_local spir_func void @_ZN11CustomClassC2Ei(%class._ZTS11CustomClass.CustomClass addrspace(4)* %this, i32 %value) unnamed_addr #3 align 2 {
+entry:
+  %data = getelementptr inbounds %class._ZTS11CustomClass.CustomClass, %class._ZTS11CustomClass.CustomClass addrspace(4)* %this, i32 0, i32 0
+  store i32 %value, i32 addrspace(4)* %data, align 4, !tbaa !5
+  ret void
+}
+
+attributes #0 = { norecurse "correctly-rounded-divide-sqrt-fp-math"="false" "disable-tail-calls"="false" "frame-pointer"="all" "less-precise-fpmad"="false" "min-legal-vector-width"="0" "no-infs-fp-math"="false" "no-jump-tables"="false" "no-nans-fp-math"="false" "no-signed-zeros-fp-math"="false" "no-trapping-math"="false" "stack-protector-buffer-size"="8" "sycl-module-id"="/localdisk2/icl/fadeeval/sycl_workspace/repo2/custom_class.cpp" "uniform-work-group-size"="true" "unsafe-fp-math"="false" "use-soft-float"="false" }
```
mlychkov, line 55:
```suggestion
attributes #0 = { norecurse "correctly-rounded-divide-sqrt-fp-math"="false" "disable-tail-calls"="false" "frame-pointer"="all" "less-precise-fpmad"="false" "min-legal-vector-width"="0" "no-infs-fp-math"="false" "no-jump-tables"="false" "no-nans-fp-math"="false" "no-signed-zeros-fp-math"="false" "no-trapping-math"="false" "stack-protector-buffer-size"="8" "uniform-work-group-size"="true" "unsafe-fp-math"="false" "use-soft-float"="false" }
```

test/custom_class.ll
```diff
@@ -0,0 +1,75 @@
...
+
+!llvm.module.flags = !{!0}
+!opencl.spir.version = !{!1}
+!spirv.Source = !{!2}
+!llvm.ident = !{!3}
+
+!0 = !{i32 1, !"wchar_size", i32 4}
+!1 = !{i32 1, i32 2}
+!2 = !{i32 4, i32 100000}
+!3 = !{!"clang version 11.0.0 (https://github.com/intel/llvm a33ffb5ae3874b89c88f13b6cdf753ca8444c2d6)"}
```
mlychkov, line 69:
```suggestion
!3 = !{!"clang version 11.0.0"}
```

test/custom_class.ll
```diff
@@ -0,0 +1,75 @@
...
+
+%class._ZTS11CustomClass.CustomClass = type { i32 }
+%"class._ZTSZZ4mainENK3$_0clERN2cl4sycl7handlerEEUlvE_.anon" = type { i8 }
+
+$"_ZTSZZ4mainENK3$_0clERN2cl4sycl7handlerEE10SYCLCustom" = comdat any
+
+@_ZN11CustomClassC1Ei = dso_local unnamed_addr alias void (%class._ZTS11CustomClass.CustomClass addrspace(4)*, i32), void (%class._ZTS11CustomClass.CustomClass addrspace(4)*, i32)* @_ZN11CustomClassC2Ei
+
+; Function Attrs: norecurse
+define weak_odr dso_local spir_kernel void @"_ZTSZZ4mainENK3$_0clERN2cl4sycl7handlerEE10SYCLCustom"() #0 comdat !kernel_arg_addr_space !4 !kernel_arg_access_qual !4 !kernel_arg_type !4 !kernel_arg_base_type !4 !kernel_arg_type_qual !4 {
```
mlychkov, line 16: Could this function be also deleted? Or without it this test doesn't crash?

fadeeval: The bug is because of this function so I can't delete it.

mlychkov: Ok then.

test/custom_class.ll
```diff
@@ -0,0 +1,101 @@
+; RUN: llvm-as %s -o %t.bc
+; RUN: llvm-spirv %t.bc
+; ModuleID = 'custom_class.ll'
```
mlychkov, line 3:
```suggestion
```
It duplicates the line 30.

fadeeval: I forgot to delete one of them.

test/custom_class.ll
mlychkov: I found that the crash can be reproduced if you delete this function and remove "internal" specifier from definition of "_ZZZ4mainENK3$_0clERN2cl4sycl7handlerEENKUlvE_clEv" function.
However, I don't know if this is a valid modification (removing of internal linkage) for current test case.

lib/SPIRV/SPIRVWriter.cpp
```diff
@@ -1423,8 +1423,13 @@ SPIRVValue *LLVMToSPIRV::transValueWithoutDecoration(Value *V,
     if (BM->isAllowedToUseExtension(ExtensionID::SPV_INTEL_inline_assembly))
       return mapValue(V, transAsmINTEL(IA));
 
-  if (CallInst *CI = dyn_cast<CallInst>(V))
+  if (CallInst *CI = dyn_cast<CallInst>(V)) {
+    if (auto Alias =
+            dyn_cast_or_null<llvm::GlobalAlias>(CI->getCalledOperand())) {
+      CI->setCalledFunction(dyn_cast<Function>(Alias->getAliasee()));
```
AlexeySachkov, line 1429:
```suggestion
      CI->setCalledFunction(cast<Function>(Alias->getAliasee()));
```

test/custom_class.ll
```diff
@@ -0,0 +1,100 @@
+; RUN: llvm-as %s -o %t.bc
+; RUN: llvm-spirv %t.bc
+
+; --- Source code ---
+; Generated with "clang++ -emit-llvm -S"
```
AlexeySachkov, line 5: `clang++ -emit-llvm -S` won't produce `spir64` triple, and for sure it won't produce `sycldevice` environment in the triple.

I guess that you used `clang++ -fsycl -fsycl-device-only -emit-llvm -S` or something like this. Also it is probably better to state that you used clang from intel/llvm project and not from llvm/llvm-project

svenvh:
LGTM, but please make sure the squashed commit has a descriptive commit message (it seems you have only changed the title of the PR so far?).

PR #594, opened 2020-06-17
Fix Execution Mode: VecTypeHint Encoding for Scalar Types

FranklandJack:
* For a VecTypeHint with a scalar type e.g. `char`, `int`, `float`,
`double` etc. the vector length is encoded as
`0` rather than `1`. This seems inconsistent with the SPIRV spec which
states *The 16 high-order bits of the Vector Type operand specify the
number of components of the vector.*. Currently the upper 16 bits are
encoded as `0x0000` rather than `0x0001`. An example of this behavior can
be found when using `llvm-spirv` to translate the IR produced from an
OpenCL
kernel with `__attribute((vec_type_hint(char)))`.
This will produce SPIRV with `OpExecutionMode %4 VecTypeHint 0x0`
rather than `OpExecutionMode %4 VecTypeHint 0x10000`.

- [x] Fix the encoding/decoding so that the vector length of scalar hints
is encoded as `1` rather than `0`. e.g. for OpenCL
`__attribute((vec_type_hint(char)))` would be encoded as `(1 << 16) |
0`.
- [x] Add tests to check the correct scalar encoding.

### DIFF
--- a/lib/SPIRV/OCLUtil.cpp
+++ b/lib/SPIRV/OCLUtil.cpp
@@ -248,30 +248,31 @@ void decodeMDNode(MDNode *N, unsigned &X, unsigned &Y, unsigned &Z) {
 
 /// Encode LLVM type by SPIR-V execution mode VecTypeHint
 unsigned encodeVecTypeHint(Type *Ty) {
+  unsigned Size = 1;
   if (Ty->isHalfTy())
-    return 4;
+    return 4 | (Size << 16);
   if (Ty->isFloatTy())
-    return 5;
+    return 5 | (Size << 16);
   if (Ty->isDoubleTy())
-    return 6;
+    return 6 | (Size << 16);
   if (IntegerType *IntTy = dyn_cast<IntegerType>(Ty)) {
     switch (IntTy->getIntegerBitWidth()) {
     case 8:
-      return 0;
+      return 0 | (Size << 16);
     case 16:
-      return 1;
+      return 1 | (Size << 16);
     case 32:
-      return 2;
+      return 2 | (Size << 16);
     case 64:
-      return 3;
+      return 3 | (Size << 16);
     default:
       llvm_unreachable("invalid integer type");
     }
   }
   if (VectorType *VecTy = dyn_cast<VectorType>(Ty)) {
     Type *EleTy = VecTy->getElementType();
-    unsigned Size = VecTy->getNumElements();
-    return Size << 16 | encodeVecTypeHint(EleTy);
+    Size = VecTy->getNumElements();
+    return Size << 16 | (encodeVecTypeHint(EleTy) & 0xFF);
   }
   llvm_unreachable("invalid type");
   return ~0U;
@@ -301,7 +302,7 @@ Type *decodeVecTypeHint(LLVMContext &C, unsigned Code) {
     llvm_unreachable("Invalid vec type hint");
     return nullptr;
   }
-  if (VecWidth < 1)
+  if (VecWidth == 1)
     return ST;
   return VectorType::get(ST, VecWidth);
 }
--- a/test/transcoding/vec_type_hint_scalar.cl
+++ b/test/transcoding/vec_type_hint_scalar.cl
@@ -1,0 +1,72 @@
+// RUN: %clang_cc1 -triple spir64-unknown-unknown -x cl -cl-std=CL2.0 -O0 -fdeclare-opencl-builtins -finclude-default-header -emit-llvm-bc %s -o %t.bc
+// RUN: llvm-spirv %t.bc -spirv-text -o %t.spt
+// RUN: FileCheck < %t.spt %s --check-prefix=CHECK-SPIRV
+// RUN: llvm-spirv %t.bc -o %t.spv
+// RUN: spirv-val %t.spv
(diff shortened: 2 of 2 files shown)


svenvh:
>     * For a VecTypeHint with a scalar type e.g. `char`, `int`, `float`,
>       `double` etc. the vector length is encoded as
>       `0` rather than `1`. This seems inconsistent with the SPIRV spec which
>       states _The 16 high-order bits of the Vector Type operand specify the
>       number of components of the vector._.

Perhaps we need to clarify this with the SPIR-V spec itself first.  The Types section (s2.2.2) states: "_Vector_: An ordered homogeneous collection of two or more _scalars_.", so it sounds a bit odd to allow a vector length of `1` for `OpVecTypeHint`?

FranklandJack:
> > ```
> > * For a VecTypeHint with a scalar type e.g. `char`, `int`, `float`,
> >   `double` etc. the vector length is encoded as
> >   `0` rather than `1`. This seems inconsistent with the SPIRV spec which
> >   states _The 16 high-order bits of the Vector Type operand specify the
> >   number of components of the vector._.
> > ```
> 
> Perhaps we need to clarify this with the SPIR-V spec itself first. The Types section (s2.2.2) states: "_Vector_: An ordered homogeneous collection of two or more _scalars_.", so it sounds a bit odd to allow a vector length of `1` for `OpVecTypeHint`?

Thank you for your feedback. I agree, it does seem odd. Even having a vector hint with a scalar argument / vector of length 1  in the first place seems strange to me. We do currently allow for vectors of one element though, this change only changes how they are encoded.

svenvh:
> We do currently allow for vectors of one element though

Maybe that needs reconsidering, as I am not sure if emitting an operand representing a scalar/1-element vector should be allowed or not. Unfortunately the spec isn't explicit about this and `spirv-val` currently doesn't seem to do any validation on the `OpVecTypeHint` operand.

Wondering if others have any opinion on this?

FranklandJack:
I am happy to update the PR to remove scalar hints completely, they would then be lost in an LLVM->SPIRV->LLVM translation though, I don't know if that's an issue?

> Wondering if others have any opinion on this?
I'm not sure who would be an appropriate developer to tag? This is my first contribution to this project :)

svenvh:
> I am happy to update the PR to remove scalar hints completely, they would then be lost in an LLVM->SPIRV->LLVM translation though, I don't know if that's an issue?

Looking at the OpenCL documentation of the `vec_type_hint` attribute, the default is `vec_type_hint(int)`, i.e. a scalar int; we probably don't want to lose support for the default case!  The OpenCL spec also explicitly allows scalar types, so we should probably keep scalar hints.

So the question is whether we should use `0` or `1` for a scalar `vec_type_hint`.  Is there a particular problem you are facing due to the use of `0`?  As far as I can tell, with the current implementation we can still support scalar operands, but perhaps I am missing something.

We should probably see if we can clarify the scalar case for `OpVecTypeHint` in the SPIR-V specification with Khronos.

FranklandJack:
> So the question is whether we should use 0 or 1 for a scalar vec_type_hint.

Since the SPIR-V spec says:

> The 16 high-order bits of the Vector Type operand specify the number of components of the vector. 

I would have thought conformant encoding would have scalar hints as vectors of length 1 so that the upper 16 bits were `0x0001`, a scalar hint being a vector of one component.

svenvh:
> > The 16 high-order bits of the Vector Type operand specify the number of components of the vector.
> 
> I would have thought conformant encoding would have scalar hints as vectors of length 1 so that the upper 16 bits were `0x0001`, a scalar hint being a vector of one component.

That doesn't sound unreasonable, although SPIR-V doesn't seem to allow vectors of one component.  In various places (e.g. s2.2.2 "Vectors",  s3.36.6 "OpTypeVector") there is an explicit requirement to have at least two components for something to be considered a vector, which is why I am hesitant to introduce a 1-element vector representation here.  Encoding the scalar as `0` elements doesn't quite fulfill the `>=2` requirement either, but it might better convey the intent that "this has 0 components, so it's not a vector"?

FranklandJack:
Ah right, I see your point, it does seem weird to allow a vector of length one in this context if they are required to have `>=2` components by the spec. In this case, I think your suggestion of clarifying the scalar case for `OpVecTypeHint` in the SPIR-V spec is a good one.

For some context, we have our own consumer of SPIR-V that assumed a vector a length 1 would be a scalar (the length 0 case wasn't handled) which we have now changed to treat 0 as the scalar case, so this isn't urgent. It looked from the definition of `OpVecTypeHint` that this was a mistake in the SPIR-V being produced, but we didn't consider the possiblity it could be deliberately encoded as length zero to avoid having vectors of one component.

svenvh:
> I think your suggestion of clarifying the scalar case for OpVecTypeHint in the SPIR-V spec is a good one.

Thanks, I have raised this with Khronos.  I will update this thread if there is any news.

FranklandJack:
Great, thank you.

dneto0:
> Unfortunately the spec isn't explicit about this and spirv-val currently doesn't seem to do any validation on the OpVecTypeHint operand.

The SPIR-V Tools validator does almost nothing for Kernel-specific modules (due to lack of contribution).

Issue #506, opened 2020-04-23
Usages of VectorType::getNumElements should be updated


AlexeySachkov:
There is upcoming change which removes `getNumElements` method from `VectorType` class. Instead, we should use `FixedVectorType` class to get number of elements in vector.

See [this thread on llvm-dev](http://lists.llvm.org/pipermail/llvm-dev/2020-April/141162.html) and [D78127](https://reviews.llvm.org/D78127)

PR #597, opened 2020-06-23
Use LLVM FixedVectorType::get

svenvh:
Update for commit eb81c85afdc ("[SVE] Deprecate default false variant
of VectorType::get", 2020-06-16).


svenvh:
> This doesn't fully resolve #506, right?

That's correct, this PR only updates creation and not `getNumElements()` uses.

PR #600, opened 2020-06-25
Add common OCL address spaces for VectorCompute globals

NikitaRudenkoIntel:
This commit allows to use UniformConstant and CrossWorkgroup
storage classes for VectorCompute globals

### DIFF
--- a/lib/SPIRV/VectorComputeUtil.cpp
+++ b/lib/SPIRV/VectorComputeUtil.cpp
@@ -120,6 +120,10 @@ getVCGlobalVarStorageClass(SPIRAddressSpace AddressSpace) noexcept {
     return StorageClassPrivate;
   case SPIRAS_Local:
     return StorageClassWorkgroup;
+  case SPIRAS_Global:
+    return StorageClassCrossWorkgroup;
+  case SPIRAS_Constant:
+    return StorageClassUniformConstant;
   default:
     assert(false && "Unexpected address space");
     return StorageClassPrivate;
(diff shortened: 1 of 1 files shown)


AlexeySotkin:
I think the switch can be replaced with `SPIRVMap`

NikitaRudenkoIntel:
No, I do not think so. I need a map between address space and storage classes. But SPIRVMap emulates static behavior, which means that there cannot be 2 maps with same types. And there is already such map in SPIRVInternal.h:216

lib/SPIRV/VectorComputeUtil.cpp
```diff
@@ -120,6 +120,10 @@ getVCGlobalVarStorageClass(SPIRAddressSpace AddressSpace) noexcept {
     return StorageClassPrivate;
   case SPIRAS_Local:
     return StorageClassWorkgroup;
+  case SPIRAS_Global:
+    return StorageClassCrossWorkgroup;
+  case SPIRAS_Constant:
+    return StorageClassUniformConstant;
```
AlexeySachkov, line 126: Note: this will most likely require `Shader` capability to be declared in SPIR-V module

NikitaRudenkoIntel: Hmm... I hope it is not. At least in the spec, the lists of enabling capabilities for StorageClassUniformConstant and StorageClassCrossWorkgroup are empty...

AlexeySachkov: Oh, my guess was incorrect then, sorry for confusion

PR #609, opened 2020-07-03
Fix crash on debug info when -spirv-allow-unknow-intrinsics option enabled.

NikitaRudenkoIntel:
Bug is caused by translating of known intrinsics declarations, which should not be translated.

### DIFF
--- a/lib/SPIRV/SPIRVWriter.cpp
+++ b/lib/SPIRV/SPIRVWriter.cpp
@@ -505,7 +505,8 @@ SPIRVFunction *LLVMToSPIRV::transFunctionDecl(Function *F) {
   if (auto BF = getTranslatedValue(F))
     return static_cast<SPIRVFunction *>(BF);
 
-  if (F->isIntrinsic() && !SPIRVAllowUnknownIntrinsics) {
+  if (F->isIntrinsic() &&
+      (!SPIRVAllowUnknownIntrinsics || isKnownIntrinsic(F->getIntrinsicID()))) {
     // We should not translate LLVM intrinsics as a function
     assert(none_of(F->user_begin(), F->user_end(),
                    [this](User *U) { return getTranslatedValue(U); }) &&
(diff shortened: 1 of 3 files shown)


Hopetech:
Nit: In your commit message s/unknow/unknown/

lib/SPIRV/SPIRVWriter.cpp
```diff
@@ -1829,6 +1862,8 @@ SPIRVValue *LLVMToSPIRV::transIntrinsicInst(IntrinsicInst *II,
     return MemoryAccess;
   };
 
+  // Before adding intrinsic here, consider registering it at
```
AlexeySotkin, line 1865: I think with this patch we have to "register" known intrinsics at the two places at the same time.

NikitaRudenkoIntel: Yes. I did not find a satisfactory solution without this.

lib/SPIRV/SPIRVWriter.cpp
```diff
@@ -1808,6 +1809,38 @@ void addIntelFPGADecorationsForStructMember(
   }
 }
 
+bool LLVMToSPIRV::isTranslateIntrinsicDecl(Intrinsic::ID Id) {
```
AlexeySotkin, line 1812:
```suggestion
bool LLVMToSPIRV::isKnownIntrinsic(Intrinsic::ID Id) {
```

lib/SPIRV/SPIRVWriter.cpp
```diff
@@ -1829,6 +1862,8 @@ SPIRVValue *LLVMToSPIRV::transIntrinsicInst(IntrinsicInst *II,
     return MemoryAccess;
   };
 
+  // Before adding intrinsic here, consider registering it at
+  // isTranslateIntrinsicDecl function
```
AlexeySotkin, line 1866:
```suggestion
  // LLVM intrinsics with known translation to SPIR-V are handled here. They also must be registered
  // at isKnownIntrinsic function in order to make -spirv-allow-unknown-intrinsics work correctly.
```

AlexeySotkin:
Please fix clang-format https://travis-ci.org/github/KhronosGroup/SPIRV-LLVM-Translator/jobs/706129500#L592 and squash your commits.
Other than that LGTM. Thanks!

lib/SPIRV/SPIRVWriter.cpp
```diff
@@ -1808,6 +1809,38 @@ void addIntelFPGADecorationsForStructMember(
...
+  case Intrinsic::dbg_declare:
+  case Intrinsic::dbg_value:
+  case Intrinsic::annotation:
+  case Intrinsic::var_annotation:
+  case Intrinsic::ptr_annotation:
+  case Intrinsic::invariant_start:
+  case Intrinsic::invariant_end:
+  case Intrinsic::dbg_label:
+  case Intrinsic::trap:
+    return false;
```
svenvh, line 1837: Isn't this contradicting the function name?  For all of these intrinsics, I would expect a function "isKnownIntrinsic" to return `true` because they are known.

NikitaRudenkoIntel: Ah... Will fix. It happened because previously this function was called "isTranslateIntrinsicDecl"

NikitaRudenkoIntel: fixed

PR #612, opened 2020-07-07
[SYCL][FPGA] Translate the LSU controls built-in

AGindinson:
The LSU controls (interchangeable with "memory access controls") apply to a pointer variable; in LLVM IR, this is expressed via `llvm.ptr.annotation` intrinsics:
`T* llvm.ptr.annotation.*(T* %, "{params:}{cache-size:..}")`.

As far as the annotation algorithm is concerned, LSU controls go down the similar path as FPGA memory attributes. The primary                                                                                                                 differences are:
1) the "on/off" parameters are encoded in a bitmask. Meanwhile, auxilary parameters like LSU cache size should be found in additional annotation blocks, if enabled by a corresponding flag;
2) in SPIR-V, the memory access decorations are applied to the results of particular load instructions/pointer access instructions, as opposed to memory attributes, which are applied to static entities (static variables/struct members).

The "Intel FPGA memory accesses" SPIR-V extension specification can be found at https://github.com/KhronosGroup/SPIRV-Registry/blob/master/extensions/INTEL/SPV_INTEL_fpga_memory_accesses.html


AGindinson:
Tagging @mlychkov @AlexeySotkin @MrSidims @vmaksimo

test/transcoding/IntelFPGAMemoryAccesses.ll
```diff
@@ -0,0 +1,218 @@
...
+@.str.1 = private unnamed_addr constant [14 x i8] c"<invalid loc>\00", section "llvm.metadata"
+; "params" bitmask doesn't hold 0x2 (cache size ON), so cache-size can be dropped
+; during translation
+; CHECK-LLVM: [[PARAM_12_CACHE_0:@\.str[0-9_.]*]] = {{.*}}{params:12}
+@.str.2 = private unnamed_addr constant [26 x i8] c"{params:12}{cache-size:0}\00", section "llvm.metadata"
+; CHECK-LLVM: [[PARAM_2_CACHE_127:@\.str[0-9_.]*]] = {{.*}}{params:2}{cache-size:127}
+@.str.3 = private unnamed_addr constant [27 x i8] c"{params:2}{cache-size:127}\00", section "llvm.metadata"
+; CHECK-LLVM: [[PARAM_3_CACHE_127:@\.str[0-9_.]*]] = {{.*}}{params:3}{cache-size:127}
+@.str.4 = private unnamed_addr constant [27 x i8] c"{params:3}{cache-size:127}\00", section "llvm.metadata"
+; Since "params" bitmask is set to 0, the next string isn't required to be preserved
```
MrSidims, line 74: Do we need to CHECK-NOT something here then?

test/transcoding/IntelFPGAMemoryAccesses.ll
```diff
@@ -0,0 +1,218 @@
...
+target datalayout = "e-i64:64-v16:16-v24:32-v32:32-v48:64-v96:128-v192:256-v256:256-v512:512-v1024:1024-n8:16:32:64"
+target triple = "spir64-unknown-unknown-sycldevice"
+
+%"class._ZTSZ4mainE3$_0.anon" = type { i8 }
+%struct._ZTS5State.State = type { float, i32 }
+
+; CHECK-LLVM: [[PARAM_3_CACHE_0:@\.str[0-9_.]*]] = {{.*}}{params:3}{cache-size:0}
+@.str = private unnamed_addr constant [25 x i8] c"{params:3}{cache-size:0}\00", section "llvm.metadata"
+@.str.1 = private unnamed_addr constant [14 x i8] c"<invalid loc>\00", section "llvm.metadata"
+; "params" bitmask doesn't hold 0x2 (cache size ON), so cache-size can be dropped
```
MrSidims, line 66: I think this logic shall be mirrored to the translator sources.

test/transcoding/IntelFPGAMemoryAccesses.ll
```diff
@@ -0,0 +1,218 @@
...
+; struct State {
+;  float Field1;
+;  int Field2;
+;  };
+;
+; void foo(float *A, int *B, State *C) {
+;   float *x;
+;   int *y;
+;   State *z;
+;   x = __builtin_intel_fpga_mem(A, 1 | 2, 0);
```
MrSidims, line 13: Please add comments what are 1, 2, 4 and 8 here.

lib/SPIRV/SPIRVWriter.cpp
```diff
@@ -1639,13 +1639,56 @@ SPIRVValue *LLVMToSPIRV::oclTransSpvcCastSampler(CallInst *CI,
...
+    int CacheSize = 0;
+    if (CacheSizeInfo.hasValue()) {
+      CacheSize = CacheSizeInfo.getValue();
+      auto DecorationString = std::to_string(CacheSize);
+      ResultVec.emplace_back(DecorationCacheSizeINTEL, DecorationString);
+    }
+    int PrefetchSize = 0;
+    if (PrefetchInfo.hasValue()) {
+      PrefetchSize = PrefetchInfo.getValue();
+      ResultVec.emplace_back(DecorationPrefetchINTEL,
```
MrSidims, line 1673: I like this version more, then one for `CacheSize` from above. But what about making it be a one-liner? (getting rid of temp variables).

test/transcoding/IntelFPGAMemoryAccesses.ll
AGindinson: It actually is - `{cache-size:..}` isn't generated when the `CacheSizeINTEL` decoration is not discovered in SPIR-V.

test/transcoding/IntelFPGAMemoryAccesses.ll
AGindinson: As it stands, we'd better not: the obsolete annotation still gets translated forwards and backwards as `annotate` would be, without affecting SPIR-V semantics. Moreover, the FPGA BE is going to disregard the obsolete annotations, thus no stability issues are expected.

Since this is more of an implementation detail and not a requirement, I chose to go without positive or negative checks for this particular case.

lib/SPIRV/SPIRVWriter.cpp
AGindinson: Sure. An even better approach could be to have the prefetch/cache-size's `Optional`s hold a `StringRef`, thus avoiding the string -> integer -> string conversions. What do you think here?

test/transcoding/IntelFPGAMemoryAccesses.ll
MrSidims: Oh, I meant to be mirrored with the appropriate comment.

lib/SPIRV/SPIRVWriter.cpp
MrSidims: I like it

AGindinson: I've decided to go with your initial suggestion so that the intermediate "integer step" can be used as a value verification at caller site.

lib/SPIRV/SPIRVReader.cpp
```diff
@@ -3100,15 +3120,38 @@ void generateIntelFPGAAnnotationForStructMember(
...
+    LSUParamsBitmask |= 0x2;
+    ParamsOut << "{cache-size:" << Result << "}";
+  }
+  if (E->hasMemberDecorate(DecorationDontStaticallyCoalesceINTEL, 0,
+                           MemberNumber))
+    LSUParamsBitmask |= 0x4;
+  if (E->hasMemberDecorate(DecorationPrefetchINTEL, 0, MemberNumber, &Result)) {
+    LSUParamsBitmask |= 0x8;
+    // TODO: Enable prefetch size backwards translation
+    // once it is supported
```
mlychkov, line 3140:
```suggestion
    // TODO: Enable prefetch size backwards translation once it is supported
```

lib/SPIRV/SPIRVReader.cpp
```diff
@@ -3100,15 +3120,38 @@ void generateIntelFPGAAnnotationForStructMember(
     Out << E->getMemberDecorationStringLiteral(DecorationUserSemantic,
                                                MemberNumber)
                .front();
+
+  unsigned LSUParamsBitmask = 0;
+  llvm::SmallString<32> AdditionalParamsStr;
+  llvm::raw_svector_ostream ParamsOut(AdditionalParamsStr);
+  if (E->hasMemberDecorate(DecorationBurstCoalesceINTEL, 0, MemberNumber))
+    LSUParamsBitmask |= 0x1;
```
mlychkov, line 3128: Sorry for the late idea: may we add enum for these bitmask values?

AGindinson: It's actually a valuable idea, and I've also been thinking of doing this at earlier stages. I'll see if I can create some `enum` that can be used both by the Writer and the Reader.

lib/SPIRV/SPIRVWriter.cpp
```diff
@@ -1704,8 +1763,10 @@ std::vector<SPIRVWord> getBankBitsFromString(StringRef S) {
 void addIntelFPGADecorations(
     SPIRVEntry *E,
     std::vector<std::pair<Decoration, std::string>> &Decorations) {
-  if (!E->getModule()->isAllowedToUseExtension(
-          ExtensionID::SPV_INTEL_fpga_memory_attributes))
+  SPIRVModule *M = E->getModule();
+  if (!M->isAllowedToUseExtension(
```
AlexeySachkov, line 1767: Note: I guess that in our stack both extensions most likely will be enabled/disabled together, but strictly speaking, this `if` is not fully correct: it will allow to use functionality from `SPV_INTEL_fpga_memory_attributes` even if it is disabled, but `SPV_INTEL_fpga_memory_accesses` enabled

test/transcoding/IntelFPGAMemoryAccesses.ll
AlexeySachkov: As another suggestion, you can declare an `enum` or use `#define` to avoid using magic numbers

lib/SPIRV/SPIRVWriter.cpp
AGindinson: I see. I don't know if there's a quick way to separate these conditions, but at the very least, I should add a `TODO` for addressing this.

mlychkov: I suppose that solution is to use || instead of &&.

mlychkov: Oh, no, this check is for all decorations. Forget it.

AGindinson: I've entirely removed the check from the function, as the same check is performed at caller site. The corresponding `TODO` was added to reflect the issue.

test/transcoding/IntelFPGAMemoryAccesses.ll
AGindinson: Gone with the `#define` variant, thanks for the suggestion!

lib/SPIRV/OCLUtil.h
```diff
@@ -254,6 +254,13 @@ const OCLMemOrderKind OCLLegacyAtomicMemOrder = OCLMO_seq_cst;
 /// OCL 1.x atomic memory scope when translated to 2.0 atomics.
 const OCLScopeKind OCLLegacyAtomicMemScope = OCLMS_device;
 
+enum IntelFPGAMemoryVal {
```
mohammadfawaz, line 257:
```suggestion
enum IntelFPGAMemoryAccessesVal {
```

mohammadfawaz:
That's Artem. LGTM.

test/transcoding/IntelFPGAMemoryAccesses.ll
AGindinson: @MrSidims, could you please take a look at the updated patch and make a GH suggestion if your comment persists? I have to confess I couldn't figure out an appropriate resolution here.

PR #614, opened 2020-07-07
Enable translation of ac_fixed functions

vmaksimo:
This patch introduces SPV_INTEL_arbitrary_precision_fixed_point extension,
which adds operations for arbitrary precision fixed point numbers.

The ac_fixed datatype is represented as a pseudo type using OpTypeInt. Its
parameters are:
* W - total width of the datatype. It's encoded in the width of the OpTypeInt.
* I - determines the position of the decimal point.
* S - determines if this is a signed or an unsigned number.

Full specification (https://github.com/intel/llvm/pull/1934):
https://github.com/intel/llvm/blob/2f6e965e686354fbb25f9c177a667a646de302eb/sycl/doc/extensions/SPIRV/SPV_INTEL_arbitrary_precision_fixed_point.asciidoc


lib/SPIRV/SPIRVReader.cpp
```diff
@@ -2399,6 +2412,55 @@ Value *SPIRVToLLVM::transValueWithoutDecoration(SPIRVValue *BV, Function *F,
...
+                                   Int1Ty,
+                                   Int32Ty,
+                                   Int32Ty,
+                                   Int32Ty,
+                                   Int32Ty};
+
+  FunctionType *FT = FunctionType::get(RetTy, ArgTys, false);
+
+  Op OpCode = Inst->getOpCode();
+  Function *Func = Function::Create(FT, GlobalValue::ExternalLinkage,
```
AlexeySachkov, line 2442: Do you have a test where the same fixed point funciton/opcode is encountered twice? I would expect that `Funciton::Create` will declare the same buit-in each time it is called and it will add `.1, `.2`, etc. suffixes to declaration name. You probably interested in `llvm::Module::getOrInsertFunction`

vmaksimo: Oh, that's a good point, thanks! I will fix that and add the appropriate test case.

vmaksimo: This `getOrInsertFunction` is not perfect, though. If the same function is called twice - it works fine, but if the same function has different signatures and return types - a huge bitcast is being created:
`
call i13 bitcast (i5 (i13, i1, i32, i32, i32, i32)* @intel_arbitrary_fixed_sqrt to i13 (i5, i1, i32, i32, i32, i32)*)(i5 %5, i1 false, i32 2, i32 2, i32 0, i32 0)
`
Not sure what is better - bitcast or `.N` at the end of the function name. What do you think of this?

AlexeySachkov: It depends on what can be handled by the consumer of that LLVM IR which is translated from the SPIR-V.

I guess that both variants are not really acceptable. Usually, such overloads are distinguished by mangling and `.1`, `.2` is some kind of it, also it is common that LLVM IR representation after the translation is similar to/the same as LLVM IR  representation before translation - probably this is a way you need to go?

lib/SPIRV/SPIRVReader.cpp
```diff
@@ -2424,26 +2424,31 @@ CallInst *SPIRVToLLVM::transFixedPointInst(SPIRVInstruction *BI,
...
-                                   Int32Ty,
-                                   Int32Ty,
-                                   Int32Ty};
-
+  SmallVector<Type *, 7> ArgTys = {InTy,    Int1Ty,  Int32Ty,
+                                   Int32Ty, Int32Ty, Int32Ty};
   FunctionType *FT = FunctionType::get(RetTy, ArgTys, false);
 
+  std::stringstream Suffix;
+  Suffix << ".i" << RetTy->getIntegerBitWidth() << ".i"
```
mlychkov, line 2439: Could you please add comment why we create function with special name?

lib/SPIRV/SPIRVReader.cpp
```diff
@@ -2424,26 +2424,31 @@ CallInst *SPIRVToLLVM::transFixedPointInst(SPIRVInstruction *BI,
...
-  Function *Func = Function::Create(FT, GlobalValue::ExternalLinkage,
-                                    SPIRVFixedPointIntelMap::rmap(OpCode), M);
-  Func->setCallingConv(CallingConv::SPIR_FUNC);
-  if (isFuncNoUnwind())
-    Func->addFnAttr(Attribute::NoUnwind);
+  std::string FuncName = SPIRVFixedPointIntelMap::rmap(OpCode) + Suffix.str();
+
+  FunctionCallee FCallee = M->getOrInsertFunction(FuncName, FT);
+
+  if (Function *Fn = dyn_cast<Function>(FCallee.getCallee())) {
```
mlychkov, line 2451: Should we check if this is Function*? What if this is not so?

test/capability-arbitrary-precision-fixed-point-numbers.ll
```diff
@@ -0,0 +1,583 @@
...
+!opencl.used.extensions = !{!3}
+!opencl.used.optional.core.features = !{!3}
+!opencl.compiler.options = !{!3}
+!llvm.ident = !{!4}
+
+!0 = !{i32 1, !"wchar_size", i32 4}
+!1 = !{i32 1, i32 2}
+!2 = !{i32 4, i32 100000}
+!3 = !{}
+!4 = !{!"Intel(R) oneAPI DPC++ Compiler 2021.1 (YYYY.x.0.MMDD)"}
```
mlychkov, line 547: I suppose that this test should be recompiled by opensource version of clang.

lib/SPIRV/SPIRVReader.cpp
AlexeySotkin: We check that it is a `Function*` with `dyn_cast<Function>`. Probably we might want to check that `FCallee.getCallee()` doesn't return `nullptr`. We should use `dyn_cast_or_null` in this case.

lib/SPIRV/SPIRVReader.cpp
```diff
@@ -2399,6 +2412,64 @@ Value *SPIRVToLLVM::transValueWithoutDecoration(SPIRVValue *BV, Function *F,
   }
 }
 
+CallInst *SPIRVToLLVM::transFixedPointInst(SPIRVInstruction *BI,
+                                           BasicBlock *BB) {
+  // LLVM fixed point functions return value:
+  // iN
```
AlexeySotkin, line 2418: What is `N`?

lib/SPIRV/SPIRVReader.cpp
mlychkov: Well, actually we expect that it will be always Function* because we insert function call on the line above.
I would rather use just `cast<Function>` call without an if  check.

mlychkov:
Updated review has been created in PR 653.

AlexeySotkin:
> Updated review has been created in PR 653.

Should we close this one and follow up in #653 ?

mlychkov:
> > Updated review has been created in PR 653.
> 
> Should we close this one and follow up in #653 ?

Yes, it would be good to close this one as outdated.

PR #618, opened 2020-07-09
Remove extra intrinsics from "isKnownIntrinsic" function

NikitaRudenkoIntel:
They were added by mistake when backported from TOT.

### DIFF
--- a/lib/SPIRV/SPIRVWriter.cpp
+++ b/lib/SPIRV/SPIRVWriter.cpp
@@ -1607,14 +1607,10 @@ void addIntelFPGADecorationsForStructMember(
 bool LLVMToSPIRV::isKnownIntrinsic(Intrinsic::ID Id) {
   // Known intrinsics usually do not need translation of their declaration
   switch (Id) {
-  case Intrinsic::assume:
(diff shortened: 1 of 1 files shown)


lib/SPIRV/SPIRVWriter.cpp
```diff
@@ -1607,14 +1607,10 @@ void addIntelFPGADecorationsForStructMember(
 bool LLVMToSPIRV::isKnownIntrinsic(Intrinsic::ID Id) {
   // Known intrinsics usually do not need translation of their declaration
   switch (Id) {
-  case Intrinsic::assume:
```
AlexeySachkov, line 1610: Note: this one actually can be translated into SPIR-V if corresponding extension is enabled

NikitaRudenkoIntel: On llvm_release_100 branch?

AlexeySachkov: Oops, I've missed the fact that this is not for `master`. Then no, it can't be translated on that branch anyway

PR #620, opened 2020-07-09
Move SPIRVAllowUnknownIntrinsics option to TranslatorOpts class

NikitaRudenkoIntel:
* Move SPIRVAllowUnknownIntrinsics option to TranslatorOpts class
* Extend LLVMSPIRVOpts.h interface allowing pass options to spirv passes

### DIFF
--- a/lib/SPIRV/SPIRVWriter.cpp
+++ b/lib/SPIRV/SPIRVWriter.cpp
@@ -92,11 +92,6 @@ using namespace OCLUtil;
 
 namespace SPIRV {
 
-cl::opt<bool> SPIRVAllowUnknownIntrinsics(
-    "spirv-allow-unknown-intrinsics", cl::init(false),
-    cl::desc("Unknown LLVM intrinsics will be translated as external function "
-             "calls in SPIR-V"));
-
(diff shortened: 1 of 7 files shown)


lib/SPIRV/SPIRVWriter.cpp
```diff
@@ -123,9 +118,9 @@ static SPIRVMemoryModelKind getMemoryModel(Module &M) {
   return SPIRVMemoryModelKind::MemoryModelMax;
 }
 
-LLVMToSPIRV::LLVMToSPIRV(SPIRVModule *SMod)
+LLVMToSPIRV::LLVMToSPIRV(SPIRVModule *SMod, bool SPIRVAllowUnknownIntrinsics)
```
AlexeySachkov, line 121: I wonder why this change is needed? Please note that `TranslatorOpts` instance is available within `SPIRVModule` and you can query desired property from the module instead.

For example:

https://github.com/KhronosGroup/SPIRV-LLVM-Translator/blob/2846475f19c56ae034d6e418fac48b117ab68bc2/lib/SPIRV/libSPIRV/SPIRVModule.h#L467-L469

PR #629, opened 2020-07-13
Fix type mapping for pointers with Device/HostOnlyINTEL storage classes

MrSidims:
a7b763b265d introduces a bug, when it could be possible to map a
single pointer type in LLVM IR to two different pointer types in
SPIR-V (when SPV_INTEL_usm_storage_classes extension is not allowed).
This patch fixes this bug.


MrSidims:
Tagging @AlexeySotkin, @mlychkov, @AGindinson and @vmaksimo

PR #630, opened 2020-07-14
Add DW_OP_and operation to SPIRV.debug.h

NikitaRudenkoIntel:
This operation cannot be merged yet because "Debug Information Extended Instruction Set" is lacking this operation:
https://www.khronos.org/registry/spir-v/specs/unified1/DebugInfo.html#_debug_operations_a_id_operation_a

It would be great if the specification could be amended with this operation

### DIFF
--- a/lib/SPIRV/libSPIRV/SPIRV.debug.h
+++ b/lib/SPIRV/libSPIRV/SPIRV.debug.h
@@ -105,7 +105,8 @@ enum ExpressionOpCode {
   Xderef     = 6,
   StackValue = 7,
   Constu     = 8,
-  Fragment   = 9
+  Fragment   = 9,
+  And        = 10,
 };
 
 enum ImportedEntityTag {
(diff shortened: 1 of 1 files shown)


AlexeySotkin:
The latest version of the debug info spec is https://www.khronos.org/registry/spir-v/specs/unified1/OpenCL.DebugInfo.100.html#_debug_operations_a_id_operation_a

I'm afraid we can't just add an operation to the spec without breaking backward compatibility. Typically in this case we should create a new version of the spec (`OpenCL.DebugInfo.101`?) or create an extension to the spec. @dneto0, @johnkslang, @bashbaug what do you think?

johnkslang:
Right, we've been saying that adding new functionality will require a new version.  Otherwise, existing tools correctly accepting the existing version will fail to work correctly rather than say they don't support the extension.

AlexeySotkin:
As a short term workaround, https://github.com/KhronosGroup/SPIRV-LLVM-Translator/pull/679 should be able to replace this patch.
In the long term, we should have a new spec or maybe some extension to the existing one.

AlexeySachkov:
I suggest to close this one in favor of #679 (see documentation for it in #749). @NikitaRudenkoIntel, do you have any objections?

Issue #210, opened 2019-06-04
Manage SPIR-V headers as an external dependency


kpet:
An old copy of (some of) the SPIR-V headers is currently committed under `lib/SPIRV/libSPIRV/{spirv.hpp,OpenCL.std.h}`.

Should the headers be managed as an external dependency, most likely a submodule, instead?

This would make it much easier to track header changes.

neildhickey:
Agreed, there are some discrepancies between the upstream khronos headers and the ones committed, so that would need cleaning up too

svenvh:
@AlexeySotkin @AlexeySachkov, llvm-spirv's copy of `spirv.hpp` has diverged from the official Khronos version (https://github.com/KhronosGroup/SPIRV-Headers/blob/master/include/spirv/unified1/spirv.hpp), in particular due to Intel-specific extensions it seems (e.g. `OpFunctionPointerINTEL`, `CapabilityFPGARegINTEL`, `DecorationMergeINTEL` etc).

It would be great if we could reduce this divergence again; do you have any plans to do so?

AlexeySotkin:
I agree that the translator should use the official spir-v headers from the Khronos repository.
We do plan to upstream the Intel-specific extensions the spir-v headers repo. But I can't provide ETA for this activity. Meanwhile, I think we can move Intel-specific extensions into a separate header in this repo to unblock the switch.

Technically, I think we could manage the headers as a cmake external project - https://cmake.org/cmake/help/latest/module/ExternalProject.html

AlexeySachkov:
@svenvh, we have actually started the activity of upstreaming our extensions into KhronosGroup/SPIRV-Headers, two PRs have been merged so far: KhronosGroup/SPIRV-Headers#150 and KhronosGroup/SPIRV-Headers#152

I honestly don't know what amount of work is left there, but @MrSidims should be able to answer that question

MrSidims:
There are several other Intel extensions that mere committed since April 1st (on top of my previous upstreaming work, there is a discussion raised in https://github.com/KhronosGroup/SPIRV-Headers/issues/151, but it no longer a blocker).

To fully replace local spirv.hpp we need 4 things:
1. Upstream rest extensions;
2. Get rid of InternalLinkage type in the local copy, which is not a part of SPIR-V spec (I have started working on this, but didn't have much time to finish, the idea is to move it into spirv-internal.hpp);
3. Get rid of OpForward instruction (same);
[...]

PR #635, opened 2020-07-15
Update gen_spirv.bash

svenvh:
Update `gen_spirv.bash` to align it with the current code base.  This PR does not synchronize the generated files with the script output yet; I plan to do that in a followup pull request including a cleanup of generated functions that are not used. By keeping that separate, the changes should be easier to review.

This PR includes:
- Special-case builtins in gen_spirv.bash
- Do not generate NameMap or isValid for SPIR-V ops
- Align gen_spirv.bash with current formatting
- Derive output file name from type argument
- Remove td file generation from gen_spirv.bash


AlexeySotkin:
@svenvh, could you please provide example of the script output?

svenvh:
> @svenvh, could you please provide example of the script output?

The full output is the entire `lib/SPIRV/libSPIRV/SPIRVNameMapEnum.h` file when invoked as `bash tools/spirv-tool/gen_spirv.bash lib/SPIRV/libSPIRV/spirv.hpp NameMap`, but then updated for recent additions to `spirv.hpp`. It's too big to paste it here, but to give you an idea, after invoking it, it will rewrite `lib/SPIRV/libSPIRV/SPIRVNameMapEnum.h` such that the diff looks like:

```
diff --git a/lib/SPIRV/libSPIRV/SPIRVNameMapEnum.h b/lib/SPIRV/libSPIRV/SPIRVNameMapEnum.h
index a1a1d19..b6f2d60 100644
--- a/lib/SPIRV/libSPIRV/SPIRVNameMapEnum.h
+++ b/lib/SPIRV/libSPIRV/SPIRVNameMapEnum.h
@@ -59,6 +59,8 @@ template <> inline void SPIRVMap<SourceLanguage, std::string>::init() {
   add(SourceLanguageGLSL, "GLSL");
   add(SourceLanguageOpenCL_C, "OpenCL_C");
   add(SourceLanguageOpenCL_CPP, "OpenCL_CPP");
+  add(SourceLanguageHLSL, "HLSL");
+  add(SourceLanguageMax, "Max");
 }
 SPIRV_DEF_NAMEMAP(SourceLanguage, SPIRVSourceLanguageNameMap)
 
@@ -70,6 +72,15 @@ template <> inline void SPIRVMap<ExecutionModel, std::string>::init() {
   add(ExecutionModelFragment, "Fragment");
   add(ExecutionModelGLCompute, "GLCompute");
   add(ExecutionModelKernel, "Kernel");
+  add(ExecutionModelTaskNV, "TaskNV");
+  add(ExecutionModelMeshNV, "MeshNV");

... etc ...
```

Similarly, here is the start of a diff after invoking the script to generate the isValid functions: `bash tools/spirv-tool/gen_spirv.bash lib/SPIRV/libSPIRV/spirv.hpp isValid`:
```
diff --git a/lib/SPIRV/libSPIRV/SPIRVIsValidEnum.h b/lib/SPIRV/libSPIRV/SPIRVIsValidEnum.h
index 76a7e81..c9d822e 100644
--- a/lib/SPIRV/libSPIRV/SPIRVIsValidEnum.h
+++ b/lib/SPIRV/libSPIRV/SPIRVIsValidEnum.h
@@ -60,6 +60,8 @@ inline bool isValid(spv::SourceLanguage V) {
   case SourceLanguageGLSL:
   case SourceLanguageOpenCL_C:
   case SourceLanguageOpenCL_CPP:
+  case SourceLanguageHLSL:
+  case SourceLanguageMax:
     return true;
   default:
     return false;

... etc ...
```

AlexeySotkin:
What is supposed usage model for this script?

AlexeySachkov:
> What is supposed usage model for this script? I know that the script is here for quite a long time, but I've never used it :-)

If I understand correctly, the correct way of updating out copy of `spirv.hpp` is launching this script instead of manual changes of the file.

@svenvh, am I right that this PR is a step towards resolving #210? I guess it should be launched regularly (or even during build time) to generate `spirv.hpp` from `.json` files

svenvh:
> What is supposed usage model for this script? I know that the script is here for quite a long time, but I've never used it :-)

This script is supposed to be used to generate `lib/SPIRV/libSPIRV/SPIRVIsValidEnum.h` and `lib/SPIRV/libSPIRV/SPIRVNameMapEnum.h`, as can be seen from a comment in those files:

```
// WARNING:
//
// This file has been generated using `tools/spirv-tool/gen_spirv.bash` and
// should not be modified manually. If the file needs to be updated, edit the
// script and any other source file instead, before re-generating this file.
```

So the intended usage is when spirv.hpp is updated, this script is run to generate the aforementioned SPIRV*Enum.h files.  No one has been doing that for a while (I didn't know about this script either), so the header files and the script are out of sync at the moment.  This PR brings them closer together, but doesn't update the header files yet.  The reason for not bringing them in sync is this PR is that I want to cleanup the generated code to reduce llvm-spirv build times a bit, and syncing them now would only add more unused code (so I want to do the sync and the cleanup in the next PR, which should also make it easier to review).

> the correct way of updating out copy of spirv.hpp is launching this script instead of manual changes of the file.

Almost, this script should be run *after* updating spirv.hpp.

> @svenvh, am I right that this PR is a step towards resolving #210? I guess that the script should be launched regularly (or even during build time) to generate spirv.hpp from .json files

It won't really solve #210, but longer term we should perhaps make it part of the build system. Such that llvm-spirv's header files are regenerated whenever spirv.hpp is modified.

svenvh:
(had to rebase due to a merge conflict)

PR #636, opened 2020-07-16
Update LLVM version from 11 to 12

svenvh:


svenvh:
apt.llvm.org does not have packages for LLVM 12 yet.

AlexeySachkov:
@svenvh, it seems that packages have been updated for LLVM 12 (I've restarted one of Travis CI jobs: https://travis-ci.org/github/KhronosGroup/SPIRV-LLVM-Translator/jobs/710293991)

However, build fails due to API change (should be fixed by #661)

I suggest that we merge this PR to switch to new LLVM and then apply build fix from #661, what do you think?

svenvh:
Sorry for the delayed response, thanks for merging this!

PR #646, opened 2020-07-19
Reset !spirv.ExecutionNode in PreprocessMetadata::visit()

bbrezillon:
Not sure if that's a valid use case, but there's currently nothing
preventing us from calling regularizeLlvmForSpirv() before writeSpirv(),
which already takes care of the LLVM regularization internally. Doing
that seems to cause duplicate entries in the !spirv.ExecutionMode node.

The comment in PreprocessMetadata::visit() seems to imply that
!spirv.ExecutionNode should be empty when we start, but addNamedMD()
actually returns the existing node if there's one. Let's delete any
pre-existing version of this metadata node so addNamedMD() always
returns a fresh NamedMDNode.

Signed-off-by: Boris Brezillon <boris.brezillon@collabora.com>


bbrezillon:
Hm, looks like `test/ExecutionMode.ll` expects the `!spirv.ExecutionMode` node to be preserved.

AlexeySotkin:
LGTM.

AlexeySotkin:
I think it makes sense to write a LIT test  checking that the metadata is erased. The test can run `llvm-spirv -s` to produce regularized LLVM IR.
@bbrezillon, if you will update the PR, please rebase it in order to run CI tests with new CMake verison.

PR #647, opened 2020-07-20
CI update: fix compilation issues

Hopetech:
Since [D78646](https://reviews.llvm.org/D78646), LLVM require CMake 3.13.4 to build.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -73,7 +73,6 @@ matrix:
 
     - env: BUILD_EXTERNAL=1 CHECK_TIDY=1
   allow_failures:
-    - compiler: clang
     - os: osx
   fast_finish: true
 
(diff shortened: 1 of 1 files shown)


Hopetech:
@AlexeySotkin I think this series would be helpful.
We should push patches only if the CI have passed.
FYI @svenvh @AnastasiaStulova

.travis.yml
```diff
@@ -60,17 +63,16 @@ matrix:
   include:
     - os: osx
       env: BUILD_TYPE=Release BUILD_EXTERNAL=0 MAKE_TARGETS="llvm-spirv" MAKE_TEST_TARGET="check-llvm-spirv"
-      osx_image: xcode9.3
+      osx_image: xcode12
```
svenvh, line 66: Actually I noticed the xcode jobs are timing out after 50m in your travis run, whereas they used to take about 30m.

Hopetech: Right, quite strange.
I will compare the 2 travis run.

AlexeySotkin:
LGTM, Thanks!

PR #649, opened 2020-07-20
Fix translation of user functions into SPIR-V instructions

MochalovaAn:
User may get their function a name the same as SPIR-V instruction.
In that case thier function call will be translated to SPIR-V instruction call.
The only functions that should be translated to SPIR-V instructions are SPIR-V built-ins
and they should have __spirv prefix.
Add checking for __spirv prefix in demangler to filter SPIR-V built-ins from other function calls

### DIFF
--- a/test/transcoding/BitReversePref.ll
+++ b/test/transcoding/BitReversePref.ll
@@ -1,0 +1,64 @@
+; RUN: llvm-as < %s -o %t.bc
+; RUN: llvm-spirv %t.bc -spirv-text -o - | FileCheck %s
+
+;CHECK:  Decorate [[#FUNC_NAME:]] LinkageAttributes "_Z10BitReversei"
+;CHECK-NOT: BitReverse
+;CHECK: FunctionCall [[#]] [[#]] [[#FUNC_NAME]]
+
+target datalayout = "e-i64:64-v16:16-v24:32-v32:32-v48:64-v96:128-v192:256-v256:256-v512:512-v1024:1024-n8:16:32:64"
+target triple = "spir64-unknown-linux-sycldevice"
+
+%"class._ZTSZ4mainE3$_0.anon" = type { i8 }
+
+$_Z10BitReversei = comdat any
+
+; Function Attrs: norecurse
+define dso_local spir_kernel void @_ZTSZ4mainE15kernel_function() #0 !kernel_arg_addr_space !4 !kernel_arg_access_qual !4 !kernel_arg_type !4 !kernel_arg_base_type !4 !kernel_arg_type_qual !4 {
+entry:
+  %call = call spir_func i32 @_Z10BitReversei(i32 1)
+  ret void
+}
+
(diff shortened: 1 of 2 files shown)


test/transcoding/BitReversePref.ll
```diff
@@ -0,0 +1,1584 @@
+; RUN: llvm-as < %s -o %t.bc
```
mlychkov, line 1: Is it possible to reduce this test example to the minimum required amount of lines to reproduce the issue? It is very hard to understand what and where is checked actually.

AlexeySotkin:
@MochalovaAn you need to accept the CLA in order to contribute to this repository.
Please rebase your branch to re-run the CI tests with new CMake version. Also I'd appreciate if you squash your commits.

test/transcoding/BitReversePref.ll
```diff
@@ -0,0 +1,64 @@
...
+
+%"class._ZTSZ4mainE3$_0.anon" = type { i8 }
+
+$_Z10BitReversei = comdat any
+
+; Function Attrs: norecurse
+define dso_local spir_kernel void @_ZTSZ4mainE15kernel_function() #0 !kernel_arg_addr_space !4 !kernel_arg_access_qual !4 !kernel_arg_type !4 !kernel_arg_base_type !4 !kernel_arg_type_qual !4 {
+entry:
+  %call = call spir_func i32 @_Z10BitReversei(i32 1)
+ ret void
```
mlychkov, line 19:
```suggestion
  ret void
```

mlychkov:
Sorry for late comment, you may ignore it if you wish. But I've thought that commit message header "Bug fix in translator" is acceptable for a lot of commits, hence it doesn't identify clearly the purpose of current fix.
Current description of commit could be moved to the header, whilst it is more helpful to describe a problem to be fixed by this commit.
I.e. the whole commit message could be:
`Fix translation of user functions into SPIR-V instructions

User may get their function a name the same as SPIR-V instruction.
In that case thier function call will be translated to SPIR-V instruction call.
The only functions that should be translated to SPIR-V instructions are SPIR-V built-ins
and they should have __spirv_ prefix.
Add checking for __spirv_ prefix in demangler to filter SPIR-V built-ins from other function calls.`

If something is not clear or inaccurate then other reviewers feel free to correct the description.

test/transcoding/BitReversePref.ll
mlychkov: Thank you. Perfect test reduction.

svenvh:
> Sorry for late comment, you may ignore it if you wish. But I've thought that commit message header "Bug fix in translator" is acceptable for a lot of commits, hence it doesn't identify clearly the purpose of current fix.
> It is more helpful to describe a problem to be fixed by this commit.

Thanks for bringing this up @mlychkov, I agree it would be good to have a more descriptive commit message.

@MochalovaAn it would be great if you could squash your commits into one and use @mlychkov's suggestion.

PR #650, opened 2020-07-21
Update DebugInfo tests after LLVM changes

svenvh:
Update a test for LLVM commit df952cb914e ("[llvm-readobj] Print error
when executed with no input files", 2020-07-20).

Update a DebugInfo test according to LLVM commit ce6de3747bc ("[DebugInfo]
Drop location ranges for variables which exist entirely outside the
variable's scope", 2020-07-22).


test/DebugInfo/X86/xray-split-dwarf-interaction.ll
```diff
@@ -11,7 +11,7 @@
 ; RUN:     -function-sections -data-sections \
 ; RUN:     -relocation-model=pic -filetype=obj \
 ; RUN:     -generate-type-units -o - %s | \
```
AlexeySotkin, line 13: There is a bug. There should be `%t.ll` instead of `%s`. But the test fails if I fix it. I suggest to remove this test, because SPIR-V format (debug info) doesn't preserve XRay instrumentation.

svenvh: Good catch! I've removed the test.

AlexeySotkin:
This change is unique for the development(master) branch. I think we should fork `llvm_release_110` before we merge this PR, if we want to avoid reverting it on `llvm_release_110`

svenvh:
> This change is unique for the development(master) branch. I think we should fork `llvm_release_110` before we merge this PR, if we want to avoid reverting it on `llvm_release_110`

Agreed.  Both this PR and #636 should only be merged once we have forked the 11.0 release branch.

PR #652, opened 2020-07-24
Improve SPIRVToLLVM::transOCLBuiltinFromVariable

AlexeySotkin:
The method did not handle correctly the case when addrspacecast instruction had
more than one usage.

### DIFF
--- a/lib/SPIRV/SPIRVReader.cpp
+++ b/lib/SPIRV/SPIRVReader.cpp
@@ -243,11 +243,30 @@ Value *SPIRVToLLVM::mapFunction(SPIRVFunction *BF, Function *F) {
 
 // Variable like GlobalInvolcationId[x] -> get_global_id(x).
 // Variable like WorkDim -> get_work_dim().
+// Replace the following pattern:
+// %a = addrspacecast i32 addrspace(1)* @__spirv_BuiltInSubgroupMaxSize to
+// i32 addrspace(4)*
+// %b = load i32, i32 addrspace(4)* %a, align 4
+// %c = load i32, i32 addrspace(4)* %a, align 4
+// With:
+// %b = call spir_func i32 @_Z22get_max_sub_group_sizev()
+// %c = call spir_func i32 @_Z22get_max_sub_group_sizev()
+
+// And replace the following pattern:
+// %a = addrspacecast <3 x i64> addrspace(1)* @__spirv_BuiltInWorkgroupId to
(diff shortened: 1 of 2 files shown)


lib/SPIRV/SPIRVReader.cpp
```diff
@@ -265,48 +283,70 @@ bool SPIRVToLLVM::transOCLBuiltinFromVariable(GlobalVariable *GV,
...
+  for (auto *UI : GV->users()) {
+    // There might or might not be an addrspacecast instruction.
+    if (auto *ASCast = dyn_cast<AddrSpaceCastInst>(UI)) {
+      Casts.push_back(ASCast);
+      for (auto *CastUser : ASCast->users() ) {
+        if (auto *LD = dyn_cast<LoadInst>(CastUser)) {
+          FindAndReplace(LD);
+        }
+      }
+    } else if (auto *LD = cast<LoadInst>(UI)) {
```
svenvh, line 332:
```suggestion
    } else if (auto *LD = dyn_cast<LoadInst>(UI)) {
```

AlexeySotkin: Thanks! Fixed.

lib/SPIRV/SPIRVReader.cpp
```diff
@@ -265,48 +285,70 @@ bool SPIRVToLLVM::transOCLBuiltinFromVariable(GlobalVariable *GV,
...
+
+  // If HasIndexArg is true, we are going over users of the Load instruction,
+  // which are expected to be ExtractElement instructions. The result of the
+  // ExtractElement instructions is considered as a usage of the GV and should
+  // be replaced with Func.
+  // If HasIndexArg is false, the result of the Load instruction is the value
+  // which should be replaced with the Func.
+  auto FindAndReplace = [&](LoadInst *LD) {
+    Loads.push_back(LD);
+    if (HasIndexArg) {
```
AlexeySachkov, line 311: Minor: suggest to use early exit here:

```c++
if (!HasIndexArg) {
  Replace({}, LD);
  return;
}
for (/* ... */) {
  /* ... */
}
```

AlexeySotkin: I'd prefer to leave it as is if you don't mind.

AlexeySachkov: No objections to leave as-is, this is a minor comment

PR #653, opened 2020-07-27
[SYCL][FPGA] Enable translation of ac_fixed functions

mlychkov:
This patch introduces SPV_INTEL_arbitrary_precision_fixed_point extension,
which adds operations for arbitrary precision fixed point numbers.

The ac_fixed datatype is represented as a pseudo type using OpTypeInt. Its
parameters are:
* W - total width of the datatype. It's encoded in the width of the OpTypeInt.
* I - determines the position of the decimal point.
* S - determines if this is a signed or an unsigned number.

Full specification (https://github.com/intel/llvm/pull/1934):
https://github.com/intel/llvm/blob/2f6e965e686354fbb25f9c177a667a646de302eb/sycl/doc/extensions/SPIRV/SPV_INTEL_arbitrary_precision_fixed_point.asciidoc


mlychkov:
CThis review is created as a continue of #614

mlychkov:
@AlexeySotkin @AlexeySachkov @vmaksimo @MrSidims Please review.

vmaksimo:
LGTM

test/capability-arbitrary-precision-fixed-point-numbers.ll
```diff
@@ -0,0 +1,581 @@
...
+;     exp<44, 34, false, 20, 20>();
+;   });
+;   return 0;
+; }
+
+; RUN: llvm-as %s -o %t.bc
+; RUN: llvm-spirv %t.bc --spirv-ext=+SPV_INTEL_arbitrary_precision_integers,+SPV_INTEL_arbitrary_precision_fixed_point -o %t.spv
+; RUN: llvm-spirv %t.spv -to-text -o - | FileCheck %s --check-prefix=CHECK-SPIRV
+
+; RUN: llvm-spirv %t.bc --spirv-ext=+SPV_INTEL_arbitrary_precision_integers -spirv-text -o - | FileCheck %s --check-prefix=CHECK-SPIRV-NEGATIVE
```
AlexeySotkin, line 98: Will we produce a valid SPIR-V in this case? If no, I think we should report an error.

AlexeySotkin:
Please resolve the conflicts.

test/capability-arbitrary-precision-fixed-point-numbers.ll
mlychkov: In case if we translate to SPIR-V without SPV_INTEL_arbitrary_precision_fixed_point then those built-ins are translated as user defined function calls. I suppose that it is technically valid but linkage will be failed.
I may add error generation, but the same mechanism should be added for hls_float built-ins too. Sinse that feature has already merged I would address this change in a separate PR for both features later.

mlychkov:
@AlexeySachkov Could you please review?

PR #655, opened 2020-07-28
[SYCL][FPGA] Lift the restriction on LSU intrinsic operands

AGindinson:
The initial implementation had required that the first
ptr.annotation operand be a result of a load intruction.
However, it may be the case that the LSU built-in gets applied
to a cast result/to a result of a function call, therefore
this patch drops the unfeasible requirement.


AGindinson:
@mohammadfawaz FYI

mohammadfawaz:
Thank you for quickly fixing this!

PR #656, opened 2020-07-29
Fix translation of LLVM intrinsics when their args are stored

AGindinson:
Replacement of the read-from operand with the result of its
intrinsic call user should only happen for `llvm.ptr.annotation`
calls.

### DIFF
--- a/test/transcoding/intrinsic_result_store.ll
+++ b/test/transcoding/intrinsic_result_store.ll
@@ -1,0 +1,45 @@
+; Regression test for intrinsic calls' translation edge cases
+; with subsequent store instructions.
+
+; RUN: llvm-as %s -o %t.bc
+; RUN: llvm-spirv %t.bc -o %t.spv
+; RUN: llvm-spirv -r %t.spv -o %t.rev.bc
+; RUN: llvm-dis < %t.rev.bc | FileCheck %s --check-prefix=CHECK-LLVM
+
+target datalayout = "e-i64:64-v16:16-v24:32-v32:32-v48:64-v96:128-v192:256-v256:256-v512:512-v1024:1024"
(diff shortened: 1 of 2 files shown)


test/transcoding/intrinsic_result_store.ll
```diff
@@ -0,0 +1,45 @@
...
+; RUN: llvm-as %s -o %t.bc
+; RUN: llvm-spirv %t.bc -o %t.spv
+; RUN: llvm-spirv -r %t.spv -o %t.rev.bc
+; RUN: llvm-dis < %t.rev.bc | FileCheck %s --check-prefix=CHECK-LLVM
+
+target datalayout = "e-i64:64-v16:16-v24:32-v32:32-v48:64-v96:128-v192:256-v256:256-v512:512-v1024:1024"
+target triple = "spir64-unknown-unknown"
+
+; Function Attrs: convergent noinline nounwind optnone
+define spir_kernel void @test_memset(i32 addrspace(1)* %data, i32 %input) #0 !kernel_arg_addr_space !1 !kernel_arg_access_qual !5 !kernel_arg_type !6 !kernel_arg_base_type !6 !kernel_arg_type_qual !7 {
```
mlychkov, line 13: Is %input neccessary for this test?

AGindinson: Since a `ptr.annotation` would need to be applied, I've decided that having it on a function parameter akin to `%input` would be best. I might as well have gone with some local, or even self-assigning `%data`, yet at the end of the day, it doesn't seem to make much difference.

mlychkov: I meant the second parameter of this kernel, is it necessary?

PR #657, opened 2020-07-31
Support of SPV_INTEL_variable_length_array

AlexeySotkin:
The extension spec is at https://github.com/intel/llvm/pull/2217


test/negative/llvm.fma.ll
```diff
@@ -4,7 +4,8 @@
 ; RUN: llvm-as %s -o %t.bc
 ; RUN: not llvm-spirv %t.bc 2>&1 | FileCheck %s
 
-; CHECK: InvalidFunctionCall: Unexpected llvm intrinsic: llvm.fma.f32
+; CHECK: InvalidFunctionCall: Unexpected llvm intrinsic:
```
MrSidims, line 7: I kind of like one-liners more. I probably see logic to move intrinsic name to the next line (to prevent double ':' nesting in a single line), but I think while it fits into 80 symbols limitation - there is no need to split the string.

AlexeySachkov: This was done not because of too long line, but because of changes in the translator output: it now emits `\n` between this phrase and actual intrinsic name: the next check it not just `CHECK`, but `CHECK-NEXT`

lib/SPIRV/SPIRVWriter.cpp
```diff
@@ -1261,11 +1261,26 @@ SPIRVValue *LLVMToSPIRV::transValueWithoutDecoration(Value *V,
             transValue(Sel->getFalseValue(), BB, true, FuncTransMode::Pointer),
             BB));
 
-  if (AllocaInst *Alc = dyn_cast<AllocaInst>(V))
+  if (AllocaInst *Alc = dyn_cast<AllocaInst>(V)) {
+    if (Alc->isArrayAllocation()) {
```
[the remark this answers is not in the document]
AlexeySotkin: Do you mean the case like the following?
```
%vla = alloca i32, i64 2
```
If yes, It looks rather like an artificial case. I think it is not a critical issue and can be fixed later.

AlexeySachkov (approved):

PR #658, opened 2020-07-31
[Backport] travis: Move to 18.04 (Bionic)

AlexeySotkin:
The Xenial binaries on apt.llvm.org have been outdated for a while,
whereas the Bionic binaries seem up to date now.  Switch Travis to
test on Bionic.

Drop the pyenv fix from df75cab ("Fix Travis CI build.", 2019-07-11)
as it no longer seems to be needed for 18.04.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,8 +3,8 @@ language: cpp
 os:
   - linux
 
-# Use Ubuntu 16.04 LTS (Xenial) as the Linux testing environment.
-dist: xenial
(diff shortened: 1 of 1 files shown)


AlexeySotkin:
> Both Xenial and Bionic are stuck at 15 July according to apt.llvm.org

Right, but my intention was to fix another CI issue, on `llvm_release_90`(not `master`) branch: https://travis-ci.org/github/KhronosGroup/SPIRV-LLVM-Translator/jobs/713616094#L367.

PR #659, opened 2020-07-31
Add OpenCL.DebugInfo.100 as alias to SPIRV.debug according to specification

NikitaRudenkoIntel:
https://www.khronos.org/registry/spir-v/specs/unified1/OpenCL.DebugInfo.100.html


NikitaRudenkoIntel:
Yes, backwards compatibility is the reason why I made "new" set instead of renaming the old one...

AlexeySachkov:
> > Yes, backwards compatibility is the reason why I made "new" set instead of renaming the old one...
> 
> That might be worth adding to the commit message.

Done

PR #663, opened 2020-08-04
SPV_INTEL_function_pointers update

zuban32:
- Make OpFunctionPointerINTEL constant (corresponds to the spec change https://github.com/intel/llvm/pull/2254)
- Function pointers translation refactoring


lib/SPIRV/SPIRVWriter.cpp
```diff
@@ -728,17 +730,19 @@ SPIRVValue *LLVMToSPIRV::transConstant(Value *V) {
 }
 
 SPIRVValue *LLVMToSPIRV::transValue(Value *V, SPIRVBasicBlock *BB,
-                                    bool CreateForward) {
+                                    bool CreateForward,
+                                    FuncTransMode FuncTrans) {
   LLVMToSPIRVValueMap::iterator Loc = ValueMap.find(V);
-  if (Loc != ValueMap.end() && (!Loc->second->isForward() || CreateForward))
+  if (Loc != ValueMap.end() && (!Loc->second->isForward() || CreateForward) &&
+      !(FuncTrans == FuncTransMode::Pointer && isa<Function>(V)))
```
AlexeySachkov, line 737: Am I right that the purpose of this "if" is to avoid returning forward-declarations of functions when `OpConstFunctionPointerINTEL` is needed to be generated?

lib/SPIRV/SPIRVWriter.cpp
```diff
@@ -1053,8 +1046,18 @@ SPIRVValue *LLVMToSPIRV::transValueWithoutDecoration(Value *V,
     return BB;
   }
 
-  if (auto F = dyn_cast<Function>(V))
-    return transFunctionDecl(F);
+  if (auto *F = dyn_cast<Function>(V)) {
+    if (FuncTrans == FuncTransMode::Decl)
+      return transFunctionDecl(F);
+    else {
```
AlexeySachkov, line 1052: No need in `else` after return

lib/SPIRV/libSPIRV/SPIRVFunction.h
```diff
@@ -171,6 +171,36 @@ class SPIRVFunction : public SPIRVValue, public SPIRVComponentExecutionModes {
 
 typedef SPIRVEntryOpCodeOnly<OpFunctionEnd> SPIRVFunctionEnd;
 
+
+class SPIRVConstFunctionPointerINTEL : public SPIRVValue {
```
AlexeySachkov, line 175: I wonder, why this is not a `SPIRVInstruction` anymore?

test/transcoding/SPV_INTEL_function_pointers/const-function-pointer.ll
```diff
@@ -0,0 +1,68 @@
+; RUN: llvm-as %s -o %t.bc
+; RUN: llvm-spirv %t.bc -spirv-text --spirv-ext=+SPV_INTEL_function_pointers -o %t.spt
+; RUN: FileCheck < %t.spt %s --check-prefix=CHECK-SPIRV
+; RUN: llvm-spirv %t.bc --spirv-ext=+SPV_INTEL_function_pointers -o %t.spv
```
AlexeySachkov, line 4:
```suggestion
; RUN: llvm-spirv %t.bc --spirv-ext=+SPV_INTEL_function_pointers -o %t.spv
; RUN: llvm-spirv %t.spv -to-text -o - | FileCheck %s --check-prefix=CHECK-SPIRV
```

lib/SPIRV/libSPIRV/SPIRVFunction.h
zuban32: Due to the same reason why other OpConstant* are not as well

lib/SPIRV/SPIRVWriter.cpp
zuban32: Yep

test/transcoding/SPV_INTEL_function_pointers/const-function-pointer.ll
zuban32: Fixed

lib/SPIRV/SPIRVWriter.cpp
zuban32: Fixed

lib/SPIRV/SPIRVWriter.h
```diff
@@ -117,10 +127,13 @@ class LLVMToSPIRV : public ModulePass {
   void transFPContract();
   SPIRVValue *transConstant(Value *V);
   SPIRVValue *transValue(Value *V, SPIRVBasicBlock *BB,
-                         bool CreateForward = true);
+                         bool CreateForward = true,
+                         FuncTransMode FuncTrans = FuncTransMode::Decl);
```
svenvh, line 131: It feels a bit inelegant to add a new argument to a generic function like `transValue` and `transValueWithoutDecoration` just to control function pointer translation.  I haven't really looked into any alternative ways so maybe it's the only reasonable way; but I'm wondering if you have evaluated different approaches?

zuban32: Well, IMO it's way better than the previous way of function pointers translation. As you may see, this arg allowed us to remove a good amount of code duplication caused by a need to support function pointers in a various instructions, e.g. for Store and InsertElement (and unary ops, etc) we had ~the same funcptrs related code. And as we don't have any globals translation state to keep this option in, I don't see any feasible alternative.

lib/SPIRV/SPIRVWriter.cpp
zuban32: Done

lib/SPIRV/SPIRVWriter.h
svenvh: Yes I see the removed code duplication which is a good thing.  I was hoping you could deduce the correct behaviour inside transValue itself from the arguments, but I suppose you need the context of the Values involved, so that may not be possible.

zuban32: Yes, basically we need both Use and User here to determine how we want to translate the User. Adding that would result in a major refactoring of all the code using it, doesn't look doable.

PR #664, opened 2020-08-05
README/CI adjustments on llvm_release_110 branch

AlexeySachkov:


README.md
```diff
@@ -32,7 +32,7 @@ sudo apt-get install llvm-11-dev llvm-11-tools clang-11 libclang-11-dev
 ```
 The installed version of LLVM will be used by default for out-of-tree build of the translator.
 ```
-git clone https://github.com/KhronosGroup/SPIRV-LLVM-Translator.git
+git clone https://github.com/KhronosGroup/SPIRV-LLVM-Translator.git -b llvm_release_110
```
AlexeySotkin, line 35: Probably we might use asciidoctor substitutions here https://asciidoctor.org/docs/asciidoc-syntax-quick-reference/#attributes-and-substitutions

AlexeySachkov: Unfortunately, we are using Markdown here and not asciidoctor.

We could try and use that feature, but I afraid that GitHub render doesn't it and I think that most of users look into this README on GitHub and not into manually rendered version

PR #665, opened 2020-08-06
Add support for usub.sat intrinsic

Fznamznon:
Translate usub.sat(a, b) as (a > b) ? a - b : 0 sequence.

### DIFF
--- a/lib/SPIRV/SPIRVWriter.cpp
+++ b/lib/SPIRV/SPIRVWriter.cpp
@@ -2023,6 +2023,19 @@ SPIRVValue *LLVMToSPIRV::transIntrinsicInst(IntrinsicInst *II,
     return BM->addBinaryInst(OpFAdd, Ty, Mul,
                              transValue(II->getArgOperand(2), BB), BB);
   }
+  case Intrinsic::usub_sat: {
+    // usub.sat(a, b) -> (a > b) ? a - b : 0
+    SPIRVType *Ty = transType(II->getType());
+    Type *BoolTy = IntegerType::getInt1Ty(M->getContext());
+    SPIRVValue *Sub =
+        BM->addBinaryInst(OpISub, Ty, transValue(II->getArgOperand(0), BB),
+                          transValue(II->getArgOperand(1), BB), BB);
+    SPIRVValue *Cmp = BM->addCmpInst(
+        OpUGreaterThan, transType(BoolTy), transValue(II->getArgOperand(0), BB),
+        transValue(II->getArgOperand(1), BB), BB);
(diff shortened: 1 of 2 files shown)


test/usub-sat.ll
```diff
@@ -0,0 +1,37 @@
...
+
+; CHECK: Constant {{[0-9]+}} [[three:[0-9]+]] 3
+; CHECK: Constant {{[0-9]+}} [[zero:[0-9]+]] 0
+
+; CHECK: Function
+; CHECK: FunctionParameter {{[0-9]+}} [[x:[0-9]+]]
+; CHECK: ISub {{[0-9]+}} [[sub_res:[0-9]+]] [[three]] [[x]]
+; CHECK: UGreaterThan {{[0-9]+}} [[cmp_res:[0-9]+]] [[three]] [[x]]
+; CHECK: Select {{[0-9]+}} [[sel_res:[0-9]+]] [[cmp_res]] [[sub_res]] [[zero]]
+
```
AlexeySachkov, line 24: Could you please also check that `[[sel_res]]` is used as a return value?

lib/SPIRV/SPIRVWriter.cpp
```diff
@@ -2023,6 +2023,19 @@ SPIRVValue *LLVMToSPIRV::transIntrinsicInst(IntrinsicInst *II,
...
+  case Intrinsic::usub_sat: {
+    // usub.sat(a, b) -> (a > b) ? a - b : 0
+    SPIRVType *Ty = transType(II->getType());
+    Type *BoolTy = IntegerType::getInt1Ty(M->getContext());
+    SPIRVValue *Sub =
+        BM->addBinaryInst(OpISub, Ty, transValue(II->getArgOperand(0), BB),
+                          transValue(II->getArgOperand(1), BB), BB);
+    SPIRVValue *Cmp = BM->addCmpInst(
+        OpUGreaterThan, transType(BoolTy), transValue(II->getArgOperand(0), BB),
+        transValue(II->getArgOperand(1), BB), BB);
```
AlexeySachkov, line 2035: Minor: would it be better to store translated arguments to local variables to avoid calling `transValue` twice for the same value?

Note: this is not a performance improvement as `transValue` already has a cache of translated values, this is just code style question

test/usub-sat.ll
Fznamznon: Ok, done.

lib/SPIRV/SPIRVWriter.cpp
Fznamznon: Okay, done.

PR #671, opened 2020-08-10
Add OpenCL.DebugInfo.100 as alias to SPIRV.debug on llvm_release_70 branch

NikitaRudenkoIntel:
Docs: https://www.khronos.org/registry/spir-v/specs/unified1/OpenCL.DebugInfo.100.html

The incorrect "SPIRV.debug" name is preserved for backward compatibility: to allow newer
translator versions consume SPIR-V files generated by the older translator versions.


AlexeySachkov:
Do we really need to backport this to 7.0 branch?

Please also note that this is a breaking change: SPIR-V generated by newer version from llvm_release_70 won't be consumed by any current released translator from llvm_release_70 branch

NikitaRudenkoIntel:
Yes, we need it on all branches including llvm_release_70.
The problem is that if don't submit it on this branch, the translator here will not be able to consume any spirv from any other branches.
For example, in one of our projects we have a LLVM8---[llvm_release_80 translator]--->spirv---[llvm_release_70 translator]--->LLVM7 path.

AlexeySotkin (approved):

PR #672, opened 2020-08-10
Support llvm.is.constant

AlexeySachkov:
This might be not the best level of support for the intrinsic, as further optimizations could help to get more precise results (i.e. constant propagation, inlining, etc.), but at the same time it should be good enough to preserve functionality of modules containing `@llvm.is.constant` intrinsic.

The issue were reported in intel/llvm#2122 and can be reproduced by compiling SYCL app with C++ headers from GCC9 (they contain calls to `__builtin_constant_p` which produces `@llvm.is.constant` intrinsic)

### DIFF
--- a/lib/SPIRV/SPIRVWriter.cpp
+++ b/lib/SPIRV/SPIRVWriter.cpp
@@ -48,6 +48,7 @@
 #include "SPIRVFunction.h"
 #include "SPIRVInstruction.h"
 #include "SPIRVInternal.h"
+#include "SPIRVLLVMUtil.h"
 #include "SPIRVMDWalker.h"
 #include "SPIRVModule.h"
 #include "SPIRVType.h"
@@ -2744,6 +2745,13 @@ SPIRVValue *LLVMToSPIRV::transIntrinsicInst(IntrinsicInst *II,
     // llvm.trap intrinsic is not implemented. But for now don't crash. This
     // change is pending the trap/abort intrinsic implementation.
     return nullptr;
+  case Intrinsic::is_constant: {
+    auto *CO = dyn_cast<Constant>(II->getOperand(0));
+    if (CO && isManifestConstant(CO))
+      return transValue(ConstantInt::getTrue(II->getType()), BB, false);
+    else
+      return transValue(ConstantInt::getFalse(II->getType()), BB, false);
+  }
   default:
     if (BM->isSPIRVAllowUnknownIntrinsicsEnabled())
       return BM->addCallInst(
--- a/lib/SPIRV/libSPIRV/SPIRVLLVMUtil.h
+++ b/lib/SPIRV/libSPIRV/SPIRVLLVMUtil.h
@@ -1,0 +1,36 @@
+//===- SPIRVLLVMUtil.h - SPIR-V LLVM-specific Utility Functions -*- C++ -*-===//
+//
+// Has inclusions from the LLVM Project, under the Apache License v2.0 with LLVM
+// Exceptions. See https://llvm.org/LICENSE.txt for license information.
+// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+//
+//===----------------------------------------------------------------------===//
+/// \file
+///
+/// This file defines utility functions dedicated to processing LLVM classes
+///
+//===----------------------------------------------------------------------===//
+
+#ifndef SPIRV_LIBSPIRV_SPIRVLLVMUTIL_H
+#define SPIRV_LIBSPIRV_SPIRVLLVMUTIL_H
+
+#include "llvm/IR/Constants.h"
+#include "llvm/Support/Casting.h"
+
+namespace SPIRV {
+inline bool isManifestConstant(const llvm::Constant *C) {
+  if (llvm::isa<llvm::ConstantData>(C)) {
+    return true;
+  } else if (llvm::isa<llvm::ConstantAggregate>(C) ||
+             llvm::isa<llvm::ConstantExpr>(C)) {
+    for (const llvm::Value *Subc : C->operand_values()) {
+      if (!isManifestConstant(llvm::cast<llvm::Constant>(Subc)))
+        return false;
+    }
+    return true;
+  }
+  return false;
+}
+} // namespace SPIRV
+
+#endif // SPIRV_LIBSPIRV_SPIRVLLVMUTIL_H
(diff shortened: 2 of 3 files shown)


lib/SPIRV/SPIRVWriter.cpp
```diff
@@ -259,6 +259,19 @@ static bool recursiveType(const StructType *ST, const Type *Ty) {
   return Run(Ty);
 }
 
+static bool isManifestConstant(const Constant *c) {
```
svenvh, line 262: This seems like a plain copy of an LLVM function.  This function is not exported by LLVM, so I'm wondering if it would be better to call a higher level function (something from `llvm/include/llvm/Analysis/ConstantFolding.h` perhaps)?  That will also give the benefit of actually folding anything it can.

AlexeySotkin:
@jinge90 please answer to the [comment ](https://github.com/KhronosGroup/SPIRV-LLVM-Translator/pull/672#discussion_r468422442) from @svenvh and sign the CLA.

lib/SPIRV/SPIRVWriter.cpp
AlexeySachkov: Looking at https://llvm.org/doxygen/ConstantFolding_8h.html I cannot find anything that can actually be used here. I mean we don't know what exactly is passed `Intrinsic::is_constant` and there is no generic function `tryToFoldWhateverItIs`

bader:
> @jinge90 please answer to the [comment ](https://github.com/KhronosGroup/SPIRV-LLVM-Translator/pull/672#discussion_r468422442) from @svenvh and sign the CLA.

@AlexeySachkov, could you address this request, please?

lib/SPIRV/SPIRVWriter.cpp
jinge90: Hi, @svenvh 
Yes, "isManifestConstant" is a plain copy of LLVM function from ConstantFolding.cpp. I understand that code duplicate won't make us happy here. However, "isManifestConstant" is a very simple function with few lines of code, it seems that bringing "isManifestConstant" into spirv translator won't introduce much code redundancy.
And I searched the invocations to "isManifestConstant" in llvm-project but couldn't find a general higher level function which meets our requirements here. "isManifestConstant" is only invoked by "ConstantFoldScalarCall1" which is not exported by llvm either.
Thank you very much.

AlexeySachkov:
> > @jinge90 please answer to the [comment ](https://github.com/KhronosGroup/SPIRV-LLVM-Translator/pull/672#discussion_r468422442) from @svenvh and sign the CLA.
> 
> @AlexeySachkov, could you address this request, please?

CLA appears to be signed already

lib/SPIRV/SPIRVWriter.cpp
svenvh: I agree it's a small function so from that perspective it shouldn't be much of a problem.  A remaining concern is on the attribution/licensing side: I am not a lawyer, but this file seems to carry an outdated header so I think we should be careful about what we add to this file.

jinge90, line 262: Hi, @AlexeySachkov 
Based the pre-ci log, we have some tidying errors here, could you help replace the input parameter "const Constant  *c" with "const Constant *C"?
Thanks very much.

lib/SPIRV/SPIRVWriter.cpp
```diff
@@ -259,6 +259,19 @@ static bool recursiveType(const StructType *ST, const Type *Ty) {
...
+  if (isa<ConstantData>(c)) {
+    return true;
+  } else if (isa<ConstantAggregate>(c) || isa<ConstantExpr>(c)) {
+    for (const Value *subc : c->operand_values()) {
```
jinge90, line 266: The same for "subc" here, should change to "Subc". The error log from pre-ci system:
Found tidying errors!
/home/travis/build/KhronosGroup/SPIRV-LLVM-Translator/lib/SPIRV/SPIRVWriter.cpp:262:48: warning: invalid case style for parameter 'c' [readability-identifier-naming]
static bool isManifestConstant(const Constant *c) {
                                               ^~
                                               C
/home/travis/build/KhronosGroup/SPIRV-LLVM-Translator/lib/SPIRV/SPIRVWriter.cpp:266:23: warning: invalid case style for variable 'subc' [readability-identifier-naming]
    for (const Value *subc : c->operand_values()) {
                      ^~~~
Thanks very much.

lib/SPIRV/SPIRVWriter.cpp
jinge90: Hi, @svenvh 
Thanks very much for your reminder, do we need to go through some legal process for this patch?
Thanks very much.

AGindinson: @svenvh, after discussion with @AlexeySachkov, I've factored out the "borrowed" code into a separate utility header. I've given that an Apache 2.0 w/ LLVM exceptions header (same as in LLVM's `lib/Analysis/ConstantFolding.cpp`). Could you please take a look at the new approach and approve the PR if this looks safe enough?

lib/SPIRV/libSPIRV/SPIRVUtil.h
```diff
@@ -40,6 +40,8 @@
 #ifndef SPIRV_LIBSPIRV_SPIRVUTIL_H
 #define SPIRV_LIBSPIRV_SPIRVUTIL_H
 
+#include "SPIRVLLVMUtil.h" // LLVM-specific utilities
```
svenvh, line 43: As this transitively includes a lot of LLVM headers, preferably include this only where you really need it, so in SPIRVWriter.cpp.  This helps keeping llvm-spirv compilation times reasonable.

Also, the comment is not really needed in my opinion.

lib/SPIRV/SPIRVWriter.cpp
svenvh: @AGindinson thanks.  I am not a lawyer, but this looks like a reasonable solution.  Please make sure to squash this PR into one commit, so that the wrong license header does not end up in the git history.

Other than a minor comment posted below, I'll be happy to approve this PR.

lib/SPIRV/libSPIRV/SPIRVUtil.h
AGindinson: Fair enough, done.

lib/SPIRV/SPIRVWriter.cpp
AGindinson: Applied both the comment and the squash. Thanks!

svenvh (approved):

PR #673, opened 2020-08-10
Remove unused NameMaps

svenvh:
For many enums in `spirv.hpp`, a NameMap was generated in `SPIRVNameMapEnum.h` even though that NameMap was never used.  Remove them to reduce build times and maintenance burden.

Sync SPIRVNameMapEnum.h with gen_spirv.bash by regenerating SPIRVNameMapEnum.h using `gen_spirv.bash`.

This is a followup from https://github.com/KhronosGroup/SPIRV-LLVM-Translator/pull/635


lib/SPIRV/libSPIRV/SPIRVNameMapEnum.h
```diff
@@ -53,254 +53,14 @@ using namespace spv;
 
 namespace SPIRV {
 
-template <> inline void SPIRVMap<SourceLanguage, std::string>::init() {
```
AlexeySachkov, line 56: Actually, absence of uses for those maps means lack of feature support in the translator: for example, I would imagine that we could use those to nicely print SourceLanguage information in textual SPIR-V representation.

This is for sure not a request for any changes and just a one more starting point for further improvements

svenvh: Agreed, though I'm under the impression that we may eventually move away from the translator's textual SPIR-V representation, in favor of the SPIR-V tools representation: see https://github.com/KhronosGroup/SPIRV-LLVM-Translator/issues/211

PR #675, opened 2020-08-12
Fix ptr.annotation translation for pointers on non-struct types

mlychkov:


mlychkov:
Tagging @AlexeySachkov @AlexeySotkin @AGindinson @MrSidims

test/transcoding/fpga_lsu_optimized.ll
```diff
@@ -0,0 +1,166 @@
...
+;   int output_data[2];
+;   for (size_t i = 0; i < 2; i++) {
+;     output_data[i] = -1;
+;   }
+;
+;   int input_data[2];
+;   for (size_t i = 0; i < 2; i++) {
+;     input_data[i] = i + 1;
+;   }
+;
```
AlexeySachkov, line 17:
```suggestion
```

I think there is no need to put host code in there as it anyway is not emitted to LLVM IR. The same applies to code that validates results after kernel enqueue

mlychkov: Thank you for a quick feedback. Removed.

MrSidims:
Minor: I suggest to add 'vendor name_' prefix for vendor-specific tests.

mlychkov:
> Minor: I suggest to add 'vendor name_' prefix for vendor-specific tests.

Yes, I agree, good suggestion. Thank you.

PR #676, opened 2020-08-12
Fix emitting SPIRV metadata for non-kernel functions in PreprocessMetadata.cpp

NikitaRudenkoIntel:
SPIRV metadata contains pointers to relevant functions
Passes like SPIRVLowerOCLBlocks and SPIRVRegularizeLLVM can erase
declarations with no usage and invalidate those pointers

### DIFF
--- a/lib/SPIRV/PreprocessMetadata.cpp
+++ b/lib/SPIRV/PreprocessMetadata.cpp
@@ -255,6 +255,9 @@ void PreprocessMetadata::preprocessVectorComputeMetadata(Module *M,
   auto EM = B->addNamedMD(kSPIRVMD::ExecutionMode);
 
   for (auto &F : *M) {
+    if (F.getCallingConv() != CallingConv::SPIR_KERNEL)
+      continue;
+
     // Add VC float control execution modes
     // RoundMode and FloatMode are always same for all types in VC
     // While Denorm could be different for double, float and half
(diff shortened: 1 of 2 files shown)


AlexeySachkov:
> This pass creates metadata entries with pointers to LLVM IR functions
> Passes which come after could modify IR and break these pointers

@NikitaRudenkoIntel, could you please adjust commit message and PR title/description to align it with the update changes?

NikitaRudenkoIntel:
Sure. Also added a test case. It does not check any spirv entries, but crashes without this patch.

AlexeySachkov:
The problem seems to appear because of `SPIRVRegularizeLLVM` pass, which erases unused declarations: reference to a function from a metadata is not considered as `Use` (it is not clear to me why so?).

However, function in the test in this PR has `spir_kernel` calling convention and we have added some metadata that points to it - later when we perform translation of execution mode, we crash because `isa` stumbles on `nullptr`

I think that I'm okay with this particular change, but I still wonder:
- do we have some particular documented interface for defining kernels in LLVM IR? I mean we could consider this code as invalid input and in this case we need another patch that will emit error for such functions
- do we need to change our way of detecting kernels which we use to create metadata for encoding execution modes?

Tagging @svenvh  and @AlexeySotkin here

NikitaRudenkoIntel:
I looks like I mistreated the problem.
The translator will crash on previous version of attached test:
```
declare dllexport spir_kernel void @test_func(i32) #0
attributes #0 = { "VCSLMSize"="0" }
```
But I guess that kernel without definition is an exceptional case.
The original problem was crashing on regular function due to emitting execution mode metadata for it with following DCE. This patch fixes this case.

AlexeySachkov (approved):

PR #678, opened 2020-08-14
Fix translation of llvm.ptr.annotation

vmaksimo:
This patch follows from 71039dd3c5dbfafab78f6b86a5bb06ba2af93c89

* Replace the destination operand with the result of its intrinsic call for store inst.
* Replace the read-from operand with the result of its intrinsic call for load inst.

### DIFF
--- a/lib/SPIRV/SPIRVReader.cpp
+++ b/lib/SPIRV/SPIRVReader.cpp
@@ -1760,6 +1760,15 @@ Value *SPIRVToLLVM::transValueWithoutDecoration(SPIRVValue *BV, Function *F,
           Src = II;
       }
     }
+    // A pointer annotation may have been generated for the destination variable.
+    if (Use *SingleUse = Dst->getSingleUndroppableUse()) {
+      if (auto *II = dyn_cast<IntrinsicInst>(SingleUse->getUser())) {
+        if (II->getIntrinsicID() == Intrinsic::ptr_annotation &&
+            II->getType() == Dst->getType())
+          // Overwrite the future store operand with the intrinsic call result.
+          Dst = II;
+      }
+    }
     bool isVolatile = BS->SPIRVMemoryAccess::isVolatile();
     uint64_t AlignValue = BS->SPIRVMemoryAccess::getAlignment();
     if (0 == AlignValue)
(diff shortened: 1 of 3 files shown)


vmaksimo:
Tagging @mlychkov @AGindinson

mlychkov:
Could you please also replace TODOs in test/transcoding/intel_fpga_lsu_optimized.ll file with actual load/store commands?

lib/SPIRV/SPIRVReader.cpp
```diff
@@ -1760,6 +1760,15 @@ Value *SPIRVToLLVM::transValueWithoutDecoration(SPIRVValue *BV, Function *F,
           Src = II;
       }
     }
+    // A pointer annotation may have been generated for the destination variable.
```
mlychkov, line 1763: Maybe to move this in a separate function? It is used in 3 places.

vmaksimo:
As I see, CI error is not related to this patch

lib/SPIRV/SPIRVReader.cpp
```diff
@@ -1412,6 +1412,18 @@ Value *SPIRVToLLVM::oclTransConstantPipeStorage(
                             GlobalValue::NotThreadLocal, SPIRAS_Global);
 }
 
+static void replaceOperandWithIntrinsicCallResult(Value *&V) {
```
AlexeySachkov, line 1415:
```suggestion
static void replaceOperandWithAnnotationIntrinsicCallResult(Value *&V) {
```

Please also add a comment about the purpose of this function, because it relies on some particular LLVM IR representation which is only used for FPGA-specific extensions.

mlychkov, line 1415: This is minor comment, applying of it is at your discretion.
I think that it is clearer that a function changes its argument if the call looks like `Val = getAnnotationIntrinsicResult(Val);`

mlychkov:
> Could you please also replace TODOs in test/transcoding/intel_fpga_lsu_optimized.ll file with actual load/store commands?

Thank you very much.

AlexeySachkov (approved):

mohammadfawaz:
Is this ready to be merged? I'm only asking because we need this patch in our downstream version of the reverse translator. Thanks!

AlexeySachkov:
> Is this ready to be merged? I'm only asking because we need this patch in our downstream version of the reverse translator. Thanks!

Sure, done

PR #679, opened 2020-08-17
Support missing DWARF opcodes

asavonic:
Signed-off-by: Andrew Savonichev <andrew.savonichev@intel.com>


asavonic:
> This operation must be added to the spec, probably as an extension.

What about the existing DW_OP_fragment?

AlexeySotkin:
> > This operation must be added to the spec, probably as an extension.
> 
> What about the existing DW_OP_fragment?

It is in the spec https://www.khronos.org/registry/spir-v/specs/unified1/OpenCL.DebugInfo.100.html#_debug_operations_a_id_operation_a

lib/SPIRV/LLVMToSPIRVDbgTran.cpp
```diff
@@ -968,6 +973,11 @@ SPIRVEntry *LLVMToSPIRVDbgTran::transDbgExpression(const DIExpression *Expr) {
     assert(OpCountMap.find(OC) != OpCountMap.end() &&
            "unhandled opcode found in DIExpression");
     unsigned OpCount = OpCountMap[OC];
+    if (isUnsupportedDIE(OC)) {
+      assert(OpCount >= 1);
+      I += OpCount - 1;
+      continue;
```
AlexeySotkin, line 979: I think, if we just skip an operation we get a wrong expression, don't we?
If we want to avoid a crash, we might:
 * Skip the intrinsic(s)(e.g. `llvm.dbg.value`) which refers to the expression with unsupported operation.
 * Lower the expression so it contains only supported operations, like it is done in [this test](https://github.com/llvm/llvm-project/blob/7b64765cd19a60d6d56106769d06cdff5b660479/llvm/test/DebugInfo/X86/convert-debugloc.ll#L37)

test/DebugInfo/X86/convert-debugloc.ll
```diff
@@ -0,0 +1,46 @@
+; RUN: llvm-as < %s -o %t.bc
+; RUN: llvm-spirv %t.bc -o %t.spv
+; RUN: llvm-spirv -r %t.spv -o - | llvm-dis -o %t.ll
+; FileCheck %s --implicit-check-not=DW_OP_convert --input-file=%t.ll
+
+; RUN: llc -mtriple=%triple -dwarf-version=5 -filetype=obj -O0 < %t.ll
+; RUN: llc -mtriple=%triple -dwarf-version=4 -filetype=obj -O0 < %t.ll
+
```
AlexeySotkin, line 8: Why there is no `CHECK:` lines in this test?

lib/SPIRV/LLVMToSPIRVDbgTran.cpp
asavonic: > Lower the expression so it contains only supported operations, like it is done in this test

That sounds good, but I cannot find even basic operations like mul or shift in the list of supported [operations](https://www.khronos.org/registry/spir-v/specs/unified1/OpenCL.DebugInfo.100.html#_debug_operations_a_id_operation_a). Am I missing something or none of them is supported?
To lower DW_OP_convert we need DW_OP_dup, shr, lit0, not, mul, or.

AlexeySotkin: > > Lower the expression so it contains only supported operations, like it is done in this test
> 
> That sounds good, but I cannot find even basic operations like mul or shift in the list of supported [operations](https://www.khronos.org/registry/spir-v/specs/unified1/OpenCL.DebugInfo.100.html#_debug_operations_a_id_operation_a). Am I missing something or none of them is supported?
> To lower DW_OP_convert we need DW_OP_dup, shr, lit0, not, mul, or.

You are right, they are not supported in SPIR-V. It seems this option doesn't work in our case.

AlexeySotkin:
Please fix the lit test

AlexeySotkin:
The operations added with patch are not compliant with the [OpenCL.DebugInfo.100](https://www.khronos.org/registry/spir-v/specs/unified1/OpenCL.DebugInfo.100.html) spec. I think, if we want to add these operations we must also add an option which will enable their generation.(do not generate these operations by default).

AlexeySotkin:
It would be nice to have some documentation for the operations or a reference to the DWARF spec.

AlexeySotkin:
As far as I understand the purpose of this patch is to cover all possible debug operations. So we are adding all operations from the DWARF 5. But note that LLVM may add its own operations, with different name, semantics or number of operands. For example, `DW_OP_LLVM_convert` has 2 operands, while `DW_OP_convert` has only 1. My point is, there is a good chance we will have to update the header if/when LLVM will add more operations.

lib/SPIRV/libSPIRV/SPIRV.debug.h
```diff
@@ -498,16 +808,171 @@ typedef SPIRVMap<dwarf::LocationAtom, SPIRVDebug::ExpressionOpCode>
...
+  add(dwarf::DW_OP_plus,                SPIRVDebug::Plus);
+  add(dwarf::DW_OP_minus,               SPIRVDebug::Minus);
+  add(dwarf::DW_OP_plus_uconst,         SPIRVDebug::PlusUconst);
+  add(dwarf::DW_OP_bit_piece,           SPIRVDebug::BitPiece);
+  add(dwarf::DW_OP_swap,                SPIRVDebug::Swap);
+  add(dwarf::DW_OP_xderef,              SPIRVDebug::Xderef);
+  add(dwarf::DW_OP_stack_value,         SPIRVDebug::StackValue);
+  add(dwarf::DW_OP_constu,              SPIRVDebug::Constu);
+  add(dwarf::DW_OP_LLVM_fragment,       SPIRVDebug::Fragment);
+  add(dwarf::DW_OP_LLVM_convert,        SPIRVDebug::Convert);
```
AlexeySotkin, line 821: Seems like `DW_OP_LLVM_tag_offset` and `DW_OP_LLVM_entry_value` are missing.
https://github.com/llvm/llvm-project/blob/master/llvm/include/llvm/BinaryFormat/Dwarf.h#L116-L125

lib/SPIRV/libSPIRV/SPIRV.debug.h
```diff
@@ -498,16 +808,171 @@ typedef SPIRVMap<dwarf::LocationAtom, SPIRVDebug::ExpressionOpCode>
...
+  add(dwarf::DW_OP_implicit_value,      SPIRVDebug::ImplicitValue);
+  add(dwarf::DW_OP_implicit_pointer,    SPIRVDebug::ImplicitPointer);
+  add(dwarf::DW_OP_addrx,               SPIRVDebug::Addrx);
+  add(dwarf::DW_OP_constx,              SPIRVDebug::Constx);
+  add(dwarf::DW_OP_entry_value,         SPIRVDebug::EntryValue);
+  add(dwarf::DW_OP_const_type,          SPIRVDebug::ConstTypeOp);
+  add(dwarf::DW_OP_regval_type,         SPIRVDebug::RegvalType);
+  add(dwarf::DW_OP_deref_type,          SPIRVDebug::DerefType);
+  add(dwarf::DW_OP_xderef_type,         SPIRVDebug::XderefType);
+  add(dwarf::DW_OP_reinterpret,         SPIRVDebug::Reinterpret);
```
AlexeySotkin, line 975: `dwarf::DW_OP_convert` is missing
https://github.com/llvm/llvm-project/blob/master/llvm/include/llvm/BinaryFormat/Dwarf.def#L664

lib/SPIRV/libSPIRV/SPIRV.debug.h
```diff
@@ -432,16 +587,171 @@ enum {
...
+  { Plus,               1 },
+  { Minus,              1 },
+  { PlusUconst,         2 },
+  { BitPiece,           3 },
+  { Swap,               1 },
+  { Xderef,             1 },
+  { StackValue,         1 },
+  { Constu,             2 },
+  { Fragment,           3 },
+  { Convert,            3 },
```
AlexeySachkov, line 600: Why 3 operands? `DW_OP_LLVM_convert` takes two and `DW_OP_convert` takes only one

AlexeySotkin: Numbers in this map are number of operands +1.

AlexeySachkov: Thanks, this is just a bit confusing

asavonic:
> As far as I understand the purpose of this patch is to cover all possible debug operations. So we are adding all operations from the DWARF 5. 

Correct.

> But note that LLVM may add its own operations, with different name, semantics or number of operands. For example, `DW_OP_LLVM_convert` has 2 operands, while `DW_OP_convert` has only 1. 

DW_OP_LLVM_convert and DW_OP_convert have identical semantic, just different representation.
LLVM one is easier to represent with the current DebugInfo spec, so I opted to use it instead of the DWARF one.

> My point is, there is a good chance we will have to update the header if/when LLVM will add more operations.

Remember that whatever LLVM uses in IR metadata must be lowered to DWARF. Therefore unless a new operation is hardware specific, there is a good chance we can express it with existing DWARF operations.

lib/SPIRV/libSPIRV/SPIRV.debug.h
asavonic: DW_OP_LLVM_convert is used instead.

lib/SPIRV/libSPIRV/SPIRV.debug.h
asavonic: `DW_OP_LLVM_entry_value` is only used for MIR. I'll check `DW_OP_LLVM_tag_offset`.

AlexeySotkin:
> Remember that whatever LLVM uses in IR metadata must be lowered to DWARF. Therefore unless a new operation is hardware specific, there is a good chance we can express it with existing DWARF operations.

Fair enough.

lib/SPIRV/libSPIRV/SPIRV.debug.h
```diff
@@ -432,16 +587,171 @@ enum {
...
+  { Mod,                1 },
+  { Mul,                1 },
+  { Neg,                1 },
+  { Not,                1 },
+  { Or,                 1 },
+  { Shl,                1 },
+  { Shr,                1 },
+  { Shra,               1 },
+  { Xor,                1 },
+  { Bra,                2 },
```
AlexeySachkov, line 628: This one should probably be excluded as well

asavonic:
> The operations being added with this patch are not compliant with the [OpenCL.DebugInfo.100](https://www.khronos.org/registry/spir-v/specs/unified1/OpenCL.DebugInfo.100.html) spec. I think, if we want to add these operations we must also add an option which will enable their generation, i.e. do not generate these operations by default.

Support new operations is now controlled by `--spirv-allow-extra-diexpressions` option (disabled by default).

lib/SPIRV/libSPIRV/SPIRV.debug.h
asavonic: Thanks, done.

tools/llvm-spirv/llvm-spirv.cpp
```diff
@@ -183,6 +183,11 @@ cl::opt<bool> SPIRVAllowUnknownIntrinsics(
     cl::desc("Unknown LLVM intrinsics will be translated as external function "
              "calls in SPIR-V"));
 
+static cl::opt<bool> SPIRVAllowExtraDIExpressions(
+    "spirv-allow-extra-diexpressions", cl::init(false),
+    cl::desc("Allow DIExpression operations not listed in the SPIR-V DebugInfo "
+             "specification (experimental)"));
```
AlexeySotkin, line 189:
```suggestion
    cl::desc("Allow DWARF operations not listed in the OpenCL.DebugInfo.100 "
             "specification (experimental, may produce incompatible SPIR-V "
             "module)"));
```

lib/SPIRV/LLVMToSPIRVDbgTran.cpp
```diff
@@ -961,8 +961,12 @@ SPIRVEntry *LLVMToSPIRVDbgTran::transDbgExpression(const DIExpression *Expr) {
...
     SPIRVDebug::ExpressionOpCode OC =
         SPIRV::DbgExpressionOpCodeMap::map(DWARFOpCode);
-    assert(OpCountMap.find(OC) != OpCountMap.end() &&
-           "unhandled opcode found in DIExpression");
+    if (OpCountMap.find(OC) == OpCountMap.end()) {
+      report_fatal_error("unknown opcode found in DIExpression");
+    }
+    if (OC > SPIRVDebug::Fragment && !BM->allowExtraDIExpressions()) {
+      report_fatal_error("unsupported opcode found in DIExpression");
+    }
```
AlexeySotkin, line 969: Nitpick: curly braces aren't needed.

tools/llvm-spirv/llvm-spirv.cpp
asavonic: Done.

lib/SPIRV/LLVMToSPIRVDbgTran.cpp
asavonic: Done.

AlexeySotkin (approved):

PR #683, opened 2020-08-17
Fix handling of phi nodes in LowerConstExprs.

zuban32:
Put newly created insertElement instructions at the ends
of the incoming blocks instead of putting them immediately before
the phi thus breaking the module.

### DIFF
--- a/lib/SPIRV/SPIRVLowerConstExpr.cpp
+++ b/lib/SPIRV/SPIRVLowerConstExpr.cpp
@@ -162,6 +162,8 @@ void SPIRVLowerConstExpr::visit(Module *M) {
                          [LowerOp](Value *V) { return LowerOp(V); });
           Value *Repl = nullptr;
           unsigned Idx = 0;
+          auto *PhiII = dyn_cast<PHINode>(II);
+          auto *InsPoint = PhiII ? &PhiII->getIncomingBlock(OI)->back() : II;
           std::list<Instruction *> ReplList;
           for (auto V : OpList) {
             if (auto *Inst = dyn_cast<Instruction>(V))
@@ -169,7 +171,7 @@ void SPIRVLowerConstExpr::visit(Module *M) {
             Repl = InsertElementInst::Create(
                 (Repl ? Repl : UndefValue::get(Vec->getType())), V,
                 ConstantInt::get(Type::getInt32Ty(M->getContext()), Idx++), "",
-                II);
+                InsPoint);
           }
           II->replaceUsesOfWith(Op, Repl);
           WorkList.splice(WorkList.begin(), ReplList);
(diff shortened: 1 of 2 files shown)


lib/SPIRV/SPIRVLowerConstExpr.cpp
```diff
@@ -162,14 +162,16 @@ void SPIRVLowerConstExpr::visit(Module *M) {
                          [LowerOp](Value *V) { return LowerOp(V); });
           Value *Repl = nullptr;
           unsigned Idx = 0;
+          auto *PhiII = dyn_cast<PHINode>(II);
+          auto *InsPoint = PhiII ? &PhiII->getIncomingBlock(OI)->back() : II;
```
AlexeySachkov, line 166: What about using [`getFirstInsertionPt`](https://llvm.org/doxygen/classllvm_1_1BasicBlock.html#a161fd4e9fa5367f64c2a4c9e921c3ad3)?

test/constexpr_phi.ll
```diff
@@ -0,0 +1,56 @@
+; RUN: llvm-as %s -o %t.bc
+; RUN: llvm-spirv %t.bc -spirv-text --spirv-ext=+SPV_INTEL_function_pointers -o %t.spt
+; RUN: FileCheck < %t.spt %s --check-prefix=CHECK-SPIRV
+; RUN: llvm-spirv %t.bc --spirv-ext=+SPV_INTEL_function_pointers -o %t.spv
```
AlexeySachkov, line 4:
```suggestion
; RUN: llvm-spirv %t.bc --spirv-ext=+SPV_INTEL_function_pointers -o %t.spv
; RUN: llvm-spirv %t.spv -to-text -o %t.spt
; RUN: FileCheck < %t.spt %s --check-prefix=CHECK-SPIRV

```

lib/SPIRV/SPIRVLowerConstExpr.cpp
zuban32: AFAIR getFirstInsertionPt return the first instruction in the BB we can insert new inst before, but we can't do that here as phi should use that instructions

test/constexpr_phi.ll
zuban32: Fixed

lib/SPIRV/SPIRVLowerConstExpr.cpp
AlexeySachkov: > AFAIR getFirstInsertionPt returns the first instruction in the BB we can insert new inst before, but we can't do that here as phi already uses that instructions

Ah, got it - we need to insert new instruction into a predecessor BB, so it is already known/computed prior encountering phi - right?

AlexeySachkov:
@zuban32, could you please rebase your PR one more time to incorporate build fixes in it?

PR #697, opened 2020-08-21
Fix translation of a spec constant with multiple uses

bader:
If spec constant has multiple uses, it's duplicated and decoration group is used to set the ID.
We need to materialize group decoration instructions to be able to specialize spec constant values.


bader:
Resolved by https://github.com/KhronosGroup/SPIRV-LLVM-Translator/pull/727.
Closing.

PR #677, opened 2020-08-14
Float controls decs

DmitryBushev:
Adding new decorations from spec: https://github.com/intel/llvm/pull/1611/

PR #699, opened 2020-08-24
Fix release build fail

DmitryBushev:
Fixed bug leaded to release build fail occured by
Float Control decs commit: https://github.com/KhronosGroup/SPIRV-LLVM-Translator/pull/677

### DIFF
--- a/lib/SPIRV/SPIRVReader.cpp
+++ b/lib/SPIRV/SPIRVReader.cpp
@@ -3951,7 +3951,7 @@ bool SPIRVToLLVM::transVectorComputeMetadata(SPIRVFunction *BF) {
           static_cast<SPIRVDecorateFunctionFloatingPointModeINTEL const *>(
               FloatModes.at(0));
       auto FloatingMode = DecFlt->getOperationMode();
-#ifdef NDEBUG
+#ifndef NDEBUG
       for (auto *DecPreCast : FloatModes) {
         auto *Dec =
             static_cast<SPIRVDecorateFunctionFloatingPointModeINTEL const *>(


AlexeySachkov (approved):
I wonder why we haven't spotted the original build fail in CI.

@svenvh, @AlexeySotkin, looks like we don't have a configuration without `NDEBUG` set (should be equal to LLVM release without assertions)

svenvh:
> I wonder why we haven't spotted the original build fail in CI.
> 
> @svenvh, @AlexeySotkin, looks like we don't have a configuration without `NDEBUG` set (should be equal to LLVM release without assertions)

That might be worth adding indeed, as it would align better with LLVM configurations.

Perhaps it's a good opportunity to revise the configurations being tested, because right now there are quite a few configurations.  I am not sure there is value in testing all combinations for the majority of the pull requests as it consumes quite a bit of Travis time, which is particularly noticeable when there are many pull requests in flight.

AlexeySachkov:
> I am not sure there is value in testing all combinations for the majority of the pull requests as it consumes quite a bit of Travis time, which is particularly noticeable when there are many pull requests in flight.

Totally agree here.

BTW, we might want to try GitHub Actions - one thing I love about it is that you can have several config files each of them describing different workflows (and each of them could be triggered by different actions).

For example, having separate workflow for launching clang-format could save us some time there as we don't need to install all those packages to get clang-format and it allows us, for example, attach a patch-file which need to be applied to fix formatting. See https://github.com/intel/llvm/actions/runs/222983731 as an example

svenvh:
> BTW, we might want to try GitHub Actions - one thing I love about it is that you can have several config files each of them describing different workflows (and each of them could be triggered by different actions).

I am not familiar with GitHub Actions, but what you describe sounds useful indeed.  I wonder if it would also allow testing based on the patch?  For example if a patch touches CMakeLists.txt then it would run more configurations compared with a patch touching only .cpp files.

Sorry for sidetracking in this PR btw, perhaps we should continue the discussion in a new issue about CI changes?

AlexeySachkov:
> Sorry for sidetracking in this PR btw, perhaps we should continue the discussion in a new issue about CI changes?

@svenvh, I've submitted #716 to discuss CI changes

PR #703, opened 2020-08-24
Fix LoopControl parameters order

MrSidims:
3.23 Loop Control:
If there are multiple following operands indicated, they are ordered: Those
indicated by smaller-numbered bits appear first.

This rule was violated in previous implementation.

### DIFF
--- a/lib/SPIRV/SPIRVReader.cpp
+++ b/lib/SPIRV/SPIRVReader.cpp
@@ -864,6 +864,7 @@ void SPIRVToLLVM::setLLVMLoopMetadata(const LoopInstType *LM,
         Parameters.push_back(SafelenMDOp);
       Metadata.push_back(llvm::MDNode::get(*Context, Parameters));
     }
+    ++NumParam;
   }
   if (LC & LoopControlPipelineEnableINTELMask) {
     Metadata.push_back(llvm::MDNode::get(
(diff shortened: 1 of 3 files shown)


MrSidims:
Tagging @mlychkov @vmaksimo @AGindinson @AlexeySachkov

lib/SPIRV/SPIRVReader.cpp
```diff
@@ -864,6 +864,7 @@ void SPIRVToLLVM::setLLVMLoopMetadata(const LoopInstType *LM,
         Parameters.push_back(SafelenMDOp);
       Metadata.push_back(llvm::MDNode::get(*Context, Parameters));
     }
+    ++NumParam;
```
AGindinson, line 867: Why is an additional increment needed, given the ones at lines 795 & 797?

MrSidims: Because on lines 795 and 797 NumParam is used immediately for IVDep translation. When it's finished we need to increase it one more time to shift index of the parsed parameters for the next LoopControl mask to handle.

AGindinson: Right, so the bug would show up if IVDep parameters were listed prior to some other attribute's parameters? I'm not implying it need not be fixed, but just trying to determine if this condition could be true _today_, when IVDep's mask bit is the greatest and its parameters come last.

AGindinson: UPD: I've blissfully missed the new attributes that occupy greater bits and thus can be affected. Thanks for the explanation!

AGindinson:
@MrSidims, formatting issues remain, I suppose.

AlexeySachkov:
@MrSidims, please fix issues reported by clang-tidy

PR #712, opened 2020-08-24
[NFC] Avoid verifying module in release builds

AlexeySachkov:
This is achieved by guarding most (only covered LLVM IR -> SPIR-V passes)
`verifyModule` calls by corresponding command line options, which are
disabled/enabled in accordance with _SPIRVDBG define.

Made _SPIRVDBG define disabled in release builds by default

### DIFF
--- a/lib/SPIRV/libSPIRV/SPIRVDebug.cpp
+++ b/lib/SPIRV/libSPIRV/SPIRVDebug.cpp
@@ -39,6 +39,11 @@
 
 #include "SPIRVDebug.h"
 
+#include "llvm/IR/Verifier.h"
+#include "llvm/Support/raw_ostream.h"
+
+#define DEBUG_TYPE "spirv-regularization"
+
 using namespace SPIRV;
 
 bool SPIRV::SPIRVDbgEnable = false;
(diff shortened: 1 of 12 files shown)


AlexeySotkin (commented):
What about having _one_ command line option which enables or disables validation/verification of LLVM module after _every_ pass?

lib/SPIRV/libSPIRV/SPIRVDebug.h
```diff
@@ -72,7 +84,43 @@ inline spv_ostream &spvdbgs() {
...
+
+// Output stream for SPIRV debug information.
+inline dev_null_stream &spvdbgs() {
+  static dev_null_stream Out;
+  return Out;
+}
+
 #endif
 
+#define VERIFY_REGULARIZATION_PASS(Module, PassName)                           \
```
svenvh, line 112: Any reason why this has to be a macro?

lib/SPIRV/libSPIRV/SPIRVDebug.cpp
```diff
@@ -45,3 +45,8 @@ bool SPIRV::SPIRVDbgEnable = false;
 SPIRV::SPIRVDbgErrorHandlingKinds SPIRV::SPIRVDbgError =
     SPIRVDbgErrorHandlingKinds::Exit;
 bool SPIRV::SPIRVDbgErrorMsgIncludesSourceInfo = true;
+
+llvm::cl::opt<bool> SPIRV::VerifyRegularizationPasses(
+    "spirv-verify-regularize-passes", llvm::cl::init(_SPIRVDBG),
+    llvm::cl::desc(
+        "Validate module after each pass in LLVM regularization phase"));
```
svenvh, line 52:
```suggestion
        "Verify module after each pass in LLVM regularization phase"));
```

lib/SPIRV/libSPIRV/SPIRVDebug.h
AlexeySachkov: > Any reason why this has to be a macro?

No particular reason: I agree that macro is completely unnecessary, will refactor into a helper function instead

AlexeySachkov: Replaced macro with a helper function in 33c039b949569438afeba506ef8dfa7c56ed2a17

lib/SPIRV/libSPIRV/SPIRVDebug.cpp
AlexeySachkov: Applied in 33c039b949569438afeba506ef8dfa7c56ed2a17

svenvh (approved):

PR #719, opened 2020-08-27
Support 1024-bit constants

AlexeySachkov:
Adjusted SPRIVConstantBase class to hold up to 8 SPIR-V words in a
constant.

Motivation for this change is that optimized LLVM IR can represent
types like `[3 x i32]` as `i96` and in combination with
SPV_INTEL_arbitrary_precision_integers extension it generates pretty
valid SPIR-V. However, during its consumption the translator reads more
than 2 SPIR-V words from an input and stores them in array for only 2
SPIR-V words, which causes memory corruption.

Discovered by debugging an issue with SYCL compiler from intel/llvm
project where internal representation of `cl::sycl::vec<int, 8>` class
were transformed into `i256` and stored in a constant.

### DIFF
--- a/lib/SPIRV/libSPIRV/SPIRVType.h
+++ b/lib/SPIRV/libSPIRV/SPIRVType.h
@@ -183,7 +183,12 @@ protected:
   _SPIRV_DEF_ENCDEC3(Id, BitWidth, IsSigned)
   void validate() const override {
     SPIRVEntry::validate();
-    assert(BitWidth > 1 && BitWidth <= 64 && "Invalid bit width");
+    assert(BitWidth > 1 &&
+           (BitWidth <= 64 ||
+            (Module->isAllowedToUseExtension(
+                 ExtensionID::SPV_INTEL_arbitrary_precision_integers) &&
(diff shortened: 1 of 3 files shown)


AlexeySotkin (commented):
Why now we limit it by 256, why not 512 or 1024?

AlexeySachkov:
> Why now we limit it by 256, why not 512 or 1024?

This is a good question. I've chosen 256 because it is the highest number I've seen so far.

AlexeySotkin:
> > Why now we limit it by 256, why not 512 or 1024?
> 
> This is a good question. I've chosen 256 because it is the highest number I've seen so far.

I suggest to use 1024. Just in case we will need extra capacity next time :)

AlexeySachkov:
> > > Why now we limit it by 256, why not 512 or 1024?
> > 
> > 
> > This is a good question. I've chosen 256 because it is the highest number I've seen so far.
> 
> I suggest to use 1024. Just in case we will need extra capacity next time :)

Done

AlexeySotkin (approved):
Please update the commit message. Other than that LGTM!

PR #721, opened 2020-08-31
Lower UIToFPInst with bool operand

Fznamznon:
UIToFPInst LLVM instruction is translated to OpConvertUToF SPIRV
instruction. OpConvertUToF instruction cannot receive boolean operand,
so replace boolean operand with with select instruction.

### DIFF
--- a/lib/SPIRV/SPIRVLowerBool.cpp
+++ b/lib/SPIRV/SPIRVLowerBool.cpp
@@ -107,7 +107,7 @@ public:
   virtual void visitUIToFPInst(UIToFPInst &I) {
     auto Op = I.getOperand(0);
     if (isBoolType(Op->getType())) {
-      auto Ty = Type::getInt16Ty(*Context);
+      auto Ty = Type::getInt32Ty(*Context);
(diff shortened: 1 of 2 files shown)


lib/SPIRV/SPIRVLowerBool.cpp
```diff
@@ -104,6 +104,17 @@ class SPIRVLowerBool : public ModulePass, public InstVisitor<SPIRVLowerBool> {
       replace(&I, Sel);
     }
   }
+  virtual void visitUIToFPInst(UIToFPInst &I) {
+    auto Op = I.getOperand(0);
+    if (isBoolType(Op->getType())) {
+      auto Ty = Type::getInt16Ty(*Context);
```
bader, line 110: Let's use i32. i16 requires extra capability.

bader (approved):

lib/SPIRV/SPIRVLowerBool.cpp
Fznamznon: Done.

PR #723, opened 2020-09-01
[SYCL][FPGA] Replace the IVDep index group validity assertion

AGindinson:
The assertion now gives way to a simple check until all known
LLVM IR edge cases that do not accord to the condition can be
handled.


AGindinson:
Tagging @MrSidims @AlexeySotkin

MrSidims (approved):
What is so special about SYCL and FPGA from the translator's point of view? :) Please remove it from the title.
Otherwise LGTM. Not sure if the test is required here.

AlexeySachkov (approved):

PR #724, opened 2020-09-01
Implement OpTypeBufferSurfaceINTEL

NikitaRudenkoIntel:
Implement OpTypeBufferSurfaceINTEL according to the spec:
https://github.com/intel/llvm/pull/1612


lib/SPIRV/libSPIRV/SPIRVType.h
```diff
@@ -829,6 +829,38 @@ bool isType(const T1 *Ty, unsigned Bits = 0) {
...
+
+protected:
+  _SPIRV_DEF_ENCDEC2(Id, AccessKind)
+  void validate() const override {
+    assert(OpCode == OC);
+    assert(WordCount == FixedWC);
+  }
+
+private:
+  SPIRVAccessQualifierKind AccessKind;
```
AlexeySotkin, line 861: There is no `AccessKind` operand in the spec.

NikitaRudenkoIntel: Yes, but the spec will be updated with this operand.

NikitaRudenkoIntel: Spec was updated

AlexeySotkin (approved):

PR #725, opened 2020-09-01
[NFC] Replace some std containers with llvm ones

AlexeySachkov:
This change aims to improve performance of the translator.

Tested this patch on the following example: 
[1024_512_256_source.zip](https://github.com/KhronosGroup/SPIRV-LLVM-Translator/files/5157501/1024_512_256_source.zip) contains 20 MB of auto-generated LLVM bitcode from SYCL application. I wrapped content of `llvm::writeSpirv` function into a loop with 50 iterations and this patch speeds up LLVM -> SPIR-V translation by ~30%

Using `std::unordered_map` & `std_unordered_set` for these two types gives only ~15% improvement on the test case mentioned above

### DIFF
--- a/lib/SPIRV/libSPIRV/SPIRVModule.cpp
+++ b/lib/SPIRV/libSPIRV/SPIRVModule.cpp
@@ -48,6 +48,9 @@
 #include "SPIRVType.h"
 #include "SPIRVValue.h"
 
+#include <llvm/ADT/DenseMap.h>
+#include <llvm/ADT/DenseSet.h>
+
 #include <set>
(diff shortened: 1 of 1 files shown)


keryell:
Interesting data points!

AlexeySotkin:
@AlexeySachkov, there are failing tests in travis.ci

AlexeySachkov:
> @AlexeySachkov, there are failing tests in travis.ci

Yes, new containers have different ordering, so, some tests should be updated to use more `-DAG` FileCheck directives, I'm working on that

PR #727, opened 2020-09-01
Do not emit OpDecorationGroup and OpGroupDecorate

Fznamznon:
They are deprecated in the later version of the spec. Consumption of
them is saved for backward compatibility.


test/group-decorate.ll
```diff
@@ -1,39 +0,0 @@
-; RUN: llvm-as %s -o %t.bc
```
AlexeySotkin, line 1: I think it make sense to have a test to check that translator can consume SPIR-V module with `OpGroupDecorate` and `OpDecorationGroup`. We could transform this test to `group-decorate.spt` (SPIR-V text representation) and test the consumption functionality with it.

Fznamznon: There are a lot of other tests in spt form which use `OpGroupDecorate` and `OpDecorationGroup`, but ok, I can add a specific one.

Fznamznon: Added in https://github.com/KhronosGroup/SPIRV-LLVM-Translator/pull/727/commits/db355467d533f3f5145e93baf94914fb3be7e344

PR #730, opened 2020-09-03
Do not return error on empty LLVM module

AlexeySotkin:
Return empty SPIR-V module instead.

### DIFF
--- a/test/empty_file.ll
+++ b/test/empty_file.ll
@@ -1,0 +1,4 @@
+; RUN: llvm-as < %s -o %t.bc
+; RUN: llvm-spirv %t.bc -o %t.spv
+; RUN: FileCheck --allow-empty --input-file=%t.spv %s
+; CHECK: {{$}}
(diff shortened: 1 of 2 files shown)


test/empty_file.ll
```diff
@@ -0,0 +1,4 @@
+; RUN: llvm-as < %s -o %t.bc
+; RUN: llvm-spirv %t.bc -o %t.spv
+; RUN: FileCheck --allow-empty --input-file=%t.spv %s
```
AlexeySachkov, line 3: This produces an invalid SPIR-V. From [universal validation rules](https://www.khronos.org/registry/spir-v/specs/unified1/SPIRV.html#_a_id_validation_a_validation_rules):

> There is at least one OpEntryPoint instruction, unless the Linkage capability is being used.

Also it breaks physical layout (missed magic number, version, generator's magic number, etc.)

I think that it worth at least printing a warning for an empty file

StuartDBrady: That's a good point. What's the best thing to do in this case, then? If empty SPIR-V modules are invalid without the Linkage capability, isn't an error the correct thing to return?

If we do need to allow accept empty IR modules, then it would make more sense to me to produce a valid empty SPIR-V file with the Linkage capability.

AlexeySotkin: Ok, I tend to agree with you. I'll rework this PR.

AlexeySachkov:
Submitted as #748

PR #736, opened 2020-09-07
Remove spirv-tools from travis on llvm_release_80 branch

NikitaRudenkoIntel:


AlexeySachkov:
I'm not a huge fan of this change, because we have added `spirv-tools` package into our CI on purpose, in order to increase testing coverage for the transaltor.

It seems that `https://packages.lunarg.com/vulkan xenial main` doesn't contain `spirv-tools` for `xenial` anymore and from my experiments in #738 it [looks like](https://travis-ci.org/github/KhronosGroup/SPIRV-LLVM-Translator/jobs/725575279) `llvm-8-tools` is built incorrectly and depends on non-existing package `python2`

Having that said and the fact that 8.0 branch is pretty old (i.e. I hope that at some point we will stop cherry-picking patches to it), I will probably go ahead and merge this PR if there are no strong objections or other suggestions about how to fix the CI in there

AlexeySachkov (approved):

PR #738, opened 2020-09-08
Switch Travis CI to Ubuntu 18.04 (Bionic)

AlexeySachkov:
packages.lunarg.com doesn't support 16.04 Xenial anymore which means
that in order to use spirv-tools package we have to move to newer Ubuntu
release.

Also updated URLs to use `https` for downloading apt packages and keys

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,8 +3,8 @@ language: cpp
 os:
   - linux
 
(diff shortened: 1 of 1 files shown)


AlexeySachkov:
It seems to me that `llvm-8-tools` has dependency on non-existing `python2` package, which causes failures in CI

PR #743, opened 2020-09-11
Implement lowering of llvm.umul.with.overflow.* intrinsic

vmaksimo:
Since the intrinsic can't be mapped to SPIR-V directly, its semantics has been implemented as a function in LLVM IR. This patch introduces a pass which links the implementation function into the IR being translated to SPIR-V.


lib/SPIRV/SPIRVLowerUmulWithOverflow.cpp
```diff
@@ -0,0 +1,158 @@
...
+using namespace llvm;
+using namespace SPIRV;
+
+namespace SPIRV {
+cl::opt<bool> SPIRVLowerUmulWithOverflowValidate(
+    "spv-lower-umulwithoverflow-validate",
+    cl::desc("Validate module after lowering llvm.umul.with.overflow.*"
+             "intrinsics"));
+
+class SPIRVLowerUmulWithOverflow
```
AlexeySachkov, line 64: To me, this seems like an overkill to invent the whole new pass to lower one more intrinsic. I would like to hear feedback from @AlexeySotkin  as well, but personally I would suggest to generalize existing `SPIRVLowerSaddWithOverflow` pass so it can lower more than just a single intrinsic

AlexeySotkin:
I think, _maybe_ we can translate this intrinsic to [OpUMulExtended](https://www.khronos.org/registry/spir-v/specs/unified1/SPIRV.html#OpUMulExtended) instruction? These two have very similar semantic.

If we decide to follow this approach. I'm concerned about reverse translation of OpUMulExtended. 

- Should we follow recommendation for LLVM [mul](https://llvm.org/docs/LangRef.html#mul-instruction) instruction?
  >  If a full product (e.g. i32 * i32 -> i64) is needed, the operands should be ~~sign-extended or~~ zero-extended as appropriate to the width of the full product.

   In this case we (1) get other type of the result, (2) overflow goes unnoticed, (3) output has more LLVM intstructions.

- Or should we translate it to `llvm.llvm.umul.with.overflow.*`?
   In this case we loose the high-order bits of the multiplication, which should be provided as the result (per the SPIR-V spec).

lib/SPIRV/SPIRVLowerUmulWithOverflow.cpp
AlexeySotkin: I agree with @AlexeySachkov. If we want to add handling for another instrinsic, we better to generalize `SPIRVLowerSaddWithOverflow` pass,

test/llvm-intrinsics/llvm.umul.with.overflow.ll
```diff
@@ -0,0 +1,91 @@
...
+; CHECK-SPIRV: INotEqual [[#]] [[CMP_RES:[0-9]+]] [[VAR_A]] [[DIV_RES]]
+; CHECK-SPIRV: Select [[#]] [[SELECT_RES:[0-9]+]] [[CMP_RES]]
+; CHECK-SPIRV: CompositeInsert [[#]] [[INSERT_RES:[0-9]+]] [[MUL_RES]]
+; CHECK-SPIRV: CompositeInsert [[#]] [[INSERT_RES_1:[0-9]+]] [[SELECT_RES]] [[INSERT_RES]]
+; CHECK-SPIRV: ReturnValue [[INSERT_RES_1]]
+
+; Function Attrs: nofree nounwind writeonly
+define dso_local spir_func void @_Z5foo32jjPj(i32 %a, i32 %b, i32* nocapture %c) local_unnamed_addr #0 {
+entry:
+  ; CHECK-LLVM: call [[UMUL_32_TY]] @llvm.umul.with.overflow.i32
```
AlexeySachkov, line 49: Please add check that there is a call of replacement function in SPIR-V module. The same comment applies for `@umulo_v2i64`

test/llvm-intrinsics/llvm.umul.with.overflow.ll
```diff
@@ -0,0 +1,91 @@
...
+target datalayout = "e-p:32:32-i64:64-v16:16-v24:32-v32:32-v48:64-v96:128-v192:256-v256:256-v512:512-v1024:1024"
+target triple = "spir"
+
+; CHECK-LLVM: [[UMUL_8_TY:%structtype]] = type { i8, i1 }
+; CHECK-LLVM: [[UMUL_32_TY:%structtype.[0-9]+]] = type { i32, i1 }
+; CHECK-LLVM: [[UMUL_VEC64_TY:%structtype.[0-9]+]] = type { <2 x i64>, <2 x i1> }
+
+; On LLVM level, we'll check that the intrinsics were generated again in reverse translation,
+; replacing the SPIR-V level implementations.
+; CHECK-LLVM-NOT: declare {{.*}} @spirv.llvm_umul_with_overflow_{{.*}}
```
AlexeySachkov, line 20: Probably it is better to move this into `--implicit-check-not`, because scope of `CHEKC-NOT` is actually bound by surrounding `CHECK` directives so it won't check the whole file

AlexeySachkov (commented):
I do like this approach better than the previous one.

I haven't yet reviewed the logic of the generated replacement function

test/llvm-intrinsics/llvm.umul.with.overflow.ll
vmaksimo: If I understood you correctly, these checks are already present on [lines 7-9](https://github.com/KhronosGroup/SPIRV-LLVM-Translator/pull/743/files#diff-53b5d620593beaa14591098d7df1b5f8fba7dbdb0a68a382c0d11c0cae94ef1aR7-R9)

AlexeySachkov: >If I understood you correctly, these checks are already present on lines 7-9

No, I mean that you have a check that `call` is generated in LLVM IR, but you don't have a check that `FunctionCall` is generated in SPIR-V - just like it is checked in `_Z4foo8hhPh` for `llvm.umul.with.overflow.i8`

vmaksimo: Oh, my bad, I missed these cases. Uploaded new patch with a fix

lib/SPIRV/SPIRVRegularizeLLVM.cpp
```diff
@@ -207,6 +210,51 @@ void SPIRVRegularizeLLVM::lowerFunnelShiftLeft(IntrinsicInst *FSHLIntrinsic) {
...
+  // Do unsigned multiplication Mul = A * B.
+  // Then check if unsigned division Div = Mul / A is not equal to B.
+  // If so, then overflow has happened.
+  auto *Mul = Builder.CreateNUWMul(FirstArg, SecondArg);
+  auto *Div = Builder.CreateUDiv(Mul, FirstArg);
+  auto *Cmp = Builder.CreateICmpNE(FirstArg, Div);
+
+  Type *BoolTy = UMulFunc->getReturnType()->getContainedType(1);
+  auto *TrueVal = Constant::getAllOnesValue(BoolTy);
+  auto *FalseVal = Constant::getNullValue(BoolTy);
```
AlexeySachkov, line 233: Probably `ConstantInt::getTrue` and `ConstantInt::getFalse` would be better from readability point here

lib/SPIRV/SPIRVRegularizeLLVM.cpp
```diff
@@ -207,6 +210,51 @@ void SPIRVRegularizeLLVM::lowerFunnelShiftLeft(IntrinsicInst *FSHLIntrinsic) {
...
+
+  auto *Overflow = Builder.CreateSelect(Cmp, TrueVal, FalseVal);
```
AlexeySachkov, line 235: Do we really need this select? why can't we just use `Cmp` as overflow flag?

lib/SPIRV/SPIRVRegularizeLLVM.cpp
```diff
@@ -207,6 +210,51 @@ void SPIRVRegularizeLLVM::lowerFunnelShiftLeft(IntrinsicInst *FSHLIntrinsic) {
...
+
+  // umul.with.overflow intrinsic return a structure, where the first element
+  // is the multiplication result, and the second is an overflow bit.
+  auto *StructTy = StructType::create({FirstArg->getType(), BoolTy});
```
AlexeySachkov, line 239:
```suggestion
  auto *StructTy = UMulFunc->getReturnType();
```

lib/SPIRV/SPIRVRegularizeLLVM.cpp
vmaksimo: Nice catch, thanks! I'll remove it

AlexeySachkov (approved):

Issue #206, opened 2019-05-30
Translation of BIFs from different frontends


AnastasiaStulova:
Currently Translator uses IR metadata to detect the original source of translation module. Do we need any different solution?

AlexeySachkov:
As @Naghasan pointed [here](https://github.com/KhronosGroup/SPIRV-LLVM-Translator/pull/74#issuecomment-444155988) we cannot rely on original source of translation module.

May be it is better to work on SPIR-V friendly LLVM IR to make it universal enough to handle OpenCL C, OpenCL C++, SYCL and different shader languages?

AnastasiaStulova:
> As @Naghasan pointed here we cannot rely on original source of translation module.

I actually mean the translation from LLVM IR to SPIR-V.

> May be it is better to work on SPIR-V friendly LLVM IR to make it universal enough to handle OpenCL C, OpenCL C++, SYCL and different shader languages?

I think we need to see how it aligns with Clang CodeGen phase. Do you have concrete ideas how it could work?

Btw Clang IL was discussed on cfe-dev recently http://lists.llvm.org/pipermail/cfe-dev/2019-February/061450.html. I found it quite interesting and I was wondering if it can work for SPIR-V generation.

Naghasan:
> Btw Clang IL was discussed on cfe-dev recently http://lists.llvm.org/pipermail/cfe-dev/2019-February/061450.html. I found it quite interesting and I was wondering if it can work for SPIR-V generation.

I believe generating SPIR-V from Clang is what the DX Shader compiler is doing 
https://github.com/microsoft/DirectXShaderCompiler/tree/master/tools/clang/lib/SPIRV

> I think we need to see how it aligns with Clang CodeGen phase.

I'm not sure if it is that interesting to be clang based rather than llvm based:
* this force users to be clang based, so defacto exclude any approache that emit directly LLVM without building a clang AST
* will the clang community accept a different target for clang ?

AnastasiaStulova:
**Input needed:** understanding of current issues in the translator for translation of BIFs from various frontends.

AlexeySachkov:
> I actually mean the translation from LLVM IR to SPIR-V.

I agree that PR I linked might not be very relevant, but anyway we have several places checking LLVM IR metadata to understand source language and pre-process LLVM IR:

[...]

PR #746, opened 2020-09-15
Fix emission of SPIR-V friendly IR for OpenCL EIS

AlexeySachkov:
We already have `--spirv-target-env` option to switch between different representations of SPIR-V in LLVM IR, but it doesn't cover all functionality. This PR adds possibility to lower instructions from [OpenCL extended instruction set](https://www.khronos.org/registry/spir-v/specs/unified1/OpenCL.ExtendedInstructionSet.100.html) into SPIR-V friendly IR representation.

This is only a single step on our way to be able to lower all SPIR-V instructions into SPIR-V friendly IR form and I believe that more unification and refactoring could be done: I plan to continue working on this topic and probably the more places I will fix, the more refactoring I will perform. So, even this code might be changed and affected by further work in this direction, but I cannot say anything about this for sure yet.

Anyway, please let me know what do you think.


AnastasiaStulova:
Sorry can I get more background please - as far as I can see you are trying to produce IR that is a bit more generic than what Clang would typically generate for OpenCL? And this issue has been registered from SYCL. Is there any place where the original problem is described?

As far as I can see this generally seems to align with the conclusions we made on https://github.com/KhronosGroup/SPIRV-LLVM-Translator/issues/206 i.e. we would keep maintaining "old style of IR" for OpenCL in clang for the time being but will also look for a new format that new frontends and languages can explore. 

I am just trying to understand whether there is any need to revisit this strategy...

AlexeySachkov:
@AnastasiaStulova, sure

> Sorry can I get more background please - as far as I can see you are trying to produce IR that is a bit more generic than what Clang would typically generate for OpenCL? 

That is correct. I have a SYCL program, which produces SPIR-V file that cannot be translated back into LLVM IR using KhronosGroup/SPIRV-LLVM-Translator.

Also, this issue seems to be described by @Naghasan in [this comment](https://github.com/KhronosGroup/SPIRV-LLVM-Translator/issues/206#issuecomment-595320879):

>Let assume I have lib.spv and my_kernel.cpp which both uses a builtin A (A being the name in the spec).
> By design, we are enforcing that this is never guarantee to work
> ```
> llvm-spirv -r lib.spv -o lib.bc 
> clang -emit-llvm my_kernel.cpp -o my_kernel.bc 
> llvm-link my_kernel.bc  lib.bc -o my_kernel.bc 
> ```
> builtin A now appears in 2 different forms in the same module.


> And this issue has been registered from SYCL. Is there any place where the original problem is described?

Not really, but let me describe it here. The following SYCL example:

```c++
#include <CL/sycl.hpp>
#include <iostream>

float isfinite(float x) { return cl::sycl::isfinite(x); }

using namespace sycl;

class KernelName;

auto exception_handler = [](sycl::exception_list exceptions) {
  for (std::exception_ptr const &e : exceptions) {
    try {
      std::rethrow_exception(e);
    } catch (sycl::exception const &e) {
      std::cout << "Caught asynchronous SYCL exception:\n"
                << e.what() << std::endl;
    }
  }
};

int main() {
  try {
    queue q(exception_handler);
    buffer<float> b(range<1>{1});
    q.submit([&](handler &h) {
       auto acc = b.get_access<access::mode::discard_write>(h);
       h.single_task<KernelName>([=]() { acc[0] = ::isfinite(10.0); });
     }).wait_and_throw();

    auto acc = b.get_access<access::mode::discard_write>();
    std::cout << acc[0] << std::endl;
  } catch (sycl::exception &e) {
    std::cout << e.what() << std::endl;
  }

  return 0;
}
```

It results in the following SPIR-V (attached: [spvasm.txt](https://github.com/KhronosGroup/SPIRV-LLVM-Translator/files/5291934/spvasm.txt)):

```
%call_0 = OpFunctionCall %float %_Z8isfinitef %float_10
; ...
%_Z8isfinitef = OpFunction %float None %184
; ...
%call1 = OpIsFinite %bool %329
```

So, when the translator lowers `%_Z8isfinitef = OpFunction %float None %184` it produces `float _Z8isfinitef` function in LLVM IR. But later, when the translation comes to `%call1 = OpIsFinite %bool %329` it is being lowered to `int _Z8isfinitef`, which causes redefinition error.

To summarize: 
With LLVM IR coming out of intel/llvm SYCL compiler we don't have a problem with distinguishing between built-ins and user-defined function, as intel/llvm FE/header files designed to emit SPIR-V friendly IR, i.e. some stable canonical LLVM-SPIRV representation.
However, during backward translation, we have to lower SPIR-V into the same canonical LLVM-SPIRV representation and not into something which will be produced by clang for OpenCL C, because of potential names clash.

This PR is kind of a technical thing: it fixes lowering of SPIR-V into SPIR-V friendly IR format. I think it can be considered as related to #206 in context of what is our stable canonical LLVM-SPIRV representation? At the moment, I'm just re-using [the existing one](https://github.com/KhronosGroup/SPIRV-LLVM-Translator/blob/master/docs/SPIRVRepresentationInLLVM.rst) and not inventing anything new

AnastasiaStulova:
Ok thank you for the explanation. That seems to be right. 

I guess one approach could be to reuse source language information to allow decoding the original mangled names? We touch this issue a bit in #206 (https://github.com/KhronosGroup/SPIRV-LLVM-Translator/issues/206#issuecomment-599727461)

> Also while translating from SPIR-V I guess we will need to know the original source language information too i.e. assuming I translated to SPIR-V from SYCL source then if I want to convert back to LLVM IR for linking/optimizations etc I would need to get the mangled names of builtin functions right in order to be able to link them correctly in... or would we have to switch all the builtin functions to intrinsics or maintain two different flows with SPIR-V intrinsics and w/o?

But we've highlighted that the source language information is not guaranteed to be preserved in SPIR-V so it is not possible to rely on it at present.

In general, this opens up the same discussion for which we didn't find any good solution yet, so your current approach makes sense to me.

Naghasan:
This seems rational to me and goes toward a better modularisation (get SPIR-V representation then OpenCL C or other).

> I guess one approach could be to reuse source language information to allow decoding the original mangled names? We touch this issue a bit in #206 ([#206 (comment)](https://github.com/KhronosGroup/SPIRV-LLVM-Translator/issues/206#issuecomment-599727461))
> 
> > Also while translating from SPIR-V I guess we will need to know the original source language information too i.e. assuming I translated to SPIR-V from SYCL source then if I want to convert back to LLVM IR for linking/optimizations etc I would need to get the mangled names of builtin functions right in order to be able to link them correctly in... or would we have to switch all the builtin functions to intrinsics or maintain two different flows with SPIR-V intrinsics and w/o?
> 
> But we've highlighted that the source language information is not guaranteed to be preserved in SPIR-V so it is not possible to rely on it at present.
> 

I think it should an option given by the user. Driver / build configuration (e.g. clang or CMake) will make things consistent across the whole build.

Using OpSource would make the translator behave differently depending on the value which goes against the spec. And in this case, a non-semantic op code would dictate if a module can be consumed or not.

AnastasiaStulova:
> Using OpSource would make the translator behave differently depending on the value which goes against the spec. And in this case, a non-semantic op code would dictate if a module can be consumed or not.

Can you elaborate on this, please? I never understood why this ended up this way in the spec. It seems something that could help with tooling. This issue came up multiple times in various places. I don't entirely understand the point of adding something that should not be used.

Naghasan:
> > Using OpSource would make the translator behave differently depending on the value which goes against the spec. And in this case, a non-semantic op code would dictate if a module can be consumed or not.
> 
> Can you elaborate on this, please? I never understood why this ended up this way in the spec. It seems something that could help with tooling. This issue came up multiple times in various places. I don't entirely understand the point of adding something that should not be used.

Answered here https://github.com/KhronosGroup/SPIRV-LLVM-Translator/issues/206#issuecomment-702229396

AlexeySachkov:
CI issues doesn't seem to be related  to the PR

AlexeySotkin (approved):
LGTM, Thanks.

svenvh (approved):
LGTM

PR #750, opened 2020-09-21
Implement SingleElementVectorINTEL decoration

NikitaRudenkoIntel:
According to the spec
https://github.com/intel/llvm/pull/1612

### DIFF
--- a/lib/SPIRV/SPIRVWriter.cpp
+++ b/lib/SPIRV/SPIRVWriter.cpp
@@ -621,6 +621,15 @@ void LLVMToSPIRV::transVectorComputeMetadata(Function *F) {
     BF->addDecorate(DecorationVectorComputeCallableFunctionINTEL);
   }
 
+  if (Attrs.hasAttribute(AttributeList::ReturnIndex,
+                         kVCMetadata::VCSingleElementVector)) {
+    auto *RT = BF->getType();
+    assert(RT->isTypeBool() || RT->isTypeFloat() || RT->isTypeInt() ||
+           RT->isTypePointer() &&
+               "This decoration is valid only for Scalar or Pointer types");
+    BF->addDecorate(DecorationSingleElementVectorINTEL);
+  }
+
   for (Function::arg_iterator I = F->arg_begin(), E = F->arg_end(); I != E;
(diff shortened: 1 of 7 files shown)


test/transcoding/SPV_INTEL_vector_compute/decoration_single_element_vector.ll
```diff
@@ -0,0 +1,39 @@
+; RUN: llvm-as %s -o %t.bc
+; RUN: llvm-spirv %t.bc -o %t.spv --spirv-ext=+SPV_INTEL_vector_compute --spirv-allow-unknown-intrinsics
+; RUN: llvm-spirv -r %t.spv -o %t.bc
+; RUN: llvm-dis %t.bc -o %t.ll
+; RUN: FileCheck %s --input-file %t.ll  -check-prefix=LLVM
```
AlexeySachkov, line 5: Please add checks for SPIR-V representation of the attribute as well

NikitaRudenkoIntel: Done

lib/SPIRV/SPIRVWriter.cpp
```diff
@@ -617,6 +617,10 @@ void LLVMToSPIRV::transVectorComputeMetadata(Function *F) {
     BF->addDecorate(DecorationSIMTCallINTEL, SIMTMode);
   }
 
+  if (Attrs.hasAttribute(AttributeList::ReturnIndex,
+                         kVCMetadata::VCSingleElementVector))
+    BF->addDecorate(DecorationSingleElementVectorINTEL);
```
AlexeySachkov, line 622: I've just realized that this implementation is not really conformant with the spec, as per extension spec this decoration should be applied to a type and not to function or its argument:

> Apply to a 'Scalar' type to indicate that the type represents a single-element vector type in the high-level language.

NikitaRudenkoIntel: Spec was updated, now the patch conforms with it

AlexeySachkov (approved):

PR #752, opened 2020-09-23
Fix translation of APInt constants

MrSidims:
Previously APInt constants were being stored into uint64_t value
with following encoding/decoding. Now they are being packed into
SPIRVWords array directly.

### DIFF
--- a/lib/SPIRV/SPIRVReader.cpp
+++ b/lib/SPIRV/SPIRVReader.cpp
@@ -1497,10 +1497,27 @@ Value *SPIRVToLLVM::transValueWithoutDecoration(SPIRVValue *BV, Function *F,
     }
     switch (BT->getOpCode()) {
     case OpTypeBool:
-    case OpTypeInt:
+    case OpTypeInt: {
+      const unsigned NumBits = BT->getBitWidth();
+      if (NumBits > 64) {
+        // Translate arbitrary precision integer constants
+        const unsigned RawDataNumWords = BConst->getNumWords();
+        const unsigned BigValNumWords = (RawDataNumWords + 1) / 2;
+        std::vector<uint64_t> BigValVec(BigValNumWords);
+        const SPIRVWord *RawData = BConst->getSPIRVWords();
+        // SPIRV words are integers of 32-bit width, meanwhile llvm::APInt
+        // is storing data using an array of 64-bit words. Here we pack SPIRV
+        // words into 64-bit integer array.
+        for (size_t I = 0; I != RawDataNumWords; ++I)
+          BigValVec[I / 2] =
+              (I % 2) ? BigValVec[I / 2] | ((uint64_t)RawData[I] << 32)
+                      : BigValVec[I / 2] | ((uint64_t)RawData[I]);
+        return mapValue(BV, ConstantInt::get(LT, APInt(NumBits, BigValVec)));
+      }
       return mapValue(
           BV, ConstantInt::get(LT, ConstValue,
                                static_cast<SPIRVTypeInt *>(BT)->isSigned()));
+    }
     case OpTypeFloat: {
       const llvm::fltSemantics *FS = nullptr;
       switch (BT->getFloatBitWidth()) {
--- a/lib/SPIRV/libSPIRV/SPIRVModule.h
+++ b/lib/SPIRV/libSPIRV/SPIRVModule.h
@@ -50,6 +50,10 @@
 #include <unordered_set>
 #include <vector>
 
+namespace llvm {
+class APInt;
+} // namespace llvm
+
 namespace SPIRV {
 
 template <Op> class SPIRVConstantBase;
@@ -254,6 +258,7 @@ public:
                                                    SPIRVFunction *F) = 0;
   virtual SPIRVValue *addConstant(SPIRVValue *) = 0;
   virtual SPIRVValue *addConstant(SPIRVType *, uint64_t) = 0;
+  virtual SPIRVValue *addConstant(SPIRVType *, llvm::APInt) = 0;
   virtual SPIRVValue *addSpecConstant(SPIRVType *, uint64_t) = 0;
   virtual SPIRVValue *addDoubleConstant(SPIRVTypeFloat *, double) = 0;
   virtual SPIRVValue *addFloatConstant(SPIRVTypeFloat *, float) = 0;
--- a/lib/SPIRV/libSPIRV/SPIRVValue.h
+++ b/lib/SPIRV/libSPIRV/SPIRVValue.h
@@ -47,6 +47,10 @@
 #include "SPIRVEntry.h"
 #include "SPIRVType.h"
 
+namespace llvm {
+class APInt;
+} // namespace llvm
+
 #include <iostream>
 #include <map>
 #include <memory>
@@ -148,6 +152,9 @@ public:
     recalculateWordCount();
     validate();
   }
+  // Incomplete constructor for AP integer constant
+  SPIRVConstantBase(SPIRVModule *M, SPIRVType *TheType, SPIRVId TheId,
+                    llvm::APInt &TheValue);
   // Complete constructor for float constant
   SPIRVConstantBase(SPIRVModule *M, SPIRVType *TheType, SPIRVId TheId,
                     float TheValue)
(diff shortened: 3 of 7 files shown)


MrSidims:
@AlexeySotkin @mlychkov @vmaksimo

test/capability-arbitrary-precision-integers.ll
```diff
@@ -50,13 +50,13 @@ entry:
   store i30 1, i30* %a.addr, align 4
 ; CHECK-LLVM: store i48 -4294901761, i48 addrspace(1)* @c
   store i48 -4294901761, i48 addrspace(1)* @c, align 8
-  store i96 1, i96 addrspace(1)* @d, align 8
-; CHECK-LLVM: store i96 1, i96 addrspace(1)* @d
+  store i96 18446744073709551620, i96 addrspace(1)* @d, align 8
```
mlychkov, line 53: Is there present check of how this value is stored in SPIR-V file?

MrSidims: Yes, line 22: ; CHECK-SPIRV-DAG: Constant [[#I96]] [[#]] 4 0 1

NOTE: 18446744073709551620 = 2^64 + 4

mlychkov: I got it, thank you.

mlychkov (approved):

lib/SPIRV/libSPIRV/SPIRVValue.h
```diff
@@ -148,6 +149,19 @@ template <spv::Op OC> class SPIRVConstantBase : public SPIRVValue {
...
+  // Complete constructor for AP integer constant
+  SPIRVConstantBase(SPIRVModule *M, SPIRVType *TheType, SPIRVId TheId,
+                    llvm::APInt TheValue)
+      : SPIRVValue(M, 0, OC, TheType, TheId) {
+    const uint64_t *BigValArr = TheValue.getRawData();
+    for (size_t I = 0; I != TheValue.getNumWords(); ++I) {
+      Union.Words[I * 2 + 1] =
+          (uint32_t)((BigValArr[I] & 0xFFFFFFFF00000000LL) >> 32);
+      Union.Words[I * 2] = (uint32_t)(BigValArr[I] & 0xFFFFFFFFLL);
+    }
```
AlexeySachkov, line 161: What about using `std::memcpy` here?

lib/SPIRV/libSPIRV/SPIRVModule.h
```diff
@@ -43,6 +43,8 @@
 #include "LLVMSPIRVOpts.h"
 #include "SPIRVEntry.h"
 
+#include "llvm/ADT/APInt.h"
```
AlexeySachkov, line 46: In spite of #755, can we just forward-declare `APInt` in this header and put actual include into corresponding `.cpp` file?

lib/SPIRV/SPIRVReader.cpp
```diff
@@ -1497,10 +1497,27 @@ Value *SPIRVToLLVM::transValueWithoutDecoration(SPIRVValue *BV, Function *F,
...
+        const unsigned BigValNumWords = (RawDataNumWords + 1) / 2;
+        std::vector<uint64_t> BigValVec(BigValNumWords);
+        const SPIRVWord *RawData = BConst->getSPIRVWords();
+        // SPIRV words are integers of 32-bit width, meanwhile llvm::APInt
+        // is storing data using an array of 64-bit words. Here we pack SPIRV
+        // words into 64-bit integer array.
+        for (size_t I = 0; I != RawDataNumWords; ++I)
+          BigValVec[I / 2] =
+              (I % 2) ? BigValVec[I / 2] | ((uint64_t)RawData[I] << 32)
+                      : BigValVec[I / 2] | ((uint64_t)RawData[I]);
```
AlexeySachkov, line 1514: Can we just use `std::memcpy`?

AlexeySachkov (commented):
Overall looks good to me.

I though that we would completely replace `Words` field of `UnionType` union with `APInt` to fully eliminate the size limit for constants, but we can do that later

lib/SPIRV/libSPIRV/SPIRVValue.h
AlexeySotkin: Are you sure we get correct SPIR-V representation for the constant literal? According to https://www.khronos.org/registry/spir-v/specs/unified1/SPIRV.html#OpConstant :

> Value is the bit pattern for the constant. Types 32 bits wide or smaller take one word. Larger types take multiple words, with low-order words appearing first.

For `i96` for example, does `BigValArr[2]` appear first? Am I missing something?

AlexeySachkov: Probably `I` should be incremented more, i.e. `I += 2`, as we fill two words within a single iteration. I still think that `std::memcpy` is better. Or, we could do `reinterpret_cast` to align types and then perform a simpler per-element copy using either 32- or 64-bit words to simplify the loop

MrSidims: It's possible to implement this via memcpy. When I schemed the change, on paper memcpy was looking more difficult because of the necessity to have the correct SPIRWord's order. I don't mind to try it in the code.

MrSidims:
> Overall looks good to me.
> 
> I though that we would completely replace `Words` field of `UnionType` union with `APInt` to fully eliminate the size limit for constants, but we can do that later

Problem here is that this requires to make a lot of reworks in encode/decode functions: APInt is mapped to just SPIRVInt, hence the default flow shall be also re-worked.

lib/SPIRV/libSPIRV/SPIRVValue.h
AlexeySachkov: > When I schemed the change, on paper memcpy was looking more difficult because of the necessity to have the correct SPIRWord's order.

I've just looked into the spec and indeed lower-order words should appear first. Unfortunately, I can't find any docs about how data is stored within `APInt`. Judging by the code from this PR and from llvm.org it seems that the ordering is the same, but because of the difference in word size we need to swap two halfs of 64-bit `APInt` word before putting it into SPIR-V - right?

If that's the case then I indeed suggest to leave everything as-is

MrSidims: Right.

AlexeySotkin (approved):

lib/SPIRV/libSPIRV/SPIRVValue.h
```diff
@@ -66,7 +68,6 @@ class SPIRVValue : public SPIRVEntry {
              SPIRVType *TheType)
       : SPIRVEntry(M, TheWordCount, TheOpCode), Type(TheType) {
     setHasNoId();
-    SPIRVValue::validate();
```
AlexeySachkov, line 69: Why this line was removed?

lib/SPIRV/libSPIRV/SPIRVValue.h
```diff
@@ -47,6 +47,8 @@
 #include "SPIRVEntry.h"
 #include "SPIRVType.h"
 
+#include "llvm/ADT/APInt.h"
```
AlexeySachkov, line 50: It is good that we don't have this include in `SPIRVModule.h` anymore, but `SPIRVValue.h` is a pretty "popular" header as well - would be good to avoid including `APInt` here as well: should be doable by outlining new `SPIRVConstantBase` constructor into corresponding `.cpp` file

AlexeySachkov (commented):
I don't have strong objections, a few questions/comments

lib/SPIRV/libSPIRV/SPIRVValue.h
MrSidims: I guess, because of incidental double "d".

MrSidims: Fixed

lib/SPIRV/libSPIRV/SPIRVValue.h
MrSidims: Removed. If you don't mind I went by taking an APInt reference.

AlexeySachkov: No objections, thanks!

AlexeySachkov (approved):

lib/SPIRV/libSPIRV/SPIRVValue.h
MrSidims: Looks like I made it wrong. Interesting that it compiled successfully locally.

PR #753, opened 2020-09-23
updates for OpenCL C 3.0

bashbaug:
This PR has a few updates to the SPIR-V LLVM Translator to prepare for OpenCL C 3.0 changes in Clang.

Primarily, OpenCL C 3.0 is a derivative of OpenCL C 2.0 and not OpenCL C++.  This means:

* We should run the OpenCL C++ passes for OpenCL 2.1 only, and not for versions newer than OpenCL 2.0.
* We should run similar OpenCL C 2.0 passes for OpenCL C 3.0.

At some point we could consider renaming some of the passes or the "CL21" version, but I would prefer to do that as a separate PR.

### DIFF
--- a/lib/SPIRV/PreprocessMetadata.cpp
+++ b/lib/SPIRV/PreprocessMetadata.cpp
@@ -207,8 +207,8 @@ void PreprocessMetadata::preprocessOCLMetadata(Module *M, SPIRVMDBuilder *B,
   // !{x} = !{i32 3, i32 102000}
   B->addNamedMD(kSPIRVMD::Source)
       .addOp()
-      .add(CLVer < kOCLVer::CL21 ? spv::SourceLanguageOpenCL_C
-                                 : spv::SourceLanguageOpenCL_CPP)
+      .add(CLVer == kOCLVer::CL21 ? spv::SourceLanguageOpenCL_CPP
+                                  : spv::SourceLanguageOpenCL_C)
       .add(CLVer)
       .done();
   if (EraseOCLMD)
--- a/lib/SPIRV/OCL20ToSPIRV.cpp
+++ b/lib/SPIRV/OCL20ToSPIRV.cpp
@@ -329,7 +329,7 @@ bool OCL20ToSPIRV::runOnModule(Module &Module) {
     return false;
 
   CLVer = std::get<1>(Src);
-  if (CLVer > kOCLVer::CL20)
+  if (CLVer == kOCLVer::CL21)
     return false;
 
   LLVM_DEBUG(dbgs() << "Enter OCL20ToSPIRV:\n");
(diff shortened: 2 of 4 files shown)


AlexeySotkin (approved):

lib/SPIRV/PreprocessMetadata.cpp
```diff
@@ -207,8 +207,8 @@ void PreprocessMetadata::preprocessOCLMetadata(Module *M, SPIRVMDBuilder *B,
   // !{x} = !{i32 3, i32 102000}
   B->addNamedMD(kSPIRVMD::Source)
       .addOp()
-      .add(CLVer < kOCLVer::CL21 ? spv::SourceLanguageOpenCL_C
-                                 : spv::SourceLanguageOpenCL_CPP)
+      .add(CLVer == kOCLVer::CL21 ? spv::SourceLanguageOpenCL_CPP
```
AnastasiaStulova, line 210: I am very confused about this statement, why CL version 2.1 corresponds to `SourceLanguageOpenCL_CPP`?

I think I am generally confused because the Translator seems to have more versions than what clang produces right now i.e. clang doesn't generate version 2.1 or 2.2.

In addition I don't know why we keep maintaining OpenCL C++. I think it only exists in a prototype which is on old version of LLVM and probably won't be compatible with IR on the newer releases. Would it be better if we just drop it completely?

svenvh: I agree with @AnastasiaStulova, I think we can make this simpler by just saying:

```suggestion
      .add(spv::SourceLanguageOpenCL_C)
```

and ignoring `spv::SourceLanguageOpenCL_CPP`.

bashbaug: I'd be fine with this.  I wasn't sure who all is using "CL21"... especially since OpenCL C 2.1 doesn't exist!

Could we make this change as a follow-on PR?  IMHO we should remove the OCL21ToSPIRV passes also, so this would be part of the larger cleanup I described in the initial PR description.

svenvh: > Could we make this change as a follow-on PR?

That sounds reasonable to me, as your commit doesn't change this aspect.

svenvh (approved):

lib/SPIRV/PreprocessMetadata.cpp
AnastasiaStulova: I agree, it is better to have clean up separately.

AnastasiaStulova (approved):

lib/SPIRV/PreprocessMetadata.cpp
AnastasiaStulova: FYI it was discussed on the call Oct 1st that `OCL21ToSPIRV` and `CL21`/`CL22` versions can likely be removed.

It seems however that `SourceLanguageOpenCL_CPP` is still used for SYCL. It is not clear whether this can be renamed, considering that there is no known frontend actively supporting OpenCL C++. I am looping in @bader for more feedback.

kpet: Linking related discussion with @bader on https://github.com/intel/llvm/issues/2427

AlexeySachkov: I have a PR to remove `CL21` pass in #766

PR #632, opened 2020-07-15
Fix global ctor/dtor translation

hewj03:
See issue #631 

Fix global ctor/dtor translation from SPIRV to LLVM IR

Translate global variable if its storage is StorageClassFunction and its name is either llvm.global_ctors or llvm.global_dtors.
Apply appending linkage to the global variable.
SPV_INTEL_function_pointers extension is required for the translation.

Issue #184, opened 2019-05-03
Support for global constructors and destructors


kpet:
All kernel functions in `@llvm.global_ctors` and `@llvm.global_dtors` should have their execution mode marked as, respectively, `Initializer` or `Finalizer`. Anything else?

This depends on SPIRV 1.1.

PR #757, opened 2020-09-28
Use Initializer execution mode for global constructors

svenvh:
This PR adds support to translate global constructors from LLVM's `llvm.global_ctors` variable to SPIR-V's `Initializer` execution mode, and vice versa.

* Only emit Initializer and Finalizer from SPIR-V 1.1
* Translate Initializer to llvm.global_ctors
* Set Initializer ExecutionMode for global ctors
* Conditionalize translation of llvm.global_ctors/dtors

### DIFF
--- a/lib/SPIRV/PreprocessMetadata.cpp
+++ b/lib/SPIRV/PreprocessMetadata.cpp
@@ -68,6 +68,8 @@ public:
 
   bool runOnModule(Module &M) override;
   void visit(Module *M);
+  void preprocessCXXStructorList(SPIRVMDBuilder::NamedMDWrapper &EM,
+                                 GlobalVariable *V, ExecutionMode EMode);
   void preprocessOCLMetadata(Module *M, SPIRVMDBuilder *B, SPIRVMDWalker *W);
   void preprocessVectorComputeMetadata(Module *M, SPIRVMDBuilder *B,
                                        SPIRVMDWalker *W);
@@ -94,6 +96,24 @@ bool PreprocessMetadata::runOnModule(Module &Module) {
   return true;
 }
 
+void PreprocessMetadata::preprocessCXXStructorList(
+    SPIRVMDBuilder::NamedMDWrapper &EM, GlobalVariable *V,
+    ExecutionMode EMode) {
+  auto *List = dyn_cast_or_null<ConstantArray>(V->getInitializer());
+  if (!List)
+    return;
+
+  for (Value *V : List->operands()) {
+    auto *Structor = cast<ConstantStruct>(V);
+
+    // Each entry in the list is a struct containing 3 members:
+    // (priority, function, data), with function being the entry point.
+    auto *Kernel = cast<Function>(Structor->getOperand(1));
+
+    EM.addOp().add(Kernel).add(EMode).done();
+  }
+}
+
 void PreprocessMetadata::visit(Module *M) {
   SPIRVMDBuilder B(*M);
   SPIRVMDWalker W(*M);
@@ -104,6 +124,10 @@ void PreprocessMetadata::visit(Module *M) {
   // Create metadata representing (empty so far) list
   // of OpExecutionMode instructions
   auto EM = B.addNamedMD(kSPIRVMD::ExecutionMode); // !spirv.ExecutionMode = {}
+
+  // Process special variables in LLVM IR module.
+  if (auto *GV = M->getGlobalVariable("llvm.global_ctors"))
+    preprocessCXXStructorList(EM, GV, spv::ExecutionModeInitializer);
 
   // Add execution modes for kernels. We take it from metadata attached to
   // the kernel functions.
--- a/lib/SPIRV/SPIRVWriter.cpp
+++ b/lib/SPIRV/SPIRVWriter.cpp
@@ -3002,10 +3002,19 @@ bool LLVMToSPIRV::transExecutionMode() {
 
       switch (EMode) {
       case spv::ExecutionModeContractionOff:
-      case spv::ExecutionModeInitializer:
-      case spv::ExecutionModeFinalizer:
         BF->addExecutionMode(BM->add(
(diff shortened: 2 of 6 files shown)


lib/SPIRV/PreprocessMetadata.cpp
```diff
@@ -94,6 +96,28 @@ bool PreprocessMetadata::runOnModule(Module &Module) {
...
+    SPIRVMDBuilder::NamedMDWrapper &EM, GlobalVariable *V,
+    ExecutionMode EMode) {
+  auto *List = dyn_cast_or_null<ConstantArray>(V->getInitializer());
+  if (!List)
+    return;
+
+  for (Value *V : List->operands()) {
+    auto *Structor = dyn_cast<ConstantStruct>(V);
+    if (!Structor)
+      continue;
```
AlexeySachkov, line 109: Two questions:
- do we really need to check each operand? `ConstantArray` holds objects of the same time
- do we even need a runtime check? Would an assertion be enough?

lib/SPIRV/PreprocessMetadata.cpp
```diff
@@ -94,6 +96,28 @@ bool PreprocessMetadata::runOnModule(Module &Module) {
...
+
+    // Each entry in the list is a struct containing 3 members:
+    // (priority, function, data), with function being the entry point.
+    auto *Kernel = dyn_cast<Function>(Structor->getOperand(1));
+    if (!Kernel)
+      continue;
```
AlexeySachkov, line 115: Can it be converted into an assertion or we expect non-functions in some cases?

lib/SPIRV/PreprocessMetadata.cpp
```diff
@@ -105,6 +129,12 @@ void PreprocessMetadata::visit(Module *M) {
   // of OpExecutionMode instructions
   auto EM = B.addNamedMD(kSPIRVMD::ExecutionMode); // !spirv.ExecutionMode = {}
 
+  // Process special variables in LLVM IR module.
+  for (auto &GV : M->globals()) {
+    if (GV.getName() == "llvm.global_ctors")
```
AlexeySachkov, line 134: What about using [`Module::getGlobalVariable`](https://llvm.org/doxygen/classllvm_1_1Module.html#aeb548ad4e336f7f3d3cc91578b5de0a8)?

lib/SPIRV/SPIRVWriter.cpp
```diff
@@ -2676,7 +2676,14 @@ bool LLVMToSPIRV::transGlobalVariables() {
   for (auto I = M->global_begin(), E = M->global_end(); I != E; ++I) {
     if ((*I).getName() == "llvm.global.annotations")
       transGlobalAnnotation(&(*I));
-    else if (MDNode *IO = ((*I).getMetadata("io_pipe_id")))
+    else if ((I->getName() == "llvm.global_ctors" ||
```
AlexeySachkov, line 2679: Am I right that the translator handles these variables as follows:
- if SPIR-V 1.1 is enabled, emit `Initializer`/`Finalizer`
- else if function pointers are available, then translate variables as-is using corresponding extensions
- otherwise, ignore them

right?

lib/SPIRV/PreprocessMetadata.cpp
svenvh: Perhaps a non-dynamic `cast` would be sufficient indeed (which will include the assert in itself).

lib/SPIRV/PreprocessMetadata.cpp
svenvh: Likewise, I will change this into a `cast`.

lib/SPIRV/PreprocessMetadata.cpp
svenvh: Thanks, good suggestion!

lib/SPIRV/SPIRVWriter.cpp
svenvh: Almost.  After this PR, the state will be like this:

LLVM -> SPIR-V:
 * if function pointers are available, then translate variables as-is using corresponding extensions
 * regardless of function pointers, if SPIR-V 1.1 is enabled, emit `Initializer` (emitting `Finalizer` is not part of this PR)
 * else report error

SPIR-V -> LLVM:
 * if the SPIR-V does not contain a `llvm.global_ctors` variable, construct it in LLVM IR from `Initializer` entry points.

Does that make sense?

AlexeySachkov: > else report error

Seems like this happens only if user explicitly stated that function should be emitted as `Initializer` through corresponding metadata from SPIR-V friendly IR, right? I.e. `llvm.global_ctors` is being silently ignored if function pointers are not available.

> Does that make sense?

Yes, looks good to me

svenvh: > I.e. llvm.global_ctors is being silently ignored if function pointers are not available.

`llvm-spirv --spirv-ext=+SPV_INTEL_function_pointers --spirv-max-version=1.0` will also attempt to emit `Initializer`, so it will fail in the same way as without the `--spirv-ext...` option.  That is because the `Initializer` mode is set by the `PreprocessMetadata` pass early in the translator.  I can relax that if you need `llvm.global_ctors` with function pointers before SPIR-V 1.1?

AlexeySachkov: > I can relax that if you need llvm.global_ctors with function pointers before SPIR-V 1.1?

It seems like a nice addition, but I'm not really asking to do that, because I'm not sure that we have such use-cases at the moment. Probably @hewj03 can comment based on #632

AlexeySotkin (approved):
LGTM

kpet:
Contributes to https://github.com/KhronosGroup/SPIRV-LLVM-Translator/issues/184

PR #758, opened 2020-09-29
Add llvm.abs.i32 translation 

MochalovaAn:
Add llvm.abs.i32 translation to "s_abs" instruction.
Add test for this.

### DIFF
--- a/lib/SPIRV/SPIRVUtil.cpp
+++ b/lib/SPIRV/SPIRVUtil.cpp
@@ -1549,6 +1549,22 @@ bool checkTypeForSPIRVExtendedInstLowering(IntrinsicInst *II, SPIRVModule *BM) {
     }
     break;
   }
+  case Intrinsic::abs: {
+        Type *Ty = II->getType();
+        int NumElems = 1;
+        if (auto *VecTy = dyn_cast<FixedVectorType>(Ty)) {
+            NumElems = VecTy->getNumElements();
+            Ty = VecTy->getElementType();
+          
+    }
+        if ((!Ty->isIntegerTy()) ||
+        ((NumElems > 4) &&
+        (NumElems != 8) && (NumElems != 16))) {
+      `        BM->getErrorLog().checkError(false, SPIRVEC_InvalidFunctionCall,
+                                   II->getCalledOperand()->getName().str(), "",
+                                   __FILE__, __LINE__);
+    }
+  }
   default:
     break;
   }
--- a/lib/SPIRV/SPIRVUtil.cpp
+++ b/lib/SPIRV/SPIRVUtil.cpp
@@ -1550,19 +1550,19 @@ bool checkTypeForSPIRVExtendedInstLowering(IntrinsicInst *II, SPIRVModule *BM) {
     break;
   }
   case Intrinsic::abs: {
-        Type *Ty = II->getType();
-        int NumElems = 1;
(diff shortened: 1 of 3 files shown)


lib/SPIRV/SPIRVWriter.cpp
```diff
@@ -2049,6 +2050,17 @@ SPIRVValue *LLVMToSPIRV::transIntrinsicInst(IntrinsicInst *II,
     return BM->addExtInst(STy, BM->getExtInstSetId(SPIRVEIS_OpenCL), ExtOp, Ops,
                           BB);
   }
+  case Intrinsic::abs: {
+    if (!checkTypeForSPIRVExtendedInstLowering(II, BM))
```
bader, line 2054: I think `checkTypeForSPIRVExtendedInstLowering` should be extended to validate `Intrinsic::abs` parameters.

bader:
> Add llvm.abs.i32 translation to "s_abs" instruction.

I think this implementation is able to handle any type, not just i32.

svenvh:
> > Add llvm.abs.i32 translation to "s_abs" instruction.
> 
> I think this implementation is able to handle any type, not just i32.

I think so too; it would be good to test another type (e.g. `i8`) in the test that was added.

@MochalovaAn, there is a spurious merge commit in this pull request too.  When you update the change, could you remove that and rebase on the latest master branch?

lib/SPIRV/SPIRVUtil.cpp
```diff
@@ -1549,6 +1549,22 @@ bool checkTypeForSPIRVExtendedInstLowering(IntrinsicInst *II, SPIRVModule *BM) {
...
     break;
   }
+  case Intrinsic::abs: {
+        Type *Ty = II->getType();
+        int NumElems = 1;
+        if (auto *VecTy = dyn_cast<FixedVectorType>(Ty)) {
+            NumElems = VecTy->getNumElements();
+            Ty = VecTy->getElementType();
+          
+    }
```
AlexeySotkin, line 1559: Please fix formatting. I guess, tabs were used instead of spaces.

lib/SPIRV/SPIRVUtil.cpp
```diff
@@ -1549,6 +1549,21 @@ bool checkTypeForSPIRVExtendedInstLowering(IntrinsicInst *II, SPIRVModule *BM) {
...
+            NumElems = VecTy->getNumElements();
+            Ty = VecTy->getElementType();
+    }
+        if ((!Ty->isIntegerTy()) ||
+        ((NumElems > 4) &&
+        (NumElems != 8) && (NumElems != 16))) {
+      `        BM->getErrorLog().checkError(false, SPIRVEC_InvalidFunctionCall,
+                                   II->getCalledOperand()->getName().str(), "",
+                                   __FILE__, __LINE__);
+    }
```
AlexeySotkin, line 1565: This code can't be compiled by Travis CI https://travis-ci.org/github/KhronosGroup/SPIRV-LLVM-Translator/jobs/732684148#L559, because of some unexpected unicode symbols I guess.

Please follow the [LLVM coding standard](https://llvm.org/docs/CodingStandards.html#whitespace), in particular [this rule](https://llvm.org/docs/CodingStandards.html#golden-rule).

AlexeySotkin (approved):
LGTM, Thanks!

PR #766, opened 2020-10-02
Cleanup OCL*ToSPIRV passes

AlexeySachkov:
Renamed `OCL20ToSPIRV` into `OCLToSPIRV` as it better represents its content.
Removed `OCL21ToSPIRV` as it is not tested and basically should be a dead code.

This is a follow-up from [this thread](https://github.com/KhronosGroup/SPIRV-LLVM-Translator/pull/753#discussion_r498553306) in #753.


lib/SPIRV/SPIRVWriter.cpp
```diff
@@ -3574,11 +3574,11 @@ void addPassesForSPIRV(legacy::PassManager &PassMgr,
   if (Opts.isSPIRVMemToRegEnabled())
     PassMgr.add(createPromoteMemoryToRegisterPass());
   PassMgr.add(createPreprocessMetadata());
-  PassMgr.add(createOCL21ToSPIRV());
+  //PassMgr.add(createOCL21ToSPIRV());
```
svenvh, line 3577:
```suggestion

```

svenvh (commented):
Thanks for doing this.  Just a minor comment, otherwise LGTM.

lib/SPIRV/SPIRVWriter.cpp
AlexeySachkov: Thanks, fixed in 7694b73631bad0e52217cda32365d382e35bb768

AlexeySachkov:
Rebased and applied clang-format to new `OCLToSPIRV.cpp` file in 726b64e04972fcd60a3eeed689ae5bad8b10d9c1, because Travis complained about formatting (I guess because file is considered to be changed by its renaming, i.e. it appears in diff)

svenvh (approved):
LGTM.

PR #767, opened 2020-10-04
[SYCL][FPGA] Fix IVDep translation for accesses to kernel closure fields

AGindinson:
When IVDep is applied to a regular kernel-scope array/pointer variable,
the GEP instructions that access this variable are marked with index
group metadata.
When IVDep is applied to a captured argument of a kernel, however,
the pointer operand of index group-marked instructions becomes a GEP
itself. The latter GEP accesses the closure of a kernel lambda with
an offset that signifies the number of the captured parameter in the
closure layout.

To treat the second case:
1. In forward translation, we ensure that for each memory block, all
pointer accesses marked into IVDep index groups are recognized and
translated into LoopControl parameters. Previously, only the latest
pointer access would get listed in the SPIR-V representation of
attribute parameters.
2. In backward translation, we also differentiate between GEP
instructions that access an array variable directly and the GEPs
accessing the "closure field accesses". An additional map stores
1 index group metadata node per each "pointer to closure + integer
offset" pair. This ensures that all accesses to the same kernel
parameter are marked into the same index group on a given loop nest
level.

### DIFF
--- a/lib/SPIRV/SPIRVReader.cpp
+++ b/lib/SPIRV/SPIRVReader.cpp
@@ -53,6 +53,7 @@
 #include "VectorComputeUtil.h"
 
 #include "llvm/ADT/DenseMap.h"
+#include "llvm/ADT/SmallSet.h"
 #include "llvm/Analysis/LoopInfo.h"
 #include "llvm/BinaryFormat/Dwarf.h"
 #include "llvm/IR/Constants.h"
@@ -803,8 +804,8 @@ void SPIRVToLLVM::setLLVMLoopMetadata(const LoopInstType *LM,
            "Missing loop control parameter!");
   }
   if (LC & LoopControlDependencyArrayINTELMask) {
-    // Collect array variable <-> safelen information
-    std::map<Value *, unsigned> ArraySflnMap;
+    // Collect pointer variable <-> safelen information
+    std::map<Value *, unsigned> PointerSflnMap;
     unsigned NumOperandPairs = LoopControlParameters[NumParam];
     unsigned OperandsEndIndex = NumParam + NumOperandPairs * 2;
     assert(OperandsEndIndex <= LoopControlParameters.size() &&
@@ -812,9 +813,9 @@ void SPIRVToLLVM::setLLVMLoopMetadata(const LoopInstType *LM,
     SPIRVModule *M = LM->getModule();
     while (NumParam < OperandsEndIndex) {
       SPIRVId ArraySPIRVId = LoopControlParameters[++NumParam];
-      Value *ArrayVar = ValueMap[M->getValue(ArraySPIRVId)];
+      Value *PointerVar = ValueMap[M->getValue(ArraySPIRVId)];
       unsigned Safelen = LoopControlParameters[++NumParam];
-      ArraySflnMap.emplace(ArrayVar, Safelen);
+      PointerSflnMap.emplace(PointerVar, Safelen);
     }
 
     // A single run over the loop to retrieve all GetElementPtr instructions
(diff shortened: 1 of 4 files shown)


AGindinson:
TODO:

- [ ] Extend the SPIR-V-level checks

AGindinson:
@AlexeySotkin @AlexeySachkov @MrSidims @mlychkov could you please start reviewing this?

lib/SPIRV/SPIRVReader.cpp
```diff
@@ -53,6 +53,7 @@
 #include "VectorComputeUtil.h"
 
 #include "llvm/ADT/DenseMap.h"
+#include "llvm/ADT/SmallSet.h"
```
AGindinson, line 56: I wonder if using small sets instead of vectors is a good idea (applicable both to SPIRVReader and SPIRVWriter). This does result in less verbose code (a simple `SmallSet<>::insert()` instead of `std::find()` + subsequent iterator checks). The doubts arise from the fact that these collections have a very low number of elements, so a set could impact the performance. I know that some optimizations take place for small numbers of elements in `SmallSet`, but not the nature of these.

Anyway, if there's any input on reviewers' side here, I would appreciate it.

AlexeySotkin: According to https://llvm.org/doxygen/SmallSet_8h_source.html#l00134 `SmallSet` uses vector under the hood if number of elements is less then template parameter `N`. `N` = 4 in your case.

So, I think it is just fine to use `SmalSet` in this case.

test/transcoding/FPGAIVDepLoopAttrOnClosure.ll
```diff
@@ -0,0 +1,553 @@
...
+;     }
+;   });
+;
+;   return 0;
+; }
+
+; RUN: llvm-as < %s > %t.bc
+; RUN: llvm-spirv %t.bc --spirv-ext=+SPV_INTEL_fpga_loop_controls -o %t.spv
+; RUN: llvm-spirv -to-text %t.spv -o %t.spt
+; RUN: FileCheck < %t.spt %s --check-prefix=CHECK-SPIRV
```
MrSidims, line 40: CHECK-SPIRV does almost nothing in the test.

lib/SPIRV/SPIRVReader.cpp
```diff
@@ -826,24 +827,54 @@ void SPIRVToLLVM::setLLVMLoopMetadata(const LoopInstType *LM,
...
       }
     }
 
-    // Create index group metadata nodes - one per each array
+    // Create index group metadata nodes - one per each of the array
     // variables. Mark each GEP accessing a particular array variable
     // into a corresponding index group
-    std::map<unsigned, std::vector<MDNode *>> SafelenIdxGroupMap;
+    std::map<unsigned, SmallSet<MDNode *, 4>> SafelenIdxGroupMap;
+    // Whenever a kernel closure field access is pointed to instead of
```
MrSidims, line 842: I'm unsure, but ...

```suggestion
    // Whenever a kernel closure field access is pointer to instead of
```

test/transcoding/FPGAIVDepLoopAttrOnClosure.ll
MrSidims: Ah, wait, there is a TODO above

MrSidims (commented):
LGTM, but need to test SPIR-V forward transformation.

lib/SPIRV/SPIRVReader.cpp
AGindinson: The passive voice is intended here, implying that "a result of an access to a kernel closure field is itself accessed via a pointer"

test/transcoding/FPGAIVDepLoopAttrOnClosure.ll
AGindinson: Yep, coming soon!

AGindinson:
> LGTM, but need to test SPIR-V forward transformation. 

Done!

PR #769, opened 2020-10-05
[NFC] Refactor OCLBuiltinFuncMangleInfo

AlexeySachkov:
The most notable change is that usages of `std::string` were replaced
with `llvm::StringRef` to simplify code and probably speedup it a little
bit.


svenvh (approved):
LGTM

Issue #716, opened 2020-08-26
Revise scope of CI testing


AlexeySachkov:
This issue aims to collect feedback about current scope of testing done in CI in order to improve the situation

From #699:

@AlexeySachkov:
> I wonder why we haven't spotted the original build fail in CI.
>
> @svenvh, @AlexeySotkin, looks like we don't have a configuration without NDEBUG set (should be equal to LLVM release without assertions)

@svenvh:
> That might be worth adding indeed, as it would align better with LLVM configurations.
>
> Perhaps it's a good opportunity to revise the configurations being tested, because right now there are quite a few configurations. I am not sure there is value in testing all combinations for the majority of the pull requests as it consumes quite a bit of Travis time, which is particularly noticeable when there are many pull requests in flight.

@AlexeySachkov:
> Totally agree here.
>
> BTW, we might want to try GitHub Actions - one thing I love about it is that you can have several config files each of them describing different workflows (and each of them could be triggered by different actions).
>
> For example, having separate workflow for launching clang-format could save us some time there as we don't need to install all those packages to get clang-format and it allows us, for example, attach a patch-file which need to be applied to fix formatting. See https://github.com/intel/llvm/actions/runs/222983731 as an example

@svenvh:
> I am not familiar with GitHub Actions, but what you describe sounds useful indeed. I wonder if it would also allow testing based on the patch? For example if a patch touches CMakeLists.txt then it would run more configurations compared with a patch touching only .cpp files.
>
> Sorry for sidetracking in this PR btw, perhaps we should continue the discussion in a new issue about CI changes?

I propose the following short-term changes:

- add configuration, which perform build without assertions (classic Release build) to catch errors like improper `#ifdef NDEBUG`, unused variables around `assert`-ions, etc.
- drop almost half of existing configurations: I don't think that we need to run all configurations twice with g++ and clang++ compilers - if we care about both, I suggest we focus on one and add 1 or 2 configurations with another, just to have basic acceptance check - in 99% cases results should be the same

Also I propose to consider switching to [GitHub Actions](https://docs.github.com/en/actions) which might get us the following benefits:
- we can have several [workflows](https://docs.github.com/en/actions/configuring-and-managing-workflows/configuring-a-workflow) for different purposes like performing clang-format check and build + tests
  - this should simplify our CI config files as each workflow is described in a separate config file - there will be no more bunch of `if`s to detect what exactly we need to do
[...]

PR #770, opened 2020-10-07
Move clang-format and cland-tidy to GitHub Actions

AlexeySachkov:
Made in scope of #716

The main advantages I see:
Now clang-format and tidy will be reported as a separate item under "Checks" - this is intended to simplify checking of results, because functional and styling checks will be separated.
Also, new workflow is optimized to not be launched on unrelated changes like documentation updates.
One more feature is that you can now download `.patch` file with clang-format fixes, which can be applied locally to fix formatting.

Additional notable changes:
- both `tidy` and `format` were enabled on `lib/Mangler/` and `lib/runtime/` subdirectories
- `tidy` was enabled on the translator header files
- both `tidy` and `format` were enabled on `OpenCL.std.h` and `spirv.hpp`, which are "external" files and it is not necessary that they conform with our coding standards. `clang-format` was disabled in those files in place, but `clang-tidy` will still be launched for them

Maintenance-wide I don't expect a lot of efforts required: GitHub Actions are well documented and I tried to make workflow as much verbose as possible to highlight potential tricky parts and describe all workarounds I had to apply.


.github/workflows/check-code-style.yml
```diff
@@ -0,0 +1,134 @@
...
+        # commit. Depth of 2 is enough, because GitHub Actions supply us with
+        # merge commit as {{ github.sha }}, i.e. the second commit is a merge
+        # base between target branch and PR
+        fetch-depth: 2
+
+    - name: Gather list of changes
+      id: gather-list-of-changes
+      run: |
+        # GitHub Actions marks step as failed even if one single command has
+        # returned non-zero exit code. This might happen, for example, it we
```
svenvh, line 44:
```suggestion
        # returned non-zero exit code. This might happen, for example, if we
```

.github/workflows/check-code-style.yml
```diff
@@ -0,0 +1,134 @@
...
+        # pipeline git diff --name-only and subsequent grep for files with
+        # particular extensions: if user didn't modify any .cpp files, grep
+        # won't find any matches and return non-zero exit code.
+        # TODO: consider offloading files extension filter to clang-format-diff
```
svenvh, line 48: Good point, why don't we leave the filtering and running work in custom scripts?  That will make it easier to run those scripts locally too.

svenvh (commented):
Thanks for working on this!

.github/workflows/check-code-style.yml
AlexeySachkov: Done in 1620536. There are two files (`OpenCL.std.h` and `spirv.hpp`), which are kind of external to us, so we shouldn't expect for them to be compliant with our coding standards (at least this is my understanding). So, I've disabled `clang-format` for them using in-place comments, but `clang-tidy` will still be launched to check changes in there - I think it shouldn't be a blocker, but still a notable change

.travis.yml
```diff
@@ -103,45 +95,17 @@ script:
...
-          -DBUILD_SHARED_LIBS=${SHARED_LIBS} \
-          -DLLVM_BUILD_TOOLS=ON \
-          -DLLVM_EXTERNAL_LIT="/usr/lib/llvm-${LLVM_VERSION}/build/utils/lit/lit.py" \
-          -DLLVM_INCLUDE_TESTS=ON \
-          -DCMAKE_INSTALL_PREFIX=../install/ \
-          -DCMAKE_EXPORT_COMPILE_COMMANDS=ON \
-          -G "Unix Makefiles"
-      fi
-    else
-      BUILDDIR=$(pwd)
```
AlexeySachkov, line 118: This `else` branch should be preserved

svenvh (approved):

AlexeySotkin (approved):

PR #787, opened 2020-10-14
Further improvements to generation of SPIRV*Enum.h

svenvh:
For many enums in `spirv.hpp`, an isValid function was generated in
`SPIRVIsValidEnum.h` even though that function was never used.  Remove
them to reduce build times and maintenance burden.

If a removed isValid function is needed again in the future, the enum
name needs to be added back to `gen_spirv.bash` followed by a rerun of
the script.

Regenerate SPIRVIsValidEnum.h and SPIRVNameMapEnum.h using the
updated `gen_spirv.bash` script.

Generate all .h files in one invocation

Move enum lists to top of script.

### DIFF
--- a/lib/SPIRV/libSPIRV/SPIRVNameMapEnum.h
+++ b/lib/SPIRV/libSPIRV/SPIRVNameMapEnum.h
@@ -139,6 +139,8 @@ template <> inline void SPIRVMap<Decoration, std::string>::init() {
   add(DecorationHlslSemanticGOOGLE, "HlslSemanticGOOGLE");
   add(DecorationUserSemantic, "UserSemantic");
   add(DecorationUserTypeGOOGLE, "UserTypeGOOGLE");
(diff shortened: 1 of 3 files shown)


lib/SPIRV/libSPIRV/SPIRVNameMapEnum.h
```diff
@@ -499,9 +500,6 @@ template <> inline void SPIRVMap<Capability, std::string>::init() {
   add(CapabilityUSMStorageClassesINTEL, "USMStorageClassesINTEL");
   add(CapabilityFPGAMemoryAccessesINTEL, "FPGAMemoryAccessesINTEL");
   add(CapabilityIOPipeINTEL, "IOPipeINTEL");
-  add(CapabilityArbitraryPrecisionFloatingPointINTEL,
```
svenvh, line 502: This is actually a duplicate of line 488.

AlexeySachkov (approved):

PR #780, opened 2020-10-12
[NFC] Re-organize tests a bit

AlexeySachkov:
Renamed tests for LLVM intrinsics from "something" to "llvm.something" to better represent their idea tot test particular llvm intrinsic.
Moved negative tests into corresponding folder

PR #790, opened 2020-10-14
Translate the llvm.minnum intrinsic function

AGindinson:
Lower such calls to a corresponding OpenCL ExtInst.

### DIFF
--- a/lib/SPIRV/SPIRVWriter.cpp
+++ b/lib/SPIRV/SPIRVWriter.cpp
@@ -1971,6 +1971,7 @@ bool LLVMToSPIRV::isKnownIntrinsic(Intrinsic::ID Id) {
   case Intrinsic::experimental_constrained_fcmps:
   case Intrinsic::experimental_constrained_fmuladd:
   case Intrinsic::fmuladd:
+  case Intrinsic::maxnum:
   case Intrinsic::minnum:
   case Intrinsic::memset:
   case Intrinsic::memcpy:
--- a/lib/SPIRV/SPIRVWriter.cpp
+++ b/lib/SPIRV/SPIRVWriter.cpp
@@ -1971,6 +1971,7 @@ bool LLVMToSPIRV::isKnownIntrinsic(Intrinsic::ID Id) {
   case Intrinsic::experimental_constrained_fcmps:
   case Intrinsic::experimental_constrained_fmuladd:
   case Intrinsic::fmuladd:
+  case Intrinsic::minnum:
   case Intrinsic::memset:
   case Intrinsic::memcpy:
   case Intrinsic::nearbyint:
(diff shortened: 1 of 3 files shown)


test/minnum.ll
```diff
@@ -0,0 +1,39 @@
...
+target triple = "spir64-unknown-unknown"
+
+; CHECK-SPIRV: Function
+; CHECK-SPIRV: FunctionParameter {{[0-9]+}} [[x:[0-9]+]]
+; CHECK-SPIRV: FunctionParameter {{[0-9]+}} [[y:[0-9]+]]
+; CHECK-SPIRV: ExtInst {{[0-9]+}} [[res:[0-9]+]] {{[0-9]+}} fmin [[x]] [[y]]
+; CHECK-SPIRV: ReturnValue [[res]]
+define spir_func float @TestHalf(float %x, float %y) {
+entry:
+  ; CHECK-LLVM: call spir_func float @[[TestHalfName:.+]](float %{{.+}}, float %{{.+}})
```
bader, line 19: What is this check suppose to validate?
I think we should look for exact name.

bader (approved):

test/minnum.ll
AGindinson: Got it, I'll update this to match the exact mangling.

jeffhammond:
Is fshl going to be added to this or is that a more involved fix?

lib/SPIRV/SPIRVWriter.cpp
```diff
@@ -2232,6 +2232,16 @@ SPIRVValue *LLVMToSPIRV::transIntrinsicInst(IntrinsicInst *II,
     return BM->addExtInst(STy, BM->getExtInstSetId(SPIRVEIS_OpenCL), ExtOp, Ops,
                           BB);
   }
+  case Intrinsic::minnum: {
```
AlexeySotkin, line 2235: Please add `minnum` to `LLVMToSPIRV::isKnownIntrinsic`.

AlexeySotkin (changes requested):

lib/SPIRV/SPIRVWriter.cpp
AGindinson: Right. Should I add `maxnum` as well with a separate commit, or a separate patch would be desirable?

AlexeySotkin: I think we can do `maxnum` with a separate commit in scope of this PR.

AGindinson: Done.

test/minnum.ll
```diff
@@ -0,0 +1,39 @@
+; RUN: llvm-as %s -o %t.bc
```
AlexeySotkin, line 1: For the test naming please take a look at https://github.com/KhronosGroup/SPIRV-LLVM-Translator/pull/780
And for the test location please mind https://github.com/KhronosGroup/SPIRV-LLVM-Translator/pull/780#issuecomment-708219326

AGindinson: There's a bunch of other intrinsic-related tests that should be moved likewise. Would it be fine if I renamed/moved the new test, and factored out all other related updates into a separate PR?

AlexeySachkov: @AGindinson, I haven't yet decided about particular test location for LLVM intrinsics, so, it is okay to me to leave the test in its current location. However, I do agree that it would be better if you add `llvm.` prefix to its name

AGindinson: Fine, I've moved `llvm.minnum.ll` back to the original directory, and am ready to follow-up with a renaming PR for other files.

AlexeySotkin (approved):

AGindinson:
Closing; another patch will be commited to open-source that includes this intrinsic and many others.

PR #800, opened 2020-10-20
Translate the llvm.fshl intrinsic function

AGindinson:
"Funnel shift left" doesn't have an analogue in the OpenCL
ExtInst set. We unroll `llvm.fshl.i<n>(i<n>, i<n>, i<n>)`
into a small algorithm that performs the actual funnel shift.

A detailed description of FSHL can be found at
https://llvm.org/docs/LangRef.html#llvm-fshl-intrinsic

### DIFF
--- a/lib/SPIRV/SPIRVRegularizeLLVM.cpp
+++ b/lib/SPIRV/SPIRVRegularizeLLVM.cpp
@@ -85,6 +85,24 @@ public:
   /// @spirv.llvm_memset_* and replace it with @llvm.memset.
   void lowerMemset(MemSetInst *MSI);
 
+  /// No SPIR-V counterpart for @llvm.fshl.* intrinsic. It will be lowered
+  /// to a newly generated @spirv.llvm_fshl_* function.
+  ///
+  /// Conceptually, FSHL:
+  /// 1. concatenates the ints, the first one being the more significant;
+  /// 2. performs a left shift-rotate on the resulting doubled-sized int;
+  /// 3. returns the most significant bits of the shift-rotate result,
+  ///    the number of bits being equal to the size of the original integers.
+  /// If FSHL operates on a vector type instead, the same operations are
+  /// performed for each set of corresponding vector elements.
+  ///
+  /// The actual implementation algorithm will be slightly different for
+  /// simplification purposes.
+  void lowerFunnelShiftLeft(IntrinsicInst *FSHLIntrinsic);
+  void buildFunnelShiftLeftFunc(Function *FSHLFunc);
+
+  static std::string lowerLLVMIntrinsicName(IntrinsicInst *II);
+
   static char ID;
 
 private:
@@ -94,17 +112,22 @@ private:
 
 char SPIRVRegularizeLLVM::ID = 0;
 
-void SPIRVRegularizeLLVM::lowerMemset(MemSetInst *MSI) {
-  if (isa<Constant>(MSI->getValue()) && isa<ConstantInt>(MSI->getLength()))
-    return; // To be handled in LLVMToSPIRV::transIntrinsicInst
-  Function *IntrinsicFunc = MSI->getCalledFunction();
+std::string SPIRVRegularizeLLVM::lowerLLVMIntrinsicName(IntrinsicInst *II) {
+  Function *IntrinsicFunc = II->getCalledFunction();
   assert(IntrinsicFunc && "Missing function");
   std::string FuncName = IntrinsicFunc->getName().str();
   std::replace(FuncName.begin(), FuncName.end(), '.', '_');
   FuncName = "spirv." + FuncName;
+  return FuncName;
+}
+
+void SPIRVRegularizeLLVM::lowerMemset(MemSetInst *MSI) {
+  if (isa<Constant>(MSI->getValue()) && isa<ConstantInt>(MSI->getLength()))
+    return; // To be handled in LLVMToSPIRV::transIntrinsicInst
+
+  std::string FuncName = lowerLLVMIntrinsicName(MSI);
   if (MSI->isVolatile())
     FuncName += ".volatile";
-
   // Redirect @llvm.memset.* call to @spirv.llvm_memset_*
   Function *F = M->getFunction(FuncName);
   if (F) {
@@ -135,6 +158,53 @@ void SPIRVRegularizeLLVM::lowerMemset(MemSetInst *MSI) {
   expandMemSetAsLoop(cast<MemSetInst>(MemSet));
   MemSet->eraseFromParent();
   return;
+}
+
+void SPIRVRegularizeLLVM::buildFunnelShiftLeftFunc(Function *FSHLFunc) {
+  if (!FSHLFunc->empty())
+    return;
+
+  auto *RotateBB = BasicBlock::Create(M->getContext(), "rotate", FSHLFunc);
+  IRBuilder<> Builder(RotateBB);
+  Type *Ty = FSHLFunc->getReturnType();
+  // Build the actual funnel shift rotate logic.
+  // In the comments, "int" is used interchangeably with "vector of int
+  // elements".
+  FixedVectorType *VectorTy = dyn_cast<FixedVectorType>(Ty);
+  Type *IntTy = VectorTy ? VectorTy->getElementType() : Ty;
+  unsigned BitWidth = IntTy->getIntegerBitWidth();
+  ConstantInt *BitWidthConstant = Builder.getInt({BitWidth, BitWidth});
+  Value *BitWidthForInsts =
+      VectorTy ? Builder.CreateVectorSplat(VectorTy->getNumElements(),
+                                           BitWidthConstant)
+               : BitWidthConstant;
+  auto *RotateModVal =
+      Builder.CreateURem(/*Rotate*/ FSHLFunc->getArg(2), BitWidthForInsts);
+  // Shift the more significant number left, the "rotate" number of bits
+  // will be 0-filled on the right as a result of this regular shift.
+  auto *ShiftLeft = Builder.CreateShl(FSHLFunc->getArg(0), RotateModVal);
+  // We want the "rotate" number of the second int's MSBs to occupy the
+  // rightmost "0 space" left by the previous operation. Therefore,
+  // subtract the "rotate" number from the integer bitsize...
+  auto *SubRotateVal = Builder.CreateSub(BitWidthForInsts, RotateModVal);
+  // ...and right-shift the second int by this number, zero-filling the MSBs.
+  auto *ShiftRight = Builder.CreateLShr(FSHLFunc->getArg(1), SubRotateVal);
+  // A simple binary addition of the shifted ints yields the final result.
+  auto *FunnelShiftRes = Builder.CreateOr(ShiftLeft, ShiftRight);
+  Builder.CreateRet(FunnelShiftRes);
+}
+
+void SPIRVRegularizeLLVM::lowerFunnelShiftLeft(IntrinsicInst *FSHLIntrinsic) {
+  // Get a separate function - otherwise, we'd have to rework the CFG of the
+  // current one. Then simply replace the intrinsic uses with a call to the new
+  // function.
+  FunctionType *FSHLFuncTy = FSHLIntrinsic->getFunctionType();
+  Type *FSHLRetTy = FSHLFuncTy->getReturnType();
+  const std::string FuncName = lowerLLVMIntrinsicName(FSHLIntrinsic);
+  Function *FSHLFunc =
+      getOrCreateFunction(M, FSHLRetTy, FSHLFuncTy->params(), FuncName);
+  buildFunnelShiftLeftFunc(FSHLFunc);
+  FSHLIntrinsic->setCalledFunction(FSHLFunc);
 }
 
 bool SPIRVRegularizeLLVM::runOnModule(Module &Module) {
@@ -170,8 +240,11 @@ bool SPIRVRegularizeLLVM::regularize() {
           Function *CF = Call->getCalledFunction();
           if (CF && CF->isIntrinsic()) {
             removeFnAttr(Call, Attribute::NoUnwind);
-            if (auto *MSI = dyn_cast<MemSetInst>(Call))
+            auto *II = cast<IntrinsicInst>(Call);
+            if (auto *MSI = dyn_cast<MemSetInst>(II))
               lowerMemset(MSI);
+            else if (II->getIntrinsicID() == Intrinsic::fshl)
+              lowerFunnelShiftLeft(II);
           }
         }
 
--- a/test/llvm-intrinsics/fshl.ll
+++ b/test/llvm-intrinsics/fshl.ll
@@ -1,0 +1,116 @@
+; RUN: llvm-as %s -o %t.bc
+; RUN: llvm-spirv %t.bc -spirv-text -o - | FileCheck %s --check-prefix=CHECK-SPIRV
+; RUN: llvm-spirv %t.bc -o %t.spv
+; RUN: llvm-spirv -r %t.spv -o %t.rev.bc
+; RUN: llvm-dis %t.rev.bc -o - | FileCheck %s --check-prefix=CHECK-LLVM
+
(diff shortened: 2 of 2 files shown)


AGindinson:
TODO:

- [ ] Hash the generated SPIR-V functions for each integer size to avoid re-generation in case of multiple intrinsics.

(possibly as a separate PR, but I'm getting to this right away)

lib/SPIRV/SPIRVWriter.cpp
```diff
@@ -2227,6 +2228,77 @@ SPIRVValue *LLVMToSPIRV::transIntrinsicInst(IntrinsicInst *II,
...
+  case Intrinsic::fshl: {
+    // We will implement fshl(a, b, c) in SPIR-V instructions.
+    // "Funnel shift rotate" operates on same-sized integers; its return type
+    // also matches.
+    SPIRVType *Ty = transType(II->getType());
+    // Create a separate function - otherwise we'd have to rework the CFG
+    // of the current function. Then simply translate the intrinsic call
+    // into the new function.
+    auto *FSHLFuncTy = BM->addFunctionType(Ty, {Ty, Ty, Ty});
+    auto *FSHLFunc = BM->addFunction(FSHLFuncTy);
```
AlexeySachkov, line 2240: I suggest to add a name to this function and query it first to check if it exists already (might be the case if there are several calls to `fshl`). For ideas how to "mangle" function name you can take a look at https://github.com/KhronosGroup/SPIRV-LLVM-Translator/pull/614/files#diff-7990d99acb2f8325d245dfecdab758eae5aebae2b54c04fd8e906212eb80ae10R2438

Also, by having specially-defined function name for `fshl` replacement we could replace call to that function back to `fshl` intrinsic in SPIR-V Reader (this is an idea for future improvement - not asking to do this right now)

AGindinson: Is it a good enough practice to query such functions by name instead of, say, hashing the functions by integer bitsize? The latter wouldn't interfere with naming the functions properly, so it's merely a question of whether names should be used for such queries.

AlexeySachkov: > Is it a good enough practice to query such functions by name instead of, say, hashing the functions by integer bitsize?

I can't say exactly whether it is the best thing from the performance prospective, but this (querying functions by name) is at least widely-used mechanism:

https://github.com/KhronosGroup/SPIRV-LLVM-Translator/blob/e1405604253a8569375c8031a73e020af2da9bd1/lib/SPIRV/OCLUtil.h#L523-L525

The snippet above is called from 5 different places in the translator. Also you can find 6 more direct uses of [`llvm::Module::getOrInsertFunction`](https://llvm.org/doxygen/classllvm_1_1Module.html#abb107e4edbf05eae92936cba6801c2d9) like:

https://github.com/KhronosGroup/SPIRV-LLVM-Translator/blob/e1405604253a8569375c8031a73e020af2da9bd1/lib/SPIRV/SPIRVReader.cpp#L1453

Personally, I like the idea of querying by names more because of its scalability: if you decide to hash functions by integer bitsize, then you need to have a separate map for each function you would like to hash and I believe that we will have more unsupported intrinsics which we will replace with corresponding implementation in the translator.

Also, having predefined names allows to do some optimizations like:

> we could replace call to that function back to fshl intrinsic in SPIR-V Reader 

An example of such would be:

https://github.com/KhronosGroup/SPIRV-LLVM-Translator/blob/e1405604253a8569375c8031a73e020af2da9bd1/lib/SPIRV/OCLUtil.cpp#L1557-L1562

> The latter wouldn't interfere with naming the functions properly, so it's merely a question of whether names should be used for such queries.

We usually use something like this:

https://github.com/KhronosGroup/SPIRV-LLVM-Translator/blob/e1405604253a8569375c8031a73e020af2da9bd1/lib/SPIRV/SPIRVInternal.h#L356-L360

However, querying whether function exists in `SPIRVModule` will require updating existing infrastructure: right now functions are just stored in a vector. Such query is much easier to do at LLVM IR level and as I can see, all functions are stored within LLVM `Module` using `StringMap` (`StringRef` -> `Value`) - you might want to consider moving this piece of code into `SPIRVRegularize` pass

AGindinson: As a matter of fact, couldn't I just do `FSHLIntrinsicPtr->getDeclaration()->replaceAllUsesWith(ImplementedFSHLPtr)`? This definitely turns the tables in `SPIRVRegularizeLLVM`'s favor.

AGindinson: Done, thanks for the detailed suggestion!
Should I also add a test for function re-usage scenario/another integer type?

AlexeySachkov: > As a matter of fact, couldn't I just do FSHLIntrinsicPtr->getDeclaration()->replaceAllUsesWith(ImplementedFSHLPtr)?
> UPD: This exact trick wouldn't work for call instructions, but I'm still going to try out your suggestions.

I'm not sure that this is absolutely correct to replace all uses of declaration with another declaration. However, replacing all uses of particular `CallInst` with another `Value` (like another `CallInst)` should work fine

AGindinson: > I'm not sure that this is absolutely correct to replace all uses of declaration with another declaration. However, replacing all uses of particular CallInst with another Value (like another CallInst) should work fine

I've ended up mimicking `lowerMemset()`'s behavior, i.e. using `replaceCalledFunction()`

lib/SPIRV/SPIRVRegularizeLLVM.cpp
```diff
@@ -137,6 +156,75 @@ void SPIRVRegularizeLLVM::lowerMemset(MemSetInst *MSI) {
   return;
 }
 
+void SPIRVRegularizeLLVM::buildFunnelShiftLeftFunc(Function *FSHLFunc) {
+  if (!FSHLFunc->empty())
+    return;
+
+  auto *IntTy = dyn_cast<IntegerType>(FSHLFunc->getReturnType());
+  assert(IntTy && "llvm.fshl: expected an integer return type");
```
AlexeySachkov, line 164:
```suggestion
  auto *IntTy = cast<IntegerType>(FSHLFunc->getReturnType());
```

`cast` already includes an assertion within it.

lib/SPIRV/SPIRVRegularizeLLVM.cpp
```diff
@@ -137,6 +156,75 @@ void SPIRVRegularizeLLVM::lowerMemset(MemSetInst *MSI) {
...
+  assert(FSHLFunc->arg_size() == 3 && "llvm.fshl: expected 3 arguments");
+  for (Argument &Arg : FSHLFunc->args())
+    assert(Arg.getType()->getTypeID() == IntTy->getTypeID() &&
+           "llvm.fshl: mismatched return type and argument types");
```
AlexeySachkov, line 168: I'm not sure that these asserts are really necessary, because this stuff should be guaranteed by LLVM IR already

lib/SPIRV/SPIRVWriter.cpp
AlexeySachkov: > Should I also add a test for function re-usage scenario/another integer type?

Yes, please. I think both tests (re-usage of existing type and using another integer type) are very useful as we can encounter such use-cases in real apps

AlexeySachkov (commented):
A few minor comments about extra assertions + there is a comment in separate comments thread here about improving tests.

Otherwise, LGTM in terms of approach and implementation, but I would need some additional time to double-check correctness of the LLVM IR which is being built within `buildFunnelShiftLeftFunc`

test/llvm.fshl.ll
```diff
@@ -0,0 +1,63 @@
+; RUN: llvm-as %s -o %t.bc
```
AlexeySachkov, line 1: One more comment regarding filename in spite of recently merged #780: this can be just named `fshl.ll`, but should be placed within `llvm-intrinsics` subfolder

lib/SPIRV/SPIRVWriter.cpp
AGindinson: Done.

lib/SPIRV/SPIRVRegularizeLLVM.cpp
AGindinson: Removed all asserts.

test/llvm.fshl.ll
AGindinson: Sure, I've rebased & moved the file.

lib/SPIRV/SPIRVRegularizeLLVM.cpp
```diff
@@ -137,6 +156,69 @@ void SPIRVRegularizeLLVM::lowerMemset(MemSetInst *MSI) {
...
+  auto *CondBB = BasicBlock::Create(M->getContext(), "cond", FSHLFunc);
+  auto *RotateBB =
+      BasicBlock::Create(M->getContext(), "rotate", FSHLFunc); // Main logic
+  auto *PhiBB = BasicBlock::Create(M->getContext(), "phi", FSHLFunc);
+
+  IRBuilder<> Builder(CondBB);
+  // If the number of bits to rotate for is divisible by the bitsize,
+  // the shift becomes useless, and we should bypass the main logic in that
+  // case.
+  auto *IntTy = cast<IntegerType>(FSHLFunc->getReturnType());
```
AlexeySachkov, line 174: I've just realized that intrinsic can accept vectors as arguments: in this case we need to apply desired transformation to each vector element.

If I understand correctly, this shouldn't be a huge problem, because each operation used supports vector as arguments, except `br`: we can't perform a jump based on vector of booleans. Having vectors in mind we probably need to either drop this optimization for the case when rotate is not needed at all or go for it only if the rotate is not needed for any vector element

AlexeySachkov (changes requested):
Logic seems to be correct, but replacement function doesn't work with vector types

lib/SPIRV/SPIRVRegularizeLLVM.cpp
AGindinson: The vector catch is a really nice one. I'm also thinking it's not too much of a problem, so I'll add vector support as a separate commit.
As far as `br` is concerned, I'm leaning towards
> go for it only if the rotate is not needed for any vector element

AGindinson: Done. Entirely dropping the condition-based optimization shouldn't cause too much problems, after all - it only affects a minority of cases for non-LLVM-based backends.

lib/SPIRV/SPIRVRegularizeLLVM.cpp
```diff
@@ -137,6 +160,53 @@ void SPIRVRegularizeLLVM::lowerMemset(MemSetInst *MSI) {
...
+    return;
+
+  auto *RotateBB = BasicBlock::Create(M->getContext(), "rotate", FSHLFunc);
+  IRBuilder<> Builder(RotateBB);
+  Type *Ty = FSHLFunc->getReturnType();
+  // Build the actual funnel shift rotate logic.
+  // In the comments, "int" is used interchangeably with "vector of int
+  // elements".
+  FixedVectorType *VectorTy = dyn_cast<FixedVectorType>(Ty);
+  Type *IntTy = VectorTy ? VectorTy->getElementType() : Ty;
```
AlexeySachkov, line 174:
```suggestion
  Type *ElementTy = VectorTy ? VectorTy->getElementType() : Ty;
```

test/llvm-intrinsics/fshl.ll
```diff
@@ -0,0 +1,115 @@
...
+; CHECK-SPIRV: FunctionParameter [[TYPE_VEC_INT_16]] [[Y_FSHL:[0-9]+]]
+; CHECK-SPIRV: FunctionParameter [[TYPE_VEC_INT_16]] [[ROT:[0-9]+]]
+
+; Check that the intrinsics were generated again in reverse translation,
+; replacing the SPIR-V level implementations.
+; CHECK-LLVM-NOT: declare {{.*}} @spirv.llvm_fshl_{{.*}}
+
+; CHECK-LLVM: declare i32 @llvm.fshl.i32(i32, i32, i32)
+; Function Attrs: nounwind readnone speculatable willreturn
+declare i32 @llvm.fshl.i32(i32, i32, i32) #1
```
AlexeySachkov, line 96: I think it would be also good to check for `call i32 @llvm.fshl.i32` just to be sure that this is not an unused declaration

AlexeySachkov (approved):
A few minor comments, other than this, LGTM

lib/SPIRV/SPIRVRegularizeLLVM.cpp
AGindinson: Technically, this variable gets used for a "non-element" type in the simple integer scenario....

lib/SPIRV/SPIRVRegularizeLLVM.cpp
```diff
@@ -85,6 +85,24 @@ class SPIRVRegularizeLLVM : public ModulePass {
...
+  /// Conceptually, FSHL:
+  /// 1. concatenates the ints, the first one being the more significant;
+  /// 2. performs a left shift-rotate on the resulting doubled-sized int;
+  /// 3. returns the most significant bits of the shift-rotate result,
+  ///    the number of bits being equal to the size of the original integers.
+  /// If FSHL operates on a vector type instead, the same operations are
+  /// performed for each set of corresponding vector elements.
+  ///
+  /// The actual implementation algorithm will be slightly different to speed
+  /// things up.
```
AlexeySotkin, line 100: I guess this sentence is not actual anymore.

lib/SPIRV/SPIRVRegularizeLLVM.cpp
```diff
@@ -85,6 +85,24 @@ class SPIRVRegularizeLLVM : public ModulePass {
...
+  void lowerFunnelShiftLeft(IntrinsicInst *FSHLIntrinsic);
+  void buildFunnelShiftLeftFunc(Function *FSHLFunc);
```
AlexeySotkin, line 102: Why do need 2 functions? Do you anticipate other usages for `buildFunnelShiftLeftFunc`?

lib/SPIRV/SPIRVRegularizeLLVM.cpp
AGindinson: I'd argue it is, because I'm performing a left shift + right shift + bitwise addition instead of the algorithm described here.

lib/SPIRV/SPIRVRegularizeLLVM.cpp
AGindinson: At the moment of writing these, the logic of building FSHL involved 3 basic blocks, so I've chosen to split the functions for more concision. With `buildFunnelShiftLeftFunc` having become simpler, I personally still prefer this split: `lowerFunnelShiftLeft()` remains dedicated to function name production, function extraction and call replacement. Would it be preferable if I merged the functions instead?

lib/SPIRV/SPIRVRegularizeLLVM.cpp
AlexeySotkin: Does it "speed things up"?

lib/SPIRV/SPIRVRegularizeLLVM.cpp
AlexeySotkin: I don't have strong preference here. Up to you.

lib/SPIRV/SPIRVRegularizeLLVM.cpp
AGindinson: I'd speculate that a real MSB extraction would require more instructions. I think I should replace it with "simplification".

test/llvm-intrinsics/fshl.ll
AGindinson: Done. I've ended up replacing the declaration checks with call checks - retaining both check types and making FileCheck work correctly proved tricky.

AlexeySotkin (approved):

AGindinson:
Squashed the commits, merged the commit messages.

PR #805, opened 2020-10-28
Fix ptr/var/global annotations after https://reviews.llvm.org/D88645

MrSidims:

### DIFF
--- a/lib/SPIRV/SPIRVReader.cpp
+++ b/lib/SPIRV/SPIRVReader.cpp
@@ -2563,6 +2563,9 @@ Value *SPIRVToLLVM::transValueWithoutDecoration(SPIRVValue *BV, Function *F,
           RetTy = Int8PtrTyPrivate;
           ValAsArg = Builder.CreateBitCast(Val, Int8PtrTyPrivate);
         }
+        Value *Args[] = {ValAsArg, GS, UndefInt8Ptr, UndefInt32, UndefInt8Ptr};
+        auto *IntrinsicCall = Builder.CreateIntrinsic(IID, RetTy, Args);
+        return mapValue(BV, IntrinsicCall);
       }
     }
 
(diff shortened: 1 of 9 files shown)


MrSidims:
Tagging @AlexeySachkov  @vmaksimo

lib/SPIRV/SPIRVReader.cpp
```diff
@@ -3658,9 +3661,10 @@ void SPIRVToLLVM::transIntelFPGADecorations(SPIRVValue *BV, Value *V) {
     Type *Int8PtrTyPrivate = Type::getInt8PtrTy(*Context, SPIRAS_Private);
     IntegerType *Int32Ty = Type::getInt32Ty(*Context);
 
-    llvm::Constant *Fields[4] = {
+    llvm::Constant *Fields[5] = {
```
vmaksimo, line 3664: You can remove array size to avoid this line change in the future (if the signature is changed again)

test/transcoding/intel_fpga_lsu_optimized.ll
```diff
@@ -1,4 +1,4 @@
-; LLVM IR generated by Intel SYCL Clang compiler (https://github.com/intel/llvm)
+;f LLVM IR generated by Intel SYCL Clang compiler (https://github.com/intel/llvm)
```
vmaksimo, line 1: Please remove this symbol

vmaksimo:
LGTM, thanks!

lib/SPIRV/SPIRVReader.cpp
MrSidims: That sounds good, but it will require some style alignment, for example in `transUserSemantic` there is `llvm::Constant *Fields[5] = {`. If we want to remove the sizes - it's better to do in a single NFC patch.

test/transcoding/intel_fpga_lsu_optimized.ll
MrSidims: Oops, thanks!

lib/SPIRV/SPIRVReader.cpp
MrSidims: Well, actualy, I've started leaning to appose this suggestion. If/when signature changes again, we would need to change this array anyway, and may be with the size set it would be harder to pass by this code during initial eye-debugging.

AlexeySachkov (approved):

PR #806, opened 2020-10-30
Fix issues found by a static analysis tool

MochalovaAn:

### DIFF
--- a/lib/SPIRV/SPIRVLowerSPIRBlocks.cpp
+++ b/lib/SPIRV/SPIRVLowerSPIRBlocks.cpp
@@ -209,6 +209,7 @@ private:
       Value *CtxAlign = nullptr;
       getBlockInvokeFuncAndContext(CallBlkBind, &InvF, &Ctx, &CtxLen,
                                    &CtxAlign);
+      assert (Ctx && "Invalid context");
       for (auto II = CallBlkBind->user_begin(), EE = CallBlkBind->user_end();
            II != EE;) {
         auto BlkUser = *II++;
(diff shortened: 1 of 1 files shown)


lib/SPIRV/SPIRVLowerSPIRBlocks.cpp
```diff
@@ -247,6 +247,7 @@ class SPIRVLowerSPIRBlocks : public ModulePass {
     if (!Ctx)
       getBlockInvokeFuncAndContext(CallGetBlkCtx->getArgOperand(0), nullptr,
                                    &Ctx);
+    assert(Ctx && "Invalid block");
```
AlexeySachkov, line 250:
```suggestion
    assert(Ctx && "Invalid context");
```

lib/SPIRV/SPIRVLowerSPIRBlocks.cpp
```diff
@@ -226,7 +226,7 @@ class SPIRVLowerSPIRBlocks : public ModulePass {
             assert(CI->getArgOperand(0) == CallBlkBind);
             Changed |= lowerGetBlockInvoke(CI, cast<Function>(InvF));
           } else if (Name == SPIR_INTRINSIC_GET_BLOCK_CONTEXT) {
-            assert(CI->getArgOperand(0) == CallBlkBind);
+            assert(Ctx && CI->getArgOperand(0) == CallBlkBind);
```
AlexeySachkov, line 229: I suggest to place an assertion for `Ctx` right after call to `getBlockInvokeFuncAndContext` at line 210: I think it is better to check result of the function right after call to it and `Ctx` is also used on line 234, where it is not covered by the proposed change

AlexeySachkov:
Seems like force-push didn't trigger Travis CI build: I will try closing and re-opening the PR to see if it helps

AlexeySachkov:
I've just enabled GitHub Actions CI - I will try to trigger it by closing and re-opening the PR. This PR should be an NFC so as soon as builds pass we can merge this

AlexeySachkov (approved):
According to testing results all builds passed and testing failures are similar to what I see in other PRs. I'm going to proceed with accepting this PR as it is fairly simple and unlikely to introduce any significant regressions

PR #807, opened 2020-11-02
[SYCL][FPGA] Fix IVDep translation for pointer parameters

AGindinson:
The original implementation of IVDep for kernel closure parameters
(github.com/KhronosGroup/SPIRV-LLVM-Translator/commit/f0e3545e1f0e3545e)
has only resolved array parameters. When the attribute is applied
to a pointer instead, an additional load instruction gets emitted
for each access by the FE compiler. Accesses to different load
results have been treated as different arrays, while still referring
to the same memory. As a consequence, accesses to the same pointer
have been marked into different `!llvm.index.group` metadata nodes -
essentially, this issue is what the new patch aims to resolve.

Additionally, the IR for the test was bumped up to the new LLVM
version.


AGindinson:
@MrSidims, @AlexeySachkov, could you please take a look?

AlexeySachkov:
> github.com/KhronosGroup/SPIRV-LLVM-Translator/commit/f0e3545e1f0e3545e

FYI, the best way to reference other commits on GitHub is `[user|org]/repo@hash`: KhronosGroup/SPIRV-LLVM-Translator@f0e3545e1f0e3545e

AlexeySachkov:
Also, the Travis CI seems to be broken (or even disabled?). Tagging @svenh for awareness. This is not a problem of this PR, but might be a blocker for incoming contributions

lib/SPIRV/SPIRVReader.cpp
```diff
@@ -835,6 +835,9 @@ void SPIRVToLLVM::setLLVMLoopMetadata(const LoopInstType *LM,
           continue;
 
         Value *AccessedPointer = GEP->getPointerOperand();
+        if (auto *LI = dyn_cast<LoadInst>(AccessedPointer))
+          if (Value *LoadOp = LI->getPointerOperand())
```
AlexeySachkov, line 839: Do we really need this as an `if` statement? I mean is it expected to encounter a `load` instruction without its pointer operand?

AGindinson:
I've updated the commit message to better reference the past commit; also moved the test into the right directory.

MrSidims (commented):
I'm no longer a fan of using SPIR-V for FPGA: it feels like you need to implement a second backend just to translate a single loop attribute...

MrSidims (approved):

keryell:
> I'm no longer a fan of using SPIR-V for FPGA: it feels like you need to implement a second backend just to translate a single loop attribute...

This is a trade-off. Probably that if Intel and Xilinx HLS backends were using the ToT Clang/LLVM like any sensible software project would do, it would be useless to go through SPIR-V or handled with dirty IR version downgrading.
If you want to experiment the other way, feel free to contribute to https://github.com/triSYCL/sycl/tree/sycl/unified/next/llvm/lib/SYCL ;-)

AlexeySachkov:
I've just enabled GitHub Actions CI - I will try to trigger it by closing and re-opening the PR

AlexeySotkin (approved):

PR #810, opened 2020-11-06
Update spirv.hpp

MrSidims:
New spirv.hpp is generated from https://github.com/KhronosGroup/SPIRV-Headers
using following PRs:
https://github.com/KhronosGroup/SPIRV-Headers/pull/176
https://github.com/KhronosGroup/SPIRV-Headers/pull/177
with an exception of SPV_INTEL_fpga_loop_fuse and
SPV_INTEL_long_constant_composite extensions definitions that were added manually
in this PR.


lib/SPIRV/libSPIRV/spirv.hpp
```diff
@@ -1,4 +1,3 @@
-// clang-format off
```
svenvh, line 1: @MrSidims if I understand correctly this is a clean file from the SPIRV-Headers project? ("clean" in the sense of the current version + your 2 pending SPIRV-Headers PRs, so no translator-specific changes were made)
If so, @AlexeySachkov perhaps we should add this file to the ignore list for the formatting checks?

MrSidims: Yes, it is. It was generated via ./bin/makeHeaders tool, so I think we may ignore the checks.
There are few extensions missing still, I'll remove WIP when it's ready.

MrSidims:
Locally the header is buildable and LITs are passing.

MrSidims:
I would prefer to merge this PR before https://github.com/KhronosGroup/SPIRV-LLVM-Translator/pull/813 . Also https://github.com/KhronosGroup/SPIRV-LLVM-Translator/pull/813 is now blocked, because the mentioned PRs to SPIR-V Headers need to land first, and also there are 2 Intel extensions that aren't yet upstreamed: SPV_INTEL_optimization_hints and SPV_INTEL_linkonce_odr.

lib/SPIRV/libSPIRV/spirv.hpp
svenvh: Okay. Given that you're removing spirv.hpp in #813 there is no value in updating the checks to ignore the file now, so we can just ignore the formatting issues in this PR.

svenvh (approved):
LGTM

MrSidims:
Updated the header. @AlexeySachkov I suggest to merge this PR.

AlexeySachkov (approved):

PR #811, opened 2020-11-09
[NFC] Fix a static analysis issue

AGindinson:
Fix a potential nullptr dereference issue that static analysis
finds in prod build mode (where assertions are disabled). This
is a follow-up for [KhronosGroup/SPIRV-LLVM-Translator@ee145be].

### DIFF
--- a/lib/SPIRV/SPIRVRegularizeLLVM.cpp
+++ b/lib/SPIRV/SPIRVRegularizeLLVM.cpp
@@ -114,7 +114,8 @@ char SPIRVRegularizeLLVM::ID = 0;
 
 std::string SPIRVRegularizeLLVM::lowerLLVMIntrinsicName(IntrinsicInst *II) {
   Function *IntrinsicFunc = II->getCalledFunction();
-  assert(IntrinsicFunc && "Missing function");
+  if (!IntrinsicFunc)
+    return "";
   std::string FuncName = IntrinsicFunc->getName().str();
   std::replace(FuncName.begin(), FuncName.end(), '.', '_');
   FuncName = "spirv." + FuncName;


lib/SPIRV/SPIRVRegularizeLLVM.cpp
```diff
@@ -114,7 +114,8 @@ char SPIRVRegularizeLLVM::ID = 0;
 
 std::string SPIRVRegularizeLLVM::lowerLLVMIntrinsicName(IntrinsicInst *II) {
   Function *IntrinsicFunc = II->getCalledFunction();
-  assert(IntrinsicFunc && "Missing function");
```
MrSidims, line 117: I tend to think, that this static analysis issue can be ignored (in LLVM you may see a lot of assertions that serves for the purpose of guarding possible nullptr dereference) or this code shall be re-written differently.

Assertions are used to show the assumptions that a programmer is making. For this one the assumption is that `getCalledFunction` never returns null. Here you change this assumption (and hence the logic of this code block), making it to be a valid case. It's still OK. From LLVM point of view it actually could happen (I mean, the function returning nullptr, see: https://llvm.org/doxygen/classllvm_1_1CallBase.html#a83b8c9e572130f403445c0b0f69b837c) in case of indirect function call. But I would suggest not to return an empty string in this case, but to write something like:
```
if (!IntrinsicFunc) {
    Value* InderectCall = II->getCalledValue();
    Value* CalledFun = InderectCall->stripPointerCasts();
    StringRef FunName = CalledFun->getName();
    // Do something with FunName / merge it with FuncName from below
}
```

MrSidims: Also, if you still want to replace this assert, please remove "NFC" from the title.

AlexeySachkov: `getCalledFunction` returns `nullptr` only if it was an indirect call. I'm not sure that it is possible to have an `IntrinsicInst` representing indirect call in LLVM IR, so it is unclear to me why this assertion should be replaced with a runtime check

AGindinson: I agree with the points that getting `nullptr` is neither expected inside this function, nor likely to happen for the reasons described by Alexey. I believe this static analysis issue had best be disregarded.

PR #812, opened 2020-11-10
Add OpModuleProcessed token handling

NikitaRudenkoIntel:
Translator fails to read module generated by spirv-link, because spirv-link adds OpModuleProcessed token which is not supported. This patch fix the problem.

### DIFF
--- a/lib/SPIRV/libSPIRV/SPIRVEntry.h
+++ b/lib/SPIRV/libSPIRV/SPIRVEntry.h
@@ -572,6 +572,23 @@ protected:
   std::string Str;
 };
 
+class SPIRVModuleProcessed : public SPIRVEntryNoId<OpModuleProcessed> {
+  static const SPIRVWord FixedWC = 1;
+
+public:
+  SPIRVModuleProcessed(SPIRVModule *M, const std::string &TheProcess)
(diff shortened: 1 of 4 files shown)


lib/SPIRV/libSPIRV/SPIRVEntry.h
```diff
@@ -560,6 +560,25 @@ class SPIRVMemberName : public SPIRVAnnotation<OpName> {
   std::string Str;
 };
 
+class SPIRVModuleProcessed : public SPIRVAnnotation<OpModuleProcessed> {
+public:
+  static const SPIRVWord FixedWC = 1;
+  // Complete constructor
+  SPIRVModuleProcessed(const SPIRVEntry *TheTarget,
```
AlexeySotkin, line 567: I'm not sure we should inherit from `SPIRVAnnotation` here. 
What should be given as `TheTarget` argument for `SPIRVModuleProcessed` constructor?

NikitaRudenkoIntel: Fixed
I used SPIRVEntryNoId as base

NikitaRudenkoIntel:
Implemented in https://github.com/KhronosGroup/SPIRV-LLVM-Translator/pull/924

PR #813, opened 2020-11-10
Make spirv.hpp be an external dependency

MrSidims:


CMakeLists.txt
```diff
@@ -3,6 +3,29 @@ cmake_minimum_required(VERSION 3.3)
 set (BASE_LLVM_VERSION 12.0.0)
 set(LLVM_SPIRV_VERSION ${BASE_LLVM_VERSION}.0)
 
+include(ExternalProject)
+
+# Download spirv.hpp from the official SPIRV-Headers repository.
+# One can skip this step in case of EXTERNAL_SPIRV_HEADERS_DIR path is set.
+if (NOT DEFINED EXTERNAL_SPIRV_HEADERS_DIR)
+  message(STATUS "SPIR-V Headers placement is not defined. Will try to download
+          spirv.hpp frin github.com")
```
AlexeySotkin, line 12:
```suggestion
  message(STATUS "SPIR-V Headers location is not specified. Will try to download
          spirv.hpp from https://github.com/KhronosGroup/SPIRV-Headers")
```

MrSidims: Thanks, done!

CONTRIBUTING.md
```diff
@@ -36,6 +36,25 @@ Coding Standards]. Compliance of your code is checked automatically using
 Travis CI. See [clang-format] and [clang-tidy] configs for more details about
 coding standards.
 
+## How to add an extension
+
+First of all please make sure you have added a link to the specification for the
+extension in your PR. Then to add definitions of new Op Codes you shall modify
+[SPIR-V Header] which is an external dependency for this project. To do so you
```
AlexeySotkin, line 43:
```suggestion
[spirv.hpp] which is an external dependency for this project. To do so you
```

CONTRIBUTING.md
```diff
@@ -79,5 +98,9 @@ it can be merged:
 [LLVM Coding Standards]: https://llvm.org/docs/CodingStandards.html
 [clang-format]: [.clang-format]
 [clang-tidy]: [.clang-tidy]
+[SPIR-V Header]: https://github.com/KhronosGroup/SPIRV-Headers/blob/master/include/spirv/unified1/spirv.hpp
```
AlexeySotkin, line 101:
```suggestion
[spirv.hpp]: https://github.com/KhronosGroup/SPIRV-Headers/blob/master/include/spirv/unified1/spirv.hpp
```

CONTRIBUTING.md
```diff
@@ -36,6 +36,25 @@ Coding Standards]. Compliance of your code is checked automatically using
 Travis CI. See [clang-format] and [clang-tidy] configs for more details about
 coding standards.
 
+## How to add an extension
+
+First of all please make sure you have added a link to the specification for the
+extension in your PR. Then to add definitions of new Op Codes you shall modify
+[SPIR-V Header] which is an external dependency for this project. To do so you
+would need to add new your definitions in [json grammar file] and rebuild the
```
AlexeySotkin, line 44:
```suggestion
should add new definitions to [json grammar file], rebuild the
```

CONTRIBUTING.md
```diff
@@ -36,6 +36,25 @@ Coding Standards]. Compliance of your code is checked automatically using
...
+header following the instructions in [SPIR-V Headers repository] and push your
```
AlexeySotkin, line 45:
```suggestion
header following the [instructions](https://github.com/KhronosGroup/SPIRV-Headers#generating-headers-from-the-json-grammar-for-the-spir-v-core-instruction-set) in [SPIR-V Headers repository] and push your
```

CONTRIBUTING.md
```diff
@@ -36,6 +36,25 @@ Coding Standards]. Compliance of your code is checked automatically using
...
+changes there. Once they are merged a brand-new spirv.hpp will be downloaded
```
AlexeySotkin, line 46:
```suggestion
changes for review, i.e. make a PR. Once the PR is merged a brand-new spirv.hpp will be downloaded
```

CONTRIBUTING.md
MrSidims: Thanks! Applied.

CMakeLists.txt
```diff
@@ -3,6 +3,29 @@ cmake_minimum_required(VERSION 3.3)
 set (BASE_LLVM_VERSION 12.0.0)
 set(LLVM_SPIRV_VERSION ${BASE_LLVM_VERSION}.0)
 
+include(ExternalProject)
+
+# Download spirv.hpp from the official SPIRV-Headers repository.
+# One can skip this step in case of EXTERNAL_SPIRV_HEADERS_DIR path is set.
+if (NOT DEFINED EXTERNAL_SPIRV_HEADERS_DIR)
```
AlexeySachkov, line 10: I suggest that we re-use functionality provided by LLVM, i.e. we assume that SPIRV-Headers repo is an external project in terms of LLVM monorepo:

To check if it is available we use `LLVM_TOOL_SPIRV_HEADERS_BUILD` variable, which is automatically defined for us in  [llvm/CMakeLists.txt:127-131](https://github.com/llvm-project/llvm/blob/master/llvm/CMakeLists.txt#L127)

To get path to sources we use `LLVM_EXTERNAL_SPIRV_HEADERS_SOURCE_DIR` variable, which is either supplied by user, or if `SPIRV-Headers` project was put into `llvm/projects` subdirectory, then it is automatically set by LLVM in  [llvm/cmake/modules/AddLLVM.cmake:1355-1348](https://github.com/llvm-project/llvm/blob/master/llvm/cmake/modules/AddLLVM.cmake#L1355)

CONTRIBUTING.md
```diff
@@ -36,6 +36,25 @@ Coding Standards]. Compliance of your code is checked automatically using
 Travis CI. See [clang-format] and [clang-tidy] configs for more details about
 coding standards.
 
+## How to add an extension
```
AlexeySachkov, line 39: I think it also makes sense to update build instructions: we probably need to mention that there is a dependency on `SPIRV-Headers` which can be either configured manually or it will be downloaded automatically

CMakeLists.txt
AGindinson: Done. I've ended up checking for `LLVM_EXTERNAL_SPIRV_HEADERS_SOURCE_DIR` definition directly, because that variable can be configured manually to point to an external headers' installation without `LLVM_TOOL_SPIRV_HEADERS_BUILD` being set.

CONTRIBUTING.md
AGindinson: Done.

CMakeLists.txt
```diff
@@ -14,6 +16,35 @@ if (NOT DEFINED LLVM_SPIRV_BUILD_EXTERNAL)
...
+  message(STATUS "SPIR-V Headers location is not specified. Will try to download
+          spirv.hpp from https://github.com/KhronosGroup/SPIRV-Headers into
+          ${LLVM_EXTERNAL_SPIRV_HEADERS_SOURCE_DIR}")
+  ExternalProject_Add(spirv-headers
+    GIT_REPOSITORY    https://github.com/KhronosGroup/SPIRV-Headers.git
+    GIT_TAG           origin/master
+    SOURCE_DIR        ${LLVM_EXTERNAL_SPIRV_HEADERS_SOURCE_DIR}
+    CONFIGURE_COMMAND ""
+    BUILD_COMMAND     ""
+    INSTALL_COMMAND   ""
```
AlexeySachkov, line 39: Am I right that there are nothing to build in `SPIRV-Headers` project? Do we need to leave a comment here explaining why all those configure/build/install commands are empty?

BTW, is it possible to remove those options completely?

CMakeLists.txt
```diff
@@ -14,6 +16,35 @@ if (NOT DEFINED LLVM_SPIRV_BUILD_EXTERNAL)
...
+# Download spirv.hpp from the official SPIRV-Headers repository.
+# One can skip this step by manually setting
+# LLVM_EXTERNAL_SPIRV_HEADERS_SOURCE_DIR path.
+if (NOT DEFINED LLVM_EXTERNAL_SPIRV_HEADERS_SOURCE_DIR)
+  if (LLVM_SPIRV_BUILD_EXTERNAL)
+    set(LLVM_EXTERNAL_SPIRV_HEADERS_SOURCE_DIR
+      "${CMAKE_CURRENT_BINARY_DIR}/SPIRV-Headers")
+  else() # In-tree build
+    set(LLVM_EXTERNAL_SPIRV_HEADERS_SOURCE_DIR
+      "${CMAKE_SOURCE_DIR}/projects")
```
AlexeySachkov: Even though `llvm/projects` directory is listed in [`.gitignore`](https://github.com/llvm/llvm-project/blob/main/llvm/.gitignore) in the upstream LLVM, I'm not sure that we should update it every time we do the build - I suggest downloading headers only to `build` area regardless of in-tree/out-of-tree setup

CONTRIBUTING.md
```diff
@@ -36,6 +36,25 @@ Coding Standards]. Compliance of your code is checked automatically using
...
+should add new definitions to [json grammar file], rebuild the header following
+the [instructions] in [SPIR-V Headers repository] and push your changes for
+review, i.e. make a PR. Once the PR is merged a brand-new spirv.hpp will be
+downloaded during build of the translator.
+
+It's highly recommended to add the definitions to [SPIR-V Headers repository]
+first, but if you don't want to bring it there yet, you can define new Op Codes
+in the [internal SPIR-V header file].
+
+For local testing you can copy spirv.hpp to {PATH}/include/spirv/unified1 and
```
AlexeySachkov, line 53: What is `{PATH}` here?

CONTRIBUTING.md
```diff
@@ -36,6 +36,25 @@ Coding Standards]. Compliance of your code is checked automatically using
...
+modify it. To use it for your local build please specify
+EXTERNAL_SPIRV_HEADERS_DIR cmake variable with your {PATH} to build the
```
AlexeySachkov, line 55: `EXTERNAL_SPIRV_HEADERS_DIR` is not referenced in `CMakeLists.txt` anymore, the doc should be updated

README.md
```diff
@@ -93,6 +93,26 @@ Building clang from sources takes time and resources and it can be avoided:
...
+etc. onto numeric tokens. The official header version is available at
+[KhronosGroup/SPIRV-Headers](https://github.com/KhronosGroup/SPIRV-Headers).
+There are several options for accessing the header file:
+- By default, the header file repository will be downloaded from
+Khronos Group GitHub and put into `<build_dir>/SPIRV-Headers` (for
+out-of-tree builds) or into `llvm/projects` (for in-tree builds).
+- If you are building the translator in-tree, you can manually
+  download the SPIR-V Headers repo into `llvm/projects` - this
+  location will be automatically picked up by the translator build
+  scripts.
```
AlexeySachkov, line 109: Note here: this is rather would be done by upstream LLVM CMake scripts: for (almost) each directory in `projects`, [`add_llvm_subdirectory`](https://github.com/llvm/llvm-project/blob/main/llvm/cmake/modules/AddLLVM.cmake#L1337) is called (through `add_llvm_external_project`), which defines `LLVM_EXTERNAL_project_name_SOURCE_DIR` which we later use.

`project_name` is generated automatically in this case and we should probably require particular name for the headers subdirectory to make it work

README.md
```diff
@@ -93,6 +93,26 @@ Building clang from sources takes time and resources and it can be avoided:
...
+- Any build type can also use an external installation of SPIR-V
+  Headers - if you have the headers downloaded somewhere in your
+  system and want to use that version, simply extend your CMake
+  command with
+  `-DLLVM_EXTERNAL_SPIRV_HEADERS_SOURCE_DIR=</path/to/headers_dir>`.
```
AlexeySachkov, line 114: Technically, there is difference between putting a directory into `llvm/projects` directory and setting `LLVM_EXTERNAL_SPIRV_HEADERS_SOURCE_DIR` yourself - former option likely to enable _build_ of the subproject by default (if there are `CMakeLists.txt`) while the latter requires `LLVM_EXTERNAL_PROJECTS_ENABLE` to be also set in order to build the project.

More details about how `llvm/project` works you can find in documentation to `add_llvm_subdirectory`  I linked above

CMakeLists.txt
AGindinson: Removing these options results in the build system's attempt to use "default" commands for all these steps, which leads build failures precisely because there is nothing to build/install. I'll add a comment to explain this nuance.

README.md
AGindinson: Updated.

CONTRIBUTING.md
AGindinson: Done.

CONTRIBUTING.md
AGindinson: Renamed to be more explicit.

README.md
AGindinson: Updated the doc.

CMakeLists.txt
AGindinson: Done.

CMakeLists.txt
```diff
@@ -14,6 +16,50 @@ if (NOT DEFINED LLVM_SPIRV_BUILD_EXTERNAL)
...
+  ExternalProject_Add(spirv-headers
+    SOURCE_DIR ${LLVM_EXTERNAL_SPIRV_HEADERS_SOURCE_DIR}
+    DOWNLOAD_COMMAND  ""
+    # Explicit specification of empty commands is needed to avoid "default"
+    # configure/build/install commands being run by the build system.
+    CONFIGURE_COMMAND ""
+    BUILD_COMMAND     ""
+    INSTALL_COMMAND   ""
+    STEP_TARGETS      build
+  )
```
AlexeySachkov, line 60: Could you please explain the idea of this branch in more details? Because it seems like I got confused. Particular questions I have:
- why do we need `ExternalProject_Add` here without download command? Do we assume that there is like 100% guarantee that required files are there where we expect them?
- why do we need to do any extra actions for `projects/SPIRV-Headers` path at all? I though that it will be picked up by LLVM automatically

README.md
```diff
@@ -93,6 +93,25 @@ Building clang from sources takes time and resources and it can be avoided:
...
 
+## Configuring SPIR-V Headers
+
+The translator build is dependent on the official Khronos header file
+`spirv.hpp` that maps SPIR-V extensions, decorations, instructions,
+etc. onto numeric tokens. The official header version is available at
+[KhronosGroup/SPIRV-Headers](https://github.com/KhronosGroup/SPIRV-Headers).
+There are several options for accessing the header file:
+- By default, the header file repository will be downloaded from
+Khronos Group GitHub and put into `<build_dir>/SPIRV-Headers`.
```
AlexeySachkov, line 104:
```suggestion
  Khronos Group GitHub and put into `<build_dir>/SPIRV-Headers`.
```

lib/SPIRV/CMakeLists.txt
```diff
@@ -45,12 +45,14 @@ add_llvm_library(LLVMSPIRVLib
...
   DEPENDS
     intrinsics_gen
+    spirv-headers
 )
 
 target_include_directories(LLVMSPIRVLib
   PRIVATE
     ${LLVM_INCLUDE_DIRS}
     ${LLVM_SPIRV_INCLUDE_DIRS}
+    ${LLVM_EXTERNAL_SPIRV_HEADERS_SOURCE_DIR}/include/spirv/unified1
```
AlexeySachkov, line 55: From what I see, `SPIRV-Headers` are [defined](https://github.com/KhronosGroup/SPIRV-Headers/blob/master/CMakeLists.txt#L72) as [interface library](https://cmake.org/cmake/help/v3.4/command/add_library.html#interface-libraries), which simplifies its usage.

 If I understand correctly (educated myself using [this](http://mariobadr.com/creating-a-header-only-library-with-cmake.html) material), the only thing we need is to "link" the translator with SPIR-V headers:
 ```
target_link_libraries(LLVMSPIRVLib SPIRV-Headers)
# or probably add SPIRV-Headers as an argument to add_llvm_library above
 ```
 
 And since `SPIRV-Headers` is an interface library, we will inherit `target_include_directories` from it, which are defined [here](https://github.com/KhronosGroup/SPIRV-Headers/blob/master/CMakeLists.txt#L73).
 
 I don't have experience with interface libraries, but if we would be able to simplify this code, that would be great

CMakeLists.txt
AGindinson: > Why do we need ExternalProject_Add here without download command? Do we assume that there is like 100% guarantee that required files are there where we expect them?

What this command achieves is basically defining the `spirv-headers` target, which is used as a dependency for `LLVMSPIRVLib` from another branch. Even using the target_link_libraries approach from your [other comment](https://github.com/KhronosGroup/SPIRV-LLVM-Translator/pull/813/commits/92911948587adeac713ada01953080bc2dfdb47d#r662430738) would hardly resolve the need for defining this target - the download process still has to happen strictly before the `LLVMSPIRVLib` build/linkage.

> Why do we need to do any extra actions for projects/SPIRV-Headers path at all? I though that it will be picked up by LLVM automatically

If you're talking about the subbranch with `LLVM_TOOL_SPIRV_HEADERS_BUILD` defined, then unfortunately, LLVM does not define the path automatically. And there's still this point with the `spirv-headers` target requiring definition.

AlexeySachkov: Ok, let it be that it requires re-definition, but what if there are no headers in expected location? Should we add some kind of check (like to at least check if directory exists or if there is a `CMakeLists.txt`) and report an error if there are no headers in expected location?

AGindinson: That's a fair point. Essentially, my logic has been that if `LLVM_TOOL_SPIRV_HEADERS_BUILD` has been defined, then the `CMakeLists.txt` presence has [already been checked](https://github.com/llvm/llvm-project/blob/main/llvm/cmake/modules/AddLLVM.cmake#L1350-L1354) by `add_llvm_subdirectory`. I agree that a check for the assumption of the `llvm/projects` location is still needed, and a check for `LLVM_SPIRV_BUILD_EXTERNAL OFF` also wouldn't hurt in this context.

CMakeLists.txt
```diff
@@ -14,6 +16,50 @@ if (NOT DEFINED LLVM_SPIRV_BUILD_EXTERNAL)
...
+if(NOT DEFINED LLVM_TOOL_SPIRV_HEADERS_BUILD AND
+   NOT DEFINED LLVM_EXTERNAL_SPIRV_HEADERS_SOURCE_DIR)
+  set(LLVM_EXTERNAL_SPIRV_HEADERS_SOURCE_DIR
+    "${CMAKE_CURRENT_BINARY_DIR}/SPIRV-Headers")
+  message(STATUS "SPIR-V Headers location is not specified. Will try to download
+          spirv.hpp from https://github.com/KhronosGroup/SPIRV-Headers into
+          ${LLVM_EXTERNAL_SPIRV_HEADERS_SOURCE_DIR}")
+  ExternalProject_Add(spirv-headers
+    GIT_REPOSITORY    https://github.com/KhronosGroup/SPIRV-Headers.git
+    GIT_TAG           origin/master
```
svenvh, line 31: Relying on the `master` branch assumes that the translator will always build with the latest upstream headers.  I'm not sure if that's always the case?  I wonder if we should instead rely on a known good revision of the spirv-headers repo, and maintain a SHA-1 somewhere (similar to what [SPIRV-Tools](https://github.com/KhronosGroup/SPIRV-Tools/blob/a95bc460f924a9541dc3a85745dcb15140c8fcf3/DEPS#L9) does with its `DEPS` file)?

svenvh (commented):
Thanks for working on this!  Mostly looking fine, but I've raised a question about how we should maintain the dependency.

CMakeLists.txt
AGindinson: That's a good point, we could switch to using a hash. This should also include a mention in the docs for implementing extensions, i.e. a note about updating the hash after the developer's patch has been accepted into SPIR-V Headers. In a way, this could solve the maintenance problem, as the newer hashes of the SPIR-V Headers will have be tested with the translator sources by the extension developers, who could solve the unexpected build failures alongside their implementations.

lib/SPIRV/CMakeLists.txt
AGindinson: At last I have succeeded in implementing this interface library approach locally. The only issue I'm facing with finalizing it is that SPIRV-Headers exports the `include/` directory, and `spirv.hpp` lies in `include/spirv/unified1`. Using triangle brackets for `spirv.hpp` inclusion within the translator sources doesn't seem to work, so one of the following edits is required:
1. Export `include/spirv/unified1` as well (needs a patch to the SPIRV-Headers repo)
2. Use `#include "spirv/unified1/spirv.hpp"` in translator sources.
@svenvh, would it be acceptable if I went with **2** as a temporary solution and re-designed this in favor of option **1** as a follow-up?

svenvh: > @svenvh, would it be acceptable if I went with 2 as a temporary solution and re-designed this in favor of option 1 as a follow-up?

No objections from me.  I probably wouldn't even mind if 2 becomes a permanent solution, as it makes it clearer which spirv.hpp we're actually including.

CMakeLists.txt
MrSidims: Agree

lib/SPIRV/CMakeLists.txt
AGindinson: There are exporting issues for the `llvm/projects/SPIRV-Headers` approach when using the header-only INTERFACE "linking". I wasn't able to resolve it without touching LLVM/SPIRV-Headers build scripts, so I've decided to go with the earlier directory inclusion approach, while retaining configure-time download and adding a TODO for switching to the INTERFACE library approach.

AlexeySachkov (approved):
LGTM,

@AGindinson, GitHub says that there are conflicts - could you please resolve them?

AGindinson:
> GitHub says that there are conflicts - could you please resolve them?

Done; also squashed the commits into one.

.github/workflows/check-in-tree-build.yml
```diff
@@ -66,6 +66,12 @@ jobs:
           repository: llvm/llvm-project
           ref: main
           path: llvm-project
+      - name:  Checkout SPIR-V Headers
```
svenvh, line 69: Is this needed actually, or would it be done automatically by cmake for in-tree builds?

CMakeLists.txt
svenvh: We will need to have a single point where we maintain that hash, because both CMakeLists.txt and github CI need it.  Unfortunately I do not have a good suggestion how to do so.

.github/workflows/check-in-tree-build.yml
AGindinson: The idea behind the manual installation for GH Actions has originally come from @AlexeySachkov, in particular, aiming to minimize the risk of proxy issues during configure time (as different workflow steps are being performed in separate, newly configured shells). On the other hand, we end up _not_ testing the configure-time download in CI, yet I'm not sure which workflow to base such testing off - creating a separate check sounds like an overkill to me.

CMakeLists.txt
AGindinson: Fair point; I'll try and come up with something.

.github/workflows/check-in-tree-build.yml
svenvh: I believe these in-tree actions are already testing the configure-time download anyway, because you do not specify any SPIRV-Headers path to cmake for in-tree builds currently.  So the checkout done here is not used, unless I am missing something?

For example the [Windows (Release) CI log](https://github.com/KhronosGroup/SPIRV-LLVM-Translator/pull/813/checks?check_run_id=3054935392)  from your latest run seems to confirm my suspicion:
```
 -- SPIR-V Headers location is not specified. Will try to download
          spirv.hpp from https://github.com/KhronosGroup/SPIRV-Headers into
          D:/a/SPIRV-LLVM-Translator/SPIRV-LLVM-Translator/build/projects/SPIRV-LLVM-Translator/SPIRV-Headers
Microsoft (R) Build Engine version 16.10.2+857e5a733 for .NET Framework
Copyright (C) Microsoft Corporation. All rights reserved.

  Checking Build System
  Creating directories for 'spirv-headers-populate'
  Performing download step (git clone) for 'spirv-headers-populate'
  Cloning into 'SPIRV-Headers'...
```

I don't have a strong preference if we do the download during configure-time or separately upfront; but I'd rather not do extra clones that aren't used. ;-)

AGindinson: The in-tree build part seems to be a bug - I was intending to make it a non-download workflow as well. I'll look into this shortly; meanwhile, the clang-format action should actually suffice for testing the configure-time download.

CMakeLists.txt
AGindinson: Done; switched to using a dedicated `.conf` file to read from both CMake and YML scrupts.

.github/workflows/check-in-tree-build.yml
AGindinson: Never mind the Windows & MacOS thing - as pointed out by @AlexeySachkov, it was just a matter of updating the corresponding workflows. Fixed now.

AlexeySachkov: Let's compare timings now to see if separate checkout through corresponding GitHub Actions is more efficient that checkout through CMake or not. We had 2:02 for configure and let's say 0 for checkout

AlexeySachkov: And now we have 2:34 + 0:04. It is hard to compare performance on Windows on some virtual HW in the cloud :)

AGindinson: Yep, I'm also suspecting that we can't hope for consistency with a vastly loaded cloud server.

svenvh (approved):

.github/workflows/check-in-tree-build.yml
```diff
@@ -70,6 +70,17 @@ jobs:
         uses: actions/checkout@v2
         with:
           path: llvm-project/llvm/projects/SPIRV-LLVM-Translator
+      - name:  Get tag for SPIR-V Headers
+        uses: juliangruber/read-file-action@v1
```
AlexeySotkin, line 74: Are we ok with using an [action](https://github.com/marketplace/actions/read-file) which is not certified by GitHub? Do we really need an extra step to read the file?

AGindinson: Unfortunately, I wasn't able to find an alternative that would be as concise - retaining the tag in the separate file felt like the primary objective. Naturally, I'm open to any suggestions.

AGindinson: UPD: Thanks for the suggestion, performed the switch in https://github.com/KhronosGroup/SPIRV-LLVM-Translator/commit/9ea401169f95972c7457e3e3fe5ef647cfc6ca8b.

AlexeySotkin (approved):

AlexeySotkin:
LGTM, Thanks!

PR #818, opened 2020-11-19
[CI] Enable GitHub Actions for builds and tests

### LINKED ISSUE #726: Add Windows to the CI
The build of the translator regularly breaks on Windows due to the lack of testing. This makes it hard for projects embedding the translator to support Windows.

AlexeySachkov:
There are two main reasons this PR was created:
1. We have a problem with Travis CI and it has been basically disabled for almost a two weeks for now
2. From my point of view, GitHub Actions provide more features to make our CI jobs even better

(2) is being discussed in #716 and this PR can be considered as a temporary, i.e. if we don't like how GitHub Actions work we can always fallback to Travis CI once it is fixed.

Anyway, this PR features the following:
- two separate workflows for in-tree and out-of-tree builds
- support for master and release branches as well as for nightly builds
- in-tree build is enabled on Linux, Windows (this tries to address #726, at least partially) and macOS
- these workflows are not launched for documentation-only changes

I added links to GitHub Actions documentation and tried to add comments to all potentially confusing places.

There are for sure more thing that could be improved, like adding `-Werror` configuration or adding GitHub Actions badges to README page, but this can be done as follow-up PRs, I think: my main intent here was to unblock commits to the translator by enabling the CI back.

A few obvious downsides:
Each job is now displayed as a separate check, which seems to be an overkill, but there is [no workaround](https://github.community/t/hide-pr-checks-for-skipped-jobs/124881/4) for this at the moment.
Also, we can't control the order in which jobs are launched, which means list of checks on the PR page might be different each time. "Checks" page is better: the order of jobs is fixed there (it follows their appearance in workflow file), but the ordering of workflows is not fixed, which means that you still can't use muscle memory to quickly look at in-tree build results

### DIFF
--- a/.github/workflows/check-in-tree-build.yml
+++ b/.github/workflows/check-in-tree-build.yml
@@ -1,5 +1,5 @@
-# This workflow is intended to check if PR doesn't introduce any regressions
-# into in-tree build in various configurations.
+# This workflow is intended to check that in-tree build of the translator is
+# healthy and all tests pass. It is used in pre-commits and nightly builds.
 #
 # Documentation for GitHub Actions:
 # [workflow-syntax]: https://docs.github.com/en/free-pro-team@latest/actions/reference/workflow-syntax-for-github-actions
@@ -50,13 +50,13 @@ jobs:
           echo "deb https://apt.llvm.org/bionic/ llvm-toolchain-bionic main" | sudo tee -a /etc/apt/sources.list
           echo "deb https://packages.lunarg.com/vulkan bionic main" | sudo tee -a /etc/apt/sources.list
           sudo apt-get update
-          # Linux systems in GitHub Actions already has pre-installed clang is
-          # not needed for the translator as we rely on newer version
+          # Linux systems in GitHub Actions already have a pre-installed clang
+          # that is not needed for the translator as we rely on newer version
           sudo apt-get --purge remove clang-8 clang-9
           sudo apt-get -yq --no-install-suggests --no-install-recommends install \
             clang-${{ env.LLVM_VERSION }} \
             spirv-tools
-          sudo update-alternatives --install /usr/bin/clang clang /usr/bin/clang-12 1000
+          sudo update-alternatives --install /usr/bin/clang clang /usr/bin/clang-${{ env.LLVM_VERSION }} 1000
       - name: Checkout LLVM sources
         uses: actions/checkout@v2
         with:
--- a/.github/workflows/check-in-tree-build.yml
+++ b/.github/workflows/check-in-tree-build.yml
@@ -1,0 +1,175 @@
+# This workflow is intended to check if PR doesn't introduce any regressions
+# into in-tree build in various configurations.
+#
+# Documentation for GitHub Actions:
+# [workflow-syntax]: https://docs.github.com/en/free-pro-team@latest/actions/reference/workflow-syntax-for-github-actions
+# [context-and-expression-syntax]: https://docs.github.com/en/free-pro-team@latest/actions/reference/context-and-expression-syntax-for-github-actions
+name: In-tree build & tests
+
+on:
+  push:
+    branches:
+      - master
+      - llvm_release_*
+    paths-ignore: # no need to check build for documentation changes
+      - 'docs/**'
+      - '**.md'
+  pull_request:
+    branches:
+      - master
+      - llvm_release_*
+    paths-ignore: # no need to check build for documentation changes
+      - 'docs/**'
+      - '**.md'
+  schedule:
+    # Ideally, we might want to simplify our regular nightly build as we
+    # probably don't need every configuration to be built every day: most of
+    # them are only necessary in pre-commits to avoid breakages
+    - cron: 0 0 * * *
+
+env:
+  LLVM_VERSION: 12
+
+jobs:
+  build_and_test_linux:
+    name: Linux
+    strategy:
+      matrix:
+        build_type: [Release, Debug]
+        shared_libs: [NoSharedLibs]
+        include:
+          - build_type: Release
+            shared_libs: EnableSharedLibs
+      fail-fast: false
+    runs-on: ubuntu-18.04
+    steps:
+      - name: Install dependencies
+        run: |
+          curl -L "https://apt.llvm.org/llvm-snapshot.gpg.key" | sudo apt-key add -
+          curl -L "https://packages.lunarg.com/lunarg-signing-key-pub.asc" | sudo apt-key add -
+          echo "deb https://apt.llvm.org/bionic/ llvm-toolchain-bionic main" | sudo tee -a /etc/apt/sources.list
+          echo "deb https://packages.lunarg.com/vulkan bionic main" | sudo tee -a /etc/apt/sources.list
+          sudo apt-get update
+          # Linux systems in GitHub Actions already has pre-installed clang is
+          # not needed for the translator as we rely on newer version
+          sudo apt-get --purge remove clang-8 clang-9
+          sudo apt-get -yq --no-install-suggests --no-install-recommends install \
+            clang-${{ env.LLVM_VERSION }} \
+            spirv-tools
+          sudo update-alternatives --install /usr/bin/clang clang /usr/bin/clang-12 1000
+      - name: Checkout LLVM sources
+        uses: actions/checkout@v2
+        with:
+          repository: llvm/llvm-project
+          ref: master
+          path: llvm-project
+      - name:  Checkout the translator sources
+        uses: actions/checkout@v2
+        with:
+          path: llvm-project/llvm/projects/SPIRV-LLVM-Translator
+      - name: Configure
+        run: |
+          mkdir build && cd build
+          # ON/OFF specifically weren't used as a values for shared_libs matrix
+          # field to improve usability of PR page: instead of (Release, ON) a
+          # job will be displayed as (Release, EnableSharedLibs)
+          SHARED_LIBS=OFF
+          if [[ "${{ matrix.shared_libs }}" == "EnableSharedLibs" ]]; then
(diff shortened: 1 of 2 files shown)


.github/workflows/check-in-tree-build.yml
```diff
@@ -0,0 +1,175 @@
...
+      - master
+      - llvm_release_*
+    paths-ignore: # no need to check build for documentation changes
+      - 'docs/**'
+      - '**.md'
+  schedule:
+    # Ideally, we might want to simplify our regular nightly build as we
+    # probably don't need every configuration to be built every day: most of
+    # them are only necessary in pre-commits to avoid breakages
+    - cron: 0 0 * * *
```
svenvh, line 28: Will this mean it will run nightly too?  That contradicts the comment at the top of this file a bit, which says it's only for testing a PR.

.github/workflows/check-in-tree-build.yml
```diff
@@ -0,0 +1,175 @@
...
+    runs-on: ubuntu-18.04
+    steps:
+      - name: Install dependencies
+        run: |
+          curl -L "https://apt.llvm.org/llvm-snapshot.gpg.key" | sudo apt-key add -
+          curl -L "https://packages.lunarg.com/lunarg-signing-key-pub.asc" | sudo apt-key add -
+          echo "deb https://apt.llvm.org/bionic/ llvm-toolchain-bionic main" | sudo tee -a /etc/apt/sources.list
+          echo "deb https://packages.lunarg.com/vulkan bionic main" | sudo tee -a /etc/apt/sources.list
+          sudo apt-get update
+          # Linux systems in GitHub Actions already has pre-installed clang is
```
svenvh, line 53:
```suggestion
          # Linux systems in GitHub Actions already have a pre-installed clang that is
```

svenvh (commented):
I'm not an expert on GitHub Actions, but it looks good to me and it appears to work for the PR.

While travis is still unavailable for this repository, I support the idea of enabling this soon, such that new and pending PRs can be subject to some automated testing again.

.github/workflows/check-in-tree-build.yml
```diff
@@ -0,0 +1,175 @@
...
+            -DBUILD_SHARED_LIBS=${SHARED_LIBS} \
+            -DLLVM_TARGETS_TO_BUILD="X86" \
+            -DSPIRV_SKIP_CLANG_BUILD=ON \
+            -DSPIRV_SKIP_DEBUG_INFO_TESTS=ON \
+            -DLLVM_LIT_ARGS="-sv --no-progress-bar" \
+            -G "Unix Makefiles"
+      - name: Build
+        run: |
+          cd build
+          make llvm-spirv -j2
```
AlexeySotkin, line 91: I think we can skip this step. Build should be done automatically for the `check-llvm-spirv` target, right?

.github/workflows/check-in-tree-build.yml
```diff
@@ -0,0 +1,175 @@
...
+          echo "deb https://apt.llvm.org/bionic/ llvm-toolchain-bionic main" | sudo tee -a /etc/apt/sources.list
+          echo "deb https://packages.lunarg.com/vulkan bionic main" | sudo tee -a /etc/apt/sources.list
+          sudo apt-get update
+          # Linux systems in GitHub Actions already has pre-installed clang is
+          # not needed for the translator as we rely on newer version
+          sudo apt-get --purge remove clang-8 clang-9
+          sudo apt-get -yq --no-install-suggests --no-install-recommends install \
+            clang-${{ env.LLVM_VERSION }} \
+            spirv-tools
+          sudo update-alternatives --install /usr/bin/clang clang /usr/bin/clang-12 1000
```
AlexeySotkin, line 59:
```suggestion
          sudo update-alternatives --install /usr/bin/clang clang /usr/bin/clang-${{ env.LLVM_VERSION }} 1000
```

.github/workflows/check-out-of-tree-build.yml
```diff
@@ -0,0 +1,72 @@
...
+          sudo apt-get update
+          # Linux systems in GitHub Actions already has pre-installed clang is
+          # not needed for the translator as we rely on newer version
+          sudo apt-get --purge remove clang-8 clang-9
+          sudo apt-get -yq --no-install-suggests --no-install-recommends install \
+            clang-${{ env.LLVM_VERSION }} \
+            llvm-${{ env.LLVM_VERSION }}-dev \
+            llvm-${{ env.LLVM_VERSION }}-tools \
+            spirv-tools
+          sudo update-alternatives --install /usr/bin/clang clang /usr/bin/clang-12 1000
```
AlexeySotkin, line 54:
```suggestion
          sudo update-alternatives --install /usr/bin/clang clang /usr/bin/clang-${{ env.LLVM_VERSION }} 1000
```

.github/workflows/check-in-tree-build.yml
```diff
@@ -0,0 +1,175 @@
...
+jobs:
+  build_and_test_linux:
+    name: Linux
+    strategy:
+      matrix:
+        build_type: [Release, Debug]
+        shared_libs: [NoSharedLibs]
+        include:
+          - build_type: Release
+            shared_libs: EnableSharedLibs
```
AlexeySotkin, line 42: Could you clarify why we don't test debug configuration with enabled shared libs?

.github/workflows/check-out-of-tree-build.yml
```diff
@@ -0,0 +1,72 @@
...
+    runs-on: ubuntu-18.04
+    steps:
+      - name: Install dependencies
+        run: |
+          curl -L "https://apt.llvm.org/llvm-snapshot.gpg.key" | sudo apt-key add -
+          curl -L "https://packages.lunarg.com/lunarg-signing-key-pub.asc" | sudo apt-key add -
+          echo "deb https://apt.llvm.org/bionic/ llvm-toolchain-bionic main" | sudo tee -a /etc/apt/sources.list
+          echo "deb https://packages.lunarg.com/vulkan bionic main" | sudo tee -a /etc/apt/sources.list
+          sudo apt-get update
+          # Linux systems in GitHub Actions already has pre-installed clang is
```
AlexeySotkin, line 46: The same comment as @svenvh left for in-tree build config
```suggestion
          # Linux systems in GitHub Actions already have a pre-installed clang that is
```

.github/workflows/check-out-of-tree-build.yml
```diff
@@ -0,0 +1,72 @@
...
+
+env:
+  LLVM_VERSION: 12
+
+jobs:
+  build_and_test:
+    name: Linux
+    strategy:
+      matrix:
+        build_type: [Release, Debug]
```
AlexeySotkin, line 35: Do we link against static or shared LLVM libraries here? In Travis CI we had both options. Should we have the same behavior in GH Actions or having just one (shared libs for example) is enough for us?

.github/workflows/check-in-tree-build.yml
AlexeySachkov: > Build should be done automatically for the `check-llvm-spirv` target, right?

Yes, it would be done. However, I prefer to leave it as-is, because it allows you to understand whether CI job is failed because the project isn't buildable or there is some problem with tests. Note: in Travis CI config we also have two separate targets for build and tests

.github/workflows/check-in-tree-build.yml
AlexeySachkov: We can easily enable those - I just though that it is not that important to have such configuration

.github/workflows/check-out-of-tree-build.yml
AlexeySachkov: > Do we link against static or shared LLVM libraries here? In Travis CI we had both options. 

Let me take a more detailed look into this: the exact modes which are available for us in out-of-tree configuration are actually limited by LLVM distribution, i.e. if there were no shared libraries in it, we won't be able to test such setup. But I guess we have everything in place if we had both configurations in Travis CI

.github/workflows/check-in-tree-build.yml
AlexeySachkov: I've updated the description in b360b08f00ea63d7a488c22037fe18ff7b5ca446

.github/workflows/check-in-tree-build.yml
AlexeySachkov: Fixed in b360b08f00ea63d7a488c22037fe18ff7b5ca446

.github/workflows/check-in-tree-build.yml
AlexeySachkov: Fixed in b360b08f00ea63d7a488c22037fe18ff7b5ca446

.github/workflows/check-out-of-tree-build.yml
AlexeySachkov: Fixed in b360b08f00ea63d7a488c22037fe18ff7b5ca446

.github/workflows/check-out-of-tree-build.yml
AlexeySachkov: Fixed in b360b08f00ea63d7a488c22037fe18ff7b5ca446

.github/workflows/check-out-of-tree-build.yml
AlexeySachkov: > Let me take a more detailed look into this

But I suggest that we merge this earlier and enable more configurations later as it might take a few days for me to find time to look into this

AlexeySachkov:
BTW, Travis CI seem to be fixed: https://travis-ci.com/github/KhronosGroup/SPIRV-LLVM-Translator

AlexeySachkov:
It seems that Travis CI went offline again - I will go ahead and merge this PR to enable at least some CI back

PR #820, opened 2020-11-20
Add intel::nofusion loop attribute translation

MrSidims:
The new loop control bit NoFusionINTEL is enabled by the FPGALoopControlsINTEL
capability. If this bit is set, it indicates that the loop should not be fused
with any adjacent loop.

Spec:
https://github.com/KhronosGroup/SPIRV-Registry/pull/85/


MrSidims:
- [ ] Propagate the bit to SPIR-V Headers

AlexeySotkin (approved):

PR #823, opened 2020-11-23
Support atomic ordering for LLVM load/store instructions

AlexeySotkin:
LLVM IR `store atomic` and `load atomic` instructions are now lowered
to the corresponding SPIR-V atomics.

Signed-off-by: Andrew Savonichev <andrew.savonichev@intel.com>

### DIFF
--- a/test/atomic-load-store.ll
+++ b/test/atomic-load-store.ll
@@ -1,0 +1,35 @@
+; RUN: llvm-as < %s -o %t.bc
+; RUN: llvm-spirv %t.bc -o %t.spv
+; RUN: spirv-val %t.spv
+; RUN: llvm-spirv -to-text %t.spv -o - | FileCheck %s
+
+; CHECK-DAG: Constant [[#]] [[#Relaxed:]] 0
+; CHECK-DAG: Constant [[#]] [[#DeviceScope:]] 1
+; CHECK-DAG: Constant [[#]] [[#Acquire:]] 2
+; CHECK-DAG: Constant [[#]] [[#Release:]] 4
+; CHECK-DAG: Constant [[#]] [[#SequentiallyConsistent:]] 16
+
+target datalayout = "e-i64:64-v16:16-v24:32-v32:32-v48:64-v96:128-v192:256-v256:256-v512:512-v1024:1024"
+target triple = "spir64"
+
+; Function Attrs: nounwind
+define dso_local spir_func void @test() {
+entry:
+; CHECK: Variable [[#]] [[#PTR:]]
+  %0 = alloca i32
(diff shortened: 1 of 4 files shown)


test/atomic-load-store.ll
```diff
@@ -0,0 +1,35 @@
...
+  %0 = alloca i32
+
+; CHECK: AtomicStore [[#PTR]] [[#DeviceScope]] [[#Relaxed]] [[#]]
+  store atomic i32 0, i32* %0 monotonic, align 4
+; CHECK: AtomicStore [[#PTR]] [[#DeviceScope]] [[#Release]] [[#]]
+  store atomic i32 0, i32* %0 release, align 4
+; CHECK: AtomicStore [[#PTR]] [[#DeviceScope]] [[#SequentiallyConsistent]] [[#]]
+  store atomic i32 0, i32* %0 seq_cst, align 4
+
+; CHECK: AtomicLoad 6 {{[0-9]+}} [[#PTR]] [[#DeviceScope]] [[#Relaxed]]
```
MrSidims, line 28: I suggest to either mask result type with `{{[0-9]+}}` (as it was done for result id) or actually check it.

test/atomic-load-store.ll
```diff
@@ -0,0 +1,35 @@
+; RUN: llvm-as < %s -o %t.bc
+; RUN: llvm-spirv %t.bc -o %t.spv
+; RUN: spirv-val %t.spv
+; RUN: llvm-spirv -to-text %t.spv -o - | FileCheck %s
```
MrSidims, line 4: I see, that purpose of this patch is to add a missing functionality in forward translation, but does it make sense to also add a test for a reversed translation to double check it?

MrSidims (commented):
Functionally LGTM, just few test related questions.

test/atomic-load-store.ll
AlexeySotkin: The reverse translation is covered by [test/transcoding/atomics.spt](https://github.com/KhronosGroup/SPIRV-LLVM-Translator/blob/master/test/transcoding/atomics.spt#L44-L45)

test/atomic-load-store.ll
AlexeySotkin: Fixed in 2fbad5e

test/atomic-load-store.ll
AlexeySotkin: Though test/transcoding/atomics.spt can be extended with --spirv-target-env="SPV-IR", which I have done in 4104901

MrSidims (approved):

PR #793, opened 2020-10-19
Translate llvm.fmuladd to fma SPIRV intrinsic

fadeeval:
Translate llvm.fmuladd to fma SPIRV intrinsic

Signed-off-by: Aleksander Fadeev <aleksander.fadeev@intel.com>

PR #824, opened 2020-11-23
Add possibility to lower llvm.fmuladd into mad from OpenCL extinst

AlexeySachkov:
This is a PR #793, which I had to re-create in order to enable Travis CI check (I hope it will work)


lib/SPIRV/SPIRVWriter.cpp
```diff
@@ -2231,6 +2231,22 @@ SPIRVValue *LLVMToSPIRV::transIntrinsicInst(IntrinsicInst *II,
     // For llvm.fmuladd.* fusion is not guaranteed. If a fused multiply-add
     // is required the corresponding llvm.fma.* intrinsic function should be
     // used instead.
+    if (!checkTypeForSPIRVExtendedInstLowering(II, BM))
```
AlexeySotkin, line 2234: There is no handling for `fmuladd` in `checkTypeForSPIRVExtendedInstLowering`, it will always return `true`. Should we extend `checkTypeForSPIRVExtendedInstLowering`?

AlexeySachkov: I don't think that this is necessary, because `llvm.fmuladd.*` is only defined for floating-point scalars, which perfectly matches requirements of `fma` extinst. We don't even need to check for vector length, because `llvm.fmuladd.*` is not defined for vectors

lib/SPIRV/SPIRVWriter.cpp
```diff
@@ -2231,6 +2231,22 @@ SPIRVValue *LLVMToSPIRV::transIntrinsicInst(IntrinsicInst *II,
...
+      break;
+
+    // If allowed, let's replace llvm.fmuladd.* with mad from OpenCL extended
+    // instruction set, as it has the same semantic for FULL_PROFILE OpenCL
+    // devices (and for EMBEDDED_PROFILE precision is implementaiton-defined).
```
mlychkov, line 2239:
```suggestion
    // devices (and for EMBEDDED_PROFILE precision is implementation-defined).
```

tools/llvm-spirv/llvm-spirv.cpp
```diff
@@ -587,6 +593,17 @@ int main(int Ac, char **Av) {
     }
   }
 
+  if (SPIRVReplaceLLVMFmulAddWithOpenCLMad.getNumOccurrences() != 0) {
+    if (IsReverse) {
+      errs() << "Note: --spirv-replace-fmuladd-with-ocl-mad option ignored as "
+                "it only "
+                "affects translation from LLVM IR to SPIR-V";
```
mlychkov, line 600:
```suggestion
                "it only affects translation from LLVM IR to SPIR-V";
```

tools/llvm-spirv/llvm-spirv.cpp
```diff
@@ -202,6 +202,12 @@ static cl::opt<SPIRV::DebugInfoEIS> DebugEIS(
                    "extended instruction set. This version of SPIR-V debug "
                    "info format is compatible with the SPIRV-Tools")));
 
+static cl::opt<bool> SPIRVReplaceLLVMFmulAddWithOpenCLMad(
+    "spirv-replace-fmuladd-with-ocl-mad",
+    cl::desc("Allow replacement of llvm.fmuladd.* intrinsic with OpenCL mad "
+             "instruction from OpenCL extended instruction set"),
+    cl::init(true));
```
AGindinson, line 209: How likely is it that similar user-requested replacements will be added for other intrinsics? If such plans (or such bits of code) exist already, would a single "joint" option like `-spirv-replace-intrinsics-with-ocl-ext` be preferable?

tools/llvm-spirv/llvm-spirv.cpp
```diff
@@ -202,6 +202,12 @@ static cl::opt<SPIRV::DebugInfoEIS> DebugEIS(
                    "extended instruction set. This version of SPIR-V debug "
                    "info format is compatible with the SPIRV-Tools")));
 
+static cl::opt<bool> SPIRVReplaceLLVMFmulAddWithOpenCLMad(
```
AGindinson, line 205: Is there a reason to make it a `bool`-accepting option instead of a simple flag?
(I personally don't think it matters much, but just for the sake of consistency)

lib/SPIRV/SPIRVWriter.cpp
AGindinson: Or even:
```suggestion
    // devices (implementation-defined for EMBEDDED_PROFILE precision ).
```

tools/llvm-spirv/llvm-spirv.cpp
AlexeySachkov: > How likely is it that similar user-requested replacements will be added for other intrinsics?

This patch is more like a workaround for the lack of fine-grained fp-contraction support in SPIR-V spec, comparing it to OpenCL spec and LLVM IR representation. So, in general I wouldn't expect this to happen often, but still there might be a few more similar situations (right now I can't thing of any)

> If such plans (or such bits of code) exist already, would a single "joint" option like `-spirv-replace-intrinsics-with-ocl-ext` be preferable?

Since this is more like a workaround and I'm not aware of any similar things I guess it would be better to go with this particular option for `fmuladd`

tools/llvm-spirv/llvm-spirv.cpp
AlexeySachkov: > Is there a reason to make it a `bool`-accepting option instead of a simple flag?

No particular reason from command line interface point of view, but from technical point of view I need to take a more careful look at LLVM's CommandLine library to see if it is possible

AlexeySachkov: Looks like [`boolOrDefault`](https://llvm.org/docs/CommandLine.html#boolean-flags) is what I'm looking for here

AlexeySachkov: Ok, I was too quick with `boolOrDefault`: I don't need it, because I don't really care about whether the option was passed or not, but I care about its value.

Also, CommandLine library [supports](https://llvm.org/docs/CommandLine.html#boolean-arguments) flags through boolean options, i.e. `-spirv-replace-fmuladd-with-ocl-mad` is equivalent to `-spirv-replace-fmuladd-with-ocl-mad=true`.

From my point of view the current way is good enough: most likely no one will ever use that switch and I don't want to overcomplicate the code just to get a `-no-spirv-replace-fmuladd-with-ocl-mad` alias (moreover, due to CommandLine library it would be really confusing to be able to use `-no-spirv-replace-fmuladd-with-ocl-mad=false`)

include/LLVMSPIRVOpts.h
```diff
@@ -192,6 +200,10 @@ class TranslatorOpts {
   bool AllowExtraDIExpressions = false;
 
   DebugInfoEIS DebugInfoVersion = DebugInfoEIS::OpenCL_DebugInfo_100;
+
+  // Controls whether llvm.fmuladd.* should be replace with mad from OpenCL
```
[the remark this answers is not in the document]
AlexeySachkov: Fixed in 05d38e03deb11d274068805048ddbab164455924, thanks

lib/SPIRV/SPIRVWriter.cpp
AlexeySachkov: Went with suggestion from Artem in 05d38e03deb11d274068805048ddbab164455924

tools/llvm-spirv/llvm-spirv.cpp
AlexeySachkov: Applied in 05d38e03deb11d274068805048ddbab164455924

tools/llvm-spirv/llvm-spirv.cpp
AGindinson: Thanks for expanding on this, SGTM.

lib/SPIRV/SPIRVWriter.cpp
AlexeySotkin: `checkTypeForSPIRVExtendedInstLowering` takes `II` as a parameter. This function is based on `switch` on intrinsic id and there is no `case` for `Intrinsic::fmuladd`. We either should add `case Intrinsic::fmuladd:`, or don't call `checkTypeForSPIRVExtendedInstLowering` at all.

AlexeySachkov: Fair enough, I will remove the call

AlexeySotkin (approved):

Issue #801, opened 2020-10-21
Discussion about clang-tidy CI check


AlexeySachkov:
Hi,

This would be a follow-up from #770: so far we had only two failures of [Check code-style workflow](https://github.com/KhronosGroup/SPIRV-LLVM-Translator/actions?query=workflow%3A%22Check+code+style%22): [first](https://github.com/KhronosGroup/SPIRV-LLVM-Translator/runs/1281193465?check_suite_focus=true), [second](https://github.com/KhronosGroup/SPIRV-LLVM-Translator/runs/1275816326?check_suite_focus=true).

In both cases, `clang-tidy` failed, but logs don't contain any errors, but only warnings. The workflow failed because it just checks if log is non-empty.

Therefore, I have a few questions:

- do we care about warnings from `clang-tidy`? If yes, then we can leave workflow file as-is or update it to pass something like `-Werror` to `clang-tidy` (if such exists) and grep for `error:`. Otherwise, we need to update the workflow to grep for `error:`
- do we need to re-evaluate list of checks enabled? Probably we do not care about warnings, but would like to turn `readability-identifier-naming` into an error, i.e. make it a mandatory check?

It would be good to have a way to highlight somehow that check was completed successfully, but there are some warnings. However, I'm not sure that this is possible (or easily possible) with current GitHub Actions functionality.

Tagging @svenvh  and @AlexeySotkin for feedback.

svenvh:
> do we care about warnings from clang-tidy?

I think so; this is how the check worked with Travis too, right?  I think it helps to keep the code follow the coding style.  If we see warnings that we do not care about, then we should probably try to suppress them, but currently I am not aware of any that we want to suppress.

svenvh:
> I've decided to turn warnings from the checks we have enabled to errors to better clarify that these are must-fix things.

That makes sense.

> Also, clang-tidy messages will now be published directly on "Files changed" page

That looks very handy indeed!

PR #825, opened 2020-11-23
Support the SPV_EXT_shader_atomic_float_add extension

AGindinson:
The new `AtomicFAddEXT` instruction will be mapped onto
`__spirv_AtomicFAddEXT()` external calls in LLVM IR. No
additional logic is required to facilitate this - the
existing infrastructure does the replacement by default
based on the `__spirv` prefix.

The full specification can be found at
github.com/KhronosGroup/SPIRV-Registry/blob/master/extensions/EXT/SPV_EXT_shader_atomic_float_add.asciidoc


MrSidims (commented):
That's a little annoying, how CI inlines comments about failed jobs and hides the actual changes... 
... Or there are no changes in lib/SPIRV/libSPIRV/SPIRVOpCodeEnum.h and lib/SPIRV/libSPIRV/SPIRVEntry.h ?

test/AtomicFAddExt.ll
```diff
@@ -0,0 +1,118 @@
+; RUN: llvm-as < %s > %t.bc
+; RUN: llvm-spirv %t.bc --spirv-ext=+SPV_INTEL_fpga_loop_controls -o %t.spv
```
MrSidims, line 2:
```suggestion
; RUN: llvm-spirv %t.bc --spirv-ext=+SPV_EXT_shader_atomic_float_add -o %t.spv
```

MrSidims: And I see a problem here... I assume, even with the incorrect line, the test is passing on your side, right? That is because there is no code, that checks if the extension is supported. There are two ways, how to do this:
1. Ask a SPIR-V module if it supports the extension. But I'm unsure, how it can work in your case...
2. Add `getRequiredExtension` method for an instruction class, like this:
```
llvm::Optional<ExtensionID> getRequiredExtension() const override {
  return ExtensionID::SPV_EXT_shader_atomic_float_add;
}
```

Problem here is that you are modifying already existing class (that was added for translation of core atomic instructions). So going this way requires to add another class for this particular extension.

MrSidims (changes requested):

test/AtomicFAddExt.ll
AGindinson: Whoops, thanks for noticing. It was a lazy `Ctrl + V` from another test, hence the misspell.
Good point with the extension check as well! I could probably add an `if (getOpCode() == OpAtomicFAddEXT)` check to the override you've suggested.

AGindinson:
> That's a little annoying, how CI inlines comments about failed jobs and hides the actual changes...
> ... Or there are no changes in lib/SPIRV/libSPIRV/SPIRVOpCodeEnum.h and lib/SPIRV/libSPIRV/SPIRVEntry.h ?

AFAIU, the CI comments have not interfered with the actual changes that are present in all of the files marked relevant.

test/AtomicFAddExt.ll
AGindinson: Done; I've also checked that leaving the extension unspecified in the CL results in a correponding LIT failure.

MrSidims (approved):

AlexeySachkov:
> > That's a little annoying, how CI inlines comments about failed jobs and hides the actual changes...
> > ... Or there are no changes in lib/SPIRV/libSPIRV/SPIRVOpCodeEnum.h and lib/SPIRV/libSPIRV/SPIRVEntry.h ?
> 
> AFAIU, the CI comments have not interfered with the actual changes that are present in all of the files marked relevant.

Yes, those annotations should behave in a similar way to review comments, i.e. they shouldn't hide any source file - I still believe that this is a cool feature, but I agree that particular errors reported in this PR is a bug in our clang-tidy configuration and they shouldn't have been reported.

Anyway, this functionality was implemented in #802 as a result of discussion in #801. If you have any feedback please post it in the PR or issue (you can also use #716 as a generic tracker for CI-related discussions) so we can collect it and improve things

lib/SPIRV/libSPIRV/SPIRVInstruction.h
```diff
@@ -2745,6 +2762,9 @@ class SPIRVAtomicInstBase : public SPIRVInstTemplateBase {
     if (getOpCode() == OpAtomicStore &&
         getOperand(ValueOperandIndex)->getType()->isTypeInt(64))
       Module->addCapability(CapabilityInt64Atomics);
+    if (getOpCode() == OpAtomicFAddEXT)
```
AlexeySachkov, line 2765: There is a comment above this method that says that it exists only because of `OpAtomicStore` - this is not true anymore, please update the comment as well

lib/SPIRV/libSPIRV/SPIRVInstruction.h
```diff
@@ -2721,15 +2721,32 @@ _SPIRV_OP(PowN, true, 10)
...
+
     // Most of atomic instructions do not require any capabilities
     // ... unless they operate on 64-bit integers.
-    if (hasType() && getType()->isTypeInt(64)) {
+    if (getType()->isTypeInt(64)) {
       // In SPIRV 1.2 spec only 2 atomic instructions have no result type:
       // 1. OpAtomicStore - need to check type of the Value operand
       // 2. OpAtomicFlagClear - doesn't require Int64Atomics capability.
       CapVec.push_back(CapabilityInt64Atomics);
+    } else if (getOpCode() == OpAtomicFAddEXT) {
```
AlexeySachkov, line 2742:
```suggestion
    } else if (getOpCode() == OpAtomicFAddEXT) {
      // ... or unless they operate on floating-point types
```

Actually, I'm not entirely sure about this comment, because the `if` might be expanded later to accommodate other atomic instructions from various extensions, so probably we should leave just one, but generic comment, which explains possible reasons why some capabilities might be required?

test/AtomicFAddExt.ll
```diff
@@ -0,0 +1,118 @@
+; RUN: llvm-as < %s > %t.bc
```
AlexeySachkov, line 1: Why not just
```suggestion
; RUN: llvm-as %s -o %t.bc
```

?

To me, it seems more readable

test/AtomicFAddExt.ll
```diff
@@ -0,0 +1,118 @@
...
+%"class._ZTSN2cl4sycl6detail5arrayILi1EEE.cl::sycl::detail::array" = type { [1 x i64] }
+%"class._ZTSN2cl4sycl2idILi1EEE.cl::sycl::id" = type { %"class._ZTSN2cl4sycl6detail5arrayILi1EEE.cl::sycl::detail::array" }
+
+$_ZTSZZ3addIfEvvENKUlRN2cl4sycl7handlerEE19_14clES3_EUlNS1_4itemILi1ELb1EEEE23_37 = comdat any
+
+$_ZTSZZ3addIdEvvENKUlRN2cl4sycl7handlerEE19_14clES3_EUlNS1_4itemILi1ELb1EEEE23_37 = comdat any
+
+@__spirv_BuiltInGlobalInvocationId = external dso_local local_unnamed_addr addrspace(1) constant <3 x i64>, align 32
+
+; CHECK-SPIRV: Capability AtomicFloat32AddEXT
```
AlexeySachkov, line 22: Please also add check for `OpExtension`

AlexeySachkov:
> That's a little annoying, how CI inlines comments about failed jobs and hides the actual changes...
> ... Or there are no changes in lib/SPIRV/libSPIRV/SPIRVOpCodeEnum.h and lib/SPIRV/libSPIRV/SPIRVEntry.h ?

I've just found that you can press "options" (three dots) on the right side of the title of each file in "Files changed" page to open a menu and disable annotations for that file

MrSidims:
> > > That's a little annoying, how CI inlines comments about failed jobs and hides the actual changes...
> > > ... Or there are no changes in lib/SPIRV/libSPIRV/SPIRVOpCodeEnum.h and lib/SPIRV/libSPIRV/SPIRVEntry.h ?
> > 
> > 
> > AFAIU, the CI comments have not interfered with the actual changes that are present in all of the files marked relevant.
> 
> Yes, those annotations should behave in a similar way to review comments, i.e. they shouldn't hide any source file - I still believe that this is a cool feature, but I agree that particular errors reported in this PR is a bug in our clang-tidy configuration and they shouldn't have been reported.
> 
> Anyway, this functionality was implemented in #802 as a result of discussion in #801. If you have any feedback please post it in the PR or issue (you can also use #716 as a generic tracker for CI-related discussions) so we can collect it and improve things



> > > That's a little annoying, how CI inlines comments about failed jobs and hides the actual changes...
> > > ... Or there are no changes in lib/SPIRV/libSPIRV/SPIRVOpCodeEnum.h and lib/SPIRV/libSPIRV/SPIRVEntry.h ?
> > 
> > 
> > AFAIU, the CI comments have not interfered with the actual changes that are present in all of the files marked relevant.
> 
> Yes, those annotations should behave in a similar way to review comments, i.e. they shouldn't hide any source file - I still believe that this is a cool feature, but I agree that particular errors reported in this PR is a bug in our clang-tidy configuration and they shouldn't have been reported.
> 
> Anyway, this functionality was implemented in #802 as a result of discussion in #801. If you have any feedback please post it in the PR or issue (you can also use #716 as a generic tracker for CI-related discussions) so we can collect it and improve things

I was a little bit harsh by saying "annoying", sorry. Main purpose of my comment was to understand if there are any changes in the file or not. If there are changes - it's definitely a bug of clang-tidy and I'll report it. If it's not - well, actually, I even think, that leaving this behavior is OK since it will make us clean up the code (me - for sure, since I'm annoyed I would prepare NFC patch with the fixes to become not annoyed :-) ).

AlexeySachkov:
> > Yes, those annotations should behave in a similar way to review comments, i.e. they shouldn't hide any source file - I still believe that this is a cool feature, but I agree that particular errors reported in this PR is a bug in our clang-tidy configuration and they shouldn't have been reported.
> > Anyway, this functionality was implemented in #802 as a result of discussion in #801. If you have any feedback please post it in the PR or issue (you can also use #716 as a generic tracker for CI-related discussions) so we can collect it and improve things
> 
> I was a little bit harsh by saying "annoying", sorry. Main purpose of my comment was to understand if there are any changes in the file or not. If there are changes (that are hidden) - it's definitely a bug of clang-tidy and I'll report it. If it's not - well, actually, I even think, that leaving this behavior is OK since it will make us clean up the code (me - for sure, since I'm annoyed I would prepare NFC patch with the fixes to become not annoyed :-) ).

By default annotations are only shown in files which were changed, but on "Files changed" page there is a separate section at the bottom, which contains annotations for files which weren't changed in the PR. It says this is some kind of "Beta" functionality

MrSidims:
> By default annotations are only shown in files which were changed, but on "Files changed" page there is a separate section at the bottom, which contains annotations for files which weren't changed in the PR. It says this is some kind of "Beta" functionality

Ah, got it, thanks!

lib/SPIRV/libSPIRV/SPIRVInstruction.h
AGindinson: Thanks for bringing this one up! Replaced the comment with a more general one.

test/AtomicFAddExt.ll
AGindinson: Done; must've been an oblivious copy & paste :)

test/AtomicFAddExt.ll
AGindinson: Sure, done.

test/AtomicFAddExt.ll
```diff
@@ -0,0 +1,119 @@
...
+  %0 = getelementptr inbounds %"class._ZTSN2cl4sycl2idILi1EEE.cl::sycl::id", %"class._ZTSN2cl4sycl2idILi1EEE.cl::sycl::id"* %_arg_3, i64 0, i32 0, i32 0, i64 0
+  %1 = load i64, i64* %0, align 8
+  %add.ptr.i29 = getelementptr inbounds float, float addrspace(1)* %_arg_, i64 %1
+  %2 = getelementptr inbounds %"class._ZTSN2cl4sycl2idILi1EEE.cl::sycl::id", %"class._ZTSN2cl4sycl2idILi1EEE.cl::sycl::id"* %_arg_8, i64 0, i32 0, i32 0, i64 0
+  %3 = load i64, i64* %2, align 8
+  %add.ptr.i = getelementptr inbounds float, float addrspace(1)* %_arg_4, i64 %3
+  %4 = load <3 x i64>, <3 x i64> addrspace(4)* addrspacecast (<3 x i64> addrspace(1)* @__spirv_BuiltInGlobalInvocationId to <3 x i64> addrspace(4)*), align 32, !noalias !5
+  %5 = extractelement <3 x i64> %4, i64 0
+  ; CHECK-SPIRV: 7 AtomicFAddEXT [[TYPE_FLOAT_32]]
+  ; CHECK-LLVM: call spir_func float @[[FLOAT_FUNC_NAME:_Z21__spirv_AtomicFAddEXT[[:alnum:]]+]]({{.*}})
```
AlexeySotkin, line 40: Do I get it right that there is no equivalent OpenCL 1.2 or 2.0 builtin for this instruction?
What IR will be generated if I run:
1. `llvm-spirv -r --spirv-target-env=CL1.2 %t.spv -o %t.rev.bc`
2. `llvm-spirv -r --spirv-target-env=CL2.0 %t.spv -o %t.rev.bc`
3. `llvm-spirv -r --spirv-target-env=SPV-IR %t.spv -o %t.rev.bc` ?

lib/SPIRV/libSPIRV/SPIRVInstruction.h
```diff
@@ -2721,30 +2721,51 @@ _SPIRV_OP(PowN, true, 10)
 
 class SPIRVAtomicInstBase : public SPIRVInstTemplateBase {
 public:
+  llvm::Optional<ExtensionID> getRequiredExtension() const override {
+    if (getOpCode() == OpAtomicFAddEXT)
+      return ExtensionID::SPV_EXT_shader_atomic_float_add;
+    return {};
+  }
```
AlexeySotkin, line 2728: As a future improvement I think it makes sense to create a map `OpCode` -> `ExtensionID`.

AGindinson: To clarify, is your suggestion to map all of the `OpCode`s requiring some extension to the corresponding `ExtensionID`s? Unless there are cases where the extension's requirement is conditional, I also believe this could be a great improvement. For one, this could drop the need for having `getRequiredExtension()` as `virtual`.

As for this particular bit, I've actually had a slightly different refactoring idea, namely to specialize functions for the macro-generated `SPIRVAtomic*` classes (since the `RequiredCapability` specialization also wants for some improvement). This would probably include some rework of the `__SPIRV_OP` macro - I haven't included such attempts into this PR, but creating a follow-up could be a plan.

test/AtomicFAddExt.ll
AGindinson: Apparently, no corresponding builtin exists in OpenCL 1.2/2.0. The IR resulting from backwards translation retains the same `__spirv_AtomicFAddExt*()` form in all 3 cases that you have brought up.

Tagging @aratajew just in case there's any misalignment on OpenCL 1.2/2.0 builtin expectations here.

AlexeySotkin (approved):

PR #826, opened 2020-11-24
Support of OpSpecConstantComposite instruction

AlexeySotkin:

### DIFF
--- a/test/transcoding/SpecConstantComposite.ll
+++ b/test/transcoding/SpecConstantComposite.ll
@@ -1,0 +1,136 @@
+; RUN: llvm-as < %s -o %t.bc
+; RUN: llvm-spirv %t.bc -spirv-text -o %t.spt
+; RUN: FileCheck < %t.spt %s --check-prefix=CHECK-SPIRV
+; RUN: llvm-spirv %t.bc -o %t.spv
+; RUN: spirv-val %t.spv
+; RUN: llvm-spirv %t.spv -r -o - | llvm-dis -o %t.ll
+; RUN: FileCheck < %t.ll %s --check-prefix=CHECK-LLVM
+
+; CHECK-SPIRV: Decorate [[#SC3:]] SpecId 3
+; CHECK-SPIRV: Decorate [[#SC4:]] SpecId 4
+; CHECK-SPIRV: Decorate [[#SC6:]] SpecId 6
+; CHECK-SPIRV: Decorate [[#SC7:]] SpecId 7
+; CHECK-SPIRV: Decorate [[#SC10:]] SpecId 10
+; CHECK-SPIRV: Decorate [[#SC11:]] SpecId 11
+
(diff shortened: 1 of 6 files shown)


test/transcoding/SpecConstantComposite.ll
```diff
@@ -0,0 +1,135 @@
+; RUN: llvm-as < %s -o %t.bc
+; RUN: llvm-spirv %t.bc -spirv-text -o %t.spt
+; RUN: FileCheck < %t.spt %s --check-prefix=CHECK-SPIRV
+; RUN: llvm-spirv %t.bc -o %t.spv
```
AlexeySachkov, line 4: What do you think about adding `spirv-val` invocation here?

test/transcoding/SpecConstantComposite.ll
```diff
@@ -0,0 +1,135 @@
...
+; RUN: llvm-spirv %t.spv -r -o - | llvm-dis -o %t.ll
```
AlexeySachkov, line 5: I suggest adding a separate test to check `--spec-const` and `--spec-const-info` command line options with composite spec constants

test/transcoding/SpecConstantComposite.ll
AlexeySotkin: Makes sense. I will add it.

test/transcoding/SpecConstantComposite.ll
AlexeySotkin: Good idea, thanks.

test/transcoding/SpecConstantComposite.ll
AlexeySotkin: Added in 4a0fc3a.

test/transcoding/SpecConstantComposite.ll
AlexeySotkin: I have added the test in 994ea05.
`--spec-const-info` doesn't show information about composite specialization constants. Intention of this option was to provide information required for specialization via OpenCL API for example. But, since, we can't specialize a composite directly there was no point to add handling for OpSpecConstantComposite. Please let me know if you think we should extend `--spec-const-info` to show composite specialization constants (that may require changing or adding new API to LLVMSPIRVLib).

AlexeySachkov: I think there is no need to print info about composite spec constants since we can't set them through OpenCL API, but I wanted to check that this functionality won't break in presence of composite spec constant, i.e. a simple test that all elements of a composite are correctly printed by this option

AlexeySotkin: I'm confused a bit. @AlexeySachkov the test added in 994ea05 - is it enough?

AlexeySachkov: > @AlexeySachkov the test added in 994ea05 - is it enough?

Yes

AlexeySachkov (approved):

PR #830, opened 2020-11-24
Align with LLORG on sret/byval parameter attribute types

AGindinson:
With the latest community changes, the StructRet/ByVal parameter
attributes are now required to have a type. Fix the build by:
1. updating the LIT IR to match this policy;
2. adding correct attribute types in backwards translation.

### DIFF
--- a/lib/SPIRV/SPIRVReader.cpp
+++ b/lib/SPIRV/SPIRVReader.cpp
@@ -2885,7 +2885,10 @@ Function *SPIRVToLLVM::transFunction(SPIRVFunction *BF) {
         AttrTy = cast<PointerType>(I->getType())->getElementType();
       else if (LLVMKind == Attribute::AttrKind::StructRet)
         AttrTy = I->getType();
-      I->addAttr(Attribute::get(*Context, LLVMKind, AttrTy));
+      // Make sure to use a correct constructor for a typed/typeless attribute
+      auto A = AttrTy ? Attribute::get(*Context, LLVMKind, AttrTy)
+                      : Attribute::get(*Context, LLVMKind);
+      I->addAttr(A);
     });
 
     SPIRVWord MaxOffset = 0;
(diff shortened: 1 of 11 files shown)


AGindinson:
@AlexeySachkov @AlexeySotkin @svenvh FYI

lib/SPIRV/SPIRVReader.cpp
```diff
@@ -2879,7 +2879,16 @@ Function *SPIRVToLLVM::transFunction(SPIRVFunction *BF) {
     BA->foreachAttr([&](SPIRVFuncParamAttrKind Kind) {
       if (Kind == FunctionParameterAttributeNoWrite)
         return;
-      F->addAttribute(I->getArgNo() + 1, SPIRSPIRVFuncParamAttrMap::rmap(Kind));
+      Attribute::AttrKind LLVMKind = SPIRSPIRVFuncParamAttrMap::rmap(Kind);
```
AlexeySachkov, line 2882: Am I right that this change is covered by some existing tests? I.e. we do not have an exact check that we generated typed attribute, but since this is a requirement of LLVM and we have tests that use such attributes we have this change covered by the fact that some existing tests are not crashing during backward translation, right?

AlexeySachkov (approved):
LGTM, thanks for working on this!

lib/SPIRV/SPIRVReader.cpp
AGindinson: Yes, I believe that's a very clean description.

Before the change was up, LLVM validation was failing when assembling the `.bc` file due to attributes' lacking the now-mandatory type; other tests were crashing on a `SPIRVReader`-stemming assertion because the LLVM library had been expecting a non-`nullptr` type.

The fact that the `isa<> applied to nullptr` assertion no longer fires can prove that the type gets set; the fact that `llvm-dis` is able to produce a textual IR file proves that `.bc` passes the validation.

MrSidims (approved):
Thanks!

AlexeySotkin (approved):

PR #832, opened 2020-11-25
Move assume/expect to spirv_internal.hpp

MrSidims:
SPV_INTEL_optimization_hints isn't going to be upstreamed to SPIR-V
Headers repository, therefore this patch is needed. Also it serves
a purpose to show, how can extensions be implemented using
spirv_internal.hpp.


lib/SPIRV/SPIRVReader.cpp
```diff
@@ -1795,7 +1795,8 @@ Value *SPIRVToLLVM::transValueWithoutDecoration(SPIRVValue *BV, Function *F,
   }
 
   // Translation of instructions
-  switch (BV->getOpCode()) {
+  int OpCode = BV->getOpCode();
+  switch (OpCode) {
```
AlexeySachkov, line 1799: Why do you need this change? If I understand correctly, `internal::OpAssumeTrueINTEL` should already have the same type as `OpBranch`, i.e. `Op`, so no explicit conversion to `int` should be needed to align types. Am I missing something?

MrSidims: `internal::OpAssumeTrueINTEL` isn't member of `spv::Op`, instead it's a member of `spv::internal::Op`, hence there is a compiler warning.

MrSidims: Double check:
/localdisk2/sidorovd/xmain/ics-ws/SYCLOS/llvm/llvm-spirv/lib/SPIRV/SPIRVReader.cpp:2412:3: warning: case value '5630' not in enumerated type 'spv::Op' [-Wswitch]
   case internal::OpAssumeTrueINTEL: {

MrSidims: Double check:
lib/SPIRV/SPIRVReader.cpp:2412:3: warning: case value '5630' not in enumerated type 'spv::Op' [-Wswitch]
case internal::OpAssumeTrueINTEL: {

AlexeySachkov: Got, thanks

AlexeySachkov (approved):

svenvh (approved):
LGTM

PR #833, opened 2020-11-26
Add GitHub Actions badges

AlexeySotkin:


README.md
```diff
@@ -2,6 +2,10 @@
 
 [![Build Status](https://travis-ci.org/KhronosGroup/SPIRV-LLVM-Translator.svg?branch=master)](https://travis-ci.org/KhronosGroup/SPIRV-LLVM-Translator)
 
+![Out-of-tree build & tests](https://github.com/KhronosGroup/SPIRV-LLVM-Translator/workflows/Out-of-tree%20build%20&%20tests/badge.svg?branch=master)
```
AlexeySachkov, line 5: There are various reasons why workflow can be triggered, including direct push to the branch, opening a PR or launching nightly build.

I suggest that in README we display status of nightly build only and not just some general status (which might correspond to the latest faulty PR). To do so, you need to select `schedule` event in badge generator (or just add `&event=schedule` to badge URL)

AlexeySachkov (approved):

PR #827, opened 2020-11-24
[Feature] Added translation of VCFCEntry attribute

DmitryBushev:
Added translation of VCFCEntry kernel attribute to execution mode
VectorComputeFastCompositeKernelINTEL according to this spec: https://github.com/intel/llvm/blob/693199d47868141cdcc69f8ebd80f149b3e28729/sycl/doc/extensions/SPIRV/SPV_INTEL_vector_compute.asciidoc

PR #836, opened 2020-11-26
[Backport] Added translation of VCFCEntry attribute(#827) branch 110

DmitryBushev:
Added translation of VCFCEntry kernel attribute to execution mode
VectorComputeFastCompositeKernelINTEL

Backport for https://github.com/KhronosGroup/SPIRV-LLVM-Translator/pull/827

### DIFF
--- a/lib/SPIRV/libSPIRV/spirv.hpp
+++ b/lib/SPIRV/libSPIRV/spirv.hpp
@@ -171,6 +171,7 @@ enum ExecutionMode {
     ExecutionModeMaxWorkDimINTEL = 5894,
     ExecutionModeNoGlobalOffsetINTEL = 5895,
     ExecutionModeNumSIMDWorkitemsINTEL = 5896,
+    ExecutionModeVectorComputeFastCompositeKernelINTEL = 6088,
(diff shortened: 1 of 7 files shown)


lib/SPIRV/libSPIRV/spirv.hpp
```diff
@@ -171,6 +171,7 @@ enum ExecutionMode {
     ExecutionModeMaxWorkDimINTEL = 5894,
     ExecutionModeNoGlobalOffsetINTEL = 5895,
     ExecutionModeNumSIMDWorkitemsINTEL = 5896,
+    ExecutionModeVectorComputeFastCompositeKernelINTEL = 8088,
```
AlexeySachkov, line 174: We have recently found that there is a bug in the spec with token assignments: https://github.com/intel/llvm/pull/1612#discussion_r535084124, so this should be `6068`

AlexeySachkov (changes requested):

lib/SPIRV/libSPIRV/spirv.hpp
DmitryBushev: Sure, I 'll update all my pull requests

PR #841, opened 2020-12-02
support select i1 on spirv 1.0

airlied:
This issue has been raised internally in Khronos, but I'll publish my workaround for review.

LLVM 12 has started to generate select on vectors with a scalar
condition. SPIR-V 1.4 allows this, but SPIR-V 1.0/1 for CL don't.

This changes the translator to translate

select(bool, vec4, vec4)
into
bvec4 = (bool, bool, bool, bool)
select(bvec4, vec4, vec4)

which might not be as efficient but should work.

It also fixes the composite construct validation code to
not try and lookup the id before it's been added


lib/SPIRV/SPIRVWriter.cpp
```diff
@@ -1365,14 +1365,31 @@ SPIRVValue *LLVMToSPIRV::transValueWithoutDecoration(Value *V,
...
-        BM->addSelectInst(
-            transValue(Sel->getCondition(), BB),
-            transValue(Sel->getTrueValue(), BB, true, FuncTransMode::Pointer),
-            transValue(Sel->getFalseValue(), BB, true, FuncTransMode::Pointer),
-            BB));
+  if (SelectInst *Sel = dyn_cast<SelectInst>(V)) {
+    auto CV = transValue(Sel->getCondition(), BB);
+    auto TV = transValue(Sel->getTrueValue(), BB, true, FuncTransMode::Pointer);
+
+    if (!CV->getType()->isTypeVector() && TV->getType()->isTypeVector()) {
```
AlexeySotkin, line 1372: I think it would be nice to have a comment with short description why we are doing this.

lib/SPIRV/SPIRVWriter.cpp
```diff
@@ -1365,14 +1365,31 @@ SPIRVValue *LLVMToSPIRV::transValueWithoutDecoration(Value *V,
...
+      Type *BoolTy = IntegerType::getInt1Ty(M->getContext());
+      auto VT = transType(FixedVectorType::get(
+          BoolTy, TV->getType()->getVectorComponentCount()));
+      std::vector<SPIRVId> IdVec;
+      for (unsigned Comp = 0; Comp < TV->getType()->getVectorComponentCount();
+           Comp++)
+        IdVec.push_back(CV->getId());
```
AlexeySotkin, line 1379: I guess we can use the fill constructor here:
```suggestion
      std::vector<SPIRVId> IdVec(TV->getType()->getVectorComponentCount(),
                                 CV->getId());
```

lib/SPIRV/SPIRVWriter.cpp
```diff
@@ -1365,14 +1365,31 @@ SPIRVValue *LLVMToSPIRV::transValueWithoutDecoration(Value *V,
...
-        V,
-        BM->addSelectInst(
-            transValue(Sel->getCondition(), BB),
-            transValue(Sel->getTrueValue(), BB, true, FuncTransMode::Pointer),
-            transValue(Sel->getFalseValue(), BB, true, FuncTransMode::Pointer),
-            BB));
+  if (SelectInst *Sel = dyn_cast<SelectInst>(V)) {
+    auto CV = transValue(Sel->getCondition(), BB);
+    auto TV = transValue(Sel->getTrueValue(), BB, true, FuncTransMode::Pointer);
+
```
AlexeySotkin, line 1371: Probably it makes sense to have a variable to hold the false value.

lib/SPIRV/SPIRVWriter.cpp
```diff
@@ -1365,14 +1365,31 @@ SPIRVValue *LLVMToSPIRV::transValueWithoutDecoration(Value *V,
...
+        IdVec.push_back(CV->getId());
+      auto CompConstruct = BM->addCompositeConstructInst(VT, IdVec, BB);
+      return mapValue(
+          V, BM->addSelectInst(CompConstruct, TV,
+                               transValue(Sel->getFalseValue(), BB, true,
+                                          FuncTransMode::Pointer),
+                               BB));
+    }
+    return mapValue(V,
+                    BM->addSelectInst(transValue(Sel->getCondition(), BB), TV,
```
AlexeySotkin, line 1388:
```suggestion
                    BM->addSelectInst(CV, TV,
```

AlexeySotkin (commented):
@airlied please add a lit test.

airlied:
Okay I think I've addressed all review comment and added a basic lit test.

lib/SPIRV/SPIRVWriter.cpp
```diff
@@ -1421,14 +1421,24 @@ SPIRVValue *LLVMToSPIRV::transValueWithoutDecoration(Value *V,
...
-            BB));
+  if (SelectInst *Sel = dyn_cast<SelectInst>(V)) {
+    auto CV = transValue(Sel->getCondition(), BB);
+    auto TV = transValue(Sel->getTrueValue(), BB, true, FuncTransMode::Pointer);
+    auto FV = transValue(Sel->getFalseValue(), BB, true, FuncTransMode::Pointer);
+
+    // LLVM can generate vector selects with a bool selector operand,
+    // This isn't supported until SPIRV 1.4. For now translate them
+    // into vector selects and hope the backend can undo this.
+    if (!CV->getType()->isTypeVector() && TV->getType()->isTypeVector()) {
```
AlexeySotkin, line 1432: One more thing. Do I get it right that without this PR it works correctly if we want to generate SPIR-V 1.4 or above. If so, I was going to suggest:
```suggestion
    if (!CV->getType()->isTypeVector() && TV->getType()->isTypeVector() &&
        !BM->isAllowedToUseVersion(VersionNumber::SPIRV_1_4)) {
```
But after https://github.com/KhronosGroup/SPIRV-LLVM-Translator/pull/505/commits/1eb284b1e0b2c18ee44be6717cb11fd1a00d9531 from @bashbaug, the maximum supported version is 1.3
https://github.com/KhronosGroup/SPIRV-LLVM-Translator/blob/9326deff2916f9f566aadf5cedeb58f8d65e945e/include/LLVMSPIRVOpts.h#L49-L60

Does anybody knows if we can populate the enum with `SPIRV_1_4` at least?

AlexeySotkin: @airlied Could you please add `SPIRV_1_4 = 0x00010400,` to this enum, update `MaximumVersion` and apply my suggestion above.
Other than that, LGTM. Thanks!

PR #842, opened 2020-12-03
Replace illegal address space casts asserts with errors

MrSidims:
Purpose of this patch is to make the translator a little bit more
user-friendly. For example following SYCL code:
```
  SYCL_EXTERNAL sycl::global_ptr<int> kernel_fun(int *ptr) {
    return sycl::global_ptr<int>((int*)&ptr);
  }
```

is UB since it has an implicit cast from private address space to
global. It's hard to detect such UB in FE, hence this diagnostic is
moved to the translator (which already had it, but in form of an
assert).


MrSidims:
Tagging @bader @AlexeySotkin @AlexeySachkov

svenvh (commented):
LGTM, but probably worth adding some tests in `test/negative`.

svenvh (approved):

PR #846, opened 2020-12-04
[NFC] Instantiate atomic instructions via inheritance

AGindinson:
The objective of this change is to reduce `OpCode`-based branching in the base
class for atomic instructions, and to simplify specialization for ones that will
be added in future.

### DIFF
--- a/lib/SPIRV/libSPIRV/SPIRVInstruction.h
+++ b/lib/SPIRV/libSPIRV/SPIRVInstruction.h
@@ -2721,51 +2721,55 @@ _SPIRV_OP(PowN, true, 10)
 
 class SPIRVAtomicInstBase : public SPIRVInstTemplateBase {
 public:
-  llvm::Optional<ExtensionID> getRequiredExtension() const override {
-    if (getOpCode() == OpAtomicFAddEXT)
-      return ExtensionID::SPV_EXT_shader_atomic_float_add;
+  SPIRVCapVec getRequiredCapability() const override {
+    // Most of the atomic instructions require a specific capability when
+    // operating on 64-bit integers.
+    // In SPIRV 1.2 spec, only 2 atomic instructions have no result type:
+    // 1. OpAtomicStore - need to check type of the Value operand
+    // 2. OpAtomicFlagClear - doesn't require Int64Atomics capability.
+    // Besides, OpAtomicCompareExchangeWeak, OpAtomicFlagTestAndSet and
+    // OpAtomicFlagClear instructions require the "kernel" capability. But this
+    // capability should be added by setting the OpenCL memory model.
+    if (hasType() && getType()->isTypeInt(64))
+      return {CapabilityInt64Atomics};
     return {};
   }
 
-  SPIRVCapVec getRequiredCapability() const override {
-    SPIRVCapVec CapVec;
-    if (!hasType())
-      return CapVec;
-
-    // Most atomic instructionsrequire a specific capability when
-    // operating on 64-bit integers.
-    if (getType()->isTypeInt(64)) {
-      // In SPIRV 1.2 spec only 2 atomic instructions have no result type:
-      // 1. OpAtomicStore - need to check type of the Value operand
-      // 2. OpAtomicFlagClear - doesn't require Int64Atomics capability.
-      CapVec.push_back(CapabilityInt64Atomics);
-    } else if (getOpCode() == OpAtomicFAddEXT) {
-      if (getType()->isTypeFloat(32))
-        CapVec.push_back(CapabilityAtomicFloat32AddEXT);
-      else if (getType()->isTypeFloat(64))
-        CapVec.push_back(CapabilityAtomicFloat64AddEXT);
-      else
-        llvm_unreachable(
-            "AtomicFAddEXT can only be generated for f32 or f64 types");
-    }
-    // Per the spec OpAtomicCompareExchangeWeak, OpAtomicFlagTestAndSet and
-    // OpAtomicFlagClear instructions require kernel capability. But this
-    // capability should be added by setting OpenCL memory model.
-    return CapVec;
-  }
-
   // Overriding the following method because of particular OpAtomic*
-  // instructions. We may have to declare additional capabilities,
-  // e.g. based on operand types.
+  // instructions that declare additional capabilities, e.g. based on operand
+  // types.
+  void setOpWords(const std::vector<SPIRVWord> &TheOps) override {
+    SPIRVInstTemplateBase::setOpWords(TheOps);
+    for (auto RC : getRequiredCapability())
+      Module->addCapability(RC);
+  }
+};
+
+class SPIRVAtomicStoreInst : public SPIRVAtomicInstBase {
+public:
+  // Overriding the following method because of 'const'-related
+  // issues with overriding getRequiredCapability(). TODO: Resolve.
   void setOpWords(const std::vector<SPIRVWord> &TheOps) override {
     SPIRVInstTemplateBase::setOpWords(TheOps);
     static const unsigned ValueOperandIndex = 3;
-    if (getOpCode() == OpAtomicStore &&
-        getOperand(ValueOperandIndex)->getType()->isTypeInt(64))
+    if (getOperand(ValueOperandIndex)->getType()->isTypeInt(64))
       Module->addCapability(CapabilityInt64Atomics);
-    if (getOpCode() == OpAtomicFAddEXT)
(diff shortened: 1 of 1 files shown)


AlexeySachkov:
> To achieve this, the const qualifier has been removed from the base
method - otherwise, it wouldn't have been possible to call the non-const
getOperand().

Why can't we add a `const` version of `getOperand()`? It seems to be a better option, because I wouldn't expect a method like `getRequiredCapability()` to modify the class. Theoretically, if required capability is hard to compute, one might cache it in the class field, but I think that this is an unlikely situation.

lib/SPIRV/libSPIRV/SPIRVInstruction.h
```diff
@@ -2721,61 +2721,77 @@ _SPIRV_OP(PowN, true, 10)
...
+class SPIRVAtomicStoreInstBase : public SPIRVAtomicInstBase {
+public:
+  SPIRVCapVec getRequiredCapability() override {
+    static const unsigned ValueOperandIndex = 3;
+    auto *ValueOp = getOperand(ValueOperandIndex);
+    if (!ValueOp)
+      return {};
+    if (ValueOp->getType()->isTypeInt(64))
+      return {CapabilityInt64Atomics};
+    return {};
```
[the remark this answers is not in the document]
AGindinson: Ah, of course! Thanks :D

AGindinson:
> Why can't we add a `const` version of `getOperand()`? It seems to be a better option, because I wouldn't expect a method like `getRequiredCapability()` to modify the class. Theoretically, if required capability is hard to compute, one might cache it in the class field, but I think that this is an unlikely situation.

That has been my initial approach, but with that, I'm still unable to debug the segfaults that fire in `SPIRVEntry::get<SPIRVValue>()`. Since the uploaded version hasn't been much better, I've replaced this whole `AtomicStore` rework with a `TODO`.

lib/SPIRV/libSPIRV/SPIRVInstruction.h
```diff
@@ -2721,61 +2721,67 @@ _SPIRV_OP(PowN, true, 10)
...
+    assert(getType()->isTypeFloat(64) &&
+           "AtomicFAddEXT can only be generated for f32 or f64 types");
+    return {CapabilityAtomicFloat64AddEXT};
   }
 };
 
 #define _SPIRV_OP(x, ...)                                                      \
   typedef SPIRVInstTemplate<SPIRVAtomicInstBase, Op##x, __VA_ARGS__> SPIRV##x;
+#define _SPIRV_OP_SPECIAL(x, ...)                                              \
+  typedef SPIRVInstTemplate<SPIRV##x##InstBase, Op##x, __VA_ARGS__> SPIRV##x;
```
AGindinson, line 2779: I might have to rework this into something like:
```suggestion
#define _SPIRV_OP_SPECIAL(x, BaseName, ...)                                              \
  typedef SPIRVInstTemplate<SPIRV##BaseName##InstBase, Op##x, __VA_ARGS__> SPIRV##x;
```
to allow for more versatile specialization.

AlexeySachkov (approved):

lib/SPIRV/libSPIRV/SPIRVInstruction.h
```diff
@@ -2721,61 +2721,68 @@ _SPIRV_OP(PowN, true, 10)
...
+  // instructions that declare additional capabilities, e.g. based on operand
+  // types.
+  void setOpWords(const std::vector<SPIRVWord> &TheOps) override {
+    SPIRVInstTemplateBase::setOpWords(TheOps);
+    for (auto RC : getRequiredCapability())
+      Module->addCapability(RC);
+  }
+};
+
+class SPIRVAtomicStoreInstBase : public SPIRVAtomicInstBase {
```
AlexeySotkin, line 2748: Why do we have "Base" in `SPIRVAtomicStoreInstBase`, does it have any ancestors?

lib/SPIRV/libSPIRV/SPIRVInstruction.h
```diff
@@ -2721,61 +2721,68 @@ _SPIRV_OP(PowN, true, 10)
...
   typedef SPIRVInstTemplate<SPIRVAtomicInstBase, Op##x, __VA_ARGS__> SPIRV##x;
+#define _SPIRV_OP_SPECIAL(x, BaseClass, ...)                                   \
+  typedef SPIRVInstTemplate<SPIRV##BaseClass##Base, Op##x, __VA_ARGS__>        \
+      SPIRV##x;
 // Atomic builtins
 _SPIRV_OP(AtomicFlagTestAndSet, true, 6)
 _SPIRV_OP(AtomicFlagClear, false, 4)
 _SPIRV_OP(AtomicLoad, true, 6)
-_SPIRV_OP(AtomicStore, false, 5)
+_SPIRV_OP_SPECIAL(AtomicStore, AtomicStoreInst, false, 5)
```
AlexeySotkin, line 2785: Can we just redefine `_SPIRV_OP` for these 2 instructions?

lib/SPIRV/libSPIRV/SPIRVInstruction.h
AGindinson: It's a "base template argument" for the instruction type that will be formed as a result of macro unfolding. The names would probably clash was I to rename the class into a simple `SPIRVAtomicStoreInst`.

lib/SPIRV/libSPIRV/SPIRVInstruction.h
AGindinson: Sounds reasonable. A possible drawback might be related to "similar" instructions (like `Load` & `Store`) being somewhat separated, but on the whole, your suggestion should be preferable to a new dedicated macro.

lib/SPIRV/libSPIRV/SPIRVInstruction.h
AlexeySotkin: Sorry, what exactly names would clash?

AGindinson: Sorry, I didn't realise that `_SPIRV_OP(_SPECIAL)` as is right now would unfold into `SPIRVAtomicStore` instead of `SPIRVAtomicStoreInst`. Please discard the name clash argument.

Personally, I slightly prefer using `Base` due to the second point made in the initial reply (it is a "template base", and `_SPIRV_OP_SPECIAL` expects the `*Base` naming as a second parameter to signify that). However, if this looks overly complicated or misleading for the code reader, I'm ready to change this so that the class is named `SPIRVAtomicStoreInst` instead, and the `*Base` naming requirement is lifted from the macro parameter.

AlexeySotkin: Yes, it looks overly complicated and misleading to me.

AGindinson: Deal, fixed with the latest revision.

AlexeySotkin (approved):

AlexeySachkov (approved):

lib/SPIRV/libSPIRV/SPIRVInstruction.h
```diff
@@ -2790,9 +2793,14 @@ _SPIRV_OP(AtomicSMax, true, 7)
...
 _SPIRV_OP(AtomicOr, true, 7)
 _SPIRV_OP(AtomicXor, true, 7)
-_SPIRV_OP(AtomicFAddEXT, true, 7)
 _SPIRV_OP(MemoryBarrier, false, 3)
 #undef _SPIRV_OP
+#define _SPIRV_OP(x, BaseClass, ...)                                           \
+  typedef SPIRVInstTemplate<SPIRV##BaseClass, Op##x, __VA_ARGS__> SPIRV##x;
+// Specialized atomic builtins
+_SPIRV_OP(AtomicStore, AtomicStoreInst, false, 5)
+_SPIRV_OP(AtomicFAddEXT, AtomicFAddEXTInst, true, 7)
```
AlexeySotkin, line 2802:
```suggestion
#define _SPIRV_OP(x, ...)                                           \
  typedef SPIRVInstTemplate<SPIRV##x##Inst, Op##x, __VA_ARGS__> SPIRV##x;
// Specialized atomic builtins
_SPIRV_OP(AtomicStore, false, 5)
_SPIRV_OP(AtomicFAddEXT, true, 7)
```

AlexeySotkin (approved):

PR #847, opened 2020-12-05
Implement SPV_INTEL_fpga_loop_fuse extension

MrSidims:
It adds LoopFuseINTEL decoration Request, to the extent possible,
that loops in the function be fused if they are contained in strictly
fewer than Depth other loops in the function. Depth is a 32-bit
unsigned integer type scalar with value at least 1. Independent is a
32-bit unsigned integer type scalar. If Independent is non-zero, it
guarantees that negative distance dependencies do not exist between
loops in the function that are contained in strictly fewer than Depth
other loops within the function.

Spec:
https://github.com/KhronosGroup/SPIRV-Registry/pull/88/


MrSidims:
Tagging @mlychkov

mlychkov (approved):

lib/SPIRV/SPIRVReader.cpp
```diff
@@ -4195,6 +4195,16 @@ bool SPIRVToLLVM::transFPGAFunctionMetadata(SPIRVFunction *BF, Function *F) {
     MetadataVec.push_back(ConstantAsMetadata::get(getInt32(M, 1)));
     F->setMetadata(kSPIR2MD::StallEnable, MDNode::get(*Context, MetadataVec));
   }
+  if (BF->hasDecorate(DecorationFuseLoopsInFunctionINTEL)) {
+    std::vector<Metadata *> MetadataVec;
+    auto Literals =
+        BF->getDecorationLiterals(DecorationFuseLoopsInFunctionINTEL);
+    MetadataVec.push_back(ConstantAsMetadata::get(
+        ConstantInt::get(Type::getInt32Ty(*Context), Literals[0])));
```
AlexeySotkin, line 4203: Minor:
```suggestion
        getUInt32(M, Literals[0]);
```

lib/SPIRV/SPIRVReader.cpp
```diff
@@ -4195,6 +4195,16 @@ bool SPIRVToLLVM::transFPGAFunctionMetadata(SPIRVFunction *BF, Function *F) {
...
+    MetadataVec.push_back(ConstantAsMetadata::get(
+        ConstantInt::get(Type::getInt32Ty(*Context), Literals[1])));
```
AlexeySotkin, line 4205: Minor:
```suggestion
        getUInt32(M, Literals[0]);
```

lib/SPIRV/SPIRVReader.cpp
MrSidims: Oh, I should have remembered that from the last review, thanks!

AlexeySotkin (approved):

PR #848, opened 2020-12-06
Implement SPV_INTEL_long_constant_composite extension

Fznamznon:
It is not possible to represent constant array initializer with number of
elements greater than maximum word count (65532) in SPIR-V. Splitting big array
before SPIR-V translation doesn't seem like a robust solution and can cause
performance problems.
To fix this disadvantage we can use something like OpSource and
OpSourceContinued pair of instructions, i.e. invent a new instruction which
will hold remaining constituents of a composite constant if resulting number
of constituents is greater than maximum word count.

Spec is available under https://github.com/intel/llvm/pull/2848


Fznamznon:
@AlexeySotkin , please take a look.

lib/SPIRV/libSPIRV/SPIRVModule.cpp
```diff
@@ -1069,15 +1078,69 @@ SPIRVValue *SPIRVModuleImpl::addNullConstant(SPIRVType *Ty) {
...
+      } else {
+        SPIRVConstantComposite::ContinuedInstType Continued =
+            static_cast<SPIRVConstantComposite::ContinuedInstType>(
+                addCompositeConstantContinuedINTEL(Slice));
+        Res->addContinuedInstruction(Continued);
+      }
+    }
+    return Res;
+  }
   return addConstant(new SPIRVConstantComposite(this, Ty, getId(), Elements));
```
mlychkov, line 1103: Just a suggestion for rearrangement of code (there might be errors because I didn't compile and test it). I think it is simpler to read. You may ignore this suggestion if don't think so.
```suggestion
  const size_t MaxNumElements = 65535 - 3;
  const size_t NumElements = Elements.size();
  if (NumElements <= MaxNumElements ||
      !isAllowedToUseExtension(ExtensionID::SPV_INTEL_long_constant_composite)) {
    return addConstant(new SPIRVConstantComposite(this, Ty, getId(), Elements));
  }

  auto Start = Elements.begin();
  auto End = Start + MaxNumElements;
  std::vector<SPIRVValue *> Slice(Start, End);
  auto *Res = static_cast<SPIRVConstantComposite *>(addCompositeConstant(Ty, Slice));

  do {
    Start = End;
    End = ((Elements.end() - End) > MaxNumElements) ? End + MaxNumElements
                                                    : Elements.end();
    Slice.assign(Start, End);
    SPIRVConstantComposite::ContinuedInstType Continued =
        static_cast<SPIRVConstantComposite::ContinuedInstType>(
            addCompositeConstantContinuedINTEL(Slice));
    Res->addContinuedInstruction(Continued);
  } while ((End - Start) > MaxNumElements);

  return Res;
```

lib/SPIRV/SPIRVWriter.cpp
```diff
@@ -414,16 +414,40 @@ SPIRVType *LLVMToSPIRV::transType(Type *T) {
     if (Name == getSPIRVTypeName(kSPIRVTypeName::ConstantPipeStorage))
       return transType(getPipeStorageType(M));
 
-    auto *Struct = BM->openStructType(T->getStructNumElements(), Name.str());
+    const size_t MaxNumElements = 65535 - 2;
```
AlexeySachkov, line 417: It would be good to reduce amount of magic numbers used in here:

For `2` I suggest to extend `SPIRVTypeStruct` class definition with a filed describing its fixed word count, like it is done for `SPIRVTypePipeStorage `:

https://github.com/KhronosGroup/SPIRV-LLVM-Translator/blob/2f70dccda368494dcd6bae214d793c5c65b44c5e/lib/SPIRV/libSPIRV/SPIRVType.h#L605

For `65535` I suggest to use `INT16_MAX` from [`<cstdint>`](https://en.cppreference.com/w/cpp/types/integer).

Also, this variable can be defined as `constexpr`

lib/SPIRV/SPIRVWriter.cpp
```diff
@@ -414,16 +414,40 @@ SPIRVType *LLVMToSPIRV::transType(Type *T) {
...
+    const size_t NumElements = T->getStructNumElements();
```
AlexeySachkov, line 418:
```suggestion
    const size_t NumElements = ST->getNumElements();
```

lib/SPIRV/SPIRVWriter.cpp
```diff
@@ -414,16 +414,40 @@ SPIRVType *LLVMToSPIRV::transType(Type *T) {
...
+    size_t SPIRVStructNumElements = NumElements;
+    if (NumElements > MaxNumElements &&
+        BM->isAllowedToUseExtension(
+            ExtensionID::SPV_INTEL_long_constant_composite)) {
+      SPIRVStructNumElements = MaxNumElements;
+    }
```
AlexeySachkov, line 424: What happens if `NumElements > MaxNumElements`, but extension is not enabled? Shall we emit an error in that case?

lib/SPIRV/SPIRVWriter.cpp
```diff
@@ -414,16 +414,40 @@ SPIRVType *LLVMToSPIRV::transType(Type *T) {
...
+
+    auto *Struct = BM->openStructType(SPIRVStructNumElements, Name.str());
     mapType(T, Struct);
 
+    if (NumElements > MaxNumElements &&
+        BM->isAllowedToUseExtension(
+            ExtensionID::SPV_INTEL_long_constant_composite)) {
+      uint64_t NumOfContinuedInstructions = NumElements / MaxNumElements - 1;
```
AlexeySachkov, line 432: If `NumElements == MaxNumElements + 1`, then the result of division would be 1, which means that we need zero continued instructions and it doesn't seem to be correct. Am I missing something?

lib/SPIRV/SPIRVWriter.cpp
```diff
@@ -414,16 +414,40 @@ SPIRVType *LLVMToSPIRV::transType(Type *T) {
...
+    }
+
     SmallVector<unsigned, 4> ForwardRefs;
 
     for (unsigned I = 0, E = T->getStructNumElements(); I != E; ++I) {
       auto *ElemTy = ST->getElementType(I);
-      if ((isa<StructType>(ElemTy) || isa<ArrayType>(ElemTy) ||
-           isa<VectorType>(ElemTy) || isa<PointerType>(ElemTy)) &&
-          recursiveType(ST, ElemTy))
+      if (recursiveType(ST, ElemTy))
```
AlexeySachkov, line 450: Why this `if` was updated?

lib/SPIRV/libSPIRV/SPIRVEntry.h
```diff
@@ -857,6 +857,92 @@ template <spv::Op OC> bool isa(SPIRVEntry *E) {
...
+
+protected:
+  void validate() const override;
+  void setWordCount(SPIRVWord WordCount) override {
+    SPIRVEntry::setWordCount(WordCount);
+    Elements.resize(WordCount - 1);
+  }
+  _SPIRV_DCL_ENCDEC
+
+  std::vector<SPIRVId> Elements;
```
AlexeySachkov, line 919: About the `clang-tidy` report: some time ago we turned all warnings into errors, our logic was that if particular check is enabled by our config, then we should follow the rules it enforces. However, we didn't know which exact checks are enabled and how it would affect incoming PRs.

As I can see, the approach with protected class members is widely used in the translator, so we either need to refactor the existing code base or disable the clang-tidy check.

You can ignore this error for now until we made a decision (I will submit an issue for that)

lib/SPIRV/libSPIRV/SPIRVEntry.h
```diff
@@ -857,6 +857,92 @@ template <spv::Op OC> bool isa(SPIRVEntry *E) {
...
+      typename std::enable_if_t<_OC == OpTypeStructContinuedINTEL, T>;
+  template <spv::Op _OC, class T = void>
+  using EnableIfCompositeConst =
+      typename std::enable_if_t<_OC == OpConstantCompositeContinuedINTEL ||
+                                    _OC ==
+                                        OpSpecConstantCompositeContinuedINTEL,
+                                T>;
+  // Complete constructor
+  SPIRVContinuedInstINTELBase(SPIRVModule *M,
+                              const std::vector<SPIRVValue *> TheElements)
```
AlexeySachkov, line 874:
```suggestion
                              const std::vector<SPIRVValue *> &TheElements)
```

lib/SPIRV/libSPIRV/SPIRVEntry.h
```diff
@@ -857,6 +857,92 @@ template <spv::Op OC> bool isa(SPIRVEntry *E) {
...
+      : SPIRVEntryNoId<OC>(M, TheElements.size() + 1) {
+
+    Elements = SPIRVEntry::getIds(TheElements);
+    validate();
+  }
+
+  SPIRVContinuedInstINTELBase(SPIRVModule *M, unsigned NumOfElements)
+      : SPIRVEntryNoId<OC>(M, NumOfElements + 1) {
+    Elements.resize(NumOfElements);
```
AlexeySachkov, line 883:
```suggestion
    Elements.resize(NumOfElements, SPIRVID_INVALID);
```

lib/SPIRV/libSPIRV/SPIRVModule.cpp
```diff
@@ -1069,15 +1078,69 @@ SPIRVValue *SPIRVModuleImpl::addNullConstant(SPIRVType *Ty) {
 
 SPIRVValue *SPIRVModuleImpl::addCompositeConstant(
     SPIRVType *Ty, const std::vector<SPIRVValue *> &Elements) {
+  SPIRVConstantComposite *Res;
+  const size_t MaxNumElements = 65535 - 3;
```
AlexeySachkov, line 1082: Magic numbers again, please outline those into some constants which are shared across the codebase

lib/SPIRV/libSPIRV/SPIRVModule.cpp
```diff
@@ -1069,15 +1078,69 @@ SPIRVValue *SPIRVModuleImpl::addNullConstant(SPIRVType *Ty) {
...
+  const size_t NumElements = Elements.size();
+  if (NumElements > MaxNumElements &&
+      isAllowedToUseExtension(ExtensionID::SPV_INTEL_long_constant_composite)) {
```
AlexeySachkov, line 1085: What about error handling when extension is disabled?

lib/SPIRV/libSPIRV/SPIRVOpCodeEnum.h
```diff
@@ -326,6 +326,9 @@ _SPIRV_OP(GroupNonUniformBitwiseXor, 361)
 _SPIRV_OP(GroupNonUniformLogicalAnd, 362)
 _SPIRV_OP(GroupNonUniformLogicalOr, 363)
 _SPIRV_OP(GroupNonUniformLogicalXor, 364)
+_SPIRV_OP(ConstantCompositeContinuedINTEL, 1025)
+_SPIRV_OP(SpecConstantCompositeContinuedINTEL, 1026)
+_SPIRV_OP(TypeStructContinuedINTEL, 1027)
```
AlexeySachkov, line 331: Token numbers don't look like a correct ones.

lib/SPIRV/libSPIRV/SPIRVType.h
```diff
@@ -656,20 +665,59 @@ class SPIRVTypeStruct : public SPIRVType {
...
+    ContinuedInstructions.push_back(Inst);
+  }
+
+  void encodeChildren(spv_ostream &O) const override {
+    O << SPIRVNL();
+    for (auto &I : ContinuedInstructions)
+      O << *I;
+  }
+
+  // TODO: return a reference here?
```
AlexeySachkov, line 685: Compiler should take care of this automatically: [copy elision](https://en.cppreference.com/w/cpp/language/copy_elision)

lib/SPIRV/libSPIRV/SPIRVType.h
```diff
@@ -656,20 +665,59 @@ class SPIRVTypeStruct : public SPIRVType {
...
     std::vector<SPIRVEntry *> Operands(MemberTypeIdVec.size());
     for (size_t I = 0, E = MemberTypeIdVec.size(); I < E; ++I)
       Operands[I] = getEntry(MemberTypeIdVec[I]);
     return Operands;
   }
+  void addContinuedInstruction(ContinuedInstType Inst) {
+    ContinuedInstructions.push_back(Inst);
+  }
+
+  void encodeChildren(spv_ostream &O) const override {
```
AlexeySachkov, line 679: Where this method is called?

lib/SPIRV/SPIRVWriter.cpp
Fznamznon: It will be emitted through `validate` method of `SPIRVTypeStruct` class.

lib/SPIRV/SPIRVWriter.cpp
Fznamznon: I was looking through the code and tried to simplify it, so I could implement `OpTypeStructContinuedINTEL` as much as possible close to `OpConstantCompositeContinuedINTEL`, but I turned out I couldn't do it in case recursive type is used. However I left this simplification. I can revert it if you prefer.

lib/SPIRV/libSPIRV/SPIRVModule.cpp
Fznamznon: Same here, error will be emitted by `validate` functionality of corresponding class.

lib/SPIRV/libSPIRV/SPIRVOpCodeEnum.h
Fznamznon: D'oh! I remember correcting them, but something went wrong.

lib/SPIRV/libSPIRV/SPIRVType.h
Fznamznon: It is called by `SPIRVEntry::encodeAll` member https://github.com/KhronosGroup/SPIRV-LLVM-Translator/blob/09d5508a737fac6bd491b93840d64957dfa41f7f/lib/SPIRV/libSPIRV/SPIRVEntry.cpp#L189 .
Then `encodeAll` is called when some `SPIRVEntry` is written to a file https://github.com/KhronosGroup/SPIRV-LLVM-Translator/blob/09d5508a737fac6bd491b93840d64957dfa41f7f/lib/SPIRV/libSPIRV/SPIRVEntry.cpp#L458 .

lib/SPIRV/SPIRVWriter.cpp
Fznamznon: In the current version, this last one element will be handled on lines 438-443, where I calculate modulo. I didn't look closely through Mikhail's suggestion, but I believe it will be handled there as well.

lib/SPIRV/SPIRVWriter.cpp
AlexeySotkin: I agree, but `INT16_MAX` must be `UINT16_MAX`

lib/SPIRV/SPIRVWriter.cpp
```diff
@@ -414,16 +414,40 @@ SPIRVType *LLVMToSPIRV::transType(Type *T) {
...
+        auto *Continued = BM->addTypeStructContinuedINTEL(MaxNumElements);
+        Struct->addContinuedInstruction(
+            static_cast<SPIRVTypeStruct::ContinuedInstType>(Continued));
+      }
+      uint64_t Remains = NumElements % MaxNumElements;
+      if (Remains) {
+        auto *Continued = BM->addTypeStructContinuedINTEL(Remains);
+        Struct->addContinuedInstruction(
+            static_cast<SPIRVTypeStruct::ContinuedInstType>(Continued));
+      }
```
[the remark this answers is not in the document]
AlexeySotkin: To me, the code with duplication is more readable.

lib/SPIRV/libSPIRV/SPIRVModule.cpp
```diff
@@ -1069,15 +1078,69 @@ SPIRVValue *SPIRVModuleImpl::addNullConstant(SPIRVType *Ty) {
...
 }
 
+SPIRVEntry *SPIRVModuleImpl::addCompositeConstantContinuedINTEL(
+    const std::vector<SPIRVValue *> &Elements) {
+  return add(new SPIRVConstantCompositeContinuedINTEL(this, Elements));
+}
+
 SPIRVValue *SPIRVModuleImpl::addSpecConstantComposite(
     SPIRVType *Ty, const std::vector<SPIRVValue *> &Elements) {
+  SPIRVSpecConstantComposite *Res;
```
AlexeySotkin, line 1113: I'd prefer to move this variable closer to its usages. Or even get rid of it at all.

lib/SPIRV/SPIRVWriter.cpp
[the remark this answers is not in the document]
Fznamznon: Ok, I left it as is for now.

lib/SPIRV/SPIRVWriter.cpp
```diff
@@ -1224,19 +1248,23 @@ SPIRVValue *LLVMToSPIRV::transValueWithoutDecoration(Value *V,
...
-             << "(65532). Should the array be split?\n Original LLVM value:\n"
-             << toString(GV);
-          getErrorLog().checkError(false, SPIRVEC_InvalidWordCount, SS.str());
+      if (!BM->isAllowedToUseExtension(
+              ExtensionID::SPV_INTEL_long_constant_composite)) {
+        if (auto ArrTy = dyn_cast_or_null<ArrayType>(Init->getType())) {
+          // First 3 words of OpConstantComposite encode: 1) word count &
+          // opcode, 2) Result Type and 3) Result Id. Max length of SPIRV
+          // instruction = 65535 words.
+          const int MaxNumElements = 65535 - 3;
```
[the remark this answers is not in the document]
Fznamznon: Tried to reduce amount of magic numbers.

lib/SPIRV/SPIRVWriter.cpp
```diff
@@ -1224,19 +1248,23 @@ SPIRVValue *LLVMToSPIRV::transValueWithoutDecoration(Value *V,
...
+          if (ArrTy->getNumElements() > MaxNumElements &&
+              !isa<ConstantAggregateZero>(Init)) {
+            std::stringstream SS;
+            SS << "Global variable has a constant array initializer with a "
+                  "number"
+               << " of elements greater than OpConstantComposite can have "
+               << "(65532). Should the array be split?\n Original LLVM value:\n"
+               << toString(GV);
```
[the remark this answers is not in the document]
Fznamznon: Done (with a feedback from clang-format).

lib/SPIRV/libSPIRV/SPIRVModule.cpp
Fznamznon: I'm not sure that it is simpler. Any other opinions?

lib/SPIRV/SPIRVWriter.cpp
Fznamznon: Ok, done.

lib/SPIRV/SPIRVWriter.cpp
Fznamznon: Done.

lib/SPIRV/libSPIRV/SPIRVEntry.h
Fznamznon: Done.

lib/SPIRV/libSPIRV/SPIRVEntry.h
Fznamznon: Done.

lib/SPIRV/libSPIRV/SPIRVModule.cpp
Fznamznon: Done as you suggested.

lib/SPIRV/libSPIRV/SPIRVOpCodeEnum.h
Fznamznon: Fixed.

lib/SPIRV/libSPIRV/SPIRVType.h
Fznamznon: Ok. Removed TODO comment.

lib/SPIRV/libSPIRV/SPIRVModule.cpp
Fznamznon: Moved it closer.

Fznamznon:
I'm not able to re-request review from @mlychkov through the button.

lib/SPIRV/libSPIRV/SPIRVModule.cpp
```diff
@@ -1069,15 +1078,70 @@ SPIRVValue *SPIRVModuleImpl::addNullConstant(SPIRVType *Ty) {
...
+        Res = static_cast<SPIRVConstantComposite *>(
+            addCompositeConstant(Ty, Slice));
+      } else {
+        SPIRVConstantComposite::ContinuedInstType Continued =
+            static_cast<SPIRVConstantComposite::ContinuedInstType>(
+                addCompositeConstantContinuedINTEL(Slice));
+        Res->addContinuedInstruction(Continued);
+      }
+    }
+    return Res;
```
mlychkov, line 1101: Ok, maybe this variant?
```suggestion
    auto Start = Elements.begin();
    auto End = Start + MaxNumElements;
    std::vector<SPIRVValue *> Slice(Start, End);
    auto *Res = static_cast<SPIRVConstantComposite *>(addCompositeConstant(Ty, Slice));
    for (; End != Elements.end();) {
      Start = End;
      End = ((Elements.end() - End) > MaxNumElements) ? End + MaxNumElements
                                                      : Elements.end();
      Slice.assign(Start, End);
      SPIRVConstantComposite::ContinuedInstType Continued =
          static_cast<SPIRVConstantComposite::ContinuedInstType>(
              addCompositeConstantContinuedINTEL(Slice));
      Res->addContinuedInstruction(Continued);
    }
    return Res;
```

mlychkov: Just to:
* remove extra check for J == 0 on each iteration;
* use looping via iterators instead of index

Fznamznon: Ok, I'll give it a try.

lib/SPIRV/SPIRVReader.cpp
```diff
@@ -529,6 +529,9 @@ Type *SPIRVToLLVM::transType(SPIRVType *T, bool IsClassMember) {
     SmallVector<Type *, 4> MT;
     for (size_t I = 0, E = ST->getMemberCount(); I != E; ++I)
       MT.push_back(transType(ST->getMemberType(I), true));
+    for (auto &CI : ST->getContinuedInstructions())
+      for (size_t I = 0, E = CI->getNumElements(); I != E; ++I)
+        MT.push_back(transType(CI->getMemberType(I), true));
```
mlychkov, line 534: May range based loop be used here ?
```suggestion
      for (auto &Elem : CI->getElements())
        MT.push_back(transType(Elem, true));
```

lib/SPIRV/libSPIRV/SPIRVModule.cpp
mlychkov: Maybe it would be helpful to add comment on that before line 1103.

lib/SPIRV/SPIRVReader.cpp
Fznamznon: It cannot be used unless `gelElements` method is added for `SPIRVTypeStructContinuedINTEL` class. I didn't add such thing to make its interface closer to interface of `SPIRVTypeStruct` class.

mlychkov: Ah, sorry, missed enable_if condition.

mlychkov (commented):
Ok, thank you. LGTM.

lib/SPIRV/SPIRVWriter.cpp
AlexeySachkov: >  I can revert it if you prefer.

It just seems to be unrelated to the patch, plus it seems to introduce more calls to `recursiveType` function than it used to be (it will be called for each struct member instead of being called only for composite ones)

lib/SPIRV/SPIRVWriter.cpp
```diff
@@ -414,16 +414,43 @@ SPIRVType *LLVMToSPIRV::transType(Type *T) {
     if (Name == getSPIRVTypeName(kSPIRVTypeName::ConstantPipeStorage))
       return transType(getPipeStorageType(M));
 
-    auto *Struct = BM->openStructType(T->getStructNumElements(), Name.str());
+    const size_t MaxNumElements = MaxWordCount - SPIRVTypeStruct::FixedWC;
```
AlexeySachkov, line 417:
```suggestion
    constexpr size_t MaxNumElements = MaxWordCount - SPIRVTypeStruct::FixedWC;
```

lib/SPIRV/SPIRVWriter.cpp
```diff
@@ -1246,19 +1273,24 @@ SPIRVValue *LLVMToSPIRV::transValueWithoutDecoration(Value *V,
...
-             << "(65532). Should the array be split?\n Original LLVM value:\n"
-             << toString(GV);
-          getErrorLog().checkError(false, SPIRVEC_InvalidWordCount, SS.str());
+      if (!BM->isAllowedToUseExtension(
+              ExtensionID::SPV_INTEL_long_constant_composite)) {
+        if (auto ArrTy = dyn_cast_or_null<ArrayType>(Init->getType())) {
+          // First 3 words of OpConstantComposite encode: 1) word count &
+          // opcode, 2) Result Type and 3) Result Id. Max length of SPIRV
+          // instruction = 65535 words.
+          const int MaxNumElements =
```
AlexeySachkov, line 1282:
```suggestion
          constexpr int MaxNumElements =
```

lib/SPIRV/libSPIRV/SPIRVModule.cpp
```diff
@@ -1069,13 +1078,74 @@ SPIRVValue *SPIRVModuleImpl::addNullConstant(SPIRVType *Ty) {
 
 SPIRVValue *SPIRVModuleImpl::addCompositeConstant(
     SPIRVType *Ty, const std::vector<SPIRVValue *> &Elements) {
-  return addConstant(new SPIRVConstantComposite(this, Ty, getId(), Elements));
+  const int MaxNumElements = MaxWordCount - SPIRVConstantComposite::FixedWC;
```
AlexeySachkov, line 1081:
```suggestion
  constexpr int MaxNumElements = MaxWordCount - SPIRVConstantComposite::FixedWC;
```

lib/SPIRV/libSPIRV/SPIRVModule.cpp
```diff
@@ -1069,13 +1078,74 @@ SPIRVValue *SPIRVModuleImpl::addNullConstant(SPIRVType *Ty) {
...
+  // In case number of elements is greater than maximum WordCount and
+  // SPV_INTEL_long_constant_composite is not enabled, the error will be emitted
+  // by validate functionality of SPIRVCompositeConstant class.
+  if (NumElements <= MaxNumElements ||
+      !isAllowedToUseExtension(ExtensionID::SPV_INTEL_long_constant_composite))
+    return addConstant(new SPIRVConstantComposite(this, Ty, getId(), Elements));
+
+  auto Start = Elements.begin();
+  auto End = Start + MaxNumElements;
+  std::vector<SPIRVValue *> Slice(Start, End);
```
AlexeySachkov, line 1093: Instead of making copies, I think it would be better to use `ArrayRef` here

lib/SPIRV/libSPIRV/SPIRVModule.cpp
```diff
@@ -1069,13 +1078,74 @@ SPIRVValue *SPIRVModuleImpl::addNullConstant(SPIRVType *Ty) {
...
+  auto *Res =
+      static_cast<SPIRVConstantComposite *>(addCompositeConstant(Ty, Slice));
+  for (; End != Elements.end();) {
+    Start = End;
+    End = ((Elements.end() - End) > MaxNumElements) ? End + MaxNumElements
+                                                    : Elements.end();
+    Slice.assign(Start, End);
+    SPIRVConstantComposite::ContinuedInstType Continued =
```
AlexeySachkov, line 1101:
```suggestion
    auto Continued =
```

lib/SPIRV/libSPIRV/SPIRVModule.cpp
```diff
@@ -1069,13 +1078,74 @@ SPIRVValue *SPIRVModuleImpl::addNullConstant(SPIRVType *Ty) {
...
+  auto End = Start + MaxNumElements;
+  std::vector<SPIRVValue *> Slice(Start, End);
+  auto *Res = static_cast<SPIRVSpecConstantComposite *>(
+      addSpecConstantComposite(Ty, Slice));
+  for (; End != Elements.end();) {
+    Start = End;
+    End = ((Elements.end() - End) > MaxNumElements) ? End + MaxNumElements
+                                                    : Elements.end();
+    Slice.assign(Start, End);
+    SPIRVSpecConstantComposite::ContinuedInstType Continued =
```
AlexeySachkov, line 1138:
```suggestion
    auto Continued =
```

lib/SPIRV/libSPIRV/SPIRVModule.cpp
```diff
@@ -1069,13 +1078,74 @@ SPIRVValue *SPIRVModuleImpl::addNullConstant(SPIRVType *Ty) {
...
-  return addConstant(new SPIRVConstantComposite(this, Ty, getId(), Elements));
+  const int MaxNumElements = MaxWordCount - SPIRVConstantComposite::FixedWC;
+  const int NumElements = Elements.size();
+
+  // In case number of elements is greater than maximum WordCount and
+  // SPV_INTEL_long_constant_composite is not enabled, the error will be emitted
+  // by validate functionality of SPIRVCompositeConstant class.
+  if (NumElements <= MaxNumElements ||
+      !isAllowedToUseExtension(ExtensionID::SPV_INTEL_long_constant_composite))
+    return addConstant(new SPIRVConstantComposite(this, Ty, getId(), Elements));
```
AlexeySachkov, line 1089: It seems strange to me that here we might intentionally create "invalid" SPIR-V instruction and it will be checked second time later to emit an error. I would prefer not to create a `SPIRVConstantComposite` with huge amount of continuents if the extension is disabled and just emit an error right away.

However, this seems more like a refactoring, so not asking to do this right now

AlexeySachkov (commented):
A few more (mostly minor) comments. Haven't looked at tests yet as they are pretty huge

AlexeySachkov (commented):
> test/long-constant-composite.ll

As we discussed offline, could you please change the data type used in the test from `float` to `int` so the file will be smaller?

> test/long-spec-const-composite.ll

We can improve the check, I think. Pseudo-code:

```
; CHECK-SPIRV: TypeInt [[TInt:[0-9]+]] 32
; CHECK-SPIRV: Decoration SpecId [[First:[0-9]+]] 1000
; CHECK-SPIRV: Decoration SpecId [[Second:[0-9]+]] 71000
; CHECK-SPIRV: SpecConstant [[TInt]] [[First]]
; CHECK-SPIRV: SpecConstant [[TInt]] [[Second]]
; CHECK-SPIRV: TypeArray [[TArr:[0-9]+]]
; CHECK-SPIRV: 65535 TypeStruct  [[TStruct:[0-9]+]] [[TArr]] [[TInt]] [[TInt]] [[TInt]]
; CHECK-SPIRV-NEXT: 4469 TypeStructContinuedINTEL [[TInt]] [[TInt]] [[TInt]]
; CHECK-SPIRV: 65535 SpecConstantComposite [[TStruct]]
; CHECK-SPIRV-SAME: [[First]]
; CHECK-SPIRV-NEXT: 4470 SpecConstantCompositeContinuedINTEL
; CHECK-SPIRV-SAME: [[Second]]
```

I.e. we can actually check that particular scalar spec constants were used in continued instruction

lib/SPIRV/libSPIRV/SPIRVValue.h
```diff
@@ -287,34 +287,81 @@ class SPIRVUndef : public SPIRVConstantEmpty<OpUndef> {
 
 template <spv::Op OC> class SPIRVConstantCompositeBase : public SPIRVValue {
 public:
+  // There are always 3 words in this instruction except constituents:
+  // 1) WordCount + OpCode
+  // 2) Result type
+  // 3) Result Id
+  const static SPIRVWord FixedWC = 3;
```
mlychkov, line 294:
```suggestion
  constexpr static SPIRVWord FixedWC = 3;
```

lib/SPIRV/libSPIRV/SPIRVType.h
```diff
@@ -619,11 +619,17 @@ class SPIRVTypePipeStorage : public SPIRVType {
 
 class SPIRVTypeStruct : public SPIRVType {
 public:
+  const static Op OC = OpTypeStruct;
+  // There are always 2 words in this instruction except member types:
+  // 1) WordCount + OpCode
+  // 2) Result Id
+  const static SPIRVWord FixedWC = 2;
```
mlychkov, line 626:
```suggestion
  constexpr static SPIRVWord FixedWC = 2;
```

lib/SPIRV/libSPIRV/SPIRVModule.cpp
Fznamznon: Already existing `addCompositeConstant` functionality receives `std::vector` as a parameter since constructor of `SPIRVConstantComposite` instruction receives `std::vector`. There is a conversion of `ArrayRef` to `std::vector` that will do copy anyway, so I'm not sure that it makes sense unless we change interface of `addCompositeConstant` and `SPIRVConstantComposite` class, and I believe such change should be done in a separate patch.

lib/SPIRV/libSPIRV/SPIRVModule.cpp
Fznamznon: Add error checking to tests.

lib/SPIRV/libSPIRV/SPIRVModule.cpp
Fznamznon: Done.

lib/SPIRV/libSPIRV/SPIRVModule.cpp
Fznamznon: Done.

lib/SPIRV/libSPIRV/SPIRVModule.cpp
Fznamznon: Done.

lib/SPIRV/SPIRVWriter.cpp
Fznamznon: Done.

lib/SPIRV/SPIRVWriter.cpp
Fznamznon: Done.

lib/SPIRV/SPIRVWriter.cpp
Fznamznon: Reverted.

lib/SPIRV/libSPIRV/SPIRVValue.h
Fznamznon: Done.

lib/SPIRV/libSPIRV/SPIRVType.h
Fznamznon: Done.

lib/SPIRV/libSPIRV/SPIRVModule.cpp
AlexeySachkov: Suggestion from Mikhail allows to reduce amount of nested code blocks and from my point of view it is readable enough, but I don't have a strong opinion and fine with both ways

lib/SPIRV/libSPIRV/SPIRVModule.cpp
AlexeySachkov: Ok, I'm fine to do this in a separate PR

lib/SPIRV/libSPIRV/SPIRVModule.cpp
Fznamznon: I already applied another variant of this suggestion.

AlexeySachkov (approved):

PR #865, opened 2020-12-23
Translate cmpxchg instruction in case it is followed by store inst

vmaksimo:


vmaksimo:
Tagging @AlexeySachkov @AlexeySotkin

lib/SPIRV/SPIRVRegularizeLLVM.cpp
```diff
@@ -355,6 +359,14 @@ bool SPIRVRegularizeLLVM::regularize() {
...
               Extract->dropAllReferences();
               ToErase.push_back(Extract);
+            } else if (auto *Store = dyn_cast<StoreInst>(U)) {
+              auto *Cmp = new ICmpInst(Store, CmpInst::ICMP_EQ, Res, Comparator,
+                                       "cmpxchg.success");
+              auto *Agg = InsertValueInst::Create(
+                  UndefValue::get(Cmpxchg->getType()), Res, 0, "agg0", Store);
+              auto *AggStruct =
+                  InsertValueInst::Create(Agg, Cmp, 1, "agg1", Store);
+              Store->getOperand(0)->replaceAllUsesWith(AggStruct);
```
Fznamznon, line 369:
```suggestion
              Store->getValueOperand()->replaceAllUsesWith(AggStruct);
```

AlexeySotkin (approved):

lib/SPIRV/SPIRVRegularizeLLVM.cpp
```diff
@@ -355,6 +359,14 @@ bool SPIRVRegularizeLLVM::regularize() {
               assert(Extract->user_empty());
               Extract->dropAllReferences();
               ToErase.push_back(Extract);
+            } else if (auto *Store = dyn_cast<StoreInst>(U)) {
```
AlexeySachkov, line 362: I'm okay with this fix, but I tend to think that the whole loop over `Cmpxchg->users()` should be removed - since we are operating here on LLVM IR, let's just always create desired composite and replace `cmpxchg` instruction with that composite. So, regardless of how it is used further the code will always be correct.

AlexeySotkin: Seems to be non-optimal(extra insert instructions) in some cases, but I agree that it should work, i.e. it is a more robust solution. Should we do it with another PR?

AlexeySachkov: > Should we do it with another PR?

I'm fine with implementing it in a separate PR

test/AtomicCompareExchange.ll
```diff
@@ -40,6 +45,24 @@ cmpxchg.continue:                                 ; preds = %cmpxchg.store_expec
   ret void
 }
 
+; CHECK-SPIRV: FunctionParameter {{[0-9]+}} [[Ptr:[0-9]+]]
+; CHECK-SPIRV: FunctionParameter {{[0-9]+}} [[Store_ptr:[0-9]+]]
+
+; CHECK-SPIRV: AtomicCompareExchange [[Int]] [[Res_1:[0-9]+]] [[Ptr]] [[MemScope_Device]]
+; CHECK-SPIRV-SAME:                  [[MemSemEqual_SeqCst]] [[MemSemUnequal_Acquire]] [[Constant_456]] [[Constant_128]]
+; CHECK-SPIRV: IEqual {{[0-9]+}} [[Success_1:[0-9]+]] [[Res_1]] [[Constant_128]]
+; CHECK-SPIRV: CompositeInsert [[Struct]] [[Composite:[0-9]+]] [[Res_1]] [[UndefStruct]] 0
```
AlexeySotkin, line 54: Does it make sense to use `OpCompositeConstruct` instead?

I'm concerned though, that it will break the reverse translation, because it expects the constituents to be constants, here:
https://github.com/KhronosGroup/SPIRV-LLVM-Translator/blob/ff6d190fd0d838af8eed0ae949d504997a9a2094/lib/SPIRV/SPIRVReader.cpp#L2299-L2320

lib/SPIRV/SPIRVRegularizeLLVM.cpp
AlexeySotkin: Created https://github.com/KhronosGroup/SPIRV-LLVM-Translator/issues/866 to track it.

PR #869, opened 2020-12-30
Lower SIToFPInst with bool operand

vmaksimo:
SIToFPInst LLVM instruction is translated to OpConvertSToF SPIRV instruction.
OpConvertSToF instruction can not receive boolean operand, so replace it with
the select instruction.

### DIFF
--- a/lib/SPIRV/SPIRVLowerBool.cpp
+++ b/lib/SPIRV/SPIRVLowerBool.cpp
@@ -110,6 +110,17 @@ public:
       I.setOperand(0, Sel);
     }
   }
+  virtual void visitSIToFPInst(SIToFPInst &I) {
+    auto Op = I.getOperand(0);
+    if (isBoolType(Op->getType())) {
+      auto Ty = Type::getInt32Ty(*Context);
+      auto Zero = getScalarOrVectorConstantInt(Ty, 0, false);
+      auto One = getScalarOrVectorConstantInt(Ty, 1, false);
+      assert(Zero && One && "Couldn't create constant int");
+      auto Sel = SelectInst::Create(Op, One, Zero, "", &I);
+      I.setOperand(0, Sel);
+    }
+  }
   bool runOnModule(Module &M) override {
     Context = &M.getContext();
     visit(M);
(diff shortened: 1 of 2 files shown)


lib/SPIRV/SPIRVLowerBool.cpp
```diff
@@ -110,6 +110,17 @@ class SPIRVLowerBool : public ModulePass, public InstVisitor<SPIRVLowerBool> {
...
   }
+  virtual void visitSIToFPInst(SIToFPInst &I) {
+    auto Op = I.getOperand(0);
+    if (isBoolType(Op->getType())) {
+      auto Ty = Type::getInt32Ty(*Context);
+      auto Zero = getScalarOrVectorConstantInt(Ty, 0, false);
+      auto One = getScalarOrVectorConstantInt(Ty, 1, false);
+      assert(Zero && One && "Couldn't create constant int");
+      auto Sel = SelectInst::Create(Op, One, Zero, "", &I);
+      I.setOperand(0, Sel);
```
AlexeySachkov, line 121: This code is copy-pasted in this file several times, I suggest that we outline it into some helper function and re-use it in other `visit*` methods

lib/SPIRV/SPIRVLowerBool.cpp
```diff
@@ -77,39 +77,28 @@ class SPIRVLowerBool : public ModulePass, public InstVisitor<SPIRVLowerBool> {
...
-      assert(Zero && One && "Couldn't create constant int");
-      auto Sel = SelectInst::Create(Op, One, Zero, "", &I);
-      I.setOperand(0, Sel);
+      if (Opcode == Instruction::ZExt || Opcode == Instruction::SExt)
+        replace(&I, Sel);
+      else if (Opcode == Instruction::UIToFP || Opcode == Instruction::SIToFP)
+        I.setOperand(0, Sel);
     }
   }
+  virtual void visitZExtInst(ZExtInst &I) { handleCastInstructions(I); }
```
AlexeySachkov, line 98: What about `visitTruncInst`?

vmaksimo: It looks very different from the others. Do we actually need to unify all of the visit functions?

AlexeySachkov (approved):

PR #873, opened 2021-01-14
Fix visitUIToFPInst method in SPIRVLowerBool

NikitaRudenkoIntel:
The method did not handle correctly the case where operand of UIToFPInst instruction is vector. This patch fixes it.

### DIFF
--- a/lib/SPIRV/SPIRVLowerBool.cpp
+++ b/lib/SPIRV/SPIRVLowerBool.cpp
@@ -77,26 +77,35 @@ public:
       replace(&I, Cmp);
     }
   }
-  void handleCastInstructions(Instruction &I) {
+  void handleExtInstructions(Instruction &I) {
     auto Op = I.getOperand(0);
     if (isBoolType(Op->getType())) {
       auto Opcode = I.getOpcode();
-      auto Ty = (Opcode == Instruction::ZExt || Opcode == Instruction::SExt)
-                    ? I.getType()
-                    : Type::getInt32Ty(*Context);
+      auto Ty = I.getType();
       auto Zero = getScalarOrVectorConstantInt(Ty, 0, false);
       auto One = getScalarOrVectorConstantInt(
           Ty, (Opcode == Instruction::SExt) ? ~0 : 1, false);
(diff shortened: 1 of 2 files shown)


vmaksimo:
Please add such case to this test: https://github.com/KhronosGroup/SPIRV-LLVM-Translator/blob/master/test/uitofp-with-bool.ll

NikitaRudenkoIntel:
Done

lib/SPIRV/SPIRVLowerBool.cpp
```diff
@@ -101,8 +101,12 @@ class SPIRVLowerBool : public ModulePass, public InstVisitor<SPIRVLowerBool> {
   }
   virtual void visitUIToFPInst(UIToFPInst &I) {
     auto Op = I.getOperand(0);
-    if (isBoolType(Op->getType())) {
-      auto Ty = Type::getInt32Ty(*Context);
+    auto OpTy = Op->getType();
+    if (isBoolType(OpTy)) {
+      auto *Ty = static_cast<Type *>(Type::getInt32Ty(*Context));
```
svenvh, line 106: You should be able to avoid the `static_cast` by not relying on `auto`:
```suggestion
      Type *Ty = Type::getInt32Ty(*Context);
```

lib/SPIRV/SPIRVLowerBool.cpp
```diff
@@ -101,8 +101,12 @@ class SPIRVLowerBool : public ModulePass, public InstVisitor<SPIRVLowerBool> {
...
+      if (auto VT = dyn_cast<FixedVectorType>(OpTy))
+        Ty =
+            llvm::VectorType::get(Ty, VT->getNumElements(), /*scalable=*/false);
```
svenvh, line 109:
```suggestion
            llvm::FixedVectorType::get(Ty, VT->getNumElements());
```

AlexeySachkov (commented):
Please also rebase on top of #869, because not only `uitofp` misses vectors handling

AlexeySachkov (approved):

PR #878, opened 2021-01-19
Do not truncate subgroup return value

svenvh:
Various SPIR-V subgroup operations return a boolean value, whereas the
equivalent OpenCL builtins return a zero or non-zero integer value.

Until now, the SPIR-V to OpenCL conversion would yield a truncate to
convert from integer to boolean, which is incorrect because any
non-zero integer value should be treated as a "true" value.  Use a
compare for the conversion instead.

Also rename the helper function to reflect better what it does and add
some comments.


AlexeySotkin (approved):

AlexeySachkov:
@aratajew: FYI

PR #880, opened 2021-01-19
[CI] Only check code style in include/ and lib/

svenvh:
I noticed in https://github.com/KhronosGroup/SPIRV-LLVM-Translator/pull/879 that the code style check also runs on .cl test files, which doesn't seem desirable.


AlexeySachkov (approved):
Actually, we might want to apply `clang-tidy` to `.cl` tests just to make sure that they are in a good shape for reading them, but this would require further changes - so this is probably a topic for a separate discussion

The change looks good to me, thanks!

PR #884, opened 2021-01-20
Update LLVM version

MrSidims:


.travis.yml
```diff
@@ -47,7 +47,7 @@ env:
     - MAKEFLAGS="-j2"
     - CHECK_FORMAT=0
     - CHECK_TIDY=0
-    - LLVM_VERSION=11
+    - LLVM_VERSION=11.0.1
```
svenvh, line 50: I think you don't have to change this.

MrSidims: Thanks!

CMakeLists.txt
```diff
@@ -32,7 +32,7 @@ if(LLVM_SPIRV_BUILD_EXTERNAL)
     )
   endif(LLVM_SPIRV_INCLUDE_TESTS)
 
-  find_package(LLVM ${BASE_LLVM_VERSION} REQUIRED
+  find_package(LLVM REQUIRED
```
AlexeySotkin, line 35: Can we specify just major version of LLVM here?
```suggestion
  find_package(LLVM 11 REQUIRED
```
I guess, it will have the same effect as the workaround below.

MrSidims: Looks like it didn't work:
```
CMake Error at CMakeLists.txt:35 (find_package):
  Could not find a configuration file for package "LLVM" that is compatible
  with requested version "11".
  The following configuration files were considered but not accepted:
    /usr/lib/llvm-11/cmake/LLVMConfig.cmake, version: 11.1.0
```

AlexeySachkov: This is a bit unexpected, because according to the [docs](https://cmake.org/cmake/help/v3.12/command/find_package.html), the format is `major[.minor[.patch[.tweak]]]` and moreover, it is should be strictly checked only if `EXACT` keyword is present, which is not the case for us

MrSidims: I see. What about just setting it as 11.1.0? AFAIK it will be released soon.

AlexeySachkov: > What about just setting it as 11.1.0?

This is fine by me

MrSidims: @svenvh @AlexeySotkin what do you agree to set the version as "11.1.0"?

svenvh: No objections if that fixes the issue.

Issue #477, opened 2020-03-31
Add ability to test the translator without clang


AlexeySachkov:
Right now we are using clang via:

https://github.com/KhronosGroup/SPIRV-LLVM-Translator/blob/a4a8f159f5cade95162e3332528fd9a3d0290e14/test/lit.cfg.py#L44

It emits error if clang wasn't found. However, it has an optional argument `required`, which can be set to `False`.

I have WIP branch, which switches this flags and adds `REQUIRED` directives to corresponding `.cl` tests [here](https://github.com/AlexeySachkov/SPIRV-LLVM-Translator/tree/private/asachkov/improve-lit-tests), but local testing shows, that `required=False` is not really supported: `use_clang` crashes somewhere withing it with exception

PR #887, opened 2021-01-27
Update LLVM version from 12 to 13

svenvh:
This might not fully build yet as prebuilt binaries of LLVM 13 are not yet available.


AlexeySachkov:
> This might not fully build yet as prebuilt binaries of LLVM 13 are not yet available.

We have a dependency on `clang` even for in-tree build in order to properly launch some tests, which should be fixed, I think: #477

AlexeySachkov (approved):

PR #895, opened 2021-01-29
Fix access specifiers in debug info

AlexeySotkin:
This commit fixes two issues:

1. FlagIsProtected and FlagIsPrivate had wrong values.

2. Set access flags for class/struct/union members which don't have sccess
   flags set in input LLVM IR, because Clang "optimized" them out in case the
   flags match with default access specifier of the containing type.
   This change also required adjustment of LIT tests to reflect the fact that
   now in DWARF we emit DW_AT_accessibility attribute even for members with
   deafult access specifier.


MrSidims (approved):
LGTM

PR #897, opened 2021-02-01
Treat warnings as errors in out of tree CI builds

svenvh:
Ensure build warnings are not left unnoticed in CI builds.  Only do
this for out of tree builds, such that any warnings from compiling
LLVM sources are not breaking the build unnecessarily.

Fix some unused variable warnings in release builds.


AlexeySachkov (approved):
Thanks!

PR #905, opened 2021-02-04
[FPGA] Add support for constant integers of any bitwidth

mlychkov:
Add translation of constant integers of any bitwidth greater or equal 2.
However, they are still limited to maximum supported number of words in
one SPIR-V instruction. For constants maximum word count for literal
part is 65532. So, max integer bitwidth is 65532*32.


AlexeySotkin:
I think you also should update test/OpSpecConstant.spvasm and `parseSpecConstOpt` in tools/llvm-spirv/llvm-spirv.cpp

mlychkov:
> I think you also should update test/OpSpecConstant.spvasm and `parseSpecConstOpt` in tools/llvm-spirv/llvm-spirv.cpp

This request is aimed only to unblock FPGA to use large _ExtInt numbers. As I understand, spec constants are not used for FPGAs currently. I would propose to create separate issue for enabling support of _ExtInt types in spec constants in the future when it will become necessary.

lib/SPIRV/libSPIRV/SPIRVValue.h
```diff
@@ -175,40 +175,59 @@ template <spv::Op OC> class SPIRVConstantBase : public SPIRVValue {
   float getFloatValue() const { return Union.FloatVal; }
   double getDoubleValue() const { return Union.DoubleVal; }
   unsigned getNumWords() const { return NumWords; }
-  SPIRVWord *getSPIRVWords() { return Union.Words; }
+  const std::vector<SPIRVWord> &getBigValWords() { return BigValWords; }
 
 protected:
+  constexpr static SPIRVWord FixedWC = 3;
+  constexpr static unsigned WordBitWidth = sizeof(SPIRVWord) * CHAR_BIT;
```
AlexeySotkin, line 182: I guess we already have this constant somewhere in this repo.

lib/SPIRV/libSPIRV/SPIRVValue.h
```diff
@@ -175,40 +175,59 @@ template <spv::Op OC> class SPIRVConstantBase : public SPIRVValue {
...
     float FloatVal;
     double DoubleVal;
-    SPIRVWord Words[64];
+    SPIRVWord Words[UnionWordWidth];
     UnionType() { UInt64Val = 0; }
   } Union;
+
+private:
+  // For storing values of any bitwidth greater than 64.
+  std::vector<SPIRVWord> BigValWords;
```
AlexeySotkin, line 230: Is it possible to use it for small integers too and get rid of the union?

lib/SPIRV/SPIRVReader.cpp
```diff
@@ -1552,14 +1552,14 @@ Value *SPIRVToLLVM::transValueWithoutDecoration(SPIRVValue *BV, Function *F,
...
         // is storing data using an array of 64-bit words. Here we pack SPIRV
         // words into 64-bit integer array.
-        for (size_t I = 0; I != RawDataNumWords; ++I)
-          BigValVec[I / 2] =
-              (I % 2) ? BigValVec[I / 2] | ((uint64_t)RawData[I] << 32)
-                      : BigValVec[I / 2] | ((uint64_t)RawData[I]);
+        for (size_t I = 0; I != RawDataNumWords / 2; ++I)
+          BigValVec[I] = ((uint64_t)RawData[2 * I + 1] << 32) | RawData[2 * I];
+        if (RawDataNumWords % 2)
+          BigValVec[BigValNumWords - 1] = RawData[RawDataNumWords - 1];
```
AlexeySachkov, line 1562:
```suggestion
          BigValVec.back() = RawData.back();
```

I'm not sure about the `RawData.back()` though (I just assume you are looking for the last item in the vector)

lib/SPIRV/libSPIRV/SPIRVType.h
```diff
@@ -183,9 +183,8 @@ class SPIRVTypeInt : public SPIRVType {
     SPIRVEntry::validate();
     assert(BitWidth > 1 &&
            (BitWidth <= 64 ||
```
AlexeySachkov, line 185: To be precise it should be
```suggestion
    assert((BitWidth == 8 || BitWidth == 16 || BitWidth == 32 || BitWidth == 64 ||
```

lib/SPIRV/libSPIRV/SPIRVModule.cpp
```diff
@@ -1044,12 +1044,14 @@ SPIRVConstantBase<OC>::SPIRVConstantBase(SPIRVModule *M, SPIRVType *TheType,
...
+  BigValWords.resize(TheValue.getNumWords() * 2);
   for (size_t I = 0; I != TheValue.getNumWords(); ++I) {
-    Union.Words[I * 2 + 1] =
+    BigValWords[I * 2 + 1] =
         (uint32_t)((BigValArr[I] & 0xFFFFFFFF00000000LL) >> 32);
-    Union.Words[I * 2] = (uint32_t)(BigValArr[I] & 0xFFFFFFFFLL);
+    BigValWords[I * 2] = (uint32_t)(BigValArr[I] & 0xFFFFFFFFLL);
   }
   recalculateWordCount();
+  BigValWords.resize(NumWords);
```
AlexeySachkov, line 1054: Why do we need this resize?

lib/SPIRV/SPIRVReader.cpp
```diff
@@ -1552,14 +1552,14 @@ Value *SPIRVToLLVM::transValueWithoutDecoration(SPIRVValue *BV, Function *F,
         const unsigned RawDataNumWords = BConst->getNumWords();
```
AlexeySachkov, line 1552: GitHub doesn't allow to comment on unchanged code even not so far away from the change one, but there is a comment above on line 1551, which should probably be updated: `arbitrary precision integer` -> `huge arbitrary precision integer`

lib/SPIRV/libSPIRV/SPIRVValue.h
AlexeySachkov: The union is needed to be able to quickly reinterpret constant as `float`, `int` or `double`, so it would require some refactoring in methods like `getFloatValue()` to get rid of it.

However, I do agree that we could just put the vector into the union to unify handling of huge integers with the handling of the rest of types

lib/SPIRV/libSPIRV/SPIRVValue.h
```diff
@@ -175,40 +175,59 @@ template <spv::Op OC> class SPIRVConstantBase : public SPIRVValue {
   float getFloatValue() const { return Union.FloatVal; }
   double getDoubleValue() const { return Union.DoubleVal; }
   unsigned getNumWords() const { return NumWords; }
-  SPIRVWord *getSPIRVWords() { return Union.Words; }
+  const std::vector<SPIRVWord> &getBigValWords() { return BigValWords; }
```
AlexeySachkov, line 178: It is not clear to me why this renaming is needed: I still prefer `getSPIRVWords` a s a more generic name, because you might want to do something with instructions words even if it is not a huge constant, but calling `getBigValWords` seems strange in that context.

lib/SPIRV/SPIRVReader.cpp
mlychkov: Done

lib/SPIRV/libSPIRV/SPIRVValue.h
mlychkov: I didn't find it. I created a new one. Please let me know if I missed it somewhere.

lib/SPIRV/libSPIRV/SPIRVValue.h
mlychkov: Refactored code.

lib/SPIRV/SPIRVReader.cpp
mlychkov: Thank you for a good catch. Fixed.

lib/SPIRV/libSPIRV/SPIRVType.h
mlychkov: Done

lib/SPIRV/libSPIRV/SPIRVModule.cpp
mlychkov: Removed it after refactoring.

lib/SPIRV/libSPIRV/SPIRVValue.h
mlychkov: Reverted renaming.

lib/SPIRV/libSPIRV/SPIRVModule.cpp
```diff
@@ -1038,21 +1038,45 @@ SPIRVValue *SPIRVModuleImpl::addConstant(SPIRVType *Ty, uint64_t V) {
   return addConstant(new SPIRVConstant(this, Ty, getId(), V));
 }
 
+template <spv::Op OC> uint64_t SPIRVConstantBase<OC>::getValue() const {
+  uint64_t TheValue = static_cast<uint64_t>(Words[0]);
+  if (NumWords > 1)
+    TheValue |= static_cast<uint64_t>(Words[1]) << SpirvWordBitWidth;
```
AlexeySachkov, line 1044: I suggest adding an assertion that `NumWords` is not greater than 2

lib/SPIRV/libSPIRV/SPIRVValue.h
```diff
@@ -146,69 +146,66 @@ template <spv::Op OC> class SPIRVConstantBase : public SPIRVValue {
...
-    validate();
+    setWords(reinterpret_cast<uint64_t *>(&TheValue));
   }
   // Incomplete constructor
   SPIRVConstantBase() : SPIRVValue(OC), NumWords(0) {}
-  uint64_t getZExtIntValue() const { return Union.UInt64Val; }
-  float getFloatValue() const { return Union.FloatVal; }
-  double getDoubleValue() const { return Union.DoubleVal; }
+  uint64_t getZExtIntValue() const { return getValue(); }
+  float getFloatValue() const { return static_cast<float>(getValue()); }
```
AlexeySachkov, line 169: Not sure if this is correct: `static_cast` will perform conversion operation ([cppreference](https://en.cppreference.com/w/cpp/language/static_cast) - see `1)`, [godbolt](https://godbolt.org/z/s4sn9b)), while you are interested in `bit_cast` (C++20) here.

Other possible options:
```c++
  return reinterpret_cast<float&>(getValue());
  // 6) from https://en.cppreference.com/w/cpp/language/reinterpret_cast
  // breaks type aliasing rules
```

```c++
  float res;
  std::memcpy(res, Words.data(), sizeof(res));
  return res;
  // this is a valid (non-exploiting a UB) option, which is expected to be optimized by compilers down to bitcast without actual memory copy
```

```c++
  union temp {uint64_t ivalue; float fvalue; };
  temp value;
  value.ivalue = getValue();
  return value.fvalue; // this is a UB, even though it works everywhere as expected
  // From https://en.cppreference.com/w/cpp/language/union
  // It's undefined behavior to read from the member of the union that wasn't most recently written.
  // Many compilers implement, as a non-standard language extension, the ability to read inactive members of a union.
```

lib/SPIRV/libSPIRV/SPIRVModule.cpp
mlychkov: Refactored code. No longer relevant.

lib/SPIRV/libSPIRV/SPIRVValue.h
mlychkov: Thank you very much for the tips. I thought about another solution, but static_cast unexpectedly worked even for cast from unit64 to float.

AlexeySachkov (approved):

lib/SPIRV/libSPIRV/SPIRVEnum.h
```diff
@@ -53,6 +54,9 @@ typedef uint32_t SPIRVId;
 #define SPIRVID_MAX ~0U
 #define SPIRVID_INVALID ~0U
 #define SPIRVWORD_MAX ~0U
+static constexpr unsigned SpirvWordSize =
+    static_cast<unsigned>(sizeof(SPIRVWord));
+static constexpr unsigned SpirvWordBitWidth = SpirvWordSize * CHAR_BIT;
```
AlexeySotkin, line 59: Can we explicitly say that word size is 32 bits and add a comment with reference to the spec p 2.2.1
```suggestion
static constexpr unsigned SpirvWordBitWidth = 32;
```

lib/SPIRV/libSPIRV/SPIRVValue.h
```diff
@@ -146,69 +146,74 @@ template <spv::Op OC> class SPIRVConstantBase : public SPIRVValue {
...
-                    llvm::APInt &TheValue);
+                    const llvm::APInt &TheValue);
   // Complete constructor for float constant
   SPIRVConstantBase(SPIRVModule *M, SPIRVType *TheType, SPIRVId TheId,
                     float TheValue)
       : SPIRVValue(M, 0, OC, TheType, TheId) {
-    Union.FloatVal = TheValue;
-    recalculateWordCount();
-    validate();
+    setWords(reinterpret_cast<uint64_t *>(&TheValue));
```
AlexeySotkin, line 158: Can we use `getRawData()` here ?
```suggestion
    setWords(TheValue->getRawData());
```

lib/SPIRV/libSPIRV/SPIRVModule.cpp
```diff
@@ -1038,21 +1038,38 @@ SPIRVValue *SPIRVModuleImpl::addConstant(SPIRVType *Ty, uint64_t V) {
   return addConstant(new SPIRVConstant(this, Ty, getId(), V));
 }
 
+template <spv::Op OC>
```
AlexeySotkin, line 1041: Do we really need this method definition in this file? It seems more logical to have it in SPIRVValue.[h|cpp]
And the same question for the constructor below.

lib/SPIRV/libSPIRV/SPIRVEnum.h
```diff
@@ -43,16 +43,23 @@
 #include "LLVMSPIRVOpts.h"
 #include "SPIRVOpCode.h"
 #include "spirv.hpp"
+#include <climits>
```
AlexeySotkin, line 46: I believe we don't need this header anymore, right?

lib/SPIRV/libSPIRV/SPIRVEnum.h
mlychkov: Done

lib/SPIRV/libSPIRV/SPIRVModule.cpp
mlychkov: Moved to SPIRVValue.cpp

lib/SPIRV/libSPIRV/SPIRVEnum.h
mlychkov: Yes, sure. Thank you.

lib/SPIRV/libSPIRV/SPIRVValue.cpp
```diff
@@ -121,4 +124,36 @@ void SPIRVValue::setFPFastMathMode(SPIRVWord M) {
...
+                                         SPIRVId TheId,
+                                         const llvm::APInt &TheValue)
+    : SPIRVValue(M, 0, OC, TheType, TheId) {
+  setWords(TheValue.getRawData());
+}
+
+// To solve errors about undefined reference to template class methods
+// definitions.
+template class SPIRVConstantBase<OpConstant>;
+template class SPIRVConstantBase<OpSpecConstant>;
```
AlexeySotkin, line 157: Do we still need these lines?

mlychkov: Yes. Otherwise, compilation of libLLVMSPIRVLib.a is failed with 'undefined reference' error for setValue method.

PR #910, opened 2021-02-10
Fix translation of cmpxchg instruction

AlexeySotkin:
Previous implementation optimized produced code assuming a particular
pattern in input LLVM IR. But that assumption is not always correct.
Now we always insert extra instructions to preserve semantics of cmpxchg.
Those extra instruction can be optimized later.

### DIFF
--- a/lib/SPIRV/SPIRVRegularizeLLVM.cpp
+++ b/lib/SPIRV/SPIRVRegularizeLLVM.cpp
@@ -309,7 +309,16 @@ bool SPIRVRegularizeLLVM::regularize() {
           }
         }
         if (auto Cmpxchg = dyn_cast<AtomicCmpXchgInst>(&II)) {
-          Value *Ptr = Cmpxchg->getPointerOperand();
+          // Transform:
+          // %1 = cmpxchg i32* %ptr, i32 %comparator, i32 %0 seq_cst acquire
+          // To:
+          // %cmpxchg.res = call spir_func
+          //   i32 @_Z29__spirv_AtomicCompareExchangePiiiiii(
+          //   i32* %ptr, i32 1, i32 16, i32 2, i32 %0, i32 %comparator)
+          // %cmpxchg.success = icmp eq i32 %cmpxchg.res, %comparator
+          // %1 = insertvalue { i32, i1 } undef, i32 %cmpxchg.res, 0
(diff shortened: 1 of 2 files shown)


AlexeySachkov (approved):

lib/SPIRV/SPIRVRegularizeLLVM.cpp
```diff
@@ -309,14 +309,36 @@ bool SPIRVRegularizeLLVM::regularize() {
           }
         }
         if (auto Cmpxchg = dyn_cast<AtomicCmpXchgInst>(&II)) {
-          Value *Ptr = Cmpxchg->getPointerOperand();
+          // Transform
+          // %1 = cmpxchg i32* %ptr, i32 %comparator, i32 %0 seq_cst acquire
+          // %2 = extractvalue { i32, i1 } %1, 1
```
svenvh, line 314: The code below is not doing anything with `extractvalue` anymore, so perhaps you can remove this line?

lib/SPIRV/SPIRVRegularizeLLVM.cpp
```diff
@@ -309,14 +309,36 @@ bool SPIRVRegularizeLLVM::regularize() {
...
+          // To:
+          // %cmpxchg.res = call spir_func
+          //   i32 @_Z29__spirv_AtomicCompareExchangePiiiiii(
+          //   i32* %ptr, i32 1, i32 16, i32 2, i32 %0, i32 %comparator)
+          // %cmpxchg.success = icmp eq i32 %cmpxchg.res, %comparator
+          // %1 = insertvalue { i32, i1 } undef, i32 %cmpxchg.res, 0
+          // %2 = insertvalue { i32, i1 } %1, i1 %cmpxchg.success, 1
+          // %3 = extractvalue { i32, i1 } %2, 1
```
svenvh, line 322: ... and the same for this line.

lib/SPIRV/SPIRVRegularizeLLVM.cpp
AlexeySotkin: removed

lib/SPIRV/SPIRVRegularizeLLVM.cpp
AlexeySotkin: and here.

svenvh (approved):
LGTM

PR #912, opened 2021-02-10
Accept intrinsic namings into -spirv-allow-unknown-intrinsics

AGindinson:
Previously, enabling the option would let all intrinsics through
the translator. From now on, the option can be configured so
that only specific intrinsic calls with the given prefix are
allowed, while those not matching the prefix will result in an
error as if no option was given.

The allowed intrinsic prefixes are fed into the option as a
comma-separated list, akin to `--spirv-ext`. The list can also
be left unspecified, which, as before, will accept all intrinsic
calls into SPIR-V.

One of the relevant use cases involves llvm.genx.* intrinsics,
which stem from SYCL ESIMD code, but can also mix with common LLVM
intrinsics that stem from regular SYCL code. It would be desirable
to allow ESIMD-specific intrinsics, but still error out on other
unknown ones, To better highlight this case, all ESIMD-specific
tests have been switched to passing an explicit prefix to the
CLI option.

### DIFF
--- a/include/LLVMSPIRVOpts.h
+++ b/include/LLVMSPIRVOpts.h
@@ -39,6 +39,11 @@
 #ifndef SPIRV_LLVMSPIRVOPTS_H
 #define SPIRV_LLVMSPIRVOPTS_H
 
+#include <llvm/ADT/Optional.h>
+#include <llvm/ADT/SmallVector.h>
+#include <llvm/ADT/StringRef.h>
+#include <llvm/IR/IntrinsicInst.h>
+
 #include <cassert>
 #include <cstdint>
 #include <map>
@@ -77,6 +82,7 @@ enum class DebugInfoEIS : uint32_t { SPIRV_Debug, OpenCL_DebugInfo_100 };
 class TranslatorOpts {
 public:
   using ExtensionsStatusMap = std::map<ExtensionID, bool>;
+  using ArgList = llvm::SmallVector<llvm::StringRef, 4>;
 
   TranslatorOpts() = default;
 
@@ -139,13 +145,24 @@ public:
 
   FPContractMode getFPContractMode() const { return FPCMode; }
 
-  bool isSPIRVAllowUnknownIntrinsicsEnabled() const noexcept {
-    return SPIRVAllowUnknownIntrinsics;
+  bool isUnknownIntrinsicAllowed(llvm::IntrinsicInst *II) const noexcept {
+    if (!SPIRVAllowUnknownIntrinsics.hasValue())
+      return false;
+    const auto &IntrinsicPrefixList = SPIRVAllowUnknownIntrinsics.getValue();
+    llvm::StringRef IntrinsicName = II->getCalledOperand()->getName();
+    for (const auto Prefix : IntrinsicPrefixList) {
+      if (IntrinsicName.startswith(Prefix)) // Also true if `Prefix` is empty
+        return true;
+    }
+    return false;
   }
 
-  void
-  setSPIRVAllowUnknownIntrinsicsEnabled(bool AllowUnknownIntrinsics) noexcept {
-    SPIRVAllowUnknownIntrinsics = AllowUnknownIntrinsics;
+  bool isSPIRVAllowUnknownIntrinsicsEnabled() const noexcept {
+    return SPIRVAllowUnknownIntrinsics.hasValue();
+  }
+
+  void setSPIRVAllowUnknownIntrinsics(ArgList IntrinsicPrefixList) noexcept {
+    SPIRVAllowUnknownIntrinsics = IntrinsicPrefixList;
   }
 
   bool allowExtraDIExpressions() const noexcept {
@@ -193,7 +210,7 @@ private:
 
   // Unknown LLVM intrinsics will be translated as external function calls in
   // SPIR-V
-  bool SPIRVAllowUnknownIntrinsics = false;
+  llvm::Optional<ArgList> SPIRVAllowUnknownIntrinsics{};
 
   // Enable support for extra DIExpression opcodes not listed in the SPIR-V
   // DebugInfo specification.
--- a/include/LLVMSPIRVOpts.h
+++ b/include/LLVMSPIRVOpts.h
@@ -39,15 +39,18 @@
 #ifndef SPIRV_LLVMSPIRVOPTS_H
 #define SPIRV_LLVMSPIRVOPTS_H
 
-#include <llvm/ADT/Optional.h>
 #include <llvm/ADT/SmallVector.h>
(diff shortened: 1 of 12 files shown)


AGindinson:
TODO:

- [ ] Add tests.

AGindinson:
@AlexeySachkov, @AlexeySachkov, could you please review the PR?
@DenisBakhvalov, could you please revise the changes in `test/transcoding/AllowIntrinsics.ll` to check if the CLI is acceptable for ESIMD purposes?

tools/llvm-spirv/llvm-spirv.cpp
```diff
@@ -113,6 +113,17 @@ static cl::list<std::string>
...
 
+static cl::list<std::string> SPIRVAllowUnknownIntrinsics(
+    "spirv-allow-unknown-intrinsics", cl::CommaSeparated,
+    cl::desc(
+        "Unknown intrinsics that begin with any prefix from the "
+        "comma-separated "
+        "input list will be translated as external function calls in SPIR-V.\n"
+        "Leaving any prefix unspecified (default) would naturally allow "
+        "all unknown intrinsics"),
+    cl::value_desc("intrinsic_prefix_1,intrinsic_prefix_2>"),
```
AlexeySachkov, line 124:
```suggestion
    cl::value_desc("intrinsic_prefix_1,intrinsic_prefix_2"),
```

tools/llvm-spirv/llvm-spirv.cpp
```diff
@@ -113,6 +113,17 @@ static cl::list<std::string>
...
            cl::ValueRequired);
 
+static cl::list<std::string> SPIRVAllowUnknownIntrinsics(
+    "spirv-allow-unknown-intrinsics", cl::CommaSeparated,
+    cl::desc(
+        "Unknown intrinsics that begin with any prefix from the "
+        "comma-separated "
+        "input list will be translated as external function calls in SPIR-V.\n"
+        "Leaving any prefix unspecified (default) would naturally allow "
+        "all unknown intrinsics"),
```
AlexeySotkin, line 123: I guess it can be formatted better, right?

AGindinson: Fair point; I hope it is now!

tools/llvm-spirv/llvm-spirv.cpp
AGindinson: Done.

test/transcoding/AllowIntrinsics.ll
```diff
@@ -1,27 +1,52 @@
-; The test checks command-line option for the translator
-; which will allow to represent unknown llvm intrinsics as external function call in SPIR-V.
+; The test checks command-line option for the translator which allows to represent
+; unknown llvm intrinsics (or any custom intrinsics) as external function calls in SPIR-V.
```
DenisBakhvalov, line 2: > (or any custom intrinsics)

nit: I would remove the word "any" since we only allow specific custom intrinsics, not all of them.

DenisBakhvalov (commented):
Changes in `test/transcoding/AllowIntrinsics.ll` look good. CLI is OK. Thanks!

test/transcoding/AllowIntrinsics.ll
AGindinson: This is fair for ESIMD use-cases that have facilitated the change, however the Translator change itself may be re-used for other purposes. IMO, this note rather relates to the idea of translating any intrinsics as external function calls, be those LLVM or custom non-LLVM ones, than to how the option filters specific intrinsics out.

That said, I think there's a way to simplify the comment so as to avoid any confusion, so I'll go ahead and employ it :D

AGindinson (commented):
The topic of ABI breakages caused by `LLVMSPIRVOpts.h` changes has been hovering in the air. There are 2 factors contributing to this:
1. newly added LLVM dependencies, which are especially tricky ABI-wise due to frequent LLVM changes;
2. `TranslatorOpts` class layout changes.

I've listed some possible ways to resolve this below, so please highlight if I should proceed with any of these or just leave things as they are.

AlexeySachkov (approved):
We just discussed on OpenCL Tooling teleconference this and considering the facts that we do not claim any ABI stability and such breakages has happened already several times due to similar changes in `TranslatorOpts` object, we decided that there are no objections against merging this as-is.

The only thing we need to do is to make sure that we add some information about ABI breakages to our release notes

include/LLVMSPIRVOpts.h
```diff
@@ -139,13 +145,24 @@ class TranslatorOpts {
 
   FPContractMode getFPContractMode() const { return FPCMode; }
 
+  bool isUnknownIntrinsicAllowed(llvm::IntrinsicInst *II) const noexcept {
```
svenvh, line 148: Since this is a non-trivial implementation, I would prefer to move the definition into a .cpp file.  That would also reduce the number of includes needed.

AGindinson: OK, I'll create a .cpp file for this purpose, and try forward-declaring as much LLVM types as possible.

AGindinson: Done. I was able to get rid of all includes but the `Optional` one (pun intended) due to incomplete-typed field initialization issues.
An additional thought I had was to replace `Optional` with `std::unique_ptr` and use `get()` instead of `hasValue()` and `getValue()`. But that wouldn't seem to gain anything besides the additional smart pointer headache - neither would replacing a relatively small LLVM include with `<memory>` present much advantage.

AGindinson: UPD: also forward-declared `Optional` and moved the implementation file to be linked within `libLLVMSPIRV`.

svenvh: Thanks @AGindinson! I think there's not much more to gain indeed, so this looks good to me.

lib/SPIRV/LLVMSPIRVOpts.cpp
```diff
@@ -0,0 +1,71 @@
+//===- LLVMSPIRVOpts.cpp - Defines LLVM/SPIR-V options ----------*- C++ -*-===//
+//
+//                     The LLVM/SPIR-V Translator
+//
+// This file is distributed under the University of Illinois Open Source
+// License. See LICENSE.TXT for details.
+//
+// Copyright (c) 2014 Advanced Micro Devices, Inc. All rights reserved.
```
AlexeySachkov, line 8:
```suggestion
// Copyright (c) 2021 Intel Corporation. All rights reserved.
```

AlexeySachkov (approved):

svenvh (approved):

PR #913, opened 2021-02-11
Implement SPV_INTEL_optnone extension

Fznamznon:
This extension adds new Function Control mask value to request to not
optimize the corresponding function. It supposed to work as analog of
optnone function attribute in LLVM IR.

The spec can be found under https://github.com/intel/llvm/pull/3198


test/optnone.ll
```diff
@@ -0,0 +1,50 @@
+; RUN: llvm-as %s -o %t.bc
+; RUN: llvm-spirv %t.bc --spirv-ext=+SPV_INTEL_optnone -spirv-text -o - | FileCheck %s --check-prefix=CHECK-SPIRV
+
+; RUN: llvm-spirv --spirv-ext=+SPV_INTEL_optnone %t.bc -o %t.spv
```
AlexeySachkov, line 4:
```suggestion
; RUN: llvm-spirv --spirv-ext=+SPV_INTEL_optnone %t.bc -o %t.spv

; RUN: llvm-spirv %t.bc -spirv-text -o - | FileCheck %s --check-prefix=CHECK-SPIRV
```

I also suggest to add a check that `optnone` is correctly ignored if the corresponding extension is disabled

Fznamznon: Ok, but I don't think that we need to remove SPIR-V checking with extension enabled.

AlexeySotkin (approved):

AlexeySachkov (approved):

PR #916, opened 2021-02-15
Support the SPV_EXT_shader_atomic_float_min_max extension

AGindinson:
As with other atomic instructions, we're looking to translate
an LLVM IR call akin to `__spirv_AtomicFMaxEXT()` into a
corresponding SPIR-V OpCode akin to `AtomicFMaxEXT`.

Extension specification can be found at
https://github.com/KhronosGroup/SPIRV-Registry/blob/master/extensions/EXT/SPV_EXT_shader_atomic_float_min_max.asciidoc


AlexeySotkin (approved):

AlexeySachkov (approved):
It would be great if tests could be simplified further: I guess we don't need that many kernel arguments to see if we can correctly detect a few intrinsics and translate them.

You can try launching `-O1`/`-O2` through `opt` on those LLVM IRs, or you can also simplify source code: I see that those are generated from a SYCL app: try using `SYCL_EXTERNAL` functions instead of the whole kernels

AGindinson:
> It would be great if tests could be simplified further: I guess we don't need that many kernel arguments to see if we can correctly detect a few intrinsics and translate them.
> 
> You can try launching `-O1`/`-O2` through `opt` on those LLVM IRs, or you can also simplify source code: I see that those are generated from a SYCL app: try using `SYCL_EXTERNAL` functions instead of the whole kernels

That's a fair point - I'll attempt using the `SYCL_EXTERNAL` method, since the IR has already been generated with some early optimizations. Since `test/AtomicFAddEXT.ll` also suffers from the "excess of true SYCL", I'd prefer to create a separate PR and cover all 3 tests - would that be acceptable?

AlexeySachkov:
> Since `test/AtomicFAddEXT.ll` also suffers from the "excess of true SYCL", I'd prefer to create a separate PR and cover all 3 tests - would that be acceptable?

Sure, works for me

PR #918, opened 2021-02-19
[NFC] Disable Travis CI on master branch

AlexeySachkov:
We have GitHub Actions configured and there is no need to use two CI
systems at the same time.

The .travis.yml config file is not removed completed yet as we haven't
swithed to use GitHub Actions on release branches - this file will be
kept as some kind of reference for now.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -11,7 +11,6 @@ git:
 
 branches:
   only:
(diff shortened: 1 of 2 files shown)


svenvh (approved):
LGTM.

Issue #920, opened 2021-02-23
Doesn't translate read/write_mem_fence


jekstrand:
The simpler `mem_fence()` built-in is properly translated to a SPIR-V **OpMemoryBarrier** but, as far as I can tell, the directional `read_mem_fence()` and `write_mem_fence()` aren't implemented.

AlexeySachkov:
Hi @jekstrand, thanks for the report

Wow, I was surprised that such built-ins even exist. I guess that the reason they are not handled is because we don't have them in OpenCL CTS (KhronosGroup/OpenCL-CTS#668) and they are rarely (or better say never?) used in other OpenCL apps, but this is indeed a bug of the translator

PR #922, opened 2021-02-24
Implement read/write_mem_fence

jekstrand:
There's no real difference at the SPIR-V level between these and the
more generic mem_fence intrinsic because mem_fence is already relaxed
and SPIR-V doesn't have a concept of anything more relaxed than that.

Closes: #920


jekstrand:
I have no idea what your bugfix process is.  I've made the PR against master but this bug exists on all the older versions.  I know this commit cherry-picks cleanly onto 11.0 because that's what I'm working on today.

svenvh (commented):
LGTM, but could you add a test?  `test/transcoding/OpenCL/mem_fence.cl` seems like a good place to add the tests to, though the new tests don't have to be as exhaustive as the mem_fence case.

> I've made the PR against master but this bug exists on all the older versions. I know this commit cherry-picks cleanly onto 11.0 because that's what I'm working on today.

That's indeed the work flow, we can cherry-pick this after it lands on `master`.

jekstrand:
> LGTM, but could you add a test?

Sure.  I'll have to figure out how to run them but I'd be happy to add a test case.

jekstrand:
I just updated to add a test.  The test passes but, weirdly, somewhere in the compile pipeline my `read_mem_fence` and `write_mem_fence` intrinsic calls are getting turned into `mem_fence` so the LLVM in the test looks weird.  It only does this with my PR, though, and I have no idea why this PR would cause that.

svenvh:
> I just updated to add a test. The test passes but, weirdly, somewhere in the compile pipeline my `read_mem_fence` and `write_mem_fence` intrinsic calls are getting turned into `mem_fence` so the LLVM in the test looks weird. It only does this with my PR, though, and I have no idea why this PR would cause that.

That is because your PR maps all of `read_mem_fence`, `write_mem_fence`, and `mem_fence` to `OpMemoryBarrier`.  After reaching SPIR-V, there is no way to tell what OpenCL builtin function that `OpMemoryBarrier` originated from.  So when translating back from SPIR-V to LLVM, `visitCallSPIRVMemoryBarrier` will produce a `mem_fence` call.

Given that there is no better SPIR-V mapping for `read_mem_fence` and `write_mem_fence`, this looks like the expected result to me.

AlexeySachkov (approved):

svenvh (approved):

PR #924, opened 2021-02-26
[SPIRV] Encode debug info producer in SPIR-V

KornevNikita:
This is a workaround to keep debug info producer information throughout the
translation. We need it because the debug info specification doesn't provide
means to carry this information.


KornevNikita:
@AlexeySotkin, @AlexeySachkov, @MrSidims, @Fznamznon, review, please.

lib/SPIRV/libSPIRV/SPIRVEntry.h
```diff
@@ -950,6 +950,21 @@ template <> struct InstToContinued<OpSpecConstantComposite> {
   constexpr static spv::Op OpCode = OpSpecConstantCompositeContinuedINTEL;
 };
 
+class SPIRVModuleProcessed : public SPIRVEntryNoId<OpModuleProcessed> {
```
Fznamznon, line 953: The spec says this instruction was missing before SPIRV 1.1. Should we implement `getRequiredSPIRVVersion` then?

MrSidims (commented):
LGTM. 
Clang-tidy error can be ignored, but clang-format shall be applied.

lib/SPIRV/libSPIRV/SPIRVEntry.cpp
```diff
@@ -679,4 +679,19 @@ SPIRVType *SPIRVTypeStructContinuedINTEL::getMemberType(size_t I) const {
   return static_cast<SPIRVType *>(SPIRVEntry::getEntry(Elements[I]));
 }
 
+void SPIRVModuleProcessed::validate() const {
+  assert(WordCount == FixedWC + getSizeInWords(ProcessStr));
```
Fznamznon, line 683: I suggest to emit an actual error instead of assertion which is wiped out in release build.
Or, add some clarifying message at least.

AlexeySachkov: I think the plain assertion should be enough here, because if it fails, it means that we have malformed module, which is likely to be "reported" (in form of a segfault probably) somewhere else. I.e. this is a simple enough assumption which don't deserve a runtime check, IMHO.

Fznamznon: Okay, can we give this assert a message then?

AlexeySachkov (approved):

lib/SPIRV/SPIRVToLLVMDbgTran.cpp
```diff
@@ -1000,4 +1000,14 @@ SPIRVToLLVMDbgTran::SplitFileName::SplitFileName(const string &FileName) {
   }
 }
 
+std::string SPIRVToLLVMDbgTran::findModuleProducer() {
```
AlexeySotkin, line 1003: This function should be called in `SPIRVToLLVMDbgTran::transCompileUnit()`, right?

test/DebugInfo/DebugInfoProducer.ll
```diff
@@ -0,0 +1,59 @@
...
+
+!llvm.dbg.cu = !{!0}
+!llvm.module.flags = !{!3, !4, !5, !6}
+!opencl.used.extensions = !{!2}
+!opencl.used.optional.core.features = !{!2}
+!opencl.compiler.options = !{!2}
+!llvm.ident = !{!7}
+
+; CHECK-LLVM: !DICompileUnit
+; CHECK-SAME: producer: "clang version 13.0.0 (https://github.com/llvm/llvm-project.git 16a50c9e642fd085e5ceb68c403b71b5b2e0607c)"
```
AlexeySotkin, line 43:
```suggestion
; CHECK-LLVM-SAME: producer: "clang version 13.0.0 (https://github.com/llvm/llvm-project.git 16a50c9e642fd085e5ceb68c403b71b5b2e0607c)"
```

AlexeySotkin, line 43: Also I think it makes sense to add
```suggestion
; CHECK-LLVM-NOT: producer: "spirv"
```

lib/SPIRV/SPIRVToLLVMDbgTran.cpp
KornevNikita: Indeed, I do not know why it disappeared, I fixed it

test/DebugInfo/DebugInfoProducer.ll
KornevNikita: Thanks, applied

lib/SPIRV/libSPIRV/SPIRVEntry.cpp
```diff
@@ -679,4 +679,20 @@ SPIRVType *SPIRVTypeStructContinuedINTEL::getMemberType(size_t I) const {
   return static_cast<SPIRVType *>(SPIRVEntry::getEntry(Elements[I]));
 }
 
+void SPIRVModuleProcessed::validate() const {
+  assert(WordCount == FixedWC + getSizeInWords(ProcessStr) &&
+         "Incorrect word count in opModuleProcessed");
```
AlexeySotkin, line 684: Minor(nitpick):
```suggestion
         "Incorrect word count in OpModuleProcessed");
```

KornevNikita: Fixed

MrSidims (approved):

PR #926, opened 2021-03-01
Preserve built-in variables in SPIR-V friendly LLVM IR

Fznamznon:
Move transformation of built-in variables to calls to SPIRVToOCL pass to
produce correct SPIR-V friendly LLVM IR which can be recognized by the
translator.

### DIFF
--- a/lib/SPIRV/SPIRVReader.cpp
+++ b/lib/SPIRV/SPIRVReader.cpp
@@ -232,22 +232,6 @@ void SPIRVToLLVM::setAttrByCalledFunc(CallInst *Call) {
   Call->setAttributes(F->getAttributes());
 }
 
-bool SPIRVToLLVM::transOCLBuiltinsFromVariables() {
-  std::vector<GlobalVariable *> WorkList;
-  for (auto I = M->global_begin(), E = M->global_end(); I != E; ++I) {
-    SPIRVBuiltinVariableKind Kind;
-    if (!isSPIRVBuiltinVariable(&(*I), &Kind))
-      continue;
-    if (!transOCLBuiltinFromVariable(&(*I), Kind))
-      return false;
-    WorkList.push_back(&(*I));
-  }
-  for (auto &I : WorkList) {
-    I->eraseFromParent();
-  }
-  return true;
-}
-
 // For integer types shorter than 32 bit, unsigned/signedness can be inferred
 // from zext/sext attribute.
 MDString *SPIRVToLLVM::transOCLKernelArgTypeName(SPIRVFunctionParameter *Arg) {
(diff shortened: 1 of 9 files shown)


Fznamznon:
@AlexeySotkin @AlexeySachkov @MrSidims , please take a look.

MrSidims:
I need to take a more detailed look at the translator code first. In general, I don't see any issues with this move of the code itself (and I agree that this approach is correct, since currently " -r --spirv-target-env=SPV-IR" option works incorrectly). But in the first place I don't understand, why the translator can't recognize ` %3 = call spir_func i32 @_Z22get_max_sub_group_sizev()` generated in builtin_vars_opt.ll test after reversed translation with the mentioned option applied. Do you have any insights?

Fznamznon:
> I need to take a more detailed look at the translator code first. In general, I don't see any issues with this move of the code itself (and I agree that this approach is correct, since currently " -r --spirv-target-env=SPV-IR" option works incorrectly). But in the first place I don't understand, why the translator can't recognize ` %3 = call spir_func i32 @_Z22get_max_sub_group_sizev()` generated in builtin_vars_opt.ll test after reversed translation with the mentioned option applied. Do you have any insights?

Translator recognizes two form of LLVM IR now - the one that comes from OpenCL front-end and SPIR-V friendly LLVM IR. LLVM IR coming from OpenCL contain built-in calls with names like `_Z22get_max_sub_group_sizev`. SPIR-V friendly LLVM IR (https://github.com/KhronosGroup/SPIRV-LLVM-Translator/blob/master/docs/SPIRVRepresentationInLLVM.rst) contains built-ins in a bit other form, in case of `get_max_sub_group_size` built-in SPIR-V friendly LLVM IR contains accesses to a global variable named `__spirv_BuiltInSubgroupMaxSize`. For LLVM IR coming from OpenCL we have a pass named OCLToSPIRV, it transforms built-ins from OCL from to SPIR-V friendly LLVM IR form, so SPIRVWriter sees only SPIR-V friendly LLVM IR form and it is the only from that is recongnized by SPIRVWriter. The problem before my patch was that we produced calls to `_Z30__spirv_BuiltInSubgroupMaxSizev` instead of accesses of `__spirv_BuiltInSubgroupMaxSize` variable and it is not the form that is recognized by SPIRVWriter or OCLToSPIRV pass so these calls wasn't translated to proper instructions if they are translated to SPIR-V again.

MrSidims:
Hm, I just tried (without your patch):
llvm-as builtin_vars_opt.ll
llvm-spirv --spirv-ext=+all builtin_vars_opt.bc
llvm-spirv -r --spirv-target-env=SPV-IR builtin_vars_opt.spv

and I see: `%3 = call spir_func i32 @_Z22get_max_sub_group_sizev()` , which is IMO incorrect (I would expect to see SPIR-V friendly LLVM IR and I expect your patch to fix it). But when I translate it back in SPIR-V I don't see expected  `Decorate 4 BuiltIn 37`, and hence the translator doesn't recognize OpenCL-flavored (but mangled) built-in. I might be talking about another issue tho.

Fznamznon:
> llvm-spirv -r --spirv-target-env=SPV-IR builtin_vars_opt.spv

I just followed your commands and I see calls to `_Z30__spirv_BuiltInSubgroupMaxSizev` (with unmodified translator).
BTW, the test checks SPIR-V friendly IR using filecheck and it passes. Did I get something wrong?

MrSidims:
> > llvm-spirv -r --spirv-target-env=SPV-IR builtin_vars_opt.spv
> 
> I just followed your commands and I see calls to `_Z30__spirv_BuiltInSubgroupMaxSizev` (with unmodified translator).
> BTW, the test checks SPIR-V friendly IR using filecheck and it passes. Did I get something wrong?

Sorry for the confusion - I didn't cleared up testing dir after reversed translation with SPV-IR option...

Fznamznon:
Yes, `_Z22get_max_sub_group_sizev` as a function in SPIR-V seems like a bug. Unrelated to SPIR-V friendly IR though.
In addition, `get_max_sub_group_size` is grepped only in builtin_vars_opt test, so we don't even have a test when such function comes from OCL LLVM IR. We may need to create an issue for this.

MrSidims:
> Unrelated to SPIR-V friendly IR though.

Yeap. Initially I made a hypothesis that this behavior might relate with the bug mentioned in PR description, but now I see, that they are different.

Nuullll:
Hi @Fznamznon, are you still working on this? If not, I'm willing to help :)

Fznamznon:
> Hi @Fznamznon, are you still working on this? If not, I'm willing to help :)

Hi, I'm not working on this. But AFAIK it has been decided to go with calls instead of built-in variables, so this patch is not relevant.

Nuullll:
@Fznamznon Looks like we still need to handle builtin-variable-based SPV-IR (for backward compatibility) in SPIRVToOCL pass.
I'd like to create a new PR if you don't mind.
Also tagging @AlexeySachkov

PR #929, opened 2021-03-03
Use -fdeclare-opencl-builtins for more .cl tests

svenvh:
Since clang commit 85eb12eefdf6 ("[OpenCL] Add declarations with
enum/typedef args", 2021-02-24), Clang provides most of the OpenCL
builtin functions through the faster `-fdeclare-opencl-builtins` flag.
Update various .cl tests to make use of this flag, to speed up running
of the tests and to avoid maintaining excerpts from `opencl-c.h` in
the tests.

### DIFF
--- a/test/event_no_group_cap.cl
+++ b/test/event_no_group_cap.cl
@@ -2,7 +2,7 @@ __kernel void test_fn( const __global char *src)
 {
 	wait_group_events(0, NULL);
 }
-// RUN: %clang_cc1 -triple spir64 -x cl -cl-std=CL2.0 -finclude-default-header -O0 -emit-llvm-bc %s -o %t.bc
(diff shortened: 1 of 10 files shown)


AnastasiaStulova (approved):
LGTM!

Btw why do translator tests use `clang_cc1` instead of `clang`? Otherwise, we could remove the include flags completely. Although I imagine it is not a big deal.

AlexeySachkov (approved):

svenvh:
> Btw why do translator tests use `clang_cc1` instead of `clang`? Otherwise, we could remove the include flags completely. Although I imagine it is not a big deal.

The main reason to use cc1 is because we are passing a few frontend flags and also because the frontend is all we really need for these tests I guess.  But with the Clang driver now including the builtin headers by default, we could think about moving the tests to invoke the Clang driver, if that makes sense?

PR #931, opened 2021-03-04
Remove language check assert

MochalovaAn:
According to specification, OpSource has no semantic impact and can safely be removed from a module. Therefore, such a strict check is not required.

### DIFF
--- a/lib/SPIRV/SPIRVReader.cpp
+++ b/lib/SPIRV/SPIRVReader.cpp
@@ -4379,22 +4379,22 @@ Instruction *SPIRVToLLVM::transOCLBuiltinFromExtInst(SPIRVExtInst *BC,
 void SPIRVToLLVM::transSourceLanguage() {
   SPIRVWord Ver = 0;
   SourceLanguage Lang = BM->getSourceLanguage(&Ver);
-  if (Lang == SourceLanguageUnknown || // Allow unknown for debug info test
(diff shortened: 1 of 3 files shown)


lib/SPIRV/SPIRVReader.cpp
```diff
@@ -4377,26 +4376,25 @@ Instruction *SPIRVToLLVM::transOCLBuiltinFromExtInst(SPIRVExtInst *BC,
...
-  std::tie(Major, Minor, Rev) = decodeOCLVer(Ver);
-  SPIRVMDBuilder Builder(*M);
-  Builder.addNamedMD(kSPIRVMD::Source).addOp().add(Lang).add(Ver).done();
-  // ToDo: Phasing out usage of old SPIR metadata
-  if (Ver <= kOCLVer::CL12)
-    addOCLVersionMetadata(Context, M, kSPIR2MD::SPIRVer, 1, 2);
-  else
-    addOCLVersionMetadata(Context, M, kSPIR2MD::SPIRVer, 2, 0);
+  if (Lang == SourceLanguageUnknown || // Allow unknown for debug info test
+      Lang == SourceLanguageOpenCL_C || Lang == SourceLanguageOpenCL_CPP) {
```
AGindinson, line 4383: I'd suggest using an early `return` on the inverted statement instead.

AGindinson (approved):
LGTM, thanks!

PR #935, opened 2021-03-11
Redesign handling of forward pointers

AlexeySotkin:
In some cases the current implementation doesn't create forward pointers
when they are required. New approach is to create forward pointers
during sorting the type instructions.

TODO:
- [x]  Add a LIT test

### DIFF
--- a/lib/SPIRV/libSPIRV/SPIRVModule.cpp
+++ b/lib/SPIRV/libSPIRV/SPIRVModule.cpp
@@ -249,7 +249,6 @@ public:
   SPIRVTypeQueue *addQueueType() override;
   SPIRVTypePipe *addPipeType() override;
   SPIRVTypeVoid *addVoidType() override;
-  void createForwardPointers() override;
   SPIRVType *addSubgroupAvcINTELType(Op) override;
   SPIRVTypeVmeImageINTEL *addVmeImageINTELType(SPIRVTypeImage *T) override;
   SPIRVTypeBufferSurfaceINTEL *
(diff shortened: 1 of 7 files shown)


AlexeySotkin:
@MrSidims please take a look.

lib/SPIRV/libSPIRV/SPIRVModule.cpp
```diff
@@ -1711,19 +1692,19 @@ class TopologicalSort {
   // the entry itslef.
   void visit(SPIRVEntry *E) {
     DFSState &State = EntryStateMap[E];
+    if (E->getOpCode() == OpTypePointer) {
+      SPIRVTypePointer *Ptr = static_cast<SPIRVTypePointer *>(E);
+      if (EntryStateMap[Ptr->getElementType()] == Discovered) {
```
MrSidims, line 1697: May be it worth to add a comment, why do we insert OpTypeForwardPointer pointer if we found the same OpTypePointer in the module.

lib/SPIRV/libSPIRV/SPIRVModule.cpp
```diff
@@ -1746,8 +1727,16 @@ class TopologicalSort {
   TopologicalSort(const SPIRVTypeVec &TypeVec,
                   const SPIRVConstantVector &ConstVec,
                   const SPIRVVariableVec &VariableVec,
-                  const SPIRVForwardPointerVec &ForwardPointerVec)
-      : ForwardPointerVec(ForwardPointerVec),
+                  SPIRVForwardPointerVec &ForwardPointerVec)
```
MrSidims, line 1730: Why it's no longer a const?

MrSidims (commented):
LGTM

AlexeySachkov (approved):
LGTM, but we indeed could add a comment about how we generate forward pointers as Dmitry suggested

lib/SPIRV/libSPIRV/SPIRVModule.cpp
AlexeySachkov: Because it is now modified in this function, see line 1754

MrSidims: Indeed

lib/SPIRV/libSPIRV/SPIRVModule.cpp
AlexeySotkin: Done

MrSidims (approved):

test/transcoding/ForwardPtr.ll
```diff
@@ -0,0 +1,31 @@
...
+%class.ArgFirst = type { void (%class.A addrspace(4)*)* }
+%structArgSec= type { %class.A addrspace(4)* }
+
+declare spir_func i1 @Caller(%struct.FuncArg addrspace(4)* ) align 2
+
+define spir_func void @MainFunc(%struct.FuncArg addrspace(4)* %context) {
+  entry:
+  %call = call spir_func zeroext i1 @Caller( %struct.FuncArg addrspace(4)* undef) 
+  ret void
+}
```
svenvh, line 31: Missing newline at end of file.

AlexeySotkin: Fixed.

PR #936, opened 2021-03-12
[SPIRV] Encode CodeView checksum information in SPIR-V

KornevNikita:
This is a workaround to keep CodeView checksum information throughout the
translation. The idea is to store checksum kind and value to Ops[TextIdx]
and then parse this string while creating DIFile.

### DIFF
--- a/lib/SPIRV/SPIRVToLLVMDbgTran.cpp
+++ b/lib/SPIRV/SPIRVToLLVMDbgTran.cpp
@@ -46,6 +46,7 @@
 
 #include "llvm/IR/DIBuilder.h"
 #include "llvm/IR/Module.h"
+#include "llvm/ADT/StringExtras.h"
 
 using namespace std;
 using namespace SPIRVDebug::Operand;
@@ -65,11 +66,13 @@ void SPIRVToLLVMDbgTran::addDbgInfoVersion() {
                    DEBUG_METADATA_VERSION);
 }
 
-DIFile *SPIRVToLLVMDbgTran::getDIFile(const string &FileName) {
+DIFile *
+SPIRVToLLVMDbgTran::getDIFile(const std::string &FileName,
+                              Optional<DIFile::ChecksumInfo<StringRef>> CS) {
   return getOrInsert(FileMap, FileName, [=]() {
     SplitFileName Split(FileName);
     if (!Split.BaseName.empty())
-      return Builder.createFile(Split.BaseName, Split.Path);
+      return Builder.createFile(Split.BaseName, Split.Path, CS);
     return static_cast<DIFile *>(nullptr);
   });
 }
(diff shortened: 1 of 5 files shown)


KornevNikita:
@AlexeySotkin @MrSidims @Fznamznon please review

lib/SPIRV/SPIRVToLLVMDbgTran.cpp
```diff
@@ -1011,4 +1015,25 @@ std::string SPIRVToLLVMDbgTran::findModuleProducer() {
   return "spirv";
 }
 
+Optional<DIFile::ChecksumInfo<StringRef>>
+SPIRVToLLVMDbgTran::ParseChecksum(StringRef Text) {
+  // Example of checksum string:
+  // "//__CSK_MD5:7bb56387968a9caa6e9e35fff94eaf7b"
+  Optional<DIFile::ChecksumInfo<StringRef>> CS;
+  auto KindPos = Text.find(SPIRVDebug::ChecksumKindPrefx);
+  if (KindPos != StringRef::npos) {
```
MrSidims, line 1024: As "//__CSK" is a prefix, I suggest to use [startswith](https://llvm.org/doxygen/classllvm_1_1StringRef.html#aeb540aeadd339af860f5c217a85363b7) here.

lib/SPIRV/SPIRVToLLVMDbgTran.cpp
```diff
@@ -1011,4 +1015,25 @@ std::string SPIRVToLLVMDbgTran::findModuleProducer() {
...
+    auto ColonPos = Text.find(":", KindPos);
+    KindPos += string("//__").size();
```
MrSidims, line 1026: I suggest to just drop `//__` from the string instead of calculating the current position to parse. It can be a good in combination with [ltrim](https://llvm.org/doxygen/classllvm_1_1StringRef.html#a9db1f51e975fa287ad0fd6a44c567974), for this algorithm instead of finding ":" in the string we can just call `ltrim(':')` on a string with dropped `//__`.

MrSidims (commented):
Other than LLVM-ish API comments the patch is LGTM

lib/SPIRV/SPIRVToLLVMDbgTran.cpp
KornevNikita: The variable "Text" can start with a different prefix and contain information about the checksum, for example, in the middle, so we need to find ChecksumKindPrefx.

KornevNikita: Added a comment to the code to clarify this point.

MrSidims (approved):

lib/SPIRV/SPIRVToLLVMDbgTran.cpp
AlexeySotkin: Is this what you meant @MrSidims?

MrSidims: Yeap. Since we can't drop `//__` prefix it's LGTM.

AlexeySotkin (approved):

PR #937, opened 2021-03-12
Add HandleLLVMOptions to main CMakeLists.txt

NikitaRudenkoIntel:
MERGE WITHOUT SQUASHING
1. Fix Fix uninitialized variables warnings, which come into place after importing HandleLLVMOptions
2. Import HandleLLVMOptions to main CMakeLists.txt to propagate configuration flags. In particular, this enables Multi-threaded option in MSVC (instead of Multi-threaded DLL)

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -20,6 +20,7 @@ if(LLVM_SPIRV_BUILD_EXTERNAL)
       ${LLVM_SPIRV_VERSION}
     LANGUAGES
       CXX
+      C
   )
 
   set(CMAKE_CXX_STANDARD 14)
@@ -47,6 +48,7 @@ if(LLVM_SPIRV_BUILD_EXTERNAL)
     ${LLVM_CMAKE_DIR}
   )
   include(AddLLVM)
+  include(HandleLLVMOptions)
 
   message(STATUS "Found LLVM: ${LLVM_VERSION}")
 
--- a/lib/SPIRV/SPIRVWriter.cpp
+++ b/lib/SPIRV/SPIRVWriter.cpp
@@ -667,7 +667,7 @@ void LLVMToSPIRV::transVectorComputeMetadata(Function *F) {
     auto ArgNo = I->getArgNo();
     SPIRVFunctionParameter *BA = BF->getArgument(ArgNo);
     if (Attrs.hasAttribute(ArgNo + 1, kVCMetadata::VCArgumentIOKind)) {
-      SPIRVWord Kind;
+      SPIRVWord Kind = {};
       Attrs.getAttribute(ArgNo + 1, kVCMetadata::VCArgumentIOKind)
           .getValueAsString()
           .getAsInteger(0, Kind);
@@ -1325,7 +1325,7 @@ SPIRVValue *LLVMToSPIRV::transValueWithoutDecoration(Value *V,
     if (IsVectorCompute) {
       BVar->addDecorate(DecorationVectorComputeVariableINTEL);
       if (GV->hasAttribute(kVCMetadata::VCByteOffset)) {
-        SPIRVWord Offset;
+        SPIRVWord Offset = {};
         GV->getAttribute(kVCMetadata::VCByteOffset)
             .getValueAsString()
             .getAsInteger(0, Offset);
--- a/lib/SPIRV/libSPIRV/SPIRVUtil.h
+++ b/lib/SPIRV/libSPIRV/SPIRVUtil.h
@@ -86,7 +86,7 @@ public:
   void init();
 
   static Ty2 map(Ty1 Key) {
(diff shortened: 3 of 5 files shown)


MrSidims (approved):

CMakeLists.txt
```diff
@@ -20,6 +20,7 @@ if(LLVM_SPIRV_BUILD_EXTERNAL)
       ${LLVM_SPIRV_VERSION}
     LANGUAGES
       CXX
+      C
```
AlexeySotkin, line 23: Why do we need C language here?

lib/SPIRV/SPIRVWriter.cpp
```diff
@@ -669,7 +669,7 @@ void LLVMToSPIRV::transVectorComputeMetadata(Function *F) {
     auto ArgNo = I->getArgNo();
     SPIRVFunctionParameter *BA = BF->getArgument(ArgNo);
     if (Attrs.hasAttribute(ArgNo + 1, kVCMetadata::VCArgumentIOKind)) {
-      SPIRVWord Kind;
+      SPIRVWord Kind = SPIRVWord{};
```
AlexeySotkin, line 672: Will it work if we do `SPIRVWord Kind = {};` instead?
The same question for initializations below.

AlexeySotkin:
@NikitaRudenkoIntel could you provide example of warnings you are trying to fix?

NikitaRudenkoIntel:
First commit fixes warnings like these one (they appear after HandleLLVMOptions introduction):
```
/users/nrudenko/LLVMv/llvm-project/llvm/projects/SPIRV-LLVM-Translator/lib/SPIRV/OCLUtil.h:443:78: error: 'Val' may be used uninitialized in this function [-Werror=maybe-uninitialized]
   return OCLMemOrderMap::map(Order) | mapBitMask<OCLMemFenceMap>(MemFenceFlag);
                                                                              ^
In file included from /users/nrudenko/LLVMv/llvm-project/llvm/projects/SPIRV-LLVM-Translator/lib/SPIRV/libSPIRV/SPIRVOpCode.h:43:0,
                 from /users/nrudenko/LLVMv/llvm-project/llvm/projects/SPIRV-LLVM-Translator/lib/SPIRV/libSPIRV/SPIRVEnum.h:44,
                 from /users/nrudenko/LLVMv/llvm-project/llvm/projects/SPIRV-LLVM-Translator/lib/SPIRV/SPIRVInternal.h:43,
                 from /users/nrudenko/LLVMv/llvm-project/llvm/projects/SPIRV-LLVM-Translator/lib/SPIRV/OCLUtil.h:42,
                 from /users/nrudenko/LLVMv/llvm-project/llvm/projects/SPIRV-LLVM-Translator/lib/SPIRV/OCLUtil.cpp:40:
/users/nrudenko/LLVMv/llvm-project/llvm/projects/SPIRV-LLVM-Translator/lib/SPIRV/libSPIRV/SPIRVUtil.h:89:9: note: 'Val' was declared here
     Ty2 Val;
         ^~~
```

lib/SPIRV/SPIRVWriter.cpp

NikitaRudenkoIntel:
Second commit fixes link issues on Windows platform.
LLVM has all its projects set to "Multi-threaded" (/MT) option.
Without importing HandleLLVMOptions, in out-of-tree build, cmake will generate "Multi-threaded DLL" (/MD) option for LLVMSPIRVLib project. This will not allow to link any LLVM project with prebuilt LLVMSPIRVLib.

CMakeLists.txt
NikitaRudenkoIntel: As I understand, it is just the requirement for HandleLLVMOptions

https://github.com/llvm/llvm-project/blob/main/llvm/cmake/modules/HandleLLVMOptions.cmake#L11

lib/SPIRV/libSPIRV/SPIRVUtil.h
```diff
@@ -86,15 +86,15 @@ template <class Ty1, class Ty2, class Identifier = void> struct SPIRVMap {
   void init();
 
   static Ty2 map(Ty1 Key) {
-    Ty2 Val;
+    Ty2 Val = {};
```
AlexeySachkov, line 89: It is not entirely clear to me what was the problem here, but ok, I probably just don't know something about C++

AlexeySachkov (approved):

lib/SPIRV/libSPIRV/SPIRVUtil.h
NikitaRudenkoIntel: The problem is uninitialized `Val`, which is treated as error after importing HandleLLVMOptions

svenvh (approved):

PR #939, opened 2021-03-18
Lower atomic FP __spirv builtins into OpenCL 1.2 builtins

AGindinson:
In backwards translation, we look to treat FP-typed atomic functions
from multi-vendor extensions similarly to "regular" atomics.

### DIFF
--- a/lib/SPIRV/SPIRVToOCL.h
+++ b/lib/SPIRV/SPIRVToOCL.h
@@ -182,6 +182,9 @@ public:
   /// using separate maps for OpenCL 1.2 and OpenCL 2.0
   virtual Instruction *mutateAtomicName(CallInst *CI, Op OC) = 0;
 
+  // Transform FP atomic opcode to corresponding OpenCL function name
+  virtual std::string FPAtomicToOCLBuiltinName(Op OC) = 0;
+
   /// Transform uniform group opcode to corresponding OpenCL function name,
   /// example: GroupIAdd(Reduce) => group_iadd => work_group_reduce_add |
   /// sub_group_reduce_add
@@ -261,6 +264,9 @@ public:
   /// Transform atomic builtin name into correct ocl-dependent name
   Instruction *mutateAtomicName(CallInst *CI, Op OC) override;
 
+  // Transform FP atomic opcode to corresponding OpenCL function name
+  std::string FPAtomicToOCLBuiltinName(Op OC) override;
+
   /// Transform SPIR-V atomic instruction opcode into OpenCL 1.2 builtin name.
   /// Depending on the type, the return name starts with "atomic_" for 32-bit
   /// types or with "atom_" for 64-bit types, as specified by
@@ -320,6 +326,9 @@ public:
 
   /// Transform atomic builtin name into correct ocl-dependent name
   Instruction *mutateAtomicName(CallInst *CI, Op OC) override;
+
+  // Transform FP atomic opcode to corresponding OpenCL function name
+  std::string FPAtomicToOCLBuiltinName(Op OC) override;
 
   /// Transform __spirv_OpAtomicCompareExchange/Weak into
   /// compare_exchange_strong/weak_explicit
--- a/lib/SPIRV/SPIRVToOCL.h
+++ b/lib/SPIRV/SPIRVToOCL.h
@@ -183,7 +183,7 @@ public:
   virtual Instruction *mutateAtomicName(CallInst *CI, Op OC) = 0;
 
   // Transform FP atomic opcode to corresponding OpenCL function name
-  virtual std::string FPAtomicToOCLBuiltinName(Op OC) = 0;
+  virtual std::string mapFPAtomicName(Op OC) = 0;
 
   /// Transform uniform group opcode to corresponding OpenCL function name,
   /// example: GroupIAdd(Reduce) => group_iadd => work_group_reduce_add |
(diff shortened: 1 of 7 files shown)


lib/SPIRV/libSPIRV/SPIRVOpCode.h
```diff
@@ -62,7 +62,9 @@ inline bool isAtomicOpCode(Op OpCode) {
   static_assert(OpAtomicLoad < OpAtomicXor, "");
   return ((unsigned)OpCode >= OpAtomicLoad &&
           (unsigned)OpCode <= OpAtomicXor) ||
-         OpCode == OpAtomicFlagTestAndSet || OpCode == OpAtomicFlagClear;
+         OpCode == OpAtomicFlagTestAndSet || OpCode == OpAtomicFlagClear ||
```
MrSidims, line 65: `atomic_flag_test_and_set/clear` aren't part of OpenCL 1.2 tho (https://www.khronos.org/registry/OpenCL/sdk/1.2/docs/man/xhtml/atomicFunctions.html and https://www.khronos.org/registry/OpenCL/sdk/2.0/docs/man/xhtml/atomic_flag_test_and_set.html ) (just challenging PR name correctness). Also as you aren't changing any tests appropriate to `atomic_flag_test_and_set/clear` instructions it makes me thinking, that they are either already being translated without your change, or they aren't being translated in the translator at all.

test/transcoding/AtomicFAddEXT.ll
```diff
@@ -20,29 +20,30 @@ define dso_local spir_func float @_Z14AtomicFloatIncRf(float addrspace(4)* align
 entry:
   %0 = addrspacecast float addrspace(4)* %Arg to float addrspace(1)*
   ; CHECK-SPIRV: 7 AtomicFAddEXT [[TYPE_FLOAT_32]]
-  ; CHECK-LLVM: call spir_func float @[[FLOAT_FUNC_NAME:_Z21__spirv_AtomicFAddEXT[[:alnum:]]+]]({{.*}})
+  ; CHECK-LLVM: call spir_func float @[[FLOAT_FUNC_NAME:_Z10atomic_add[[:alnum:]]+ff]]({{.*}})
```
AlexeySotkin, line 23: I suppose we should be able to consume this variant of LLVM IR in OpenCL 1.2 environment. But there are no `atomic_[add|min|max]` builtins with floating point arguments in the OpenCL C 1.2 spec. Probably, we need a corresponding OpenCL extensions for `SPV_EXT_shader_atomic_float_add` and `SPV_EXT_shader_atomic_float_min_max`

AlexeySotkin (commented):
Can we also have a test with ` --spirv-target-env=SPV-IR` ?

lib/SPIRV/SPIRVToOCL.h
```diff
@@ -182,6 +182,9 @@ class SPIRVToOCLBase : public InstVisitor<SPIRVToOCLBase> {
   /// using separate maps for OpenCL 1.2 and OpenCL 2.0
   virtual Instruction *mutateAtomicName(CallInst *CI, Op OC) = 0;
 
+  // Transform FP atomic opcode to corresponding OpenCL function name
+  virtual std::string FPAtomicToOCLBuiltinName(Op OC) = 0;
```
AlexeySachkov, line 186: In order to fix `clang-tidy` issue you can rename this function into something like `mapFPAtomicName` to be closer to existing `mapAtomicName` which you modify here

lib/SPIRV/SPIRVToOCL12.cpp
```diff
@@ -263,6 +278,9 @@ Instruction *SPIRVToOCL12Base::mutateAtomicName(CallInst *CI, Op OC) {
 std::string SPIRVToOCL12Base::mapAtomicName(Op OC, Type *Ty) {
   std::string Prefix = Ty->isIntegerTy(64) ? kOCLBuiltinName::AtomPrefix
                                            : kOCLBuiltinName::AtomicPrefix;
+  // Map fp atomic instructions to regular OpenCL built-ins.
+  if (isFPAtomicOpCode(OC))
+    return FPAtomicToOCLBuiltinName(OC);
   return Prefix += OCL12SPIRVBuiltinMap::rmap(OC);
```
AlexeySachkov, line 284: BTW, why can't we just extend `OCL12SPIRVBuiltinMap`?

AGindinson: The problem with extending the `SPIRVMap<>` is caused by the fact that only one entry is allowed for each function name - mapping `add` to `AtomicFAddEXT` results in the overwrite of the `AtomicIAdd` entry.

Maybe we should define and use `atomic_f[add/min/max]` naming instead, similarly to what is done for `umax`/`umin`? @Fznamznon, @AlexeySotkin, what do you think here?

AGindinson: As discussed offline, we will retain `atomic_[add/min/max]` naming for now, working out a unified strategy for float/unsigned atomic functions later.

lib/SPIRV/SPIRVToOCL.h
AGindinson: Done, thanks for the suggestion!

AlexeySachkov (approved):

AlexeySotkin (approved):

PR #949, opened 2021-03-23
Fix mangling of OpAtomicUMax/OpAtomicUMin representation in SPV-IR

aratajew:


lib/SPIRV/SPIRVUtil.cpp
```diff
@@ -1634,6 +1634,12 @@ class SPIRVFriendlyIRMangleInfo : public BuiltinFuncMangleInfo {
       setArgAttr(0, SPIR::ATTR_CONST);
       addUnsignedArg(0);
       break;
+    case OpAtomicUMax:
+      LLVM_FALLTHROUGH;
+    case OpAtomicUMin:
+      addUnsignedArg(0);
+      addUnsignedArg(3);
```
Fznamznon, line 1641: Actually I can't find where the spec says explicitly that _Pointer_ and _Value_ should be unsigned. For `OpAtomicUMax` it says
> Result Type must be an integer type scalar.
The type of Value must be the same as Result Type. The type of the value pointed to by Pointer must be the same as Result Type.

But it also says
> get a New Value by finding the largest unsigned integer of Original Value and Value

I guess it adds an assumption that _Value_ should be unsigned.
See https://www.khronos.org/registry/spir-v/specs/unified1/SPIRV.html#OpAtomicUMax 

Do you think it is a bug in spec?

aratajew: > I guess it adds an assumption that Value should be unsigned.

That's how I interpreted this. You're right, specification could be more precise.

aratajew: Is the ambiguity in the specification a blocker for this commit?

AlexeySotkin: In SPIR-V modules with Kernel capability all integers types have Signedness operand = 0.

> 0 indicates unsigned, or no signedness semantics

In this case I think letter **U** in the name of the intrinsic implies that integers must be treated as unsigned. (while OpAtomic**S**Max treats integer operands as signed)

AlexeySotkin (approved):

Fznamznon (approved):

PR #950, opened 2021-03-23
Add support for SPV_KHR_linkonce_odr

Fznamznon:
This extension adds a new Linkage Type LinkOnceODR. A global variable or
function with such linkage type can be merged with equivalent global variables
or functions from other modules during the linking process. The primary use case
of this linkage type is to satisfy the C++ language One Definition Rule.

The spec
https://github.com/KhronosGroup/SPIRV-Registry/blob/master/extensions/KHR/SPV_KHR_linkonce_odr.asciidoc


Fznamznon:
@AlexeySotkin @AlexeySachkov @MrSidims please take a look.

lib/SPIRV/SPIRVReader.cpp
```diff
@@ -4468,9 +4468,11 @@ std::string SPIRVToLLVM::getOCLGenericCastToPtrName(SPIRVInstruction *BI) {
 
 llvm::GlobalValue::LinkageTypes
 SPIRVToLLVM::transLinkageType(const SPIRVValue *V) {
-  if (V->getLinkageType() == internal::LinkageTypeInternal) {
+  SPIRVLinkageTypeKind LT = V->getLinkageType();
+  if (LT == internal::LinkageTypeInternal)
```
AlexeySotkin, line 4472: Can we check this linkage type with a `case` in the `switch(LT)` statement below?

Fznamznon: Checking this in the switch statement gives a warning because `LinkageTypeInternal` is not a part of `SPIRVLinkageTypeKind` enum. It is defined in spirv_internal, not in spirv.hpp like other linkage types. I see this:
```
SPIRVReader.cpp:4475:3: warning: case value ‘2147483646’ not in enumerated type ‘SPIRV::SPIRVLinkageTypeKind {aka spv::LinkageType}’ [-Wswitch]
   case internal::LinkageTypeInternal:
```

MrSidims: As w/a I can suggest to cast LT to `int` and switch over it. No warnings shall be emitted.

MrSidims (approved):
LGTM

lib/SPIRV/SPIRVReader.cpp
Fznamznon: Okay, done.

AlexeySotkin (approved):

Fznamznon:
Ping.

AlexeySachkov (approved):

svenvh (approved):

PR #938, opened 2021-03-15
Fix llvm-spirv crash when count of Fortran metadata variables is an array

MochalovaAn:
Handle the case when Count of Fortran metadata variables is an array

Signed-off-by: amochalo <anastasiya.mochalova@intel.com>

PR #951, opened 2021-03-23
Fix crash at translation of Entity of DebugImportedEntity

KornevNikita:
This is workaround to fix crash, which caused by using dyn_cast on a nullptr,
as the result of translation of Entity operand of DebugImportedEntity, because
this operand is DebugInfoNone. Because on LLVM to SPIR-V translation stage
the entity was DIModule, which is currently not supported by the SPIR-V
debug info spec(and by the translator).


lib/SPIRV/SPIRVToLLVMDbgTran.cpp
```diff
@@ -761,6 +761,9 @@ DINode *SPIRVToLLVMDbgTran::transImportedEntry(const SPIRVExtInst *DebugInst) {
   DIFile *File = getFile(Ops[SourceIdx]);
   auto *Entity = transDebugInst<DINode>(BM->get<SPIRVExtInst>(Ops[EntityIdx]));
   if (Ops[TagIdx] == SPIRVDebug::ImportedModule) {
+    if (!Entity)
+      return Builder.createImportedModule(
+          Scope, static_cast<DIImportedEntity *>(nullptr), File, Line);
```
Fznamznon, line 766: Should this exit be done out of the `if` statement?
I see that `Entity` also `dyn_cast`ed on line 775 inside the next `if`.

KornevNikita: No, because this crash is observed when Ops[TagIdx] == SPIRVDebug::ImportedModule

KornevNikita:
@AlexeySotkin take a look, please.

lib/SPIRV/SPIRVToLLVMDbgTran.cpp
MrSidims: I think, that crash is happening due to `dyn_cast` can't work with nullptr input. If we just replace `dyn_cast` with [dyn_cast_or_null](https://llvm.org/docs/ProgrammersManual.html#the-isa-cast-and-dyn-cast-templates) will it help? Or we actually have to change the logic of `transImportedEntry`?

AlexeySotkin: In theory `Entity` can be `nullptr` for `Ops[TagIdx] == SPIRVDebug::ImportedDeclaration` too. But I think we should handle this case differently. Lets keep it as is for now?

MrSidims: Okay

MrSidims (approved):

AlexeySotkin (approved):

AlexeySotkin:
I think for this PR we can take the LIT test from #938. We need to remove debug info for local variable "a" though.

Fznamznon (approved):

test/DebugInfo/DebugInfoNoneEntity.ll
```diff
@@ -0,0 +1,42 @@
...
+!24 = distinct !DISubprogram(name: "declare_target_subroutine", linkageName: "MAIN__", scope: !10, file: !10, line: 23, type: !25, scopeLine: 23, spFlags: DISPFlagDefinition | DISPFlagMainSubprogram, unit: !9, retainedNodes: !27)
+!25 = !DISubroutineType(types: !26)
+!26 = !{null}
+!27 = !{!28, !29}
+!28 = !DILocalVariable(name: "sum", scope: !24, file: !10, line: 29, type: !19)
+!29 = !DILocalVariable(name: "x", scope: !24, file: !10, line: 29, type: !19)
+!31 = !DICompositeType(tag: DW_TAG_array_type, baseType: !19, elements: !32, dataLocation: !DIExpression(DW_OP_push_object_address, DW_OP_deref), associated: !DIExpression(DW_OP_push_object_address, DW_OP_deref, DW_OP_constu, 0, DW_OP_or))
+!32 = !{!33}
+!33 = !DISubrange(lowerBound: !DIExpression(DW_OP_push_object_address, DW_OP_plus_uconst, 64, DW_OP_deref), upperBound: !DIExpression(DW_OP_push_object_address, DW_OP_plus_uconst, 64, DW_OP_deref, DW_OP_push_object_address, DW_OP_plus_uconst, 48, DW_OP_deref, DW_OP_plus, DW_OP_constu, 1, DW_OP_minus), stride: !DIExpression(DW_OP_push_object_address, DW_OP_plus_uconst, 56, DW_OP_deref))
+!34 = !DIModule(scope: !24, name: "iso_fortran_env", isDecl: true)
```
mlychkov, line 42: A nit just to simplify the test example:
```suggestion
!llvm.module.flags = !{!1, !2, !3, !4}
!llvm.dbg.cu = !{!5}

!0 = !{}
!1 = !{i32 1, !"wchar_size", i32 4}
!2 = !{i32 7, !"PIC Level", i32 2}
!3 = !{i32 2, !"Debug Info Version", i32 3}
!4 = !{i32 2, !"Dwarf Version", i32 4}
!5 = distinct !DICompileUnit(language: DW_LANG_Fortran95, file: !6, producer: " F90 Flang -Beating Intel(R) Fortran 21.0-2297", isOptimized: false, runtimeVersion: 0, emissionKind: FullDebug, enums: !0, globals: !0, imports: !7, splitDebugInlining: false, nameTableKind: None)
!6 = !DIFile(filename: "declare_target_subroutine.F90", directory: "/test")
!7 = !{!8}
!8 = !DIImportedEntity(tag: DW_TAG_imported_module, scope: !9, entity: !12, file: !6, line: 24)
!9 = distinct !DISubprogram(name: "declare_target_subroutine", linkageName: "MAIN__", scope: !6, file: !6, line: 23, type: !10, scopeLine: 23, spFlags: DISPFlagDefinition | DISPFlagMainSubprogram, unit: !5, retainedNodes: !0)
!10 = !DISubroutineType(types: !11)
!11 = !{null}
!12 = !DIModule(scope: !9, name: "iso_fortran_env", isDecl: true)
```

mlychkov:
Could you clarify why it is a workaround? Is it some bug in another place that should be addressed later?

KornevNikita:
> Could you clarify why it is a workaround? Is it some bug in another place that should be addressed later?

I quote @AlexeySotkin - "Entity operand is DebugInfoNone because on LLVM to SPIR-V translation stage the entity was DIModule, which is currently not supported by the SPIR-V debug info spec(and by the translator)"

mlychkov:
> > Could you clarify why it is a workaround? Is it some bug in another place that should be addressed later?
> 
> I quote @AlexeySotkin - "Entity operand is DebugInfoNone because on LLVM to SPIR-V translation stage the entity was DIModule, which is currently not supported by the SPIR-V debug info spec(and by the translator)"

Ok, I see. Then it would be nice to include this explanation into the commit message.

KornevNikita:
> > > Could you clarify why it is a workaround? Is it some bug in another place that should be addressed later?
> > 
> > 
> > I quote @AlexeySotkin - "Entity operand is DebugInfoNone because on LLVM to SPIR-V translation stage the entity was DIModule, which is currently not supported by the SPIR-V debug info spec(and by the translator)"
> 
> Ok, I see. Then it would be nice to include this explanation into the commit message.

Thanks, done

KornevNikita:
Wrong button

PR #954, opened 2021-03-25
[Backport] Try to derive the size of vectors when translating DICompo…

aparshin-intel:
…siteType

SPIRV translator does not have the facilities to preserve the size specified
in the DICompositeType node for DW_TAG_array_type.  Instead, it relies on the
size of the underlying base type to calculate the intended size. In
cases when the underlying base type is typedef - this size can`t be
restored. This commit introduces a way to recursively traverse the
definition of the base type to restore the original size


svenvh:
This is included in #986.

PR #958, opened 2021-03-26
Backmerge fix for SPV-IR mangling of OpAtomicUMax/OpAtomicUMin to llvm_release_110

aratajew:


svenvh:
The first commit in here will be part of #967; it would help if we could postpone merging this.

svenvh:
This is included in #986.

PR #963, opened 2021-03-30
Fix SPIR-V friendly IR for OpGroupAsyncCopy and OpGroupWaitEvents

Fznamznon:
These instructions were translated directly to OCL even if SPIR-V
friendly LLVM IR was requested.

### DIFF
--- a/lib/SPIRV/SPIRVReader.cpp
+++ b/lib/SPIRV/SPIRVReader.cpp
@@ -684,6 +684,8 @@ bool SPIRVToLLVM::isSPIRVCmpInstTransToLLVMInst(SPIRVInstruction *BI) const {
   return isCmpOpCode(OC) && !(OC >= OpLessOrGreater && OC <= OpUnordered);
 }
 
+// TODO: Instead of direct translation to OCL we should always produce SPIR-V
+// friendly IR and apply lowering later if needed
 bool SPIRVToLLVM::isDirectlyTranslatedToOCL(Op OpCode) const {
   if (isSubgroupAvcINTELInstructionOpCode(OpCode) ||
       isIntelSubgroupOpCode(OpCode))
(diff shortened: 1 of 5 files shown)


Fznamznon:
@AlexeySotkin , @AlexeySachkov , @MrSidims , please take a look.

lib/SPIRV/SPIRVReader.cpp
```diff
@@ -696,7 +698,8 @@ bool SPIRVToLLVM::isDirectlyTranslatedToOCL(Op OpCode) const {
     // clang-consistent format in SPIRVToOCL pass.
     return !(isAtomicOpCode(OpCode) || isGroupOpCode(OpCode) ||
              isGroupNonUniformOpcode(OpCode) || isPipeOpCode(OpCode) ||
-             isMediaBlockINTELOpcode(OpCode));
+             isMediaBlockINTELOpcode(OpCode) || OpCode == OpGroupAsyncCopy ||
```
MrSidims, line 701: Does it make sense to create it to `isGroupEventAlgorithm` function?

MrSidims (approved):
Just one nit

lib/SPIRV/SPIRVReader.cpp
Fznamznon: It would make sense if we merge `SPIRVToOCL::visitCallAsyncWorkGroupCopy` and `SPIRVToOCL::visitCallGroupWaitEvents` (I added them in this patch) so we can use it in `SPIRVToOCL` pass, because I believe we will remove `isDirectlyTranslatedToOCL` function one day.

lib/SPIRV/libSPIRV/SPIRVInstruction.h
```diff
@@ -298,35 +298,26 @@ class SPIRVInstTemplateBase : public SPIRVInstruction {
...
-  // Some instructions skip literals when returning operands.
-  size_t getOperandOffset() const {
-    if (hasExecScope() && !isGroupOpCode(OpCode) && !isPipeOpCode(OpCode))
-      return 1;
-    return 0;
-  }
-
   // Get operands which are values.
   // Drop execution scope and group operation literals.
   // Return other literals as uint32 constants.
```
mlychkov, line 303: Should this comment be updated as dropping is removed from the function?

AlexeySachkov (approved):

lib/SPIRV/libSPIRV/SPIRVInstruction.h
Fznamznon: I think it can be removed, since we don't drop anything now.

PR #964, opened 2021-03-30
Fix transTypeComposite bug

KornevNikita:
This is a workaround to fix an error caused by DIBuilder which always creates
a node with the DW_TAG_struct_type tag for both class and structure.

### DIFF
--- a/lib/SPIRV/SPIRVToLLVMDbgTran.cpp
+++ b/lib/SPIRV/SPIRVToLLVMDbgTran.cpp
@@ -256,10 +256,10 @@ SPIRVToLLVMDbgTran::transTypeComposite(const SPIRVExtInst *DebugInst) {
   DICompositeType *CT = nullptr;
   switch (Ops[TagIdx]) {
   case SPIRVDebug::Class:
-    CT = Builder.createClassType(
-        ParentScope, Name, File, LineNo, Size, Align, 0, Flags, DerivedFrom,
-        DINodeArray() /*elements*/, nullptr /*VTableHolder*/,
-        nullptr /*TemplateParams*/, Identifier);
+    CT = Builder.createReplaceableCompositeType(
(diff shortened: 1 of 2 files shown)


test/DebugInfo/TransTypeCompositeCaseClass_.ll
```diff
@@ -0,0 +1,73 @@
+; Source code:
+; class A {};
+; struct B {};
+
+; int main() {
+;   A a;
+;    B b;
+;    return 0;
```
mlychkov, line 8:
```suggestion
;   B b;
;   return 0;
```

lib/SPIRV/SPIRVToLLVMDbgTran.cpp
```diff
@@ -256,10 +256,10 @@ SPIRVToLLVMDbgTran::transTypeComposite(const SPIRVExtInst *DebugInst) {
   DICompositeType *CT = nullptr;
   switch (Ops[TagIdx]) {
   case SPIRVDebug::Class:
-    CT = Builder.createClassType(
-        ParentScope, Name, File, LineNo, Size, Align, 0, Flags, DerivedFrom,
-        DINodeArray() /*elements*/, nullptr /*VTableHolder*/,
-        nullptr /*TemplateParams*/, Identifier);
+    CT = Builder.createReplaceableCompositeType(
```
AlexeySachkov, line 259: I suggest to leave a `TODO` comment here to revert this change once `DIBuilder` is fixed

AlexeySachkov (approved):
I also suggest to either file a bug in LLVM upstream about this, or propose a patch with fix

MrSidims (approved):

AlexeySachkov (approved):

AlexeySotkin (approved):

PR #966, opened 2021-03-31
Implement SPV_INTEL_memory_access_aliasing extension

MrSidims:
This extension adds analogs of alias.scope and noalias metadata
in SPIR-V. These metadata are being used to annotate non-aliasing
memory accesses and being generated, for example, during inlining
of functions with restrict pointer parameters.

This patch also casually adds OpDecorateId in the translator flow.
Spec: https://github.com/intel/llvm/pull/3426


lib/SPIRV/libSPIRV/SPIRVDecorate.h
```diff
@@ -205,6 +205,34 @@ class SPIRVDecorate : public SPIRVDecorateGeneric {
   }
 };
 
+class SPIRVDecorateId : public SPIRVDecorateGeneric {
```
MrSidims, line 208: `SPIRVDecorateId` and `SPIRVDecorate` are kind of the same. Instead of creation of another class, we can also change `SPIRVDecorate` to be templated. I didn't implement it this way tho, because addition of a template will create a bit of a mess in the code below and IMO grepping for "SPIRVDecorateId" will make implementation of other extensions faster and easier than grepping for templated "SPIRVDecorate".

lib/SPIRV/libSPIRV/SPIRVEntry.cpp
```diff
@@ -270,13 +271,30 @@ void SPIRVEntry::addDecorate(Decoration Kind, SPIRVWord Literal) {
   addDecorate(new SPIRVDecorate(Kind, this, Literal));
 }
 
+void SPIRVEntry::addDecorateId(SPIRVDecorateId *Dec) {
```
MrSidims, line 274: It's a duplication of "addDecorate" and their bodies can be merged actually. But the way I've done it now IMO is more transparent. Same is applicable to all Decorate functions below. I don't mind to change it tho.

lib/SPIRV/libSPIRV/SPIRVDecorate.h
```diff
@@ -154,7 +154,7 @@ class SPIRVDecorate : public SPIRVDecorateGeneric {
   SPIRVDecorate() : SPIRVDecorateGeneric(OC) {}
 
   llvm::Optional<ExtensionID> getRequiredExtension() const override {
-    switch (Dec) {
+    switch (static_cast<int>(Dec)) {
```
MrSidims, line 157: To remove.

lib/SPIRV/SPIRVReader.cpp
```diff
@@ -1507,6 +1509,59 @@ static void replaceOperandWithAnnotationIntrinsicCallResult(Value *&V) {
...
+        BM->get<SPIRVAliasScopeDeclINTEL>(ScopeId);
+    std::vector<SPIRVId> AliasDomainIds = AliasScope->getArguments();
+    // Currently we expect exactly one argument for aliasing scope
+    // instruction.
+    // TODO: add translation of string scope and domain operand.
+    assert(AliasDomainIds.size() == 1 &&
+           "AliasScopeDeclINTEL must have exactly one argument");
+    SPIRVId AliasDomainId = AliasDomainIds[0];
+    // Create and store unique domain and scope metadata
+    MDAliasDomainMap.emplace(std::pair<SPIRVId, MDNode *>(
```
MrSidims, line 1550: Usage of `std::unordered_map` promises an OK performance, but it's not the fasted algorithm here. A bunch of if-elses would be a bit faster, but here I'd prefer code-readability, than some extra optimizations. WDYT?

MrSidims:
Please take a look
@AlexeySotkin @AlexeySachkov @Fznamznon

lib/SPIRV/SPIRVWriter.cpp
```diff
@@ -1216,6 +1217,32 @@ SPIRVValue *LLVMToSPIRV::transAtomicLoad(LoadInst *LD, SPIRVBasicBlock *BB) {
                                           BB, transType(LD->getType())));
 }
 
+// Aliasing list MD contains several scope MD nodes whithin it. Each scope MD
+// has a selfreference and an extra MD node for aliasing domain and also it
+// can contatin and optional string operand. Domain MD contains a self-reference
```
mlychkov, line 1222: Looks like there is some typo in this comment line.

lib/SPIRV/SPIRVWriter.cpp
```diff
@@ -1368,7 +1395,12 @@ SPIRVValue *LLVMToSPIRV::transValueWithoutDecoration(Value *V,
     if (ST->isAtomic())
       return transAtomicStore(ST, BB);
 
-    std::vector<SPIRVWord> MemoryAccess(1, 0);
+    // Keep this vector to store MemoryAccess operands for both Alignment and
```
mlychkov, line 1398: MemoryAccess fullfilling is the same for load and store intructions. May it be moved into a separate function?

lib/SPIRV/libSPIRV/SPIRVMemAliasingINTEL.h
```diff
@@ -0,0 +1,76 @@
...
+protected:
+  static const SPIRVWord FixedWC = TheFixedWordCount;
+  static const Op OC = TheOpCode;
+  std::vector<SPIRVId> Args;
+  _SPIRV_DEF_ENCDEC2(Id, Args)
+};
+
+#define _SPIRV_OP(x, ...)                                                      \
+  typedef SPIRVMemAliasingINTELGeneric<internal::Op##x, __VA_ARGS__> SPIRV##x;
+// Intel Mem Alasing Instructions
```
mlychkov, line 69:
```suggestion
// Intel Memory Aliasing Instructions
```

MrSidims:
Applied suggestions

lib/SPIRV/SPIRVWriter.cpp
```diff
@@ -1368,7 +1409,12 @@ SPIRVValue *LLVMToSPIRV::transValueWithoutDecoration(Value *V,
...
+    // Aliasing information. It's quite incorrect, since the first is of a
+    // SPIRVWord type, while the second one is of SPIRVId type, but creation
+    // of the second vector would make fulfilling '3.26 Memory Operands'
+    // requirements about operands order a bit tricky.
```
AlexeySotkin, line 1416: SPIRV id is the same SPIRV word. See https://www.khronos.org/registry/spir-v/specs/unified1/SPIRV.html#Id and https://github.com/KhronosGroup/SPIRV-LLVM-Translator/blob/56988bb67213554db07932d0f0765a5f0de2e581/lib/SPIRV/libSPIRV/SPIRVEnum.h#L54-L55

lib/SPIRV/libSPIRV/SPIRVEntry.cpp
AlexeySotkin: I'd prefer to have only `addDecorate()`

AlexeySotkin: .. or call one from another.

lib/SPIRV/libSPIRV/SPIRVMemAliasingINTEL.h
```diff
@@ -0,0 +1,76 @@
+//===- SPIRVMemAliasingINTEL.h -                               --*- C++ -*-===//
+//
+//                     The LLVM/SPIRV Translator
+//
+// This file is distributed under the University of Illinois Open Source
+// License. See LICENSE.TXT for details.
```
AlexeySotkin, line 6: @MrSidims please double check that this is the right license.

lib/SPIRV/SPIRVWriter.cpp
MrSidims: I meant to say, that SPIRVId always points to some object in SPIR-V module (instruction, variable etc), whilst SPIRVWord can also be a constant literal. But I actually missed "An <id> always consumes one word". Removed this comment.

lib/SPIRV/libSPIRV/SPIRVMemAliasingINTEL.h
MrSidims: TBH, I've taken it from SPIRVAsm.h :
```
  1 //===- SPIRVAsm.h -                                           --*- C++ -*-===//
  2 //
  3 //                     The LLVM/SPIRV Translator
  4 //
  5 // This file is distributed under the University of Illinois Open Source
  6 // License. See LICENSE.TXT for details.
  7 //
  8 //===----------------------------------------------------------------------===//
```

I'll double check it.

lib/SPIRV/libSPIRV/SPIRVMemAliasingINTEL.h
```diff
@@ -1,10 +1,35 @@
 //===- SPIRVMemAliasingINTEL.h -                               --*- C++ -*-===//
 //
-//                     The LLVM/SPIRV Translator
+//                     The LLVM/SPIR-V Translator
 //
 // This file is distributed under the University of Illinois Open Source
 // License. See LICENSE.TXT for details.
 //
+// Copyright (c) 2021 Advanced Micro Devices, Inc. All rights reserved.
```
AlexeySotkin, line 8: Why AMD @MrSidims ?

lib/SPIRV/libSPIRV/SPIRVMemAliasingINTEL.h
```diff
@@ -1,10 +1,35 @@
...
+// the rights to use, copy, modify, merge, publish, distribute, sublicense,
+// and/or sell copies of the Software, and to permit persons to whom the
+// Software is furnished to do so, subject to the following conditions:
+//
+// Redistributions of source code must retain the above copyright notice,
+// this list of conditions and the following disclaimers.
+// Redistributions in binary form must reproduce the above copyright notice,
+// this list of conditions and the following disclaimers in the documentation
+// and/or other materials provided with the distribution.
+// Neither the names of Advanced Micro Devices, Inc., nor the names of its
```
AlexeySotkin, line 22: Here name of the company seems to be incorrect too.

lib/SPIRV/libSPIRV/SPIRVMemAliasingINTEL.h
MrSidims: Oops, my cover has been compromised :D

lib/SPIRV/SPIRVReader.cpp
AlexeySotkin: I don't think this is the case when we should care about performance.

> The real problem is that programmers have spent far too much time worrying about efficiency in the wrong places and at the wrong times; premature optimization is the root of all evil (or at least most of it) in programming.
> -- Donald Knuth.

Speaking about readability, I guess, we can write it as:
```suggestion
    MDAliasDomainMap[AliasDomainId] = MDB.createAnonymousAliasScopeDomain();
```

MrSidims: The suggestion won't work, since `emplace` checks if there is an element in map and add it only if it's not, and [] operator will insert/replace this element no matter what. So in the end it would be different aliasing scopes.

AlexeySotkin: Then maybe we can do 
```suggestion
    MDAliasDomainMap.emplace(AliasDomainId, MDB.createAnonymousAliasScopeDomain());
```
?

lib/SPIRV/SPIRVWriter.cpp
```diff
@@ -1391,7 +1441,12 @@ SPIRVValue *LLVMToSPIRV::transValueWithoutDecoration(Value *V,
     if (LD->isAtomic())
       return transAtomicLoad(LD, BB);
 
-    std::vector<SPIRVWord> MemoryAccess(1, 0);
+    // Keep this vector to store MemoryAccess operands for both Alignment and
```
MrSidims, line 1444: Misleading comment

lib/SPIRV/libSPIRV/SPIRVEntry.cpp
```diff
@@ -262,21 +263,41 @@ void SPIRVEntry::addDecorate(SPIRVDecorate *Dec) {
...
+    addDecorate(new SPIRVDecorateId(Kind, this, Literal));
+    return;
+  default:
+    addDecorate(new SPIRVDecorate(Kind, this, Literal));
+  }
 }
 
 void SPIRVEntry::eraseDecorate(Decoration Dec) { Decorates.erase(Dec); }
 
+void SPIRVEntry::eraseDecorateId(Decoration Dec) { DecorateIds.erase(Dec); }
```
MrSidims, line 289: To remove

lib/SPIRV/libSPIRV/SPIRVMemAliasingINTEL.h
```diff
@@ -0,0 +1,101 @@
...
+                               const std::vector<SPIRVId> &TheArgs)
+      : SPIRVEntry(TheModule, TheArgs.size() + TheFixedWordCount, TheOpCode,
+                   TheId), Args(TheArgs) {
+    SPIRVMemAliasingINTELGeneric::validate();
+    assert(TheModule && "Invalid module");
+  }
+
+  SPIRVMemAliasingINTELGeneric() : SPIRVEntry(TheOpCode) {}
+
+  void setArguments(const std::vector<SPIRVId> &A) {
```
MrSidims, line 61: Check and remove if not needed

test/transcoding/SPV_INTEL_memory_access_aliasing/intel-alias-masked-load-store.ll
```diff
@@ -0,0 +1,148 @@
...
+
+; Function Attrs: nofree norecurse nounwind willreturn mustprogress
+define dso_local spir_kernel void @_ZTSZ4mainE15kernel_restrict(i32 addrspace(1)* noalias %_arg_, i32 addrspace(1)* noalias %_arg_1, i32 addrspace(1)* noalias %_arg_3) local_unnamed_addr #0 !kernel_arg_buffer_location !4 {
+entry:
+  %0 = addrspacecast i32 addrspace(1)* %_arg_ to i32 addrspace(4)*
+  %1 = addrspacecast i32 addrspace(1)* %_arg_1 to i32 addrspace(4)*
+  %2 = addrspacecast i32 addrspace(1)* %_arg_3 to i32 addrspace(4)*
+; CHECK-LLVM: call spir_func i32 @maskedload{{.*}} !alias.scope ![[LISTMD1:[0-9]+]]
+; CHECK-LLVM: call spir_func i32 @maskedload{{.*}} !alias.scope ![[LISTMD1]]
+  %3 = call i32 @maskedload(i32 addrspace(4)* %0), !tbaa !5, !alias.scope !9
```
MrSidims, line 81: masked->wrapped

AlexeySotkin (approved):

PR #970, opened 2021-04-01
Backmerge fix for emission of SPIR-V friendly IR for vload to llvm_release_110

aratajew:


svenvh:
It would help if we could postpone merging this until after #967, thanks!

svenvh:
This is included in #986.

PR #983, opened 2021-04-05
Fix incorrect translation of FPGA decoration on arrays

MrSidims:
var.annotation shall be emitted, but instead the translator was
generating ptr.annotation. It was happening due to incorrect assumption,
that for attributed static memory variable decoration shall be applied
to alloca instruction, whilst in the reality the alloca instruction
might be hidden by several cast instructions.

### DIFF
--- a/lib/SPIRV/SPIRVReader.cpp
+++ b/lib/SPIRV/SPIRVReader.cpp
@@ -3692,8 +3692,25 @@ void SPIRVToLLVM::transIntelFPGADecorations(SPIRVValue *BV, Value *V) {
       Value *BaseInst =
           AL ? Builder.CreateBitCast(V, Int8PtrTyPrivate, V->getName()) : Inst;
 
+      // Try to find alloca instruction for statically allocated variables.
+      // Alloca might be hidden by a couple of casts.
+      bool isStaticMemoryAttribute = AL ? true : false;
+      if (!isStaticMemoryAttribute) {
+        do {
+          if (!isa<BitCastInst>(Inst) && !isa<AddrSpaceCastInst>(Inst))
+            break;
+          Inst = dyn_cast<Instruction>(Inst->getOperand(0));
+          if (!Inst)
+            break;
+          if (isa<AllocaInst>(Inst)) {
+            isStaticMemoryAttribute = true;
+            break;
+          }
+        } while (Inst);
+      }
+
       auto AnnotationFn =
-          AL // Static memory attributes if true
+          isStaticMemoryAttribute
               ? llvm::Intrinsic::getDeclaration(M, Intrinsic::var_annotation)
               : llvm::Intrinsic::getDeclaration(M, Intrinsic::ptr_annotation,
                                                 AllocatedTy);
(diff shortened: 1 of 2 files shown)


lib/SPIRV/SPIRVReader.cpp
```diff
@@ -3692,8 +3692,25 @@ void SPIRVToLLVM::transIntelFPGADecorations(SPIRVValue *BV, Value *V) {
...
+          if (!isa<BitCastInst>(Inst) && !isa<AddrSpaceCastInst>(Inst))
+            break;
+          Inst = dyn_cast<Instruction>(Inst->getOperand(0));
+          if (!Inst)
+            break;
+          if (isa<AllocaInst>(Inst)) {
+            isStaticMemoryAttribute = true;
+            break;
+          }
+        } while (Inst);
```
AlexeySachkov, line 3709: I assume you were looking for [`Value::stripPointerCasts`](https://llvm.org/doxygen/classllvm_1_1Value.html#a966eb231e7d4e572874d2cb49b18faea)?

MrSidims: Looks like it won't work. It "Strip off pointer casts, **all-zero GEPs** and address space casts", which breaks the current behavior. If we annotates structure members (to which we have access through GEP), the annotation shall be `ptr.annotation` and if we use a result of `stripPointerCasts` here, `ptr.annotation` would change to `var.annotation`.

MrSidims: Or actually, we can ask alloca what type is has.

MrSidims: Nevermind. The problem here is that for annotated struct objects we shall use var.annotation and for struct members we shall use ptr.annotation. And the cleanest way to distinguish them in LLVM IR is by looking on existence of GEP instruction (asking for a type of alloca won't help, since in both cases it would be a pointer on a structure). So in the end the existing approach can't be replaced by `stripPointerCasts`.

MrSidims:
@mlychkov @AGindinson @vmaksimo please take a look

lib/SPIRV/SPIRVReader.cpp
mlychkov, line 3709:
```suggestion
      while (!isStaticMemoryAttribute && Inst &&
             (isa<BitCastInst>(Inst) || isa<AddrSpaceCastInst>(Inst))) {
          Inst = dyn_cast<Instruction>(Inst->getOperand(0));
          isStaticMemoryAttribute = (Inst && isa<AllocaInst>(Inst));
      }

```

test/transcoding/IntelFPGAMemoryAttributes.ll
```diff
@@ -168,6 +168,9 @@
 ;   } s;
 ;   s.field = 0;
 ; }
+; void static_memory_attribute() {
+;   [[intelfpga::register]] int register_var[32];
+; }
```
vmaksimo, line 173: A bit confused with the naming - we have an `IntelFPGAMemoryAttributesForStaticVar.ll` test for static memory attributes.

lib/SPIRV/SPIRVReader.cpp
MrSidims: Thanks!

MrSidims:
Ping
Also tagging @AlexeySotkin

AlexeySachkov (approved):

MrSidims:
Ping @AlexeySotkin

PR #984, opened 2021-04-06
Fix SPIRV friendly IR for OpImageSampleExplicitLod&OpSampledImage

Fznamznon:
These instructions were translated directly to OCL even if SPIR-V
friendly LLVM IR was requested.

### DIFF
--- a/lib/SPIRV/SPIRVReader.cpp
+++ b/lib/SPIRV/SPIRVReader.cpp
@@ -690,6 +690,8 @@ bool SPIRVToLLVM::isDirectlyTranslatedToOCL(Op OpCode) const {
   if (isSubgroupAvcINTELInstructionOpCode(OpCode) ||
       isIntelSubgroupOpCode(OpCode))
     return true;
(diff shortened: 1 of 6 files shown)


Fznamznon:
@AlexeySotkin , @AlexeySachkov , @MrSidims please take a look.

lib/SPIRV/SPIRVReader.cpp
```diff
@@ -1280,60 +1282,6 @@ static char getTypeSuffix(Type *T) {
   return Suffix;
 }
 
-// ToDo: Handle unsigned integer return type. May need spec change.
```
AlexeySachkov, line 1283: This `TODO` was lost, I suggest that we preserve it if it is still actual

Fznamznon: Ok, fair enough.

Fznamznon:
Ping.

MrSidims (approved):

AlexeySachkov (approved):

Issue #977, opened 2021-04-01
Remove translation between "const" arg-type metadata and NoWrite function parameter


StuartDBrady:
The `NoWrite` SPIR-V function parameter provides an assurance that memory pointed to by a `NoWrite` pointer will not be modified _through that pointer_, as per LLVM's `readonly` argument attribute.

The `const` qualifier (represented in LLVM IR for kernel functions using `kernel_arg_type_qual` metadata) does _not_ provide any such guarantee. There is nothing that prevents a function that takes a pointer to a `const` pointee from stripping the `const` pointer away and modifying the pointee.

As such, it is not appropriate to translate from one to the other, and so the code for doing so should be removed, so that this translation is not performed either when translating from LLVM IR to SPIR-V, or when translating from SPIR-V to LLVM IR.

MrSidims:
You may assign it on me

MrSidims:
Fixed with https://github.com/KhronosGroup/SPIRV-LLVM-Translator/pull/989

PR #989, opened 2021-04-06
Remove mapping of NoWrite function parameter attribute on qual metadata

MrSidims:
The NoWrite SPIR-V function parameter provides an assurance that memory pointed to by a NoWrite pointer
will not be modified through that pointer.

The const qualifier (represented in LLVM IR for kernel functions using kernel_arg_type_qual metadata) does
not provide any such guarantee. There is nothing that prevents a function that takes a pointer to a const
pointee from stripping the const pointer away and modifying the pointee.

As such, it is not appropriate to translate from one to the other, and so the code for doing so should be
removed, so that this translation is not performed either when translating from LLVM IR to SPIR-V, or
when translating from SPIR-V to LLVM IR.

### DIFF
--- a/lib/SPIRV/SPIRVReader.cpp
+++ b/lib/SPIRV/SPIRVReader.cpp
@@ -104,7 +104,6 @@ static bool DbgSaveTmpLLVM = false;
 static const char *DbgTmpLLVMFileName = "_tmp_llvmbil.ll";
 
 namespace kOCLTypeQualifierName {
-const static char *Const = "const";
 const static char *Volatile = "volatile";
 const static char *Restrict = "restrict";
 const static char *Pipe = "pipe";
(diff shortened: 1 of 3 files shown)


MrSidims:
It addresses https://github.com/KhronosGroup/SPIRV-LLVM-Translator/issues/977 . Correct translation of the metadata will be done later.

MrSidims:
@StuartDBrady @svenvh @AlexeySachkov @AlexeySotkin please take a look

svenvh (commented):
LGTM. It would be nice if you could add the motivation/reasoning from #977 into the commit message.

MrSidims:
> LGTM. It would be nice if you could add the motivation/reasoning from #977 into the commit message.

Added to the PR description.

MrSidims:
@AlexeySotkin @svenvh I think we can merge it now

svenvh:
> @AlexeySotkin @svenvh I think we can merge it now

As I mentioned earlier, it would be helpful if you can add the motivation to the commit message (not the PR description alone), so that we also have it in the git history (and not only on GitHub).

MrSidims:
> > @AlexeySotkin @svenvh I think we can merge it now
> 
> As I mentioned earlier, it would be helpful if you can add the motivation to the commit message (not the PR description alone), so that we also have it in the git history (and not only on GitHub).

Okay, done. I thought there is a way (policy) in github to reuse PR description for comment msg during merging.

svenvh (commented):
> Okay, done. I thought there is a way (policy) in github to reuse PR description for comment msg during merging.

Thanks!  It depends on the merge strategy chosen when submitting the change.  Since this is a single commit, we may not use the "squash+merge" strategy and then there is no chance to add anything into the commit message.

Change LGTM, but I'd like to see someone else approve it too.

svenvh (approved):

AlexeySachkov:
> Thanks! It depends on the merge strategy chosen when submitting the change. Since this is a single commit, we may not use the "squash+merge" strategy and then there is no chance to add anything into the commit message.

BTW, I have squash and merge button available and therefore can change the commit message even for single commit

AlexeySachkov (approved):
Seems reasonable to me. Tagging @StuartDBrady as he was involved in related discussion

PR #991, opened 2021-04-07
Promote SPV_INTEL_optimization_hints to SPV_KHR_expect_assume

MrSidims:


MrSidims:
@mlychkov @AlexeySachkov @AlexeySotkin @Fznamznon please take a look

AlexeySotkin:
@MrSidims, I think, we still need to support translation of SPV_INTEL_optimization_hints from SPIR-V to LLVM, right?

MrSidims:
> @MrSidims, I think, we still need to support translation of SPV_INTEL_optimization_hints from SPIR-V to LLVM, right?

I don't think so, since SPV_INTEL_optimization_hints and SPV_KHR_expect_assume have the same token allocations. Unless we want to support a case when "llvm-spirv -to-binary legacy.spt" is used.

Fznamznon (approved):
LGTM

AlexeySotkin:
> > @MrSidims, I think, we still need to support translation of SPV_INTEL_optimization_hints from SPIR-V to LLVM, right?
> 
> I don't think so, since SPV_INTEL_optimization_hints and SPV_KHR_expect_assume have the same token allocations. Unless we want to support a case when "llvm-spirv -to-binary legacy.spt" is used.

You are right, we are only changing textual representation.

AlexeySotkin (approved):

MrSidims:
> > > @MrSidims, I think, we still need to support translation of SPV_INTEL_optimization_hints from SPIR-V to LLVM, right?
> > 
> > 
> > I don't think so, since SPV_INTEL_optimization_hints and SPV_KHR_expect_assume have the same token allocations. Unless we want to support a case when "llvm-spirv -to-binary legacy.spt" is used.
> 
> You are right, we are only changing textual representation.

The only thing that just started to worrying me is that OpExtension string in the module. We might want to leave declaration of SPV_INTEL_optimization_hints , I'm double checking it right now.

AlexeySachkov:
> > > > @MrSidims, I think, we still need to support translation of SPV_INTEL_optimization_hints from SPIR-V to LLVM, right?
> > > 
> > > 
> > > I don't think so, since SPV_INTEL_optimization_hints and SPV_KHR_expect_assume have the same token allocations. Unless we want to support a case when "llvm-spirv -to-binary legacy.spt" is used.
> > 
> > 
> > You are right, we are only changing textual representation.
> 
> The only thing that just started worrying me is that OpExtension string in the module. We might want to leave declaration of SPV_INTEL_optimization_hints , I'm double checking it right now.

There shouldn't be any problem with strings. If I understand correctly consumers should validate incoming SPIR-V by looking into capabilities and not into extension strings

include/LLVMSPIRVExtensions.inc
```diff
@@ -7,6 +7,7 @@ EXT(SPV_EXT_shader_atomic_float_min_max)
 EXT(SPV_KHR_no_integer_wrap_decoration)
 EXT(SPV_KHR_float_controls)
 EXT(SPV_KHR_linkonce_odr)
+EXT(SPV_KHR_expect_assume)
 EXT(SPV_INTEL_subgroups)
 EXT(SPV_INTEL_media_block_io)
 EXT(SPV_INTEL_device_side_avc_motion_estimation)
```
AlexeySachkov, line 13: I think we should either remove `SPV_INTEL_optimization_hints` from this list, or update SPIRVWriter to allow `KHR` opcodes to be generated for `INTEL` extension - which seems a bit strange, but should be fine since those are binary compatible extensions

MrSidims: I have tried it locally and a reversed translation of a module with SPV_INTEL_optimization_hints failed, erroring out, that there is an unsupported extension, see: lib/SPIRV/libSPIRV/SPIRVStream.cpp `getEntry` function
```
241   if (OpExtension == OpCode) {
242     auto *OpExt = static_cast<SPIRVExtension *>(Entry);
243     ExtensionID ExtID;
244     bool ExtIsKnown = SPIRVMap<ExtensionID, std::string>::rfind(
245         OpExt->getExtensionName(), &ExtID);
246     if (!M.getErrorLog().checkError(
247             ExtIsKnown, SPIRVEC_InvalidModule,
248             "input SPIR-V module uses unknown extension '" +
249                 OpExt->getExtensionName() + "'")) {
250       M.setInvalid();
251     }
```

MrSidims: So I left this extension in the list so pre-compiled modules could be re-used by SPIR-V consumer.

AlexeySachkov: > So I left this extension in the list so pre-compiled modules could be re-used by SPIR-V consumer.

Got it, thanks

AlexeySachkov (approved):

AlexeySachkov:
@bashbaug, @AnastasiaStulova: FYI

svenvh (approved):

lib/SPIRV/libSPIRV/SPIRVModule.cpp
```diff
@@ -1603,17 +1603,17 @@ SPIRVInstruction *SPIRVModuleImpl::addSampledImageInst(SPIRVType *ResultTy,
...
-SPIRVInstruction *SPIRVModuleImpl::addAssumeTrueINTELInst(SPIRVValue *Condition,
+SPIRVInstruction *SPIRVModuleImpl::addAssumeTrueKHRInst(SPIRVValue *Condition,
                                                           SPIRVBasicBlock *BB) {
-  return addInstruction(new SPIRVAssumeTrueINTEL(Condition->getId(), BB), BB);
+  return addInstruction(new SPIRVAssumeTrueKHR(Condition->getId(), BB), BB);
 }
 
-SPIRVInstruction *SPIRVModuleImpl::addExpectINTELInst(SPIRVType *ResultTy,
+SPIRVInstruction *SPIRVModuleImpl::addExpectKHRInst(SPIRVType *ResultTy,
                                                       SPIRVValue *Value,
```
svenvh, line 1612: Actually clang-format failed here, and in some other places too (but ignore the formatting failures in spirv.hpp).

MrSidims: Thanks! Didn't notice that.

svenvh: Thanks for fixing!  I have created #1002 to avoid this problem in the future.

PR #993, opened 2021-04-12
[CI] Run check-code-style on explicit OS version 

svenvh:


AlexeySachkov:
@svenvh, I think I might know what is wrong here: in our workflow we use `runs-on: ubuntu-latest` which is (I assume) Ubuntu 20, while we still use `bionic` apt repos for Ubuntu 18 for installing packages - probably there is a conflict somewhere?

svenvh:
> @svenvh, I think I might know what is wrong here: in our workflow we use `runs-on: ubuntu-latest` which is (I assume) Ubuntu 20, while we still use `bionic` apt repos for Ubuntu 18 for installing packages - probably there is a conflict somewhere?

Ah, good catch, I didn't notice it was running on 20.04!  That seems to be the issue indeed: the checks are running fine when I change it to `runs-on: ubuntu-18.04`.  I think it makes sense to ask for an explicit Ubuntu version so that we are not affected whenever GitHub changes the "latest".

AlexeySotkin:
> > @svenvh, I think I might know what is wrong here: in our workflow we use `runs-on: ubuntu-latest` which is (I assume) Ubuntu 20, while we still use `bionic` apt repos for Ubuntu 18 for installing packages - probably there is a conflict somewhere?
> 
> Ah, good catch, I didn't notice it was running on 20.04! That seems to be the issue indeed: the checks are running fine when I change it to `runs-on: ubuntu-18.04`. I think it makes sense to ask for an explicit Ubuntu version so that we are not affected whenever GitHub changes the "latest".

Does it make sense to switch to Ubuntu 20.04?

svenvh:
> Does it make sense to switch to Ubuntu 20.04?

We could do that in a followup PR I suppose, as it may require some more debugging/tweaking of the config.  For now, I just want to make sure the formatting checks are working again.

AlexeySotkin:
> > Does it make sense to switch to Ubuntu 20.04?
> 
> We could do that in a followup PR I suppose, as it may require some more debugging/tweaking of the config. For now, I just want to make sure the formatting checks are working again.

Ok.

AlexeySotkin (approved):

AlexeySachkov:
> > Does it make sense to switch to Ubuntu 20.04?
> 
> We could do that in a followup PR I suppose, as it may require some more debugging/tweaking of the config. For now, I just want to make sure the formatting checks are working again.

Thanks for doing that, BTW!

AlexeySachkov (approved):

PR #995, opened 2021-04-13
Update for LLVM iterator change

svenvh:
Fix the build after LLVM commit `0a92aff721f4 ("Replace uses of
std::iterator with explicit using", 2021-04-12)`.


AlexeySachkov (approved):
Thanks!

PR #996, opened 2021-04-15
Add W/A to translate OpenCL kernel_arg_type_qual metadata

MrSidims:
Since constant qualifier stored in the metadata is no longer mapped
on NoWrite function parameter attribute we need a another way to
save information about it. As work around this patch stores argument
qualifier information in OpString under
'preserve-ocl-kernel-arg-type-metadata-through-string' option. This
option should be also passed during reversed translation to obtain
the metadata.

This patch also moves already existing workaround for kernel_arg_type
metadata under the option.

### DIFF
--- a/test/transcoding/kernel_arg_type_qual.ll
+++ b/test/transcoding/kernel_arg_type_qual.ll
@@ -1,0 +1,43 @@
+; RUN: llvm-as %s -o %t.bc
+; RUN: llvm-spirv -preserve-ocl-kernel-arg-type-metadata-through-string %t.bc -o - -spirv-text | FileCheck %s --check-prefix=CHECK-SPIRV
+; RUN: llvm-spirv %t.bc -o - -spirv-text | FileCheck %s --check-prefix=CHECK-SPIRV-NEGATIVE
+; RUN: llvm-spirv -preserve-ocl-kernel-arg-type-metadata-through-string %t.bc -o %t.spv
+; RUN: llvm-spirv -r -preserve-ocl-kernel-arg-type-metadata-through-string %t.spv -o %t.rev.bc
+; RUN: llvm-dis < %t.rev.bc | FileCheck %s --check-prefix=CHECK-LLVM-WORKAROUND
+; RUN: llvm-spirv %t.bc -o %t.spv
+; RUN: llvm-spirv -r %t.spv -o %t.rev.bc
+; RUN: llvm-dis < %t.rev.bc | FileCheck %s --check-prefix=CHECK-LLVM-WORKAROUND-NEGATIVE
+
+; ModuleID = 'test.cl'
+source_filename = "test.cl"
+target datalayout = "e-m:e-p270:32:32-p271:32:32-p272:64:64-i64:64-f80:128-n8:16:32:64-S128"
+target triple = "spir64-unknown-unknown."
+
+; CHECK-SPIRV: String 12 "kernel_arg_type_qual.test.volatile,const,,"
+; CHECK-SPIRV: Name [[ARG:[0-9]+]] "g"
+; CHECK-SPIRV: Decorate [[ARG]] Volatile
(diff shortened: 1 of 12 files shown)


MrSidims:
Need to merge https://github.com/KhronosGroup/SPIRV-LLVM-Translator/pull/989 first

test/group-decorate.spt
```diff
@@ -38,6 +35,4 @@
 ; RUN: llvm-spirv -r %t.spv -o %t.bc
 ; RUN: llvm-dis < %t.bc | FileCheck %s --check-prefix=CHECK-LLVM
 
-; CHECK-LLVM: define spir_kernel void @test(<4 x i8> addrspace(1)* nocapture %src1, <4 x i8> addrspace(1)* nocapture %src2, <4 x i8> addrspace(1)* nocapture %dst) {{.*}} !kernel_arg_type_qual ![[QUALS:[0-9]+]]
-
-; CHECK-LLVM: ![[QUALS]] = !{!"const", !"const", !""}
+; CHECK-LLVM: define spir_kernel void @test(<4 x i8> addrspace(1)* nocapture %src1, <4 x i8> addrspace(1)* nocapture %src2, <4 x i8> addrspace(1)* nocapture %dst) {{.*}}
```
mlychkov, line 38: May we add string "const" to keep constant qualifier in this test?

lib/SPIRV/SPIRVWriter.h
```diff
@@ -192,6 +192,7 @@ class LLVMToSPIRVBase {
   bool isKernel(Function *F);
   bool transMetadata();
   bool transOCLMetadata();
+
```
mlychkov, line 195: Could you please remove this chabge?

MrSidims: Thanks! Missed this one

test/group-decorate.spt
MrSidims: This change relates to https://github.com/KhronosGroup/SPIRV-LLVM-Translator/pull/996/commits/b1d6306a9c4904cf3921c9dbebf2bfc673deb622 or https://github.com/KhronosGroup/SPIRV-LLVM-Translator/pull/989 . The test checks correctness of decoration groups and function parameter attribute NoWrite previously is among decorations to be grouped. As we unmap it from 'constant' qualifier we shan't check it in this tests.

mlychkov (approved):

StuartDBrady (changes requested):
Could this workaround be disabled by default, please?

test/transcoding/kernel_arg_type_qual.ll
```diff
@@ -0,0 +1,37 @@
...
+; RUN: llvm-spirv %t.bc -o %t.spv
+; RUN: llvm-spirv -r %t.spv -o %t.rev.bc
+; RUN: llvm-dis < %t.rev.bc | FileCheck %s --check-prefix=CHECK-LLVM
+
+; ModuleID = 'test.cl'
+source_filename = "test.cl"
+target datalayout = "e-m:e-p270:32:32-p271:32:32-p272:64:64-i64:64-f80:128-n8:16:32:64-S128"
+target triple = "spir64-unknown-unknown."
+
+; CHECK-SPIRV: String 12 "kernel_arg_type_qual.test.volatile,const,,"
```
AlexeySotkin, line 12: What if an argument has multiple qualifiers at the same time?

MrSidims: It's possible to have restrict volatile pointer to mark that it doesn't alias with non-volatile pointer.

MrSidims:
@StuartDBrady @svenvh W/A is disabled by default and now it requires preserve-ocl-kernel-arg-type-metadata-through-string to be passed

FYI @pszymich

Fznamznon (approved):
It would be great to mention info about `kernel_arg_type` metadata in commit message here as well.

MrSidims:
Merging, since the patch was approved in llvm_11 branch

PR #1006, opened 2021-04-20
Switch to Github Actions on llvm_release_100

AlexeySachkov:
Config files were copied from `master` and adjusted to the release branch


Hopetech:
Seems OK to me.
Can we also do the same change to llvm_release_110?

AlexeySotkin (approved):

PR #1015, opened 2021-04-22
Relax handling of boolean specialization constants

AlexeySachkov:
When SPIR-V friendly IR is used, it might be the case that it was
generated by some other LLVM tool and/or pass and for boolean types it
is possible that they are represented as i8 instead of i1 in LLVM IR
(depending on how they were used in the original source file).

This patch adjusts and relaxes handling of boolean specialization
constants from SPIR-V friendly IR to get the following signatures
acceptable: `i1 _Z20__spirv_SpecConstantia(i32, i8)` by using return
type of SPIR-V friendly IR function as specialization constant type
instead of using the type of the second argument.


MrSidims (commented):
I guess test name is missed, other than tat the patch is LGTM

AlexeySachkov:
> I guess test name is missed, other than tat the patch is LGTM

Good catch, thanks, fixed in 29c185cdc18ae2c77ef18612dc712020655444a9

MrSidims (approved):

AlexeySachkov:
@StuartDBrady, any feedback?

AlexeySotkin (approved):

AlexeySachkov:
@StuartDBrady, I've merged this without your approval, because it has been on review for a week already, but if you have any comments - I would be happy to apply them in a follow-up PR, just let me know

PR #1016, opened 2021-04-25
Fix -Werror=non-virtual-dtor -Werror=suggest-override errors

hewj03:
* "has virtual functions and accessible non-virtual destructor"
* runOnModule "can be marked override"

### DIFF
--- a/lib/SPIRV/SPIRVToOCL.h
+++ b/lib/SPIRV/SPIRVToOCL.h
@@ -182,7 +182,7 @@ protected:
   SPIRVToOCLLegacy(char &ID) : ModulePass(ID) {}
 
 public:
-  virtual bool runOnModule(Module &M) override = 0;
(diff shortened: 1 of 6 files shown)


MrSidims (approved):

lib/SPIRV/SPIRVToOCL.h
```diff
@@ -181,7 +182,7 @@ class SPIRVToOCLLegacy : public ModulePass {
   SPIRVToOCLLegacy(char &ID) : ModulePass(ID) {}
 
 public:
-  virtual bool runOnModule(Module &M) = 0;
+  virtual bool runOnModule(Module &M) override = 0;
```
AlexeySachkov, line 185:
```suggestion
  bool runOnModule(Module &M) override = 0;
```

PR #1021, opened 2021-04-28
Fix for ParseChecksum function

KornevNikita:
This is a patch to fix failure, caused by bringing as argument of ParseChecksum
StringRef with non-zero length, but with empty data.

### DIFF
--- a/lib/SPIRV/SPIRVToLLVMDbgTran.cpp
+++ b/lib/SPIRV/SPIRVToLLVMDbgTran.cpp
@@ -1007,7 +1007,8 @@ DIFile *SPIRVToLLVMDbgTran::getFile(const SPIRVId SourceId) {
          "DebugSource instruction is expected");
   SPIRVWordVec SourceArgs = Source->getArguments();
   assert(SourceArgs.size() == OperandCount && "Invalid number of operands");
-  auto ChecksumStr = !getDbgInst<SPIRVDebug::DebugInfoNone>(SourceArgs[TextIdx])
+  string ChecksumStr =
+      !getDbgInst<SPIRVDebug::DebugInfoNone>(SourceArgs[TextIdx])
                          ? getString(SourceArgs[TextIdx])
                          : "";
   StringRef Checksum(ChecksumStr);
--- a/lib/SPIRV/SPIRVToLLVMDbgTran.cpp
+++ b/lib/SPIRV/SPIRVToLLVMDbgTran.cpp
@@ -1007,7 +1007,11 @@ DIFile *SPIRVToLLVMDbgTran::getFile(const SPIRVId SourceId) {
          "DebugSource instruction is expected");
   SPIRVWordVec SourceArgs = Source->getArguments();
   assert(SourceArgs.size() == OperandCount && "Invalid number of operands");
-  StringRef Checksum(getString(SourceArgs[TextIdx]));
+  auto ChecksumStr =
+	  !getDbgInst<SPIRVDebug::DebugInfoNone>(SourceArgs[TextIdx])
+                         ? getString(SourceArgs[TextIdx])
+                         : "";
+  StringRef Checksum(ChecksumStr);
   return getDIFile(getString(SourceArgs[FileIdx]), ParseChecksum(Checksum));
 }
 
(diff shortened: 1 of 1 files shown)


KornevNikita:
@AlexeySotkin @AlexeySachkov @MrSidims @Fznamznon review please

lib/SPIRV/SPIRVToLLVMDbgTran.cpp
```diff
@@ -1037,15 +1037,18 @@ SPIRVToLLVMDbgTran::ParseChecksum(StringRef Text) {
...
-  auto KindPos = Text.find(SPIRVDebug::ChecksumKindPrefx);
-  if (KindPos != StringRef::npos) {
-    auto ColonPos = Text.find(":", KindPos);
-    KindPos += string("//__").size();
-    auto KindStr = Text.substr(KindPos, ColonPos - KindPos);
-    auto Checksum = Text.substr(ColonPos).ltrim(':');
-    if (auto Kind = DIFile::getChecksumKind(KindStr)) {
-      size_t ChecksumEndPos = Checksum.find_if_not(llvm::isHexDigit);
-      CS.emplace(Kind.getValue(), Checksum.substr(0, ChecksumEndPos));
+  if (Text.size() > SPIRVDebug::ChecksumKindPrefx.size() &&
```
AlexeySotkin, line 1040: I think checking sizes is redundant. `find` should return `npos` in this case.

lib/SPIRV/SPIRVToLLVMDbgTran.cpp
```diff
@@ -1037,15 +1037,18 @@ SPIRVToLLVMDbgTran::ParseChecksum(StringRef Text) {
...
+      Text.front() != '\0') {
```
AlexeySotkin, line 1041: Here you are checking that first symbol is not '\0'. What if second or third symbol is '\0', is it possible?

lib/SPIRV/SPIRVToLLVMDbgTran.cpp
MrSidims: If I understand StringRef correctly, "non-zero length and an empty data" can be possible only if StringRef object was constructed from nullptr. If I'm right we can just check here if Text.data() != nullptr

KornevNikita: @MrSidims Unfortunately, it doesn't work, because data is not quite empty, it contains '\0'

KornevNikita: @AlexeySotkin we need this because we are checking Text.front(), and if Text.size() is 0, calling Text.front() will cause an error. As I know, there isn't .at() method for StringRef, so we cant avoid this comparing. So I don't see another way to check it. Maybe you have another idea?

lib/SPIRV/SPIRVToLLVMDbgTran.cpp
KornevNikita: Theoretically it is possible, but, in practice, it is not yet found.

lib/SPIRV/SPIRVToLLVMDbgTran.cpp
AlexeySotkin: Sorry, I still don't understand what we are fixing. Is it the case when `Text` contains only one character `\0`?
I think `find` on the next line should return `npos` in this case. Am I missing something?

mlychkov: May we add a test for this patch? It might clarify the case that is fixed.

mlychkov, line 1040: Just a nit: may we return earlier if first two conditional checks are not met? It'd reduce nesting of code blocks.

KornevNikita: @AlexeySotkin There is one moment in `StringRef::find` - `uint8_t Last = Start[N - 1];`
Start contains only '\0', so calling `N - 1` (where `N` is size of prefix `//__CSK_`) fails.

lib/SPIRV/SPIRVToLLVMDbgTran.cpp
```diff
@@ -1007,7 +1007,11 @@ DIFile *SPIRVToLLVMDbgTran::getFile(const SPIRVId SourceId) {
          "DebugSource instruction is expected");
   SPIRVWordVec SourceArgs = Source->getArguments();
   assert(SourceArgs.size() == OperandCount && "Invalid number of operands");
-  StringRef Checksum(getString(SourceArgs[TextIdx]));
+  auto ChecksumStr =
```
mlychkov: LGTM.
Just a nit: Could you please specify a type of string here?

lib/SPIRV/SPIRVToLLVMDbgTran.cpp
```diff
@@ -1007,7 +1007,11 @@ DIFile *SPIRVToLLVMDbgTran::getFile(const SPIRVId SourceId) {
          "DebugSource instruction is expected");
   SPIRVWordVec SourceArgs = Source->getArguments();
   assert(SourceArgs.size() == OperandCount && "Invalid number of operands");
-  StringRef Checksum(getString(SourceArgs[TextIdx]));
+  string ChecksumStr =
```
mlychkov, line 1010:
```suggestion
  std::string ChecksumStr =
```
I see that 'using namespace std' is used here, but this is considered as a good practice to specify namespace (or at least namespace alias) in general.

mlychkov (approved):

MrSidims (approved):

lib/SPIRV/SPIRVToLLVMDbgTran.cpp
```diff
@@ -1007,7 +1007,11 @@ DIFile *SPIRVToLLVMDbgTran::getFile(const SPIRVId SourceId) {
          "DebugSource instruction is expected");
   SPIRVWordVec SourceArgs = Source->getArguments();
   assert(SourceArgs.size() == OperandCount && "Invalid number of operands");
-  StringRef Checksum(getString(SourceArgs[TextIdx]));
+  std::string ChecksumStr =
+      !getDbgInst<SPIRVDebug::DebugInfoNone>(SourceArgs[TextIdx])
+          ? getString(SourceArgs[TextIdx])
+          : "";
+  StringRef Checksum(ChecksumStr);
```
AlexeySotkin, line 1014: I think we can skip allocation of extra `StringRef` variable. I guess `StringRef` can be constructed from `std::string` implicitly if we pass `ChecksumStr` to `ParseChecksum()`

AlexeySotkin (approved):

PR #1041, opened 2021-05-24
Add error generation when using memmove pass with PHINode as source

KornevNikita:
This is a patch to describe a failure in visitMemMoveInst function caused by
using it with PHINode as source.

### DIFF
--- a/lib/SPIRV/SPIRVLowerMemmove.cpp
+++ b/lib/SPIRV/SPIRVLowerMemmove.cpp
@@ -62,7 +62,7 @@ public:
     auto *Dest = I.getRawDest();
     auto *Src = I.getRawSource();
     if (isa<PHINode>(Src))
-      report_fatal_error("llvm.memmove with PHINode as source not supported",
+      report_fatal_error("llvm.memmove of PHI instruction result not supported",
                          false);
     auto *SrcTy = Src->getType();
     if (!isa<ConstantInt>(I.getLength()))
--- a/lib/SPIRV/SPIRVLowerMemmove.cpp
+++ b/lib/SPIRV/SPIRVLowerMemmove.cpp
@@ -61,6 +61,9 @@ public:
     Builder.SetInsertPoint(&I);
     auto *Dest = I.getRawDest();
     auto *Src = I.getRawSource();
+    if (isa<PHINode>(Src))
+      report_fatal_error("llvm.memmove with PHINode as source not supported",
+                         false);
     auto *SrcTy = Src->getType();
     if (!isa<ConstantInt>(I.getLength()))
       // ToDo: for non-constant length, could use a loop to copy a
--- a/test/negative/memmove.with.phinode.src.ll
+++ b/test/negative/memmove.with.phinode.src.ll
@@ -1,0 +1,105 @@
+; It is assumed that the test will fail
+
+; RUN: llvm-as %s -o %t.bc
+; RUN: not --crash llvm-spirv %t.bc -o %t.spv 2>&1 | FileCheck %s
+
+; CHECK: llvm.memmove with PHINode as source not supported
+
+; ModuleID = 'test.bc'
+source_filename = "llvm-link"
+target datalayout = "e-i64:64-v16:16-v24:32-v32:32-v48:64-v96:128-v192:256-v256:256-v512:512-v1024:1024-n8:16:32:64"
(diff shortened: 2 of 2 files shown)


lib/SPIRV/SPIRVLowerMemmove.cpp
```diff
@@ -61,6 +61,9 @@ class SPIRVLowerMemmoveBase : public InstVisitor<SPIRVLowerMemmoveBase> {
     Builder.SetInsertPoint(&I);
     auto *Dest = I.getRawDest();
     auto *Src = I.getRawSource();
+    if (isa<PHINode>(Src))
+      report_fatal_error("llvm.memmove with PHINode as source not supported",
```
mlychkov, line 65:
```suggestion
      report_fatal_error("llvm.memmove of PHI instruction result not supported",
```

test/negative/memmove.with.phinode.src.ll
```diff
@@ -0,0 +1,105 @@
+; It is assumed that the test will fail
```
mlychkov, line 1: May we reduce this test?

test/negative/memmove.with.phinode.src.ll
```diff
@@ -0,0 +1,59 @@
...
+  call void @llvm.memmove.p4i8.p4i8.i64(i8 addrspace(4)* align 8 dereferenceable(32) %5, i8 addrspace(4)* align 8 dereferenceable(32) %phi, i64 32, i1 false)
+  ret void
+}
+
+; Function Attrs: argmemonly nofree nosync nounwind willreturn
+declare void @llvm.memmove.p4i8.p4i8.i64(i8 addrspace(4)* nocapture writeonly, i8 addrspace(4)* nocapture readonly, i64, i1 immarg)
+
+!opencl.spir.version = !{!0, !0, !0, !0, !0, !0, !0, !0, !0, !0, !0}
+!spirv.Source = !{!1, !1, !1, !1, !1, !1, !1, !1, !1, !1, !1}
+!llvm.ident = !{!2, !2, !2, !2, !2, !2, !2, !2, !2, !2, !2}
```
mlychkov, line 52:
```suggestion
!opencl.spir.version = !{!0}
!spirv.Source = !{!1}
!llvm.ident = !{!2}
```

test/negative/memmove.with.phinode.src.ll
```diff
@@ -0,0 +1,59 @@
...
+%"AccessorImplDevice" = type { [1 x i64], [1 x i64], [1 x i64] }
+%class.kfunc = type <{ %"accessor", %"group", i32, [4 x i8] }>
+
+$test_func_object = comdat any
+
+@InvocIndex = external local_unnamed_addr addrspace(1) constant i64, align 8
+@"func_object1" = internal addrspace(3) global %class.kfunc zeroinitializer, align 8
+
+; Function Attrs: convergent norecurse
+define weak_odr dso_local spir_kernel void @test_func_object() local_unnamed_addr comdat !kernel_arg_buffer_location !4 {
```
mlychkov, line 24: I suppose that !kernel_arg_buffer_location metadata can be dropped if there are no arguments to a function.

test/negative/memmove.with.phinode.src.ll
```diff
@@ -0,0 +1,59 @@
...
+
+; ModuleID = 'test.bc'
+source_filename = "llvm-link"
+target datalayout = "e-i64:64-v16:16-v24:32-v32:32-v48:64-v96:128-v192:256-v256:256-v512:512-v1024:1024-n8:16:32:64"
+target triple = "spir64-unknown-unknown-sycldevice"
+
+%"group" = type { [1 x i64], [1 x i64], [1 x i64], [1 x i64] }
+%"accessor" = type { %"AccessorImplDevice", i32 addrspace(1)* }
+%"AccessorImplDevice" = type { [1 x i64], [1 x i64], [1 x i64] }
+%class.kfunc = type <{ %"accessor", %"group", i32, [4 x i8] }>
```
mlychkov, line 16: All these types might be replaced by one simple type as they are not sufficient for the test.

mlychkov (commented):
LGTM.
Just a note: There is an improper copying of memory in the test, but since we only check source of memmove instruction then it may be insignificant here.

AlexeySotkin (approved):

PR #1042, opened 2021-05-24
Facilitate pre-installed Clang detection for in-tree builds

AGindinson:
Comply with LLORG community changes in llvm/llvm-project@a1e6565.

### DIFF
--- a/.github/workflows/check-in-tree-build.yml
+++ b/.github/workflows/check-in-tree-build.yml
@@ -54,9 +54,6 @@ jobs:
           echo "deb https://apt.llvm.org/bionic/ llvm-toolchain-bionic main" | sudo tee -a /etc/apt/sources.list
           echo "deb https://packages.lunarg.com/vulkan bionic main" | sudo tee -a /etc/apt/sources.list
           sudo apt-get update
-          # Linux systems in GitHub Actions already have a pre-installed clang
-          # that is not needed for the translator as we rely on newer version
-          sudo apt-get --purge remove clang-8 clang-9
           sudo apt-get -yq --no-install-suggests --no-install-recommends install \
             clang-${{ env.LLVM_VERSION }} \
             spirv-tools
@@ -95,6 +92,7 @@ jobs:
           make llvm-spirv -j2
       - name: Build tests & test
         run: |
+          export CLANG=/usr/bin/clang-${{ env.LLVM_VERSION }}
           cd build
           make check-llvm-spirv -j2
 
--- a/.github/workflows/check-in-tree-build.yml
+++ b/.github/workflows/check-in-tree-build.yml
@@ -57,6 +57,8 @@ jobs:
           sudo apt-get -yq --no-install-suggests --no-install-recommends install \
             clang-${{ env.LLVM_VERSION }} \
             spirv-tools
+          # Linux systems in GitHub Actions already have older versions of clang
+          # pre-installed. Make sure to override these with the relevant version.
           sudo update-alternatives --install /usr/bin/clang clang /usr/bin/clang-${{ env.LLVM_VERSION }} 1000
       - name: Checkout LLVM sources
         uses: actions/checkout@v2
@@ -92,9 +94,6 @@ jobs:
           make llvm-spirv -j2
       - name: Build tests & test
         run: |
-          # Linux systems in GitHub Actions already have older versions of clang
-          # pre-installed. Explicitly rely on the latest version.
-          export CLANG=/usr/bin/clang-${{ env.LLVM_VERSION }}
           cd build
           make check-llvm-spirv -j2
 
--- a/.github/workflows/check-in-tree-build.yml
+++ b/.github/workflows/check-in-tree-build.yml
@@ -92,6 +92,8 @@ jobs:
           make llvm-spirv -j2
       - name: Build tests & test
         run: |
+          # Linux systems in GitHub Actions already have older versions of clang
+          # pre-installed. Explicitly rely on the latest version.
           export CLANG=/usr/bin/clang-${{ env.LLVM_VERSION }}
           cd build
           make check-llvm-spirv -j2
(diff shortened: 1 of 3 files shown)


.github/workflows/check-in-tree-build.yml
```diff
@@ -61,6 +61,9 @@ jobs:
             clang-${{ env.LLVM_VERSION }} \
             spirv-tools
           sudo update-alternatives --install /usr/bin/clang clang /usr/bin/clang-${{ env.LLVM_VERSION }} 1000
+          # Ensure that the pre-installed Clang path gets detected by the
+          # in-tree-built llvm-lit tool
+          export CLANG=/usr/bin/clang
```
AlexeySachkov, line 66: If I understand correctly, it only applies to the particular step, because each step is launched in its separate shell.

If you are looking for global env variables see `env` section above. I also suggest inspecting/debugging `llvm-project/llvm/utils/lit/lit/llvm/config.py:428:` to see why we failed to find clang in the first place

AGindinson: > It only applies to the particular step, because each step is launched in its separate shell.

Thanks, I was struggling to interpret the failures myself.

> I also suggest inspecting/debugging llvm-project/llvm/utils/lit/lit/llvm/config.py:428: to see why we failed to find clang in the first place.

In progress; I was intending to do this anyway before converting the PR from draft.

test/lit.cfg.py
```diff
@@ -41,7 +41,7 @@
 
 llvm_config.use_default_substitutions()
 
-llvm_config.use_clang()
+llvm_config.use_clang(use_installed=True)
```
AlexeySachkov, line 44: Thanks for providing the info about corresponding llorg change. Having looked at it, I think that for us  itmight be better to use `CLANG` env variable as you suggested in your first version of this PR, because GitHub Actions machines already has pre-installed clang and we do some not very pleasant manipulations to remove outdated clang from there and replace with a newer one. By having a possibility to specifically select a desired clang version through specifying a path to it we could cleanup our GA config files.

AGindinson: Just to clarify, is the suggestion to avoid purging older Clang versions by setting the explicit CLANG path to `/usr/bin/clang-13`? I could definitely try this, although I haven't quite figured out whether this `CLANG` global variable is working without the LIT parameter...

AGindinson: In the event it requires the LIT parameter, we could still keep my LIT changes alongside with replacing purging lines with the environment variable export.

AlexeySachkov: If I understand correctly, `use_installed` option means to look into `PATH` instead of build directory and above all that is `CLANG` env variable which can override both. So, from CI point of view this is not needed anymore.

However, for local `out-of-tree` environments it makes sense to pass it, just to avoid adding new env requirements for our downstream users

AlexeySachkov (approved):
The fix LGTM, but it would be nice to see some comments, in the code/workflow files to describe the motivation, like:
- we use `CLANG` to override looking into build directory (which doesn't exists for `clang` in our CI) and to override looking into `PATH`, because there is an outdated pre-installed (by GitHub Actions) clang
- at the same time we do pass `use_installed` to enable looking into `PATH` - this is done for downstream users convenience in case someone replicates `clang`-less builds - no additional env configuration would be required for such users

test/lit.cfg.py
AGindinson: That's been my reasoning as well. I'll make sure to add comments so that all of this is reflected in the code.

AlexeySotkin:
> ... and to override looking into PATH, because there is an outdated pre-installed (by GitHub Actions) clang

Why do we have to override looking into `PATH` if we have updated the pre-installed clang with the latest version and configured it as a default one with `update-alterntives` command?

AlexeySachkov:
> Why do we have to override looking into PATH if we have updated the pre-installed clang with the latest version and configured it as a default one with update-alterntives command?

Because we have just removed that hack

AGindinson:
@AlexeySotkin, @AlexeySachkov, I've removed the `CLANG` environment setting based on the above discussion - CI results show that `update-alternatives` is sufficient for our purpose.

.github/workflows/check-in-tree-build.yml
```diff
@@ -54,12 +54,11 @@ jobs:
...
           sudo apt-get update
-          # Linux systems in GitHub Actions already have a pre-installed clang
-          # that is not needed for the translator as we rely on newer version
-          sudo apt-get --purge remove clang-8 clang-9
           sudo apt-get -yq --no-install-suggests --no-install-recommends install \
             clang-${{ env.LLVM_VERSION }} \
             spirv-tools
+          # Linux systems in GitHub Actions already have older versions of clang
+          # pre-installed. Make sure to override these with the relevant version.
           sudo update-alternatives --install /usr/bin/clang clang /usr/bin/clang-${{ env.LLVM_VERSION }} 1000
```
AlexeySotkin, line 62: Minor: We might want to use `--set` (manual mode) instead of `--instal` (automatic mode) to make choice of clang even stricter.

AlexeySotkin (approved):

AlexeySachkov (approved):

.github/workflows/check-in-tree-build.yml
AGindinson: That's worth a try - would it be fine if I uploaded and tested this in a separate PR?

AlexeySotkin: Sure

AGindinson: Created https://github.com/KhronosGroup/SPIRV-LLVM-Translator/pull/1043; will open the review after the current PR is merged.

svenvh (approved):
LGTM, thanks for looking into this!

PR #1044, opened 2021-05-25
Map OpVariable to LLVM value before initializer translation

Fznamznon:
When some global variable contains constant function pointer to a
function and this global variable is referenced from the pointed
function, translation crashed because intializer for global variable was
translated before mapping of OpVariable to LLVM value. It happened
because during translation of initializer in such case the original
variable was met for the second time, but it wasn't mapped to a proper
llvm value.

### DIFF
--- a/lib/SPIRV/SPIRVReader.cpp
+++ b/lib/SPIRV/SPIRVReader.cpp
@@ -1667,8 +1667,25 @@ Value *SPIRVToLLVM::transValueWithoutDecoration(SPIRVValue *BV, Function *F,
     bool IsConst = BVar->isConstant();
     llvm::GlobalValue::LinkageTypes LinkageTy = transLinkageType(BVar);
     SPIRVStorageClassKind BS = BVar->getStorageClass();
-    Constant *Initializer = nullptr;
     SPIRVValue *Init = BVar->getInitializer();
+    if (BS == StorageClassFunction && !Init) {
+      assert(BB && "Invalid BB");
+      return mapValue(BV, new AllocaInst(Ty, 0, BV->getName(), BB));
+    }
+
+    SPIRAddressSpace AddrSpace;
+    bool IsVectorCompute =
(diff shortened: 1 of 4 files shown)


Fznamznon:
@AlexeySotkin @MrSidims @AlexeySachkov could you please take a look?

AlexeySotkin (approved):

lib/SPIRV/SPIRVReader.cpp
```diff
@@ -1667,41 +1667,41 @@ Value *SPIRVToLLVM::transValueWithoutDecoration(SPIRVValue *BV, Function *F,
...
+    auto Res = mapValue(BV, LVar);
+    if (Init)
+      Initializer = dyn_cast<Constant>(transValue(Init, F, BB, false));
+    else if (LinkageTy == GlobalValue::CommonLinkage)
+      // In LLVM, variables with common linkage type must be initialized to 0.
+      Initializer = Constant::getNullValue(Ty);
+    else if (BS == SPIRVStorageClassKind::StorageClassWorkgroup)
+      Initializer = dyn_cast<Constant>(UndefValue::get(Ty));
+    else if ((LinkageTy != GlobalValue::ExternalLinkage) &&
+             (BS == SPIRVStorageClassKind::StorageClassCrossWorkgroup))
```
MrSidims, line 1697: That would look nasty, but I guess we also need to check for StorageClassDeviceOnlyINTEL and StorageClassHostOnlyINTEL added in https://github.com/KhronosGroup/SPIRV-LLVM-Translator/pull/579/files

Fznamznon: The only thing I did here is move. So, they weren't checked before too. I think it is a change that we need to make separately.

MrSidims: Agree

MrSidims (approved):

PR #1047, opened 2021-05-31
[SPIR-V] Stop re-writing SPIR-V type map

MrSidims:
There is a map used during forward translation. It links LLVM type
and SPIR-V type, that is being generated, LLVM type here is used as
a key. Due to a bug in type mapping (information is below), it's
possible that for a single pointer type in LLVM, the translator
will create 2 physically-indentical SPIR-V pointer types and with
previous implementation of mapType function the map would be changed
during the second type insertion. This patch prevents it, forcing
to reuse the type that was created first, though we would still have
an unused duplication of this type in SPIR-V module.

Please consider following LLVM IR sample:
%struct._ZTS4Args.Args = type { %struct._ZTS6Object.Object addrspace(4)* }
%struct._ZTS6Object.Object = type { i32 (%struct._ZTS6Object.Object addrspace(4)*, i32)* }

Here we have 2 LLVM type declarations. The translator will recursively
go through structure members and then through element types of the
pointer types, creating the appropriate SPIR-V types and mapping them
on LLVM types. Due to a recursive nature of this algorithm, here we will
have '%struct._ZTS6Object.Object addrspace(4)*' be processed twice with
2 SPIR-V type declaration instruction being created.

### DIFF
--- a/test/pointer_type_mapping.ll
+++ b/test/pointer_type_mapping.ll
@@ -1,0 +1,63 @@
+; RUN: llvm-as %s -o %t.bc
+; RUN: llvm-spirv %t.bc --spirv-ext=+all -spirv-text -o %t
+; RUN: FileCheck < %t %s
+
+; CHECK: Name [[#NAME:]] "struct._ZTS6Object.Object"
+; CHECK-COUNT-1: TypeStruct [[#NAME]]
+; TODO add check count one and remove unused, when the type mapping bug is fixed
+; CHECK: TypePointer [[#UNUSED:]] {{.*}} [[#NAME]]
+; CHECK: TypePointer [[#PTRTY:]] {{.*}} [[#NAME]]
+; CHECK-COUNT-2: TypePointer {{.*}} {{.*}} [[#PTRTY]]
+; CHECK-NOT: TypePointer {{.*}} {{.*}} [[#UNUSED]]
+
+; ModuleID = 'sycl_test.bc'
+source_filename = "sycl_test.cpp"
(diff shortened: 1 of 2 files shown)


test/pointer_type_mapping.ll
```diff
@@ -0,0 +1,63 @@
+; RUN: llvm-as %s -o %t.bc
+; RUN: llvm-spirv %t.bc --spirv-ext=+all -spirv-text -o %t
+; RUN: FileCheck < %t %s
+
+; CHECK: Name [[#NAME:]] "struct._ZTS6Object.Object"
+; CHECK-COUNT-1: TypeStruct [[#NAME]]
+; TODO add check count one and remove unused, when the type mapping bug is fixed
```
svenvh, line 7: You could already add the check and XFAIL the test perhaps?

AlexeySachkov: Since this tests is also checks that only *single* `TypeStruct` is actually being used, it wouldn't be clear why does it fail: due to one bug or another?

I mean, we have a bigger general problem with that mapping and a particular case where it causes problems. The case is workarounded and we have a test for that. The test is not entirely finished, because the bigger problem is still there.

MrSidims: TBH this PR in the current stage is more about highlighting the issue. I would prefer to fix first and push it in the same branch (unless fixing it takes a lot of time).

svenvh: Fair enough, XFAILing this test would mean the work-around is no longer tested indeed.  So fine to leave it as is then.

MrSidims: Yet if we want it be merged anyway, I should apply [this](https://github.com/intel/llvm/pull/3856#discussion_r642806327) comment

AlexeySotkin (approved):

PR #1050, opened 2021-06-03
Improve OpSpecConstantOp handling for unary and integer comparison operators

svenvh:
This is a continuation of #1037 to improve OpSpecConstantOp handling.

#### Handle OpSpecConstantOp for integer comparisons
This revealed that OpINotEqual was missing from
isSpecConstantOpAllowedOp.

#### Handle OpSpecConstantOp for SNegate, Not, and LogicalNot
Use an IRBuilder to get automatic folding where possible.

#### Extend OpSpecConstantOp testing for initializers
Co-authored-by: @StuartDBrady 

#### Use IRBuilder for folding
Rely on the IRBuilder to fold into constant expressions if possible.


StuartDBrady (approved):
Looks good to me.

PR #1051, opened 2021-06-03
[NFC] Use SPIRVBinaryInst for OpFMod and OpSMod

svenvh:
`OpFMod` and `OpSMod` were implemented using a separate class, but
seemingly without an underlying motivation.  Instead of a separate
class, reuse the `SPIRVBinaryInst` class to unify `OpFMod` and `OpSMod`
handling with other binary instructions.

This change enables supporting FMod and SMod operators with
`OpSpecConstantOp`, which will be done in a future commit.

### DIFF
--- a/lib/SPIRV/libSPIRV/SPIRVInstruction.h
+++ b/lib/SPIRV/libSPIRV/SPIRVInstruction.h
@@ -654,7 +654,6 @@ template <Op OC>
 class SPIRVBinaryInst
     : public SPIRVInstTemplate<SPIRVBinary, OC, true, 5, false> {};
 
-/* ToDo: SMod and FMod to be added */
 #define _SPIRV_OP(x) typedef SPIRVBinaryInst<Op##x> SPIRV##x;
 _SPIRV_OP(IAdd)
 _SPIRV_OP(FAdd)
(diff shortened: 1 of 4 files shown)


svenvh:
clang-tidy failure seems to be unrelated to my change.

StuartDBrady (approved):
Looks good to me.

lib/SPIRV/libSPIRV/SPIRVInstruction.h
```diff
@@ -666,7 +665,9 @@ _SPIRV_OP(UDiv)
 _SPIRV_OP(SDiv)
 _SPIRV_OP(FDiv)
 _SPIRV_OP(SRem)
+_SPIRV_OP(SMod)
 _SPIRV_OP(FRem)
+_SPIRV_OP(FMod)
```
StuartDBrady, line 670: Nit: it might be better to reorder this as "SRem, FRem, UMod, SMod, FMod" to as to preserve the pattern of listing the (signed, and then unsigned, if applicable) integer op followed by the corresponding floating point op, which occurs throughout this sequence of typedefs.

svenvh: I have followed the order of the `Op` enum values from `spirv.hpp`:
```
    OpSRem = 138,
    OpSMod = 139,
    OpFRem = 140,
    OpFMod = 141,
```
I'd argue sticking to the enum order makes more sense than inventing a different pattern.  Funny enough, the SPIR-V specification's enum values follow the pattern you're describing for most of the binary arithmetic ops.

StuartDBrady: > I'd argue sticking to the enum order makes more sense than inventing a different pattern. Funny enough, the SPIR-V specification's enum values follow the pattern you're describing for most of the binary arithmetic ops.

Okay, that makes sense, although UMod would be out of place, in that case (although this is not new).

PR #1058, opened 2021-06-08
[SPIR-V] Fix duplicate pointers

MrSidims:
Current implementation creates forward pointer only if direct pointee
was already discovered during type traversal. But it is possible to have a
cyclic dependency when pointer points to already discovered type via multiple
intermediate types.

Modified algorithm tracks cyclic dependency case via return value of the
recursive algorithm.


MrSidims:
Upstream of https://github.com/intel/llvm/pull/3883
Tagging @AlexeySotkin

AlexeySachkov (approved):

AlexeySotkin:
It seems we have a clang-tidy tooling issue not related to this patch.

PR #1060, opened 2021-06-08
Implement support for dynamic memmove

Fznamznon:
Use LLVM `expandMemMoveAsLoop` function in case memmove intrinsic use has
non-constant length.

### DIFF
--- a/lib/SPIRV/SPIRVLowerMemmove.cpp
+++ b/lib/SPIRV/SPIRVLowerMemmove.cpp
@@ -41,7 +41,6 @@
 #include "libSPIRV/SPIRVDebug.h"
 
 #include "llvm/IR/IRBuilder.h"
-#include "llvm/IR/InstVisitor.h"
 #include "llvm/IR/IntrinsicInst.h"
 #include "llvm/IR/Module.h"
 #include "llvm/IR/PassManager.h"
@@ -106,12 +105,11 @@ public:
     I.dropAllReferences();
     I.eraseFromParent();
   }
-  bool expandMemIntrinsicUses(Function &F) {
+  bool expandMemMoveIntrinsicUses(Function &F) {
     bool Changed = false;
 
-    for (auto I = F.user_begin(), E = F.user_end(); I != E;) {
-      MemMoveInst *Inst = cast<MemMoveInst>(*I);
-      ++I;
+    for (User *U : F.users()) {
+      MemMoveInst *Inst = cast<MemMoveInst>(U);
       if (!isa<ConstantInt>(Inst->getLength())) {
         expandMemMoveAsLoop(Inst);
         Inst->eraseFromParent();
@@ -132,7 +130,7 @@ public:
         continue;
 
       if (F.getIntrinsicID() == Intrinsic::memmove)
-        if (expandMemIntrinsicUses(F))
+        if (expandMemMoveIntrinsicUses(F))
           Changed = true;
     }
 
--- a/lib/SPIRV/SPIRVLowerMemmove.cpp
+++ b/lib/SPIRV/SPIRVLowerMemmove.cpp
@@ -130,8 +130,7 @@ public:
         continue;
 
       if (F.getIntrinsicID() == Intrinsic::memmove)
-        if (expandMemMoveIntrinsicUses(F))
-          Changed = true;
+        Changed |= expandMemMoveIntrinsicUses(F);
     }
 
     verifyRegularizationPass(M, "SPIRVLowerMemmove");
--- a/lib/SPIRV/SPIRVLowerMemmove.cpp
+++ b/lib/SPIRV/SPIRVLowerMemmove.cpp
@@ -108,7 +108,7 @@ public:
   bool expandMemMoveIntrinsicUses(Function &F) {
     bool Changed = false;
 
-    for (User *U : F.users()) {
+    for (User *U : make_early_inc_range(F.users())) {
       MemMoveInst *Inst = cast<MemMoveInst>(U);
       if (!isa<ConstantInt>(Inst->getLength())) {
         expandMemMoveAsLoop(Inst);
--- a/lib/SPIRV/SPIRVLowerMemmove.cpp
+++ b/lib/SPIRV/SPIRVLowerMemmove.cpp
@@ -46,17 +46,17 @@
 #include "llvm/IR/Module.h"
 #include "llvm/IR/PassManager.h"
 #include "llvm/Pass.h"
+#include "llvm/Transforms/Utils/LowerMemIntrinsics.h"
 
 using namespace llvm;
 using namespace SPIRV;
 
 namespace SPIRV {
 
-class SPIRVLowerMemmoveBase : public InstVisitor<SPIRVLowerMemmoveBase> {
+class SPIRVLowerMemmoveBase {
 public:
-  SPIRVLowerMemmoveBase() : Context(nullptr) {}
-  virtual ~SPIRVLowerMemmoveBase() {}
-  virtual void visitMemMoveInst(MemMoveInst &I) {
+  SPIRVLowerMemmoveBase() : Context(nullptr), Mod(nullptr) {}
+  void LowerMemMoveInst(MemMoveInst &I) {
     IRBuilder<> Builder(I.getParent());
     Builder.SetInsertPoint(&I);
     auto *Dest = I.getRawDest();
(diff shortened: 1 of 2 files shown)


lib/SPIRV/SPIRVLowerMemmove.cpp
```diff
@@ -111,13 +106,38 @@ class SPIRVLowerMemmoveBase : public InstVisitor<SPIRVLowerMemmoveBase> {
     I.dropAllReferences();
     I.eraseFromParent();
   }
+  bool expandMemIntrinsicUses(Function &F) {
+    bool Changed = false;
+
+    for (auto I = F.user_begin(), E = F.user_end(); I != E;) {
+      MemMoveInst *Inst = cast<MemMoveInst>(*I);
+      ++I;
```
svenvh, line 114: Perhaps worth putting a comment why the increment is done here, and not as part of the for-statement.

lib/SPIRV/SPIRVLowerMemmove.cpp
```diff
@@ -46,17 +46,17 @@
...
 #include "llvm/Pass.h"
+#include "llvm/Transforms/Utils/LowerMemIntrinsics.h"
 
 using namespace llvm;
 using namespace SPIRV;
 
 namespace SPIRV {
 
-class SPIRVLowerMemmoveBase : public InstVisitor<SPIRVLowerMemmoveBase> {
+class SPIRVLowerMemmoveBase {
```
svenvh, line 56: Can we drop the `InstVisitor.h` include now?

lib/SPIRV/SPIRVLowerMemmove.cpp
Fznamznon: I'm not sure it is needed, I replaced this loop with `for(User *U : ... `

lib/SPIRV/SPIRVLowerMemmove.cpp
Fznamznon: Sure.

lib/SPIRV/SPIRVLowerMemmove.cpp
svenvh: That's also a solution.  But perhaps you need to use `make_early_inc_range(F.users())` then, as the loop may modify the range it is iterating over?

lib/SPIRV/SPIRVLowerMemmove.cpp
```diff
@@ -65,11 +64,6 @@ class SPIRVLowerMemmoveBase : public InstVisitor<SPIRVLowerMemmoveBase> {
       report_fatal_error("llvm.memmove of PHI instruction result not supported",
```
AlexeySotkin, line 64: I guess, now we can handle this case with `expandMemMoveAsLoop`, right?

lib/SPIRV/SPIRVLowerMemmove.cpp
```diff
@@ -111,13 +105,37 @@ class SPIRVLowerMemmoveBase : public InstVisitor<SPIRVLowerMemmoveBase> {
...
-    visit(M);
+    bool Changed = false;
+
+    for (Function &F : M) {
+      if (!F.isDeclaration())
+        continue;
+
+      if (F.getIntrinsicID() == Intrinsic::memmove)
+        if (expandMemMoveIntrinsicUses(F))
+          Changed = true;
```
mlychkov, line 134:
```suggestion
        Changed |= expandMemMoveIntrinsicUses(F);
```

test/llvm-intrinsics/dynamic-memmove.ll
```diff
@@ -0,0 +1,51 @@
...
+; CHECK-SPIRV-NOT: llvm.memmove
+
+target datalayout = "e-m:e-i64:64-f80:128-n8:16:32:64-S128"
+target triple = "spir64-unknown-unknown"
+
+declare void @llvm.memmove.p1i8.p1i8.i64(i8 addrspace(1)* nocapture, i8 addrspace(1)* nocapture readonly, i64, i1)
+
+define spir_func void @memmove_caller(i8 addrspace(1)* %dst, i8 addrspace(1)* %src, i64 %n) {
+entry:
+  call void @llvm.memmove.p1i8.p1i8.i64(i8 addrspace(1)* %dst, i8 addrspace(1)* %src, i64 %n, i1 false)
```
mlychkov, line 17: Could you please add two memmove instructions to check that there is no iterator invalidation after eraseFromParent() call?

test/llvm-intrinsics/dynamic-memmove.ll
```diff
@@ -0,0 +1,51 @@
...
+  ret void
+
+; CHECK-LLVM: @memmove_caller(i8 addrspace(1)* [[DST:%.*]], i8 addrspace(1)* [[SRC:%.*]], i64 [[N:%.*]])
+; CHECK-LLVM:  [[SRC_CMP:%.*]] = ptrtoint i8 addrspace(1)* %src to i64
```
mlychkov, line 21: Is it worth checking whole loop code generated by LLVM in SPIR-V translator test?
Is it possible to left checks for only some minimal subset of instructions?

lib/SPIRV/SPIRVLowerMemmove.cpp
Fznamznon: I guess so, since I haven't seen any disclaimers that `expandMemMoveAsLoop` doesn't work with PHI instructions. I tried to keep the current PR focused on memmove of non-constant length, so I suggest to handle PHI instuctions in a separate patch.

test/llvm-intrinsics/dynamic-memmove.ll
Fznamznon: Probably not, I didn't come up with any minimal subset though. Do you have any suggestions?

test/llvm-intrinsics/dynamic-memmove.ll
Fznamznon: I've tried doing it locally, it actually doesn't crash and looks ok, but the doc for `eraseFromParent()` says "This method unlinks 'this' from the containing basic block and deletes it.", so using of `make_early_inc_range` seems necessary as well.

test/llvm-intrinsics/dynamic-memmove.ll
mlychkov: I'd left only first 6 check strings. for example. It is enough to see that some loop preparation instructions are added that use src, dst and N from memmove instruction.

test/llvm-intrinsics/dynamic-memmove.ll
```diff
@@ -0,0 +1,51 @@
...
+; CHECK-LLVM:  [[DST_CMP:%.*]] = ptrtoint i8 addrspace(1)* %dst to i64
```
mlychkov, line 22:
```suggestion
; CHECK-LLVM:  [[SRC_CMP:%.*]] = ptrtoint i8 addrspace(1)* [[SRC]] to i64
; CHECK-LLVM:  [[DST_CMP:%.*]] = ptrtoint i8 addrspace(1)* [[DST]] to i64
```

test/llvm-intrinsics/dynamic-memmove.ll
mlychkov: Probably, we may even add only check for spirv code part and skip translation back into llvm.

Fznamznon: I prefer checking in LLVM IR, since it is more readable.

test/llvm-intrinsics/dynamic-memmove.ll
Fznamznon: Done.

lib/SPIRV/SPIRVLowerMemmove.cpp
Fznamznon: I think you are right, thanks. It seems `eraseFromParent` deletes instruction and can invalidate the range.

lib/SPIRV/SPIRVLowerMemmove.cpp
Fznamznon: @AlexeySotkin , are you ok if I add support for PHI instructions in the next patch?

Fznamznon:
I cannot re-request review from @mlychkov via GitHub interface, so, Mikhail, please take a look.

lib/SPIRV/SPIRVLowerMemmove.cpp
AlexeySotkin: Sure, no problem.

mlychkov (approved):
Thank you for updates. LGTM.

AlexeySotkin (approved):

svenvh (approved):
LGTM, thanks!

PR #1062, opened 2021-06-09
Implemented SPIR-V support for loopcount attributes

pauzinl:
There are 3 metadata, which map on `LoopControlLoopCountINTELMask` . Mask has 3 parameters. Default values are `-1`.

### DIFF
--- a/lib/SPIRV/SPIRVReader.cpp
+++ b/lib/SPIRV/SPIRVReader.cpp
@@ -1000,38 +1000,41 @@ void SPIRVToLLVM::setLLVMLoopMetadata(const LoopInstType *LM,
     // as 2 SPIRVWords (int32)
     uint64_t LoopCountMin =
         static_cast<uint64_t>(LoopControlParameters[NumParam++]);
-    LoopCountMin =
-        LoopCountMin | static_cast<uint64_t>(LoopControlParameters[NumParam++])
-                           << 32;
-    if (static_cast<int64_t>(LoopCountMin) > -1) {
+    LoopCountMin |= static_cast<uint64_t>(LoopControlParameters[NumParam++])
+                    << 32;
+    if (static_cast<int64_t>(LoopCountMin) >= 0) {
       Metadata.push_back(llvm::MDNode::get(
           *Context,
           getMetadataFromNameAndParameter("llvm.loop.intel.loopcount_min",
                                           static_cast<int64_t>(LoopCountMin))));
+      assert(NumParam <= LoopControlParameters.size() &&
+             "Missing loop control parameter!");
     }
 
     uint64_t LoopCountMax =
         static_cast<uint64_t>(LoopControlParameters[NumParam++]);
-    LoopCountMax =
-        LoopCountMax | static_cast<uint64_t>(LoopControlParameters[NumParam++])
-                           << 32;
-    if (static_cast<int64_t>(LoopCountMax) > -1) {
+    LoopCountMax |= static_cast<uint64_t>(LoopControlParameters[NumParam++])
+                    << 32;
+    if (static_cast<int64_t>(LoopCountMax) >= 0) {
       Metadata.push_back(llvm::MDNode::get(
           *Context,
           getMetadataFromNameAndParameter("llvm.loop.intel.loopcount_max",
                                           static_cast<int64_t>(LoopCountMax))));
+      assert(NumParam <= LoopControlParameters.size() &&
+             "Missing loop control parameter!");
     }
 
     uint64_t LoopCountAvg =
         static_cast<uint64_t>(LoopControlParameters[NumParam++]);
-    LoopCountAvg =
-        LoopCountAvg | static_cast<uint64_t>(LoopControlParameters[NumParam++])
-                           << 32;
-    if (static_cast<int64_t>(LoopCountAvg) > -1) {
+    LoopCountAvg |= static_cast<uint64_t>(LoopControlParameters[NumParam++])
+                    << 32;
+    if (static_cast<int64_t>(LoopCountAvg) >= 0) {
       Metadata.push_back(llvm::MDNode::get(
           *Context,
           getMetadataFromNameAndParameter("llvm.loop.intel.loopcount_avg",
                                           static_cast<int64_t>(LoopCountAvg))));
+      assert(NumParam <= LoopControlParameters.size() &&
+             "Missing loop control parameter!");
     }
   }
   llvm::MDNode *Node = llvm::MDNode::get(*Context, Metadata);
--- a/lib/SPIRV/SPIRVReader.cpp
+++ b/lib/SPIRV/SPIRVReader.cpp
@@ -996,8 +996,8 @@ void SPIRVToLLVM::setLLVMLoopMetadata(const LoopInstType *LM,
   if (LC & LoopControlNoFusionINTELMask)
     Metadata.push_back(getMetadataFromName("llvm.loop.fusion.disable"));
   if (LC & spv::internal::LoopControlLoopCountINTELMask) {
-  // LoopCountINTELMask parameters are int64 and each parameter is stored
-  // as 2 SPIRVWords (int32)
+    // LoopCountINTELMask parameters are int64 and each parameter is stored
+    // as 2 SPIRVWords (int32)
     uint64_t LoopCountMin =
         static_cast<uint64_t>(LoopControlParameters[NumParam++]);
     LoopCountMin =
@@ -1009,7 +1009,7 @@ void SPIRVToLLVM::setLLVMLoopMetadata(const LoopInstType *LM,
           getMetadataFromNameAndParameter("llvm.loop.intel.loopcount_min",
                                           static_cast<int64_t>(LoopCountMin))));
     }
- 
+
     uint64_t LoopCountMax =
         static_cast<uint64_t>(LoopControlParameters[NumParam++]);
     LoopCountMax =
--- a/lib/SPIRV/SPIRVReader.cpp
+++ b/lib/SPIRV/SPIRVReader.cpp
@@ -739,12 +739,13 @@ SPIRVToLLVM::getMetadataFromNameAndParameter(std::string Name,
               ConstantInt::get(Type::getInt32Ty(*Context), Parameter))};
 }
 
-inline std::vector<llvm::Metadata *>
-SPIRVToLLVM::getMetadataFromNameAndParameter(std::string Name,
-                                             int64_t Parameter) {
-  return {MDString::get(*Context, Name),
-          ConstantAsMetadata::get(
-              ConstantInt::get(Type::getInt64Ty(*Context), Parameter))};
+inline llvm::MDNode *SPIRVToLLVM::getMetadataFromNameAndParameter(
+    std::string Name, int64_t Parameter) {
+  std::vector<llvm::Metadata *> Metadata = {
+      MDString::get(*Context, Name),
+      ConstantAsMetadata::get(
+          ConstantInt::get(Type::getInt64Ty(*Context), Parameter))};
+  return llvm::MDNode::get(*Context, Metadata);
 }
 
 template <typename LoopInstType>
@@ -998,17 +999,16 @@ void SPIRVToLLVM::setLLVMLoopMetadata(const LoopInstType *LM,
   if (LC & spv::internal::LoopControlLoopCountINTELMask) {
     // LoopCountINTELMask parameters are int64 and each parameter is stored
     // as 2 SPIRVWords (int32)
+    assert(NumParam + 6 <= LoopControlParameters.size() &&
+           "Missing loop control parameter!");
+
     uint64_t LoopCountMin =
         static_cast<uint64_t>(LoopControlParameters[NumParam++]);
     LoopCountMin |= static_cast<uint64_t>(LoopControlParameters[NumParam++])
                     << 32;
     if (static_cast<int64_t>(LoopCountMin) >= 0) {
-      Metadata.push_back(llvm::MDNode::get(
-          *Context,
-          getMetadataFromNameAndParameter("llvm.loop.intel.loopcount_min",
-                                          static_cast<int64_t>(LoopCountMin))));
-      assert(NumParam <= LoopControlParameters.size() &&
-             "Missing loop control parameter!");
+      Metadata.push_back(getMetadataFromNameAndParameter(
+          "llvm.loop.intel.loopcount_min", static_cast<int64_t>(LoopCountMin)));
     }
 
     uint64_t LoopCountMax =
@@ -1016,12 +1016,8 @@ void SPIRVToLLVM::setLLVMLoopMetadata(const LoopInstType *LM,
     LoopCountMax |= static_cast<uint64_t>(LoopControlParameters[NumParam++])
                     << 32;
     if (static_cast<int64_t>(LoopCountMax) >= 0) {
-      Metadata.push_back(llvm::MDNode::get(
-          *Context,
-          getMetadataFromNameAndParameter("llvm.loop.intel.loopcount_max",
-                                          static_cast<int64_t>(LoopCountMax))));
-      assert(NumParam <= LoopControlParameters.size() &&
-             "Missing loop control parameter!");
+      Metadata.push_back(getMetadataFromNameAndParameter(
+          "llvm.loop.intel.loopcount_max", static_cast<int64_t>(LoopCountMax)));
     }
 
     uint64_t LoopCountAvg =
@@ -1029,12 +1025,8 @@ void SPIRVToLLVM::setLLVMLoopMetadata(const LoopInstType *LM,
     LoopCountAvg |= static_cast<uint64_t>(LoopControlParameters[NumParam++])
                     << 32;
     if (static_cast<int64_t>(LoopCountAvg) >= 0) {
-      Metadata.push_back(llvm::MDNode::get(
-          *Context,
-          getMetadataFromNameAndParameter("llvm.loop.intel.loopcount_avg",
-                                          static_cast<int64_t>(LoopCountAvg))));
-      assert(NumParam <= LoopControlParameters.size() &&
-             "Missing loop control parameter!");
+      Metadata.push_back(getMetadataFromNameAndParameter(
+          "llvm.loop.intel.loopcount_avg", static_cast<int64_t>(LoopCountAvg)));
     }
   }
   llvm::MDNode *Node = llvm::MDNode::get(*Context, Metadata);
--- a/lib/SPIRV/SPIRVReader.cpp
+++ b/lib/SPIRV/SPIRVReader.cpp
@@ -995,39 +995,43 @@ void SPIRVToLLVM::setLLVMLoopMetadata(const LoopInstType *LM,
   }
   if (LC & LoopControlNoFusionINTELMask)
     Metadata.push_back(getMetadataFromName("llvm.loop.fusion.disable"));
-
   if (LC & spv::internal::LoopControlLoopCountINTELMask) {
-    uint64_t LoopCountMin = 0;
-    LoopCountMin = static_cast<uint64_t>(LoopControlParameters[NumParam++]);
-    LoopCountMin = LoopCountMin | static_cast<uint64_t>(
-                                  LoopControlParameters[NumParam++]) << 32;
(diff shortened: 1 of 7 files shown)


MrSidims:
@AlexeySotkin @AlexeySachkov  could you please approve the tests to run for this PR

lib/SPIRV/SPIRVWriter.cpp
```diff
@@ -1210,11 +1216,39 @@ LLVMToSPIRVBase::getLoopControl(const BranchInst *Branch,
...
+                                  static_cast<SPIRVWord>(LoopCountMin >> 32));
+    ParametersToSort.emplace_back(spv::internal::LoopControlLoopCountINTELMask,
+                                  static_cast<SPIRVWord>(LoopCountMax));
+    ParametersToSort.emplace_back(spv::internal::LoopControlLoopCountINTELMask,
+                                  static_cast<SPIRVWord>(LoopCountMax >> 32));
+    ParametersToSort.emplace_back(spv::internal::LoopControlLoopCountINTELMask,
+                                  static_cast<SPIRVWord>(LoopCountAvg));
+    ParametersToSort.emplace_back(spv::internal::LoopControlLoopCountINTELMask,
+                                  static_cast<SPIRVWord>(LoopCountAvg >> 32));
+  }
```
MrSidims, line 1251: This code needs to have a comment, that `LoopCountINTELMask` have int64 literal parameters and we need to store int64 into 2 SPIRVWords.

lib/SPIRV/SPIRVReader.cpp
```diff
@@ -987,6 +995,41 @@ void SPIRVToLLVM::setLLVMLoopMetadata(const LoopInstType *LM,
   }
   if (LC & LoopControlNoFusionINTELMask)
     Metadata.push_back(getMetadataFromName("llvm.loop.fusion.disable"));
+
```
MrSidims, line 998: No need for this blank line

lib/SPIRV/SPIRVReader.cpp
```diff
@@ -987,6 +995,41 @@ void SPIRVToLLVM::setLLVMLoopMetadata(const LoopInstType *LM,
...
+  if (LC & spv::internal::LoopControlLoopCountINTELMask) {
+    uint64_t LoopCountMin = 0;
+    LoopCountMin = static_cast<uint64_t>(LoopControlParameters[NumParam++]);
```
MrSidims, line 1001: `uint64_t LoopCountMin = LoopCountMin = static_cast<uint64_t>(LoopControlParameters[NumParam++]);`

same for Max and Avg

lib/SPIRV/SPIRVReader.cpp
```diff
@@ -987,6 +995,41 @@ void SPIRVToLLVM::setLLVMLoopMetadata(const LoopInstType *LM,
   }
   if (LC & LoopControlNoFusionINTELMask)
     Metadata.push_back(getMetadataFromName("llvm.loop.fusion.disable"));
+
+  if (LC & spv::internal::LoopControlLoopCountINTELMask) {
```
MrSidims, line 999: I believe here we need a comment about size of LoopCountINTELMask parameters (int64) and that they are stored as SPIRVWords (int32).

MrSidims (commented):
Overall LGTM
@AGindinson please take a look


> There are 3 metadata, which map on LoopControlLoopCountINTELMask . Mask has 3 parameters. Default values are -1.

What about:
"These attributes generated llvm.loop.intel.loopcount_min/max/avg metadata in LLVM IR, which are mapped on LoopControlLoopCountINTELMask. There are three 64-bit integer literal operands of LoopMerge instruction associated with the mask. The first operand is the minimum iteration count, the second is the maximum iteration count, and the third is the expected iteration count. Default values for the parameters are -1.

Spec: https://github.com/KhronosGroup/SPIRV-Registry/pull/115
"

lib/SPIRV/SPIRVReader.cpp
```diff
@@ -739,6 +739,14 @@ SPIRVToLLVM::getMetadataFromNameAndParameter(std::string Name,
               ConstantInt::get(Type::getInt32Ty(*Context), Parameter))};
 }
 
+inline std::vector<llvm::Metadata *>
+SPIRVToLLVM::getMetadataFromNameAndParameter(std::string Name,
+                                             int64_t Parameter) {
```
AGindinson, line 744: Would it make sense to remove the preceding overload for `SPIRVWord` (a.k.a. `unsigned`) and only leave this `int64_t` signature? It's doubtable that we'd run into any unsigned token number that would exceed 32 bits...

lib/SPIRV/SPIRVWriter.cpp
```diff
@@ -1111,6 +1111,12 @@ LLVMToSPIRVBase::getLoopControl(const BranchInst *Branch,
 
   size_t LoopControl = spv::LoopControlMaskNone;
   std::vector<std::pair<SPIRVWord, SPIRVWord>> ParametersToSort;
+  // We need to create these variables here because it might be the case when
+  // metadata is not defined for one or two of these parameters and in this
+  // case we need to store "-1" value
+  int64_t LoopCountMin = -1;
+  int64_t LoopCountMax = -1;
+  int64_t LoopCountAvg = -1;
```
AGindinson, line 1119: An alternative is to wrap these numbers in a function-scope struct with a name that would correspond to the implemented extension, something akin to:
```
struct LoopCountInfo {
  int64_t Min = -1, Max = -1, Avg = -1;
} LoopCount;
```
Not sure if it gains much in terms of code readability (the function still has to be decluttered some day), but one could argue that a single struct-typed object is easier to eye-track than 3 distinct variables.

test/transcoding/SPV_INTEL_fpga_loop_controls/FPGALoopAttr.ll
```diff
@@ -186,12 +276,21 @@ attributes #0 = { convergent noinline nounwind optnone "correctly-rounded-divide
 !9 = distinct !{!9, !10}
 !10 = !{!"llvm.loop.max_concurrency.count", i32 2}
 !11 = distinct !{!11, !8, !10}
+!12 = distinct !{!12, !13, !14}
+!13 = !{!"llvm.loop.mustprogress"}
+!14 = !{!"llvm.loop.intel.loopcount_avg", i64 12}
+!15 = distinct !{!15, !13, !16, !17, !18}
+!16 = !{!"llvm.loop.intel.loopcount_min", i64 4}
+!17 = !{!"llvm.loop.intel.loopcount_max", i64 4294967396}
```
AGindinson, line 284: A comment that would tie this number to a hexadecimal/binary counterpart would be appreciated.

AGindinson (approved):
LGTM, just some "highly optional" stylistic nits

lib/SPIRV/SPIRVReader.cpp
AlexeySachkov: I wouldn't recommend to remove `SPIRVWord` overload, because `SPIRVWord` is directly mapped to the spec and widely used in the translator implementation. Moreover, mapping some 32-bit value down to 64-bit value might confuse SPIR-V -> LLVM IR -> SPIR-V roundtrip, because LLVM IR -> SPIR-V might expect 32-bit value.

Instead, I would suggest to accept two SPIR-V words here and automatically bundle them into a single 64-bit value within the function. Note: if we go this way, we should probably leave a comment near this function that it produces a metadata with a single value and not with two values

lib/SPIRV/SPIRVWriter.cpp
AlexeySachkov:
```
struct LoopCountInfo {
  static constexpr int64_t Min = -1;
  static constexpr Max = -1;
  static constexpr Avg = -1;
};
```

Ideally, such struct (or just a bunch of constexpr variables) should reside in `spv` or `spv::internal` namespace similar to `spv::LoopControlMaskNone`. The idea is that if we have some special value for this mask, let's define it somewhere near the mask definition.

AGindinson: The only thing I'd argue with here is `static constexpr` - different loops may want to use different instances of the struct, and the value of the fields may change based on the actual attribute values.

AlexeySachkov: My understanding was that here we are talking about some generic default values, which probably should be documented in the corresponding SPIR-V extension and even if they depend on the attribute, I would assume that there are finite set of such default values for each attribute, which means that we can still define them as `constexpr`

AGindinson: > My understanding was that here we are talking about some generic default values

Not quite; these are needed as in-scope variables that are assigned to on lines 1219-1233 and read from on lines 1238-1253.

lib/SPIRV/SPIRVReader.cpp
```diff
@@ -987,6 +995,45 @@ void SPIRVToLLVM::setLLVMLoopMetadata(const LoopInstType *LM,
...
   if (LC & LoopControlNoFusionINTELMask)
     Metadata.push_back(getMetadataFromName("llvm.loop.fusion.disable"));
+  if (LC & spv::internal::LoopControlLoopCountINTELMask) {
+    // LoopCountINTELMask parameters are int64 and each parameter is stored
+    // as 2 SPIRVWords (int32)
+    uint64_t LoopCountMin =
+        static_cast<uint64_t>(LoopControlParameters[NumParam++]);
+    LoopCountMin =
+        LoopCountMin | static_cast<uint64_t>(LoopControlParameters[NumParam++])
+                           << 32;
```
MrSidims, line 1005: I believe it worth to outline the packing algorithm into a separate function, so we can reuse it later for the upcoming feature and not to re-invent a wheel.

mlychkov, line 1005:
```suggestion
    LoopCountMin |=
        static_cast<uint64_t>(LoopControlParameters[NumParam++]) << 32;
```

lib/SPIRV/SPIRVReader.cpp
```diff
@@ -987,6 +995,45 @@ void SPIRVToLLVM::setLLVMLoopMetadata(const LoopInstType *LM,
...
+    if (static_cast<int64_t>(LoopCountMin) > -1) {
```
mlychkov, line 1006: This is just a nit, but I think that it is used to write non negative checks as `>= 0`.
Otherwise, it looks like -1 is some special value there for that parameter.

lib/SPIRV/SPIRVReader.cpp
```diff
@@ -987,6 +995,45 @@ void SPIRVToLLVM::setLLVMLoopMetadata(const LoopInstType *LM,
   }
   if (LC & LoopControlNoFusionINTELMask)
     Metadata.push_back(getMetadataFromName("llvm.loop.fusion.disable"));
+  if (LC & spv::internal::LoopControlLoopCountINTELMask) {
+    // LoopCountINTELMask parameters are int64 and each parameter is stored
+    // as 2 SPIRVWords (int32)
+    uint64_t LoopCountMin =
+        static_cast<uint64_t>(LoopControlParameters[NumParam++]);
```
mlychkov, line 1002: Could you please add an assert check that NumParam <=  LoopControlParameters.size()?

lib/SPIRV/SPIRVWriter.cpp
```diff
@@ -1210,11 +1216,41 @@ LLVMToSPIRVBase::getLoopControl(const BranchInst *Branch,
...
+          LoopControl |= spv::internal::LoopControlLoopCountINTELMask;
         }
       }
     }
   }
-
+  if (LoopControl & spv::internal::LoopControlLoopCountINTELMask) {
+    // LoopCountINTELMask have int64 literal parameters and we need to store
+    // int64 into 2 SPIRVWords
+    ParametersToSort.emplace_back(spv::internal::LoopControlLoopCountINTELMask,
```
mlychkov, line 1241: From description of std::sort: "The order of equal elements is not guaranteed to be preserved."
Maybe I'm just getting worked up over nothing, but may there be a situation when second word of one of these params will be moved before first word or even before a word of another parameter?

MrSidims: > From description of std::sort: "The order of equal elements is not guaranteed to be preserved."

Unless a programmer wrote a special rule how to proceed with the sort in a lambda, and here we have:
`return CompareLeft.first < CompareRight.first`

lib/SPIRV/SPIRVReader.cpp
pauzinl: The spec says that "A negative literal operand value specifies that the respective loop iteration bound or expectation is not defined"

mlychkov: Yes, and it means that we add a metadata value only when LoopCountXXX is non-negative or >= 0.

lib/SPIRV/SPIRVWriter.cpp
mlychkov: I meant that some sort algorithms are unstable and may change order of equal elements even for strict less comparison.

mlychkov: Oh, I found that there is also present std::stable_sort. Maybe to use this sort to be sure that there won't be unexpected reshuffles of LoopCountXXX words?

test/transcoding/SPV_INTEL_fpga_loop_controls/FPGALoopAttr.ll
pauzinl: Added comment in this commit: [a7a3741](https://github.com/KhronosGroup/SPIRV-LLVM-Translator/pull/1062/commits/a7a3741d55a23ee5bb5298c087fc5eaf778edda5)

lib/SPIRV/SPIRVReader.cpp
pauzinl: Changed it in this commit: [a7a3741](https://github.com/KhronosGroup/SPIRV-LLVM-Translator/pull/1062/commits/a7a3741d55a23ee5bb5298c087fc5eaf778edda5)

lib/SPIRV/SPIRVReader.cpp
pauzinl: Changed it in this commit: [a7a3741](https://github.com/KhronosGroup/SPIRV-LLVM-Translator/pull/1062/commits/a7a3741d55a23ee5bb5298c087fc5eaf778edda5)

lib/SPIRV/SPIRVReader.cpp
pauzinl: Added in this commit: [a7a3741](https://github.com/KhronosGroup/SPIRV-LLVM-Translator/pull/1062/commits/a7a3741d55a23ee5bb5298c087fc5eaf778edda5)

lib/SPIRV/SPIRVReader.cpp
```diff
@@ -987,6 +995,48 @@ void SPIRVToLLVM::setLLVMLoopMetadata(const LoopInstType *LM,
...
+    LoopCountMin |= static_cast<uint64_t>(LoopControlParameters[NumParam++])
+                    << 32;
+    if (static_cast<int64_t>(LoopCountMin) >= 0) {
+      Metadata.push_back(llvm::MDNode::get(
+          *Context,
+          getMetadataFromNameAndParameter("llvm.loop.intel.loopcount_min",
+                                          static_cast<int64_t>(LoopCountMin))));
+      assert(NumParam <= LoopControlParameters.size() &&
```
mlychkov, line 1010: Firstly, we increase NumParam twice for each LoopCount parameter.
Secondly, assert should be checked outside of this if. Otherwise, we will skip this check if LoopCountXXX is negative.
I think that it would be enough to check that `NumParam <= LoopControlParameters.size() + 6` before reading all 3 parameters.

lib/SPIRV/SPIRVReader.cpp
pauzinl: Added comment in [a2d546c](https://github.com/KhronosGroup/SPIRV-LLVM-Translator/pull/1062/commits/a2d546cfaa70d488cdcd44be1a1ef57c08f076ca)

lib/SPIRV/SPIRVReader.cpp
mlychkov: Sorry, of course, NumParam + 6 <= LoopControlParameters.size().

lib/SPIRV/SPIRVReader.cpp
```diff
@@ -987,6 +995,48 @@ void SPIRVToLLVM::setLLVMLoopMetadata(const LoopInstType *LM,
...
+             "Missing loop control parameter!");
+    }
+
+    uint64_t LoopCountMax =
+        static_cast<uint64_t>(LoopControlParameters[NumParam++]);
+    LoopCountMax |= static_cast<uint64_t>(LoopControlParameters[NumParam++])
+                    << 32;
+    if (static_cast<int64_t>(LoopCountMax) >= 0) {
+      Metadata.push_back(llvm::MDNode::get(
```
AlexeySachkov, line 1019: You can update new overload of `getMetadataFromNameAndParameter` added by you to directly return `MDNode *` to simplify code here

lib/SPIRV/SPIRVReader.cpp
AlexeySachkov: During an offline discussion I was pointed out that this function is not always called and in order to determine whether it should be called or not, concatenation of two SPIR-V words should occur first.

Therefore, my second part of the comment should be ignored:

> Instead, I would suggest to accept two SPIR-V words here and automatically bundle them into a single 64-bit value within the function. Note: if we go this way, we should probably leave a comment near this function that it produces a metadata with a single value and not with two values

The first one is still stands

lib/SPIRV/SPIRVWriter.cpp
AlexeySachkov: > Maybe to use this sort to be sure that there won't be unexpected reshuffles of LoopCountXXX words?

Do we actually care about those re-orders?

lib/SPIRV/SPIRVReader.cpp
pauzinl: Updated in [1fbf78b](https://github.com/KhronosGroup/SPIRV-LLVM-Translator/pull/1062/commits/1fbf78bf40440051bb7dbc9459d4cb1070ea8eae)

lib/SPIRV/SPIRVReader.cpp
pauzinl: Changed assert in [1fbf78b](https://github.com/KhronosGroup/SPIRV-LLVM-Translator/pull/1062/commits/1fbf78bf40440051bb7dbc9459d4cb1070ea8eae)

lib/SPIRV/SPIRVWriter.cpp
pauzinl: Changed std::sort to std::stable_sort in [1fbf78b](https://github.com/KhronosGroup/SPIRV-LLVM-Translator/pull/1062/commits/1fbf78bf40440051bb7dbc9459d4cb1070ea8eae)

lib/SPIRV/SPIRVWriter.cpp
```diff
@@ -1111,6 +1111,12 @@ LLVMToSPIRVBase::getLoopControl(const BranchInst *Branch,
 
   size_t LoopControl = spv::LoopControlMaskNone;
   std::vector<std::pair<SPIRVWord, SPIRVWord>> ParametersToSort;
+  // We need to create these variables here because it might be the case when
+  // metadata is not defined for one or two of these parameters and in this
+  // case we need to store "-1" value
```
mlychkov, line 1116:
```suggestion
  // If only a subset of loop count parameters is defined in metadata
  // then undefined ones should have a default value -1 in SPIR-V.
  // Preset all loop count parameters with the default value.
```

pauzinl: Changed in [a12e1a0](https://github.com/KhronosGroup/SPIRV-LLVM-Translator/pull/1062/commits/a12e1a08a62ea0b5993037c27caf8aa39f72231b)

mlychkov (approved):
Thank you. No more comments from my side.

MrSidims (approved):

PR #1063, opened 2021-06-09
Backmerge SPIRV-IR related changes to llvm_release_120

aratajew:


aratajew:
@AlexeySachkov could we please merge it? Check code style fail seems to be just a warning.

PR #1067, opened 2021-06-09
[CI] Align clang-tidy/format versions to LLVM version

svenvh:
Using a different LLVM version to obtain the compile commands may
cause inclusion of compiler flags that aren't recognized by
the older version of clang-tidy.

### DIFF
--- a/.github/workflows/check-code-style.yml
+++ b/.github/workflows/check-code-style.yml
@@ -61,7 +61,9 @@ jobs:
         curl -L "https://apt.llvm.org/llvm-snapshot.gpg.key" | sudo apt-key add -
         echo "deb https://apt.llvm.org/bionic/ llvm-toolchain-bionic main" | sudo tee -a /etc/apt/sources.list
         sudo apt-get update
-        sudo apt-get install -yqq clang-format-9 clang-tidy-9 llvm-${{ env.LLVM_VERSION }}-dev libomp-${{ env.LLVM_VERSION }}-dev
+        sudo apt-get install -yqq \
+            clang-format-${{ env.LLVM_VERSION }} clang-tidy-${{ env.LLVM_VERSION }} \
+            llvm-${{ env.LLVM_VERSION }}-dev libomp-${{ env.LLVM_VERSION }}-dev
 
     - name: Generate compile command database
       if: ${{ steps.gather-list-of-changes.outputs.HAS_CHANGES }}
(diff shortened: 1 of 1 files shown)


.github/workflows/check-code-style.yml
```diff
@@ -61,7 +61,9 @@ jobs:
         curl -L "https://apt.llvm.org/llvm-snapshot.gpg.key" | sudo apt-key add -
         echo "deb https://apt.llvm.org/bionic/ llvm-toolchain-bionic main" | sudo tee -a /etc/apt/sources.list
         sudo apt-get update
-        sudo apt-get install -yqq clang-format-9 clang-tidy-9 llvm-${{ env.LLVM_VERSION }}-dev libomp-${{ env.LLVM_VERSION }}-dev
```
AlexeySachkov, line 64: Actually, I've hardcoded `-9` version here on purpose, in order to avoid situations when we use the latest version of `clang-tidy` (which might not even be released yet) in our CI, while contributors could have something installed from a built-in APT repos, which will be older and produce different result.

However, I'm not sure how quickly `clang-format` evolves and whether it is even possible to find yourself in such situation with mismatched formatting. But anyway, if we decide to switch from 9 to something newer, I would prefer if we stick with already released versions of LLVM, i.e. `LLVM_VERSION` env variable defined in this workflow should lag behind other workflows by 1.

What do you think here, @svenvh?

svenvh: I see your point, and that sounds reasonable.  But getting that to work in practice might be challenging?

- Using clang-tidy/format from an earlier release (e.g. `< LLVM_VERSION`) may lead to incompatibilities, such as the issue with the `-fno-semantic-interposition` flag that we are seeing [now](https://github.com/KhronosGroup/SPIRV-LLVM-Translator/pull/1068/checks?check_run_id=2784654437)
- If we set `LLVM_VERSION` to something earlier for this job, to match the clang-tidy/format versions, then the cmake step to generate compile_command will fail as it won't be able to find a matching LLVM version.

The currently offending flag seems to be available from clang 11, so I can try switching this job from 9 to 11.  But this might break again any time in the future I suppose, so not sure if that's a stable solution?

AlexeySachkov: > But getting that to work in practice might be challenging?

Yeah, that is a concern for sure

> Using clang-tidy/format from an earlier release (e.g. < LLVM_VERSION) may lead to incompatibilities, such as the issue with the -fno-semantic-interposition flag that we are seeing now

My idea was to align `llvm-*` packages with `clang-tidy/format` packages

> If we set LLVM_VERSION to something earlier for this job, to match the clang-tidy/format versions, then the cmake step to generate compile_command will fail as it won't be able to find a matching LLVM version.

Theoretically, we can specify different version for `cmake` invocation, but I think that it will be overcomplicated.

> The currently offending flag seems to be available from clang 11, so I can try switching this job from 9 to 11. But this might break again any time in the future I suppose, so not sure if that's a stable solution?

We have been living with `9` for a half a year, at least? I tend to think that it would actually be just fine to use `11` until it is broken again.

> I see your point, and that sounds reasonable. 

At the same time, I don't have any actual data to support the importance of my concern, so probably we can proceed with your PR as-is and see if it causes any problems? If that would be the case, then we can try to settle on particular version of `clang-format`

AlexeySotkin (approved):

PR #1069, opened 2021-06-09
Fix attachment of decoration to spec constants

AlexeySotkin:
If a phi instruction uses result of __spirv_SpecConst() call which has
not been translated yet, we created a OpForward instruction. Then we
translated the call to OpSpecConstant and attached the SpecId decoration
to it. After that, when we map LLVM value to SPIR-V values we detected
that there is a OpForward instruction and replaced Id of OpSpecConstant
with Id of the OpForward instruction, leaving the decoration targeted to
non-existing instruction.

To avoid this incorrect behaviour we should attach decorations after
replacing OpForward instruction, i.e. in transDecoration method.

### DIFF
--- a/lib/SPIRV/SPIRVWriter.cpp
+++ b/lib/SPIRV/SPIRVWriter.cpp
@@ -1902,6 +1902,15 @@ bool LLVMToSPIRVBase::transDecoration(Value *V, SPIRVValue *BV) {
   }
   transMemAliasingINTELDecorations(V, BV);
 
+  if (auto *CI = dyn_cast<CallInst>(V)) {
+    auto OC = BV->getOpCode();
+    if (OC == OpSpecConstantTrue || OC == OpSpecConstantFalse ||
+        OC == OpSpecConstant) {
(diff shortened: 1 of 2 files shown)


lib/SPIRV/SPIRVWriter.cpp
```diff
@@ -1902,6 +1902,15 @@ bool LLVMToSPIRVBase::transDecoration(Value *V, SPIRVValue *BV) {
   }
   transMemAliasingINTELDecorations(V, BV);
 
+  if (auto *CI = dyn_cast_or_null<CallInst>(V)) {
```
AlexeySachkov, line 1905: Why do we expect `V` to be `nullptr` here?

AlexeySachkov (approved):

lib/SPIRV/SPIRVWriter.cpp
AlexeySotkin: Right, I have no reason to expect it to be `nullptr`. I'll fix it.

svenvh (approved):

PR #1070, opened 2021-06-10
Handle OpSpecConstantOp for Select

svenvh:
`OpSpecConstantOp` may be used outside of a basic block (for example
in a variable initializer).  Modify the `SPIRVSelect` constructor to
accept a `nullptr` basic block.

Use an IRBuilder when translating `OpSelect`, so that the resulting
select expression will be created as a constant expression (and
possibly constant-folded) where possible.

Update some tests as the SPIR-V to LLVM translation now folds selects
more aggressively due to the use of IRBuilder.

### DIFF
--- a/lib/SPIRV/libSPIRV/SPIRVInstruction.cpp
+++ b/lib/SPIRV/libSPIRV/SPIRVInstruction.cpp
@@ -244,8 +244,14 @@ SPIRVInstruction *createInstFromSpecConstantOp(SPIRVSpecConstantOp *Inst) {
   assert(isSpecConstantOpAllowedOp(OC) &&
          "Op code not allowed for OpSpecConstantOp");
   Ops.erase(Ops.begin(), Ops.begin() + 1);
-  return SPIRVInstTemplateBase::create(OC, Inst->getType(), Inst->getId(), Ops,
-                                       nullptr, Inst->getModule());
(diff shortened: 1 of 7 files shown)


StuartDBrady (approved):
Looks good to me.

lib/SPIRV/libSPIRV/SPIRVInstruction.cpp
```diff
@@ -244,8 +244,14 @@ SPIRVInstruction *createInstFromSpecConstantOp(SPIRVSpecConstantOp *Inst) {
   assert(isSpecConstantOpAllowedOp(OC) &&
          "Op code not allowed for OpSpecConstantOp");
   Ops.erase(Ops.begin(), Ops.begin() + 1);
-  return SPIRVInstTemplateBase::create(OC, Inst->getType(), Inst->getId(), Ops,
-                                       nullptr, Inst->getModule());
+  switch (OC) {
```
AlexeySotkin, line 247: I'd like to avoid one more opcode-based switch. But to make it work we should change `SPIRVSelect` to inherit it from `SPIRVInstTemplateBase`, which is probably out of scope for this PR. Other than that, LGTM.

AlexeySotkin (approved):

PR #1071, opened 2021-06-11
Fix handling of complex constant expressions

Fznamznon:
When some constant expression is used as operand of regular instruction
and operand of other constant expression, the current algorithm of
lowering could produce incorrect order of instructions because it is not
possible to add instruction operand to a constant expression. Consider
following pseudo code example:
```
call foo(constexpr2(constexpr1), constexpr1)

// After first loop iteration through operands of the call instruction:

A = constexpr2(constexpr1)
call foo(A, constexpr1)

// Ok, instruction A is now a user of constexpr1, so, when second
// operand is processed, all uses of constexpr1 are updated, but the
// instruction that represents constexpr1 is inserted before call
// instruction because it is now being processed, so it will look like
// this:

A = constexpr2(B)
B = constexpr1
call foo(A, B)

// So, instruction B needs to be moved before all its users to get a
// valid module:

B = constexpr1
A = constexpr2(B)
call foo(A, B)

```

### DIFF
--- a/lib/SPIRV/SPIRVLowerConstExpr.cpp
+++ b/lib/SPIRV/SPIRVLowerConstExpr.cpp
@@ -157,8 +157,12 @@ void SPIRVLowerConstExprBase::visit(Module *M) {
               Users.push_back(InstUser);
           }
         }
-        for (auto &User : Users)
+        for (auto &User : Users) {
+          if (ReplInst->getParent() == User->getParent())
+            if (User->comesBefore(ReplInst))
+              ReplInst->moveBefore(User);
           User->replaceUsesOfWith(CE, ReplInst);
+        }
         return ReplInst;
       };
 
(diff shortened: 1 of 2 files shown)


AGindinson (commented):
> // So, instruction B needs to be moved after all its users to get a
> // valid module:

If I am interpreting the commit message, it should read "before all its [B's] users" instead.

Fznamznon:
> > // So, instruction B needs to be moved after all its users to get a
> > // valid module:
> 
> If I am interpreting the commit message, it should read "before all its [B's] users" instead.

Right. My bad.

AlexeySotkin (approved):

svenvh:
LGTM, but could you rebase this change on latest `master` and rerun the test? This to ensure that the test still passes after translation of `select` was changed to use IRBuilder (which may fold expressions) in #1070 .

Fznamznon:
> LGTM, but could you rebase this change on latest `master` and rerun the test? This to ensure that the test still passes after translation of `select` was changed to use IRBuilder (which may fold expressions) in #1070 .

Sure, done.

svenvh (approved):

Issue #1072, opened 2021-06-11
Avoid one more opcode-based switch


AlexeySotkin:
I'd like to avoid one more opcode-based switch. But to make it work we should change `SPIRVSelect` to inherit it from `SPIRVInstTemplateBase`, which is probably out of scope for this PR. Other than that, LGTM.

_Originally posted by @AlexeySotkin in https://github.com/KhronosGroup/SPIRV-LLVM-Translator/pull/1070#discussion_r649882432_

svenvh:
I'll have a look at this.

PR #1073 adds some more cases to the opcode switch, so we also need to change the inheritance for those classes.  I don't expect we need to modify any more classes, as it's the last PR in my series to improve `OpSpecConstantOp` handling with integer operators.

PR #1073, opened 2021-06-11
Handle OpSpecConstantOp with VectorShuffle, CompositeExtract, and CompositeInsert

svenvh:
This is (hopefully) the last PR in a series to improve `OpSpecConstantOp`
handling with integer operators.

Modify the constructors to allow passing a `nullptr` basic block (as is
the case for variable initializers).

Modify the constructors to take `SPIRVId`s instead of `SPIRVValue`s,
which better reflects what is stored in the class, and saves us an
unnecessary ID-to-Value-to-ID round trip in
`createInstFromSpecConstantOp`.

Use an IRBuilder when translating, so that the resulting
expression will be created as a constant expression (and
possibly constant-folded) where possible.

Modify `test/opundef.spt` to return the constructed value, so that it
does not get optimized out.

### DIFF
--- a/lib/SPIRV/libSPIRV/SPIRVInstruction.cpp
+++ b/lib/SPIRV/libSPIRV/SPIRVInstruction.cpp
@@ -253,6 +253,23 @@ SPIRVInstruction *createInstFromSpecConstantOp(SPIRVSpecConstantOp *Inst) {
     return new SPIRVVectorShuffle(Inst->getId(), Inst->getType(), Ops[0],
                                   Ops[1], Comp, nullptr, Inst->getModule());
   }
+  case OpCompositeExtract: {
+    std::vector<SPIRVWord> Indices;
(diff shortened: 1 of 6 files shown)


lib/SPIRV/libSPIRV/SPIRVInstruction.cpp
```diff
@@ -245,6 +245,31 @@ SPIRVInstruction *createInstFromSpecConstantOp(SPIRVSpecConstantOp *Inst) {
          "Op code not allowed for OpSpecConstantOp");
   Ops.erase(Ops.begin(), Ops.begin() + 1);
   switch (OC) {
+  case OpVectorShuffle: {
```
svenvh, line 248: As part of #1072 these classes should be changed to inherit from `SPIRVInstTemplateBase`.

StuartDBrady (approved):
Looks good to me.

The clang-tidy failure is an existing issue.

PR #1074, opened 2021-06-11
Add mapping for DWARF language codes to SourceLanguage

jzzheng22:
When translating from LLVM IR to SPIR-V, the SourceLanguage is currently 
set to the DWARF language value, which is wrong. This PR adds a mapping 
from DWARF language codes to the correct value in the SourceLanguage enum.

### DIFF
--- a/lib/SPIRV/LLVMToSPIRVDbgTran.cpp
+++ b/lib/SPIRV/LLVMToSPIRVDbgTran.cpp
@@ -490,7 +490,10 @@ LLVMToSPIRVDbgTran::transDbgCompilationUnit(const DICompileUnit *CU) {
   Ops[SPIRVDebugInfoVersionIdx] = SPIRVDebug::DebugInfoVersion;
   Ops[DWARFVersionIdx] = M->getDwarfVersion();
   Ops[SourceIdx] = getSource(CU)->getId();
-  Ops[LanguageIdx] = CU->getSourceLanguage();
+  Ops[LanguageIdx] = spv::SourceLanguage::SourceLanguageUnknown;
+  auto DwarfLang =
+      static_cast<llvm::dwarf::SourceLanguage>(CU->getSourceLanguage());
+  Ops[LanguageIdx] = convertDWARFSourceLangToSPIRV(DwarfLang);
   BM->addModuleProcessed(SPIRVDebug::ProducerPrefix + CU->getProducer().str());
   // Cache CU in a member.
   SPIRVCU = static_cast<SPIRVExtInst *>(
--- a/lib/SPIRV/LLVMToSPIRVDbgTran.cpp
+++ b/lib/SPIRV/LLVMToSPIRVDbgTran.cpp
@@ -490,7 +490,6 @@ LLVMToSPIRVDbgTran::transDbgCompilationUnit(const DICompileUnit *CU) {
   Ops[SPIRVDebugInfoVersionIdx] = SPIRVDebug::DebugInfoVersion;
   Ops[DWARFVersionIdx] = M->getDwarfVersion();
   Ops[SourceIdx] = getSource(CU)->getId();
-  Ops[LanguageIdx] = spv::SourceLanguage::SourceLanguageUnknown;
   auto DwarfLang =
       static_cast<llvm::dwarf::SourceLanguage>(CU->getSourceLanguage());
   Ops[LanguageIdx] = convertDWARFSourceLangToSPIRV(DwarfLang);
--- a/lib/SPIRV/libSPIRV/SPIRV.debug.h
+++ b/lib/SPIRV/libSPIRV/SPIRV.debug.h
@@ -1,6 +1,7 @@
 #ifndef SPIRV_DEBUG_H
 #define SPIRV_DEBUG_H
 #include "SPIRVUtil.h"
+#include "spirv.hpp"
 #include "llvm/BinaryFormat/Dwarf.h"
 
 namespace SPIRVDebug {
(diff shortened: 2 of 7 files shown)


lib/SPIRV/libSPIRV/spirv.hpp
```diff
@@ -65,6 +65,7 @@ enum SourceLanguage {
     SourceLanguageOpenCL_C = 3,
     SourceLanguageOpenCL_CPP = 4,
     SourceLanguageHLSL = 5,
+    SourceLanguageCPP_for_OpenCL = 6,
```
AlexeySotkin, line 68: There is no such value in the spec: https://www.khronos.org/registry/spir-v/specs/unified1/SPIRV.html#_a_id_source_language_a_source_language

What is the difference between `SourceLanguageCPP_for_OpenCL` and `SourceLanguageOpenCL_CPP` ?

lib/SPIRV/libSPIRV/SPIRV.debug.h
```diff
@@ -776,6 +777,15 @@ enum {
 using namespace llvm;
 
 namespace SPIRV {
+typedef SPIRVMap<dwarf::SourceLanguage, spv::SourceLanguage> SPIRVDWARFSourceLanguageMap;
+template <>
+inline void SPIRVDWARFSourceLanguageMap::init() {
```
AlexeySotkin, line 782: This is a bidirectional map and it can't have duplicated values.

lib/SPIRV/libSPIRV/spirv.hpp
svenvh: I suppose the value will be added to the spec soon, as https://github.com/KhronosGroup/SPIRV-Headers/pull/217 was recently merged into the SPIR-V headers.

> What is the difference between SourceLanguageCPP_for_OpenCL and SourceLanguageOpenCL_CPP ?

C++ for OpenCL and OpenCL C++ are different languages.  Quoting from https://github.com/KhronosGroup/OpenCL-Guide/blob/main/chapters/cpp_for_opencl.md :
> The OpenCL working group has transitioned from the original OpenCL C++ kernel language first defined in OpenCL 2.2 to the community developed C++ for OpenCL kernel language that provides improved features and compatibility with OpenCL C.

lib/SPIRV/libSPIRV/SPIRV.debug.h
jzzheng22: I've changed the mapping to use two switch statements instead.

lib/SPIRV/libSPIRV/spirv.hpp
StuartDBrady: Yes, the value is being added retrospectively to all versions of SPIR-V.  When SPIR-V Tools is built with the current SPIR-V Headers, use of the CPP_for_OpenCL language ID will pass validation, and will correctly assemble and disassemble.

It was agreed that it is appropriate to add source language IDs retrospectively.  As the language ID has no semantic content, the use of unknown languages must not cause SPIR-V to fail to be consumed in an implementation.  The SPIR-V Tools will be changed to allow disassembly and assembly of modules using unknown language IDs, using a numeric form for the ID, and the validation will be relaxed to permit unknown language IDs.  For the time being, the overly strict behavior of SPIR-V Tools should not affect those who are translating from code written in e.g. OpenCL C.

lib/SPIRV/LLVMToSPIRVDbgTran.cpp
```diff
@@ -490,7 +490,10 @@ LLVMToSPIRVDbgTran::transDbgCompilationUnit(const DICompileUnit *CU) {
   Ops[SPIRVDebugInfoVersionIdx] = SPIRVDebug::DebugInfoVersion;
   Ops[DWARFVersionIdx] = M->getDwarfVersion();
   Ops[SourceIdx] = getSource(CU)->getId();
-  Ops[LanguageIdx] = CU->getSourceLanguage();
+  Ops[LanguageIdx] = spv::SourceLanguage::SourceLanguageUnknown;
```
StuartDBrady, line 493: Ops[LanguageIdx] is assigned to twice. Shouldn't the first assignment to it be removed, now?

jzzheng22: Thanks for catching that! Fixed

test/FortranArray.ll
```diff
@@ -5,6 +5,11 @@
 ; RUN: llvm-spirv -r %t.spv -o %t.rev.bc
 ; RUN: llvm-dis %t.rev.bc -o - | FileCheck %s --check-prefix=CHECK-LLVM
 
+; XFAIL: *
```
svenvh, line 8: @MochalovaAn since you added this test in #938, maybe you have a better idea how to handle this.

This PR reveals a problem with this test, as the Fortran language cannot be represented as a SPIR-V source language, and the language affects the LLVM verification rules as explained in the comment below.  Do you have any suggestions what to do here?

AlexeySachkov: Tagging @AlexeySotkin  here to help answer the question

AlexeySachkov: From what I understood myself, it seems like we have encountered a case where we had a debug info for an array type with `count` missing (or being `nullptr`) - the PR allows to translate this missing count to debug id `none` and then translates it back to a subrange `1+`.

I assume that the test explicitly sets the Fortran language because we haven't encountered this problem with other languages. I'm not that familiar with debug info in LLVM IR enough to say if we can just replace the language field there to something else than Fortran.

svenvh: I've raised #1098 for a discussion on the wider issue of representing source languages that are not in SPIR-V's Source Language enum.

StuartDBrady (approved):
This looks good to me.

PR #1078, opened 2021-06-16
Fix SPIRV-IR for group opcodes

aratajew:


aratajew:
@Fznamznon  @AlexeySotkin please, take a look.

Fznamznon (approved):

lib/SPIRV/SPIRVUtil.cpp
```diff
@@ -1665,6 +1665,69 @@ class SPIRVFriendlyIRMangleInfo : public BuiltinFuncMangleInfo {
       addUnsignedArg(0);
       addUnsignedArg(3);
       break;
+    case OpGroupUMax:
+      LLVM_FALLTHROUGH;
```
svenvh, line 1669: You shouldn't need those `LLVM_FALLTHROUGH` lines if the `case`s are immediately following each other (which is the case for your patch).

aratajew: Just wanted to keep it consistent with the other part of this function. But you're right, usage of `LLVM_FALLTHROUGH` is not required for empty switch cases. I think we can remove all `LLVM_FALLTHROUGH`'s in this function.

svenvh: Thanks for cleaning up the `LLVM_FALLTHROUGH`s!

svenvh (approved):

aratajew:
@svenvh could you please merge this change?

PR #1082, opened 2021-06-17
Fix SYCL/OpenCL vector size in DebugInfo reverse translation

AGindinson:
The SYCL/OpenCL specifications require that 3-element vectors
are sized equally to the 4-element ones (see 4.10.2.1 and 4.10.2.6,
[khronos.org/registry/SYCL/specs/sycl-1.2.1.pdf](https://www.khronos.org/registry/SYCL/specs/sycl-1.2.1.pdf)). Meanwhile, the
logic for translating SPIR-V debug info into LLVM IR obeys a
trivial `vec_size = num_elems * elem_size` formula, which fails
for the 3-element edge case.

Example LLVM IR pre-translation:
```
!0 = !DICompositeType(tag: DW_TAG_array_type, baseType: !3, size: 128, flags: DIFlagVector, elements: !1)
!1 = !{!2}
!2 = !DISubrange(count: 3)
!3 = !DIBasicType(name: "int", size: 32, align: 32, encoding: DW_ATE_signed)
```
Faulty LLVM IR after bi-directional translation (note the size in `!0`):
```
!0 = !DICompositeType(tag: DW_TAG_array_type, baseType: !3, size: 96, flags: DIFlagVector, elements: !1)
!1 = !{!2}
!2 = !DISubrange(count: 3)
!3 = !DIBasicType(name: "int", size: 32, align: 32, encoding: DW_ATE_signed)
```
Until SPIR-V DI instructions are re-designed to store the array
size information, handle the 3-element case explicitly to favor
OpenCL/SYCL requirements.

### DIFF
--- a/lib/SPIRV/SPIRVToLLVMDbgTran.cpp
+++ b/lib/SPIRV/SPIRVToLLVMDbgTran.cpp
@@ -222,7 +222,15 @@ SPIRVToLLVMDbgTran::transTypeVector(const SPIRVExtInst *DebugInst) {
   DIType *BaseTy =
       transDebugInst<DIType>(BM->get<SPIRVExtInst>(Ops[BaseTypeIdx]));
   SPIRVWord Count = Ops[ComponentCountIdx];
-  uint64_t Size = getDerivedSizeInBits(BaseTy) * Count;
+  // FIXME: The current design of SPIR-V Debug Info doesn't provide a field
+  // for the derived memory size. Meanwhile, OpenCL/SYCL 3-element vectors
+  // occupy the same amount of memory as 4-element vectors, hence the simple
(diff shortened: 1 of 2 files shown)


AGindinson:
@AlexeySotkin, could you please take a look?

AlexeySotkin (approved):

AGindinson:
@svenvh, could you please take a look as well? Just in case there are any concerns with such a temporary approach, e.g. for non-OpenCL programming models.

svenvh (approved):
> @svenvh, could you please take a look as well? Just in case there are any concerns with such a temporary approach, e.g. for non-OpenCL programming models.

I cannot really answer for non-OpenCL programming models, but the change in itself looks good to me.

PR #1083, opened 2021-06-17
Fix OpModuleProcessed memory leak

svenvh:
Fix a small memory leak after #924; @KornevNikita could you please review this fix?


KornevNikita:
LGTM 👍🏻

PR #1085, opened 2021-06-22
Improve handling of memmove when Lenght > size of the type to move

MrSidims:
It is possible that for a single element Src type size doesn't
match memory size to move (for example we copy N elements of one
integer array to an address of a first element of another integer
array).

### DIFF
--- a/lib/SPIRV/SPIRVLowerMemmove.cpp
+++ b/lib/SPIRV/SPIRVLowerMemmove.cpp
@@ -63,34 +63,31 @@ public:
     if (isa<PHINode>(Src))
       report_fatal_error("llvm.memmove of PHI instruction result not supported",
                          false);
-    auto *SrcTy = Src->getType();
-    auto *Length = cast<ConstantInt>(I.getLength());
-    auto *S = Src;
     // The source could be bit-cast or addrspacecast from another type,
     // need the original type for the allocation of the temporary variable
-    while (isa<BitCastInst>(S) || isa<AddrSpaceCastInst>(S))
-      S = cast<CastInst>(S)->getOperand(0);
-    SrcTy = S->getType();
+    auto *SrcTy = Src->stripPointerCasts()->getType();
+    auto *Length = cast<ConstantInt>(I.getLength());
     MaybeAlign Align = I.getSourceAlign();
     auto Volatile = I.isVolatile();
     Value *NumElements = nullptr;
     uint64_t ElementsCount = 1;
     if (SrcTy->isArrayTy()) {
-      NumElements = Builder.getInt32(SrcTy->getArrayNumElements());
       ElementsCount = SrcTy->getArrayNumElements();
+      NumElements = Builder.getInt32(ElementsCount);
     }
-    if (((ElementsCount > 1) && (Mod->getDataLayout().getTypeSizeInBits(
-                                     SrcTy->getPointerElementType()) *
-                                     ElementsCount !=
-                                 Length->getZExtValue() * 8)) ||
-        ((ElementsCount == 1) &&
-         (Mod->getDataLayout().getTypeSizeInBits(
-              SrcTy->getPointerElementType()) < Length->getZExtValue() * 8)))
-      report_fatal_error("Size of the memcpy should match the allocated memory",
-                         false);
-
-    auto *Alloca =
-        Builder.CreateAlloca(SrcTy->getPointerElementType(), NumElements);
+    // Get number of bits to move and allocate memory appropriately:
+    // if lenght is bigger than a pointer base type size, then create an
+    // alloca of an array type with the same base type.
+    const uint64_t LenBits = Length->getZExtValue();
+    const uint64_t LayoutTypeBites =
+        Mod->getDataLayout().getTypeSizeInBits(SrcTy->getPointerElementType()) *
+        ElementsCount;
+    auto *AllocaTy = SrcTy->getPointerElementType();
+    if (LenBits > LayoutTypeBites) {
+      const uint64_t ArraySize = LenBits / LayoutTypeBites;
+      AllocaTy = ArrayType::get(SrcTy->getPointerElementType(), ArraySize);
+    }
+    auto *Alloca = Builder.CreateAlloca(AllocaTy, NumElements);
     if (Align.hasValue()) {
       Alloca->setAlignment(Align.getValue());
     }
--- a/test/llvm-intrinsics/memmove.ll
+++ b/test/llvm-intrinsics/memmove.ll
@@ -7,6 +7,9 @@
 
 ; CHECK-SPIRV-NOT: llvm.memmove
 
+; CHECK-SPIRV: TypeInt [[#TYPEINT:]] 32
+; CHECK-SPIRV: Constant [[#TYPEINT]] [[#ARRSIZE:]] 128
+
 ; CHECK-SPIRV: Variable {{[0-9]+}} [[mem:[0-9]+]] 7
 ; CHECK-SPIRV: Bitcast [[i8Ty:[0-9]+]] [[tmp0:[0-9]+]] [[mem]]
 ; CHECK-SPIRV: LifetimeStart [[tmp0]] [[size:[0-9]+]]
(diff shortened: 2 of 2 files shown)


MrSidims:
@Fznamznon @AlexeySotkin please take a look

lib/SPIRV/SPIRVLowerMemmove.cpp
```diff
@@ -79,16 +79,22 @@ class SPIRVLowerMemmoveBase {
...
-                                     ElementsCount !=
-                                 Length->getZExtValue() * 8)) ||
-        ((ElementsCount == 1) &&
-         (Mod->getDataLayout().getTypeSizeInBits(
-              SrcTy->getPointerElementType()) < Length->getZExtValue() * 8)))
-      report_fatal_error("Size of the memcpy should match the allocated memory",
-                         false);
-
+    {
+      // Check whether Src layout size matches a memory size parameter of
```
AlexeySotkin, line 83:
```suggestion
      // Check whether Src layout size matches the length parameter of
```

Fznamznon (commented):
Can we add a test for this?

MrSidims:
> Can we add a test for this?

Done

MrSidims:
Ping @AlexeySotkin

test/llvm-intrinsics/memmove.ll
```diff
@@ -81,6 +95,12 @@ define spir_func void @copy_struct(%struct.SomeStruct addrspace(1)* nocapture re
   ret void
 }
 
+; Function Attrs: nounwind
+define spir_kernel void @test_array(i8 addrspace(1)* %in, i8 addrspace(1)* %out) {
```
AlexeySotkin, line 99: Here is what we get with your fix after SPIRVLowerMemmove pass:
```
define spir_kernel void @test_array(i8 addrspace(1)* %in, i8 addrspace(1)* %out) {
  %1 = alloca i8, align 1
  call void @llvm.lifetime.start.p0i8(i64 -1, i8* %1)
  call void @llvm.memcpy.p0i8.p1i8.i32(i8* %1, i8 addrspace(1)* %out, i32 128, i1 false)
  call void @llvm.memcpy.p1i8.p0i8.i32(i8 addrspace(1)* %in, i8* %1, i32 128, i1 false)
  call void @llvm.lifetime.end.p0i8(i64 -1, i8* %1)
  ret void
}
```
We allocate 8 bits and copy 128*32(=4096) bit there. I believe we should see a segmentation fault (or something like this ) in this case.

AlexeySotkin (changes requested):
Please address my comment

test/llvm-intrinsics/memmove.ll
MrSidims: I see. I believe we need to have:
`%1 = alloca i8*`
instead: https://godbolt.org/z/EYY3ezacf

MrSidims:
> Please address my comment

Done.

lib/SPIRV/SPIRVLowerMemmove.cpp
```diff
@@ -76,21 +76,30 @@ class SPIRVLowerMemmoveBase {
...
-      report_fatal_error("Size of the memcpy should match the allocated memory",
+    // Check whether Src layout size matches the length parameter of
+    // llvm.memmove intrinsic. It is possible that for a single element Src
+    // type size doesn't match memory size to move (for example we copy N
+    // elements of one integer array to an address of a first element of
+    // another integer array). So we ignore this check here and handle it
+    // by setting alloca type instruction.
+    const uint64_t LenBits = Length->getZExtValue() * 8;
+    const uint64_t LayoutTypeBits =
+        Mod->getDataLayout().getTypeSizeInBits(SrcTy->getPointerElementType()) *
```
AlexeySotkin, line 90: `SrcTy` is not necessarily the original type used in `llvm.memmove`, because we went through casts at lines 71-72.

MrSidims: Yes and I believe it's done by design, for example we can call:
call void @llvm.memmove.p1i8.p1i8.i32(i8 * %dsr, i8 * %src, i32 4, i1 false)
where src and dst pointer are of i8 type whilst the real value is of i32 type (it was just bitcasted away). In this case we shall allocate not a vector [i8 x 4] but a single i32 chunk of memory for a temporary variable.

lib/SPIRV/SPIRVLowerMemmove.cpp
```diff
@@ -76,21 +76,30 @@ class SPIRVLowerMemmoveBase {
...
+    if ((ElementsCount > 1) && (LenBits != LayoutTypeBits))
+      report_fatal_error("Size of the memcpy should match the allocated "
+                         "memory",
                          false);
-
-    auto *Alloca =
-        Builder.CreateAlloca(SrcTy->getPointerElementType(), NumElements);
+    auto *AllocaTy = SrcTy->getPointerElementType();
+    // If we need more - allocate more
+    if (ElementsCount == 1 && LenBits > LayoutTypeBits) {
```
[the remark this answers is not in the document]
MrSidims: Yes, done.

MrSidims:
@mlychkov @AGindinson @KornevNikita  could you please take a look?

lib/SPIRV/SPIRVLowerMemmove.cpp
```diff
@@ -63,34 +63,31 @@ class SPIRVLowerMemmoveBase {
...
     MaybeAlign Align = I.getSourceAlign();
     auto Volatile = I.isVolatile();
     Value *NumElements = nullptr;
     uint64_t ElementsCount = 1;
     if (SrcTy->isArrayTy()) {
-      NumElements = Builder.getInt32(SrcTy->getArrayNumElements());
       ElementsCount = SrcTy->getArrayNumElements();
+      NumElements = Builder.getInt32(ElementsCount);
     }
-    if (((ElementsCount > 1) && (Mod->getDataLayout().getTypeSizeInBits(
```
MrSidims, line 82: The check was removed completely. Length parameter has no restrictions in LangRef.

test/llvm-intrinsics/memmove.ll
```diff
@@ -60,6 +70,18 @@
 ; CHECK-LLVM: [[tmp4:%[0-9]+]] = bitcast %struct.SomeStruct* [[local]] to [[type]]
 ; CHECK-LLVM: call void @llvm.lifetime.end.p0i8({{i[0-9]+}} {{-?[0-9]+}}, [[type]] [[tmp4]])
 
+; CHECK-LLVM-LABEL: @test_array
+; CHECK-LLVM: %[[#ALLOCA:]] = alloca [16 x i8], align 1
+; CHECK-LLVM: %[[#TMP0:]] = bitcast [16 x i8]* %[[#ALLOCA]] to i8*
+; CHECK-LLVM: call void @llvm.lifetime.start.p0i8(i64 -1, i8* %[[#TMP0]])
+; CHECK-LLVM: %[[#TMP1:]] = bitcast [16 x i8]* %1 to i8*
```
mlychkov, line 77: What variable is %1?

MrSidims: It's alloca, thanks!

MrSidims:
@mlychkov @AGindinson @KornevNikita friendly ping

KornevNikita:
👍🏻

mlychkov (approved):

test/llvm-intrinsics/memmove.ll
```diff
@@ -60,6 +73,18 @@
 ; CHECK-LLVM: [[tmp4:%[0-9]+]] = bitcast %struct.SomeStruct* [[local]] to [[type]]
 ; CHECK-LLVM: call void @llvm.lifetime.end.p0i8({{i[0-9]+}} {{-?[0-9]+}}, [[type]] [[tmp4]])
 
+; CHECK-LLVM-LABEL: @test_array
+; CHECK-LLVM: %[[#ALLOCA:]] = alloca [16 x i8], align 1
+; CHECK-LLVM: %[[#TMP0:]] = bitcast [16 x i8]* %[[#ALLOCA]] to i8*
+; CHECK-LLVM: call void @llvm.lifetime.start.p0i8(i64 -1, i8* %[[#TMP0]])
+; CHECK-LLVM: %[[#TMP1:]] = bitcast [16 x i8]* %[[#ALLOCA]] to i8*
+; CHECK-LLVM: call void @llvm.memcpy.p0i8.p1i8.i32(i8* %[[#TMP1]], i8 addrspace(1)* %out, i32 128, i1 false)
```
AlexeySotkin, line 81: Sorry, but I still don't understand how we can copy 128 bytes into array of 16 bytes.  @MrSidims, am I missing something?

PR #1086, opened 2021-06-22
[Backport] Replace type pointers by ids in SPIRVTypeFunction on llvm_release_120 branch

NikitaRudenkoIntel:
The type of a function parameter may be id of a forward declared pointer type.
When we decode such module, we can't construct SPIRVTypeFunction, because
actual SPIRVTypePointer has not been decoded yet. Instead we can keep ids of
paramter types which can be resolved later.


NikitaRudenkoIntel:
Closing outdated backport

PR #1091, opened 2021-06-23
Add support of new IR pattern to transOCLBuiltinFromVariable

Fznamznon:
Some LLVM optimizations started generation of ascast -> gep -> load
sequence, this patch adds support for it.

### DIFF
--- a/lib/SPIRV/SPIRVReader.cpp
+++ b/lib/SPIRV/SPIRVReader.cpp
@@ -290,6 +290,20 @@ Value *SPIRVToLLVM::mapFunction(SPIRVFunction *BF, Function *F) {
 // %5 = insertelement <3 x i64> %3, i64 %4, i32 2
 // %c = extractelement <3 x i64> %5, i32 idx
 // %d = extractelement <3 x i64> %5, i32 idx
+//
+// Replace the following pattern:
+// %0 = addrspacecast <3 x i64> addrspace(1)* @__spirv_BuiltInWorkgroupSize to
+// <3 x i64> addrspace(4)*
+// %1 = getelementptr <3 x i64>, <3 x i64> addrspace(4)* %0, i64 0, i64 0
+// %2 = load i64, i64 addrspace(4)* %1, align 32
+// With:
+// %0 = call spir_func i64 @_Z13get_global_idj(i32 0) #1
+// %1 = insertelement <3 x i64> undef, i64 %0, i32 0
+// %2 = call spir_func i64 @_Z13get_global_idj(i32 1) #1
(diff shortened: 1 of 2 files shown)


Fznamznon:
@AlexeySotkin , @MrSidims , please take a look.

lib/SPIRV/SPIRVReader.cpp
```diff
@@ -363,10 +381,27 @@ bool SPIRVToLLVM::transOCLBuiltinFromVariable(GlobalVariable *GV,
...
+      Value *Ptr = LD->getPointerOperand();
+
+      if (isa<FixedVectorType>(Ptr->getType()->getPointerElementType())) {
+        LD->replaceAllUsesWith(Vectors.back());
+      } else {
+        auto *GEP = dyn_cast<GetElementPtrInst>(Ptr);
+        assert(GEP && "Unexpected pattern!");
+        assert(GEP->getNumIndices() == 2 && "Unexpected pattern!");
+        Value *Idx = (Value *)GEP->getOperand(2);
+        Value *Vec = (Value *)Vectors.back();
```
AlexeySotkin, line 394:
```suggestion
        auto *Idx = static_cast<Value *>(GEP->getOperand(2));
        auto *Vec = static_cast<Value *>(Vectors.back());
```

AlexeySotkin (approved):

lib/SPIRV/SPIRVReader.cpp
Fznamznon: Ah, it seems we don't.

AlexeySotkin (approved):

lib/SPIRV/SPIRVReader.cpp
```diff
@@ -300,7 +314,8 @@ bool SPIRVToLLVM::transOCLBuiltinFromVariable(GlobalVariable *GV,
   } else {
     FuncName = std::string(GV->getName());
   }
-  Type *ReturnTy = GV->getType()->getPointerElementType();
+  Type *GVTy = GV->getType()->getPointerElementType();
+  Type *ReturnTy = GVTy;
```
MrSidims, line 318: This split is done just to improve code readability, correct? I don't see any GVTy/ReturnTy modifications below.

Fznamznon: No, not only for this. ReturnTy is re-initialized later in the code. I needed to keep original type of the global variable.

MrSidims: Ah, it's line 325, I've missed it.

MrSidims (approved):

PR #1092, opened 2021-06-24
Fix TopologicalSort memory leak

pauzinl:
There was memory leak in function `TopologicalSort::visit`. This patch fixes it.
@AlexeySotkin, @svenvh could you please review this fix?

### DIFF
--- a/lib/SPIRV/libSPIRV/SPIRVModule.cpp
+++ b/lib/SPIRV/libSPIRV/SPIRVModule.cpp
@@ -1774,7 +1774,7 @@ public:
   }
 
   ~TopologicalSort() {
-    for (auto F : ForwardPointerSet)
+    for (auto *F : ForwardPointerSet)
       delete F;
   }
 };
--- a/lib/SPIRV/libSPIRV/SPIRVModule.cpp
+++ b/lib/SPIRV/libSPIRV/SPIRVModule.cpp
@@ -1715,8 +1715,9 @@ class TopologicalSort {
           // cyclic dependency by inserting a forward declaration of that
           // pointer.
           SPIRVTypePointer *Ptr = static_cast<SPIRVTypePointer *>(E);
-          ForwardPointerSet.insert(new SPIRVTypeForwardPointer(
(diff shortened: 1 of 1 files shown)


lib/SPIRV/libSPIRV/SPIRVModule.cpp
```diff
@@ -1772,6 +1772,11 @@ class TopologicalSort {
     ForwardPointerVec.insert(ForwardPointerVec.end(), ForwardPointerSet.begin(),
```
svenvh, line 1772: Thanks for looking at this leak!

I'm not sure if deleting in `~TopologicalSort` is conceptually the correct fix, because the elements may live longer than `TopologicalSort` it seems.  Perhaps we need to loop over the module's `ForwardPointerVec` instead, although looking at it I'm not sure why the module has that field?

I would like to see what @AlexeySotkin thinks about this.

AlexeySotkin: I agree with @svenvh. The problem here is that at line 1718 we allocate a `SPIRVTypeForwardPointer` object but we don't add it to the module. If we add it to the module, it should be deleted in the destructor of `SPIRVTypeForwardPointer` in the loop over `EntryNoId`.

@pauzinl  could you please try to modify line 1718 to something like:
```
SPIRVModule *BM = E->getModule();
ForwardPointerSet.insert(BM->add(new SPIRVTypeForwardPointer(
              BM, Ptr, Ptr->getPointerStorageClass())));
```

pauzinl: Changed it in [5e79b32](https://github.com/KhronosGroup/SPIRV-LLVM-Translator/pull/1092/commits/5e79b3209443f6ddeb049ee62197ceb62b81bc1a)

MrSidims (approved):

AlexeySotkin (approved):

svenvh (approved):
LGTM, thanks!

PR #1094, opened 2021-06-28
Add SPV_KHR_integer_dot_product support

svenvh:
Add support for mapping the SPV_KHR_integer_dot_product extension
operations to and from SPIR-V friendly IR.

The new instructions' required capabilities depend on their argument
types.  This does not work with the current LLVM-to-SPIR-V
translation that consists of the following steps:

 1. Create the instruction without operands.
 2. Add it to the module (which queries the instruction's required
    capabilities).
 3. Add operands.

Swap steps 2 and 3, such that the instruction is fully constructed
before querying its required capabilities.


svenvh:
Remaining clang-tidy failures are unrelated.

test/transcoding/SPV_KHR_integer_dot_product-nonsat.ll
```diff
@@ -0,0 +1,292 @@
+; RUN: llvm-as %s -o %t.bc
+; RUN: llvm-spirv %t.bc -spirv-text --spirv-ext=+SPV_KHR_integer_dot_product -o %t.txt
```
MrSidims, line 2: What will happen if the extension isn't enabled?

test/transcoding/SPV_KHR_integer_dot_product-nonsat.ll
```diff
@@ -0,0 +1,292 @@
...
+; CHECK-LLVM: call spir_func i16 @_Z23__spirv_UDotKHR_Rushortiii
+; CHECK-LLVM: call spir_func i32 @_Z21__spirv_UDotKHR_Ruintiii
+; CHECK-LLVM: call spir_func i64 @_Z22__spirv_UDotKHR_Rulongiii
+
+; CHECK-LLVM: call spir_func i8 @_Z23__spirv_SUDotKHR_Ruchariii
+; CHECK-LLVM: call spir_func i16 @_Z24__spirv_SUDotKHR_Rushortiii
+; CHECK-LLVM: call spir_func i32 @_Z22__spirv_SUDotKHR_Ruintiii
+; CHECK-LLVM: call spir_func i64 @_Z23__spirv_SUDotKHR_Rulongiii
+
+; CHECK-SPIRV: 6 SDotKHR [[#I8]] [[#]] [[#]] [[#]] 0
```
MrSidims, line 41: I suggest not to check number of words

test/transcoding/SPV_KHR_integer_dot_product-nonsat.ll
svenvh: The translator will hit an assert related to `isAllowedToUseExtension`, as the translator starts with an empty `ExtensionsStatusMap` if no `--spirv-ext` is given.  So it will refuse to produce SPIR-V in that case.

test/transcoding/SPV_KHR_integer_dot_product-nonsat.ll
svenvh: Any reason why not?  I thought it would be valuable to check the number of words, because these instructions take an optional "Packed Vector Format" operand; see http://htmlpreview.github.io/?https://github.com/KhronosGroup/SPIRV-Registry/blob/master/extensions/KHR/SPV_KHR_integer_dot_product.html

MrSidims: Agree

test/transcoding/SPV_KHR_integer_dot_product-nonsat.ll
mlychkov: May we use:
`; RUN: not --crash llvm-spirv %t.bc -o %t.spv | FileCheck %s --check-prefix=CHECK-ERROR`
to test this case?

svenvh: Done.  The translator currently just asserts as I mentioned, which is not very pretty.  I will try to improve that in a separate pull request, once this has landed.

MrSidims: Yeah, I believe it's a general problem with SPIR-V friendly LLVM IR. Usual pattern like:
 IF something_in_llvm_ir
  AND IF allowed_extension
   THEN generate_some_spirv
isn't applicable in this flow.

MrSidims (approved):
Please resolve LIT failures, other than that the patch is LGTM

test/transcoding/SPV_KHR_integer_dot_product-nonsat.ll
svenvh: The CI failures show that we cannot check for this in all possible builds at the moment: for a build without asserts, we cannot check for any error messages.

svenvh: Turning the assert into a proper error seemed easy actually, so I've done it as part of this PR.

MrSidims (approved):

lib/SPIRV/libSPIRV/SPIRVInstruction.h
```diff
@@ -2871,6 +2871,74 @@ class SPIRVExpectKHRInstBase : public SPIRVInstTemplateBase {
...
+  }
+
+  void validate() const override {
+    SPIRVInstruction::validate();
+    SPIRVId Vec1 = Ops[0];
+    SPIRVId Vec2 = Ops[1];
+    (void)Vec1;
+    (void)Vec2;
+
+    assert(getValueType(Vec1) == getValueType(Vec2) &&
```
mlychkov, line 2894: Module's error log may be used instead of assert:
getModule()->getErrorLog().checkError(getValueType(Vec1) == getValueType(Vec2), SPIRVEC_InvalidInstruction, "Input vectors must have the same type");
```suggestion
    getModule()->getErrorLog().checkError(getValueType(Vec1) == getValueType(Vec2), SPIRVEC_InvalidInstruction, OpCodeNameMap::map(OpCode) + "\nInput vectors must have the same type");
```
In this case llvm-spirv will be finished without crash on error and the check will work even in build w/o asserts.

svenvh: All SPIRVInstruction validate functions are relying on asserts, none of them is using the Module error log currently.  So I prefer not to break this pattern as part of this PR.  I agree there are many ways to improve those validate functions though.

> and the check will work even in build w/o asserts.

As I understand, these validate functions are mostly there to aid translator developers.  They do not perform full validation, as there's the separate spirv-val tool for that.  So having those checks in builds without asserts isn't that important I think.

mlychkov: Ok, I see, thank you for explanation.

mlychkov (approved):

PR #1100, opened 2021-07-05
Fix using of deprecated CreateInBounds

KornevNikita:
Enable using of CreateInBounds with element type due to deprecate
GetElementPtrInst::CreateInBounds without element type.

### DIFF
--- a/lib/SPIRV/SPIRVReader.cpp
+++ b/lib/SPIRV/SPIRVReader.cpp
@@ -1402,7 +1402,7 @@ bool SPIRVToLLVM::postProcessOCLBuiltinWithArrayArguments(
           auto Zero =
               ConstantInt::getNullValue(Type::getInt32Ty(T->getContext()));
           Value *Index[] = {Zero, Zero};
-          I = GetElementPtrInst::CreateInBounds(Alloca, Index, "", CI);
+          I = GetElementPtrInst::CreateInBounds(nullptr, Alloca, Index, "", CI);
         }
         return Name.str();
       },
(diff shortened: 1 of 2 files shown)


lib/SPIRV/SPIRVReader.cpp
```diff
@@ -1402,7 +1402,7 @@ bool SPIRVToLLVM::postProcessOCLBuiltinWithArrayArguments(
           auto Zero =
               ConstantInt::getNullValue(Type::getInt32Ty(T->getContext()));
           Value *Index[] = {Zero, Zero};
-          I = GetElementPtrInst::CreateInBounds(Alloca, Index, "", CI);
+          I = GetElementPtrInst::CreateInBounds(nullptr, Alloca, Index, "", CI);
```
svenvh, line 1405: We have the desired type already in `T`:
```suggestion
          I = GetElementPtrInst::CreateInBounds(T, Alloca, Index, "", CI);
```

lib/SPIRV/SPIRVUtil.cpp
```diff
@@ -1184,7 +1184,8 @@ Value *getScalarOrArrayConstantInt(Instruction *Pos, Type *T, unsigned Len,
     new StoreInst(CA, Alloca, Pos);
     auto Zero = ConstantInt::getNullValue(Type::getInt32Ty(T->getContext()));
     Value *Index[] = {Zero, Zero};
-    auto Ret = GetElementPtrInst::CreateInBounds(Alloca, Index, "", Pos);
+    auto *Ret =
+        GetElementPtrInst::CreateInBounds(nullptr, Alloca, Index, "", Pos);
```
svenvh, line 1188: Likewise:
```suggestion
        GetElementPtrInst::CreateInBounds(AT, Alloca, Index, "", Pos);
```

KornevNikita: Thanks!

PR #1105, opened 2021-07-07
Revert "Remove mapping of NoWrite function parameter attribute on qua…

pszymich:
…l metadata"

This reverts commit 9b2cc0c6b11ed43d4a57ea426cfdfb83bd078846.

This change is needed for Intel Compute stack with LLVM 11 conformance ([compute-runtime](https://github.com/intel/compute-runtime), [intel-graphics-compiler](https://github.com/intel/intel-graphics-compiler), [opencl-clang](https://github.com/intel/opencl-clang)), since it breaks an [unit test related to `readonly` function parameter](https://github.com/intel/compute-runtime/blob/master/opencl/test/unit_test/kernel/kernel_tests.cpp#L422) in [this kernel](https://github.com/intel/compute-runtime/blob/master/opencl/test/unit_test/test_files/simple_kernels.cl#L22).

### DIFF
--- a/lib/SPIRV/SPIRVReader.cpp
+++ b/lib/SPIRV/SPIRVReader.cpp
@@ -104,6 +104,7 @@ static bool DbgSaveTmpLLVM = false;
 static const char *DbgTmpLLVMFileName = "_tmp_llvmbil.ll";
 
 namespace kOCLTypeQualifierName {
+const static char *Const = "const";
 const static char *Volatile = "volatile";
 const static char *Restrict = "restrict";
 const static char *Pipe = "pipe";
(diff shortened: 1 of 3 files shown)


MrSidims:
We'd better not to revert the patch. Previously kernal_arg_type_qual metadata was incorrectly mapped on NoWrite SPIR-V function parameter attribute and hence it was a bug. Better is to finish https://github.com/KhronosGroup/SPIRV-LLVM-Translator/pull/996 . For that https://github.com/KhronosGroup/SPIRV-LLVM-Translator/pull/996#pullrequestreview-637697168 needs to be addresses aka a translation flag should be introduced.

MrSidims:
I'm OK to revert the patch on llvm_11 branch, but need an approve from @svenvh and @StuartDBrady .

StuartDBrady:
As far as I understand, the metadata in question is not needed for SPIR-V that is to be consumed with `clCreateProgramWithIL()`, or to be compiled into a binary and then consumed with `clCreateProgramWithBinary()`.  The change in question is needed for correct behavior of the translator when used as part of the compilation flow of developers compiling OpenCL C source into SPIR-V.  The OpenCL CTS _should_ test this, but I believe it does not, currently.

Please _don't_ revert this change.  Instead, please apply the workaround that was discussed when an extra translation flag is passed, and please use this flag, assuming the workaround is sufficient.

Otherwise, a translation flag could be added to restore the previous behavior of the translator.

Either way, the default behavior seen by developers running `llvm-spirv` from prebuilt packages in e.g. Linux distributions should _not_ be nonstandard behavior needed specifically to get the Intel Compute stack working.

MrSidims:
Close in favor of https://github.com/KhronosGroup/SPIRV-LLVM-Translator/pull/1109

PR #1109, opened 2021-07-09
Add W/A to translate OpenCL kernel_arg_type_qual metadata

MrSidims:
Since constant qualifier stored in the metadata is no longer mapped
on NoWrite function parameter attribute we need a another way to
save information about it. As work around this patch stores argument
qualifier information in OpString under
'preserve-ocl-kernel-arg-type-metadata-through-string' option. This
option should be also passed during reversed translation to obtain
the metadata.

This patch also moves already existing workaround for kernel_arg_type
metadata under the option.

### DIFF
--- a/lib/SPIRV/SPIRVReader.cpp
+++ b/lib/SPIRV/SPIRVReader.cpp
@@ -3840,9 +3840,13 @@ bool SPIRVToLLVM::transNonTemporalMetadata(Instruction *I) {
 // generated and 'false' otherwise.
 static bool transKernelArgTypeMedataFromString(LLVMContext *Ctx,
                                                SPIRVModule *BM,
-                                               Function *Kernel) {
-  std::string ArgTypePrefix = std::string(SPIR_MD_KERNEL_ARG_TYPE) + "." +
-                              Kernel->getName().str() + ".";
+                                               Function *Kernel,
+                                               std::string MDName) {
+  // Run W/A translation only if the appropriate option is passed
+  if (!BM->shouldPreserveOCLKernelArgTypeMetadataThroughString())
+    return false;
+  std::string ArgTypePrefix =
+      std::string(MDName) + "." + Kernel->getName().str() + ".";
   auto ArgTypeStrIt = std::find_if(
       BM->getStringVec().begin(), BM->getStringVec().end(),
       [=](SPIRVString *S) { return S->getStr().find(ArgTypePrefix) == 0; });
@@ -3874,7 +3878,7 @@ static bool transKernelArgTypeMedataFromString(LLVMContext *Ctx,
     }
   }
 
-  Kernel->setMetadata(SPIR_MD_KERNEL_ARG_TYPE, MDNode::get(*Ctx, TypeMDs));
+  Kernel->setMetadata(MDName, MDNode::get(*Ctx, TypeMDs));
   return true;
 }
 
@@ -3993,29 +3997,32 @@ bool SPIRVToLLVM::transOCLMetadata(SPIRVFunction *BF) {
                                  return MDString::get(*Context, Qual);
                                });
   // Generate metadata for kernel_arg_type
-  if (!transKernelArgTypeMedataFromString(Context, BM, F))
+  if (!transKernelArgTypeMedataFromString(Context, BM, F,
+                                          SPIR_MD_KERNEL_ARG_TYPE))
     addOCLKernelArgumentMetadata(Context, SPIR_MD_KERNEL_ARG_TYPE, BF, F,
                                  [=](SPIRVFunctionParameter *Arg) {
                                    return transOCLKernelArgTypeName(Arg);
                                  });
   // Generate metadata for kernel_arg_type_qual
-  addOCLKernelArgumentMetadata(
-      Context, SPIR_MD_KERNEL_ARG_TYPE_QUAL, BF, F,
-      [=](SPIRVFunctionParameter *Arg) {
-        std::string Qual;
-        if (Arg->hasDecorate(DecorationVolatile))
-          Qual = kOCLTypeQualifierName::Volatile;
-        Arg->foreachAttr([&](SPIRVFuncParamAttrKind Kind) {
-          Qual += Qual.empty() ? "" : " ";
-          if (Kind == FunctionParameterAttributeNoAlias)
-            Qual += kOCLTypeQualifierName::Restrict;
+  if (!transKernelArgTypeMedataFromString(Context, BM, F,
+                                          SPIR_MD_KERNEL_ARG_TYPE_QUAL))
+    addOCLKernelArgumentMetadata(
+        Context, SPIR_MD_KERNEL_ARG_TYPE_QUAL, BF, F,
+        [=](SPIRVFunctionParameter *Arg) {
+          std::string Qual;
+          if (Arg->hasDecorate(DecorationVolatile))
+            Qual = kOCLTypeQualifierName::Volatile;
+          Arg->foreachAttr([&](SPIRVFuncParamAttrKind Kind) {
+            Qual += Qual.empty() ? "" : " ";
+            if (Kind == FunctionParameterAttributeNoAlias)
+              Qual += kOCLTypeQualifierName::Restrict;
+          });
+          if (Arg->getType()->isTypePipe()) {
+            Qual += Qual.empty() ? "" : " ";
+            Qual += kOCLTypeQualifierName::Pipe;
+          }
+          return MDString::get(*Context, Qual);
         });
-        if (Arg->getType()->isTypePipe()) {
-          Qual += Qual.empty() ? "" : " ";
-          Qual += kOCLTypeQualifierName::Pipe;
-        }
-        return MDString::get(*Context, Qual);
-      });
   // Generate metadata for kernel_arg_base_type
   addOCLKernelArgumentMetadata(Context, SPIR_MD_KERNEL_ARG_BASE_TYPE, BF, F,
                                [=](SPIRVFunctionParameter *Arg) {
--- a/test/transcoding/KernelArgTypeInOpString.ll
+++ b/test/transcoding/KernelArgTypeInOpString.ll
@@ -21,21 +21,31 @@
 ; OpString "kernel_arg_type.%kernel_name%.typename0,typename1,..."
 
 ; RUN: llvm-as %s -o %t.bc
+; RUN: llvm-spirv -preserve-ocl-kernel-arg-type-metadata-through-string %t.bc -spirv-text -o %t.spv.txt
+; RUN: FileCheck < %t.spv.txt %s --check-prefix=CHECK-SPIRV-WORKAROUND
 ; RUN: llvm-spirv %t.bc -spirv-text -o %t.spv.txt
-; RUN: FileCheck < %t.spv.txt %s --check-prefix=CHECK-SPIRV
+; RUN: FileCheck < %t.spv.txt %s --check-prefix=CHECK-SPIRV-WORKAROUND-NEGATIVE
+; RUN: llvm-spirv -preserve-ocl-kernel-arg-type-metadata-through-string %t.bc -o %t.spv
+; RUN: spirv-val %t.spv
+; RUN: llvm-spirv -r -preserve-ocl-kernel-arg-type-metadata-through-string %t.spv -o %t.rev.bc
+; RUN: llvm-dis %t.rev.bc
+; RUN: FileCheck < %t.rev.ll %s --check-prefix=CHECK-LLVM-WORKAROUND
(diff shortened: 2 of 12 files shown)


MrSidims:
@StuartDBrady @svenvh please take a look

test/transcoding/KernelArgTypeInOpString.ll
```diff
@@ -21,21 +21,31 @@
 ; OpString "kernel_arg_type.%kernel_name%.typename0,typename1,..."
 
 ; RUN: llvm-as %s -o %t.bc
-; RUN: llvm-spirv %t.bc -spirv-text -o %t.spv.txt
+; RUN: llvm-spirv -preserve-ocl-kernel-arg-type-metadata-through-string %t.bc -spirv-text -o %t.spv.txt
 ; RUN: FileCheck < %t.spv.txt %s --check-prefix=CHECK-SPIRV
```
StuartDBrady, line 25: Could you rename this, to make it more apparent that it is checking for a workaround, please?

StuartDBrady: I.e. rename CHECK-SPIRV to CHECK-SPIRV-WORKAROUND, or add a comment, or some equivalent that makes this clearer.

MrSidims: Applied.
Basically the entire KernelArgTypeInOpString.ll and KernelArgTypeInOpString2.ll tests are for the workaround (note: current workaround with constant qualifier in kernel_arg_type_qual metadata is just an expansion for the one added years before, that adds translation of a kernel_arg_type(_name?) metadata). So there is already a comment in the test:  "As a workaround we store original names in OpString instruction".

test/transcoding/cl-types.ll
```diff
@@ -19,9 +19,9 @@
...
 
 ; RUN: llvm-as %s -o %t.bc
-; RUN: llvm-spirv %t.bc -spirv-text -o %t.spv.txt
+; RUN: llvm-spirv -preserve-ocl-kernel-arg-type-metadata-through-string %t.bc -spirv-text -o %t.spv.txt
 ; RUN: FileCheck < %t.spv.txt %s --check-prefix=CHECK-SPIRV
-; RUN: llvm-spirv %t.bc -o %t.spv
+; RUN: llvm-spirv -preserve-ocl-kernel-arg-type-metadata-through-string %t.bc -o %t.spv
 ; RUN: llvm-spirv -r %t.spv -o %t.rev.bc
 ; RUN: llvm-dis %t.rev.bc
 ; RUN: FileCheck < %t.rev.ll %s --check-prefix=CHECK-LLVM
```
StuartDBrady, line 27: This seems to be a test for the preservation of metadata through round-trip translation from LLVM to SPIR-V, and then back to LLVM.  Preservation of this metadata cannot be preserved when implementing SPIR-V translation as per the SPIR-V specification.  As such, preservation of the metadata is not guaranteed through the (somewhat informal) principles of operation of the translator.

It should be made clearer in this test that it cannot be expected to work using SPIR-V as specified, i.e. without the `-preserve-ocl-kernel-arg-type-metadata-through-string` option.  This is essentially vendor-specific functionality, and not clearly stated as such (although it could also be useful functionality for testing purposes, and other special cases).

Assuming all behavior tested here is either vendor-specific or covered elsewhere, the filename should be changed to make it clear that this is testing nonstandard functionality.  Otherwise, parts of this test that would be relevant when not using the `-preserve-ocl-kernel-arg-type-metadata-through-string` option should be run both with and without that option.

Please consider whether a comment is needed at the start of the test.  (I expect that if the filename is updated, or if extra RUN lines are added with appropriate CHECK-PREFIXes to distinguish between standard and nonstandard behavior, then a comment should not be needed.)

test/transcoding/KernelArgTypeInOpString2.ll
```diff
@@ -23,21 +23,31 @@
 ; OpString "kernel_arg_type.%kernel_name%.typename0,typename1,..."
 
 ; RUN: llvm-as %s -o %t.bc
-; RUN: llvm-spirv %t.bc -spirv-text -o %t.spv.txt
+; RUN: llvm-spirv -preserve-ocl-kernel-arg-type-metadata-through-string %t.bc -spirv-text -o %t.spv.txt
 ; RUN: FileCheck < %t.spv.txt %s --check-prefix=CHECK-SPIRV
```
StuartDBrady: Again, this should be renamed to something that makes it clearer that this is nonstandard behavior, e.g. CHECK-SPIRV-WORKAROUND.

MrSidims: renamed in the second patch

test/transcoding/kernel_arg_type_qual.ll
```diff
@@ -0,0 +1,43 @@
+; RUN: llvm-as %s -o %t.bc
+; RUN: llvm-spirv -preserve-ocl-kernel-arg-type-metadata-through-string %t.bc -o - -spirv-text | FileCheck %s --check-prefix=CHECK-SPIRV
+; RUN: llvm-spirv %t.bc -o - -spirv-text | FileCheck %s --check-prefix=CHECK-SPIRV-NEGATIVE
+; RUN: llvm-spirv -preserve-ocl-kernel-arg-type-metadata-through-string %t.bc -o %t.spv
+; RUN: llvm-spirv -r %t.spv -o %t.rev.bc
+; RUN: llvm-dis < %t.rev.bc | FileCheck %s --check-prefix=CHECK-LLVM
```
StuartDBrady, line 6: The CHECK-LLVM prefix here should be renamed to make it clearer that it is testing nonstandard functionality.

test/transcoding/spirv-types.ll
```diff
@@ -1,12 +1,13 @@
 ;; Test SPIR-V opaque types
 ;;
 ; RUN: llvm-as %s -o %t.bc
-; RUN: llvm-spirv %t.bc -spirv-text -o %t.spv.txt
+; RUN: llvm-spirv -preserve-ocl-kernel-arg-type-metadata-through-string %t.bc -spirv-text -o %t.spv.txt
```
StuartDBrady, line 4: Likewise, these should be renamed, or extra RUN lines and CHECK prefixes added.

test/transcoding/kernel_arg_name.ll
```diff
@@ -1,12 +1,12 @@
-; RUN: llvm-as %s -o - | llvm-spirv -o %t.spv
+; RUN: llvm-as %s -o - | llvm-spirv -preserve-ocl-kernel-arg-type-metadata-through-string -o %t.spv
```
StuartDBrady, line 1: Would any part of this test be valid without `-preserve-ocl-kernel-arg-type-metadata-through-string`?

Either the filename should be changed to make it clear that this is testing nonstandard functionality, or an extra RUN line should be added with the CHECKs split to allow testing to be conditionalised.

test/transcoding/spirv-types.ll
MrSidims: I'll go through the tests and make the appropriate changes in RUN and CHECK strings. But I'm a bit uncertain about "non-standard" functionality definition. I acknowledge, that the way of how these two metadata are being translated is a workaround, but the metadata themselves are defined in opencl-spir-12, so it'd better to be translated somehow. Do you think, that it's better at some point to introduce a SPIR-V extension to have their representation?

lib/SPIRV/SPIRVReader.cpp
```diff
@@ -3993,29 +3994,32 @@ bool SPIRVToLLVM::transOCLMetadata(SPIRVFunction *BF) {
                                  return MDString::get(*Context, Qual);
                                });
   // Generate metadata for kernel_arg_type
-  if (!transKernelArgTypeMedataFromString(Context, BM, F))
+  if (!transKernelArgTypeMedataFromString(Context, BM, F,
```
StuartDBrady, line 3997: This nonstandard metadata should only be consumed by the translator when the metadata preservation flag is enabled. The code to ensure this doesn't seem to have been added. Have I understood correctly? If so, please conditionalize this accordingly.

lib/SPIRV/SPIRVReader.cpp
```diff
@@ -3993,29 +3994,32 @@ bool SPIRVToLLVM::transOCLMetadata(SPIRVFunction *BF) {
...
-      Context, SPIR_MD_KERNEL_ARG_TYPE_QUAL, BF, F,
-      [=](SPIRVFunctionParameter *Arg) {
-        std::string Qual;
-        if (Arg->hasDecorate(DecorationVolatile))
-          Qual = kOCLTypeQualifierName::Volatile;
-        Arg->foreachAttr([&](SPIRVFuncParamAttrKind Kind) {
-          Qual += Qual.empty() ? "" : " ";
-          if (Kind == FunctionParameterAttributeNoAlias)
-            Qual += kOCLTypeQualifierName::Restrict;
+  if (!transKernelArgTypeMedataFromString(Context, BM, F,
```
StuartDBrady, line 4004: Likewise.

StuartDBrady:
The LLVM -> SPIR-V translation part of this looks okay to me, but changes are needed to conditionalize SPIR-V -> LLVM translation, and to make it clearer in the testing that this is nonstandard functionality.

StuartDBrady (changes requested):
Changes needed.

StuartDBrady:
Note: the commit message contains a typo: "OpStirng" -> "OpString".  I have updated the PR description but not the commit message.  (Please _either_ ensure this is squashed when submitted, or force-push the change, so as to correct this.)

StuartDBrady:
Note: _all_ nonstandard metadata translation, whether from SPIR-V -> LLVM or from LLVM -> SPIR-V, should be disabled by default (i.e. unless the preservation flag is passed).

MrSidims:
@StuartDBrady applied comments. The PR doesn't claim to disable non-standard metadata generation, but it moves some of them that were translated via OpString under the option.

MrSidims:
@StuartDBrady @svenvh  friendly ping
@AlexeySachkov @Fznamznon could you please take a look as well?

lib/SPIRV/SPIRVReader.cpp
```diff
@@ -3993,29 +3997,32 @@ bool SPIRVToLLVM::transOCLMetadata(SPIRVFunction *BF) {
                                  return MDString::get(*Context, Qual);
                                });
   // Generate metadata for kernel_arg_type
-  if (!transKernelArgTypeMedataFromString(Context, BM, F))
+  if (!transKernelArgTypeMedataFromString(Context, BM, F,
+                                          SPIR_MD_KERNEL_ARG_TYPE))
```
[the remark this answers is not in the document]
MrSidims: PR to master: https://github.com/KhronosGroup/SPIRV-LLVM-Translator/pull/996
This PR just got more attention.
W/A was indeed added sometime ago. Mentioned this in the commit message.

Fznamznon (approved):

pszymich:
@StuartDBrady A swift review of the new changes would be much appreciated.

test/transcoding/spirv-types.ll
StuartDBrady: Yes, a SPIR-V extension should have been introduced in the first place for this. Please don't forget to follow up by adding a SPIR-V extension.

lib/SPIRV/SPIRVReader.cpp
StuartDBrady: Looks good.

lib/SPIRV/SPIRVReader.cpp
StuartDBrady: Looks good now.

test/transcoding/kernel_arg_name.ll
StuartDBrady: This part of the change has simply been reverted without any rationale given.  Please ensure you answer review commits.  If preservation of the kernel_arg_name metadata is now disabled by default, then why does this have neither a flag nor an explanation?  Was the flag ever needed in the first place, here?

test/transcoding/kernel_arg_type_qual.ll
StuartDBrady: Looks okay.  In hindsight, it may have been better to place all of these tests for nonstandard metadata into their own subdirectory under transcoding, perhaps with a README or at least name that indicates clearly that this metadata is nonstandard.

test/transcoding/kernel_arg_name.ll
StuartDBrady: Thank you. It seems to me that it would be good to use a single option for all nonstandard kernel arg metadata, at some point... although better still to add an extension, and use that instead (if enabled).

StuartDBrady (approved):
Looks good to me.  Please ensure the PR for the master branch is updated promptly if needed.

PR #1111, opened 2021-07-13
Fix readonly attribute after new LLVM assertion

svenvh:
After LLVM commit 5d1464cbfe90 ("[Attributes] Make type attribute
handling more generic (NFCI)", 2021-07-12), the Attribute constructor
became stricter.  This revealed that the translator was creating
`readonly` attributes with a type attached, tripping the newly added
assertion in `Attribute::get`.


MrSidims (approved):
Thanks!

PR #1112, opened 2021-07-14
Update lit.cfg.py to avoid an error

Weiwei-2021:
A KeyError gets triggered if the lit test is running in a machine which pre-installs spirv tools without setting LD_LIBRARY_PATH in env. It would be good to add a check before using.

FYI, In my case, the llvm-spirv-translator is a llvm in-tree build. And I will use "make check-llvm-spirv" or "llvm-lit *.spv" to run  thoses lit tests. I don't have a chance to test whether it will be a problem in other build methods.

### DIFF
--- a/test/lit.cfg.py
+++ b/test/lit.cfg.py
@@ -73,6 +73,7 @@ if config.spirv_tools_have_spirv_val:
 else:
     config.substitutions.append(('spirv-val', ':'))
 
-if using_spirv_tools:
+env = config.environment
+if using_spirv_tools and ('LD_LIBRARY_PATH' in env):
     new_ld_library_path = os.path.pathsep.join((config.spirv_tools_lib_dir, config.environment['LD_LIBRARY_PATH']))
     config.environment['LD_LIBRARY_PATH'] = new_ld_library_path
--- a/test/lit.cfg.py
+++ b/test/lit.cfg.py
@@ -73,7 +73,6 @@ if config.spirv_tools_have_spirv_val:
 else:
     config.substitutions.append(('spirv-val', ':'))
 
-env = config.environment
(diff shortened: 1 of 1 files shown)


MrSidims:
Approved to run the tests.

test/lit.cfg.py
```diff
@@ -73,6 +73,7 @@
 else:
     config.substitutions.append(('spirv-val', ':'))
 
-if using_spirv_tools:
+env = config.environment
+if using_spirv_tools and ('LD_LIBRARY_PATH' in env):
     new_ld_library_path = os.path.pathsep.join((config.spirv_tools_lib_dir, config.environment['LD_LIBRARY_PATH']))
     config.environment['LD_LIBRARY_PATH'] = new_ld_library_path
```
AlexeySachkov, line 79: What if `config.spirv_tools_lib_dir` contains a correct path, but `LD_LIBRARY_PATH` is not set? In that case we won't set the proper `LD_LIBRARY_PATH` for SPIR-V Tools and it could lead to errors in some valid environments.

I think that the original code is not perfect and it should be refactored, which will also help you to fix the issue: we should use existing ` llvm_config.with_system_environment` and `llvm_config.with_environment` functions. For example you can take a look how they are used in [intel/llvm repo](https://github.com/intel/llvm/blob/sycl/sycl/test/lit.cfg.py#L63)

AlexeySachkov (changes requested):

test/lit.cfg.py
Weiwei-2021: Thank you for pointing out the example in intel/llvm repo. It is a much nicer approach to check the environment. I changed my commit and please take a look.

AlexeySachkov (approved):

svenvh (approved):

PR #1114, opened 2021-07-15
Don't add aliasing MemAccess mask if metadata translation failed

MrSidims:
Otherwise we might face an assert in case if empty metadata was generated by LLVM opt.

### DIFF
--- a/test/transcoding/SPV_INTEL_memory_access_aliasing/intel-alias-empty-md.ll
+++ b/test/transcoding/SPV_INTEL_memory_access_aliasing/intel-alias-empty-md.ll
@@ -1,0 +1,41 @@
+; Check that the translator doesn't fail on a translation of empty aliasing
+; metadata
+
+; RUN: llvm-as %s -o %t.bc
+; RUN: llvm-spirv %t.bc --spirv-ext=+SPV_INTEL_memory_access_aliasing -o %t.spv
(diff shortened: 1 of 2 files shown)


MrSidims:
@Fznamznon @vmaksimo @AGindinson please take a look

Fznamznon (commented):
The change seems logical, but could you please add a test?

MrSidims:
> The change seems logical, but could you please add a test?

Sure, done.

test/transcoding/SPV_INTEL_memory_access_aliasing/intel-alias-empty-md.ll
```diff
@@ -0,0 +1,41 @@
+; Check whether the translator fails on translation of empty aliasing metadata
+; or not
```
Fznamznon, line 2: NIT: I think we all agree that translator shouldn't fail at all :)
```suggestion
; Check that the translator doesn't fail on translation of empty aliasing metadata
```

Fznamznon (approved):

test/transcoding/SPV_INTEL_memory_access_aliasing/intel-alias-empty-md.ll
MrSidims: Indeed :)

PR #1117, opened 2021-07-19
Add pass to lower Bitcast to nonstandard type instructions

KornevNikita:
It is a pass to lower bitcast instructions to non-standard SPIR-V types.
It covers only known issues. At this moment there is only one pattern
that should be covered - use of vector with unsupported number of
elements. This pattern should be handled this way:

%0 = bitcast <3 x i64> addrspace(1)* @Id to <6 x i32> addrspace(1)*
%1 = addrspacecast <6 x i32> addrspace(1)* %0 to <6 x i32> addrspace(4)*
%2 = load <6 x i32>, <6 x i32> addrspace(4)* %1, align 32
%conv = extractelement <6 x i32> %2, i32 4
%conv1 = sitofp i32 %conv to float

Must be replaced by:
%0 = addrspacecast <3 x i64> addrspace(1)* @Id to  <3 x i64> addrspace(4)*
%1 = load <3 x i64>, <3 x i64> addrspace(4)* %0, align 32
%2 = extractelement <3 x i64> %1, i32 0
%conv = trunc i64 %2 to i32
%conv1 = sitofp i32 %conv to float

It is assumed that the pass will be further developed as new patterns arise.

### DIFF
--- a/include/LLVMSPIRVLib.h
+++ b/include/LLVMSPIRVLib.h
@@ -42,7 +42,6 @@
 #define SPIRV_H
 
 #include "LLVMSPIRVOpts.h"
-#include "llvm/Pass.h"
 
 #include <iostream>
 #include <string>
@@ -67,6 +66,7 @@ void initializePreprocessMetadataLegacyPass(PassRegistry &);
 void initializeSPIRVLowerBitCastToNonStandardTypeLegacyPass(PassRegistry &);
 
 class ModulePass;
+class FunctionPass;
 } // namespace llvm
 
 #include "llvm/IR/Module.h"
--- a/include/LLVMSPIRVLib.h
+++ b/include/LLVMSPIRVLib.h
@@ -42,6 +42,7 @@
 #define SPIRV_H
 
 #include "LLVMSPIRVOpts.h"
+#include "llvm/Pass.h"
 
 #include <iostream>
 #include <string>
@@ -218,7 +219,7 @@ ModulePass *createSPIRVWriterPass(std::ostream &Str,
 
 /// Create a pass for removing bitcast instructions to non-standard SPIR-V
 /// types
-ModulePass *createSPIRVLowerBitCastToNonStandardTypeLegacy();
+FunctionPass *createSPIRVLowerBitCastToNonStandardTypeLegacy();
 
 } // namespace llvm
 
--- a/include/LLVMSPIRVLib.h
+++ b/include/LLVMSPIRVLib.h
@@ -63,6 +63,7 @@ void initializeSPIRVRegularizeLLVMLegacyPass(PassRegistry &);
 void initializeSPIRVToOCL12LegacyPass(PassRegistry &);
 void initializeSPIRVToOCL20LegacyPass(PassRegistry &);
 void initializePreprocessMetadataLegacyPass(PassRegistry &);
+void initializeSPIRVLowerBitCastToNonStandardTypeLegacyPass(PassRegistry &);
 
 class ModulePass;
 } // namespace llvm
@@ -214,6 +215,10 @@ ModulePass *createSPIRVWriterPass(std::ostream &Str);
 /// ostream.
 ModulePass *createSPIRVWriterPass(std::ostream &Str,
                                   const SPIRV::TranslatorOpts &Opts);
+
+/// Create a pass for removing bitcast instructions to non-standard SPIR-V
+/// types
+ModulePass *createSPIRVLowerBitCastToNonStandardTypeLegacy();
 
 } // namespace llvm
 
--- a/lib/SPIRV/SPIRVLowerBitCastToNonStandardType.cpp
+++ b/lib/SPIRV/SPIRVLowerBitCastToNonStandardType.cpp
@@ -53,6 +53,72 @@ using namespace llvm;
 
 namespace SPIRV {
 
+static VectorType *getVectorType(Type *Ty) {
+  if (!Ty)
+    return nullptr;
+  if (auto *ElemTy = dyn_cast<PointerType>(Ty))
+    Ty = ElemTy->getElementType();
+  return dyn_cast<VectorType>(Ty);
+}
+
+/// Since SPIR-V does not support non-standard vector types, instructions using
+/// these types should be replaced in a special way to avoid using of
+/// unsupported types.
+/// lowerBitCastToNonStdVec function is designed to avoid using of bitcast to
+/// unsupported vector types instructions and should be called if similar
+/// instructions have been encountered in input LLVM IR.
+bool lowerBitCastToNonStdVec(Instruction *Inst,
+                             std::vector<Instruction *> &InstsToErase) {
+  Instruction *OldInstIter(Inst);
+  IRBuilder<> Builder(OldInstIter);
+  Value *NewValue = OldInstIter->getOperand(0);
+  VectorType *NewVecTy = getVectorType(NewValue->getType());
+  InstsToErase.push_back(OldInstIter);
+  // Handle addrspacecast instruction after bitcast if present
+  for (auto *U : OldInstIter->users()) {
+    if (auto *ASCastInst = dyn_cast<AddrSpaceCastInst>(U)) {
+      unsigned DestAS = ASCastInst->getDestAddressSpace();
+      auto *NewVecPtrTy = NewVecTy->getPointerTo(DestAS);
+      // AddrSpaceCast is created explicitly instead of using method
+      // IRBuilder<>.CreateAddrSpaceCast because IRBuilder doesn't create
+      // separate instruction for constant values. Whereas SPIR-V translator
+      // doesn't like several nested instructions in one.
+      NewValue = new AddrSpaceCastInst(NewValue, NewVecPtrTy);
+      Builder.Insert(NewValue);
+      OldInstIter = ASCastInst;
+      InstsToErase.push_back(OldInstIter);
+      break;
+    }
+  }
+  // Handle load instruction which is following the bitcast in the pattern
+  for (auto *U : OldInstIter->users()) {
+    if (auto *LI = dyn_cast<LoadInst>(U)) {
+      NewValue = Builder.CreateLoad(NewVecTy, NewValue);
+      OldInstIter = LI;
+      InstsToErase.push_back(OldInstIter);
+      break;
+    }
+  }
+  // Handle extractelement instruction which is following the load
+  for (auto *U : OldInstIter->users()) {
+    if (auto *EEI = dyn_cast<ExtractElementInst>(U)) {
+      VectorType *OldVecTy = getVectorType(Inst->getType());
+      uint64_t NumElemsInOldVec = OldVecTy->getElementCount().getValue();
+      uint64_t NumElemsInNewVec = NewVecTy->getElementCount().getValue();
+      uint64_t ElemIdx =
+          cast<ConstantInt>(EEI->getIndexOperand())->getSExtValue() /
+          (NumElemsInOldVec / NumElemsInNewVec);
+      NewValue = Builder.CreateExtractElement(NewValue, ElemIdx);
+      NewValue = Builder.CreateTrunc(NewValue, OldVecTy->getElementType());
+      OldInstIter = EEI;
+      InstsToErase.push_back(OldInstIter);
+      break;
+    }
+  }
+  OldInstIter->replaceAllUsesWith(NewValue);
+  return true;
+}
+
 class SPIRVLowerBitCastToNonStandardTypePass {
 public:
   SPIRVLowerBitCastToNonStandardTypePass() {}
@@ -64,20 +130,31 @@ public:
     // parameter, since it added by an optimization.
     bool Changed = false;
 
-    std::vector<std::pair<Instruction *, VectorType *>> BCastsToNonStdVec;
+    std::vector<Instruction *> BCastsToNonStdVec;
     std::vector<Instruction *> InstsToErase;
     for (auto &BB : F)
       for (auto &I : BB) {
         auto *BC = dyn_cast<BitCastInst>(&I);
         if (!BC)
           continue;
-        auto *DestTy = BC->getDestTy();
-        if (auto *ElemTy = dyn_cast<PointerType>(DestTy))
-          DestTy = ElemTy->getElementType();
-        if (auto *DestVecTy = dyn_cast<VectorType>(DestTy)) {
+        auto *SrcTy = BC->getSrcTy();
+        if (auto *SrcElemTy = dyn_cast<PointerType>(SrcTy))
+          SrcTy = SrcElemTy->getElementType();
+        if (auto *SrcVecTy = dyn_cast<VectorType>(SrcTy)) {
+          uint64_t NumElemsInSrcVec = SrcVecTy->getElementCount().getValue();
+          // assert(isValidVectorSize(NumElemsInSrcVec) &&
+          //  ("Unsupported vector type with the size of: " +
+          //   NumElemsInSrcVec));
+          if (!isValidVectorSize(NumElemsInSrcVec))
+            report_fatal_error("Unsupported vector type with the size of: " +
+                                  NumElemsInSrcVec,
+                              false);
+        }
+        VectorType *DestVecTy = getVectorType(BC->getDestTy());
+        if (DestVecTy) {
           uint64_t NumElemsInDestVec = DestVecTy->getElementCount().getValue();
           if (!isValidVectorSize(NumElemsInDestVec))
-            BCastsToNonStdVec.push_back(std::make_pair(&I, DestVecTy));
+            BCastsToNonStdVec.push_back(&I);
         }
       }
 
@@ -88,61 +165,6 @@ public:
       InstsToErase[I]->eraseFromParent();
 
     return Changed ? PreservedAnalyses::none() : PreservedAnalyses::all();
-  }
-
-  bool lowerBitCastToNonStdVec(std::pair<Instruction *, VectorType *> &Inst,
-                               std::vector<Instruction *> &InstsToErase) {
-    Instruction *I = Inst.first;
-    InstsToErase.push_back(I);
-    auto *InstIter = I;
-    IRBuilder<> Builder(I);
-    Value *NewSrc = I->getOperand(0);
-    // It is assumed that the function can contain addrspacecast after handled
-    // bitcast instruction, so addrspacecast should also be handled
-    for (auto *U : I->users()) {
-      if (auto *ASCastInst = dyn_cast<AddrSpaceCastInst>(U)) {
-        unsigned DestAddrSpace = ASCastInst->getDestAddressSpace();
-        auto *SrcTy = cast<PointerType>(NewSrc->getType())->getElementType();
-        auto *NewDestPtrTy = SrcTy->getPointerTo(DestAddrSpace);
-        NewSrc = new AddrSpaceCastInst(NewSrc, NewDestPtrTy);
-        Builder.Insert(NewSrc);
-        InstIter = ASCastInst;
-        InstsToErase.push_back(InstIter);
-        break;
-      }
-    }
-    auto *SrcTy = cast<PointerType>((NewSrc)->getType())->getElementType();
-    auto *Load = Builder.CreateLoad(SrcTy, NewSrc);
-    // In the already known pattern, the bitcast is followed by load instruction
-    for (auto *U : InstIter->users()) {
-      if (auto *LI = dyn_cast<LoadInst>(U)) {
-        InstIter = LI;
-        InstsToErase.push_back(InstIter);
-        break;
-      }
-    }
-    // The bitcast is followed by extractelement instruction
-    VectorType *DestVecTy = Inst.second;
-    uint64_t NumElemsInDestVec = DestVecTy->getElementCount().getValue();
-    uint64_t NumElemsInSrcVec =
-        cast<VectorType>(SrcTy)->getElementCount().getValue();
-    int ElemIdx = 0;
-    Instruction *ExtrElInstIter = InstIter;
-    for (auto *U : InstIter->users()) {
-      if (auto *EEI = dyn_cast<ExtractElementInst>(U)) {
-        ElemIdx = cast<ConstantInt>(EEI->getIndexOperand())->getSExtValue() /
-                  (NumElemsInDestVec / NumElemsInSrcVec);
-        ExtrElInstIter = cast<Instruction>(U);
-        InstsToErase.push_back(ExtrElInstIter);
-        break;
-      }
-    }
-    auto *ExtractElement = Builder.CreateExtractElement(
-        Load, ConstantInt::get(Type::getInt64Ty(I->getContext()), ElemIdx));
-    auto *Trunc =
-        Builder.CreateTrunc(ExtractElement, DestVecTy->getElementType());
-    ExtrElInstIter->replaceAllUsesWith(Trunc);
-    return true;
   }
 };
 
--- a/lib/SPIRV/SPIRVLowerBitCastToNonStandardType.cpp
+++ b/lib/SPIRV/SPIRVLowerBitCastToNonStandardType.cpp
@@ -163,8 +163,7 @@ public:
   }
 };
 
-class SPIRVLowerBitCastToNonStandardTypeLegacy
-    : public FunctionPass {
+class SPIRVLowerBitCastToNonStandardTypeLegacy : public FunctionPass {
 public:
   static char ID;
   SPIRVLowerBitCastToNonStandardTypeLegacy() : FunctionPass(ID) {}
--- a/lib/SPIRV/SPIRVLowerBitCastToNonStandardType.cpp
+++ b/lib/SPIRV/SPIRVLowerBitCastToNonStandardType.cpp
@@ -63,7 +63,7 @@ public:
   bool runLowerBitCastToNonStandardType(Module &Module) {
     // This pass doesn't cover all possible uses of non-standard types, only
     // known
-    auto M = &Module;
+    auto *M = &Module;
     bool Changed = false;
 
     std::vector<std::pair<Instruction *, VectorType *>> BCastsToNonStdVec;
@@ -154,14 +154,14 @@ public:
         }
       }
     }
-    auto ExtractElement = ExtractElementInst::Create(
+    auto *ExtractElement = ExtractElementInst::Create(
         Load, ConstantInt::get(Type::getInt64Ty(I->getContext()), ElemIdx));
     InstsToInsert.push_back(ExtractElement);
     auto *Trunc = new TruncInst(ExtractElement, DestVecTy->getElementType());
     InstsToInsert.push_back(Trunc);
     InstsToInsert[0]->insertBefore(I);
-    for (unsigned long i = 1; i < InstsToInsert.size(); i++)
-      InstsToInsert[i]->insertAfter(InstsToInsert[i - 1]);
+    for (size_t It = 1; It < InstsToInsert.size(); It++)
+      InstsToInsert[It]->insertAfter(InstsToInsert[It - 1]);
     IIII->replaceAllUsesWith(Trunc);
     if (InstsToInsert.size())
       Changed = true;
--- a/lib/SPIRV/SPIRVLowerBitCastToNonStandardType.cpp
+++ b/lib/SPIRV/SPIRVLowerBitCastToNonStandardType.cpp
@@ -137,18 +137,13 @@ public:
         auto *BC = dyn_cast<BitCastInst>(&I);
         if (!BC)
           continue;
-        auto *SrcTy = BC->getSrcTy();
-        if (auto *SrcElemTy = dyn_cast<PointerType>(SrcTy))
-          SrcTy = SrcElemTy->getElementType();
-        if (auto *SrcVecTy = dyn_cast<VectorType>(SrcTy)) {
+        VectorType *SrcVecTy = getVectorType(BC->getSrcTy());
+        if (SrcVecTy) {
           uint64_t NumElemsInSrcVec = SrcVecTy->getElementCount().getValue();
-          // assert(isValidVectorSize(NumElemsInSrcVec) &&
-          //  ("Unsupported vector type with the size of: " +
-          //   NumElemsInSrcVec));
           if (!isValidVectorSize(NumElemsInSrcVec))
             report_fatal_error("Unsupported vector type with the size of: " +
-                                  NumElemsInSrcVec,
-                              false);
+                                   std::to_string(NumElemsInSrcVec),
+                               false);
         }
         VectorType *DestVecTy = getVectorType(BC->getDestTy());
         if (DestVecTy) {
@@ -161,8 +156,8 @@ public:
     for (auto &I : BCastsToNonStdVec)
       Changed |= lowerBitCastToNonStdVec(I, InstsToErase);
 
-    for (int I = InstsToErase.size() - 1; I >= 0; I--)
-      InstsToErase[I]->eraseFromParent();
+    for (auto I = InstsToErase.rbegin(); I != InstsToErase.rend(); ++I)
+      (*I)->eraseFromParent();
 
     return Changed ? PreservedAnalyses::none() : PreservedAnalyses::all();
   }
--- a/lib/SPIRV/SPIRVLowerBitCastToNonStandardType.cpp
+++ b/lib/SPIRV/SPIRVLowerBitCastToNonStandardType.cpp
@@ -74,6 +74,8 @@ bool lowerBitCastToNonStdVec(Instruction *Inst,
   VectorType *NewVecTy = getVectorType(NewValue->getType());
   InstsToErase.push_back(OldInstIter);
   // Handle addrspacecast instruction after bitcast if present
+  // TODO: There are can be many addrspacecast instructions using this bitcast,
+  // so that it is necessary to handle all these instructions in the same way
   for (auto *U : OldInstIter->users()) {
     if (auto *ASCastInst = dyn_cast<AddrSpaceCastInst>(U)) {
       unsigned DestAS = ASCastInst->getDestAddressSpace();
--- a/lib/SPIRV/SPIRVLowerBitCastToNonStandardType.cpp
+++ b/lib/SPIRV/SPIRVLowerBitCastToNonStandardType.cpp
@@ -53,8 +53,6 @@ using namespace llvm;
 
 namespace SPIRV {
 
-constexpr unsigned MaxRecursionDepth = 16;
-
 static VectorType *getVectorType(Type *Ty) {
   assert(Ty != nullptr && "Expected non-null type");
   if (auto *ElemTy = dyn_cast<PointerType>(Ty))
@@ -71,11 +69,12 @@ static VectorType *getVectorType(Type *Ty) {
 bool lowerBitCastToNonStdVec(Instruction *OldInst, Value *NewInst,
                              const VectorType *OldVecTy,
                              std::vector<Instruction *> &InstsToErase,
-                             IRBuilder<> &Builder, unsigned RecursionDepth) {
-  if (RecursionDepth > MaxRecursionDepth)
+                             IRBuilder<> &Builder,
+                             unsigned RecursionDepth = 0) {
+  static constexpr unsigned MaxRecursionDepth = 16;
+  if (RecursionDepth++ > MaxRecursionDepth)
     report_fatal_error(
         "The depth of recursion exceeds the maximum possible depth", false);
-  RecursionDepth++;
   bool Changed = false;
   VectorType *NewVecTy = getVectorType(NewInst->getType());
   if (NewVecTy) {
@@ -162,9 +161,8 @@ public:
     for (auto &I : BCastsToNonStdVec) {
       Value *NewValue = I->getOperand(0);
       VectorType *OldVecTy = getVectorType(I->getType());
-      unsigned RecursionDepth = 0;
       Changed |= lowerBitCastToNonStdVec(I, NewValue, OldVecTy, InstsToErase,
-                                         Builder, RecursionDepth);
+                                         Builder);
     }
 
     for (auto *I : InstsToErase)
--- a/lib/SPIRV/SPIRVLowerBitCastToNonStandardType.cpp
+++ b/lib/SPIRV/SPIRVLowerBitCastToNonStandardType.cpp
@@ -54,8 +54,7 @@ using namespace llvm;
 namespace SPIRV {
 
 static VectorType *getVectorType(Type *Ty) {
-  if (!Ty)
-    return nullptr;
+  assert(Ty != nullptr && "Expected non-null type");
   if (auto *ElemTy = dyn_cast<PointerType>(Ty))
     Ty = ElemTy->getElementType();
   return dyn_cast<VectorType>(Ty);
@@ -119,7 +118,8 @@ bool lowerBitCastToNonStdVec(Instruction *Inst,
   return true;
 }
 
-class SPIRVLowerBitCastToNonStandardTypePass {
+class SPIRVLowerBitCastToNonStandardTypePass
+    : public llvm::PassInfoMixin<SPIRVLowerBitCastToNonStandardTypePass> {
 public:
   SPIRVLowerBitCastToNonStandardTypePass() {}
 
@@ -164,8 +164,7 @@ public:
 };
 
 class SPIRVLowerBitCastToNonStandardTypeLegacy
-    : public FunctionPass,
-      public SPIRVLowerBitCastToNonStandardTypePass {
+    : public FunctionPass {
 public:
   static char ID;
   SPIRVLowerBitCastToNonStandardTypeLegacy() : FunctionPass(ID) {}
--- a/lib/SPIRV/SPIRVLowerBitCastToNonStandardType.cpp
+++ b/lib/SPIRV/SPIRVLowerBitCastToNonStandardType.cpp
@@ -76,9 +76,7 @@ public:
           DestTy = ElemTy->getElementType();
         if (auto *DestVecTy = dyn_cast<VectorType>(DestTy)) {
           uint64_t NumElemsInDestVec = DestVecTy->getElementCount().getValue();
-          if (NumElemsInDestVec != 2 && NumElemsInDestVec != 3 &&
-              NumElemsInDestVec != 4 && NumElemsInDestVec != 8 &&
-              NumElemsInDestVec != 16)
+          if (!isValidVectorSize(NumElemsInDestVec))
             BCastsToNonStdVec.push_back(std::make_pair(&I, DestVecTy));
         }
       }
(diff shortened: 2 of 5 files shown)


KornevNikita:
@AlexeySachkov @AlexeySotkin @Fznamznon take a look please

test/lower-non-standard-types.ll
```diff
@@ -0,0 +1,44 @@
+; It is assumed that the test will not fail
+
+; RUN: llvm-as %s -o %t.bc
+; RUN: llvm-spirv %t.bc -o %t.spv
+; RUN: llvm-spirv -r %t.spv -o %t.rev.bc
+
```
AlexeySotkin, line 6: Please add checks to demonstrate the result of the pass

KornevNikita: Done

lib/SPIRV/SPIRVLowerBitCastToNonStandardType.cpp
```diff
@@ -0,0 +1,208 @@
...
+
+  bool runLowerBitCastToNonStandardType(Module &Module) {
+    // This pass doesn't cover all possible uses of non-standard types, only
+    // known
+    auto *M = &Module;
+    bool Changed = false;
+
+    std::vector<std::pair<Instruction *, VectorType *>> BCastsToNonStdVec;
+    std::stack<Instruction *> InstsToErase;
+    for (auto &F : M->functions())
```
Fznamznon, line 71: Why don't we make it a function pass instead of writing this loop manually?

lib/SPIRV/SPIRVLowerBitCastToNonStandardType.cpp
```diff
@@ -0,0 +1,208 @@
...
+        if (ASCI->getOperand(0) == cast<Value>(I)) {
+          Src = new AddrSpaceCastInst(I->getOperand(0), SrcPointer);
+          InstsToInsert.push_back(Src);
+          InstsToErase.push(II);
+          break;
+        }
+      }
+    }
+    if (!II)
+      II = I;
```
Fznamznon, line 125: Can we give more informative names to variables, please?

lib/SPIRV/SPIRVLowerBitCastToNonStandardType.cpp
```diff
@@ -0,0 +1,208 @@
...
+    auto *SrcTy =
+        cast<PointerType>(cast<Value>(Src)->getType())->getElementType();
+    LoadInst *Load =
+        Src ? new LoadInst(SrcTy, Src, "", false, Align())
+            : new LoadInst(SrcTy, I->getOperand(0), "", false, Align());
+    InstsToInsert.push_back(Load);
+    auto *III = II;
+    while ((III = III->getNextNode())) {
```
Fznamznon, line 133: This pass seems to handle one hardcoded pattern, right? Why don't we go through users of faulty bitcast instead (and users of users until the bad type is gone)? I think it can make code simpler (at least I assume you won't need that much IIIs)

lib/SPIRV/SPIRVLowerBitCastToNonStandardType.cpp
```diff
@@ -0,0 +1,208 @@
...
+    // known
+    auto *M = &Module;
+    bool Changed = false;
+
+    std::vector<std::pair<Instruction *, VectorType *>> BCastsToNonStdVec;
+    std::stack<Instruction *> InstsToErase;
+    for (auto &F : M->functions())
+      for (auto &BB : F)
+        for (auto &I : BB)
+          if (BitCastInst *BC = dyn_cast<BitCastInst>(&I)) {
```
Fznamznon, line 74: You can use auto when type is obvious. + to be consistent with the following code
```suggestion
          if (auto *BC = dyn_cast<BitCastInst>(&I)) {
```

lib/SPIRV/SPIRVLowerBitCastToNonStandardType.cpp
```diff
@@ -0,0 +1,208 @@
...
+    // This pass doesn't cover all possible uses of non-standard types, only
+    // known
+    auto *M = &Module;
+    bool Changed = false;
+
+    std::vector<std::pair<Instruction *, VectorType *>> BCastsToNonStdVec;
+    std::stack<Instruction *> InstsToErase;
+    for (auto &F : M->functions())
+      for (auto &BB : F)
+        for (auto &I : BB)
```
Fznamznon, line 73: I guess we can mention that we assume that bad type won't be passed to a function as parameter, since it added by an optimization and we hope that optimizations don't change functions signature, right?

lib/SPIRV/SPIRVLowerBitCastToNonStandardType.cpp
```diff
@@ -0,0 +1,208 @@
...
+    bool Changed = false;
+    std::vector<Instruction *> InstsToInsert;
+    Instruction *I = Inst.first;
+    VectorType *DestVecTy = Inst.second;
+    uint64_t NumElemsInDestVec = DestVecTy->getElementCount().getValue();
+    InstsToErase.push(I);
+    AddrSpaceCastInst *Src{nullptr};
+    auto *II = I;
+    while ((II = II->getNextNode())) {
+      if (auto *ASCI = dyn_cast<AddrSpaceCastInst>(II)) {
```
Fznamznon, line 111: That is a bit unusual name for someone who works with LLVM code for some time
```suggestion
      if (auto *AsCastInst = dyn_cast<AddrSpaceCastInst>(II)) {
```

lib/SPIRV/SPIRVLowerBitCastToNonStandardType.cpp
```diff
@@ -0,0 +1,208 @@
...
+
+  bool lowerBitCastToNonStdVec(std::pair<Instruction *, VectorType *> &Inst,
+                               std::stack<Instruction *> &InstsToErase) {
+    bool Changed = false;
+    std::vector<Instruction *> InstsToInsert;
+    Instruction *I = Inst.first;
+    VectorType *DestVecTy = Inst.second;
+    uint64_t NumElemsInDestVec = DestVecTy->getElementCount().getValue();
+    InstsToErase.push(I);
+    AddrSpaceCastInst *Src{nullptr};
```
Fznamznon, line 108: Other variables seem to be initialized via old fashioned assignment style, why so is here?
```suggestion
    AddrSpaceCastInst *Src = nullptr;
```

Fznamznon:
I didn't dig deeply into all loops, because for me recursive loop over users of faulty bitcast seems more convenient.
We also should remember that the pass like this, i.e. changing types can be very hard to maintain, since there is always more unhandled patterns coming.

AlexeySotkin:
> We also should remember that the pass like this, i.e. changing types can be very hard to maintain, since there is always more unhandled patterns coming.

I can't think of a better solution to handle such types. @Fznamznon, if you have any ideas, please share.

lib/SPIRV/SPIRVLowerBitCastToNonStandardType.cpp
```diff
@@ -0,0 +1,202 @@
...
+    // This pass doesn't cover all possible uses of non-standard types, only
+    // known. We assume that bad type won't be passed to a function as
+    // parameter, since it added by an optimization.
+    bool Changed = false;
+
+    std::vector<std::pair<Instruction *, VectorType *>> BCastsToNonStdVec;
+    std::stack<Instruction *> InstsToErase;
+    for (auto &BB : F)
+      for (auto &I : BB)
+        if (auto *BC = dyn_cast<BitCastInst>(&I)) {
```
mlychkov, line 71: I'd suggest to rewrite it to reduce nesting level:
```suggestion
        auto *BC = dyn_cast<BitCastInst>(&I);
        if (!BC)
          continue;
```

lib/SPIRV/SPIRVLowerBitCastToNonStandardType.cpp
```diff
@@ -0,0 +1,202 @@
...
+
+  PreservedAnalyses
+  runLowerBitCastToNonStandardType(Function &F, FunctionAnalysisManager &FAM) {
+    // This pass doesn't cover all possible uses of non-standard types, only
+    // known. We assume that bad type won't be passed to a function as
+    // parameter, since it added by an optimization.
+    bool Changed = false;
+
+    std::vector<std::pair<Instruction *, VectorType *>> BCastsToNonStdVec;
+    std::stack<Instruction *> InstsToErase;
```
mlychkov, line 68: Why is stack used here? May vector be used instead?

lib/SPIRV/SPIRVLowerBitCastToNonStandardType.cpp
```diff
@@ -0,0 +1,202 @@
...
+                               std::stack<Instruction *> &InstsToErase) {
+    bool Changed = false;
+    std::vector<Instruction *> InstsToInsert;
+    Instruction *I = Inst.first;
+    VectorType *DestVecTy = Inst.second;
+    uint64_t NumElemsInDestVec = DestVecTy->getElementCount().getValue();
+    InstsToErase.push(I);
+    AddrSpaceCastInst *Src = nullptr;
+    auto *ASCastInstIter = I;
+    while ((ASCastInstIter = ASCastInstIter->getNextNode())) {
```
mlychkov, line 106: It would be nice to have comment about a purpose of blocks started at lines 106, 128, 138.

lib/SPIRV/SPIRVLowerBitCastToNonStandardType.cpp
```diff
@@ -0,0 +1,202 @@
...
+                    (NumElemsInDestVec / NumElemsInSrcVec);
+          ExtrElInstIter = ASCastInstIter;
+          InstsToErase.push(ExtrElInstIter);
+          break;
+        }
+      }
+    }
+    auto *ExtractElement = ExtractElementInst::Create(
+        Load, ConstantInt::get(Type::getInt64Ty(I->getContext()), ElemIdx));
+    InstsToInsert.push_back(ExtractElement);
```
mlychkov, line 155: It is simpler to use llvm::IRBuilder<> instead of saving instructions just for setting an order of insertion.

lib/SPIRV/SPIRVLowerBitCastToNonStandardType.cpp
```diff
@@ -0,0 +1,199 @@
...
+
+  bool lowerBitCastToNonStdVec(std::pair<Instruction *, VectorType *> &Inst,
+                               std::vector<Instruction *> &InstsToErase) {
+    bool Changed = false;
+    Instruction *I = Inst.first;
+    VectorType *DestVecTy = Inst.second;
+    uint64_t NumElemsInDestVec = DestVecTy->getElementCount().getValue();
+    InstsToErase.push_back(I);
+    auto *ASCastInstIter = I;
+    IRBuilder<> Builder(I->getParent());
```
mlychkov, line 103: IRBuilder has constructor with 'Instruction *' as input.

include/LLVMSPIRVLib.h
```diff
@@ -63,6 +64,7 @@ void initializeSPIRVRegularizeLLVMLegacyPass(PassRegistry &);
 void initializeSPIRVToOCL12LegacyPass(PassRegistry &);
 void initializeSPIRVToOCL20LegacyPass(PassRegistry &);
 void initializePreprocessMetadataLegacyPass(PassRegistry &);
+void initializeSPIRVLowerBitCastToNonStandardTypeLegacyPass(PassRegistry &);
 
 class ModulePass;
```
mlychkov, line 69:
```suggestion
class ModulePass;
class FunctionPass;
```
instead of including Pass.h file.

lib/SPIRV/SPIRVLowerBitCastToNonStandardType.cpp
```diff
@@ -0,0 +1,199 @@
...
+      for (auto &I : BB) {
+        auto *BC = dyn_cast<BitCastInst>(&I);
+        if (!BC)
+          continue;
+        auto *DestTy = BC->getDestTy();
+        if (auto *ElemTy = dyn_cast<PointerType>(DestTy))
+          DestTy = ElemTy->getElementType();
+        if (auto *DestVecTy = dyn_cast<VectorType>(DestTy)) {
+          uint64_t NumElemsInDestVec = DestVecTy->getElementCount().getValue();
+          if (NumElemsInDestVec != 2 && NumElemsInDestVec != 3 &&
```
mlychkov, line 79: There is isValidVectorSize(NumElemsInDestVec) function in SPIRVInternal.h

test/lower-non-standard-types.ll
```diff
@@ -0,0 +1,47 @@
...
+source_filename = "lower-non-standard-types.cpp"
+target datalayout = "e-i64:64-v16:16-v24:32-v32:32-v48:64-v96:128-v192:256-v256:256-v512:512-v1024:1024-n8:16:32:64"
+target triple = "spir64-unknown-unknown-sycldevice"
+
+%"simd" = type { <8 x float> }
+
+@__spirv_BuiltInGlobalInvocationId = external dso_local local_unnamed_addr addrspace(1) constant <3 x i64>, align 32
+
+; Function Attrs: convergent norecurse
+define dso_local spir_func void @vmult2(%"simd"* %a) local_unnamed_addr #0 !sycl_explicit_simd !4 !intel_reqd_sub_group_size !6 {
```
mlychkov, line 23: Variable %a is not used. Test may be reduced by removing %a and simd type definition.
It also would be nice to shorten the name of @__spirv_BuiltInGlobalInvocationId because it is insufficient here.

lib/SPIRV/SPIRVLowerBitCastToNonStandardType.cpp
```diff
@@ -0,0 +1,183 @@
...
+    std::vector<Instruction *> InstsToErase;
+    for (auto &BB : F)
+      for (auto &I : BB) {
+        auto *BC = dyn_cast<BitCastInst>(&I);
+        if (!BC)
+          continue;
+        auto *DestTy = BC->getDestTy();
+        if (auto *ElemTy = dyn_cast<PointerType>(DestTy))
+          DestTy = ElemTy->getElementType();
+        if (auto *DestVecTy = dyn_cast<VectorType>(DestTy)) {
```
mlychkov, line 77:
```suggestion
        VectorType *DestVecTy = getVectorType(BC->getDestTy());
        if (DestVecTy)
```

mlychkov: namespace {

static VectorType *getVectorType(Type *Ty) {
  if (!Ty)
    return nullptr;
  if (auto *ElemTy = dyn_cast<PointerType>(Ty))
    Ty = ElemTy->getElementType();
  return dyn_cast<VectorType>(Ty);
}

} // namespace

lib/SPIRV/SPIRVLowerBitCastToNonStandardType.cpp
```diff
@@ -0,0 +1,183 @@
...
+  SPIRVLowerBitCastToNonStandardTypePass() {}
+
+  PreservedAnalyses
+  runLowerBitCastToNonStandardType(Function &F, FunctionAnalysisManager &FAM) {
+    // This pass doesn't cover all possible uses of non-standard types, only
+    // known. We assume that bad type won't be passed to a function as
+    // parameter, since it added by an optimization.
+    bool Changed = false;
+
+    std::vector<std::pair<Instruction *, VectorType *>> BCastsToNonStdVec;
```
mlychkov, line 67: I think that using vector of pairs complicates code readability
```suggestion
    std::vector<Instruction *> BCastsToNonStdVec;
```

lib/SPIRV/SPIRVLowerBitCastToNonStandardType.cpp
```diff
@@ -0,0 +1,183 @@
...
+        auto *BC = dyn_cast<BitCastInst>(&I);
+        if (!BC)
+          continue;
+        auto *DestTy = BC->getDestTy();
+        if (auto *ElemTy = dyn_cast<PointerType>(DestTy))
+          DestTy = ElemTy->getElementType();
+        if (auto *DestVecTy = dyn_cast<VectorType>(DestTy)) {
+          uint64_t NumElemsInDestVec = DestVecTy->getElementCount().getValue();
+          if (!isValidVectorSize(NumElemsInDestVec))
+            BCastsToNonStdVec.push_back(std::make_pair(&I, DestVecTy));
```
mlychkov, line 80:
```suggestion
            BCastsToNonStdVec.push_back(&I);
```

lib/SPIRV/SPIRVLowerBitCastToNonStandardType.cpp
```diff
@@ -0,0 +1,183 @@
...
+      }
+    }
+    auto *ExtractElement = Builder.CreateExtractElement(
+        Load, ConstantInt::get(Type::getInt64Ty(I->getContext()), ElemIdx));
+    auto *Trunc =
+        Builder.CreateTrunc(ExtractElement, DestVecTy->getElementType());
+    ExtrElInstIter->replaceAllUsesWith(Trunc);
+    return true;
+  }
+};
```
mlychkov, line 147:
```suggestion
  bool lowerBitCastToNonStdVec(Instruction *Inst,
                               std::vector<Instruction *> &InstsToErase) {
    Instruction *OldInstIter{Inst};
    IRBuilder<> Builder(OldInstIter);
    Value *NewValue = OldInstIter->getOperand(0);
    VectorType *NewVecTy = getVectorType(NewValue->getType());
    InstsToErase.push_back(OldInstIter);

    // Handle addrspacecast instruction after bitcast if present
    for (auto *U : OldInstIter->users()) {
      if (auto *ASCastInst = dyn_cast<AddrSpaceCastInst>(U)) {
        unsigned DestAS = ASCastInst->getDestAddressSpace();
        auto *NewVecPtrTy = NewVecTy->getPointerTo(DestAS);
        // AddrSpaceCast is created explicitly instead of using method
        // IRBuilder<>.CreateAddrSpaceCast because IRBuilder doesn't create
        // separate instruction for constant values. Whereas SPIR-V translator
        // doesn't like several nested instructions in one.
        NewValue = new AddrSpaceCastInst(NewValue, NewVecPtrTy);
        Builder.Insert(NewValue);

        OldInstIter = ASCastInst;
        InstsToErase.push_back(OldInstIter);
        break;
      }
    }

    // Handle load instruction which is following the bitcast in the pattern
    for (auto *U : OldInstIter->users()) {
      if (auto *LI = dyn_cast<LoadInst>(U)) {
        NewValue = Builder.CreateLoad(NewVecTy, NewValue);

        OldInstIter = LI;
        InstsToErase.push_back(OldInstIter);
        break;
      }
    }

    // Handle extractelement instruction which is following the load
    for (auto *U : OldInstIter->users()) {
      if (auto *EEI = dyn_cast<ExtractElementInst>(U)) {
        VectorType *OldVecTy = getVectorType(Inst->getType());
        uint64_t NumElemsInOldVec = OldVecTy->getElementCount().getValue();
        uint64_t NumElemsInNewVec = NewVecTy->getElementCount().getValue();
        uint64_t ElemIdx =
            cast<ConstantInt>(EEI->getIndexOperand())->getSExtValue() /
                (NumElemsInOldVec / NumElemsInNewVec);
        NewValue = Builder.CreateExtractElement(NewValue, ElemIdx);
        NewValue = Builder.CreateTrunc(NewValue, OldVecTy->getElementType());

        OldInstIter = EEI;
        InstsToErase.push_back(OldInstIter);
        break;
      }
    }

    OldInstIter->replaceAllUsesWith(NewValue);
    return true;
  }
};
```

mlychkov: I also think that this function may be not a method because there are no dependencies on SPIRVLowerBitCastToNonStandardTypePass class.

lib/SPIRV/SPIRVLowerBitCastToNonStandardType.cpp
```diff
@@ -0,0 +1,205 @@
...
+
+    std::vector<Instruction *> BCastsToNonStdVec;
+    std::vector<Instruction *> InstsToErase;
+    for (auto &BB : F)
+      for (auto &I : BB) {
+        auto *BC = dyn_cast<BitCastInst>(&I);
+        if (!BC)
+          continue;
+        auto *SrcTy = BC->getSrcTy();
+        if (auto *SrcElemTy = dyn_cast<PointerType>(SrcTy))
```
mlychkov, line 141: Please use getVectorType call here.

lib/SPIRV/SPIRVLowerBitCastToNonStandardType.cpp
```diff
@@ -0,0 +1,205 @@
...
+          uint64_t NumElemsInDestVec = DestVecTy->getElementCount().getValue();
+          if (!isValidVectorSize(NumElemsInDestVec))
+            BCastsToNonStdVec.push_back(&I);
+        }
+      }
+
+    for (auto &I : BCastsToNonStdVec)
+      Changed |= lowerBitCastToNonStdVec(I, InstsToErase);
+
+    for (int I = InstsToErase.size() - 1; I >= 0; I--)
```
mlychkov, line 164: Please use rbegin/rend iterators instead of indexes.

mlychkov (approved):

lib/SPIRV/SPIRVLowerBitCastToNonStandardType.cpp
```diff
@@ -0,0 +1,200 @@
...
+
+#include <utility>
+
+using namespace llvm;
+
+namespace SPIRV {
+
+static VectorType *getVectorType(Type *Ty) {
+  if (!Ty)
+    return nullptr;
```
AlexeySachkov, line 58: Do you really expect `Ty` to be a `nullptr`? Do you have a handling for returned `nullptr` where you call `getVectorType`? I guess that answer to both questions is no and therefore this "if" could be replaced with an assertion that `Ty` is not `nullptr`

lib/SPIRV/SPIRVLowerBitCastToNonStandardType.cpp
```diff
@@ -0,0 +1,200 @@
...
+  if (auto *ElemTy = dyn_cast<PointerType>(Ty))
+    Ty = ElemTy->getElementType();
+  return dyn_cast<VectorType>(Ty);
```
AlexeySachkov, line 61: Do you have a handling on call sites that returned value is not `nullptr`? If not, then I suggest to just use `cast` here to automatically get an assertion if `Ty` is not `VectorType`

lib/SPIRV/SPIRVLowerBitCastToNonStandardType.cpp
```diff
@@ -0,0 +1,200 @@
...
+    if (auto *ASCastInst = dyn_cast<AddrSpaceCastInst>(U)) {
+      unsigned DestAS = ASCastInst->getDestAddressSpace();
+      auto *NewVecPtrTy = NewVecTy->getPointerTo(DestAS);
+      // AddrSpaceCast is created explicitly instead of using method
+      // IRBuilder<>.CreateAddrSpaceCast because IRBuilder doesn't create
+      // separate instruction for constant values. Whereas SPIR-V translator
+      // doesn't like several nested instructions in one.
+      NewValue = new AddrSpaceCastInst(NewValue, NewVecPtrTy);
+      Builder.Insert(NewValue);
+      OldInstIter = ASCastInst;
```
AlexeySachkov, line 88: What if there are several addressspace casts which use the same instruction?

lib/SPIRV/SPIRVLowerBitCastToNonStandardType.cpp
```diff
@@ -0,0 +1,200 @@
...
+      OldInstIter = EEI;
+      InstsToErase.push_back(OldInstIter);
+      break;
+    }
+  }
+  OldInstIter->replaceAllUsesWith(NewValue);
+  return true;
+}
+
+class SPIRVLowerBitCastToNonStandardTypePass {
```
AlexeySachkov, line 122: If I understand correctly, to make it a proper pass you must [inherit from `PassInfoMixin`](https://llvm.org/docs/WritingAnLLVMNewPMPass.html#basic-code-required)

lib/SPIRV/SPIRVLowerBitCastToNonStandardType.cpp
```diff
@@ -0,0 +1,200 @@
...
+    for (auto I = InstsToErase.rbegin(); I != InstsToErase.rend(); ++I)
+      (*I)->eraseFromParent();
+
+    return Changed ? PreservedAnalyses::none() : PreservedAnalyses::all();
+  }
+};
+
+class SPIRVLowerBitCastToNonStandardTypeLegacy
+    : public FunctionPass,
+      public SPIRVLowerBitCastToNonStandardTypePass {
```
AlexeySachkov, line 168:
```suggestion
    : public FunctionPass {
```

Why do you need to inherit from `SPIRVLowerBitCastToNonStandardTypePass` here?

AlexeySachkov (changes requested):

lib/SPIRV/SPIRVLowerBitCastToNonStandardType.cpp
KornevNikita: Using `cast` causes many LIT tests failures. Because when we are looking for bitcast to non-standard vector type instructions, we need to check that source vector type is valid, so we calling `VectorType *SrcVecTy = getVectorType(BC->getSrcTy())` to get this vector. But there are many tests with bitcast instructions, which source operands are not vectors, so we are getting failuers in attempting to cast this types. I suggest to keep it as is, because there are `if`'s like this:
`VectorType *SrcVecTy = getVectorType(BC->getSrcTy());`
`if (SrcVecTy) {`
...
` VectorType *DestVecTy = getVectorType(BC->getDestTy());`
`if (DestVecTy) {`
It is assumed that in other cases `getVectorType` cannot return nullptr.

mlychkov: This function is also called when we are searching for a bitcast of vector type value. However, we may look into other bitcasts during this search.

lib/SPIRV/SPIRVLowerBitCastToNonStandardType.cpp
KornevNikita: Added _TODO_ cooment in [587cc7f](https://github.com/KhronosGroup/SPIRV-LLVM-Translator/pull/1117/commits/587cc7f2a52ec020fe24677c46824f4973f58653)

lib/SPIRV/SPIRVLowerBitCastToNonStandardType.cpp
```diff
@@ -0,0 +1,200 @@
...
+/// unsupported types.
+/// lowerBitCastToNonStdVec function is designed to avoid using of bitcast to
+/// unsupported vector types instructions and should be called if similar
+/// instructions have been encountered in input LLVM IR.
+bool lowerBitCastToNonStdVec(Instruction *Inst,
+                             std::vector<Instruction *> &InstsToErase) {
+  Instruction *OldInstIter(Inst);
+  IRBuilder<> Builder(OldInstIter);
+  Value *NewValue = OldInstIter->getOperand(0);
+  VectorType *NewVecTy = getVectorType(NewValue->getType());
```
AlexeySachkov, line 74: `NewVecTy` is not checked for `nullptr` before its use on line 82

AlexeySachkov (approved):
I think that we can merge this as some kind of baseline, i.e. in general algorithm works, but in order for it to be useful in LLVM IR produced from real applications, we must ensure that we resolve `TODO` about multiple uses of non-standard vector types (from `SPIRVLowerBitCastToNonStandardType.cpp:77`)

lib/SPIRV/SPIRVLowerBitCastToNonStandardType.cpp
AlexeySotkin: > What if there are several addressspace casts which use the same instruction?

You mean we don't need `break` at the end of the `if`? If yes, I agree.

mlychkov: In this case whole function should be refactored to process several patterns "BitCast -> AScast -> Load -> ExtractElement" in an IR function.

AlexeySachkov: Yes, we can't just remove the `break`, because the code assumes that there is only one instruction which needs to be fixed

lib/SPIRV/SPIRVLowerBitCastToNonStandardType.cpp
```diff
@@ -51,6 +51,8 @@
 
 using namespace llvm;
 
+typedef std::vector<std::pair<Instruction *, Value *>> InstructionValueVec;
```
mlychkov, line 54:
```suggestion
using InstructionValueVec = std::vector<std::pair<Instruction *, Value *>>;
```

lib/SPIRV/SPIRVLowerBitCastToNonStandardType.cpp
```diff
@@ -68,15 +70,13 @@ static VectorType *getVectorType(Type *Ty) {
...
+  IRBuilder<> Builder(Inst);
+  Value *NewValue = Inst->getOperand(0);
   VectorType *NewVecTy = getVectorType(NewValue->getType());
-  InstsToErase.push_back(OldInstIter);
+  InstsToErase.push_back(Inst);
   // Handle addrspacecast instruction after bitcast if present
-  // TODO: There are can be many addrspacecast instructions using this bitcast,
-  // so that it is necessary to handle all these instructions in the same way
-  for (auto *U : OldInstIter->users()) {
+  InstructionValueVec ASCastInsts;
```
mlychkov, line 78: I'd suggest to rewrite this function to call it recursively for each next instruction in the pattern.

lib/SPIRV/SPIRVLowerBitCastToNonStandardType.cpp
```diff
@@ -0,0 +1,194 @@
...
+/// these types should be replaced in a special way to avoid using of
+/// unsupported types.
+/// lowerBitCastToNonStdVec function is designed to avoid using of bitcast to
+/// unsupported vector types instructions and should be called if similar
+/// instructions have been encountered in input LLVM IR.
+bool lowerBitCastToNonStdVec(Instruction *Inst, Value *NewValue,
+                             VectorType *OldVecTy, VectorType *NewVecTy,
+                             std::vector<Instruction *> &InstsToErase) {
+  bool Changed = false;
+  IRBuilder<> Builder(Inst);
```
mlychkov, line 73: I'd suggest to create Builder object once in runLowerBitCastToNonStandardType function and here just update insert point.

lib/SPIRV/SPIRVLowerBitCastToNonStandardType.cpp
```diff
@@ -0,0 +1,194 @@
...
+  for (auto *U : Inst->users()) {
```
mlychkov, line 74:
```suggestion
  VectorType *NewVecTy = getVectorType(NewInst->getType());
  if (NewVecTy) {
    for (auto *U : Inst->users()) {
```

lib/SPIRV/SPIRVLowerBitCastToNonStandardType.cpp
```diff
@@ -0,0 +1,194 @@
...
+    if (auto *ASCastInst = dyn_cast<AddrSpaceCastInst>(U)) {
+      unsigned DestAS = ASCastInst->getDestAddressSpace();
+      auto *NewVecPtrTy = NewVecTy->getPointerTo(DestAS);
+      // AddrSpaceCast is created explicitly instead of using method
+      // IRBuilder<>.CreateAddrSpaceCast because IRBuilder doesn't create
+      // separate instruction for constant values. Whereas SPIR-V translator
+      // doesn't like several nested instructions in one.
+      Value *LocalValue = new AddrSpaceCastInst(NewValue, NewVecPtrTy);
+      Builder.Insert(LocalValue);
+      InstsToErase.push_back(ASCastInst);
```
mlychkov, line 85:
```suggestion
```

lib/SPIRV/SPIRVLowerBitCastToNonStandardType.cpp
```diff
@@ -0,0 +1,194 @@
...
+      Changed |= lowerBitCastToNonStdVec(ASCastInst, LocalValue, OldVecTy,
+                                         NewVecTy, InstsToErase);
+    }
+    // Handle load instruction which is following the bitcast in the pattern
+    if (auto *LI = dyn_cast<LoadInst>(U)) {
+      Value *LocalValue = Builder.CreateLoad(NewVecTy, NewValue);
+      InstsToErase.push_back(LI);
```
mlychkov, line 92:
```suggestion
```

lib/SPIRV/SPIRVLowerBitCastToNonStandardType.cpp
```diff
@@ -0,0 +1,194 @@
...
+    if (auto *EEI = dyn_cast<ExtractElementInst>(U)) {
+      uint64_t NumElemsInOldVec = OldVecTy->getElementCount().getValue();
+      uint64_t NumElemsInNewVec = NewVecTy->getElementCount().getValue();
+      uint64_t ElemIdx =
+          cast<ConstantInt>(EEI->getIndexOperand())->getSExtValue() /
+          (NumElemsInOldVec / NumElemsInNewVec);
+      Value *LocalValue = Builder.CreateExtractElement(NewValue, ElemIdx);
+      LocalValue = Builder.CreateTrunc(LocalValue, OldVecTy->getElementType());
+      InstsToErase.push_back(EEI);
+      EEI->replaceAllUsesWith(LocalValue);
```
mlychkov, line 106:
```suggestion
      Changed |= lowerBitCastToNonStdVec(EEI, LocalValue, OldVecTy, InstsToErase);
```

lib/SPIRV/SPIRVLowerBitCastToNonStandardType.cpp
```diff
@@ -0,0 +1,194 @@
...
+    }
+  }
+  return Changed;
```
mlychkov, line 109:
```suggestion
  InstsToErase.push_back(OldInst);
  if (!Changed)
    OldInst->replaceAllUsesWith(NewInst);

  return true;
```

lib/SPIRV/SPIRVLowerBitCastToNonStandardType.cpp
```diff
@@ -0,0 +1,194 @@
...
+          uint64_t NumElemsInDestVec = DestVecTy->getElementCount().getValue();
+          if (!isValidVectorSize(NumElemsInDestVec))
+            BCastsToNonStdVec.push_back(&I);
+        }
+      }
+    for (auto &I : BCastsToNonStdVec) {
+      Value *NewValue = I->getOperand(0);
+      VectorType *OldVecTy = getVectorType(I->getType());
+      VectorType *NewVecTy = getVectorType(NewValue->getType());
+      InstsToErase.push_back(I);
```
mlychkov, line 150:
```suggestion
```

lib/SPIRV/SPIRVLowerBitCastToNonStandardType.cpp
```diff
@@ -0,0 +1,194 @@
...
+      Changed |= lowerBitCastToNonStdVec(I, NewValue, OldVecTy, NewVecTy,
+                                         InstsToErase);
+    }
+
+    for (auto I = InstsToErase.rbegin(); I != InstsToErase.rend(); ++I)
+      (*I)->eraseFromParent();
```
mlychkov, line 156:
```suggestion
    for (auto *I : InstsToErase)
      I->eraseFromParent();
```

lib/SPIRV/SPIRVLowerBitCastToNonStandardType.cpp
```diff
@@ -0,0 +1,199 @@
...
+          uint64_t NumElemsInDestVec = DestVecTy->getElementCount().getValue();
+          if (!isValidVectorSize(NumElemsInDestVec))
+            BCastsToNonStdVec.push_back(&I);
+        }
+      }
+    IRBuilder<> Builder(F.getContext());
+    for (auto &I : BCastsToNonStdVec) {
+      Value *NewValue = I->getOperand(0);
+      VectorType *OldVecTy = getVectorType(I->getType());
+      Builder.SetInsertPoint(I);
```
mlychkov, line 155: Could you please move this call into the beginning of lowerBitCastToNonStdVec? Dependent instructions are not guarantied to be consecutive in IR.

lib/SPIRV/SPIRVLowerBitCastToNonStandardType.cpp
```diff
@@ -0,0 +1,199 @@
...
+/// Since SPIR-V does not support non-standard vector types, instructions using
+/// these types should be replaced in a special way to avoid using of
+/// unsupported types.
+/// lowerBitCastToNonStdVec function is designed to avoid using of bitcast to
+/// unsupported vector types instructions and should be called if similar
+/// instructions have been encountered in input LLVM IR.
+bool lowerBitCastToNonStdVec(Instruction *OldInst, Value *NewInst,
+                             VectorType *OldVecTy,
+                             std::vector<Instruction *> &InstsToErase,
+                             IRBuilder<> *Builder) {
```
mlychkov, line 72:
```suggestion
                             IRBuilder<> &Builder) {
```

lib/SPIRV/SPIRVLowerBitCastToNonStandardType.cpp
```diff
@@ -0,0 +1,199 @@
...
+}
+
+/// Since SPIR-V does not support non-standard vector types, instructions using
+/// these types should be replaced in a special way to avoid using of
+/// unsupported types.
+/// lowerBitCastToNonStdVec function is designed to avoid using of bitcast to
+/// unsupported vector types instructions and should be called if similar
+/// instructions have been encountered in input LLVM IR.
+bool lowerBitCastToNonStdVec(Instruction *OldInst, Value *NewInst,
+                             VectorType *OldVecTy,
```
mlychkov, line 70:
```suggestion
                             const VectorType *OldVecTy,
```

test/lower-non-standard-types.ll
```diff
@@ -0,0 +1,58 @@
...
+; CHECK: %conv2 = sitofp i32 %6 to float
+
+; CHECK-NOT: %0 = bitcast <3 x i64> addrspace(1)* @Id to <6 x i32> addrspace(1)*
+; CHECK-NOT: %1 = addrspacecast <6 x i32> addrspace(1)* %0 to <6 x i32> addrspace(4)*
+; CHECK-NOT: %2 = load <6 x i32>, <6 x i32> addrspace(4)* %1, align 32
+; CHECK-NOT: %3 = load <6 x i32>, <6 x i32> addrspace(4)* %1, align 32
+; CHECK-NOT: %4 = extractelement <6 x i32> %2, i32 0
+; CHECK-NOT: %5 = extractelement <6 x i32> %3, i32 4
+; CHECK-NOT: %conv1 = sitofp i32 %4 to float
+; CHECK-NOT: %conv2 = sitofp i32 %5 to float
```
mlychkov, line 21: I think that it is enough to check that there is no "<6 x i32>" string in a result.

test/lower-non-standard-types.ll
```diff
@@ -0,0 +1,58 @@
+; RUN: llvm-as %s -o %t.bc
+; RUN: llvm-spirv -s %t.bc -o - | llvm-dis -o - | FileCheck %s
+
+; CHECK: %0 = addrspacecast <3 x i64> addrspace(1)* @Id to <3 x i64> addrspace(4)*
```
mlychkov, line 4: Please use variables instead of hardcoded identifiers.

test/lower-non-standard-types.ll
```diff
@@ -0,0 +1,51 @@
...
+; CHECK: [[LoadInst1:%.*]] = load <3 x i64>, <3 x i64> addrspace(4)* [[ASCastInst]], align 32
+; CHECK: [[ExtrElInst1:%.*]] = extractelement <3 x i64> [[LoadInst1]], i64 0
+; CHECK: [[TruncInst1:%.*]] = trunc i64 [[ExtrElInst1]] to i32
+; CHECK: [[LoadInst2:%.*]] = load <3 x i64>, <3 x i64> addrspace(4)* [[ASCastInst]], align 32
+; CHECK: [[ExtrElInst2:%.*]] = extractelement <3 x i64> [[LoadInst2]], i64 2
+; CHECK: [[TruncInst2:%.*]] = trunc i64 [[ExtrElInst2]] to i32
+; CHECK: %conv1 = sitofp i32 [[TruncInst1]] to float
+; CHECK: %conv2 = sitofp i32 [[TruncInst2]] to float
+
+; CHECK-NOT: <6 x i32>
```
mlychkov, line 14: Unfortunately, FileCheck won't catch a situation when <6 x i32> is present in an output before CHECK pattern if CHECK-NOT goes after CHECK directives.
You may use FileCheck option --implicit-check-not="<6 x i32>" here to be sure that there is no unsupported vector instructions in a whole output.

mlychkov (approved):
Thank you very much for your patience and willingness to accept constructive remarks.
Now all looks good to me.

KornevNikita:
Clang-tidy warning about recursion can be ignored, because the variable "RecursionDepth" is added to control the depth, so it cannot loop.

AlexeySachkov (approved):
There is still room for improvement, I think, but it looks reliable enough to be merged as initial version

test/lower-non-standard-types.ll
```diff
@@ -0,0 +1,49 @@
+; RUN: llvm-as %s -o %t.bc
+; RUN: llvm-spirv -s %t.bc -o - | llvm-dis -o - | FileCheck %s --implicit-check-not="<6 x i32>"
+
+; CHECK: [[ASCastInst:%.*]] = addrspacecast <3 x i64> addrspace(1)* @Id to <3 x i64> addrspace(4)*
+; CHECK: [[LoadInst1:%.*]] = load <3 x i64>, <3 x i64> addrspace(4)* [[ASCastInst]], align 32
+; CHECK: [[ExtrElInst1:%.*]] = extractelement <3 x i64> [[LoadInst1]], i64 0
+; CHECK: [[TruncInst1:%.*]] = trunc i64 [[ExtrElInst1]] to i32
```
AlexeySotkin, line 7: I'm not sure we are doing semantically correct transformation. Originally we casted to `<6 x i32>` and took 0th and 4th elements:
```
 v       v 
|0|1|2|3|4|5|
```
With this patch we keep it `<3 x i64>` and take 0th and 2nd elements:
```
 v       v 
|0|1|2|3|4|5|
|0  |1  |2  |
```
According to the [documentation](https://llvm.org/docs/LangRef.html#trunc-to-instruction) for `trunc` instruction:
 
> The ‘trunc’ instruction truncates the **high order** bits in value and converts the remaining bits to ty2.

If I understand it correctly we get elements corresponding to 1st and 5th elements of `<6 x i32>`
```
 v       v 
|0|1|2|3|4|5|
|0  |1  |2  |
   ^       ^
```
If this is correct, in some cases we should shift the 64-bit value right for 32 bits before the truncation.

AlexeySotkin (approved):

PR #1119, opened 2021-07-20
Improve llvm.memmove translation

AlexeySachkov:
The main change comparing to previous approach: we do not look for the underlying type which is being copied anymore. Instead, we directly allocate required amount of bytes and copy them.


lib/SPIRV/SPIRVLowerMemmove.cpp
```diff
@@ -54,54 +54,51 @@ namespace SPIRV {
...
-    auto *Alloca = Builder.CreateAlloca(AllocaTy, NumElements);
-    if (Align.hasValue()) {
-      Alloca->setAlignment(Align.getValue());
-    }
+    auto *AllocaTy = ArrayType::get(IntegerType::getInt8Ty(*Context),
+                                    Length->getZExtValue());
+    MaybeAlign SrcAlign = I.getSourceAlign();
+
+    // FIXME: What if @llvm.memmove located within a loop? Wouldn't this be a
+    // huge memory eater?
```
AlexeySotkin, line 82: Perhaps lifetime intrinsics may give a hint to optimizer, so it could re-use the memory.

lib/SPIRV/SPIRVLowerMemmove.cpp
```diff
@@ -54,54 +54,51 @@ namespace SPIRV {
...
   void LowerMemMoveInst(MemMoveInst &I) {
+    // There is no direct equivalent of @llvm.memmove in SPIR-V and the closest
+    // instructions are 'OpMemoryCopy' and 'OpMemoryCopySized'.
+    //
+    // 'OpMemoryCopy' does not accept amount of bytes to copy and infers that
+    // from type which is being copied; also it only allows to copy value of a
+    // particular type to pointer pointing to the same type.
+    //
+    // 'OpMemoryCopySized' is closer to @llvm.memmove, because it actually
+    // copies bytes, but unlike memove it doesn't support overlapping source
```
AlexeySotkin, line 68: SPIR-V spec doesn't explicitly specify whether overlapping is supported or not. I think worth asking for clarification from Khronos,

AlexeySotkin (approved):

svenvh (approved):
LGTM, thanks.

lib/SPIRV/SPIRVLowerMemmove.cpp
```diff
@@ -54,54 +54,51 @@ namespace SPIRV {
 
 class SPIRVLowerMemmoveBase {
 public:
-  SPIRVLowerMemmoveBase() : Context(nullptr), Mod(nullptr) {}
+  SPIRVLowerMemmoveBase() : Context(nullptr) {}
+
   void LowerMemMoveInst(MemMoveInst &I) {
+    // There is no direct equivalent of @llvm.memmove in SPIR-V and the closest
+    // instructions are 'OpMemoryCopy' and 'OpMemoryCopySized'.
```
MrSidims, line 61:
```suggestion
    // instructions are 'OpCopyMemory' and 'OpCopyMemorySized'.
```

same applicable below

MrSidims (approved):
Please fix instructions' names in the comment

lib/SPIRV/SPIRVLowerMemmove.cpp
AlexeySachkov: Thanks for pointing that out, indeed lifetime intrinsics should deal with this. However, as a first argument they accept the allocation size and we do not pass anything there, like the allocation is variable-sized, which I guess is not the case. But the consequences of that are not clear to me

AlexeySachkov: Replaced this FIXME with another one in dbc54dd752c04bb65e8d2e2aedaaac22c247266e

lib/SPIRV/SPIRVLowerMemmove.cpp
AlexeySachkov: Good point, I've updated the comment in dbc54dd752c04bb65e8d2e2aedaaac22c247266e

lib/SPIRV/SPIRVLowerMemmove.cpp
AlexeySachkov: Done in dbc54dd752c04bb65e8d2e2aedaaac22c247266e

lib/SPIRV/SPIRVLowerMemmove.cpp
```diff
@@ -54,54 +54,54 @@ namespace SPIRV {
...
-    }
+    auto *AllocaTy = ArrayType::get(IntegerType::getInt8Ty(*Context),
+                                    Length->getZExtValue());
+    MaybeAlign SrcAlign = I.getSourceAlign();
+
+    auto *Alloca = Builder.CreateAlloca(AllocaTy);
+    if (SrcAlign.hasValue())
+      Alloca->setAlignment(SrcAlign.getValue());
+
+    // FIXME: Do we need to pass the size of alloca here? From LangRef:
```
AlexeySotkin, line 85: I think we do.

AlexeySotkin:
Should we add spirv-val to the test?

lib/SPIRV/SPIRVLowerMemmove.cpp
AlexeySotkin: @AlexeySachkov are you going to fix it in scope of this PR?

AlexeySotkin: To be resolved in scope of https://github.com/KhronosGroup/SPIRV-LLVM-Translator/issues/1130

PR #1124, opened 2021-07-22
Fix SPV-IR for OpImageRead instruction

Nuullll:
Do not translate it directly to OCL. Move the translation logic to
SPIRVToOCL pass.


Nuullll:
Sorry for the noise, duplicated with #1106

PR #1131, opened 2021-07-27
Update minimum cmake version to 3.13.4

svenvh:
7251f3cb ("Make spirv.hpp be an external dependency (#813)",
2021-07-27) introduces use of the FetchContent module, which is only
available from CMake 3.11.

We might just as well align with LLVM's cmake requirement, so bump the
requirement to CMake 3.13.4.


MrSidims (approved):
LGTM
FYI @AGindinson

AGindinson (approved):
Nice catch, thanks!

PR #1133, opened 2021-07-28
[SPV-IR] Fix SPIRV builtin variable/call translation

Nuullll:
This PR fixes the translation of SPIRV builtin variables/calls. The translator
now is able to recognize builtin-call-based SPV-IR input.
  e.g. call  spir_func i64 @_Z33__spirv_BuiltinGlobalInvocationIdi(i32 1)

- LLVM --> SPIRV translation:

  1. We first translate OCL work item builtins to SPV-IR builtin calls in
     OCLToSPIRV. For example:
      @_Z13get_global_idj(i) -> @_Z33__spirv_BuiltinGlobalInvocationIdi(i)

  2. SPV-IR builtin calls are transformed to variables in LLVMToSPIRV. This
     step is shared by all source inputs (SPV-IR, OCL-IR). For example:
      @_Z33__spirv_BuiltinGlobalInvocationIdi(i) ->
        x = load GlobalInvocationId; extract x, i

- SPIRV --> LLVM translation:

  1. We lower SPIRV builtin variables into calls in SPIRVToLLVM so that SPV-IR
     has builtin function calls (instead of variables) by default.

  2. Builtin calls in SPV-IR can be then translated into OpenCL builtin calls,
     in SPIRVToOCL.
     e.g. @_Z33__spirv_BuiltinGlobalInvocationIdi(i) -> @_Z13get_global_idj(i)

  3. SPIRVToOCL is also capable of lower SPIRV builtin variables into calls as
     preprocessing, so that this pass can be used by users independently to
     translate SPV-IR into OCL-IR.

### DIFF
--- a/lib/SPIRV/SPIRVReader.cpp
+++ b/lib/SPIRV/SPIRVReader.cpp
@@ -245,32 +245,6 @@ void SPIRVToLLVM::setCallingConv(CallInst *Call) {
   Call->setCallingConv(F->getCallingConv());
 }
 
-void SPIRVToLLVM::setAttrByCalledFunc(CallInst *Call) {
-  Function *F = Call->getCalledFunction();
-  assert(F);
-  if (F->isIntrinsic()) {
-    return;
-  }
-  Call->setCallingConv(F->getCallingConv());
-  Call->setAttributes(F->getAttributes());
-}
-
-bool SPIRVToLLVM::transOCLBuiltinsFromVariables() {
-  std::vector<GlobalVariable *> WorkList;
-  for (auto I = M->global_begin(), E = M->global_end(); I != E; ++I) {
-    SPIRVBuiltinVariableKind Kind;
-    if (!isSPIRVBuiltinVariable(&(*I), &Kind))
-      continue;
-    if (!transOCLBuiltinFromVariable(&(*I), Kind))
-      return false;
-    WorkList.push_back(&(*I));
-  }
-  for (auto &I : WorkList) {
-    I->eraseFromParent();
-  }
-  return true;
-}
-
 // For integer types shorter than 32 bit, unsigned/signedness can be inferred
 // from zext/sext attribute.
 MDString *SPIRVToLLVM::transOCLKernelArgTypeName(SPIRVFunctionParameter *Arg) {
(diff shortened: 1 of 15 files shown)


Nuullll:
@AlexeySachkov @AlexeySotkin @Fznamznon @MrSidims Please take a look, thanks :)

AlexeySotkin:
@Fznamznon could you please take a look?

lib/SPIRV/SPIRVWriter.cpp
```diff
@@ -3443,6 +3443,9 @@ bool LLVMToSPIRVBase::translate() {
   if (isEmptyLLVMModule(M))
     BM->addCapability(CapabilityLinkage);
 
+  // Transform SPV-IR builtin calls to builtin variables.
+  transWorkItemBuiltinsToVariables(M);
```
Fznamznon, line 3447: This function is also called from OCLToSPIRV pass. I guess it is called twice because built-in calls instead variables are expected for SPIR-V friendly IR as well. So, maybe we make this transformation common and for example do only one call to this function but from some regularization pass which runs for all source languages?

lib/SPIRV/SPIRVReader.cpp
```diff
@@ -3542,7 +3325,14 @@ bool SPIRVToLLVM::translate() {
...
   transGeneratorMD();
-  if (!transOCLBuiltinsFromVariables())
+  // TODO: add an option to control the builtin format in SPV-IR.
+  // The primary format should be function calls:
+  //   e.g. call spir_func i32 @_Z29__spirv_BuiltInGlobalLinearIdv()
+  // The secondary format should be global variables:
+  //   e.g. load i32, i32* @__spirv_BuiltInGlobalLinearId, align 4
+  // If the desired format is global variables, we don't have to lower them
+  // as calls.
+  if (!lowerBuiltinVariablesToCalls(M))
```
Fznamznon, line 3335: While we don't have an option to control output format, do we need another call to `lowerBuiltinVariablesToCalls` from SPIRToOCL passes?

test/transcoding/builtin_calls.ll
```diff
@@ -0,0 +1,36 @@
+; RUN: llvm-as %s -o %t.bc
```
Fznamznon, line 1: We actually have a couple (or more) tests related to built-in variables. For example https://github.com/KhronosGroup/SPIRV-LLVM-Translator/blob/master/test/transcoding/builtin_vars_arithmetics.ll . Could you please also check that round trip translation (i.e. LLVM-SPV-IR-LLVM) now has been fixed and works for these tests?

lib/SPIRV/SPIRVUtil.cpp
```diff
@@ -1639,6 +1639,281 @@ bool checkTypeForSPIRVExtendedInstLowering(IntrinsicInst *II, SPIRVModule *BM) {
...
+}
+
+/// Translates OCL/SPR-IR work-item builtin functions to SPIRV builtin
+/// variables.
+/// Function like get_global_id(i) -> x = load GlobalInvocationId; extract x, i
+/// Function like get_work_dim() -> load WorkDim
+void transWorkItemBuiltinsToVariables(Module *M) {
+  LLVM_DEBUG(dbgs() << "Enter transWorkItemBuiltinsToVariables\n");
+  std::vector<Function *> WorkList;
+  for (auto &I : *M) {
```
AlexeySachkov, line 1863: Why `I` and not `F`? If I understand correctly, here we iterate over functions and not over instructions

lib/SPIRV/SPIRVUtil.cpp
```diff
@@ -1639,6 +1639,281 @@ bool checkTypeForSPIRVExtendedInstLowering(IntrinsicInst *II, SPIRVModule *BM) {
...
+    StringRef DemangledName;
```
AlexeySachkov, line 1864: I suggest to add an early exit if function is a definition, because we do not expect those built-ins to be defined; they should be only declared

lib/SPIRV/SPIRVUtil.cpp
```diff
@@ -1639,6 +1639,281 @@ bool checkTypeForSPIRVExtendedInstLowering(IntrinsicInst *II, SPIRVModule *BM) {
...
+  return true;
+}
+
+/// Translates OCL/SPR-IR work-item builtin functions to SPIRV builtin
+/// variables.
+/// Function like get_global_id(i) -> x = load GlobalInvocationId; extract x, i
+/// Function like get_work_dim() -> load WorkDim
+void transWorkItemBuiltinsToVariables(Module *M) {
+  LLVM_DEBUG(dbgs() << "Enter transWorkItemBuiltinsToVariables\n");
+  std::vector<Function *> WorkList;
```
AlexeySachkov, line 1862:
```suggestion
  std::vector<Function *> ToRemove;
```

lib/SPIRV/SPIRVUtil.cpp
```diff
@@ -1639,6 +1639,281 @@ bool checkTypeForSPIRVExtendedInstLowering(IntrinsicInst *II, SPIRVModule *BM) {
...
+      if (IsVec) {
+        NewValue =
+            ExtractElementInst::Create(NewValue, CI->getArgOperand(0), "", CI);
+        if (DLoc)
+          NewValue->setDebugLoc(DLoc);
+        LLVM_DEBUG(dbgs() << *NewValue << '\n');
+      }
+      NewValue->takeName(CI);
+      CI->replaceAllUsesWith(NewValue);
+      InstList.push_back(CI);
```
AlexeySachkov, line 1905: I guess that if you switch `WorkList` to store `Value *`, then you will be able to re-use it here and reduce the code a little bit by performing a cleanup in a single place

Nuullll:
A gentle ping.

lib/SPIRV/SPIRVWriter.cpp
```diff
@@ -2043,6 +2043,73 @@ bool LLVMToSPIRVBase::transBuiltinSet() {
...
+    if (!SPIRVBuiltInNameMap::rfind(Name.str(), &BVKind))
+      continue;
+    std::string BuiltinVarName =
+        std::string(kSPIRVName::Prefix) + SPIRVBuiltInNameMap::map(BVKind);
+    LLVM_DEBUG(dbgs() << "builtin variable name: " << BuiltinVarName << '\n');
+    bool IsVec = F.getFunctionType()->getNumParams() > 0;
+    Type *GVType =
+        IsVec ? FixedVectorType::get(F.getReturnType(), 3) : F.getReturnType();
+    auto *BV = new GlobalVariable(
+        *M, GVType, true, GlobalValue::ExternalLinkage, nullptr, BuiltinVarName,
```
Fznamznon, line 2081: NIT
```suggestion
        *M, GVType, /*isConstant=*/ true, GlobalValue::ExternalLinkage, nullptr, BuiltinVarName,
```
It would be great to add comments explaining what constant values mean https://llvm.org/docs/CodingStandards.html#comment-formatting .

lib/SPIRV/SPIRVWriter.cpp
```diff
@@ -2043,6 +2043,73 @@ bool LLVMToSPIRVBase::transBuiltinSet() {
...
+        0, GlobalVariable::NotThreadLocal, SPIRAS_Input);
+    for (auto UI = F.user_begin(), UE = F.user_end(); UI != UE; ++UI) {
```
Fznamznon, line 2083: NIT
```suggestion
    for (auto *U : F->users()) {
```

lib/SPIRV/SPIRVWriter.cpp
```diff
@@ -2043,6 +2043,73 @@ bool LLVMToSPIRVBase::transBuiltinSet() {
...
+      CI->replaceAllUsesWith(NewValue);
+      ToRemove.push_back(CI);
+    }
+    ToRemove.push_back(&F);
+  }
+  for (auto *V : ToRemove) {
+    if (auto *I = dyn_cast<Instruction>(V))
+      I->eraseFromParent();
+    else if (auto *F = dyn_cast<Function>(V))
+      F->eraseFromParent();
```
Fznamznon, line 2108: NIT: that would be weird if something git into ToRemove and it is not a call or a function, right?
```suggestion
    else if (auto *F = dyn_cast<Function>(V))
      F->eraseFromParent();
    else
      llvm_unreachable("Unexpected value to remove!");
```

lib/SPIRV/SPIRVWriter.cpp
```diff
@@ -2043,6 +2043,73 @@ bool LLVMToSPIRVBase::transBuiltinSet() {
...
+bool LLVMToSPIRVBase::transWorkItemBuiltinCallsToVariables() {
+  LLVM_DEBUG(dbgs() << "Enter transWorkItemBuiltinCallsToVariables\n");
+  // Store instructions and functions need to be removed.
+  SmallVector<Value *, 16> ToRemove;
+  for (auto &F : *M) {
+    // Builtins should be declaration only.
+    if (!F.isDeclaration())
+      continue;
+    StringRef DemangledName;
+    if (!oclIsBuiltin(F.getName(), DemangledName))
```
Fznamznon, line 2064: I guess it is expected that a function named **ocl**IsBuiltin returns true for SPV-IR as well? Maybe we should rename it?

lib/SPIRV/SPIRVWriter.cpp
```diff
@@ -2043,6 +2043,73 @@ bool LLVMToSPIRVBase::transBuiltinSet() {
...
 
+/// Transforms SPR-IR work-item builtin calls to SPIRV builtin variables.
+/// e.g.
+///  SPR-IR: @_Z33__spirv_BuiltInGlobalInvocationIdi(i)
+///    is transformed as:
+///  x = load GlobalInvocationId; extract x, i
+/// e.g.
+///  SPR-IR: @_Z22__spirv_BuiltInWorkDimi()
+///    is transformed as:
+///  load WorkDim
```
Fznamznon, line 2054: We usually call this representation SPV-IR, right?
```suggestion
/// Transforms SPV-IR work-item builtin calls to SPIRV builtin variables.
/// e.g.
///  SPV-IR: @_Z33__spirv_BuiltInGlobalInvocationIdi(i)
///    is transformed as:
///  x = load GlobalInvocationId; extract x, i
/// e.g.
///  SPV-IR: @_Z22__spirv_BuiltInWorkDim()
///    is transformed as:
///  load WorkDim
```
(also corrected mangling).

lib/SPIRV/SPIRVWriter.cpp
```diff
@@ -2043,6 +2043,73 @@ bool LLVMToSPIRVBase::transBuiltinSet() {
...
+bool LLVMToSPIRVBase::transWorkItemBuiltinCallsToVariables() {
+  LLVM_DEBUG(dbgs() << "Enter transWorkItemBuiltinCallsToVariables\n");
+  // Store instructions and functions need to be removed.
```
Fznamznon, line 2057:
```suggestion
  // Store instructions and functions that need to be removed.
```

lib/SPIRV/SPIRVReader.cpp
Fznamznon: Yes, I understand that, thanks.
I see that `lowerBuiltinVariablesToCalls` is also called at the end of SPIRVReader, I guess it already makes SPIRVReader to produce calls to work item built-ins (instead of variables use). After that all that is left to do in SPIRVToOCL passes is to change built-in names in visitCallSPIRVBuiltin.
But `lowerBuiltinVariablesToCalls` is called again at the beginning of SPIRVToOCL pass. Do we really need this second call, or I just misunderstood something? (Sorry If I did).

MrSidims: I guess a comment on line 3328 "// TODO: add an option to control the builtin format in SPV-IR." explains why `lowerBuiltinVariablesToCalls` is being called twice and foresees a time when it actually won't be a case . BTW, why do we need to call `lowerBuiltinVariablesToCalls` in the reader in the first place? I'd prefer to always have GV-like built-ins in SPV-IR, or I just don't have enough data?

Fznamznon: > I guess a comment on line 3328 "// TODO: add an option to control the builtin format in SPV-IR." explains why lowerBuiltinVariablesToCalls is being called twice and foresees a time when it actually won't be a case . 
Yes, as I already mentioned in my first comment, while we don't have an option, can we make only one call?

> BTW, why do we need to call lowerBuiltinVariablesToCalls in the reader in the first place? I'd prefer to always have GV-like built-ins in SPV-IR
The thing is, we already have users that use calls, so we don't want to break their workflow.

test/transcoding/builtin_calls.ll
Fznamznon: Thank you.

lib/SPIRV/SPIRVReader.cpp
Nuullll: > Yes, as I already mentioned in my first comment, while we don't have an option, can we make only one call?

Sorry I didn't make it clear. The thing is, we have users that use the `SPIRVToOCL` pass independently, translating SPV-IR into OCL-IR. And for these users, the `SPIRVToOCL` pass may receive SPV-IR of both formats (builtin calls / variables) as input. For example, clang is still emitting builtin-variable-based SPV-IR.
So I think this call in `SPIRVToOCL` is probably unavoidable, at least for the current stage :(

lib/SPIRV/SPIRVWriter.cpp
Nuullll: Typo fixed.

lib/SPIRV/SPIRVWriter.cpp
Nuullll: Thanks, done.

lib/SPIRV/SPIRVWriter.cpp
Nuullll: Yes. I think we better do it (renaming `oclIsBuiltin()`) in a separate PR to avoid irrelevant changes.

lib/SPIRV/SPIRVWriter.cpp
```diff
@@ -2043,6 +2043,73 @@ bool LLVMToSPIRVBase::transBuiltinSet() {
...
+    LLVM_DEBUG(dbgs() << "Function demangled name: " << DemangledName << '\n');
+    SPIRVBuiltinVariableKind BVKind;
+    SmallVector<StringRef, 2> Postfix;
+    // Deprefix "__spirv_"
+    StringRef Name = dePrefixSPIRVName(DemangledName, Postfix);
+    // Lookup SPIRV Builtin Variable Kind.
+    if (!SPIRVBuiltInNameMap::rfind(Name.str(), &BVKind))
+      continue;
+    std::string BuiltinVarName =
+        std::string(kSPIRVName::Prefix) + SPIRVBuiltInNameMap::map(BVKind);
```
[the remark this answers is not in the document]
Nuullll: Thanks, done! Eliminated `BVKind` as well!

lib/SPIRV/SPIRVWriter.cpp
Nuullll: Done.

lib/SPIRV/SPIRVWriter.cpp
Nuullll: Thanks, done.

lib/SPIRV/SPIRVWriter.cpp
Nuullll: Done.

lib/SPIRV/SPIRVReader.cpp
Fznamznon: Ah, I understand. Thank you.

Fznamznon (approved):
Thank you!

AlexeySachkov (approved):

PR #1134, opened 2021-07-28
Update LLVM version from 13 to 14

svenvh:


.travis.yml
```diff
@@ -57,11 +57,11 @@ matrix:
   include:
     - os: osx
       env: BUILD_TYPE=Release BUILD_EXTERNAL=0 MAKE_TARGETS="llvm-spirv" MAKE_TEST_TARGET="check-llvm-spirv"
-      osx_image: xcode13
+      osx_image: xcode14
```
AlexeySotkin, line 60: Do we really need to change version of xcode?

svenvh: Good point, there is not even such a thing as xcode14 it seems.  I'll revert this.

PR #1137, opened 2021-07-29
Remove travis CI on llvm_release_120 branch

NikitaRudenkoIntel:


AlexeySotkin:
@NikitaRudenkoIntel should we close this PR?

NikitaRudenkoIntel:
> should we close this PR?
I created this to check how to fix CI on llvm_release_120 branch.
This can be closed, yes.
Though on this branch we have both github testing and travis. Are both of them necessary?

AlexeySotkin:
> Though on this branch we have both github testing and travis. Are both of them necessary?

This is a good point. I think we can disable Travis CI on this branch already. @AlexeySachkov what do you think?

svenvh (approved):
Fine with me, assuming we don't need this for releases (@AlexeySotkin you probably know more about that anyway).

AlexeySachkov:
> > Though on this branch we have both github testing and travis. Are both of them necessary?
> 
> This is a good point. I think we can disable Travis CI on this branch already. @AlexeySachkov what do you think?

GitHub Actions CI seems to be stable/reliable enough so we can remove Travis. We indeed don't yet have a way to publish pre-built binaries from GitHub Actions, but that should be technically doable

PR #1144, opened 2021-08-04
Fix getSPIRVBuiltin assertion failure

Nuullll:
It's possible that the underscore-delimited postfix is not empty when we
do this check in `SPIRVToOCLBase::visitCallInst`. Should return false
instead of reporting an assertion failure.

### DIFF
--- a/lib/SPIRV/SPIRVUtil.cpp
+++ b/lib/SPIRV/SPIRVUtil.cpp
@@ -478,7 +478,8 @@ bool getSPIRVBuiltin(const std::string &OrigName, spv::BuiltIn &B) {
   SmallVector<StringRef, 2> Postfix;
   StringRef R(OrigName);
   R = dePrefixSPIRVName(R, Postfix);
-  assert(Postfix.empty() && "Invalid SPIR-V builtin Name");
+  if (!Postfix.empty())
+    return false;
   return getByName(R.str(), B);
 }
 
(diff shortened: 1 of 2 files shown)


Nuullll:
@AlexeySachkov @AlexeySotkin @Fznamznon @MrSidims 
Please help take a look. This is an emergency regression of #1133 :P
Thank you.

AlexeySachkov (approved):

lib/SPIRV/SPIRVUtil.cpp
```diff
@@ -478,7 +478,8 @@ bool getSPIRVBuiltin(const std::string &OrigName, spv::BuiltIn &B) {
   SmallVector<StringRef, 2> Postfix;
   StringRef R(OrigName);
   R = dePrefixSPIRVName(R, Postfix);
-  assert(Postfix.empty() && "Invalid SPIR-V builtin Name");
+  if (!Postfix.empty())
```
Fznamznon, line 481: I was first confused by this requirement because in SPV IR it is pretty normal that function name has underscores, for example in case SPIR-V instruction may return different types, so the corresponding SPV IR function will have _R{return_type} postfix. But I guess this function is for built-in variables only, so it is ok to expect empty prefix, am I right?

Nuullll: Yes, I think you're right. Only use `getSPIRVBuiltin` when you expect builtin variables, otherwise, the content of `Postfix` might be meaningless.

Fznamznon (approved):
Thank you!

AlexeySotkin:
The tests failed due to problems in CI, which was fixed by https://github.com/KhronosGroup/SPIRV-LLVM-Translator/pull/1134.
@Nuullll could you please rebase your PR, to restart CI tests?

Nuullll:
> The tests failed due to problems in CI, which was fixed by #1134.
> @Nuullll could you please rebase your PR, to restart CI tests?

Sure, done.

PR #1145, opened 2021-08-04
Fix SPV-IR for OpBuildNDRange

Fznamznon:
It is translated to a function with unmangled name
__spirv_BuildNDRange_{1|2|3}D with struct return parameter and array
arguments, since translator only translates it properly to SPIR-V with
this signature. _ND postfix is requred because array arguments are mangled
in the same way, so if there was no postfix, translator would produce
functions with same name for different dimensions.

### DIFF
--- a/lib/SPIRV/SPIRVInternal.h
+++ b/lib/SPIRV/SPIRVInternal.h
@@ -1037,6 +1037,23 @@ bool lowerBuiltinVariableToCall(GlobalVariable *GV,
 // Transform all builtin variables into calls
 bool lowerBuiltinVariablesToCalls(Module *M);
 
+/// \brief Post-process OpenCL or SPIRV builtin function returning struct type.
+///
+/// Some builtin functions are translated to SPIR-V instructions with
+/// struct type result, e.g. NDRange creation functions. Such functions
+/// need to be post-processed to return the struct through sret argument.
+bool postProcessBuiltinReturningStruct(Function *F);
+
+/// \brief Post-process OpenCL or SPIRV builtin function having array argument.
+///
+/// These functions are translated to functions with array type argument
+/// first, then post-processed to have pointer arguments.
+bool postProcessBuiltinWithArrayArguments(Function *F, StringRef DemangledName);
+
+bool postProcessBuiltinsReturningStruct(Module *M, bool IsCpp = false);
+
+bool postProcessBuiltinsWithArrayArguments(Module *M, bool IsCpp = false);
+
 } // namespace SPIRV
 
 #endif // SPIRV_SPIRVINTERNAL_H
--- a/lib/SPIRV/SPIRVToOCL.cpp
+++ b/lib/SPIRV/SPIRVToOCL.cpp
@@ -135,6 +135,10 @@ void SPIRVToOCLBase::visitCallInst(CallInst &CI) {
     visitCallSPIRVAvcINTELInstructionBuiltin(&CI, OC);
     return;
   }
+  if (OC == OpBuildNDRange) {
+    visitCallBuildNDRangeBuiltIn(&CI, OC, DemangledName);
+    return;
+  }
   if (OC == OpGenericCastToPtrExplicit) {
     visitCallGenericCastToPtrExplicitBuiltIn(&CI, OC);
     return;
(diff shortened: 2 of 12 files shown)


Fznamznon:
@aratajew , @Nuullll , @AlexeySotkin , @MrSidims , I'd appreciate if you take a look

lib/SPIRV/SPIRVUtil.cpp
```diff
@@ -1863,6 +1863,97 @@ bool lowerBuiltinVariablesToCalls(Module *M) {
...
+                                          StringRef DemangledName) {
+  LLVM_DEBUG(dbgs() << "[postProcessOCLBuiltinWithArrayArguments] " << *F
+                    << '\n');
+  auto Attrs = F->getAttributes();
+  auto Name = F->getName();
+  mutateFunction(
+      F,
+      [=](CallInst *CI, std::vector<Value *> &Args) {
+        auto FBegin =
+            CI->getParent()->getParent()->begin()->getFirstInsertionPt();
```
MrSidims, line 1907:
```suggestion
            CI->getFunction()->begin()->getFirstInsertionPt();
```

lib/SPIRV/SPIRVWriter.cpp
```diff
@@ -4329,5 +4329,6 @@ bool llvm::regularizeLlvmForSpirv(Module *M, std::string &ErrMsg,
   legacy::PassManager PassMgr;
   addPassesForSPIRV(PassMgr, Opts);
   PassMgr.run(*M);
+
```
MrSidims, line 4332: An artifact was found

lib/SPIRV/SPIRVUtil.cpp
```diff
@@ -1863,6 +1863,97 @@ bool lowerBuiltinVariablesToCalls(Module *M) {
...
+          getOrCreateFunction(M, Type::getVoidTy(*Context), ArgTys, Name);
+      NewF->addParamAttr(0, Attribute::get(*Context,
+                                           Attribute::AttrKind::StructRet,
+                                           F->getReturnType()));
+      NewF->setCallingConv(F->getCallingConv());
+      auto Args = getArguments(CI);
+      Args.insert(Args.begin(), ST->getPointerOperand());
+      auto *NewCI = CallInst::Create(NewF, Args, CI->getName(), CI);
+      NewCI->setCallingConv(CI->getCallingConv());
+      ST->eraseFromParent();
```
MrSidims, line 1889: Lets also drop all references.

MrSidims (commented):
A bunch of comments. I'll need another round to look on the patch, but overall LGTM.

lib/SPIRV/SPIRVInternal.h
```diff
@@ -1037,6 +1037,23 @@ bool lowerBuiltinVariableToCall(GlobalVariable *GV,
 // Transform all builtin variables into calls
 bool lowerBuiltinVariablesToCalls(Module *M);
 
+/// \brief Post-process OpenCL or SPIRV builtin function returning struct type.
+///
+/// Some builtin functions are translated to SPIR-V instructions with
```
MrSidims, line 1042: The comment is a bit confusing. Does it mean, that in SPV-IR we have a function, that have some structure return type and we replace it with a function of void type, or we already have a void function, which `postProcessBuiltinReturningStruct` function modifies to add `sret` attribute?

Fznamznon: Some SPIR-V instructions may have OpTypeStruct as the type of the result. SPIRVReader translates such instructions to functions that return struct via return value. No matter of chosen output IR format (SPV-IR or OCL), `postProcessBuiltinReturningStruct` makes a function returning struct via return value return it via first argument with sret attribute. It is the only way to return a struct from a function that SPIR-V translator accepts when reads input IR, I guess that is because clang emits functions returning struct this way.

lib/SPIRV/SPIRVToOCL.cpp
```diff
@@ -576,6 +580,32 @@ void SPIRVToOCLBase::visitCallSPIRVImageMediaBlockBuiltin(CallInst *CI, Op OC) {
       },
       &Attrs);
 }
+void SPIRVToOCLBase::visitCallBuildNDRangeBuiltIn(CallInst *CI, Op OC,
+                                                  StringRef DemangledName) {
+  AttributeList Attrs = CI->getCalledFunction()->getAttributes();
+  mutateCallInstOCL(
+      M, CI,
+      [=](CallInst *Call, std::vector<Value *> &Args) {
+        auto *GWS = Args[0];
```
Nuullll, line 589: NIT:
- Add an assert.
- Use more readable var names.
```suggestion
        assert(Args.size() == 3);
        auto *GlobalWorkSize = Args[0];
```

lib/SPIRV/SPIRVUtil.cpp
```diff
@@ -1863,6 +1863,97 @@ bool lowerBuiltinVariablesToCalls(Module *M) {
...
 
+bool postProcessBuiltinReturningStruct(Function *F) {
+  Module *M = F->getParent();
+  LLVMContext *Context = &M->getContext();
+  std::string Name = F->getName().str();
+  F->setName(Name + ".old");
+  for (auto *U : F->users()) {
+    if (auto *CI = dyn_cast<CallInst>(U)) {
+      auto *ST = dyn_cast<StoreInst>(*(CI->user_begin()));
+      assert(ST);
```
Nuullll, line 1874: NIT: `cast<>` is equivalent to `dyn_cast<> + assert()`.
```suggestion
      auto *ST = cast<StoreInst>(*(CI->user_begin()));
```

Nuullll (approved):
Mostly LGTM.

lib/SPIRV/SPIRVToOCL.cpp
Fznamznon: Done, thanks

lib/SPIRV/SPIRVUtil.cpp
Fznamznon: Done, thanks

lib/SPIRV/SPIRVUtil.cpp
Fznamznon: Done, thanks

lib/SPIRV/SPIRVWriter.cpp
Fznamznon: Oops, removed :)

lib/SPIRV/SPIRVUtil.cpp
Fznamznon: Done, thanks

MrSidims (approved):

Issue #598, opened 2020-06-24
Incorrect Handling of Kernel Function Calling Another Kernel Function


bashbaug:
See:

https://github.com/KhronosGroup/SPIRV-Registry/issues/74

In summary, the SPIRV-LLVM-Translator is currently generating invalid SPIR-V when one kernel function calls another kernel function, in some cases.  For example:

```c
kernel void func( global uint* dst ) {
    dst[0] = 42;
}

kernel void kern( global uint* dst ) {
    func(dst);
}
```

Compiling with:

```sh
clang -c -cl-std=CL1.2 -target spir64 -emit-llvm -Xclang -finclude-default-header -g0 -O0 call_kernel.cl
llvm-spirv call_kernel.bc -o call_kernel.spv
```

Generates:

```
; SPIR-V
; Version: 1.0
; Generator: Khronos LLVM/SPIR-V Translator; 14
; Bound: 25
; Schema: 0
               OpCapability Addresses
               OpCapability Kernel
               OpCapability Int64
          %1 = OpExtInstImport "OpenCL.std"
               OpMemoryModel Physical64 OpenCL
               OpEntryPoint Kernel %6 "func"
               OpEntryPoint Kernel %16 "kern"
         %22 = OpString "kernel_arg_type.func.uint*,"
         %23 = OpString "kernel_arg_type.kern.uint*,"
               OpSource OpenCL_C 102000
               OpDecorate %24 Alignment 8
         %24 = OpDecorationGroup
               OpGroupDecorate %24 %10 %19
       %uint = OpTypeInt 32 0
      %ulong = OpTypeInt 64 0
    %ulong_0 = OpConstant %ulong 0
    %uint_42 = OpConstant %uint 42
       %void = OpTypeVoid
%_ptr_CrossWorkgroup_uint = OpTypePointer CrossWorkgroup %uint
          %5 = OpTypeFunction %void %_ptr_CrossWorkgroup_uint
%_ptr_Function__ptr_CrossWorkgroup_uint = OpTypePointer Function %_ptr_CrossWorkgroup_uint
          %6 = OpFunction %void DontInline %5
          %7 = OpFunctionParameter %_ptr_CrossWorkgroup_uint
          %8 = OpLabel
         %10 = OpVariable %_ptr_Function__ptr_CrossWorkgroup_uint Function
               OpStore %10 %7 Aligned 8
[...]

PR #1147, opened 2021-08-09
add an entry point wrapper function around functions.

airlied:
SPIR-V spec states:

However the SPIR-V generated from LLVM violates this and spirv-val fails

This is a first attempt to fix that problem, by creating the entrypoint
as a wrapper around the function.

I'm just creating an MR to get some input on this plan.

### DIFF
--- a/lib/SPIRV/SPIRVWriter.cpp
+++ b/lib/SPIRV/SPIRVWriter.cpp
@@ -576,9 +576,7 @@ SPIRVFunction *LLVMToSPIRVBase::transFunctionDecl(Function *F) {
   BF->setFunctionControlMask(transFunctionControlMask(F));
   if (F->hasName())
     BM->setName(BF, F->getName().str());
-  if (isKernel(F))
-    BM->addEntryPoint(ExecutionModelKernel, BF->getId());
-  else if (F->getLinkage() != GlobalValue::InternalLinkage)
+  if (F->getLinkage() != GlobalValue::InternalLinkage)
     BF->setLinkageType(transLinkageType(F));
 
   // Translate OpenCL/SYCL buffer_location metadata if it's attached to the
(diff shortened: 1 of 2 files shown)


MrSidims:
Linked issues:
https://github.com/KhronosGroup/SPIRV-LLVM-Translator/issues/598
https://github.com/KhronosGroup/SPIRV-Registry/issues/74

Requested review from myself as I recently thinking about the issue as well.

I don't have strong objections against modifications in the SPIR-V emitter itself to address the issue, but yet I slightly lean towards another way of solving it: introduction of LLVM pass that should be run before translation, because by the nature of the issue we should align spir LLVM IR module form with the expected SPIR-V module form. Also on LLVM IR transformation level it's much easier to copy function and function parameter attributes, call instruction metadata and etc then 'right before the translation happens'.

airlied:
Where would such an LLVM IR pass live, in LLVM itself?
I don't object to that idea either, I'm going to try and fix this up at least to go two way so the tests pass, but I'm interested in any solution that is deemed acceptable to most.

MrSidims:
>  I suppose we could do this in SPIRV-LLVM-Translator in those passes it adds

Indeed. May be even just a modification of SPIRVRegularizeLLVM

airlied:
One downside I see about doing this in LLVM is I think the wrapper function will need a new name different from the function name, and I'm not sure how that'll interact. I suppose the pass could add an prefix and then the emitter could rip the prefix off. (I'm a bit rusty on how meaningful entrypoint names are from the runtime pov).

MrSidims:
> One downside I see about doing this in LLVM is I think the wrapper function will need a new name different from the function name, and I'm not sure how that'll interact. I suppose the pass could add an prefix and then the emitter could rip the prefix off. (I'm a bit rusty on how meaningful entrypoint names are from the runtime pov).

I was thinking of moving entrypoint/kernel's body to a wrapper and calling this wrapper from entrypoint/kernel. In this case the name of the kernel preserves and we have no issues in OpenCL runtime for both of the approaches.

Now I see that your patch is doing a bit of a different work - it's recreating the entrypoint. But there is a naming issue as well - it uses the same name as the original function. Since we still have it the module we now have two functions with the same name (unless I'm missing something, guess, I should make some experiments with your patch and without it to make sure). May be on SPIR-V level it's fine, since 'Name' isn't something that SPIR-V tools checks (and indeed IDs are more valuable), but during reverse translation from SPIR-V to LLVM IR we will end up with two conflicting functions.

airlied:
> > One downside I see about doing this in LLVM is I think the wrapper function will need a new name different from the function name, and I'm not sure how that'll interact. I suppose the pass could add an prefix and then the emitter could rip the prefix off. (I'm a bit rusty on how meaningful entrypoint names are from the runtime pov).
> 
> I was thinking of moving entrypoint/kernel's body to a wrapper (with a different name) and calling this wrapper from entrypoint/kernel. In this case the name of the kernel preserves and we have no issues in OpenCL (or whatever) runtime for both of the approaches.
> 
> Now I see that your patch is doing a bit of a different work - it's recreating the entrypoint. But there is a naming issue as well - it uses the same name as the original function. Since we still have it in the module we now have two functions with the same name (unless I'm missing something, guess, I should make some experiments with your patch and without it to make sure). May be on SPIR-V level it's fine, since 'Name' isn't something that SPIR-V tools checks (and indeed IDs are more valuable), but during reverse translation from SPIR-V to LLVM IR we will end up with two conflicting functions.


Yes I've started writing the SPIRVReader.cpp code to merge those back into a single LLVM function if it detects the pattern, but it isn't fully working yet (or there might be other bugs in my way).

I just wasn't sure if the name of the entrypoint needed to correspond to anything for the runtime, whereas for linking we have to preserve the function names. If we don't have to preserve the function names then we can rework the entrypoint name, I'll have to experiment a bit more.

MrSidims:
> I just wasn't sure if the name of the entrypoint needed to correspond to anything for the runtime, whereas for linking we have to preserve the function names.

Ah, I mixed up my explanation from a comment above. It does make sense to preserve the Name from runtime perspective, since programmers use it on the OpenCL host side.

airlied:
Okay I've made a second MR with the alternate approach, this is passing just as much of CTS tests in my impl as the other one.

Tests are gonna need some work either way, but let's get happy with the approach first.

kpet:
Should this be closed now that https://github.com/KhronosGroup/SPIRV-LLVM-Translator/pull/1149 was merged?

PR #1148, opened 2021-08-09
Linkage type fix for global variables translation

KornevNikita:
This is a workaround to fix translation of linkage type of "llvm.used" and
"llvm.compiler.used" global variables that should be "appending" instead of
"export".

### DIFF
--- a/test/appending-linkage-type.ll
+++ b/test/appending-linkage-type.ll
@@ -1,0 +1,33 @@
+; RUN: llvm-as < %s -o %t.bc
+; RUN: llvm-spirv %t.bc -o %t.spv -spirv-ext=+all
+; RUN: llvm-spirv -r %t.spv -o %t.rev.bc
+; RUN: llvm-dis %t.rev.bc -o %t.rev.ll
+; RUN: FileCheck %s --input-file %t.rev.ll --check-prefix CHECK-LLVM
+
+; ModuleID = 'appending-linkage-type'
+target datalayout = "e-p:64:64:64-i1:8:8-i8:8:8-i16:16:16-i32:32:32-i64:64:64-f32:32:32-f64:64:64-v16:16:16-v24:32:32-v32:32:32-v48:64:64-v64:64:64-v96:128:128-v128:128:128-v192:256:256-v256:256:256-v512:512:512-v1024:1024:1024"
+target triple = "spir64-unknown-unknown"
+
+%union.int_double = type { i64 }
+
+@_c0 = internal addrspace(1) constant %union.int_double { i64 4607182418800017408 }, align 8 #0
+@._c0.ref = internal constant %union.int_double addrspace(1)* @_c0 #0
+; CHECK-LLVM: @llvm.compiler.used = appending global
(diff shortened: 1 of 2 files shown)


KornevNikita:
@AlexeySotkin @AlexeySachkov @mlychkov @MrSidims please take a look

MrSidims (approved):

AlexeySotkin (approved):

test/appending-linkage-type.ll
```diff
@@ -0,0 +1,33 @@
+; RUN: llvm-as < %s -o %t.bc
+; RUN: llvm-spirv %t.bc -o %t.spv -spirv-ext=+all
```
svenvh, line 2: Is `-spirv-ext=+all` really needed?

svenvh (commented):
LGTM, but one nit about the test.

test/appending-linkage-type.ll
KornevNikita: Yes, otherwise **@llvm.compiler.used** disappears during LLVM-IR -> SPIR-V -> LLVM-IR translation. But in general, we only need to make sure that **@llvm.compiler.used** retains **appending** linkage type after SPIR-V -> LLVM-IR translation, so we can replace this **.ll** test by **.spt**. Should we do it this way?

svenvh: > Yes, otherwise @llvm.compiler.used disappears during LLVM-IR -> SPIR-V -> LLVM-IR translation.

Right, in that case you can perhaps explicitly list the extension(s) that this relies on instead of enabling `all`.

> we can replace this .ll test by .spt. Should we do it this way?

I find .ll files easier to understand, so I prefer keeping the .ll (but that's my personal opinion).

test/appending-linkage-type.ll
```diff
@@ -0,0 +1,33 @@
...
+target triple = "spir64-unknown-unknown"
+
+%union.int_double = type { i64 }
+
+@_c0 = internal addrspace(1) constant %union.int_double { i64 4607182418800017408 }, align 8 #0
+@._c0.ref = internal constant %union.int_double addrspace(1)* @_c0 #0
+; CHECK-LLVM: @llvm.compiler.used = appending global
+@llvm.compiler.used = appending global [1 x i8*] [i8* bitcast (%union.int_double addrspace(1)** @._c0.ref to i8*)] #0
+
+attributes #0 = { "VCGlobalVariable" }
```
AlexeySotkin, line 18: We can translate it to SPIR-V without any extensions if we remove this attribute. We don't need it for the purpose of this test, right?

KornevNikita: Yes, we need it. I tried to remove this attribute and then translate, but @llvm.compiler.used also disappeared. But I found extensions, which are needed to translate this test insted of all.

test/appending-linkage-type.ll
KornevNikita: Done

AlexeySachkov (approved):

PR #1149, opened 2021-08-10
add an entry point wrapper around functions (llvm pass)

airlied:
SPIR-V spec states:

In order to satisfy SPIR-V that entrypoints and functions
must be different, this introduces an entrypoint wrapper around
functions at the LLVM IR level, then fixes up a few things like
naming at the SPIRV translation.


airlied:
This the alternate version of #1147 using an LLVM pass.

airlied:
@MrSidims just wanted to make sure this version is on your radar.

MrSidims:
@airlied yes, I'll take a look later this day

lib/SPIRV/SPIRVRegularizeLLVM.cpp
```diff
@@ -453,6 +460,68 @@ void SPIRVRegularizeLLVMBase::lowerFuncPtr(Module *M) {
...
+  for (auto &F : *M) {
+    if (F.getCallingConv() != CallingConv::SPIR_KERNEL)
+      continue;
+
+    Work.push_back(&F);
+  }
+  for (auto &F: Work) {
+
+    // for declarations just make them into SPIR functions.
+    if (F->empty()) {
```
MrSidims, line 476:
```suggestion
    if (F->isDeclaration()) {
```

lib/SPIRV/SPIRVRegularizeLLVM.cpp
```diff
@@ -453,6 +460,68 @@ void SPIRVRegularizeLLVMBase::lowerFuncPtr(Module *M) {
...
+    IRBuilder<> Builder(CallBB);
+
+    SmallVector<Value *, 1> Args;
+    for (auto AI = WrapFn->arg_begin(), End = WrapFn->arg_end(); AI != End; ++AI) {
+      Value *A = AI;
+      Args.emplace_back(A);
+    }
+    F->setCallingConv(CallingConv::SPIR_FUNC);
+    auto CI = CallInst::Create(F, ArrayRef<Value *>(Args), F->getReturnType()->isVoidTy() ? "" : "call", CallBB);
+    CI->setCallingConv(F->getCallingConv());
```
MrSidims, line 497: I suggest to try [CloneFunction](https://llvm.org/doxygen/CloneFunction_8cpp_source.html#l00283)

lib/SPIRV/SPIRVRegularizeLLVM.cpp
```diff
@@ -453,6 +460,68 @@ void SPIRVRegularizeLLVMBase::lowerFuncPtr(Module *M) {
...
+
+    // for declarations just make them into SPIR functions.
+    if (F->empty()) {
+      F->setCallingConv(CallingConv::SPIR_FUNC);
+      continue;
+    }
+
+    // Otherwise add a wrapper around the function to act as an entry point.
+    FunctionType *FType = F->getFunctionType();
+    std::string wrap_name = "__spirv_entry_" + static_cast<std::string>(F->getName());
```
MrSidims, line 483: I believe the idea should be not to create a new function, that is being used as an entry point, but instead to create a separate spir function, which is a copy of the kernel/entry point and then reuse this function to:
1. replace all function body of the kernel/entry point with a call to this function.
2. call this new function instead of a kernel/entry point, if it was called by another function.

I see that in the writer '__spirv_entry_' is being find and erased, but instead of it we can just have:
__kernel fun1() { // some code }
__kernel fun2() { call fun2 }

be transformed into

__kernel fun1_body() { // some code }
__kernel fun1() { call fun1_body() }    // we don't change a name of a kernel
__kernel fun2() { call fun1_body() }

airlied: @MrSidims I'm not sure how this will work with linking though.
if I have two files with forward declerations I forsee issues

the same name needs to be in entry point and linkage attributes but pointing at two different functions.

if I have SPIR-V output from the above won't I get linkageattributes "fun2" for the function name, and any spir-v trying to link against this will be looking for fun1 linkage. (or am I missing something? happy to hear I am)

lib/SPIRV/SPIRVRegularizeLLVM.cpp
```diff
@@ -453,6 +460,68 @@ void SPIRVRegularizeLLVMBase::lowerFuncPtr(Module *M) {
...
+
+    Work.push_back(&F);
+  }
+  for (auto &F: Work) {
+
+    // for declarations just make them into SPIR functions.
+    if (F->empty()) {
+      F->setCallingConv(CallingConv::SPIR_FUNC);
+      continue;
+    }
```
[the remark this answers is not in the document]
airlied: fixed up locally.

lib/SPIRV/SPIRVRegularizeLLVM.cpp
```diff
@@ -453,6 +460,68 @@ void SPIRVRegularizeLLVMBase::lowerFuncPtr(Module *M) {
...
 
+void SPIRVRegularizeLLVMBase::addKernelEntryPoint(Module *M) {
+  std::vector<Function *> Work;
+
+  // Get a list of all functions that have SPIR kernel calling conv
+  for (auto &F : *M) {
+    if (F.getCallingConv() != CallingConv::SPIR_KERNEL)
+      continue;
+
+    Work.push_back(&F);
```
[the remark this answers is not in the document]
airlied: fixed up locally.

lib/SPIRV/SPIRVRegularizeLLVM.cpp
airlied: I don't think CloneFunction does what I want here, it seems to be for cloning functions between modules, I'm only cloning the decl and adding a call.

lib/SPIRV/SPIRVRegularizeLLVM.cpp
airlied: file a.cl:
__kernel a() { //some code} 

file b.cl
__kernel a(); //fwd decl
__kernel b { // some code, a(), //more code }

Now how would b work out the renamed function a() for linkage here?

airlied: @MrSidims any more info or do you think I'm wrong here.

MrSidims: Sorry, I was on vacation back then and missed your answer. You are right.

lib/SPIRV/SPIRVRegularizeLLVM.cpp
mlychkov, line 479:
```suggestion
    F->setCallingConv(CallingConv::SPIR_FUNC);
    // for declarations just make them into SPIR functions.
    if (F->empty())
      continue;
```

lib/SPIRV/SPIRVRegularizeLLVM.cpp
mlychkov, line 483: It would be good to create const with "__spirv_entry_" value and use it instead of hard code.

lib/SPIRV/SPIRVRegularizeLLVM.cpp
```diff
@@ -453,6 +460,68 @@ void SPIRVRegularizeLLVMBase::lowerFuncPtr(Module *M) {
...
+
+    auto *CallBB = BasicBlock::Create(M->getContext(), "", WrapFn);
+    IRBuilder<> Builder(CallBB);
+
+    SmallVector<Value *, 1> Args;
+    for (auto AI = WrapFn->arg_begin(), End = WrapFn->arg_end(); AI != End; ++AI) {
+      Value *A = AI;
+      Args.emplace_back(A);
+    }
+    F->setCallingConv(CallingConv::SPIR_FUNC);
```
mlychkov, line 495: Calling conv will be set at the beginning of the loop for functions in Work if first suggestion is applied..
```suggestion
```

lib/SPIRV/SPIRVRegularizeLLVM.cpp
```diff
@@ -453,6 +460,68 @@ void SPIRVRegularizeLLVMBase::lowerFuncPtr(Module *M) {
...
+    }
+    WrapFn->setCallingConv(CallingConv::SPIR_KERNEL);
+    WrapFn->setLinkage(llvm::GlobalValue::InternalLinkage);
+
+    Builder.CreateRet(F->getReturnType()->isVoidTy() ? nullptr : CI);
+
+    /* have to find the spir-v metadata for execution mode and transfer it to the wrapper */
+    if (auto NMD = SPIRVMDWalker(*M).getNamedMD(kSPIRVMD::ExecutionMode)) {
+      while (!NMD.atEnd()) {
+	Function *MDF = nullptr;
```
mlychkov, line 515: Please increase indentation.

lib/SPIRV/SPIRVRegularizeLLVM.cpp
```diff
@@ -69,6 +70,11 @@ class SPIRVRegularizeLLVMBase {
   // Lower functions
   bool regularize();
 
+  /// Erase cast inst of function and replace with the function.
+  /// Assuming F is a SPIR-V builtin function with op code \param OC.
+  //  void addKernelEntryPoint(Function *F, Op OC);
```
mlychkov, line 75: Is this comment relevant?

Fznamznon:
@airlied , there was not that much of a progress for about a month. Are you going to finish this? In case you need some help, I'm ready to help.

lib/SPIRV/SPIRVRegularizeLLVM.cpp
```diff
@@ -453,6 +460,68 @@ void SPIRVRegularizeLLVMBase::lowerFuncPtr(Module *M) {
     lowerFuncPtr(I.first, I.second);
 }
 
+void SPIRVRegularizeLLVMBase::addKernelEntryPoint(Module *M) {
+  std::vector<Function *> Work;
+
+  // Get a list of all functions that have SPIR kernel calling conv
```
Fznamznon, line 466: Should we process only called kernels? I believe that the kernels which are not called might not need this transformation.

airlied: can we tell that, especially for linkages, if something is called from another module?

lib/SPIRV/SPIRVRegularizeLLVM.cpp
airlied: done

lib/SPIRV/SPIRVRegularizeLLVM.cpp
airlied: oops cut-n-paste fail, wrote a proper constant.

airlied:
@Fznamznon thanks for the offer, I was waiting to see everyone was happy with the approach taken. I think I've resolved all the open review comments, however I think the big work is going to be fixing up the tests.

Happy to accept any help in getting it over the line.

airlied:
So this does leave the question should this do something different on the reader side?

airlied:
I've added a reader side patch, and I've fixed up a bunch of tests that were expecting EntryPoint functions to have contents and matched on Name instead. There are still some tests failing, I'll have to dig some more into them.

airlied:
Okay making this non-draft, I've fixed up all the tests to deal with the new entrypoint vs function stuff. Those patches should be reviewed by someone who understands what those tests are meant to do, in case I did something wrong.

lib/SPIRV/SPIRVRegularizeLLVM.cpp
Fznamznon: I wasn't sure that OpenCL actually allows to call the kernels from other translation units. Reading comments to https://github.com/KhronosGroup/SPIRV-Registry/issues/74 , seems this might be the case. If someone really links on SPIR-V level (and I guess it is also possible) we should really wrap them all...

lib/SPIRV/SPIRVWriter.cpp
```diff
@@ -596,11 +596,16 @@ SPIRVFunction *LLVMToSPIRVBase::transFunctionDecl(Function *F) {
...
       static_cast<SPIRVFunction *>(mapValue(F, BM->addFunction(BFT)));
   BF->setFunctionControlMask(transFunctionControlMask(F));
-  if (F->hasName())
-    BM->setName(BF, F->getName().str());
-  if (isKernel(F))
-    BM->addEntryPoint(ExecutionModelKernel, BF->getId());
-  else if (F->getLinkage() != GlobalValue::InternalLinkage)
+  if (F->hasName()) {
+    if (isKernel(F)) {
+      /* strip the prefix as the runtime will be looking for this name */
```
Fznamznon, line 601: So, we get two functions with the same name in resulting SPIR-V, right? Maybe it will be better if they will have different names and the original kernel that contains body will have the prefix? I wonder if it may confuse some third party SPIR-V consumers.

lib/SPIRV/SPIRVReader.cpp
```diff
@@ -2681,6 +2681,26 @@ Function *SPIRVToLLVM::transFunction(SPIRVFunction *BF) {
     return Loc->second;
 
   auto IsKernel = isKernel(BF);
+
+  if (IsKernel) {
```
Fznamznon, line 2685: Maybe we don't need this step at all? I guess inline optimizations should be able to deal with it.

test/mem2reg.cl
```diff
@@ -1,10 +1,11 @@
 // RUN: %clang_cc1 -O0 -S -triple spir-unknown-unknown -cl-std=CL2.0 -x cl -disable-O0-optnone %s -emit-llvm-bc -o %t.bc
 // RUN: llvm-spirv -s %t.bc
-// RUN: llvm-dis < %t.bc | FileCheck %s --check-prefixes=CHECK,CHECK-WO
+// RUN: llvm-dis < %t.bc | FileCheck %s --check-prefixes=CHECK-WO
 // RUN: llvm-spirv -s -spirv-mem2reg %t.bc -o %t.opt.bc
-// RUN: llvm-dis < %t.opt.bc | FileCheck %s --check-prefixes=CHECK,CHECK-W
-// CHECK-LABEL: spir_kernel void @foo
+// RUN: llvm-dis < %t.opt.bc | FileCheck %s --check-prefixes=CHECK-W
+// CHECK-W-LABEL: spir_func void @foo
```
Fznamznon, line 6: Sorry, but I don't see why the calling convention have changed with mem2reg enabled. Could you please elaborate?

lib/SPIRV/SPIRVRegularizeLLVM.cpp
```diff
@@ -529,6 +536,74 @@ void SPIRVRegularizeLLVMBase::lowerFuncPtr(Module *M) {
...
+    auto *CallBB = BasicBlock::Create(M->getContext(), "", WrapFn);
+    IRBuilder<> Builder(CallBB);
+
+    Function::arg_iterator DestI = WrapFn->arg_begin();
+    for (const Argument &I : F->args()) {
+      DestI->setName(I.getName());
+      DestI++;
+    }
+    SmallVector<Value *, 1> Args;
+    for (auto AI = WrapFn->arg_begin(), End = WrapFn->arg_end(); AI != End;
```
Fznamznon, line 570: Why don't we use range-based for loop here as well?

lib/SPIRV/SPIRVRegularizeLLVM.cpp
```diff
@@ -529,6 +536,74 @@ void SPIRVRegularizeLLVMBase::lowerFuncPtr(Module *M) {
...
+         ++AI) {
+      Value *A = AI;
+      Args.emplace_back(A);
+    }
+    auto *CI =
+        CallInst::Create(F, ArrayRef<Value *>(Args),
+                         F->getReturnType()->isVoidTy() ? "" : "call", CallBB);
```
Fznamznon, line 577: I think kernels should always have void return type.

lib/SPIRV/SPIRVRegularizeLLVM.cpp
```diff
@@ -529,6 +536,74 @@ void SPIRVRegularizeLLVMBase::lowerFuncPtr(Module *M) {
...
+    for (auto MD = MDs.begin(), End = MDs.end(); MD != End; ++MD) {
+      WrapFn->addMetadata(MD->first, *MD->second);
+    }
+    WrapFn->setCallingConv(CallingConv::SPIR_KERNEL);
+    WrapFn->setLinkage(llvm::GlobalValue::InternalLinkage);
+
+    Builder.CreateRet(F->getReturnType()->isVoidTy() ? nullptr : CI);
+
+    /* have to find the spir-v metadata for execution mode and transfer it to
+       the wrapper */
```
Fznamznon, line 594: Just a NIT
```suggestion
   // Have to find the spir-v metadata for execution mode and transfer it to
   // the wrapper.
```

lib/SPIRV/SPIRVWriter.cpp
airlied: I can't figure out a way that is compatible with consumers of this library and has separate name. You can't change the non-kernel function because it breaks external linkage, and you can't change the kernel entrypoint name because it breaks runtime consumers of the translator which expect the entrypoints to have the name.

lib/SPIRV/SPIRVReader.cpp
airlied: At least for the SPIRV-LLVM-Translator tests we need it, since they don't get opt passes.

test/mem2reg.cl
airlied: The mem2reg test may not be testing what we expect, but the first pass just looks the raw clang outputted LLVM IR before it's been processed by regularize AFAICS, or llvm-spirv -s input.bc doesn't do what the test author things.

The fix is the first IR we test is from clang and the second is post regularizing, which suggests to me the test might actually be wrong in the first case.

lib/SPIRV/SPIRVWriter.cpp
jenatali: > You can't change the non-kernel function because it breaks external linkage

I'm not following why this is the case. If you have module A declaring a kernel, and module B calling it, if you do the same name change to both, wouldn't that then link correctly?

lib/SPIRV/SPIRVRegularizeLLVM.cpp
airlied: sorry C++ isn't my native language sometimes, I'm used a range base loop here and in the reader side patch as well, thanks for the suggestion.

lib/SPIRV/SPIRVRegularizeLLVM.cpp
airlied: indeed they do, fixed that.

lib/SPIRV/SPIRVRegularizeLLVM.cpp
airlied: fixed.

lib/SPIRV/SPIRVWriter.cpp
airlied: what happens if you generate SPIR-V from something else to link in?

jenatali: Is that a valid way to construct an OpenCL program at the API? I know we do that with the libclc helpers, but since none of those are kernels, they wouldn't be affected, would they? Unless (I didn't really read the code) this was applying a name change to all non-kernel functions, rather than just kernel function wrappers?

airlied: I think it's legal to declare a __kernel in one file, and a non-kernel with the same name as a fwd decl in another file, and expect call to work between them, but I'd have to confirm that.

We don't rename all functions. So linkage from one file that doesn't have __kernel on it won't know to rename the underlying function.

jenatali: SPIR embeds different calling conventions for kernels vs functions. I don't think it's valid to forward-declare a function without `__kernel` and call it that way, and define it in another file with `__kernel`. That'll produce a calling convention mismatch in SPIR, and it can generate a name mismatch in SPIR-V.

airlied: you can create multiple programs some with CreateProgramWithIL, with SPIR-V not generated from SPIRV-LLVM-Translator, you can link those against CreateProgramWithSource that do go through this. I think the linkage attributes would have to be correct for that to function properly.

Fznamznon: > I think it's legal to declare a __kernel in one file, and a non-kernel with the same name as a fwd decl in another file, and expect call to work between them, but I'd have to confirm that.

I can't find any info about it in the OpenCL spec where it defines __kernel attribute.
Though, I tend to agree with @jenatali , that in general it shouldn't be allowed. I've tried compiling a simple OCL example:
```
// source 1
void foo(global int *_B);     
void kernel A(global int *B) {
  foo(B);                     
}
// source 2
void kernel foo(global int *B) {
  B[0] = 1;
}                        
```
I compiled both these files to LLVM IR using `clang` and then linked them using `llvm-link`. In the resulting LLVM IR `foo` is defined as `spir_kernel`, but it is called with `spir_func` calling convention (it changes to `spir_kernel` if I add `kernel` attribute to the declaration of `foo` in source1):
```
; Function Attrs: mustprogress nofree norecurse nosync nounwind willreturn writeonly
define dso_local spir_kernel void @foo(i32 addrspace(1)* nocapture %B) local_unnamed_addr #0 !kernel_arg_addr_space !3 !kernel_arg_access_qual !4 !kernel_
arg_type !5 !kernel_arg_base_type !5 !kernel_arg_type_qual !6 {
entry:
  store i32 1, i32 addrspace(1)* %B, align 4, !tbaa !7
  ret void
}

; Function Attrs: convergent norecurse nounwind
define dso_local spir_kernel void @A(i32 addrspace(1)* %B) local_unnamed_addr #1 !kernel_arg_addr_space !3 !kernel_arg_access_qual !4 !kernel_arg_type !5
!kernel_arg_base_type !5 !kernel_arg_type_qual !6 {
entry:
  call spir_func void @foo(i32 addrspace(1)* %B) #2
  ret void
}
```
Even though LLVM verifier doesn't complain about this, LLVM langref says that:

> The calling convention of the call must match the calling convention of the target function, or else the behavior is undefined.

I don't think that UB case should be somehow allowed or taken into account.
It seems unfortunately this can't be diagnosed by front-ends but maybe it can be reported by linker.
From the other hand SPIR-V spec says:
```
The Linkage Attributes Decoration must not be applied to functions targeted by an OpEntryPoint instruction.
```
After that I wonder if it is meant to call a kernel from a separate translation unit at all. (@bashbaug , I'd appreciate if you could help here).

test/mem2reg.cl
Fznamznon: Yes, the test seems weird. I really didn't notice that the first check really checks LLVM IR from clang. Now I get why the change has happened, since in your changes spir_func calling convention is applied to the original kernel. Thanks.

lib/SPIRV/SPIRVWriter.cpp
jenatali: See https://github.com/KhronosGroup/SPIRV-Registry/issues/74: This is required by OpenCL, but disallowed by SPIR-V, which is the whole point of this change.

The only question in my mind at this point is, can you take one `clProgram` which is source (created from `CreateProgramWithSource`), and expect to link it with another created with `CreateProgramWithIL`. I'd argue no, but I don't see anything explicitly saying yes or no in the CL spec. That probably needs a separate issue on https://github.com/KhronosGroup/OpenCL-Registry.

bashbaug: > After that I wonder if it is meant to call a kernel from a separate translation unit at all.

Good question.  I'm guessing this is a corner case nobody has encountered / considered before.  I can see how this would be complicated given the SPIR-V constraints.  Should this scenario be dis-allowed?  Would that enable this PR to move forward?

> The only question in my mind at this point is, can you take one clProgram which is source (created from CreateProgramWithSource), and expect to link it with another created with CreateProgramWithIL. I'd argue no, but I don't see anything explicitly saying yes or no in the CL spec.

This is interesting too, because I'd argue that this should be supported.  😄 

Rationale is: the program object created from source or IL must be compiled (via `clCompileProgram`) before it can be linked, which transitions the state of the program object from `BINARY_TYPE_NONE` to `BINARY_TYPE_COMPILED_OBJECT`.  After the program object has been compiled I can't think of any reason to differentiate between program objects that originated from source and program objects that originate from IL, so the linking should "just work".  Is there a detail I'm missing here?

I agree both of these topics should have separate issues for more discussion but they should be filed in in [OpenCL-Docs](https://github.com/KhronosGroup/OpenCL-Docs/), not [OpenCL-Registry](https://github.com/KhronosGroup/OpenCL-Registry).

jenatali: > After the program object has been compiled I can't think of any reason to differentiate between program objects that originated from source and program objects that originate from IL, so the linking should "just work". Is there a detail I'm missing here?

C++ compilers have similar problems here, where a "compiled object" from one version of a compiler doesn't necessarily have the same ABI as a "compiled object" from another version of a compiler, and therefore the two objects can't necessarily be linked. This is similar, if we treat an IL program as an object from one version of a compiler, which might not be the same as the platform's source compiler.

> Should this scenario be dis-allowed? Would that enable this PR to move forward?

I don't think this is strictly necessary, what airlied has here now (two functions with the same name, one exported, one as a kernel) is probably fine, and might even be desirable since it round-trips better. While writing this comment, I've come around to this POV, but I still think it's good to probe on what the expected behavior here is - and if it is to be explicitly required, it absolutely should be tested by the CTS.

airlied: I'm pretty sure it's allowed to have IL and non-IL mixed in one path. So I think this should remain as-is.

Fznamznon:
@MrSidims , could you please help with review while I'm on vacation?

MrSidims (commented):
Functionally the current approach is LGTM. Please add a stand-alone test for it. Also, if you don't mind, I'll open a PR in https://github.com/intel/llvm/pulls with this patch just to run some E2E testing.

lib/SPIRV/SPIRVRegularizeLLVM.cpp
```diff
@@ -529,6 +536,70 @@ void SPIRVRegularizeLLVMBase::lowerFuncPtr(Module *M) {
...
+void SPIRVRegularizeLLVMBase::addKernelEntryPoint(Module *M) {
+  std::vector<Function *> Work;
+
+  // Get a list of all functions that have SPIR kernel calling conv
+  for (auto &F : *M) {
+    if (F.getCallingConv() == CallingConv::SPIR_KERNEL)
+      Work.push_back(&F);
+  }
+  for (auto &F : Work) {
+
```
MrSidims, line 548: Code-style nit: please remove this line

lib/SPIRV/SPIRVWriter.cpp
bashbaug: Just FYI, I've created two OpenCL-Docs issues based on the above discussion:

* https://github.com/KhronosGroup/OpenCL-Docs/issues/675
* https://github.com/KhronosGroup/OpenCL-Docs/issues/676

MrSidims:
Also created https://github.com/intel/llvm/pull/4602 to make some E2E testing

airlied:
I can't see how these test fails could be related to this MR.

MrSidims:
They are failing in other PRs as well

airlied:
is there anything preventing this from landing? @MrSidims @Fznamznon

MrSidims:
I'd like to hear opinion from ARM folks as well. @svenvh @StuartDBrady

MrSidims (approved):
Sorry for a long waiting, I'll resolve merge conflicts and restart testing. I've also restarted testing in https://github.com/intel/llvm/pull/4602 to see some end-to-end results.

airlied:
@MrSidims where is this up to? I'd really like to see it land.

MrSidims:
@airlied could you please address LIT failures. They are looking fishy for me:
readonly.ll
```
CHECK-LLVM: expected string not found in input
; CHECK-LLVM: spir_kernel void @_ZTSZ4mainE15kernel_function(i32 addrspace(1)* readonly %{{.*}})
              ^
<stdin>:1:1: note: scanning from here
; ModuleID = '<stdin>'
^
<stdin>:13:7: note: possible intended match here
 call spir_func void @_ZTSZ4mainE15kernel_function(i32 addrspace(1)* readonly %_arg_) #0

Input was: 
define spir_kernel void @__spirv_entry__ZTSZ4mainE15kernel_function(i32 addrspace(1)* readonly %_arg_) #0 !kernel_arg_addr_space !6 !kernel_arg_access_qual !7 !kernel_arg_type !8 !kernel_arg_type_qual !9 !kernel_arg_base_type !8 { 
```
so we end up with the different kernel name. Shouldn't it cause an issue with OpenCL runtime making `clCreateKernel` to pass `CL_​INVALID_​KERNEL_​NAME` ?

airlied:
looking into them now, it's frustrating that this MR is allowed to go stale enough that I have to fully reengineer it each time.

airlied:
@MrSidims okay got it fixed, was one line conflict with the entrypoint spirv-1.6 changes. hopefully all the tests pass now.

MrSidims:
> @MrSidims okay got it fixed, was one line conflict with the entrypoint spirv-1.6 changes. hopefully all the tests pass now.

Thanks! Will merge it by EOD (unless someone will have strong lastminute objections).
Sorry for taking it so long. It's a bit dangerous change, so we had to double check everything (and I was a bit paranoid) before merging this.

lib/SPIRV/SPIRVRegularizeLLVM.cpp
```diff
@@ -605,6 +612,69 @@ void SPIRVRegularizeLLVMBase::lowerFuncPtr(Module *M) {
...
 
+void SPIRVRegularizeLLVMBase::addKernelEntryPoint(Module *M) {
+  std::vector<Function *> Work;
+
+  // Get a list of all functions that have SPIR kernel calling conv
+  for (auto &F : *M) {
+    if (F.getCallingConv() == CallingConv::SPIR_KERNEL)
+      Work.push_back(&F);
+  }
+  for (auto &F : Work) {
```
mlychkov, line 623: Sorry, I've just thought about this little improvement. It's just a nit:
```suggestion
  for (auto &F : *M) {
    // Skip functions that have not a SPIR kernel calling conv
    if (F.getCallingConv() != CallingConv::SPIR_KERNEL)
      continue;
```

MrSidims: I'll apply this comment after the original request is merged.

svenvh: Not sure if that's a valid suggestion, as the loop may modify the Module's function list (using `getOrCreateFunction`) while its being iterated over.

StuartDBrady:
Now that this change has been merged, can KhronosGroup/SPIRV-Registry#74 be closed?

PR #1152, opened 2021-08-12
Remove sycldevice target triple from tests

KornevNikita:
This is a patch to update tests in scope of patch
intel/llvm#3929 that removes sycldevice triple
environment component, so this triple should be also removed from tests.


mlychkov:
As I see there are about 46 test files with sycldevice target triple component:
```
$ grep -rnI sycldevice SPIRV-LLVM-Translator | wc -l
46
```

Сould you please also write short description in commit message clarifying why those changes are done in addition to link on intel/llvm review?

KornevNikita:
@mlychkov thanks for correction

MrSidims (commented):
Since https://github.com/intel/llvm/pull/3929 remove the triple preserving backward compatibility can we also leave (or add) one test with legacy triple?

MrSidims (approved):

mlychkov (approved):

PR #1160, opened 2021-08-17
Fix leak in LLVMToSPIRV::transExecutionMode()

enaydano:
In case spv::ExecutionModeSharedLocalMemorySizeINTEL the new SPIRVExecutionMode was not registered via SPIRVModule::add() and was leaking


MrSidims:
Good catch, thanks!

@enaydano please sign CLA. Also could you please add a test for this bug fix? May be we should modify some tests from test/transcoding/SPV_INTEL_vector_compute directory, @NikitaRudenkoIntel could you please assist here?

AlexeySotkin (approved):

PR #1162, opened 2021-08-18
Update for LLVM addRetAttr API change

svenvh:
Fix build after LLVM commit `3f4d00bc3b51 ("[NFC] More
get/removeAttribute() cleanup", 2021-08-17)`.


Fznamznon (approved):
I'm not sure why out-of-tree build still uses some old version of LLVM, but the fix is LGTM

svenvh:
> I'm not sure why out-of-tree build still uses some old version of LLVM, but the fix is LGTM

The out-of-tree builds use packages from apt.llvm.org that are updated only daily or so:
```Bionic LTS (18.04) - Last update : Tue, 17 Aug 2021 19:05:06 UTC / Revision: 20210817091414+1689dade4218```
That revision didn't have the API-change commit yet, hence those builds fail.

MrSidims (approved):

PR #1163, opened 2021-08-18
Fix production of OpPtrCastToGeneric instruction

Fznamznon:
In LLVM it is valid if addrspace cast result pointer has not only other
address space, but also other element type. However spec states that
for Storage Class changing instructions Result Type and Pointer must point
to the same type. So, this patch adds a regularization step that adds
an additional bitcast in case address space cast changes pointer element
type as well and then everything can be easily translated to valid
SPIR-V that is accepted by spirv-val.

### DIFF
--- a/lib/SPIRV/SPIRVRegularizeLLVM.cpp
+++ b/lib/SPIRV/SPIRVRegularizeLLVM.cpp
@@ -351,6 +351,8 @@ bool SPIRVRegularizeLLVMBase::regularize() {
             II.setMetadata(MDName, nullptr);
           }
         }
+        // Add an additional bitcast in case address space cast also changes
+        // pointer element type.
         if (auto *ASCast = dyn_cast<AddrSpaceCastInst>(&II)) {
           Type *DestTy = ASCast->getDestTy();
           Type *SrcTy = ASCast->getSrcTy();
--- a/lib/SPIRV/SPIRVRegularizeLLVM.cpp
+++ b/lib/SPIRV/SPIRVRegularizeLLVM.cpp
@@ -351,6 +351,23 @@ bool SPIRVRegularizeLLVMBase::regularize() {
             II.setMetadata(MDName, nullptr);
           }
         }
+        if (auto *ASCast = dyn_cast<AddrSpaceCastInst>(&II)) {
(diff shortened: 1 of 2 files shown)


lib/SPIRV/SPIRVRegularizeLLVM.cpp
```diff
@@ -351,6 +351,23 @@ bool SPIRVRegularizeLLVMBase::regularize() {
...
+              SrcTy->getPointerElementType()) {
+            PointerType *InterTy =
+                PointerType::get(DestTy->getPointerElementType(),
+                                 SrcTy->getPointerAddressSpace());
+            BitCastInst *NewBCast = new BitCastInst(
+                ASCast->getPointerOperand(), InterTy, /*NameStr=*/"", ASCast);
+            AddrSpaceCastInst *NewASCast =
+                new AddrSpaceCastInst(NewBCast, DestTy, /*NameStr=*/"", ASCast);
+            ToErase.push_back(ASCast);
+            ASCast->dropAllReferences();
```
MrSidims, line 367: TBH I feel like dropAllReferences should be even applied to the content of ToErase vector itself, but it goes out of a scope of this patch.

MrSidims (approved):
LGTM, thanks!

lib/SPIRV/SPIRVRegularizeLLVM.cpp
```diff
@@ -351,6 +351,23 @@ bool SPIRVRegularizeLLVMBase::regularize() {
             II.setMetadata(MDName, nullptr);
           }
         }
+        if (auto *ASCast = dyn_cast<AddrSpaceCastInst>(&II)) {
```
svenvh, line 354: It would be nice to have a comment giving a one-line summary of what is happening inside this if-block.  Something like (mostly taken from your description):
"Add an additional bitcast in case address space cast also changes pointer element type".

Fznamznon: Sure, done.

svenvh: Thanks!

AlexeySotkin (approved):

PR #1165, opened 2021-08-19
Add support for SPV_INTEL_joint_matrix extension

AlexeySotkin:
The spec is available at https://github.com/intel/llvm/pull/4373


lib/SPIRV/SPIRVRegularizeLLVM.cpp
```diff
@@ -291,6 +293,46 @@ void SPIRVRegularizeLLVMBase::lowerUMulWithOverflow(
...
+    P = P.second.split(", ");
+    StringRef Cols = P.first.take_while(llvm::isDigit);
+    SPVName << kSPIRVTypeName::PostfixDelim << Cols.str();
+    P = P.second.split(", ");
+    StringRef Layout = P.first.substr(P.first.rfind(')') + 1);
+    SPVName << kSPIRVTypeName::PostfixDelim << Layout.str();
+    P = P.second.split(", ");
+    StringRef Scope = P.first.substr(P.first.rfind(')') + 1);
+    SPVName << kSPIRVTypeName::PostfixDelim << Scope.str();
+    ST->setName(SPVName.str());
```
AlexeySachkov, line 332: Wow, that is rough. Comments explaining which part of the demangling name we are trying to handle at each line would be helpful, especially if they all were referring to the same example you have atop of the enclosing if

lib/SPIRV/SPIRVRegularizeLLVM.cpp
```diff
@@ -291,6 +293,46 @@ void SPIRVRegularizeLLVMBase::lowerUMulWithOverflow(
...
+    std::string DemangledName = llvm::demangle(MangledName.str());
+    StringRef Name(DemangledName);
+    Name = Name.slice(Name.find('<') + 1, Name.rfind('>'));
+    std::stringstream SPVName;
+    SPVName << kSPIRVTypeName::PrefixAndDelim
+            << kSPIRVTypeName::JointMatrixINTEL << kSPIRVTypeName::Delimiter;
+    auto P = Name.split(", ");
+    StringRef ElemType = P.first;
+    if (size_t Space = ElemType.rfind(' '))
+      ElemType = ElemType.substr(Space + 1);
```
AlexeySachkov, line 318: What if there are no space? In which cases you expect the condition to return `false`?

lib/SPIRV/SPIRVUtil.cpp
```diff
@@ -148,7 +149,12 @@ std::string mapLLVMTypeToOCLType(const Type *Ty, bool Signed) {
     Ss << mapLLVMTypeToOCLType(EleTy, Signed) << Size;
     return Ss.str();
   }
-  return "invalid_type";
+  // Just mangle the type
```
AlexeySachkov, line 152: Some comment (or even an `assert`) about which types we expect here would be helpful

lib/SPIRV/SPIRVRegularizeLLVM.cpp
```diff
@@ -291,6 +293,46 @@ void SPIRVRegularizeLLVMBase::lowerUMulWithOverflow(
   UMulIntrinsic->setCalledFunction(UMulFunc);
 }
 
+void SPIRVRegularizeLLVMBase::adaptStructTypes(StructType *ST) {
+  if (!ST->hasName())
+    return;
+  StringRef STName = ST->getName();
+  STName.consume_front("struct.");
```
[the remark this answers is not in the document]
AlexeySotkin: I think we shouldn't exit here in this case. It should work just fine if there is no `.struct` in the front (or someone already removed it for us).

lib/SPIRV/SPIRVRegularizeLLVM.cpp
AlexeySotkin: It is ok if there is no space. This `if` handles cases when we have type modifier, like in the example `signed char`. In this case we take only last word "char". If there is no space it means that there is no modifiers and `ElemType` already holds the name of the type we want.

lib/SPIRV/SPIRVRegularizeLLVM.cpp
AlexeySotkin: Added more comments in e549d01

lib/SPIRV/SPIRVUtil.cpp
AlexeySotkin: Added a comment in e549d01

Issue #1151, opened 2021-08-10
New release for llvm 13.0.0


AlexeySotkin:
We should create a branch corresponding to `release/13.x` branch of llvm/llvm-project repository.

concatime:
Up, I would like to build mesa with clang 13, but the feature `gallium-opencl` requires `libclc`, which requires `llvm-spirv`.

kloczek:
+1

PR #1166, opened 2021-08-20
Fork llvm_release_130 branch

AlexeySotkin:
Also cherry-picking latest commits from master(main) branch.

### DIFF
--- a/.github/workflows/check-code-style.yml
+++ b/.github/workflows/check-code-style.yml
@@ -24,7 +24,7 @@ env:
   # We need compile command database in order to perform clang-tidy check. So,
   # in order to perform configure step we need to setup llvm-dev package. This
   # env variable used to specify desired version of it
-  LLVM_VERSION: 14
+  LLVM_VERSION: 13
 
 jobs:
   clang-format-and-tidy:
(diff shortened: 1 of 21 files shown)


AlexeySotkin:
Fixes #1151

.github/workflows/check-code-style.yml
```diff
@@ -24,7 +24,7 @@ env:
   # We need compile command database in order to perform clang-tidy check. So,
   # in order to perform configure step we need to setup llvm-dev package. This
   # env variable used to specify desired version of it
-  LLVM_VERSION: 14
+  LLVM_VERSION: 13
```
svenvh, line 27: A similar change is needed for `.travis.yml`

AlexeySotkin: Should we remove `.travis.yml`? It seems nobody needs binaries anyway.

svenvh: No objections personally, but we should probably decide that elsewhere for visibility (perhaps tooling call?).

While waiting for a decision, I'm okay to leave Travis CI broken/disabled, so that we can proceed with the llvm_release_130 branch.

svenvh (approved):

.github/workflows/check-code-style.yml
AlexeySotkin: Agree about the visibility.

PR #1174, opened 2021-08-24
Integer dot product translation

Quetzonarch:
When dot function is called in OpenCL-like LLVM IR (eg. _Z3_dot()) it's nearly always translated into OpDot (which works on floats), even when arguments are vectors of integers. This proposed change adds additional translation for dot function so that it's translated into a proper OpCode from SPV_KHR_integer_dot_product extension, based on the arguments used.


StuartDBrady:
In my view, this optimization is beyond the scope of the SPIR-V/LLVM Translator, and should instead be implemented in `spirv-opt` (in SPIR-V Tools) and/or as part of a standalone optimizer for LLVM IR making use of OpenCL builtins.  Such an optimizer could make use of the SPIR-V/LLVM Translator library interface, if needed, but it does not belong in this repository.  It _certainly_ shouldn't be enabled by default, given that 1) this is an optimization beyond the scope of the translator and 2) the translator needs to support consumers of SPIR-V that do not support the IDP extension.

StuartDBrady:
At first glance, the description led me to believe this was an optimization which was using pattern matching to detect cases where a floating point dot product has been used with floating point vectors that have been converted from integer vectors, i.e. optimizing OpDot(OpConvertStoF(_some_int_)) → OpSDot(_some_int_).

I now see that this is not what was intended, but that this is a correctness fix for translation of IR produced from source using the cl_khr_integer_dot_product extension.  I haven't looked very closely at the tests, but the code looks fine as far as I can see.  However, I'd request that the commit message and description be updated to make it clearer that this is adding support for cl_khr_integer_dot_product, rather than adding an optimization.

lib/SPIRV/OCLToSPIRV.cpp
```diff
@@ -1419,6 +1436,105 @@ void OCLToSPIRVBase::visitCallDot(CallInst *CI) {
...
+      // dot_acc_sat(uint, int, int, int) @_Z11dot_acc_satjiii
+      // dot_acc_sat(uint, uint, int, int) @_Z11dot_acc_satjjii 
+      IsFirstSigned = (IsDot) ? (MangledName[MangledName.size() - 3] == 'i')
+                              : (MangledName[MangledName.size() - 4] == 'i');
+      IsSecondSigned = (IsDot) ? (MangledName[MangledName.size() - 2] == 'i')
+                               : (MangledName[MangledName.size() - 3] == 'i');
+  }
+  std::string NamePref =
+      (IsFirstSigned != IsSecondSigned ? "SU" : ((IsFirstSigned) ? "S" : "U"));
+  std::string FinName = "_Z" +
```
svenvh, line 1511: It might be better to use `mutateCallInstSPIRV` inside this method, which will handle the mangling for you.

test/transcoding/SPV_KHR_integer_dot_product_OCLtoSPIRV_char4.ll
```diff
@@ -0,0 +1,83 @@
+; RUN: llvm-as < %s -o %t.bc
+; RUN: llvm-spirv -s %t.bc -o %t.regularized.bc
+; RUN: llvm-dis %t.regularized.bc -o - | FileCheck %s --check-prefix=CHECK-LLVM
+; RUN: llvm-spirv %t.bc -spirv-text --spirv-ext=+SPV_KHR_integer_dot_product -o - | FileCheck %s --check-prefix=CHECK-SPIRV
+
```
svenvh, line 5: Is it possible to add a `spirv-val` step too in those tests?

test/transcoding/SPV_KHR_integer_dot_product_OCLtoSPIRV_char4.ll
```diff
@@ -0,0 +1,83 @@
...
+!llvm.ident = !{!2}
+
+!0 = !{i32 1, !"wchar_size", i32 4}
+!1 = !{i32 2, i32 0}
+!2 = !{!"clang version 11.0.0 (https://github.com/c199914007/llvm.git 8b94769313ca84cb9370b60ed008501ff692cb71)"}
+!3 = !{i32 0, i32 0, i32 0, i32 0, i32 0, i32 0}
+!4 = !{!"none", !"none", !"none", !"none", !"none", !"none"}
+!5 = !{!"char4", !"uchar4", !"char4", !"uchar4", !"char4", !"uchar4"}
+!6 = !{!"char __attribute__((ext_vector_type(4)))", !"uchar __attribute__((ext_vector_type(4)))", !"char __attribute__((ext_vector_type(4)))", !"uchar __attribute__((ext_vector_type(4)))", !"char __attribute__((ext_vector_type(4)))", !"uchar __attribute__((ext_vector_type(4)))"}
+!7 = !{!"", !"", !"", !"", !"", !""}
```
svenvh, line 83: Please add missing newline at EOF to the new tests.

svenvh (commented):
Like @StuartDBrady I initially also got misled by the description.  As I understand now, this is not an optimization but actually fixing a **bug** in the translator: it currently translates calls such as `dot(uchar4, uchar4)` into `OpDot %i8`, which is wrong because `OpDot` only operates on floating point types. Clarifying in the description that the current behaviour is wrong would be great, as Stuart mentioned.

I think we should work towards accepting this PR and am keen to help; we've left this for too long!

@Quetzonarch I guess first you should try to rebase onto latest `main` and then you'll probably hit some issues.  I have left some comments that should help you solve some problems that I expect will appear after rebasing, plus some other improvements.

Quetzonarch:
Thank you for the comments and suggestions regarding the change. Yes, the point of this is to fix incorrect translations of IR produced from source using the cl_khr_integer_dot_product extension, sorry for the misleading description. I've begun implementing suggested changes. I'll try to push the update as soon as possible.

Quetzonarch:
I've rebased the branch onto the latest main and applied suggested changes. Everything seemed to work alright, but some parts might need some additional checking (for example my use of mutateCallInstSPIRV function, as I wasn't too familiar with that one).

test/transcoding/SPV_KHR_integer_dot_product_OCLtoSPIRV_char4.ll
```diff
@@ -0,0 +1,85 @@
+; RUN: llvm-as < %s -o %t.bc
+; RUN: llvm-spirv -s %t.bc -o %t.regularized.bc
```
svenvh, line 2: You will need to add `--spirv-ext=+SPV_KHR_integer_dot_product` when converting a `.bc` that will be using the extension.

svenvh (commented):
Thanks for the update @Quetzonarch!  I've left a few more comments and the CI has found some issues; other than that it looks good to me.

Quetzonarch:
I've seen that CI had found some issues, so I've applied small changes to the last commit (slight code formatting in OCLToSPIRV and missing spir_func words in tests).

svenvh (approved):

PR #1176, opened 2021-08-24
Fix support of SPV_INTEL_vector_compute

Fznamznon:
The previous patch has fixed only case with 1-element vectors, however
after closer look at the spec it became clear that
SPV_INTEL_vector_compute actually allows any number of vector elements
(capability VectorAnyINTEL), so this is the proper fix.

### DIFF
--- a/lib/SPIRV/SPIRVLowerBitCastToNonStandardType.cpp
+++ b/lib/SPIRV/SPIRVLowerBitCastToNonStandardType.cpp
@@ -160,11 +160,11 @@ public:
         VectorType *SrcVecTy = getVectorType(BC->getSrcTy());
         if (SrcVecTy) {
           uint64_t NumElemsInSrcVec = SrcVecTy->getElementCount().getValue();
-          // SPV_INTEL_vector_compute allows 1-element vectors
+          // SPV_INTEL_vector_compute allows to use vectors with any number of
+          // components
           if (!isValidVectorSize(NumElemsInSrcVec) &&
-              !(Opts.isAllowedToUseExtension(
-                    ExtensionID::SPV_INTEL_vector_compute) &&
-                NumElemsInSrcVec == 1))
+              !Opts.isAllowedToUseExtension(
+                  ExtensionID::SPV_INTEL_vector_compute))
             report_fatal_error("Unsupported vector type with the size of: " +
                                    std::to_string(NumElemsInSrcVec),
                                false);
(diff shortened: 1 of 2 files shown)


lib/SPIRV/SPIRVLowerBitCastToNonStandardType.cpp
```diff
@@ -160,23 +160,23 @@ class SPIRVLowerBitCastToNonStandardTypePass
...
         if (SrcVecTy) {
           uint64_t NumElemsInSrcVec = SrcVecTy->getElementCount().getValue();
-          // SPV_INTEL_vector_compute allows 1-element vectors
+          // SPV_INTEL_vector_compute allows to use vectors with any number of
+          // components
           if (!isValidVectorSize(NumElemsInSrcVec) &&
-              !(Opts.isAllowedToUseExtension(
-                    ExtensionID::SPV_INTEL_vector_compute) &&
-                NumElemsInSrcVec == 1))
+              !Opts.isAllowedToUseExtension(
```
mlychkov, line 166: In case if this extension is included we doesn't lowering any instructions with vector types.
Maybe to move this check just to the beginning of this method?

Fznamznon: I was just thinking that maybe this pass will lower not only vector instructions one day, so I left it here. Do you think it is possible?

mlychkov: I think that we might rename this concrete method to use it for non standard vectors lowering only. And to write new functions for lowering other non supported patterns. It would make these changes atomic to each other.

Fznamznon: Ok, I've moved check to the beginning.

lib/SPIRV/SPIRVLowerBitCastToNonStandardType.cpp
```diff
@@ -150,6 +150,13 @@ class SPIRVLowerBitCastToNonStandardTypePass
     // parameter, since it added by an optimization.
     bool Changed = false;
 
+    // SPV_INTEL_vector_compute allows to use vectors with any number of
+    // components. Since this method only lowers vectors with non-standard
+    // in pure SPIR-V number of components, there is no need to do anything in
+    // case SPV_INTEL_vector_compute is enabled.
+    if (Opts.isAllowedToUseExtension(ExtensionID::SPV_INTEL_vector_compute))
+      return PreservedAnalyses::all();
```
AlexeySotkin, line 158: It is out of scope of this patch, but it is strange that we return `PreservedAnalyses`, because it is not an analysis pass AFAIK,

AlexeySotkin (approved):

lib/SPIRV/SPIRVLowerBitCastToNonStandardType.cpp
Fznamznon: I was following the logic that was made here before. I'm a newbie in passes that run through new passmanager. Probably @KornevNikita can help to understand it better once returns from vacation.

mlychkov (approved):

lib/SPIRV/SPIRVLowerBitCastToNonStandardType.cpp
mlychkov: I suppose current method better to return just bool "changed or not".
In main run method PreservedAnalyses should be created depending on changed value.

MrSidims (approved):

PR #1178, opened 2021-08-24
Add API to query capabilities, memory model and version

pauzinl:
This is one of the possible implementations of the API

### DIFF
--- a/include/LLVMSPIRVLib.h
+++ b/include/LLVMSPIRVLib.h
@@ -139,7 +139,7 @@ bool getSpecConstInfo(std::istream &IS,
 
 struct SPIRVHeaderData {
   std::uint32_t MemoryModel = 0x7fffffff; // MemoryModelMax
-  std::uint32_t Version = 0;
+  SPIRV::VersionNumber Version;
   std::vector<std::uint32_t> Capabilities;
 };
 /// \Brief Partially load SPIR-V from the stream and decode memory model,
--- a/include/LLVMSPIRVLib.h
+++ b/include/LLVMSPIRVLib.h
@@ -137,6 +137,17 @@ using SpecConstInfoTy = std::pair<uint32_t, uint32_t>;
 bool getSpecConstInfo(std::istream &IS,
                       std::vector<SpecConstInfoTy> &SpecConstInfo);
 
+struct SPIRVHeaderData {
+  std::uint32_t MemoryModel = 0x7fffffff; // MemoryModelMax
+  std::uint32_t Version = 0;
+  std::vector<std::uint32_t> Capabilities;
+};
+/// \Brief Partially load SPIR-V from the stream and decode memory model,
+/// version and capabilities.
+/// \return true if succeeds.
+bool readSPIRVHeader(std::istream &IS, SPIRVHeaderData &Result,
+                     std::string &ErrMsg);
+
 /// \brief Convert a SPIRVModule into LLVM IR.
 /// \returns null on failure.
 std::unique_ptr<Module>
--- a/include/LLVMSPIRVLib.h
+++ b/include/LLVMSPIRVLib.h
@@ -139,11 +139,12 @@ bool getSpecConstInfo(std::istream &IS,
 
 struct SPIRVHeaderData {
   std::uint32_t MemoryModel = 0x7fffffff; // MemoryModelMax
+  std::uint32_t AddressingModel = 0x7fffffff;
   SPIRV::VersionNumber Version;
   std::vector<std::uint32_t> Capabilities;
 };
-/// \Brief Partially load SPIR-V from the stream and decode memory model,
-/// version and capabilities.
+/// \brief Partially load SPIR-V from the stream and decode memory model,
+/// addressing model, version and capabilities.
 /// \return true if succeeds.
 bool readSPIRVHeader(std::istream &IS, SPIRVHeaderData &Result,
                      std::string &ErrMsg);
--- a/include/LLVMSPIRVLib.h
+++ b/include/LLVMSPIRVLib.h
@@ -138,9 +138,9 @@ bool getSpecConstInfo(std::istream &IS,
                       std::vector<SpecConstInfoTy> &SpecConstInfo);
 
 struct SPIRVHeaderData {
-  std::uint32_t MemoryModel = 0x7fffffff; // MemoryModelMax
-  std::uint32_t AddressingModel = 0x7fffffff;
-  SPIRV::VersionNumber Version;
+  std::uint32_t MemoryModel = ~0U; // SPIRVWORD_MAX
+  std::uint32_t AddressingModel = ~0U;
+  SPIRV::VersionNumber Version = static_cast<SPIRV::VersionNumber>(~0U);
   std::vector<std::uint32_t> Capabilities;
 };
 /// \brief Partially load SPIR-V from the stream and decode memory model,
--- a/lib/SPIRV/SPIRVReader.cpp
+++ b/lib/SPIRV/SPIRVReader.cpp
@@ -4364,6 +4364,51 @@ bool llvm::getSpecConstInfo(std::istream &IS,
   return !IS.bad();
 }
 
+bool llvm::readSPIRVHeader(std::istream &IS, SPIRVHeaderData &Result,
+                           std::string &ErrMsg) {
+  std::unique_ptr<SPIRVModule> BM(SPIRVModule::createSPIRVModule());
+  BM->setAutoAddExtensions(false);
+  SPIRVDecoder D(IS, *BM);
+  SPIRVWord Magic;
+  SPIRVWord Buffer;
+  bool Success;
+  D >> Magic;
+  if (!BM->getErrorLog().checkError(Magic == MagicNumber, SPIRVEC_InvalidModule,
+                                    "invalid magic number")) {
+    ErrMsg = "invalid magic number";
+    return false;
+  }
+
+  D >> Buffer;
+  Result.Version = Buffer;
+  D.ignore(3);
+  while (D.getWordCountAndOpCode()) {
+    Success = 0;
+    switch (D.OpCode) {
+    case spv::OpMemoryModel:
+      D.ignore(1);
+      Success = D.getNextSPIRVWord(Buffer);
+      if (Success)
+        Result.MemoryModel = Buffer;
+      break;
+    case spv::OpCapability:
+      Success = D.getNextSPIRVWord(Buffer);
+      if (Success)
+        Result.Capabilities.push_back(Buffer);
+      break;
+    default:
+      if (spv::OpSource == D.OpCode || spv::OpName == D.OpCode ||
+          spv::OpMemberName == D.OpCode || spv::OpDecorate == D.OpCode ||
+          spv::OpMemberDecorate == D.OpCode || spv::OpFunction == D.OpCode) {
+        return true;
+      } else {
+        D.ignoreInstruction();
+      }
+    }
+  }
+  return true;
+}
+
 // clang-format off
 const StringSet<> SPIRVToLLVM::BuiltInConstFunc {
   "convert", "get_work_dim", "get_global_size", "sub_group_ballot_bit_count",
--- a/lib/SPIRV/SPIRVReader.cpp
+++ b/lib/SPIRV/SPIRVReader.cpp
@@ -4371,7 +4371,6 @@ bool llvm::readSPIRVHeader(std::istream &IS, SPIRVHeaderData &Result,
   SPIRVDecoder D(IS, *BM);
   SPIRVWord Magic;
   SPIRVWord Buffer;
-  bool Success;
   D >> Magic;
   if (!BM->getErrorLog().checkError(Magic == MagicNumber, SPIRVEC_InvalidModule,
                                     "invalid magic number")) {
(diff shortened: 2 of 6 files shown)


AlexeySachkov:
Tagging @PawelJurek, @aratajew, @cdai2, @Nuullll, @svenvh and @tiwaria1.

Folks, in some of our (Intel) OpenCL runtimes we have a functionality of pre-validation of SPIR-V file we are going to consume: that includes checking its version, memory model and capabilities to compare them with the list of supported ones and reject SPIR-V file if we encounter something unsupported.

We would like to upstream that functionality into the translator to make it widely available for more vendors and we would like to hear your feedback about the proposed API - would it be useful to you? Is there anything additional which we should add to the list of extracted info? Do we need some alternative forms of the API to retrieve SPIR-V version and used capabilities through separate API calls?

lib/SPIRV/SPIRVReader.cpp
```diff
@@ -4364,6 +4364,51 @@ bool llvm::getSpecConstInfo(std::istream &IS,
...
+  bool Success;
+  D >> Magic;
+  if (!BM->getErrorLog().checkError(Magic == MagicNumber, SPIRVEC_InvalidModule,
+                                    "invalid magic number")) {
+    ErrMsg = "invalid magic number";
+    return false;
+  }
+
+  D >> Buffer;
+  Result.Version = Buffer;
```
AlexeySachkov, line 4383: Why not `D >> Result.Version` ?

pauzinl: Actually I just wanted to assign values to struct fields in the same way using variable `Buffer`. I will change it after I get another requested changes just not to spam with commits.

include/LLVMSPIRVLib.h
```diff
@@ -137,6 +137,17 @@ using SpecConstInfoTy = std::pair<uint32_t, uint32_t>;
 bool getSpecConstInfo(std::istream &IS,
                       std::vector<SpecConstInfoTy> &SpecConstInfo);
 
+struct SPIRVHeaderData {
+  std::uint32_t MemoryModel = 0x7fffffff; // MemoryModelMax
+  std::uint32_t Version = 0;
```
svenvh, line 142: We should have `SPIRV::VersionNumber` available here I think?  So we could use that instead of `uint32_t`.

lib/SPIRV/SPIRVReader.cpp
```diff
@@ -4364,6 +4364,51 @@ bool llvm::getSpecConstInfo(std::istream &IS,
...
+  D.ignore(3);
+  while (D.getWordCountAndOpCode()) {
+    Success = 0;
+    switch (D.OpCode) {
+    case spv::OpMemoryModel:
+      D.ignore(1);
+      Success = D.getNextSPIRVWord(Buffer);
+      if (Success)
+        Result.MemoryModel = Buffer;
+      break;
```
svenvh, line 4393: Can't we just return here, once we have seen `OpMemoryModel`?  See "2.4. Logical Layout of a Module" of the SPIR-V spec.

svenvh (commented):
> Tagging @PawelJurek, @aratajew, @cdai2, @Nuullll, @svenvh and @tiwaria1.
> 
> Folks, in some of our (Intel) OpenCL runtimes we have a functionality of pre-validation of SPIR-V file we are going to consume: that includes checking its version, memory model and capabilities to compare them with the list of supported ones and reject SPIR-V file if we encounter something unsupported.
> 
> We would like to upstream that functionality into the translator to make it widely available for more vendors and we would like to hear your feedback about the proposed API - would it be useful to you? Is there anything additional which we should add to the list of extracted info? Do we need some alternative forms of the API to retrieve SPIR-V version and used capabilities through separate API calls?

I don't have any objections against the current proposal with a single new API call returning a struct; I've added some minor comments in case you decide to go with that.

However I wonder if you have considered a different approach: capture what is allowed in terms of memory model and capabilities in `TranslatorOpts` and abort translation if you encounter anything that is not allowed.  That means we don't need a different API entry point and should also allow you to write tests for the new functionality.

lib/SPIRV/libSPIRV/SPIRVStream.cpp
```diff
@@ -222,6 +222,37 @@ bool SPIRVDecoder::getWordCountAndOpCode() {
...
+bool SPIRVDecoder::getNextSPIRVWord(SPIRVWord &Word) {
+  if (IS.eof()) {
+    Word = -1;
+    SPIRVDBG(spvdbgs() << "[SPIRVDecoder] getNextSPIRVWord EOF " << Word
+                       << '\n');
+    return false;
+  }
+  *this >> Word;
+#ifdef _SPIRV_SUPPORT_TEXT_FMT
+  if (SPIRVUseTextFormat) {
```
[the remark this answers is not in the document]
mlychkov: @pauzinl Could you please clarify why code on lines 235-241 duplicates code on lines 244-250?

lib/SPIRV/libSPIRV/SPIRVStream.cpp
```diff
@@ -222,6 +222,37 @@ bool SPIRVDecoder::getWordCountAndOpCode() {
...
+  }
+#endif
+  assert(!IS.bad() && "SPIRV stream is bad");
+  if (IS.fail()) {
+    Word = -1;
+    SPIRVDBG(spvdbgs() << "[SPIRVDecoder] getNextSPIRVWord FAIL " << Word
+                       << '\n');
+    return false;
+  }
+  SPIRVDBG(spvdbgs() << "[SPIRVDecoder] getWordCountAndOpCode " << Word
```
mlychkov, line 251:
```suggestion
  SPIRVDBG(spvdbgs() << "[SPIRVDecoder] getNextSPIRVWord " << Word
```

lib/SPIRV/SPIRVReader.cpp
```diff
@@ -4364,6 +4364,44 @@ bool llvm::getSpecConstInfo(std::istream &IS,
...
+    } else if (D.OpCode == OpMemoryModel) {
+      D.ignore(1);
+      Success = D.getNextSPIRVWord(Buffer);
+      if (Success) {
+        Result.MemoryModel = Buffer;
+        return true;
+      }
+    }
+  }
+  ErrMsg = "Memory model not specified";
```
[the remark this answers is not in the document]
pauzinl: What error message would you recommend to use here?

lib/SPIRV/libSPIRV/SPIRVStream.cpp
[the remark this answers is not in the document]
pauzinl: @mlychkov I thought it should work the same as `getWordCountAndOpCode`. Anyway, I will remove this code (233-243) because we don't need to work with text SPIR-V.

lib/SPIRV/SPIRVReader.cpp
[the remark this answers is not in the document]
AlexeySotkin: Please ignore my comment.

lib/SPIRV/SPIRVReader.cpp
```diff
@@ -4364,6 +4364,44 @@ bool llvm::getSpecConstInfo(std::istream &IS,
...
+    ErrMsg = "invalid magic number";
+    return false;
+  }
+
+  D >> Buffer;
+  Result.Version = static_cast<VersionNumber>(Buffer);
+  // Skip the rest of the header
+  D.ignore(3);
+  while (D.getWordCountAndOpCode()) {
+    Success = false;
```
mlychkov, line 4387: Success may be removed because we may write the code as:
`if (D.getNextSPIRVWord(Buffer))`

lib/SPIRV/SPIRVReader.cpp
```diff
@@ -4364,6 +4364,44 @@ bool llvm::getSpecConstInfo(std::istream &IS,
...
+    if (D.OpCode == OpCapability) {
+      Success = D.getNextSPIRVWord(Buffer);
+      if (Success)
+        Result.Capabilities.push_back(Buffer);
+    } else if (D.OpCode == OpMemoryModel) {
+      D.ignore(1);
```
mlychkov, line 4393: Please add a comment about what is ignored here.

lib/SPIRV/SPIRVReader.cpp
```diff
@@ -4364,6 +4364,44 @@ bool llvm::getSpecConstInfo(std::istream &IS,
...
+  Result.Version = static_cast<VersionNumber>(Buffer);
+  // Skip the rest of the header
+  D.ignore(3);
+  while (D.getWordCountAndOpCode()) {
+    Success = false;
+    if (D.OpCode == OpCapability) {
+      Success = D.getNextSPIRVWord(Buffer);
+      if (Success)
+        Result.Capabilities.push_back(Buffer);
+    } else if (D.OpCode == OpMemoryModel) {
```
mlychkov, line 4392: If there is no MemoryModel instruction in a SPIR-V file then we will read a whole file. May there be some instruction after that there is no sense in searching MemoryModel and we can exit the loop earlier?

AlexeySotkin: I think if we encounter any instruction which is not one of `OpCapability`, `OpExtension`, `OpExtInstImport`, `OpMemoryModel` we can exit.

include/LLVMSPIRVLib.h
```diff
@@ -137,6 +137,17 @@ using SpecConstInfoTy = std::pair<uint32_t, uint32_t>;
 bool getSpecConstInfo(std::istream &IS,
                       std::vector<SpecConstInfoTy> &SpecConstInfo);
 
+struct SPIRVHeaderData {
+  std::uint32_t MemoryModel = 0x7fffffff; // MemoryModelMax
+  SPIRV::VersionNumber Version;
+  std::vector<std::uint32_t> Capabilities;
+};
```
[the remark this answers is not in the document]
AlexeySotkin: I think it really would be useful to get information whether the module has Logical Physical32 or Physical64 addressing model

include/LLVMSPIRVLib.h
```diff
@@ -137,6 +137,18 @@ using SpecConstInfoTy = std::pair<uint32_t, uint32_t>;
 bool getSpecConstInfo(std::istream &IS,
                       std::vector<SpecConstInfoTy> &SpecConstInfo);
 
+struct SPIRVHeaderData {
+  std::uint32_t MemoryModel = 0x7fffffff; // MemoryModelMax
+  std::uint32_t AddressingModel = 0x7fffffff;
+  SPIRV::VersionNumber Version;
+  std::vector<std::uint32_t> Capabilities;
```
mlychkov, line 144: SPIRVWord type may be used for uint_32 values and SPIRVWORD_MAX as default value.

lib/SPIRV/SPIRVReader.cpp
```diff
@@ -4364,6 +4364,43 @@ bool llvm::getSpecConstInfo(std::istream &IS,
...
+                           std::string &ErrMsg) {
+  std::unique_ptr<SPIRVModule> BM(SPIRVModule::createSPIRVModule());
+  BM->setAutoAddExtensions(false);
+  SPIRVDecoder D(IS, *BM);
+  SPIRVWord Magic;
+  SPIRVWord Buffer;
+  D >> Magic;
+  if (!BM->getErrorLog().checkError(Magic == MagicNumber, SPIRVEC_InvalidModule,
+                                    "invalid magic number")) {
+    ErrMsg = "invalid magic number";
```
mlychkov, line 4377: Should we save ErrMsg explicitly if we already printed it in checkError()?

lib/SPIRV/SPIRVReader.cpp
```diff
@@ -4364,6 +4364,43 @@ bool llvm::getSpecConstInfo(std::istream &IS,
...
+bool llvm::readSPIRVHeader(std::istream &IS, SPIRVHeaderData &Result,
+                           std::string &ErrMsg) {
+  std::unique_ptr<SPIRVModule> BM(SPIRVModule::createSPIRVModule());
+  BM->setAutoAddExtensions(false);
+  SPIRVDecoder D(IS, *BM);
+  SPIRVWord Magic;
+  SPIRVWord Buffer;
+  D >> Magic;
+  if (!BM->getErrorLog().checkError(Magic == MagicNumber, SPIRVEC_InvalidModule,
+                                    "invalid magic number")) {
```
mlychkov, line 4376:
```suggestion
                                    "Invalid magic number")) {
```

include/LLVMSPIRVLib.h
```diff
@@ -137,6 +137,18 @@ using SpecConstInfoTy = std::pair<uint32_t, uint32_t>;
 bool getSpecConstInfo(std::istream &IS,
                       std::vector<SpecConstInfoTy> &SpecConstInfo);
 
+struct SPIRVHeaderData {
```
mlychkov, line 140: Would it be better to define this struct somewhere in `lib/SPIRV/libSPIRV` subdir to get access to `SPIRVWord` and `SPIRVWORD_MAX` definitions?

MrSidims: I believe it won't work. The closest header by meaning would be SPIRVEntry.h. So we would need to include SPIRVEntry.h here, but this very file serves a purpose of compilation of a driver aka llvm-spirv binary, while SPIRVEntry.h (as well as other files in libSPIRV) describes SPIR-V constructs: instructions, types etc.

mlychkov: Maybe, my concern wasn't clear. I think that SPIRV header struct shouldn't be defined in include/LLVM* files, because it has nothing with LLVM infrastructure support. It SPIR-V file specific only.
So, it would be better to be defined in lib/SPIRV/* files.

MrSidims:
Closing as outdated

PR #1180, opened 2021-08-26
[SPV-IR] Ignore SPIR-V linkage name for BuiltIn variable

Nuullll:
The BuiltIn variable/call name in SPV-IR should stick to
"__spirv_BuiltIn*", no matter what the SPIR-V linkage name is.

This is an issue exposed by PR #1133.


Nuullll:
ping

PR #1185, opened 2021-08-30
Fix SPIRV->LLVM translation of variable lenght arrays

Fznamznon:
VLAs were translated in section where non-instruction SPIR-V
values are translated. However translation of them produce LLVM IR
instructions. This patch moves translation of VLAs to the proper place
where other instructions are created, so the placeholder creation logic
also works for them. This fixes the case when VLA used in complex
control flow graph caused producing of invalid LLVM module during
reverse translation.

### DIFF
--- a/lib/SPIRV/SPIRVReader.cpp
+++ b/lib/SPIRV/SPIRVReader.cpp
@@ -1510,21 +1510,6 @@ Value *SPIRVToLLVM::transValueWithoutDecoration(SPIRVValue *BV, Function *F,
     return Res;
   }
 
-  case OpVariableLengthArrayINTEL: {
-    auto *VLA = static_cast<SPIRVVariableLengthArrayINTEL *>(BV);
-    llvm::Type *Ty = transType(BV->getType()->getPointerElementType());
-    llvm::Value *ArrSize = transValue(VLA->getOperand(0), F, BB, false);
-    return mapValue(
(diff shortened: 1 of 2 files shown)


Fznamznon:
@AlexeySotkin , @MrSidims , could you please take a look?

MrSidims (approved):
LGTM

Fznamznon:
> LGTM
> Just a side question: does SPV-IR translation flow works for these instructions? Aka what would the appropriate option produce, alloca or __spirv_VariableLengthArrayINTEL?

Looking at the code, it seems OpVariableLengthArrayINTEL is translated only to/from alloca instructions (I haven't tried to feed some IR with `__spirv_VariableLengthArrayINTEL` to the Writer though). But I think we have a bunch of other instructions that are translated directly to LLVM instructions. I think SPIR-V friendly IR is needed to cover the case when there is no LLVM IR instruction covering the functionality of SPIR-V one.

AlexeySotkin (commented):
In general LGTM. But it is unclear how checks in the test are related to VLA functionality.

Fznamznon:
> In general LGTM. But it is unclear how checks in the test are related to VLA functionality.

They check that alloca instruction for VLA is added and used from the proper basic blocs. Before this patch, it was added to the module in incorrect place, so it was broken.

AlexeySotkin (approved):

PR #1191, opened 2021-08-31
Improve LLVM annotations translation to SPIR-V

MrSidims:
https://reviews.llvm.org/D88645 introduces additional parameter
to var/ptr/global annotation instruction, which can be set by
clang::annotation attribute that has variadic constant expression
argument. Right now we can't translate it properly and an extension
is being discussed, but yet we need a W/A for this. As a W/A the
current patch makes this argument (in case if it is const int) be
translatable as a string put in UserSemantic decorations, so the
IL flow is looking like this:
Annotation in source-code:
[[clang::annotate("custom", 30, 60)]] int *Var;

LLVM IR before:
 @.str = private unnamed_addr constant [7 x i8] c"custom\00"
 @.args = private unnamed_addr constant { i32, i32 } { i32 30, i32 60 }
 // ...
 call void @llvm.var.annotation(i8* ...,
       i8* getelementptr inbounds ([7 x i8], [7 x i8]* @.str, i32 0, i32 0),
       i8* ... , i32 ..., i8* bitcast ({ i32, i32 }* @.args to i8*))
 // ...

SPIR-V:
  Decorate %GEP% UserSemantic "custom: 30, 60"

LLVM IR after:
  @3 = private unnamed_addr constant [15 x i8] c"custom: 30, 60\00"
  // ...
  call void @llvm.var.annotation(i8* ..., i8* getelementptr inbounds (
      [15 x i8], [15 x i8]* @3, i32 0, i32 0), i8* ..., i32 ..., i8* undef)
  // ...

As you see the annotation remains to be a string after the translation. It's
done so because we wouldn't be able to distinguish
[[clang::annotate("custom", 30, 60)]] from [[clang::annotate("custom, 30, 60")]]
as they both would result in the same UserSemantic decoration.

Also this patch brings a bit of refactoring of add Intel FPGA decoration functions,
since their infrustructure can (and should) be reused here, solving an issue of
incorrect allowed FPGA memory SPIR-V extensions processing.

TODO:
1. Replace this W/A with an extension, when it's ready;
2. Global annotation instruction isn't handled properly yet, its handling requires even
   more refactoring, which I plan to do, when the extension is ready;
3. Filename annotation is not being always and properly translated. To fix this additional
   round of refactoring is required, yet again it will be done with the new extension implementation.

### DIFF
--- a/lib/SPIRV/SPIRVWriter.cpp
+++ b/lib/SPIRV/SPIRVWriter.cpp
@@ -3078,6 +3078,7 @@ SPIRVValue *LLVMToSPIRVBase::transIntrinsicInst(IntrinsicInst *II,
     }
 
     std::string AnnotationString;
+    // Process main annotation with the attached optional integer parameters
     if (GetElementPtrInst *GEP =
             dyn_cast<GetElementPtrInst>(II->getArgOperand(1))) {
       if (Constant *C = dyn_cast<Constant>(GEP->getOperand(0))) {
@@ -3456,11 +3457,10 @@ bool LLVMToSPIRVBase::transGlobalVariables() {
                      Return = true;
                      break;
                    default:
-                     Return = false;
-                     break;
+                     return false;
                    }
                  } else
-                   Return = false;
+                   return false;
                }
                return Return;
              }())
--- a/lib/SPIRV/SPIRVWriter.cpp
+++ b/lib/SPIRV/SPIRVWriter.cpp
@@ -2200,7 +2200,7 @@ SPIRVValue *LLVMToSPIRVBase::oclTransSpvcCastSampler(CallInst *CI,
 
 using DecorationsInfoVec = std::vector<std::pair<Decoration, std::string>>;
 
-struct IntelFPGADecorations {
+struct AnnotationDecorations {
   DecorationsInfoVec MemoryAttributesVec;
   DecorationsInfoVec MemoryAccessesVec;
 };
@@ -2242,9 +2242,47 @@ struct IntelLSUControlsInfo {
   llvm::Optional<unsigned> PrefetchInfo;
 };
 
-IntelFPGADecorations
-tryParseIntelFPGAAnnotationString(StringRef AnnotatedCode) {
-  IntelFPGADecorations Decorates;
+void processOptionalAnnotationInfo(Constant *Const,
+                                   std::string &AnnotationString) {
+  // Handle optional parameter. It can be for example
+  // { %struct.S, i8*, void ()* } { %struct.S undef, i8* null,
+  //                                void ()* @_Z4blahv }
+  // Now we will just handle integer constants (wrapped in a constant
+  // struct, that is being bitcasted to i8*), converting them to string.
+  // TODO: remove this workaround when/if an extension spec that allows or adds
+  // variadic-arguments UserSemantic decoration
+  if (!Const->getNumOperands())
+    return;
+  if (auto *CStruct = dyn_cast_or_null<ConstantStruct>(Const->getOperand(0))) {
+    uint32_t NumOperands = CStruct->getNumOperands();
+    if (!NumOperands)
+      return;
+    if (auto *CInt = dyn_cast<ConstantInt>(CStruct->getOperand(0))) {
+      AnnotationString += ": ";
+      AnnotationString += std::to_string(CInt->getSExtValue());
+    }
+    for (uint32_t I = 1; I != CStruct->getNumOperands(); ++I) {
+      if (auto *CInt = dyn_cast<ConstantInt>(CStruct->getOperand(I))) {
+        AnnotationString += ", ";
+        AnnotationString += std::to_string(CInt->getSExtValue());
+      }
+    }
+  }
+}
+
+AnnotationDecorations tryParseAnnotationString(SPIRVModule *BM,
+                                               StringRef AnnotatedCode) {
+  AnnotationDecorations Decorates;
+  const bool AllowFPGAMemAccesses =
+      BM->isAllowedToUseExtension(ExtensionID::SPV_INTEL_fpga_memory_accesses);
+  const bool AllowFPGAMemAttr = BM->isAllowedToUseExtension(
+      ExtensionID::SPV_INTEL_fpga_memory_attributes);
+  if (!(AllowFPGAMemAccesses || AllowFPGAMemAttr)) {
+    Decorates.MemoryAttributesVec.emplace_back(DecorationUserSemantic,
+                                               AnnotatedCode.str());
+    return Decorates;
+  }
+
   IntelLSUControlsInfo LSUControls;
 
   // Intel FPGA decorations are separated into
@@ -2258,25 +2296,27 @@ tryParseIntelFPGAAnnotationString(StringRef AnnotatedCode) {
     // Drop the braces surrounding the actual decoration
     const StringRef AnnotatedDecoration = AnnotatedCode.substr(
         DecorationsIt->position() + 1, DecorationsIt->length() - 2);
+
     std::pair<StringRef, StringRef> Split = AnnotatedDecoration.split(':');
     StringRef Name = Split.first, ValueStr = Split.second;
-
-    if (Name == "params") {
-      unsigned ParamsBitMask = 0;
-      bool Failure = ValueStr.getAsInteger(10, ParamsBitMask);
-      assert(!Failure && "Non-integer LSU controls value");
-      (void)Failure;
-      LSUControls.setWithBitMask(ParamsBitMask);
-    } else if (Name == "cache-size") {
-      if (!LSUControls.CacheSizeInfo.hasValue())
-        continue;
-      unsigned CacheSizeValue = 0;
-      bool Failure = ValueStr.getAsInteger(10, CacheSizeValue);
-      assert(!Failure && "Non-integer cache size value");
-      (void)Failure;
-      LSUControls.CacheSizeInfo = CacheSizeValue;
-    } // TODO: Support LSU prefetch size, which currently defaults to 0
-    else /* Memory attributes */ {
+    if (AllowFPGAMemAccesses) {
+      if (Name == "params") {
+        unsigned ParamsBitMask = 0;
+        bool Failure = ValueStr.getAsInteger(10, ParamsBitMask);
+        assert(!Failure && "Non-integer LSU controls value");
+        (void)Failure;
+        LSUControls.setWithBitMask(ParamsBitMask);
+      } else if (Name == "cache-size") {
+        if (!LSUControls.CacheSizeInfo.hasValue())
+          continue;
+        unsigned CacheSizeValue = 0;
+        bool Failure = ValueStr.getAsInteger(10, CacheSizeValue);
+        assert(!Failure && "Non-integer cache size value");
+        (void)Failure;
+        LSUControls.CacheSizeInfo = CacheSizeValue;
+      } // TODO: Support LSU prefetch size, which currently defaults to 0
+    }
+    if (AllowFPGAMemAttr) {
       StringRef Annotation;
       Decoration Dec;
       if (Name == "pump") {
@@ -2322,13 +2362,8 @@ std::vector<SPIRVWord> getBankBitsFromString(StringRef S) {
   return Bits;
 }
 
-void addIntelFPGADecorations(SPIRVEntry *E, DecorationsInfoVec &Decorations) {
+void addAnnotationDecorations(SPIRVEntry *E, DecorationsInfoVec &Decorations) {
   SPIRVModule *M = E->getModule();
-  if (!M->isAllowedToUseExtension(
-          ExtensionID::SPV_INTEL_fpga_memory_attributes) &&
-      !M->isAllowedToUseExtension(ExtensionID::SPV_INTEL_fpga_memory_accesses))
-    return;
-
   for (const auto &I : Decorations) {
     // Such decoration already exists on a type, skip it
     if (E->hasDecorate(I.first, /*Index=*/0, /*Result=*/nullptr)) {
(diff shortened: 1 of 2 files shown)


MrSidims:
Moved to draft, since I want to address fixing of one more bug

MrSidims:
@AlexeySotkin @Fznamznon @mlychkov @vmaksimo please take a look

AlexeySotkin:
How are you going to distinguish `[[clang::annotate("custom", 30, 60)]]` from `[[clang::annotate("custom, 30, 60")]]`? Or you don't need to distinguish between them?

MrSidims:
> How are you going to distinguish `[[clang::annotate("custom", 30, 60)]]` from `[[clang::annotate("custom, 30, 60")]]`? Or you don't need to distinguish between them?

Right now:
1. there is no need for that;
2. it's not possible right now, the extension I mentioned in the description will add a decoration with variadic number of Literal parameters, which can be applied simultaneously with UserSemantic decoration. So these two cases will be distinguished like this:
```
[[clang::annotate("custom, 30, 60")]]
UserSemantic %TARGET% "custom, 30, 60"
```
Vs
```
[[clang::annotate("custom", 30, 60)]]
UserSemantic %TARGET% "custom"
%DecorationName% %TARGET% 30 60
```

lib/SPIRV/SPIRVWriter.cpp
```diff
@@ -2242,9 +2242,47 @@ struct IntelLSUControlsInfo {
...
+    return;
+  if (auto *CStruct = dyn_cast_or_null<ConstantStruct>(Const->getOperand(0))) {
+    uint32_t NumOperands = CStruct->getNumOperands();
+    if (!NumOperands)
+      return;
+    if (auto *CInt = dyn_cast<ConstantInt>(CStruct->getOperand(0))) {
+      AnnotationString += ": ";
+      AnnotationString += std::to_string(CInt->getSExtValue());
+    }
+    for (uint32_t I = 1; I != CStruct->getNumOperands(); ++I) {
```
mlychkov, line 2264:
```suggestion
    for (uint32_t I = 1; I != NumOperands; ++I) {
```

lib/SPIRV/SPIRVWriter.cpp
```diff
@@ -2242,9 +2242,47 @@ struct IntelLSUControlsInfo {
...
+      if (auto *CInt = dyn_cast<ConstantInt>(CStruct->getOperand(I))) {
```
mlychkov, line 2265: If some CStruct param is not an integer then it will be just dropped. I.e. '30, "word", 60' will be translated to '30, 60'.
Maybe it would be better to translate optional parameters only when all of them are integers and none of them if there is at least one non integer value.

lib/SPIRV/SPIRVWriter.cpp
```diff
@@ -2971,35 +3034,41 @@ SPIRVValue *LLVMToSPIRVBase::transIntrinsicInst(IntrinsicInst *II,
       SV = transValue(II->getOperand(0), BB);
     }
 
-    GetElementPtrInst *GEP = cast<GetElementPtrInst>(II->getArgOperand(1));
-    Constant *C = cast<Constant>(GEP->getOperand(0));
-    StringRef AnnotationString;
-    getConstantStringInfo(C, AnnotationString);
+    std::string AnnotationString;
+    if (GetElementPtrInst *GEP =
+            dyn_cast<GetElementPtrInst>(II->getArgOperand(1))) {
```
mlychkov, line 3039: May there be a case when var_annotation has non GEP 2nd argument? Should we try to process 5th argument if we couldn't translate 2nd?

lib/SPIRV/SPIRVWriter.cpp
MrSidims: I'd prefer to translate as much as possible within the current approach.

Also, right now it's not a real-life scenario (or at least not something that our backend dev team desire (and since there is no much interest externally I'm guessing that it's not a real-life scenario)). So taking in account, that this code will be replaced with another code within few months I wouldn't bother about it (yet it doesn't bring any bugs/crashes etc).

lib/SPIRV/SPIRVWriter.cpp
```diff
@@ -3008,19 +3077,21 @@ SPIRVValue *LLVMToSPIRVBase::transIntrinsicInst(IntrinsicInst *II,
...
+        StringRef StrRef;
+        getConstantStringInfo(C, StrRef);
+        AnnotationString += StrRef.str();
+      }
+    }
+    if (BitCastInst *Cast = dyn_cast<BitCastInst>(II->getArgOperand(4)))
+      if (Constant *C = dyn_cast_or_null<Constant>(Cast->getOperand(0)))
+        processOptionalAnnotationInfo(C, AnnotationString);
+    AnnotationDecorations Decorations =
+        tryParseAnnotationString(BM, AnnotationString);
```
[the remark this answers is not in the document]
MrSidims: Sure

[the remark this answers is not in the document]
MrSidims: I probably left this code inlined because at some point I wanted to resolve:
```
Filename annotation is not being always and properly translated. To fix this additional
round of refactoring is required, yet again it will be done with the new extension implementation.
```
issue, and draft code was introducing some differences between these two blocks. But in the end I won't fix the issue in this PR and hence the code can and should be outlined in a different function.

lib/SPIRV/SPIRVWriter.cpp
MrSidims: >> May there be a case when var_annotation has non GEP 2nd argument?

If I understand correctly, it's not a real-life scenario. LangRef says, that var/ptr/annotation should always have 2nd and 3rd arguments to be pointers to a global string and we always accesses to a global string via GEP instruction aka via a pointer to [0] element with a specified memory size.

lib/SPIRV/SPIRVWriter.cpp
MrSidims: Thanks

lib/SPIRV/SPIRVWriter.cpp
```diff
@@ -3008,19 +3077,21 @@ SPIRVValue *LLVMToSPIRVBase::transIntrinsicInst(IntrinsicInst *II,
       AnnotSubj = cast<CastInst>(AnnotSubj)->getOperand(0);
```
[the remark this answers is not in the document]
MrSidims: Indeed!

lib/SPIRV/SPIRVWriter.cpp
```diff
@@ -3370,6 +3441,30 @@ bool LLVMToSPIRVBase::transGlobalVariables() {
...
+                     Return = true;
+                     break;
+                   default:
+                     return false;
+                   }
+                 } else
+                   return false;
+               }
+               return Return;
+             }())
```
Fznamznon, line 3460:
```suggestion
               // Check if the GV is used only in var/ptr instructions. If yes -
               // skip processing of this since it's only an annotation GV.
               for (auto *U : I->users()) {
                 Value *V = U->stripPointerCasts();
                 if (auto *GEP =
                         dyn_cast_or_null<GetElementPtrInst>(V))
                   V = GEP;
                 if (auto *II = dyn_cast_or_null<IntrinsicInst>(V)) {
                   switch (II->getIntrinsicID()) {
                   case Intrinsic::var_annotation:
                   case Intrinsic::ptr_annotation:
                     continue;
                   default:
                     return false;
                   }
                 }
                 return false;
               }
               return true;
             }())
```

lib/SPIRV/SPIRVWriter.cpp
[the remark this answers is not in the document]
MrSidims: Hah, actually, no.
https://llvm.org/doxygen/classllvm_1_1Value.html
```
stripPointerCasts () const
Strip off pointer casts, all-zero GEPs and address space casts
```
we don't want to skip GEPs

lib/SPIRV/SPIRVWriter.cpp
```diff
@@ -2242,9 +2242,47 @@ struct IntelLSUControlsInfo {
...
+}
+
+AnnotationDecorations tryParseAnnotationString(SPIRVModule *BM,
+                                               StringRef AnnotatedCode) {
+  AnnotationDecorations Decorates;
+  const bool AllowFPGAMemAccesses =
+      BM->isAllowedToUseExtension(ExtensionID::SPV_INTEL_fpga_memory_accesses);
+  const bool AllowFPGAMemAttr = BM->isAllowedToUseExtension(
+      ExtensionID::SPV_INTEL_fpga_memory_attributes);
+  if (!(AllowFPGAMemAccesses || AllowFPGAMemAttr)) {
```
Fznamznon, line 2296: It is totally up to you, but reverse condition kinda looks easier to understand
```suggestion
  if (!AllowFPGAMemAccesses && !AllowFPGAMemAttr) {
```

lib/SPIRV/SPIRVWriter.cpp
```diff
@@ -2242,9 +2242,47 @@ struct IntelLSUControlsInfo {
...
+  // Handle optional parameter. It can be for example
+  // { %struct.S, i8*, void ()* } { %struct.S undef, i8* null,
+  //                                void ()* @_Z4blahv }
+  // Now we will just handle integer constants (wrapped in a constant
+  // struct, that is being bitcasted to i8*), converting them to string.
+  // TODO: remove this workaround when/if an extension spec that allows or adds
+  // variadic-arguments UserSemantic decoration
+  if (!Const->getNumOperands())
+    return;
+  if (auto *CStruct = dyn_cast_or_null<ConstantStruct>(Const->getOperand(0))) {
```
Fznamznon, line 2256: It seems we expect that `Const` has at least one operand, what is the point to use `dyn_cast_or_null` then?

lib/SPIRV/SPIRVWriter.cpp
```diff
@@ -2339,48 +2374,76 @@ void addIntelFPGADecorations(SPIRVEntry *E, DecorationsInfoVec &Decorations) {
     case DecorationUserSemantic:
       E->addDecorate(new SPIRVDecorateUserSemanticAttr(E, I.second));
       break;
-    case DecorationMemoryINTEL:
-      E->addDecorate(new SPIRVDecorateMemoryINTELAttr(E, I.second));
-      break;
+    case DecorationMemoryINTEL: {
+      if (M->isAllowedToUseExtension(
+              ExtensionID::SPV_INTEL_fpga_memory_attributes))
```
Fznamznon, line 2395: Before your change this function early exited if `SPV_INTEL_fpga_memory_attributes` and `SPV_INTEL_fpga_memory_accesses` both not allowed. Now there is checks for only one extension for most of decorations. Looks like it changes the previous logic, was it an intention?

lib/SPIRV/SPIRVWriter.cpp
```diff
@@ -3008,19 +3077,21 @@ SPIRVValue *LLVMToSPIRVBase::transIntrinsicInst(IntrinsicInst *II,
...
+    if (GetElementPtrInst *GEP =
+            dyn_cast<GetElementPtrInst>(II->getArgOperand(1))) {
+      if (Constant *C = dyn_cast<Constant>(GEP->getOperand(0))) {
+        StringRef StrRef;
+        getConstantStringInfo(C, StrRef);
+        AnnotationString += StrRef.str();
+      }
+    }
+    if (BitCastInst *Cast = dyn_cast<BitCastInst>(II->getArgOperand(4)))
+      if (Constant *C = dyn_cast_or_null<Constant>(Cast->getOperand(0)))
```
Fznamznon: Just a NIT
```suggestion
    if (auto *Cast = dyn_cast<BitCastInst>(II->getArgOperand(4)))
      if (auto *C = dyn_cast_or_null<Constant>(Cast->getOperand(0)))
```

lib/SPIRV/SPIRVWriter.cpp
```diff
@@ -3008,19 +3077,21 @@ SPIRVValue *LLVMToSPIRVBase::transIntrinsicInst(IntrinsicInst *II,
...
-      // of extension A even if extension B is the only one enabled.
-      // With the existing stack, these are always enabled/disabled
-      // simultaneously, however we should find a way to separate the
-      // actions for each individual extension.
-      Decorations = tryParseIntelFPGAAnnotationString(AnnotationString);
+    std::string AnnotationString;
+    // Process main annotation with the attached optional integer parameters
+    if (GetElementPtrInst *GEP =
+            dyn_cast<GetElementPtrInst>(II->getArgOperand(1))) {
+      if (Constant *C = dyn_cast<Constant>(GEP->getOperand(0))) {
```
Fznamznon: Just a NIT
```suggestion
    if (auto *GEP =
            dyn_cast<GetElementPtrInst>(II->getArgOperand(1))) {
      if (auto *C = dyn_cast<Constant>(GEP->getOperand(0))) {
```

lib/SPIRV/SPIRVWriter.cpp
```diff
@@ -2971,35 +3034,41 @@ SPIRVValue *LLVMToSPIRVBase::transIntrinsicInst(IntrinsicInst *II,
...
+        StringRef StrRef;
+        getConstantStringInfo(C, StrRef);
+        AnnotationString += StrRef.str();
+      }
+    }
+    if (BitCastInst *Cast = dyn_cast<BitCastInst>(II->getArgOperand(4)))
+      if (Constant *C = dyn_cast_or_null<Constant>(Cast->getOperand(0)))
```
Fznamznon: Just a NIT
```suggestion
    if (auto *Cast = dyn_cast<BitCastInst>(II->getArgOperand(4)))
      if (auto *C = dyn_cast_or_null<Constant>(Cast->getOperand(0)))
```

lib/SPIRV/SPIRVWriter.cpp
MrSidims: Yes, it was an intention, there are some decorations that are allowed only for `SPV_INTEL_fpga_memory_attributes` and some for `SPV_INTEL_fpga_memory_accesses`, so previously implementation wasn't separating them, which is a bug. From the description:
```
Also this patch brings a bit of refactoring of add Intel FPGA decoration functions,
since their infrustructure can (and should) be reused here, solving an issue of
incorrect allowed FPGA memory SPIR-V extensions processing.
```
Sorry for a mixing this up in a single patch, I was just trying to fit this patch in the existing code and 'handle FPGA memory decorations' infrastructure is looking natural to be reused. And since I was going to reuse this infrustructure I fixed this bug as well.

lib/SPIRV/SPIRVWriter.cpp
MrSidims: Ouch, I actually need to apply logic from [here](https://github.com/KhronosGroup/SPIRV-LLVM-Translator/pull/1191#discussion_r699990957) first, then check for `user_empty` and only then apply this comment :)

lib/SPIRV/SPIRVWriter.cpp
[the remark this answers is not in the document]
MrSidims: Applied. Yet I left
```
    AnnotationDecorations Decorations =
        tryParseAnnotationString(BM, AnnotationString);
```

outside of this function, since I'd expect these string to be unique for var, ptr and global annotations at some point.

lib/SPIRV/SPIRVWriter.cpp
MrSidims: No need for that, changed to dyn_cast

lib/SPIRV/SPIRVWriter.cpp
```diff
@@ -3370,6 +3434,32 @@ bool LLVMToSPIRVBase::transGlobalVariables() {
   for (auto I = M->global_begin(), E = M->global_end(); I != E; ++I) {
     if ((*I).getName() == "llvm.global.annotations")
       transGlobalAnnotation(&(*I));
+    else if ([&]() -> bool {
```
mlychkov, line 3437: May we capture only `I` variable?

mlychkov (approved):

lib/SPIRV/SPIRVWriter.cpp
MrSidims: Applied

lib/SPIRV/SPIRVWriter.cpp
MrSidims: And actually I had to go through GEP users as well.

Fznamznon (approved):

lib/SPIRV/SPIRVWriter.cpp
```diff
@@ -3370,6 +3434,34 @@ bool LLVMToSPIRVBase::transGlobalVariables() {
...
+                 auto *GEP = dyn_cast_or_null<GetElementPtrInst>(V);
+                 if (!GEP)
+                   return false;
+                 for (auto *GEPU : GEP->users()) {
+                   auto *II = dyn_cast<IntrinsicInst>(GEPU);
+                   if (!II)
+                     return false;
+                   switch (II->getIntrinsicID()) {
+                   case Intrinsic::var_annotation:
+                   case Intrinsic::ptr_annotation:
```
MrSidims, line 3455: `case Intrinsic::global_annotation:`

PR #1192, opened 2021-08-31
Amend SingleElementVectorINTEL decoration use cases according to spec update

DmitryBushev:
Handling of SingleElementVectorINTEL decoration was amended to match new state of a spec: https://github.com/intel/llvm/pull/1612.

### DIFF
--- a/lib/SPIRV/SPIRVReader.cpp
+++ b/lib/SPIRV/SPIRVReader.cpp
@@ -1510,6 +1510,11 @@ Value *SPIRVToLLVM::transValueWithoutDecoration(SPIRVValue *BV, Function *F,
         LVar->addAttribute(kVCMetadata::VCByteOffset, utostr(Offset));
       if (BVar->hasDecorate(DecorationVolatile))
         LVar->addAttribute(kVCMetadata::VCVolatile);
+      SPIRVWord StarsOnElement;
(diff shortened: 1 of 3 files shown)


lib/SPIRV/SPIRVReader.cpp
```diff
@@ -220,6 +220,30 @@ Value *SPIRVToLLVM::getTranslatedValue(SPIRVValue *BV) {
...
+  auto VecDecorateSEV = BV->getDecorations(DecorationSingleElementVectorINTEL);
+  assert(VecDecorateSEV.size() == 1 &&
+         "Entry must have no more than one SingleElementVectorINTEL "
+         "decoration");
+  auto *DecorateSEV = VecDecorateSEV.back();
+  auto LiteralCount = DecorateSEV->getLiteralCount();
+  assert(LiteralCount <= 1 && "SingleElementVectorINTEL decoration must "
+                              "have no more than one literal");
+
+  SPIRVWord StarsOnElement =
```
mlychkov, line 239: This is just a nit, but I think that it would be better if the name of this var contained parameter name 'Levels' instead of stars.

mlychkov (approved):

MrSidims (approved):

PR #1195, opened 2021-09-02
Add convergent attribute during reverse translation of OpControlBarrier

KornevNikita:
There is a patch to keep convergent attr during translation of
OpControlBarrier to prevent optimization passes from making barrier calls
control-dependent.

### DIFF
--- a/test/spirv-ocl-builtins-version.spt
+++ b/test/spirv-ocl-builtins-version.spt
@@ -64,5 +64,5 @@
 ; CHECK-LLVM-SPV-IR: call spir_func void @_Z22__spirv_ControlBarrieriii(i32 2, i32 2, i32 784) [[attr]]
 ; CHECK-LLVM-SPV-IR: call spir_func void @_Z22__spirv_ControlBarrieriii(i32 2, i32 2, i32 2320) [[attr]]
 ; CHECK-LLVM-SPV-IR: call spir_func void @_Z22__spirv_ControlBarrieriii(i32 2, i32 2, i32 2832) [[attr]]
-; FIXME: shall we apply convergent attribute to SPIR-V friendly IR representaiton as well?
-; CHECK-LLVM-SPV-IR: attributes [[attr]] = { nounwind }
+; FIXME: shall we apply convergent attribute to SPIR-V friendly IR representaiton as well? (There is workaround to keep convergent attr for OpControlBarrier)
+; CHECK-LLVM-SPV-IR: attributes [[attr]] = { convergent nounwind }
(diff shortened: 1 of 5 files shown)


MrSidims:
WDYT about 'Fix translation of OpControlBarrier' -> 'Add convergent attribute during reverse translation of OpControlBarrier' ?

test/spirv-ocl-builtins-version.spt
```diff
@@ -64,5 +64,5 @@
 ; CHECK-LLVM-SPV-IR: call spir_func void @_Z22__spirv_ControlBarrieriii(i32 2, i32 2, i32 784) [[attr]]
 ; CHECK-LLVM-SPV-IR: call spir_func void @_Z22__spirv_ControlBarrieriii(i32 2, i32 2, i32 2320) [[attr]]
 ; CHECK-LLVM-SPV-IR: call spir_func void @_Z22__spirv_ControlBarrieriii(i32 2, i32 2, i32 2832) [[attr]]
-; FIXME: shall we apply convergent attribute to SPIR-V friendly IR representaiton as well?
-; CHECK-LLVM-SPV-IR: attributes [[attr]] = { nounwind }
+; FIXME: shall we apply convergent attribute to SPIR-V friendly IR representaiton as well? (There is workaround to keep convergent attr for OpControlBarrier)
```
Fznamznon, line 67: Is it a workaround?

MrSidims: I believe - yes. We don't have `convergent` analog in SPIR-V and hence we can't translate it properly from LLVM IR to SPIR-V and back to LLVM IR.

MrSidims: But yeah, even if we had this analog, I can barely imagine, how it can be associated with `ControlBarrier` instruction (for example it has no result id to apply a decoration).

So, I'd say it's yet a workaround unless proven otherwise :)

Fznamznon: Isn't something that we have to do based on definition of certain SPIR-V operations and `convergent` attribute? It seems that both translator and clang add convergent attribute to OpenCL builtins and this is an intention.

> Meanwhile, SPIR-V spec doesn't enforces ControlBarrier to be convergent

Yes, because I believe it doesn't know what is LLVM convergent attribute. I think it prevents LLVM optimizations from breaking the logic of barrier operation.
BTW, looking at the code in SPIR-V reader, it seems no one needed SPIR-V convergent attribute analog during translation of group/subgroup opcodes, but somehow barrier was lost in this logic.

MrSidims: Okay, so I'd say we should either add 'workaround' to other parts of SPIRVReader or to remove it here? Well, I don't mind to remove it here :)

KornevNikita:
@MrSidims @Fznamznon @AlexeySachkov @mlychkov please take a look

MrSidims:
@KornevNikita please rebase the PR

MrSidims (approved):
LGTM
But I believe, that @Fznamznon concern wasn't addressed. I'm okay to remove W/A string.

PR #1196, opened 2021-09-03
Don't try to apply aliasing decoration to SPIR-V instruction without …

MrSidims:
…return id

Signed-off-by: Dmitry Sidorov <dmitry.sidorov@intel.com>

PR #1203, opened 2021-09-07
[Backport] Don't try to apply aliasing decoration to SPIR-V instruction without …

MrSidims:
…return id (#1196)


MrSidims:
FYI @DmitryBushev

MrSidims:
No need in this patch

PR #1204, opened 2021-09-09
Fix duplication of global annotation strings in reverse translation

KornevNikita:
This is a patch to avoid duplication of global annotation strings in reverse
translation caused by the fact that translator translates each
UserSemantic decoration, even if a similar decoration with the same string
has also been translated.

### DIFF
--- a/lib/SPIRV/SPIRVReader.cpp
+++ b/lib/SPIRV/SPIRVReader.cpp
@@ -3400,9 +3400,9 @@ void SPIRVToLLVM::transIntelFPGADecorations(SPIRVValue *BV, Value *V) {
     SmallString<256> AnnotStr;
     generateIntelFPGAAnnotation(BV, AnnotStr);
     if (!AnnotStr.empty()) {
-      Constant* GS = nullptr;
+      Constant *GS = nullptr;
       std::string StringAnnotStr = AnnotStr.c_str();
-      static unordered_map<std::string, Constant*> Annotations;
+      static unordered_map<std::string, Constant *> Annotations;
       if (Annotations.find(StringAnnotStr) != Annotations.end())
         GS = Annotations[StringAnnotStr];
       else {
--- a/lib/SPIRV/SPIRVReader.cpp
+++ b/lib/SPIRV/SPIRVReader.cpp
@@ -3402,12 +3402,11 @@ void SPIRVToLLVM::transIntelFPGADecorations(SPIRVValue *BV, Value *V) {
     if (!AnnotStr.empty()) {
       Constant *GS = nullptr;
       std::string StringAnnotStr = AnnotStr.c_str();
-      static unordered_map<std::string, Constant *> Annotations;
-      if (Annotations.find(StringAnnotStr) != Annotations.end())
-        GS = Annotations[StringAnnotStr];
-      else {
+      if (AnnotationsMap.find(StringAnnotStr) != AnnotationsMap.end()) {
+        GS = AnnotationsMap[StringAnnotStr];
+      } else {
         GS = Builder.CreateGlobalStringPtr(AnnotStr);
-        Annotations.insert(std::make_pair(StringAnnotStr, GS));
+        AnnotationsMap.insert(std::make_pair(StringAnnotStr, GS));
       }
 
       Value *BaseInst =
--- a/lib/SPIRV/SPIRVReader.cpp
+++ b/lib/SPIRV/SPIRVReader.cpp
@@ -3400,7 +3400,15 @@ void SPIRVToLLVM::transIntelFPGADecorations(SPIRVValue *BV, Value *V) {
     SmallString<256> AnnotStr;
     generateIntelFPGAAnnotation(BV, AnnotStr);
     if (!AnnotStr.empty()) {
-      auto *GS = Builder.CreateGlobalStringPtr(AnnotStr);
+      Constant* GS = nullptr;
+      std::string StringAnnotStr = AnnotStr.c_str();
+      static unordered_map<std::string, Constant*> Annotations;
+      if (Annotations.find(StringAnnotStr) != Annotations.end())
+        GS = Annotations[StringAnnotStr];
+      else {
+        GS = Builder.CreateGlobalStringPtr(AnnotStr);
+        Annotations.insert(std::make_pair(StringAnnotStr, GS));
+      }
 
       Value *BaseInst =
           AL ? Builder.CreateBitCast(V, Int8PtrTyPrivate, V->getName()) : Inst;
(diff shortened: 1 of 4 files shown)


KornevNikita:
@MrSidims @AlexeySachkov @Fznamznon @mlychkov please take a look

lib/SPIRV/SPIRVReader.cpp
```diff
@@ -3400,7 +3400,15 @@ void SPIRVToLLVM::transIntelFPGADecorations(SPIRVValue *BV, Value *V) {
     SmallString<256> AnnotStr;
     generateIntelFPGAAnnotation(BV, AnnotStr);
     if (!AnnotStr.empty()) {
-      auto *GS = Builder.CreateGlobalStringPtr(AnnotStr);
+      Constant *GS = nullptr;
+      std::string StringAnnotStr = AnnotStr.c_str();
+      static unordered_map<std::string, Constant *> Annotations;
```
Fznamznon, line 3405: In general `static` variables are not appreciated (for example in LLVM coding guidelines https://llvm.org/docs/CodingStandards.html#do-not-use-static-constructors ), because order of their initialization is not defined by C++ standard. Can we for example make this map a field of `SPIRVReader`?

lib/SPIRV/SPIRVReader.cpp
```diff
@@ -3400,7 +3400,15 @@ void SPIRVToLLVM::transIntelFPGADecorations(SPIRVValue *BV, Value *V) {
...
+      if (Annotations.find(StringAnnotStr) != Annotations.end())
+        GS = Annotations[StringAnnotStr];
+      else {
```
Fznamznon, line 3408: Just to be consistent with the `else` block.
```suggestion
      if (Annotations.find(StringAnnotStr) != Annotations.end()) {
        GS = Annotations[StringAnnotStr];
      } else {
```

lib/SPIRV/SPIRVReader.cpp
KornevNikita: Check new commit please, This is what you're talking about?

lib/SPIRV/SPIRVReader.cpp
```diff
@@ -3400,7 +3400,14 @@ void SPIRVToLLVM::transIntelFPGADecorations(SPIRVValue *BV, Value *V) {
...
     if (!AnnotStr.empty()) {
-      auto *GS = Builder.CreateGlobalStringPtr(AnnotStr);
+      Constant *GS = nullptr;
+      std::string StringAnnotStr = AnnotStr.c_str();
+      if (AnnotationsMap.find(StringAnnotStr) != AnnotationsMap.end()) {
+        GS = AnnotationsMap[StringAnnotStr];
+      } else {
+        GS = Builder.CreateGlobalStringPtr(AnnotStr);
+        AnnotationsMap.insert(std::make_pair(StringAnnotStr, GS));
+      }
```
mlychkov, line 3410:
```suggestion
      std::string StringAnnotStr = AnnotStr.c_str();
      auto AnnotItr = AnnotationsMap.find(StringAnnotStr);
      if (AnnotItr != AnnotationsMap.end()) {
        GS = AnnotItr->second();
      } else {
        GS = Builder.CreateGlobalStringPtr(AnnotStr);
        AnnotationsMap.emplace(std::move(StringAnnotStr), GS);
      }
```

lib/SPIRV/SPIRVReader.cpp
Fznamznon: Yup

Fznamznon (approved):

lib/SPIRV/SPIRVReader.cpp
```diff
@@ -3400,7 +3400,15 @@ void SPIRVToLLVM::transIntelFPGADecorations(SPIRVValue *BV, Value *V) {
     SmallString<256> AnnotStr;
     generateIntelFPGAAnnotation(BV, AnnotStr);
     if (!AnnotStr.empty()) {
-      auto *GS = Builder.CreateGlobalStringPtr(AnnotStr);
+      Constant *GS = nullptr;
+      std::string StringAnnotStr = AnnotStr.c_str();
+      auto AnnotItr = AnnotationsMap.find(StringAnnotStr);
+      if (AnnotItr != AnnotationsMap.end()) {
+        GS = AnnotItr->second();
```
MrSidims, line 3407:
```suggestion
        GS = AnnotItr->second;
```

MrSidims (approved):
Please fix the build error

lib/SPIRV/SPIRVReader.h
```diff
@@ -159,6 +159,7 @@ class SPIRVToLLVM {
   SPIRVToLLVMPlaceholderMap PlaceholderMap;
   std::unique_ptr<SPIRVToLLVMDbgTran> DbgTran;
   std::vector<Constant *> GlobalAnnotations;
```
mlychkov, line 161: It would be nice to have a comment about GlobalAnnotations and AnnotationsMap fields to understand what is the difference between them.

KornevNikita: Done. Looks better now?

mlychkov: Ok, let me rephrase this: May we join GlobalAnnotations and AnnotationsMap? And why if we cann't?

KornevNikita: Do you mean to use only one container instead of these two?

mlychkov: Yes.

lib/SPIRV/SPIRVReader.h
```diff
@@ -159,6 +159,15 @@ class SPIRVToLLVM {
...
```
mlychkov, line 161:
```suggestion
  // GlobalAnnotations vector collects structures of
  // annotations arguments for global variables and functions.
  // They are used in translation of llvm.global.annotations instruction.
  std::vector<Constant *> GlobalAnnotations;
```

lib/SPIRV/SPIRVReader.h
```diff
@@ -159,6 +159,15 @@ class SPIRVToLLVM {
...
+  // There is also GlobalAnnotations vector which is used to store arrays of
+  // arguments for llvm.global.annotations instructions. But SPIR-V may also
+  // contain annotations for llvm.ptr.annotation.
+  // During reverse translation we need to check if a similar annotation with
+  // the same string has already been translated to avoid duplication of these
+  // global strings in LLVM-IR, which can be caused by multiple translation of
+  // UserSemantic decorations with the same literal. This map is used to store
+  // global strings by their annotations.
+  std::unordered_map<std::string, Constant *> AnnotationsMap;
```
mlychkov, line 170:
```suggestion
  // AnnotationsMap helps to translate annotation strings for local variables.
  // Map values are pointers on global strings in LLVM-IR.
  // It is used to avoid duplication of these annotation strings in LLVM-IR,
  // which can be caused by multiple translation of
  // UserSemantic decorations with the same literal.
  std::unordered_map<std::string, Constant *> AnnotationsMap;
```

lib/SPIRV/SPIRVReader.h
KornevNikita: Done

mlychkov (approved):
LGTM, thank you.

PR #1213, opened 2021-09-16
Fix for a bunch of fixed point integer SPIR-V instructions

pauzinl:
There could be cases when return type of intrinsic is greater than 64 moves return value into argument list as sret parameter.

TODO: currently sret parameter attribute is being lost during reverse translation for the mentioned case, need to fix it.

### DIFF
--- a/lib/SPIRV/SPIRVWriter.cpp
+++ b/lib/SPIRV/SPIRVWriter.cpp
@@ -4127,15 +4127,38 @@ LLVMToSPIRVBase::transBuiltinToInstWithoutDecoration(Op OC, CallInst *CI,
     // Literal S Literal I Literal rI Literal Q Literal O
 
     Type *ResTy = CI->getType();
-    SPIRVValue *Input =
-        transValue(CI->getOperand(0) /* A - integer input of any width */, BB);
+    SPIRVValue *Input;
+    std::vector<Value *> Operands(5);
 
-    std::vector<Value *> Operands = {
-        CI->getOperand(1) /* S - bool value, indicator of signedness */,
-        CI->getOperand(2) /* I - location of the fixed-point of the input */,
-        CI->getOperand(3) /* rI - location of the fixed-point of the result*/,
-        CI->getOperand(4) /* Quantization mode */,
-        CI->getOperand(5) /* Overflow mode */};
+    // If return type of intrinsic is greater than 64 then FE moves return
+    // variable into arguments list
+    if (CI->hasStructRetAttr()) {
+      assert(ResTy->isVoidTy() && "Return type is not void");
+      ResTy = cast<PointerType>(CI->getOperand(0)->getType())->getElementType();
+      Input = transValue(CI->getOperand(1) /* A - integer input of any width */,
+                         BB);
+
+      Operands[0] =
+          CI->getOperand(2); /* S - bool value, indicator of signedness */
+      Operands[1] =
+          CI->getOperand(3); /* I - location of the fixed-point of the input */
+      Operands[2] =
+          CI->getOperand(4); /* rI - location of the fixed-point of the result*/
+      Operands[3] = CI->getOperand(5); /* Quantization mode */
+      Operands[4] = CI->getOperand(6); /* Overflow mode */
+    } else {
+      Input = transValue(CI->getOperand(0) /* A - integer input of any width */,
+                         BB);
+
+      Operands[0] =
+          CI->getOperand(1); /* S - bool value, indicator of signedness */
+      Operands[1] =
+          CI->getOperand(2); /* I - location of the fixed-point of the input */
+      Operands[2] =
+          CI->getOperand(3); /* rI - location of the fixed-point of the result*/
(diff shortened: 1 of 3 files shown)


pauzinl:
Hi, @MrSidims, @mlychkov. Could you please review this fix?

MrSidims:
Title comment: sincos_fixed() - is a SYCL math API. Here you bring a fix for a bunch of fixed point integer SPIR-V instructions.
Please also add a description of the bug and a test.

MrSidims (commented):
I believe, we should face the same issue for `ArbitraryFloat` instructions as well. I suggest to fix it simultaneously. Also some code from `ArbitraryFloat` can be adopted for fixed integer point instructions, but better to sync with @mlychkov about it.

lib/SPIRV/SPIRVWriter.cpp
```diff
@@ -4127,15 +4127,38 @@ LLVMToSPIRVBase::transBuiltinToInstWithoutDecoration(Op OC, CallInst *CI,
...
-        CI->getOperand(2) /* I - location of the fixed-point of the input */,
-        CI->getOperand(3) /* rI - location of the fixed-point of the result*/,
-        CI->getOperand(4) /* Quantization mode */,
-        CI->getOperand(5) /* Overflow mode */};
+    SPIRVValue *Input;
+    std::vector<Value *> Operands(5);
+
+    // If return type of intrinsic is greater than 64 then FE moves return
+    // variable into arguments list
+    if (CI->hasStructRetAttr()) {
```
MrSidims, line 4135: To avoid code duplication I suggest to introduce extra variable:
`uint32_t FirstOperand = 0;`

If we have 'sret' attribute, under
`if (CI->hasStructRetAttr()) {`
we can set it to be
`FirstOperand = 1;`
later it can be used like this:
```
      Operands[0] =
          CI->getOperand(1 + FirstOperand); /* S - bool value, indicator of signedness */
```
so we will have a unified code for 'sret' and 'no sret' cases.

MrSidims: but yet some lambda magic from `ArbitraryFloat` handling can (should?) be embraced here.
```
    std::transform(OpItr, OpEnd, std::back_inserter(Literals), [](auto *O) {
      return cast<llvm::ConstantInt>(O)->getZExtValue();
    });
```

MrSidims (commented):
Functionally and test-wise LGTM, thanks!
I just have some re-wording suggestion, I'll post next.

lib/SPIRV/SPIRVWriter.cpp
```diff
@@ -4127,19 +4127,25 @@ LLVMToSPIRVBase::transBuiltinToInstWithoutDecoration(Op OC, CallInst *CI,
     // Literal S Literal I Literal rI Literal Q Literal O
 
     Type *ResTy = CI->getType();
+
+    auto OpItr = CI->value_op_begin();
+    auto OpEnd = OpItr + CI->getNumArgOperands();
+
+    // If return type of intrinsic is greater than 64 then FE moves return
+    // variable into arguments list
```
MrSidims, line 4135:
```suggestion
    // If the return type of an instruction is wider than 64-bit, then this instruction will return
    // via 'sret' argument added into the arguments list. Here we reverse this, removing 'sret'
    // argument and restoring the original return type.
```

MrSidims (commented):
Suggesting in the code comment is also applicable to the PR comment.

lib/SPIRV/SPIRVWriter.cpp
```diff
@@ -4127,42 +4127,25 @@ LLVMToSPIRVBase::transBuiltinToInstWithoutDecoration(Op OC, CallInst *CI,
...
-    std::vector<Value *> Operands(5);
+
+    auto OpItr = CI->value_op_begin();
+    auto OpEnd = OpItr + CI->getNumArgOperands();
 
     // If return type of intrinsic is greater than 64 then FE moves return
     // variable into arguments list
     if (CI->hasStructRetAttr()) {
       assert(ResTy->isVoidTy() && "Return type is not void");
       ResTy = cast<PointerType>(CI->getOperand(0)->getType())->getElementType();
```
mlychkov, line 4138: May we use OpItr here instead of getOperand?

mlychkov (approved):
Thank you.

lib/SPIRV/SPIRVWriter.cpp
pauzinl: Sure. Changed it in [80b3711](https://github.com/KhronosGroup/SPIRV-LLVM-Translator/pull/1213/commits/80b371198f9d197467986749273b905ffeb3dbff)

lib/SPIRV/SPIRVWriter.cpp
pauzinl: Changed comments in [80b3711](https://github.com/KhronosGroup/SPIRV-LLVM-Translator/pull/1213/commits/80b371198f9d197467986749273b905ffeb3dbff)

MrSidims (approved):

PR #1214, opened 2021-09-17
Address some static analyzer concerns

MrSidims:
The assertions are actually not required here, since all these
functions are being called from visitCallInst, which has the
appropriate check. But for the code consistency lets do not
trust functions' inputs.

### DIFF
--- a/lib/SPIRV/SPIRVToOCL.cpp
+++ b/lib/SPIRV/SPIRVToOCL.cpp
@@ -582,6 +582,7 @@ void SPIRVToOCLBase::visitCallSPIRVImageMediaBlockBuiltin(CallInst *CI, Op OC) {
 }
 void SPIRVToOCLBase::visitCallBuildNDRangeBuiltIn(CallInst *CI, Op OC,
                                                   StringRef DemangledName) {
+  assert(CI->getCalledFunction() && "Unexpected indirect call");
   AttributeList Attrs = CI->getCalledFunction()->getAttributes();
   mutateCallInstOCL(
       M, CI,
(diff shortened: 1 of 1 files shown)


MrSidims:
@mlychkov @pauzinl please take a look

pauzinl (approved):

lib/SPIRV/SPIRVToOCL.cpp
```diff
@@ -582,6 +582,7 @@ void SPIRVToOCLBase::visitCallSPIRVImageMediaBlockBuiltin(CallInst *CI, Op OC) {
 }
 void SPIRVToOCLBase::visitCallBuildNDRangeBuiltIn(CallInst *CI, Op OC,
                                                   StringRef DemangledName) {
+  assert(CI->getCalledFunction() && "Unexpected indirect call");
```
mlychkov, line 585: As I see in almost all cases Attrs is requested here and passed into mutateCallInstOCL function.
May we add mutateCallInstOCL version without Attrs parameter and initialize this parameter there with the assert check to reduce code duplication?

Fznamznon: `visitCallBuildNDRangeBuiltIn` is called from `visitCallInst` which early exits in case the called function is `nullptr`.

lib/SPIRV/SPIRVToOCL.cpp
```diff
@@ -611,6 +612,7 @@ void SPIRVToOCLBase::visitCallBuildNDRangeBuiltIn(CallInst *CI, Op OC,
 
 void SPIRVToOCLBase::visitCallGenericCastToPtrExplicitBuiltIn(CallInst *CI,
                                                               Op OC) {
+  assert(CI->getCalledFunction() && "Unexpected indirect call");
```
Fznamznon, line 615: Yeah, I think my previous comment works for all the assertions in this patch. They don't seem to be necessary.

MrSidims: Indeed, but please check the PR description :)
There are several visit ... functions in the file, some of them have the assertions, others don't. So I'd say we'd better to either remove all of them or add new assertions. I'm kind of on adding extra, to express 'not-trust-function inputs' strategy. But yeah, all of these assertions aren't needed actually.

AlexeySotkin (approved):

PR #1138, opened 2021-08-02
Add validation of vloada_half instruction

MrSidims:
The instruction is for vector types and not for scalar types.
It's a kind of a follow up for:
https://github.com/KhronosGroup/OpenCL-CTS/issues/1002

Signed-off-by: Dmitry Sidorov <dmitry.sidorov@intel.com>

PR #1215, opened 2021-09-20
Revert "Add validation of vloada_half instruction (#1138)"

StuartDBrady:
This reverts commit 1a7f1420db4dc21a9628b85e9c06073c718ebfc4.

The `vloada_halfn` functions return `float` vectors:

* `floatn vloada_halfn(size_t offset, const __global half *p)`
* `floatn vloada_halfn(size_t offset, const __local half *p)`
* `floatn vloada_halfn(size_t offset, const __constant half *p)`
* `floatn vloada_halfn(size_t offset, const __private half *p)`

- For the generic address space:
`floatn vloada_halfn(size_t offset, const half *p)`

These functions exist specifically for implementations that lack the `cl_khr_fp16` extension, and so do not include the `halfn` vector types, but do include a scalar `half` type for use as a pointer argument _only_.

Likewise, in the _SPIR-V OpenCL Extended Instruction Set_, the result type for `vloada_halfn` is defined as follows:

> _Result Type_ must be _vector(2,3,4,8,16)_ of _float_ values.

The assertion added to `isBuiltinTransToExtInst(...)` is therefore incorrect.

The `vloada_halfn` SPIR-V OpenCL Extended Instruction Set instructions that are emitted for calls to these builtins are accepted by the SPIR-V Validator.


StuartDBrady:
Looking more closely, I notice that original assertion in `LLVMToSPIRVBase::isBuiltinTransToExtInst(...)` in `SPIRVWriter.cpp` checks that the return type is a `half` vector type, but these functions return `float` vectors.

StuartDBrady:
Some background, from the original description for this PR:

The `vloada_halfn` functions take pointers to scalar `half` values:

* `floatn vloada_halfn(size_t offset, const __global half *p)`
* `floatn vloada_halfn(size_t offset, const __local half *p)`
* `floatn vloada_halfn(size_t offset, const __constant half *p)`
* `floatn vloada_halfn(size_t offset, const __private half *p)`

- For the generic address space:
`floatn vloada_halfn(size_t offset, const half *p)`

These functions exist specifically for implementations that lack the `cl_khr_fp16` extension, and therefore do not include the `halfn` vector types, but do include a scalar `half` type for use as a pointer argument *only*.  With the `cl_khr_fp16` extension, `vloada_halfn` is not needed:

```
   float2 alt_vloada_half2(size_t offset, const __private half *p)
   {
      const __private half2 *p2 = (const __private half2 *)p + offset;
      return convert_floatn(*p2);
   }
```

Likewise, in the _SPIR-V OpenCL Extended Instruction Set_, `vloada_halfn` is defined as taking an <id> _p_, such that:

> _p_ must be a pointer(global, local, private, constant, generic) to `half`.

Note that more generally, the entire purpose for all of the `vload*` and `vstore*` functions is to allow operations taking a pointer to the scalar type where the vector type can/should not be used directly, due to alignment constraints or limited availability of vector types.

StuartDBrady:
The main reason for me wanting to revert this rather than trying to fix it is that the negative tests make no sense whatsoever to me: the first argument to `vloada_halfn` is a `size_t` offset, and I cannot understand what these tests are trying to do with `half %val` or `<2 x float> %val`, and what this has to do with the assertions being added.

MrSidims (approved):

StuartDBrady:
(The latest force push was simply a minor change to the commit message, to put back some information that I had previously removed.)

StuartDBrady:
(The latest force push is just a rebase, intended to show tests passing.)

MrSidims (approved):

PR #1220, opened 2021-09-22
Clean up translation of islessgreater() and OpLessOrGreater

StuartDBrady:
Translate `islessgreater()` to _OpFOrdNotEqual_ instead of _OpLessOrGreater_, which is deprecated as of SPIR-V 1.5 revision 2, and has identical semantics.

Translate _OpLessOrGreater_ to `fcmp one` instead of `islessgreater()`, which has identical semantics.

### DIFF
--- a/lib/SPIRV/OCLUtil.cpp
+++ b/lib/SPIRV/OCLUtil.cpp
@@ -293,7 +293,7 @@ template <> void SPIRVMap<std::string, Op, SPIRVInstruction>::init() {
   _SPIRV_OP(isgreaterequal, FOrdGreaterThanEqual)
   _SPIRV_OP(isless, FOrdLessThan)
   _SPIRV_OP(islessequal, FOrdLessThanEqual)
-  _SPIRV_OP(islessgreater, LessOrGreater)
+  _SPIRV_OP(islessgreater, FOrdNotEqual)
   _SPIRV_OP(isordered, Ordered)
   _SPIRV_OP(isunordered, Unordered)
   _SPIRV_OP(isfinite, IsFinite)
--- a/lib/SPIRV/SPIRVReader.cpp
+++ b/lib/SPIRV/SPIRVReader.cpp
@@ -598,11 +598,6 @@ SPIRVToLLVM::transValue(const std::vector<SPIRVValue *> &BV, Function *F,
   return V;
 }
 
-bool SPIRVToLLVM::isSPIRVCmpInstTransToLLVMInst(SPIRVInstruction *BI) const {
-  auto OC = BI->getOpCode();
-  return isCmpOpCode(OC) && OC != OpLessOrGreater;
-}
-
 void SPIRVToLLVM::setName(llvm::Value *V, SPIRVValue *BV) {
   auto Name = BV->getName();
   if (!Name.empty() && (!V->hasName() || Name != V->getName()))
@@ -1113,6 +1108,9 @@ Value *SPIRVToLLVM::transCmpInst(SPIRVValue *BV, BasicBlock *BB, Function *F) {
   if (BB) {
     Builder.SetInsertPoint(BB);
   }
+
+  if (OP == OpLessOrGreater)
+    OP = OpFOrdNotEqual;
 
   if (BT->isTypeVectorOrScalarInt() || BT->isTypeVectorOrScalarBool() ||
       BT->isTypePointer())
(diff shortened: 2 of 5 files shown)


StuartDBrady:
@DmitryBushev, here are the changes for OpLessOrGreater that I mentioned, if you would be interested in reviewing them.

DmitryBushev (approved):

MrSidims (approved):

MrSidims:
LGTM but as a side note:

> SPIR-V 1.5 revision 2

Currently max supported version for the translator is SPIR-V 1.4

MrSidims:
Withdraw my approve.
The patch is still LGTM, but I do believe, that the translator should either claim to support the latest SPIR-V version and generate the appropriate SPIR-V or, generate different SPIR-V depending on max supported SPIR-V version.

MrSidims:
May be the logic should be like this:
**Default** SPIR-V version _should be set_ as 1.4 or 1.5 (now it's 1.0) 
(if the **Default** SPIR-V version is 1.5) IF _--spirv-max-version_ is not set OR IF it's set as 1.5 or higher
THEN generate **OpFOrdNotEqual**
ELSE generate **OpLessOrGreater**

same logic should be applicable for any other deprecated instruction.

StuartDBrady:
> LGTM but as a side note:
> 
> > SPIR-V 1.5 revision 2
> 
> Currently max supported version for the translator is SPIR-V 1.4

Yes, I'm aware of that – I've submitted this ahead of time, rather than waiting until after we increase the maximum SPIR_V version.

> Withdraw my approve.
> The patch is still LGTM, but I do believe, that the translator should either claim to support the latest SPIR-V version and generate the appropriate SPIR-V or, generate different SPIR-V depending on max supported SPIR-V version option.

Why should this generate "different SPIR-V"?  What is wrong with the SPIR-V that it generates?

> May be the logic should be like this:
> **Default** SPIR-V version _should be set_ as 1.4 or 1.5 (now it's 1.0)
> (if the **Default** SPIR-V version is 1.5) IF _--spirv-max-version_ is not set OR IF it's set as 1.5 or higher
> THEN generate **OpFOrdNotEqual**
> ELSE generate **OpLessOrGreater**
> 
> same logic should be applicable for any other deprecated instruction.

_OpFOrdNotEqual_ has identical semantics to _OpLessOrGreater_.  _OpFOrdNotEqual_ is available in every version of SPIR-V, is not deprecated, and as a minor detail, does not require the _Kernel_ capability.  _OpLessOrGreater_ is _Kernel_-specific and deprecated (meaning you shouldn't use it for any version of SPIR-V given a semantically equivalent alternative).  The reason we deprecated _OpLessOrGreater_ is that there is (and was) no reason ever to use it.

Yes, SPIR-V 1.5 revision 2 is the document that deprecated it, but it is a unified specification and did so for all versions of SPIR-V.

MrSidims (approved):
My bad. I was thinking about general instruction deprecation/replacement process and what would happen if we map some LLVM IR construct on a certain SPIR-V instruction before, for example, SPIR-V 1.4 and on another instruction after SPIR-V 1.4, and this ew instruction was added after SPIR-V 1.4 as well.

StuartDBrady:
> My bad. I was thinking about general instruction deprecation/replacement process and what would happen if we map some LLVM IR construct on a certain SPIR-V instruction before, for example, SPIR-V 1.4 and on another instruction after SPIR-V 1.4, and this new instruction was added after SPIR-V 1.4 as well.

No worries.  I agree that this would be a really good thing to have in place _before_ we need it!  There are some cases where older SPIR-V versions do not have the semantics required for correct translation and our handling of this is not graceful (and others for which it may be _too_ graceful) which I may raise in the coming weeks, as time permits.

StuartDBrady:
The only change since the previous version is that the commit message has been updated to more clearly state that _OpFOrdNotEqual_ and _OpLessOrGreater_ have identical semantics, and that `islessgreater()` and `fcmp one` have identical semantics (in both cases, ignoring SNaNs as they are not supported).

hewj03:
@StuartDBrady this commit breaks standalone use of createSPIRVToOCL20Legacy pass because OpLessOrGreater can't be found in OCLSPIRVBuiltinMap any more.
Can we avoid generating OpLessOrGreater at all since it is deprecated? E.g. _Z21__spirv_LessOrGreaterDv16_dS_ will have OpFOrdNotEqual instead of OpLessOrGreater. This will change spirv binary, which might brings risk.
Then OpLessOrGreater won't be used in LLVM IR -> SPIRV -> LLVM IR pipeline, although it would still be needed for translating old spirv binaries that contains OpLessOrGreater.
This approach can fix standalone use of createSPIRVToOCL20Legacy pass. How do you think?

PR #1221, opened 2021-09-23
Fix mistranslation of OpAtomicCompareExchangeWeak

StuartDBrady:
The semantics of _OpAtomicCompareExchangeWeak_ are the same as those of _OpAtomicCompareExchange_. Each of the ops should be translated to `atomic_compare_exchange_strong_explicit()` when using OpenCL 2.0 builtins, and not `atomic_compare_exchange_weak_explicit()`, as use of the latter may result in unreported spurious failures which are not detectable when using _OpAtomicCompareExchangeWeak_, which does not give a direct indication of its success.

### DIFF
--- a/lib/SPIRV/SPIRVToOCL.h
+++ b/lib/SPIRV/SPIRVToOCL.h
@@ -187,7 +187,7 @@ public:
 
   /// Transform __spirv_OpAtomicCompareExchange and
   /// __spirv_OpAtomicCompareExchangeWeak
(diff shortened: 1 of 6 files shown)


svenvh (commented):
LGTM, but would be good to have approval from someone else too.

MrSidims (approved):

StuartDBrady:
I have raised issue #1230 to address the translation from LLVM IR to SPIR-V that includes _OpAtomicCompareExchangeWeak_, was removed in SPIR-V 1.4 and is deprecated for all prior versions.

PR #1223, opened 2021-09-24
Set generator ID in generated SPIR-V

djdeath:
This is useful they you want to identify a particular version of a generator to apply workarounds.

### DIFF
--- a/lib/SPIRV/SPIRVWriter.cpp
+++ b/lib/SPIRV/SPIRVWriter.cpp
@@ -3438,6 +3438,7 @@ bool isEmptyLLVMModule(Module *M) {
 }
 
 bool LLVMToSPIRVBase::translate() {
+  BM->setGeneratorId(kTranslatorId);
   BM->setGeneratorVer(KTranslatorVer);
 
   if (isEmptyLLVMModule(M))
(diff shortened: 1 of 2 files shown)


lib/SPIRV/SPIRVWriter.cpp
```diff
@@ -3438,6 +3438,7 @@ bool isEmptyLLVMModule(Module *M) {
 }
 
 bool LLVMToSPIRVBase::translate() {
+  BM->setGeneratorId(kTranslatorId);
```
AlexeySotkin, line 3441: Please note that we already set generator id in `SPIRVModuleImpl`'s constructor: https://github.com/KhronosGroup/SPIRV-LLVM-Translator/blob/08884df4c33aeb3ff36bf39a21a14e5fb37d905a/lib/SPIRV/libSPIRV/SPIRVModule.cpp#L71

Having said that we probably should refactor to align setting translator version and translator id: both in the constructor or both here in the `translate` method.

djdeath: Thanks for pointing this out. I may I have missed something out then.

My workflow involves LLVM-SPIRV-Translator -> SPIRV-Tools-Linker. Turns out the linker doesn't set the generator ID correctly so might be the issue I'm running into.

Looks like SPIRVModule should take care of this already.

djdeath:
As @AlexeySotkin pointed out, this should be set already.

PR #1225, opened 2021-09-26
Use StorageClassConstant for constant global variables

llandwerlin-intel:
The following OpenCL code :

```
   struct DummyStruct {
      uint a;
   };
   const uint my_var = 42;

   kernel void
   test(global struct DummyStruct *a_ptr)
   {
      a_ptr->a = my_var;
   }
```

Currently generates this :

```
                            OpDecorate %my_var Constant
                 %uint_42 = OpConstant %uint 42
%_ptr_CrossWorkgroup_uint = OpTypePointer CrossWorkgroup %uint
                  %my_var = OpVariable %_ptr_CrossWorkgroup_uint CrossWorkgroup %uint_42
```

But those global constant variables could be turned into
UniformConstant.

Signed-off-by: Lionel Landwerlin <lionel.g.landwerlin@intel.com>

### DIFF
--- a/lib/SPIRV/SPIRVWriter.cpp
+++ b/lib/SPIRV/SPIRVWriter.cpp
@@ -1466,7 +1466,11 @@ LLVMToSPIRVBase::transValueWithoutDecoration(Value *V, SPIRVBasicBlock *BB,
     bool IsVectorCompute =
         BM->isAllowedToUseExtension(ExtensionID::SPV_INTEL_vector_compute) &&
         GV->hasAttribute(kVCMetadata::VCGlobalVariable);
-    if (IsVectorCompute)
+    if (AddressSpace == SPIRAS_Global &&
+        AddressSpace == GV->isConstant()) {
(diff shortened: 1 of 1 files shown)


llandwerlin-intel:
My first iteration targeted all types of variable. This new one only applies to global variables.

jenatali:
I believe that this breaks semantics for kernel arguments, which is a no-no since those need to match application's expectations. E.g. you can have a `__global const foo *` as a kernel arg, and [requesting the argument info out of the CL API](https://www.khronos.org/registry/OpenCL/specs/3.0-unified/html/OpenCL_API.html#clGetKernelArgInfo) needs to reflect that it's global (`CL_KERNEL_ARG_ADDRESS_GLOBAL`) and const (`CL_KERNEL_ARG_TYPE_CONST`).

llandwerlin-intel:
Thanks @jenatali. Closing, I can't see a good way to do this in the translator atm.

PR #1227, opened 2021-09-30
Add an option to recognize that the input file is already LLVM

KornevNikita:
This is a patch to introduce new option which, when used with -r, allows
llvm-spirv to recognize if the input file is already LLVM and copy it to
output file. If the input file is SPIR-V it will run normal conversion.


KornevNikita:
I assume that there may be a better way to detect if input is LLVM/SPIR-R and I would be glad if someone could give me some advice.

KornevNikita:
@AlexeySotkin @AlexeySachkov @MrSidims @mlychkov take a look, please

test/IR_passthrough.ll
```diff
@@ -0,0 +1,61 @@
+; RUN: llvm-as < %s -o %t.bc
+; RUN: llvm-spirv -r -ir-passthrough %t.bc -o %t.ir-passthrough.bc
+; RUN: llvm-dis %t.ir-passthrough.bc -o %t.ir-passthrough.ll
+; RUN: FileCheck %s --input-file %t.ir-passthrough.ll --check-prefix CHECK-LLVM-COPY
+
+; ModuleID = 'souce.cpp'
+; CHECK-LLVM-COPY: source_filename = "souce.cpp"
+; CHECK-LLVM-COPY: target datalayout = "e-i64:64-v16:16-v24:32-v32:32-v48:64-v96:128-v192:256-v256:256-v512:512-v1024:1024-n8:16:32:64"
```
MrSidims, line 8:
```suggestion
; CHECK-LLVM-COPY-NEXT: target datalayout = "e-i64:64-v16:16-v24:32-v32:32-v48:64-v96:128-v192:256-v256:256-v512:512-v1024:1024-n8:16:32:64"
```
same is applicable below as well

tools/llvm-spirv/llvm-spirv.cpp
```diff
@@ -382,6 +389,57 @@ static int regularizeLLVM(SPIRV::TranslatorOpts &Opts) {
   return 0;
 }
 
+static int copyInputLLVMToOutput() {
+  LLVMContext Context;
```
MrSidims, line 393: May be it's worth adding `LLVM_DEBUG(dbgs() <<` for significant steps.

AlexeySachkov:
> I'd like to understand the motivation behind this option a bit better, because it's adding quite a bit of code for optionally doing nothing, essentially. I don't know how you integrate the translator on your side, but if you have LLVM IR coming in, can you simply not invoke the translator at all as an alternative solution?

Tagging @AGindinson @mdtoguchi here - they should be able to explain the integration better than me.

On high-level, it looks like this: in DPC++ toolchain (intel/llvm GitHub repo) we sometimes do some batch processing of files through clang driver, i.e. some of our tools generates a number of files, which are then passed to another tool one by one, or something like this.

If I understand correctly, some new functionality leads to situations where we can produce both SPIR-V and LLVM IR files into a single list of files to process. Since clang driver can't dynamically decide which tool to launch - the list of commands to launch and their argument is statically built at the beginning and can only be blindly executed later. Therefore, we would like to be able to launch the translator on all files from that list, but we would like it to not fail if there is nothing to do: like if we try to convert to LLVM IR a file, which already contains an LLVM IR.

test/IR_passthrough.ll
KornevNikita: Done

svenvh:
> @AlexeySachkov @AlexeySotkin @svenvh So, we decided to close this PR?

If you can solve the problem on your end in a different way, then *not* merging this PR would be my preference.

PR #1206, opened 2021-09-13
Enforce UserSemantic decoration if no FPGA decorations found

MrSidims:
Signed-off-by: Dmitry Sidorov <dmitry.sidorov@intel.com>

PR #1228, opened 2021-09-30
Don't emit UserSemantic with full annotation str if FPGA decorations found

MrSidims:
https://github.com/KhronosGroup/SPIRV-LLVM-Translator/pull/1206 with a
change of 'AnnotatedDecoration' to 'AnnotatedCode' brought a regression,
when the annotation information started to be duplicated in both
UserSemantic decoration and FPGA-specific decorations. This patch fixes
that.

Yet if there is an annotation string that is split in blocks like Intel
FPGA annotation, it's not necessarily an FPGA annotation. Translate the
whole string as UserSemantic decoration in this case.


MrSidims:
@AlexeySotkin @mlychkov @vmaksimo please take a look

mlychkov (approved):

AlexeySotkin (approved):

PR #1229, opened 2021-10-01
[NFC] Optimize OCLUtil::toString usage in checkError function

vmaksimo:
`toString` function is called to convert an LLVM IR instruction to a string,
mostly in order to include it into an error message in case LLVM IR is invalid.
The problem is that this function is called even when IR is valid which
significantly slows translator's work.

This patch reworks `toString` usage in `checkError` method to generate a string
only in case when LLVM IR is invalid.

### DIFF
--- a/lib/SPIRV/SPIRVWriter.cpp
+++ b/lib/SPIRV/SPIRVWriter.cpp
@@ -1871,9 +1871,8 @@ LLVMToSPIRVBase::transValueWithoutDecoration(Value *V, SPIRVBasicBlock *BB,
     AtomicRMWInst::BinOp Op = ARMW->getOperation();
     if (!BM->getErrorLog().checkError(
             !AtomicRMWInst::isFPOperation(Op) && Op != AtomicRMWInst::Nand,
-            SPIRVEC_InvalidInstruction,
-            toString(V) + "\nAtomic " +
-                AtomicRMWInst::getOperationName(Op).str() +
+            SPIRVEC_InvalidInstruction, V,
+            "Atomic " + AtomicRMWInst::getOperationName(Op).str() +
                 " is not supported in SPIR-V!\n"))
       return nullptr;
 
@@ -3158,10 +3157,10 @@ SPIRVValue *LLVMToSPIRVBase::transIntrinsicInst(IntrinsicInst *II,
       return BM->addInstTemplate(OpSaveMemoryINTEL, BB, Ty);
     }
     BM->getErrorLog().checkError(
-        BM->isUnknownIntrinsicAllowed(II), SPIRVEC_InvalidFunctionCall,
-        toString(II) + "\nTranslation of llvm.stacksave intrinsic requires "
-                       "SPV_INTEL_variable_length_array extension or "
-                       "-spirv-allow-unknown-intrinsics option.");
+        BM->isUnknownIntrinsicAllowed(II), SPIRVEC_InvalidFunctionCall, II,
+        "Translation of llvm.stacksave intrinsic requires "
+        "SPV_INTEL_variable_length_array extension or "
+        "-spirv-allow-unknown-intrinsics option.");
     break;
   }
   case Intrinsic::stackrestore: {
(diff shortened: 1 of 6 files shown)


lib/SPIRV/libSPIRV/SPIRVError.h
```diff
@@ -109,12 +110,33 @@ class SPIRVErrorLog {
...
+inline bool SPIRVErrorLog::checkError(bool Cond, SPIRVErrorCode ErrCode,
+                                      llvm::Instruction *Inst,
+                                      const std::string &Msg,
+                                      const char *CondString,
+                                      const char *FileName, unsigned LineNo) {
+  if (Cond)
+    return Cond;
+  // Do not overwrite previous failure.
+  if (ErrorCode != SPIRVEC_Success)
+    return Cond;
```
svenvh, line 134: Is this not already handled by the `checkError` function that you're calling below?

vmaksimo: Yes, it is, but the idea is that we should not call `toString` method earlier than it's actually needed. So, we need these 2 checks before the call. 
Maybe it can be done better to fully avoid code duplication, but I thought that such approach will result in more code and will worsen readability.

vmaksimo:
Tagging @AlexeySachkov @MrSidims, please take a look

lib/SPIRV/SPIRVWriter.cpp
```diff
@@ -1877,7 +1872,7 @@ LLVMToSPIRVBase::transValueWithoutDecoration(Value *V, SPIRVBasicBlock *BB,
     if (!BM->getErrorLog().checkError(
             !AtomicRMWInst::isFPOperation(Op) && Op != AtomicRMWInst::Nand,
             SPIRVEC_InvalidInstruction,
-            OCLUtil::toString(V) + "\nAtomic " +
+            toString(V) + "\nAtomic " +
```
AlexeySotkin, line 1875:
```suggestion
            V + "\nAtomic " +
```

lib/SPIRV/libSPIRV/SPIRVError.h
AlexeySotkin: Another way to solve the problem might be to move checking of the condition out of this function, so instead of 
```
      getErrorLog().checkError(
          DestAddrSpace != SPIRAS_Constant, SPIRVEC_InvalidModule, U,
          "Casts from generic address space to constant are illegal\n");
```
we would do something like:
```
      if (DestAddrSpace != SPIRAS_Constant)
        getErrorLog().reportErrorForInstruction(SPIRVEC_InvalidModule, U,
          "Casts from generic address space to constant are illegal\n");
```

lib/SPIRV/SPIRVWriter.cpp
vmaksimo: Do you mean such change? It'll not work now as `checkError` accepts an `Instruction`, not `Value`.
```suggestion
            V, "\nAtomic " +
```

AlexeySotkin: Should we change `checkError` to accept `Value`?

vmaksimo: Applied in ff6cabe.

lib/SPIRV/libSPIRV/SPIRVError.h
svenvh: > Yes, it is, but the idea is that we should not call toString method earlier than it's actually needed.

Right, I see your point.  In that case it's worth adding a comment that you're doing the `Cond` check first for speed purposes.  Otherwise someone might simplify the function and remove it again one day.

MrSidims (approved):
LGTM

lib/SPIRV/libSPIRV/SPIRVUtil.h
```diff
@@ -422,6 +424,14 @@ getOrInsert(MapTy &Map, typename MapTy::key_type Key, FuncTy Func) {
   return NF;
 }
 
+template <typename T> std::string toString(const T *Object) {
+  std::string S;
```
AlexeySotkin, line 428: I think it worth checking that `Object` != `nullptr`.

test/negative/atomicrmw-unsupported-operation.ll
```diff
@@ -2,8 +2,8 @@
 ; RUN: not llvm-spirv %t.bc -o %t.spv 2>&1 | FileCheck %s
 
 ; CHECK: InvalidInstruction: Can't translate llvm instruction:
-; CHECK: atomicrmw nand i32 addrspace(1)* @ui, i32 42 acq_rel
 ; CHECK: Atomic nand is not supported in SPIR-V!
+; CHECK: atomicrmw nand i32 addrspace(1)* @ui, i32 42 acq_rel
```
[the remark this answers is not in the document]
vmaksimo: It's related to usage of new `checkError` method. Unfortunately, it doesn't support insertion of instruction IR to any place of error message - now the result of `toString(Inst)` is attached at the end of error message.
There are 2 options how to fix this:
1. to revert this change back https://github.com/KhronosGroup/SPIRV-LLVM-Translator/pull/1229#discussion_r721488847
2. to add a bool flag, which will modify where to place result of `toString()` - at the end or at the beginning.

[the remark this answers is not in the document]
AlexeySotkin: I think it is not a major issue and can be addressed in a separate patch.

PR #1232, opened 2021-10-05
Ignore llvm.instrprof.increment intrinsic

KornevNikita:
This is a workaround to avoid crash when translating
llvm.instrprof.increment which is not supported by SPIR-V and can not be
allowed by related option due to using llvm-spirv as a toolchain node.


KornevNikita:
@AlexeySotkin @MrSidims @AlexeySachkov take a look please, do we need test here?

MrSidims:
> do we need test here?

I'd say yes. See: test/llvm-intrinsics/

lib/SPIRV/SPIRVWriter.cpp
```diff
@@ -3189,6 +3189,8 @@ SPIRVValue *LLVMToSPIRVBase::transIntrinsicInst(IntrinsicInst *II,
   case Intrinsic::invariant_end:
   case Intrinsic::dbg_label:
   case Intrinsic::trap:
+  // llvm.instrprof.increment is not supported
```
AlexeySotkin, line 3192: Please rearrange the lines so comment about `llvm.trap`  appears just before `case Intrinsic::trap`.

lib/SPIRV/SPIRVWriter.cpp
```diff
@@ -3189,6 +3189,8 @@ SPIRVValue *LLVMToSPIRVBase::transIntrinsicInst(IntrinsicInst *II,
...
+  case Intrinsic::instrprof_increment:
```
AlexeySotkin, line 3193: There are more `llvm.instrprof.*` intrinsics. I believe, if we want to add this one we should add the others too.

MrSidims (approved):

AlexeySotkin (approved):

AlexeySotkin:
Probably it is not too late to squash them on the `master` branch if nobody objects :)

PR #1233, opened 2021-10-06
[NFC] Optimize adding of decorations

Fznamznon:
Use std::vector instead of std::multiset with custom comparator.
It is a bit unclear why std::multiset was required since there is no
order requirement for OpDecorate instructions in SPIR-V spec, in addition old
data structure was highly inefficient.


lib/SPIRV/libSPIRV/SPIRVDecorate.h
```diff
@@ -378,17 +350,17 @@ class SPIRVDecorationGroup : public SPIRVEntry {
...
       const_cast<SPIRVDecorateGeneric *>(I)->setOwner(this);
     Decs.clear();
   }
 
-  SPIRVDecorateSet &getDecorations() { return Decorations; }
+  SPIRVDecorateVec &getDecorations() { return Decorations; }
 
 protected:
-  SPIRVDecorateSet Decorations;
+  SPIRVDecorateVec Decorations;
```
Fznamznon, line 363: The patch changes only name here. I think the problem that clang-tidy complains about is out of scope of this patch.

PR #1241, opened 2021-10-08
Fix pointer to pointer mangling substitution.

aratajew:
This change fixes generation of SPV-IR for OpEnqueueMarker opcode and
VME related opcodes.


aratajew:
@svenvh could you please take a look? :)

svenvh:
> > LGTM, but I didn't look at the VME stuff.
> 
> Thanks! Would you like someone else to take a look at VME stuff? If so, who could do this?

Yeah, I suppose that should be someone from Intel.

PR #1243, opened 2021-10-11
Update dbg-prolog-end test after LLVM change

svenvh:
Update test for LLVM commit 9f93f2bfbd3f ("Do not emit prologue_end
for line 0 locs if there is a non-zero loc present", 2021-10-07).


AlexeySotkin:
Thanks!

Issue #978, opened 2021-04-04
Bugs when there are two consecutive get_local_id


drcut:
Hi, I have tried to support converting NVPTX into SPIRV and then convert into LLVM IR. And when I convert the following code
`int row = threadIdx.y;`
, it can generate the corret LLVM bitcode
`%0 = call spir_func i64 @_Z12get_local_idj(i32 1) #1`

However, when the input code is 
`int row = threadIdx.y;
int col = threadIdx.x;`
The generated LLVM will be 
`%0 = call spir_func i64 @_Z12get_local_idj(i32 1)`
`%2 = load <3 x i64>, <3 x i64> addrspace(5)* @__spirv_BuiltInLocalInvocationId1`
`%3 = extractelement <3 x i64> %2, i32 0 `
So I think when converting a SPIRV file into LLVM IR, when there are two consecutive get_local_id, it will raise error.
Would you please give me some suggestions?
Thanks in advance!

AlexeySachkov:
Hi @drcut,

> Would you please give me some suggestions?

Unfortunately, I can't suggest anything right now, but having that SPIR-V file which triggers this issue would definitely help

drcut:
Thanks, here are the links of my spv file
[naive_matmul.spv](https://drive.google.com/file/d/1vEqgJInlfNHOixPkiKGQfHKLOmxGeQ1T/view?usp=sharing).
I also put its corresponding cuda file, as it may be helpful.
[naive_matmul.cu](https://drive.google.com/file/d/1C7m1Djcme-h_tuWx3P6AiQlNTRLhbB1M/view?usp=sharing)

drcut:
@AlexeySachkov Do you have any suggestions for why there is an error when there is more than one get_local_idj? I'd like to study this issue recently, and want to get some hints from you. Thanks in advance

drcut:
So I used a quite ugly way to solve the bug. I create some variable (in the below code, the `LocalInvocationId`) to store the generated GlobalVariable, so that each time we meet a built-in function, we first check whether its corresponding GlobalVariable existing, if not, we will create a new one, else, we directly use the previous generated GlobalVariable
```cpp
    if (BuiltinVarName == "__spirv_BuiltInLocalInvocationId") {
      if (!LocalInvocationId)
        LocalInvocationId = new GlobalVariable(
            *M, GVType, true, GlobalValue::ExternalLinkage, nullptr,
[...]

PR #1245, opened 2021-10-11
Implement SPV_INTEL_hw_thread_queries extension

MrSidims:
The extension adds BuiltIn decorations to query maximum number
of H/W threads per a sub-device (tile), a sub-device ID and a H/W
thread ID within the sub-device. These instructions combined can be
used to calculate a unique H/W thread on which the code is being launched.

Spec: https://github.com/intel/llvm/pull/4747
Some info about Intel multi-tile cards:
https://github.com/intel/llvm/blob/sycl/sycl/doc/MultiTileCardWithLevelZero.md

### DIFF
--- a/lib/SPIRV/SPIRVWriter.cpp
+++ b/lib/SPIRV/SPIRVWriter.cpp
@@ -1529,11 +1529,10 @@ LLVMToSPIRVBase::transValueWithoutDecoration(Value *V, SPIRVBasicBlock *BB,
             internal::BuiltInMaxHWThreadIDPerSubDeviceINTEL) {
       if (!BM->isAllowedToUseExtension(
               ExtensionID::SPV_INTEL_hw_thread_queries)) {
-        std::stringstream SS;
-        SS << "Intel HW thread queries must be enabled by "
-           << "SPV_INTEL_hw_thread_queries extension.\n"
-           << "LLVM value that is being translated:\n";
-        getErrorLog().checkError(false, SPIRVEC_InvalidModule, V, SS.str());
+        std::string ErrorStr = "Intel HW thread queries must be enabled by "
+                               "SPV_INTEL_hw_thread_queries extension.\n"
+                               "LLVM value that is being translated:\n";
+        getErrorLog().checkError(false, SPIRVEC_InvalidModule, V, ErrorStr);
       }
       BM->addExtension(ExtensionID::SPV_INTEL_hw_thread_queries);
     }


lib/SPIRV/SPIRVWriter.cpp
```diff
@@ -1524,6 +1524,21 @@ LLVMToSPIRVBase::transValueWithoutDecoration(Value *V, SPIRVBasicBlock *BB,
     spv::BuiltIn Builtin = spv::BuiltInPosition;
     if (!GV->hasName() || !getSPIRVBuiltin(GV->getName().str(), Builtin))
       return BVar;
+    if (static_cast<uint32_t>(Builtin) >= internal::BuiltInSubDeviceIDINTEL &&
+        static_cast<uint32_t>(Builtin) <=
+            internal::BuiltInMaxHWThreadIDPerSubDeviceINTEL) {
+      if (!BM->isAllowedToUseExtension(
+              ExtensionID::SPV_INTEL_hw_thread_queries)) {
+        std::stringstream SS;
+        SS << "Intel HW thread queries must be enabled by "
```
MrSidims, line 1533: Alternatively, in case if the extension is not enabled, SubDeviceIDINTEL and MaxHWThreadIDPerSubDeviceINTEL can be replaced with OpConstant carrying 0, and HWThreadIDINTEL can be replaced with either 0 or GlobalLineralId, WDYT?

MrSidims:
@AlexeySotkin @Fznamznon please take a look

lib/SPIRV/SPIRVWriter.cpp
MrSidims: Or even HWThreadIDINTEL can also be replaced with 0.

MrSidims:
> Please, fix typo in commit header 'extenion'.

Thanks!

test/transcoding/SPV_INTEL_hw_thread_queries/intel_hw_thread_queries.ll
```diff
@@ -0,0 +1,39 @@
+; Test for SPV_INTEL_hw_thread_queries feature
+; SPIR-V friendly LLVM IR to SPIR-V translation and vice versa
+; RUN: llvm-as %s -o %t.bc
+; RUN: llvm-spirv %t.bc --spirv-ext=+SPV_INTEL_hw_thread_queries -o %t.spv
+; RUN: llvm-spirv %t.spv -to-text -o %t.spt
+; RUN: FileCheck < %t.spt %s --check-prefix=CHECK-SPIRV
+
+; RUN: llvm-spirv -r %t.spv -o %t.rev.bc
+; RUN: llvm-dis < %t.rev.bc | FileCheck %s --check-prefix=CHECK-LLVM
```
Fznamznon, line 9: I guess the new built-ins don't have any form that is not SPIR-V friendly IR, right? I wonder what is the purpose of checking it twice?

test/transcoding/SPV_INTEL_hw_thread_queries/intel_hw_thread_queries.ll
```diff
@@ -0,0 +1,39 @@
...
+@__spirv_BuiltInHWThreadIDINTEL = external addrspace(1) global i32
+@__spirv_BuiltInMaxHWThreadIDPerSubDeviceINTEL = external addrspace(1) global i32
+
+; Function Attrs: nounwind readnone
+define spir_kernel void @foo() {
+entry:
+  %0 = load i32, i32 addrspace(4)* addrspacecast (i32 addrspace(1)* @__spirv_BuiltInSubDeviceIDINTEL to i32 addrspace(4)*), align 4
+  %1 = load i32, i32 addrspace(4)* addrspacecast (i32 addrspace(1)* @__spirv_BuiltInHWThreadIDINTEL to i32 addrspace(4)*), align 4
+  %2 = load i32, i32 addrspace(4)* addrspacecast (i32 addrspace(1)* @__spirv_BuiltInMaxHWThreadIDPerSubDeviceINTEL to i32 addrspace(4)*), align 4
+  ; CHECK-LLVM: call spir_func i32 @_Z31__spirv_BuiltInSubDeviceIDINTELv() #1
```
Fznamznon, line 35: Can we also check that LLVM IR with calls but not builtin variables is correctly recognized by the translator?

lib/SPIRV/SPIRVWriter.cpp
Fznamznon: I think this is a bit of "magic". I would not feed SPIR-V with an extension to the device that doesn't support it at all. Or it should be documented somewhere with like... giant red letters.

test/transcoding/SPV_INTEL_hw_thread_queries/intel_hw_thread_queries.ll
MrSidims: Originally, there was no such check in the patch. But then I noticed, that during translation to OCL - the built-ins are not mangled, and I actually need to add extra code for reverse translation. So I added extra checks to reflect this.

lib/SPIRV/SPIRVWriter.cpp
MrSidims: Agree

test/transcoding/SPV_INTEL_hw_thread_queries/intel_hw_thread_queries.ll
MrSidims: Added a separate test for that. But it does look like if in the module there are both representations of builtins simultaneously - it confuses the translator and it generated not correct SPIR-V (not only for this extension). Please consider following example:
```
@__spirv_BuiltInSubgroupMaxSize = external addrspace(1) global i32
define spir_kernel void @foo() {
entry:
  %0 = call spir_func i32 @_Z30__spirv_BuiltInSubgroupMaxSizev() #1
  ret void
}
define spir_kernel void @boo() {
entry:
  %0 = load i32, i32 addrspace(4)* addrspacecast (i32 addrspace(1)* @__spirv_BuiltInSubgroupMaxSize to i32 addrspace(4)*)
  ret void
}
declare dso_local spir_func i32 @_Z30__spirv_BuiltInSubgroupMaxSizev()
```
 Translates to:
```
 12 Decorate 4 LinkageAttributes "__spirv_BuiltInSubgroupMaxSize" Import
 4 Decorate 4 BuiltIn 37
 13 Decorate 6 LinkageAttributes "__spirv_BuiltInSubgroupMaxSize.1" Import
 3 Decorate 6 Constant
```
and then to:
```
%0 = load i32, i32 addrspace(7)* @__spirv_BuiltInSubgroupMaxSize.1, align 4
// ...
%0 = call spir_func i32 @_Z22get_max_sub_group_sizev()
```
I believe, it might relates to https://github.com/KhronosGroup/SPIRV-LLVM-Translator/issues/978 .

lib/SPIRV/libSPIRV/SPIRVDecorate.h
```diff
@@ -102,6 +102,17 @@ class SPIRVDecorateGeneric : public SPIRVAnnotationGeneric {
     }
   }
 
+  llvm::Optional<ExtensionID> getRequiredExtension() const override {
+    switch (static_cast<size_t>(Dec)) {
+    case internal::BuiltInSubDeviceIDINTEL:
```
Fznamznon, line 107: I was surprised by this change. Don't we have any extensions that bring new decorations already?

lib/SPIRV/SPIRVWriter.cpp
```diff
@@ -1524,6 +1524,20 @@ LLVMToSPIRVBase::transValueWithoutDecoration(Value *V, SPIRVBasicBlock *BB,
     spv::BuiltIn Builtin = spv::BuiltInPosition;
     if (!GV->hasName() || !getSPIRVBuiltin(GV->getName().str(), Builtin))
       return BVar;
+    if (static_cast<uint32_t>(Builtin) >= internal::BuiltInSubDeviceIDINTEL &&
+        static_cast<uint32_t>(Builtin) <=
+            internal::BuiltInMaxHWThreadIDPerSubDeviceINTEL) {
+      if (!BM->isAllowedToUseExtension(
+              ExtensionID::SPV_INTEL_hw_thread_queries)) {
+        std::stringstream SS;
```
mlychkov, line 1532: Is stringstream necessary here? May we use just string?

PR #1246, opened 2021-10-12
Add load to obtain input for ap/hls instructions in case if its type > 64bits

MrSidims:
In the LLVM IR if input is an integer with a width more than 64 bits
the function will take this input through 'byval' pointer. If we
leave the translation as is it will violate the extension spec,
which says, that the input must by of an integer type. Add an extra load
to fix this.

### DIFF
--- a/lib/SPIRV/SPIRVWriter.cpp
+++ b/lib/SPIRV/SPIRVWriter.cpp
@@ -4152,7 +4152,7 @@ LLVMToSPIRVBase::transBuiltinToInstWithoutDecoration(Op OC, CallInst *CI,
       OpItr++;
     }
 
-    // A - integer input of any width or 'byval' pointer to this integer
+    // Input - integer input of any width or 'byval' pointer to this integer
     SPIRVValue *Input = transValue(*OpItr, BB);
     if (OpItr->getType()->isPointerTy())
       Input = BM->addLoadInst(Input, {}, BB);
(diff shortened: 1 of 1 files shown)


MrSidims:
@pauzinl @AlexeySotkin please take a look

lib/SPIRV/SPIRVWriter.cpp
```diff
@@ -4219,7 +4224,13 @@ LLVMToSPIRVBase::transBuiltinToInstWithoutDecoration(Op OC, CallInst *CI,
       OpItr++;
     }
 
-    SPIRVValue *InA = transValue(*OpItr++ /* A - input */, BB);
+    // A - integer input of any width or 'byval' pointer to this integer
```
AlexeySotkin, line 4225: A?

lib/SPIRV/SPIRVWriter.cpp
```diff
@@ -4287,12 +4298,24 @@ LLVMToSPIRVBase::transBuiltinToInstWithoutDecoration(Op OC, CallInst *CI,
...
```
AlexeySotkin, line 4297: A -> InA ?

lib/SPIRV/SPIRVWriter.cpp
```diff
@@ -4135,8 +4135,13 @@ LLVMToSPIRVBase::transBuiltinToInstWithoutDecoration(Op OC, CallInst *CI,
       OpItr++;
     }
 
-    SPIRVValue *Input =
-        transValue(*OpItr++ /* A - integer input of any width */, BB);
+    // A - integer input of any width or 'byval' pointer to this integer
```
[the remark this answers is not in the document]
MrSidims: Renamed here and after, thanks!

test/transcoding/capability-arbitrary-precision-fixed-point-numbers.ll
```diff
@@ -166,6 +168,11 @@
 ; CHECK-SPIRV: 6 Load [[Ty_34]] [[SinCos_InId:[0-9]+]]
 ; CHECK-SPIRV-NEXT: 9 FixedSinCosINTEL [[Ty_66]] [[#]] [[SinCos_InId]] 1 3 2 0 0
 
+; CHECK-SPIRV: 6 Load [[Ty_68]] [[ResId:[0-9]+]]
+; CHECK-SPIRV-NEXT: 5 Store [[PtrId:[0-9]+]] [[ResId]]
+; CHECK-SPIRV-NEXT: 4 Load [[Ty_68]] [[ExpInId2:[0-9]+]] [[PtrId]]
```
AlexeySotkin, line 173: It would be nice if we could eliminate redundant load-store sequence.

AlexeySotkin:
@MrSidims please resolve the conflict.

AlexeySotkin (approved):

PR #1265, opened 2021-11-03
Rename OpConstFunctionPointerINTEL to OpConstantFunctionPointerINTEL

MrSidims:
Spec update: https://github.com/intel/llvm/pull/4883/files
SPIR-V Headers update: https://github.com/KhronosGroup/SPIRV-Headers/pull/251
SPIR-V Tools update: https://github.com/KhronosGroup/SPIRV-Tools/pull/4613

Signed-off-by: Dmitry Sidorov <dmitry.sidorov@intel.com>

PR #1279, opened 2021-11-15
Support DW_OP_LLVM_arg DWARF operation

vmaksimo:


vmaksimo:
Tagging @AlexeySotkin @MrSidims

test/DebugInfo/DebugInfoUnsupported.ll
```diff
@@ -0,0 +1,42 @@
...
+; TODO: remove this test as soon as more DWARF operations are supported.
+
+; RUN: llvm-as %s -o %t.bc
+; RUN: llvm-spirv %t.bc -o %t.spv
+; RUN: llvm-spirv %t.spv -to-text -o %t.spt
+; RUN: FileCheck < %t.spt %s --check-prefix=CHECK-SPIRV
+; RUN: llvm-spirv -r %t.spv -o %t.rev.bc
+; RUN: llvm-dis < %t.rev.bc | FileCheck %s --check-prefix=CHECK-LLVM
+
+; CHECK-SPIRV: DebugInfoNone
```
AlexeySotkin, line 12: This check doesn't seem really useful, this instruction may be used by many other instructions. I think check for `DebugValue` and its operands makes more sense here.

test/DebugInfo/DebugInfoUnsupported.ll
```diff
@@ -0,0 +1,42 @@
...
+
+target datalayout = "e-m:o-i64:64-f80:128-n8:16:32:64-S128"
+target triple = "spir64-unknown-unknown"
+
+declare void @llvm.dbg.value(metadata, metadata, metadata) nounwind readnone speculatable
+
+define void @DbgIntrinsics() sanitize_memtag {
+entry:
+  %x = alloca i32, align 4
+  call void @llvm.dbg.value(metadata !DIArgList(i32* %x, i32* %x), metadata !6, metadata !DIExpression(DW_OP_LLVM_arg, 0, DW_OP_LLVM_arg, 1, DW_OP_plus)), !dbg !10
```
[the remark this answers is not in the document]
vmaksimo: Updated in 715e6bb

test/DebugInfo/DebugInfoUnsupported.ll
vmaksimo: Updated in 715e6bb

test/DebugInfo/DebugInfoUnsupported.ll
```diff
@@ -0,0 +1,49 @@
...
+target triple = "spir64-unknown-unknown"
+
+declare void @llvm.dbg.value(metadata, metadata, metadata) nounwind readnone speculatable
+
+define void @DbgIntrinsics() sanitize_memtag {
+entry:
+  %x = alloca i32, align 4
+; CHECK-LLVM-NOT: DW_OP_LLVM_arg
+; CHECK-LLVM: call void @llvm.dbg.value(metadata i32* %x, metadata ![[#]], metadata !DIExpression())
+; CHECK-LLVM-NOT: DW_OP_LLVM_arg
```
MrSidims, line 29: Doesn't the first CHECK-LLVM-NOT: DW_OP_LLVM_arg scan all of the module ? I mean the second check seems to be not needed.

lib/SPIRV/LLVMToSPIRVDbgTran.cpp
```diff
@@ -1010,9 +1010,16 @@ SPIRVEntry *LLVMToSPIRVDbgTran::transDbgExpression(const DIExpression *Expr) {
   for (unsigned I = 0, N = Expr->getNumElements(); I < N; ++I) {
     using namespace SPIRVDebug::Operand::Operation;
     auto DWARFOpCode = static_cast<dwarf::LocationAtom>(Expr->getElement(I));
+    SPIRVDebug::ExpressionOpCode OC = {};
+
+    // Skip handling of debug intrinsics with unsupported metadata.
+    // TODO: replace this check to ::map call when more DWARF operations are
+    // supported.
+    if (!SPIRV::DbgExpressionOpCodeMap::find(DWARFOpCode, &OC)) {
+      Operations.clear();
```
MrSidims, line 1019: I'm not an expert in debug info processing in the translator and its infrastructure. Yet I have (may be a naive) suggestion not to clear all of the Operations vector in case if a single operation is supported, but instead erasing/not adding this particular operation to it, leaving others to be processed. I can also imagine, that it's a bad suggestion, that can result in a misleading debug information in the module, WDYT?

test/DebugInfo/DebugInfoUnsupported.ll
vmaksimo: No, the behavior you described works only when `--implicit-check-not` option is enabled in FileCheck. Otherwise, it works like that:

> The “CHECK-NOT:” directive is used to verify that a string doesn’t occur between two matches (or before the first match, or after the last match)

Although, it's possible to use the option and to replace two checks with one.

MrSidims: No need to replace then, thanks

lib/SPIRV/LLVMToSPIRVDbgTran.cpp
vmaksimo: Firstly I also thought that it'd be enough to just skip the unsupported operation and proceed further. 
But there are such DWARF expressions which operands should be processed together, otherwise, they'll make no sense, e.g.:
`!DIExpression(DW_OP_LLVM_arg, 0, DW_OP_LLVM_arg, 1, DW_OP_minus, DW_OP_stack_value)` - if we just ignore `DW_OP_LLVM_arg`, then `0` and `1` values lost their meaning. 

[Language reference ](https://llvm.org/docs/LangRef.html#diexpression) says:
> `DW_OP_LLVM_arg, N` is used in debug intrinsics that refer to more than one value, such as one that calculates the sum of two registers. This is always used in combination with an ordered list of values, such that `DW_OP_LLVM_arg, N` refers to the N``th element in that list.

So, the chosen approach is safer for debug information and that information won't be misleading.

AlexeySotkin: I also have no understanding how debugger will interpret such modification of debug info, but my _guess_ is when we remove the expression completely the debugger will something like "I don't know what the value is", and if we remove just one operation from the expression it will give us some value which is not the real/actual/correct. I personally prefer the first option, because the second one is very misleading.

MrSidims:
D:\a\SPIRV-LLVM-Translator\SPIRV-LLVM-Translator\llvm-project\llvm\projects\SPIRV-LLVM-Translator\lib\SPIRV\libSPIRV\SPIRVOpCodeEnum.h(347,1): error C2065: 'OpConstFunctionPointerINTEL': undeclared identifier [D:\a\SPIRV-LLVM-Translator\SPIRV-LLVM-Translator\build\projects\SPIRV-LLVM-Translator\lib\SPIRV\LLVMSPIRVLib.vcxproj]

This error is caused by https://github.com/KhronosGroup/SPIRV-Headers/pull/251 and should be fixed by https://github.com/KhronosGroup/SPIRV-LLVM-Translator/pull/1265 , but honestly I don't understand why the only build is failing, I'd expect of all the builds to fail.

MrSidims (approved):

AlexeySotkin:
> D:\a\SPIRV-LLVM-Translator\SPIRV-LLVM-Translator\llvm-project\llvm\projects\SPIRV-LLVM-Translator\lib\SPIRV\libSPIRV\SPIRVOpCodeEnum.h(347,1): error C2065: 'OpConstFunctionPointerINTEL': undeclared identifier [D:\a\SPIRV-LLVM-Translator\SPIRV-LLVM-Translator\build\projects\SPIRV-LLVM-Translator\lib\SPIRV\LLVMSPIRVLib.vcxproj]
> 
> This error is caused by [KhronosGroup/SPIRV-Headers#251](https://github.com/KhronosGroup/SPIRV-Headers/pull/251) and should be fixed by #1265 , but honestly I don't understand why the only build is failing, I'd expect of all the builds to fail.

Seems like it fixes this kind of build but breaks all other.

test/DebugInfo/DebugInfoUnsupported.ll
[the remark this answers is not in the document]
AlexeySotkin: What happened with DIArgList? It seems right, but I don't recall if we added handling for this type of metadata.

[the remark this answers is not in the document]
vmaksimo: You're right, there is no special handling for this metadata.
The translator just takes the variable (from the second intrinsic argument) instead of handling the first operand of intrinsic which actually contains the `DIArgList`:
https://github.com/KhronosGroup/SPIRV-LLVM-Translator/blob/1cad18c0c448650d1cc3f18f6077360a98a752f9/lib/SPIRV/LLVMToSPIRVDbgTran.cpp#L120-L121
So, this behavior should be changed once we start to support DIArgList, but it's not related to the changes in this PR.

aparshin-intel:
@vmaksimo , @AlexeySotkin , @MrSidims 

Could you elaborate on why this one was approved? I don't think that silently skipping debug information is a good idea. Should not users be aware that the translated LLVM IR can't be represented as a SPIRV?

AlexeySotkin:
> @vmaksimo , @AlexeySotkin , @MrSidims
> 
> Could you elaborate on why this one was approved? I don't think that silently skipping debug information is a good idea. Should not users be aware that the translated LLVM IR can't be represented as a SPIRV?
I agree, users should be aware, but I think we don't have a mechanism to report warnings in the translator, do we?

AlexeySotkin:
> > I agree, users should be aware, but I think we don't have a mechanism to report warnings in the translator, do we?
> 
> I'm not aware of this mechanism. Moreover, I don't think this should be a warning.
> 
> It is responsibility of a client to provide an LLVM IR that can be represented in SPIR-V. If a client has some debug instruction that uses an DIArgList - it is there for a reason. DIArgList is not supported by SPIRV specification, thus it is the responsibility of a client to get rid of those.

The problem is not in `DIArgList` but in the `DW_OP_LLVM_arg`. But it doesn't change the matter. This patch is rather a workaround to avoid a crash in translation of debug info with unsupported debug expression operation. The proper fix would require update of the SPIR-V debug info spec. Meanwhile, perhaps a better solution might be to add `DW_OP_LLVM_arg` to https://github.com/KhronosGroup/SPIRV-LLVM-Translator/blob/1cad18c0c448650d1cc3f18f6077360a98a752f9/lib/SPIRV/libSPIRV/SPIRV.debug.h#L866
(and a couple of other places in the same file). @vmaksimo What do you think?

vmaksimo:
> The problem is not in `DIArgList` but in the `DW_OP_LLVM_arg`. But it doesn't change the matter. This patch is rather a workaround to avoid a crash in translation of debug info with unsupported debug expression operation. The proper fix would require update of the SPIR-V debug info spec. Meanwhile, perhaps a better solution might be to add `DW_OP_LLVM_arg` to
> 
> https://github.com/KhronosGroup/SPIRV-LLVM-Translator/blob/1cad18c0c448650d1cc3f18f6077360a98a752f9/lib/SPIRV/libSPIRV/SPIRV.debug.h#L866
> 
> 
> (and a couple of other places in the same file). @vmaksimo What do you think?

Unfortunately, it won't work that simple.
This is the case when the DWARF expression with all its operands should be processed together.
`!DIExpression(DW_OP_LLVM_arg, 0, DW_OP_LLVM_arg, 1, DW_OP_minus, DW_OP_stack_value)` - if we just processed `DW_OP_LLVM_arg`, then the program will crash on `0` and `1` values, because they are not 'expressions' without DW_OP_LLVM_arg.
According to [language reference ](https://llvm.org/docs/LangRef.html#diexpression):
> `DW_OP_LLVM_arg, N` is used in debug intrinsics that refer to more than one value, such as one that calculates the sum of two registers. This is always used in combination with an ordered list of values, such that `DW_OP_LLVM_arg, N` refers to the N``th element in that list.

It seems that we should either support the DWARF expression fully (it'll involve support of `DIArgList` too) or just ignore the unknown expression if we don't want the translator to crash.
I agree that the better way is to update the SPIR-V debug info spec, but until then the translator will fail with such an input.

aparshin-intel:
>  just ignore the unknown expression if we don't want the translator to crash.

A proper behavior would be to return an error in this case.
This "workaround" silently discards some IR constructs. I don't believe this is what the user expects.

AlexeySotkin (approved):

MrSidims (approved):

Issue #1282, opened 2021-11-20
13.0.0 build failure


0-wiz-0:
I tried building 13.0.0 against llvm 13.0.0 and SPIRV-Headers sdk-1.2.198.0 on NetBSD-9.99.92/amd64 and see:
```
In file included from /scratch/parallel/spirv-llvm-translator/work/SPIRV-LLVM-Translator-13.0.0/lib/SPIRV/libSPIRV/SPIRVEnum.h:44,                    
                 from /scratch/parallel/spirv-llvm-translator/work/SPIRV-LLVM-Translator-13.0.0/lib/SPIRV/libSPIRV/SPIRVEntry.h:44,                   
                 from /scratch/parallel/spirv-llvm-translator/work/SPIRV-LLVM-Translator-13.0.0/lib/SPIRV/libSPIRV/SPIRVModule.h:44,                  
                 from /scratch/parallel/spirv-llvm-translator/work/SPIRV-LLVM-Translator-13.0.0/lib/SPIRV/LLVMToSPIRVDbgTran.h:42,                    
                 from /scratch/parallel/spirv-llvm-translator/work/SPIRV-LLVM-Translator-13.0.0/lib/SPIRV/LLVMToSPIRVDbgTran.cpp:38:                  
/scratch/parallel/spirv-llvm-translator/work/SPIRV-LLVM-Translator-13.0.0/lib/SPIRV/libSPIRV/SPIRVOpCodeEnum.h: In member function 'void SPIRV::SPIRVM
ap<Ty1, Ty2, Identifier>::init() [with Ty1 = spv::Op; Ty2 = std::__cxx11::basic_string<char>; Identifier = void]':                                    
/scratch/parallel/spirv-llvm-translator/work/SPIRV-LLVM-Translator-13.0.0/lib/SPIRV/libSPIRV/SPIRVOpCode.h:52:31: error: 'OpConstFunctionPointerINTEL'
 was not declared in this scope                                                                                                                       
   52 | #define _SPIRV_OP(x, ...) add(Op##x, #x);                                                                                                     
      |                               ^~                                                                                                              
/scratch/parallel/spirv-llvm-translator/work/SPIRV-LLVM-Translator-13.0.0/lib/SPIRV/libSPIRV/SPIRVOpCodeEnum.h:347:1: note: in expansion of macro '_SP
IRV_OP'                                                                                                                                               
[...]

PR #1284, opened 2021-11-22
Rename OpConstFunctionPointerINTEL to OpConstantFunctionPointerINTEL

MrSidims:


MrSidims:
Addresses https://github.com/KhronosGroup/SPIRV-LLVM-Translator/issues/1282 . I probably won't cherry-pick the patch to the older branches because they still use local copy of spirv.hpp

PR #1291, opened 2021-11-26
Fix memory leak after OpInBoundsPtrAccessChain

pauzinl:
There was memory leak after creating `OpInBoundsPtrAccessChain` instruction during reverse translation. This patch fixes it.
Note: `OpInBoundsPtrAccessChain` creates from `OpSpecConstantOp` with same ID, so we can't add it to `IdEntryMap`

### DIFF
--- a/lib/SPIRV/libSPIRV/SPIRVInstruction.cpp
+++ b/lib/SPIRV/libSPIRV/SPIRVInstruction.cpp
@@ -244,8 +244,12 @@ SPIRVInstruction *createInstFromSpecConstantOp(SPIRVSpecConstantOp *Inst) {
   assert(isSpecConstantOpAllowedOp(OC) &&
          "Op code not allowed for OpSpecConstantOp");
   Ops.erase(Ops.begin(), Ops.begin() + 1);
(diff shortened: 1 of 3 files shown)


pauzinl:
Hi, @AlexeySotkin, @svenvh, @KornevNikita could you please review this fix?

svenvh (approved):

KornevNikita (approved):

AlexeySotkin:
What if we just `delete BI` after mapping at
https://github.com/KhronosGroup/SPIRV-LLVM-Translator/blob/ef0015c775791d57c10d573d1de64b423827a225/lib/SPIRV/SPIRVReader.cpp#L1474
?

pauzinl:
> What if we just `delete BI` after mapping at
> 
> https://github.com/KhronosGroup/SPIRV-LLVM-Translator/blob/ef0015c775791d57c10d573d1de64b423827a225/lib/SPIRV/SPIRVReader.cpp#L1471-L1475
> 
> 
> ?

@AlexeySotkin, If I understood correctly, you suggest something like:
`   Value *RetVal =  mapValue(BV, transValue(BI, nullptr, nullptr, false));
    delete BI;
    return RetVal;`
It caused test failure with incorrect generated IR

PR #1294, opened 2021-11-26
Add --spirv-tools-dis option to disassemble using SPIRV-Tools

svenvh:
llvm-spirv's text rendering of a SPIR-V module uses an internal
non-standard format that is not compatible with any other tools.

Add support for emitting SPIR-V assembly using the more widely used
SPIR-V Tools format.  This is achieved by linking to SPIRV-Tools and
invoking its disassembler.  For now, this functionality is only
available if the SPIRV-Tools package is found at build configuration
time.

The main motivation for this change is the proposed integration of
llvm-spirv in Clang (https://reviews.llvm.org/D112404 and
https://reviews.llvm.org/D112410).  Particularly, we would like Clang invocations
such as `clang -S -target spirv` to produce the standard SPIR-V
assembly format instead of llvm-spirv's internal format.

### DIFF
--- a/tools/llvm-spirv/CMakeLists.txt
+++ b/tools/llvm-spirv/CMakeLists.txt
@@ -23,3 +23,9 @@ target_include_directories(llvm-spirv
     ${LLVM_INCLUDE_DIRS}
     ${LLVM_SPIRV_INCLUDE_DIRS}
 )
+
+if(SPIRV_TOOLS_FOUND)
+  target_compile_definitions(llvm-spirv PRIVATE LLVM_SPIRV_HAVE_SPIRV_TOOLS=1)
+  target_include_directories(llvm-spirv PRIVATE ${SPIRV_TOOLS_INCLUDE_DIRS})
+  target_link_libraries(llvm-spirv PRIVATE ${SPIRV_TOOLS_LDFLAGS})
+endif(SPIRV_TOOLS_FOUND)
--- a/tools/llvm-spirv/llvm-spirv.cpp
+++ b/tools/llvm-spirv/llvm-spirv.cpp
@@ -59,6 +59,10 @@
 #include "llvm/Support/Signals.h"
 #include "llvm/Support/ToolOutputFile.h"
 
+#ifdef LLVM_SPIRV_HAVE_SPIRV_TOOLS
+#include "spirv-tools/libspirv.hpp"
+#endif
+
 #ifndef _SPIRV_SUPPORT_TEXT_FMT
 #define _SPIRV_SUPPORT_TEXT_FMT
 #endif
@@ -144,6 +148,10 @@ static cl::opt<bool>
         cl::desc("Preserve OpenCL kernel_arg_type and kernel_arg_type_qual "
                  "metadata through OpString"));
 
+static cl::opt<bool>
+    SPIRVToolsDis("spirv-tools-dis", cl::init(false),
+                  cl::desc("Emit textual assembly using SPIRV-Tools"));
+
 using SPIRV::ExtensionID;
 
 #ifdef _SPIRV_SUPPORT_TEXT_FMT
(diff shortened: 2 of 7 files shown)


tools/llvm-spirv/llvm-spirv.cpp
```diff
@@ -246,6 +278,48 @@ static int convertLLVMToSPIRV(const SPIRV::TranslatorOpts &Opts) {
...
+    std::string DisOutput;
+
+    // Serialize the SPIR-V module to a uint32_t array and then invoke the
+    // SPIR-V tools disassembler to obtain textual assembly.
+    std::string Err;
+    StreambufToArray OutStreamBuf;
+    std::ostream OutStream(&OutStreamBuf);
+    bool Success = writeSpirv(M.get(), Opts, OutStream, Err);
+    if (!Success) {
+      errs() << "Failed to serialize SPIR-V: " << Err << '\n';
```
AlexeySotkin, line 299: Basically `writeSpirv` does 2 things: translates LLVM to SPIR-V and serializes it. AFAIK, we see errors from translation stage way more often then from serialization.

tools/llvm-spirv/CMakeLists.txt
```diff
@@ -1,3 +1,6 @@
+option(LLVM_SPIRV_LIBSPIRV_DIS
+  "Use libspirv for disassembly." OFF)
```
AlexeySotkin, line 2: Perhaps it's worth clarifying that `libspirv` is from SPIRV-Tools.

AlexeySotkin (approved):
Other than a couple of nits, LGTM. Thanks!

tools/llvm-spirv/llvm-spirv.cpp
```diff
@@ -227,6 +235,30 @@ static std::string removeExt(const std::string &FileName) {
...
+protected:
+  std::streamsize xsputn(const char *s, std::streamsize count) override {
+    for (std::streamsize I = 0; I < count; I++) {
+      v.push_back(s[I]);
+    }
+    return count;
+  }
+
+private:
+  std::vector<char> v;
```
AnastasiaStulova, line 258: maybe it can be named `buffer` or something else more descriptive instead?

AnastasiaStulova: Also should variable names be in the upper case or does it not apply to members?

tools/llvm-spirv/llvm-spirv.cpp
svenvh: Good point, I have rephrased it.

tools/llvm-spirv/CMakeLists.txt
svenvh: Yeah, done.

AnastasiaStulova (commented):
LGTM in general!

Should we add any test for this?

tools/llvm-spirv/llvm-spirv.cpp
svenvh: > Should we add any test for this?

Ideally yes, but it's not trivial to do so, because the test should only run when llvm-spirv was built with the new configuration option enabled.

A related question is whether we should add another build to the CI for the new build configuration?

> maybe it can be named buffer or something else more descriptive instead?

Sure, I'll change the name.

AnastasiaStulova: > 



> Ideally yes, but it's not trivial to do so, because the test should only run when llvm-spirv was built with the new configuration option enabled.

> A related question is whether we should add another build to the CI for the new build configuration?

Is there some `REQUIRES` guards in tests already available? Yeah I imagine if we want to run this in CI we might need a new CI job or something? Although it would only run one test I imagine which is a bit wasteful... but if we want clang to use this it should be tested somewhere.... So maybe we can just extend an existing CI to run this too?

svenvh: > Is there some REQUIRES guards in tests already available?

There's not, but I have added one.

tools/llvm-spirv/CMakeLists.txt
```diff
@@ -23,3 +26,10 @@ target_include_directories(llvm-spirv
     ${LLVM_INCLUDE_DIRS}
     ${LLVM_SPIRV_INCLUDE_DIRS}
 )
+
+if(LLVM_SPIRV_LIBSPIRV_DIS)
+  pkg_search_module(SPIRV_TOOLS REQUIRED SPIRV-Tools)
```
MrSidims, line 31: May be something like this can work (not sure about the correctness of pkg_search_module semantic)?

```suggestion
  pkg_search_module(SPIRV_TOOLS QUIET SPIRV-Tools SPIRV_TOOLS_FOUND)
  if(SPIRV_TOOLS_FOUND)
    add_definitions(LLVM_SPIRV_LIBSPIRV_DIS)
```

We would end up with two possible builds, but with REQUIRES specifier in the tests we can test this functionality without spawning extra CI jobs.

svenvh: In the Khronos OpenCL Tooling call yesterday we discussed enabling the new cmake option in CI.  But following your suggestion, we can just enable the functionality if SPIRV-Tools is available without the need for a cmake option.  This has the additional benefit that we can hoist the SPIRV-Tools check to the top-level `CMakeLists.txt`, and reuse its result in `test/CMakeLists.txt`.  I'll update the PR for this.

AnastasiaStulova (approved):
LGTM! Thanks

MrSidims (approved):

PR #1300, opened 2021-11-29
Convert ldexp scalar arg to vector

haonanya:
For OpenCL built-in math functions 'floatn ldexp(floatn x, int k)' and
'doublen ldexp(doublen x, int k)', convert scalar arg to vector to keep
consistency with SPIRV spec. Regarding to SPIRV OpenCL Extended
Instruction set, k operand must have the same component count as Result
Type and x operands

Signed-off-by: Haonan Yang <haonan.yang@intel.com>


haonanya:
Hi, @AlexeySachkov, @AlexeySotkin, could you please review the patch?
Thanks very much.

svenvh (commented):
LGTM, but please shorten the commit title (so that it doesn't get broken up over multiple lines).  I'd suggest something like "Convert ldexp scalar arg to vector"; further clarification can then be provided in the rest of the commit message.

Since this is OpenCL specific, I would have written a .cl test instead of .ll, because .cl tests are easier to maintain.

svenvh (approved):
LGTM, thanks!

haonanya:
Hi, @AlexeySotkin, do you have further comments or would you like someone else to take a look? 
Thanks very much.

lib/SPIRV/OCLToSPIRV.cpp
```diff
@@ -1883,6 +1890,31 @@ void OCLToSPIRVBase::visitSubgroupAVCBuiltinCallWithSampler(
...
+    Type *Type0 = Args[0]->getType();
+    Type *Type1 = Args[1]->getType();
+    // For OpenCL built-in math functions 'floatn ldexp(floatn x, int k)' and
+    // 'doublen ldexp(doublen x, int k)', convert scalar arg to vector to keep
+    // consistency with SPIRV spec. Regarding to SPIRV OpenCL Extended
+    // Instruction set, k operand must have the same component count as Result
+    // Type and x operands
+    if (auto *FixedVecType0 = dyn_cast<FixedVectorType>(Type0)) {
+      auto ScalarTypeID = Type0->getScalarType()->getTypeID();
+      if ((ScalarTypeID == llvm::Type::FloatTyID ||
```
MrSidims, line 1906: Should we also handle Half type here? It exists as an extension and can be found here: https://www.khronos.org/registry/OpenCL/specs/2.2/html/OpenCL_Ext.html

test/transcoding/ldexp.cl
```diff
@@ -0,0 +1,15 @@
+// Check that translator converts scalar arg to vector for ldexp math instructions
+// RUN: %clang_cc1 -triple spir-unknown-unknown -O1 -cl-std=CL2.0 -fdeclare-opencl-builtins -finclude-default-header -emit-llvm-bc %s -o %t.bc
+// RUN: llvm-spirv %t.bc -o %t.spv
+// RUN: spirv-val %t.spv
+// RUN: llvm-spirv -to-text %t.spv -o %t.spt
+// RUN: FileCheck < %t.spt %s --check-prefix=CHECK-SPIRV
+// RUN: llvm-spirv -r %t.spv -o %t.rev.bc
+// RUN: llvm-dis %t.rev.bc -o - | FileCheck %s --check-prefixes=CHECK-LLVM-CL20
+
+__kernel void test_kernel(float3 x, int k, __global float3* ret) {
```
MrSidims, line 10: nit: we should probably test doubles as well

lib/SPIRV/OCLToSPIRV.cpp
haonanya: done, thanks.

test/transcoding/ldexp.cl
haonanya: done, add test for doublen and halfn, thanks.

MrSidims (approved):

haonanya:
@MrSidims. Thanks for your review. Could you please help to merge it? :)

PR #1303, opened 2021-12-01
Fix return type postfix for SPIR-V Friendly IR

aratajew:


aratajew:
@Fznamznon could you please take a look? :)

lib/SPIRV/SPIRVReader.cpp
```diff
@@ -3093,24 +3093,43 @@ Instruction *SPIRVToLLVM::transSPIRVBuiltinFromInst(SPIRVInstruction *BI,
...
+  case OpSDotAccSatKHR:
+  case OpUDotAccSatKHR:
+  case OpSUDotAccSatKHR:
     AddRetTypePostfix = true;
-    if (OC == OpConvertUToF || OC == OpSatConvertUToS)
-      IsRetSigned = true;
+    break;
+  default: {
+    if ((isCvtOpCode(OC) && OC != OpGenericCastToPtrExplicit) ||
+        OC == internal::OpJointMatrixLoadINTEL)
```
MrSidims, line 3115: nit: `internal::OpJointMatrixLoadINTEL` can be added among other non-default switch cases, if we switch over `static_cast<size_t>(OC)`

MrSidims (approved):
LGTM, thanks!

lib/SPIRV/SPIRVReader.cpp
aratajew: Thanks the the tip. Implemented it, since in case we needed to add more internal opcodes to this switch in the future, the code will be much cleaner.

PR #696, opened 2020-08-21
Handle @llvm.memset.* with non-constant arguments

AlexeySotkin:
There is no SPIR-V counterpart for @llvm.memset.* intrinsic. Cases with constant
value and length arguments are emulated via "storing" a constant array to the
destination.

For other cases we wrap the intrinsic in @spirv.llvm_memset_* function and
expand the intrinsic to a loop via expandMemSetAsLoop() from
llvm/Transforms/Utils/LowerMemIntrinsics.h.

During reverse translation from SPIR-V to LLVM IR we can detect
@spirv.llvm_memset_* and replace it with @llvm.memset.

Signed-off-by: Alexey Sotkin <alexey.sotkin@intel.com>

PR #1304, opened 2021-12-01
Add lowering for llvm.bswap intrinsic

KornevNikita:


KornevNikita:
@AlexeySotkin @MrSidims @AlexeySachkov @mlychkov take a look please. I added CodeGen to CmakeLists to avoid copying of [LowerBSWAP](https://github.com/intel/llvm/blob/sycl/llvm/lib/CodeGen/IntrinsicLowering.cpp#L53) function. Can we leave it as it is, or would it be better to copy this function into the translator, WDYT?

lib/SPIRV/CMakeLists.txt
```diff
@@ -39,6 +39,7 @@ add_llvm_library(LLVMSPIRVLib
   LINK_COMPONENTS
     Analysis
     BitWriter
+    CodeGen
```
AlexeySotkin, line 42: The tradeoff here is having an extra and quite heavy LLVM library linked to llvm-spirv vs having(and maintaining) extra source code which is a copy-paste of a function from LLVM CodeGen lib.
The source we would copy is [this function](https://github.com/KornevNikita/llvm/blob/a8bb540ac7d683cdf087a0676ffdef9d3a4f7d2d/llvm/lib/CodeGen/IntrinsicLowering.cpp#L53-L145), ~100 lines.
@AlexeySachkov, @MrSidims what do you think?

svenvh: libLLVMCodeGen is indeed the biggest component of LLVM.  Perhaps we can propose to move the intrinsic lowering elsewhere (e.g. into TransformUtils)?

For us, linking against libLLVMCodeGen shouldn't be an issue, so that would still have our preference over duplicating code.

lib/SPIRV/SPIRVRegularizeLLVM.cpp
```diff
@@ -293,6 +296,11 @@ void SPIRVRegularizeLLVMBase::lowerUMulWithOverflow(
   UMulIntrinsic->setCalledFunction(UMulFunc);
 }
 
+void SPIRVRegularizeLLVMBase::lowerBSwap(CallInst *BSwapIntrinsic) {
+  IntrinsicLowering IL(M->getDataLayout());
+  IL.LowerIntrinsicCall(BSwapIntrinsic);
```
mlychkov, line 301:
```suggestion
void SPIRVRegularizeLLVMBase::lowerBSwap(const std::vector<CallInst *> &BSwapInsts) {
  IntrinsicLowering IL(M->getDataLayout());
  for (auto *CI : BSwapInsts)
    IL.LowerIntrinsicCall(CI);
```

This variant less portable maybe, but it reduces number of recreations of IntrinsicLowering object.

mlychkov: Or another variant:
```suggestion
void SPIRVRegularizeLLVMBase::lowerIntrinsic(CallInst *Intrinsic) {
  IntrinsicLowering IL(M->getDataLayout());
  IL.LowerIntrinsicCall(Intrinsic);
```
In this case we may use this method for lowering other intrinsics as well.

AlexeySotkin: I'd be very cautious calling this function for all intrinsics, because we don't control how intrinsics are lowered by this function and thus they may be lowered not in the way we want.

mlychkov: I didn't mean to call it blind for all intrinsics in input module blind. Just to reuse it in the future for some another intrinsic that should be lowered and has lowering code in LowerIntrinsicCall function.

AlexeySotkin: Oh, I misinterpret you comment. Yes we can do that.

AlexeySotkin: > 

I'd prefer moving `IntrinsicLowering` constructor out of the loop too.

AlexeySotkin:
One more thing we should do (as a follow up PR maybe?) is to recognize the bswap code during translation from SPIR-V to LLVM and replace it by the intrinsic. We already did such thing for `memset` in https://github.com/KhronosGroup/SPIRV-LLVM-Translator/pull/696 We probably should make `SPIRVRegularizeLLVM::lowerMemset()` more generic to handle other intrinsics too.

lib/SPIRV/SPIRVRegularizeLLVM.cpp
KornevNikita: I applied both suggestions in 3eb9828a1e1fa38b1d6f849381adfc92d35bd2cf

MrSidims (approved):
LGTM

lib/SPIRV/SPIRVReader.cpp
```diff
@@ -2500,8 +2500,46 @@ Value *SPIRVToLLVM::transValueWithoutDecoration(SPIRVValue *BV, Function *F,
...
+            (VarName.find("bswap.8") != std::string::npos &&
+             static_cast<SPIRVTypeInt *>(BV->getType())->getBitWidth() == 64))
+          BSwapOperandMap.emplace(
+              VarName,
+              transValue(static_cast<SPIRVBinary *>(BV)->getOperand(0), F, BB));
+        if (VarName.find("bswap.i16") != std::string::npos) {
+          IRBuilder<> Builder(BB);
+          return mapValue(BV, Builder.CreateIntrinsic(
+                                  Intrinsic::bswap, transType(BV->getType()),
+                                  BSwapOperandMap["bswap.2"]));
```
mlychkov, line 2525: This is just a nit: we may use integral value as a key of the map, for example bitwidth of swapped type.

mlychkov (approved):

lib/SPIRV/SPIRVReader.cpp
KornevNikita: We can't use bitwidth as the key for the map, because there are may be several bswap intrinsics with the same bitwidth. Вut, in this case, there is also an error in the current implementation. Thanks for pointing that out.

KornevNikita: Thanks, done

test/llvm-intrinsics/bswap.ll
```diff
@@ -0,0 +1,128 @@
...
+; CHECK-SPIRV: BitwiseOr [[#TypeInt64]]
+; CHECK-SPIRV: BitwiseOr [[#TypeInt64]]
+; CHECK-SPIRV: BitwiseOr [[#TypeInt64]] [[#RetVal:]]
+; CHECK-SPIRV: ReturnValue [[#RetVal]]
+; CHECK-SPIRV: FunctionEnd 
+
+; CHECK-LLVM: %1 = call i16 @llvm.bswap.i16(i16 %0)
+; CHECK-LLVM: %3 = call i16 @llvm.bswap.i16(i16 %2)
+; CHECK-LLVM: %5 = call i32 @llvm.bswap.i32(i32 %4)
+; CHECK-LLVM: %7 = call i64 @llvm.bswap.i64(i64 %6)
```
AlexeySotkin, line 68: Please make sure(`; CHECK-LLVM-NOT: `) there are no logical and bitwise instructions in the resulting LLVM.

test/llvm-intrinsics/bswap.ll
```diff
@@ -0,0 +1,128 @@
...
+; CHECK-SPIRV: BitwiseOr [[#TypeInt64]]
+; CHECK-SPIRV: BitwiseOr [[#TypeInt64]]
+; CHECK-SPIRV: BitwiseOr [[#TypeInt64]]
+; CHECK-SPIRV: BitwiseOr [[#TypeInt64]]
+; CHECK-SPIRV: BitwiseOr [[#TypeInt64]]
+; CHECK-SPIRV: BitwiseOr [[#TypeInt64]] [[#RetVal:]]
+; CHECK-SPIRV: ReturnValue [[#RetVal]]
+; CHECK-SPIRV: FunctionEnd 
+
+; CHECK-LLVM: %1 = call i16 @llvm.bswap.i16(i16 %0)
```
AlexeySotkin, line 65: `%number` is not a reliable check. numbers might change in the future. I suggest using regex
```suggestion
; CHECK-LLVM: {{%.*}} = call i16 @llvm.bswap.i16(i16 {{%.*}})
```
or don't check them at all
```suggestion
; CHECK-LLVM: call i16 @llvm.bswap.i16(i16
```

test/llvm-intrinsics/bswap.ll
```diff
@@ -0,0 +1,119 @@
...
+; CHECK-LLVM-NOT: lshr i16
+; CHECK-LLVM-NOT: or i16
+; CHECK-LLVM-NOT: shl i32
+; CHECK-LLVM-NOT: lshr i32
+; CHECK-LLVM-NOT: and i32
+; CHECK-LLVM-NOT: or i32
+; CHECK-LLVM-NOT: shl i64
+; CHECK-LLVM-NOT: lshr i64
+; CHECK-LLVM-NOT: and i64
+; CHECK-LLVM-NOT: or i64
```
mlychkov, line 59: May be reduced to:
```suggestion
; CHECK-LLVM-NOT: {{(shl|lshr|or|and) i(16|32|64)}}
```

mlychkov (approved):

AlexeySotkin (approved):

PR #1319, opened 2021-12-06
[docs] Add documentation for decorations metadata

steffenlarsen:
This commit adds documentation for two new named metadata nodes: `spirv.Decorations` and `spirv.ParameterDecorations`. These metadata nodes allow LLVM IR to directly specify SPIR-V decorations on global variable and function parameters respectively. The translator should traverse these metadata nodes and generate the corresponding SPIR-V decoration operations.

Note: The intention of this documentation change is to gather community feedback ahead of implementation.

### DIFF
--- a/docs/SPIRVRepresentationInLLVM.rst
+++ b/docs/SPIRVRepresentationInLLVM.rst
@@ -411,6 +411,53 @@ For example:
 are translated for image types, but they should be encoded in LLVM IR type name
 rather than function metadata.
 
+Function parameter and global variable decoration through metadata
+------------------------------------------------------------------
+
+Both function parameters and global variables can be decorated using LLVM
+metadata through the metadata names ``spirv.ParameterDecorations`` and
+``spirv.Decorations`` respectively. ``spirv.ParameterDecorations`` must be tied
+to the kernel function while ``spirv.Decorations`` is tied directly to the
+global variable.
+
+A "decoration-node" is a metadata node consisting of one or more operands. The
+first operand is an integer literal representing the SPIR-V decoration
+identifier. The other operands are either an integer or string literal
+representing the remaining extra operands of the corresponding SPIR-V
+decoration.
+
+A "decoration-list" is a metadata node consisting of references to zero or more
+decoration-nodes.
+
+``spirv.Decorations`` must refer to a decoration-list while
+``spirv.ParameterDecorations`` must refer to a metadata node that contains N
+references to decoration-lists, where N is the number of arguments of the
+function the metadata is tied to.
+
+``spirv.Decorations`` example:
+
+.. code-block:: llvm
+
+  !spirv.Decorations !1
+  ...
+  !1 = !{!2}
+  !2 = !{i32 22}
+
+decorates a global variable with ``Constant`` in SPIR-V.
+
+``spirv.ParameterDecorations`` example:
+
+.. code-block:: llvm
+
+  !spirv.ParameterDecorations !1
+  ...
+  !1 = !{!2, !3}
+  !2 = !{}
+  !2 = !{i32 19}
+
+decorates the second argument of the corresponding function with ``Restrict``
+in SPIR-V while not adding any decoration to the first argument.
+
 Debug information extension
 ===========================
 
--- a/docs/SPIRVRepresentationInLLVM.rst
+++ b/docs/SPIRVRepresentationInLLVM.rst
@@ -438,7 +438,7 @@ function the metadata is tied to.
 
 .. code-block:: llvm
 
-  @v = global i32 0,!spirv.Decorations !1
+  @v = global i32 0, !spirv.Decorations !1
   ...
   !1 = !{!2, !3}               ; decoration-list with two decoration nodes
   !2 = !{i32 22}               ; decoration-node with no extra operands
@@ -455,8 +455,8 @@ with extra operands ``"v"`` and ``Export`` in SPIR-V.
   ...
   !1 = !{!2, !3} ; metadata node with 2 decoration-lists
   !2 = !{}       ; empty decoration-list
-  !3 = !{!4}     ; decoration-list
-  !4 = !{i32 19} ; decoration-node
+  !3 = !{!4}     ; decoration-list with one decoration node
+  !4 = !{i32 19} ; decoration-node with no extra operands
 
 decorates the argument ``a`` of ``k`` with ``Restrict`` in SPIR-V while not
 adding any decoration to argument ``b``.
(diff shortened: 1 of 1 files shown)


MrSidims:
@AlexeySotkin @svenvh please take a look

docs/SPIRVRepresentationInLLVM.rst
```diff
@@ -411,6 +411,53 @@ For example:
...
+
+``spirv.ParameterDecorations`` example:
+
+.. code-block:: llvm
+
+  !spirv.ParameterDecorations !1
+  ...
+  !1 = !{!2, !3}
+  !2 = !{}
+  !2 = !{i32 19}
```
AlexeySotkin, line 456:
```suggestion
  !3 = !{i32 19}
```

docs/SPIRVRepresentationInLLVM.rst
```diff
@@ -411,6 +411,53 @@ For example:
...
+``spirv.Decorations`` must refer to a decoration-list while
+``spirv.ParameterDecorations`` must refer to a metadata node that contains N
+references to decoration-lists, where N is the number of arguments of the
+function the metadata is tied to.
+
+``spirv.Decorations`` example:
+
+.. code-block:: llvm
+
+  !spirv.Decorations !1
```
AlexeySotkin, line 441: Could you please add the variable being decorated to the example.

docs/SPIRVRepresentationInLLVM.rst
```diff
@@ -411,6 +411,53 @@ For example:
...
+  !1 = !{!2}
+  !2 = !{i32 22}
+
+decorates a global variable with ``Constant`` in SPIR-V.
+
+``spirv.ParameterDecorations`` example:
+
+.. code-block:: llvm
+
+  !spirv.ParameterDecorations !1
```
AlexeySotkin, line 452: Could you please add the function being decorated to the example.

docs/SPIRVRepresentationInLLVM.rst
```diff
@@ -411,6 +411,53 @@ For example:
...
+function the metadata is tied to.
+
+``spirv.Decorations`` example:
+
+.. code-block:: llvm
+
+  !spirv.Decorations !1
+  ...
+  !1 = !{!2}
+  !2 = !{i32 22}
```
AlexeySotkin, line 444: Could you please clarify which metadata is "decoration-list" and which is "decoration-node"?

docs/SPIRVRepresentationInLLVM.rst
```diff
@@ -411,6 +411,53 @@ For example:
...
+Both function parameters and global variables can be decorated using LLVM
+metadata through the metadata names ``spirv.ParameterDecorations`` and
+``spirv.Decorations`` respectively. ``spirv.ParameterDecorations`` must be tied
+to the kernel function while ``spirv.Decorations`` is tied directly to the
+global variable.
+
+A "decoration-node" is a metadata node consisting of one or more operands. The
+first operand is an integer literal representing the SPIR-V decoration
+identifier. The other operands are either an integer or string literal
+representing the remaining extra operands of the corresponding SPIR-V
```
AlexeySotkin, line 426: Could you please add an example with extra operands?

docs/SPIRVRepresentationInLLVM.rst
steffenlarsen: This was also wrong, so good thing you pointed it out! It has been fixed. 😄

docs/SPIRVRepresentationInLLVM.rst
steffenlarsen: I have added the variable and use that as a reference in the description. Good call.

docs/SPIRVRepresentationInLLVM.rst
steffenlarsen: Absolutely. It has been added.

docs/SPIRVRepresentationInLLVM.rst
steffenlarsen: I have added comments on each line. Is that in line with what you were thinking?

docs/SPIRVRepresentationInLLVM.rst
steffenlarsen: I have added an example with `LinkageAttributes` which is both an example showing extra operands as well as both of the accepted operand types.

docs/SPIRVRepresentationInLLVM.rst
```diff
@@ -411,6 +411,56 @@ For example:
...
+
+  define spir_kernel void @k(float %a, float %b) #0 !spirv.ParameterDecorations !1
+  ...
+  !1 = !{!2, !3} ; metadata node with 2 decoration-lists
+  !2 = !{}       ; empty decoration-list
+  !3 = !{!4}     ; decoration-list with one decoration node
+  !4 = !{i32 19} ; decoration-node with no extra operands
+
+decorates the argument ``a`` of ``k`` with ``Restrict`` in SPIR-V while not
+adding any decoration to argument ``b``.
```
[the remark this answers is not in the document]
steffenlarsen: > `!2` refers to `%a` argument, while `!3` refers to `%b` argument. Is my understanding correct? If so, then it looks like `%b` is decorated with `Restrict`, not `%a`. Please correct me if I'm wrong.

My apologies, you are absolutely right. I must have gotten them mixed around when I added the function definition. I have fixed the references. Thanks for pointing it out!

> Also why do we need an additional decoration list `!3 = !{!4}`. Couldn't we have `!3` tied to a decoration node directly without intermediate one element decoration list? Like: `!3 = !{i32 19}`

The additional layer is to allow each parameter zero or more decorations. In this example we only use one for `!3` (namely the decoration in node `!4`) but we could have had a decoration in a `!5 = !{...}` and have had `!3 = !{!4, !5}`. Of course, for the cases where we have a single decoration we could remove the intermediate list, but for consistency I personally prefer this structure.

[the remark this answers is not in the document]
aratajew: Thanks for fixing this.
> Of course, for the cases where we have a single decoration we could remove the intermediate list, but for consistency I personally prefer this structure.

Sure, I agree. I suppose that SPIR-V Friendly IR generator would require special handling for optimizing out such a one element lists which is not worth.

AlexeySotkin (approved):

svenvh (approved):

MrSidims (approved):

MrSidims:
@aratajew were you questions answered? Could you please take another look?

aratajew (approved):

aratajew:
> @aratajew were you questions answered? Could you please take another look?

Yes, looks good to me :)

PR #1321, opened 2021-12-07
Move OpenCL relational builtin translation to SPIRVToOCL

hewj03:
Move OpenCL relational builtin translation to SPIRVToOCL

Motivation is to let standalone use of SPIRVToOCL20Legacy pass support translating relational builtin in Spirv Friendly IR.
Standalones use of SPIRVToOCL20Legacy pass may receive IR in spirv friendly IR:
```
 %call3.i.i.i = tail call spir_func <16 x i8> @_Z20__spirv_FOrdLessThanDv16_dS_(<16 x double> %1, <16 x double> %2)
```
Previously "Translate SPIR-V builtins to OCL 2.0 builtins (spvtoocl20)" pass will transform it into:
```
%call3.i.i.i = call spir_func <16 x i8> @_Z6islessDv16_dS_(<16 x double> %1, <16 x double> %2)
```
After this patch, the pass will transform it into:
```
%call3.i.i.i11 = call spir_func <16 x i64> @_Z6islessDv16_dS_(<16 x double> %1, <16 x double> %2) 
```

This patch also aligns relational builtin in SPIRV Friendly IR with spirv spec. i8 is used for vector boolean type, since clang frontend can't generate i1 vector type for spirv builtin yet.
E.g.
```
  i32 @_Z11__spirv_AnyDv2_i(<2 x i32>
```
is changed to
```
  i1 @_Z11__spirv_AnyDv2_c(<2 x i8>
```

### DIFF
--- a/test/relationals.ll
+++ b/test/relationals.ll
@@ -14,17 +14,17 @@ declare dso_local spir_func <4 x i8> @_Z16__spirv_IsFiniteIDv4_aDv4_fET_T0_(<4 x
 declare dso_local spir_func <4 x i8> @_Z16__spirv_IsNormalIDv4_aDv4_fET_T0_(<4 x float>)
 declare dso_local spir_func <4 x i8> @_Z18__spirv_SignBitSetIDv4_aDv4_fET_T0_(<4 x float>)
 
-; CHECK-SPV-LLVM: call spir_func <4 x i32> @_Z13__spirv_IsNanDv4_f
-; CHECK-SPV-LLVM: call spir_func <4 x i32> @_Z13__spirv_IsInfDv4_f
-; CHECK-SPV-LLVM: call spir_func <4 x i32> @_Z16__spirv_IsFiniteDv4_f
-; CHECK-SPV-LLVM: call spir_func <4 x i32> @_Z16__spirv_IsNormalDv4_f
-; CHECK-SPV-LLVM: call spir_func <4 x i32> @_Z18__spirv_SignBitSetDv4_f
+; CHECK-SPV-LLVM: call spir_func <4 x i8> @_Z13__spirv_IsNanDv4_f
+; CHECK-SPV-LLVM: call spir_func <4 x i8> @_Z13__spirv_IsInfDv4_f
+; CHECK-SPV-LLVM: call spir_func <4 x i8> @_Z16__spirv_IsFiniteDv4_f
+; CHECK-SPV-LLVM: call spir_func <4 x i8> @_Z16__spirv_IsNormalDv4_f
+; CHECK-SPV-LLVM: call spir_func <4 x i8> @_Z18__spirv_SignBitSetDv4_f
 
-; CHECK-SPV-LLVM: declare spir_func <4 x i32> @_Z13__spirv_IsNanDv4_f(<4 x float>)
-; CHECK-SPV-LLVM: declare spir_func <4 x i32> @_Z13__spirv_IsInfDv4_f(<4 x float>)
-; CHECK-SPV-LLVM: declare spir_func <4 x i32> @_Z16__spirv_IsFiniteDv4_f(<4 x float>)
-; CHECK-SPV-LLVM: declare spir_func <4 x i32> @_Z16__spirv_IsNormalDv4_f(<4 x float>)
-; CHECK-SPV-LLVM: declare spir_func <4 x i32> @_Z18__spirv_SignBitSetDv4_f(<4 x float>)
+; CHECK-SPV-LLVM: declare spir_func <4 x i8> @_Z13__spirv_IsNanDv4_f(<4 x float>)
+; CHECK-SPV-LLVM: declare spir_func <4 x i8> @_Z13__spirv_IsInfDv4_f(<4 x float>)
+; CHECK-SPV-LLVM: declare spir_func <4 x i8> @_Z16__spirv_IsFiniteDv4_f(<4 x float>)
+; CHECK-SPV-LLVM: declare spir_func <4 x i8> @_Z16__spirv_IsNormalDv4_f(<4 x float>)
+; CHECK-SPV-LLVM: declare spir_func <4 x i8> @_Z18__spirv_SignBitSetDv4_f(<4 x float>)
 
 ; CHECK-SPIRV: {{[0-9]+}} TypeBool [[TBool:[0-9]+]]
 ; CHECK-SPIRV: {{[0-9]+}} TypeVector [[TBoolVec:[0-9]+]] [[TBool]]
--- a/test/transcoding/OpAllAny.ll
+++ b/test/transcoding/OpAllAny.ll
@@ -3,54 +3,104 @@
 ; RUN: FileCheck < %t.txt %s --check-prefix=CHECK-SPIRV
 ; RUN: llvm-spirv %t.bc -o %t.spv
 ; RUN: spirv-val %t.spv
+; RUN: llvm-spirv -r %t.spv -o %t.rev.spv.bc --spirv-target-env=SPV-IR
+; RUN: llvm-dis < %t.rev.spv.bc | FileCheck %s --check-prefix=CHECK-SPV-IR
 ; RUN: llvm-spirv -r %t.spv -o %t.rev.bc
 ; RUN: llvm-dis < %t.rev.bc | FileCheck %s --check-prefix=CHECK-LLVM
-; RUN: llvm-spirv -r %t.spv -o %t.rev.bc --spirv-target-env=SPV-IR
-; RUN: llvm-dis < %t.rev.bc | FileCheck %s --check-prefix=CHECK-SPV-LLVM
 
-; CHECK-LLVM: call spir_func i32 @_Z3allDv2_i(
-; CHECK-LLVM: call spir_func i32 @_Z3anyDv2_i(
-
-; CHECK-SPV-LLVM: call spir_func i32 @_Z11__spirv_AllDv2_i(
-; CHECK-SPV-LLVM: call spir_func i32 @_Z11__spirv_AnyDv2_i(
-
-; CHECK-SPV-LLVM: declare spir_func i32 @_Z11__spirv_AllDv2_i(<2 x i32>)
-; CHECK-SPV-LLVM: declare spir_func i32 @_Z11__spirv_AnyDv2_i(<2 x i32>)
+; This test checks SYCL relational builtin any and all with vector input types.
 
 ; CHECK-SPIRV: 2 TypeBool [[BoolTypeID:[0-9]+]]
-; CHECK-SPIRV: 4 All [[BoolTypeID]]
+
 ; CHECK-SPIRV: 4 Any [[BoolTypeID]]
+; CHECK-SPIRV: 4 Any [[BoolTypeID]]
+; CHECK-SPIRV: 4 Any [[BoolTypeID]]
+; CHECK-SPIRV: 4 Any [[BoolTypeID]]
+; CHECK-SPIRV: 4 All [[BoolTypeID]]
+; CHECK-SPIRV: 4 All [[BoolTypeID]]
+; CHECK-SPIRV: 4 All [[BoolTypeID]]
+; CHECK-SPIRV: 4 All [[BoolTypeID]]
+
+; CHECK-SPV-IR: call spir_func i1 @_Z11__spirv_AnyDv2_c(<2 x i8>
+; CHECK-SPV-IR: call spir_func i1 @_Z11__spirv_AnyDv2_c(<2 x i8>
+; CHECK-SPV-IR: call spir_func i1 @_Z11__spirv_AnyDv2_c(<2 x i8>
+; CHECK-SPV-IR: call spir_func i1 @_Z11__spirv_AnyDv2_c(<2 x i8>
+; CHECK-SPV-IR: call spir_func i1 @_Z11__spirv_AllDv2_c(<2 x i8>
+; CHECK-SPV-IR: call spir_func i1 @_Z11__spirv_AllDv2_c(<2 x i8>
+; CHECK-SPV-IR: call spir_func i1 @_Z11__spirv_AllDv2_c(<2 x i8>
+; CHECK-SPV-IR: call spir_func i1 @_Z11__spirv_AllDv2_c(<2 x i8>
+
+; CHECK-LLVM: call spir_func i32 @_Z3anyDv2_c(<2 x i8>
+; CHECK-LLVM: call spir_func i32 @_Z3anyDv2_c(<2 x i8>
+; CHECK-LLVM: call spir_func i32 @_Z3anyDv2_c(<2 x i8>
+; CHECK-LLVM: call spir_func i32 @_Z3anyDv2_c(<2 x i8>
+; CHECK-LLVM: call spir_func i32 @_Z3allDv2_c(<2 x i8>
+; CHECK-LLVM: call spir_func i32 @_Z3allDv2_c(<2 x i8>
+; CHECK-LLVM: call spir_func i32 @_Z3allDv2_c(<2 x i8>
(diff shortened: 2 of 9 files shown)


hewj03:
Hi @AlexeySotkin @AlexeySachkov , could you please take a look?

MrSidims:
From the bottom of my heart I support this fix, but I quite don't understand the reasoning put in the description. I'd expect the pass not to change the type, unless it's special OpenCL type, like image or sampler. Also could you please elaborate more about changes of LLVM IR in the existing tests? I believe, they should be modified in a way to check SPIR-V friendly IR and 'OpenCL-friendly' IR under translation options, for example like here https://github.com/KhronosGroup/SPIRV-LLVM-Translator/pull/1329/

hewj03:
@MrSidims Thank you for the review.

> From the bottom of my heart I support this fix, but I quite don't understand the reasoning put in the description. I'd expect the pass not to change the type, unless it's special OpenCL type, like image or sampler.

Let me take SYCL builtin isfinite as example, according to sycl spec:
```
declare spir_func <16 x i64> @_Z8isfiniteDv2_d(<16 x double>)
```
It is implemented by SPIRV builtin in https://github.com/intel/llvm/blob/cabb43fdb137f6cc65fc6486a1b6be1300185df7/clang/lib/Sema/SPIRVBuiltins.td#L859 and the builtin's return type is <16 x i8>, although according to https://github.com/intel/llvm/pull/5031#issuecomment-984613321 the return type should be <16 x i1> instead.
```
%call1.i.i.i = tail call spir_func <16 x i8> @_Z16__spirv_IsFiniteDv16_d(<16 x double> %9)
```
Then <16 x i8> return value is converted to <16 x i64> in sycl builtin header.
spvtoocl20 pass has to modify _Z16__spirv_IsFiniteDv16_d return value type so that it matches with OpenCL builtin declaration.
> Also could you please elaborate more about changes of LLVM IR in the existing tests? I believe, they should be modified in a way to check SPIR-V friendly IR and 'OpenCL-friendly' IR under translation options, for example like here #1329

Added new "--spirv-target-env=SPV-IR" test.
"CHECK-SPV-IR" will checks that builtin declaration follows SPIRV spec and "CHECK-LLVM" checks that builtin declaration follows OpenCL spec.

test/transcoding/OpAllAny.ll
```diff
@@ -3,54 +3,104 @@
...
+; CHECK-SPV-IR: call spir_func i1 @_Z11__spirv_AllDv2_b(<2 x i1>
+
+; CHECK-LLVM: call spir_func i32 @_Z3anyDv2_i(<2 x i32>
+; CHECK-LLVM: call spir_func i32 @_Z3anyDv2_i(<2 x i32>
+; CHECK-LLVM: call spir_func i32 @_Z3anyDv2_i(<2 x i32>
+; CHECK-LLVM: call spir_func i32 @_Z3anyDv2_i(<2 x i32>
+; CHECK-LLVM: call spir_func i32 @_Z3allDv2_i(<2 x i32>
+; CHECK-LLVM: call spir_func i32 @_Z3allDv2_i(<2 x i32>
+; CHECK-LLVM: call spir_func i32 @_Z3allDv2_i(<2 x i32>
+; CHECK-LLVM: call spir_func i32 @_Z3allDv2_i(<2 x i32>
```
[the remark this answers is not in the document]
hewj03: Maybe the test is a little confusing.

Sycl header in https://github.com/intel/llvm won't generate <2 x i64> input for any/all. Sycl header already trunc i64 input to i8 because spirv builtin only accepts i8 input, see https://github.com/intel/llvm/blob/cabb43fdb137f6cc65fc6486a1b6be1300185df7/clang/lib/Sema/SPIRVBuiltins.td#L856

spirv-translator itself won't generate <2 x i64> input neither, because it follows spirv spec.

Are there any other source that will generate <2 x i64>? I'm not aware of that. If such source may exist, we can use i64 return type in SPIRVToOCLBase::visitCallSPIRVAnyAll, what's your opinion?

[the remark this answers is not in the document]
aratajew: OpenCL C specification:
```
int all(igentype x)
The argument type igentype refers to the built-in signed integer types i.e. char, charn, short, shortn, int, intn, long and longn.
```
Clang can generate <2 x i64>: https://godbolt.org/z/rbYz1xnKf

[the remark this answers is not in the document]
hewj03: If the input IR is <2 x i64> any/all from OpenCL, spvtoocl20 doesn't do transformation and IR is not modified.
spvtoocl20 translates spirv friendly IR, so its input type is defined by spirv friendly IR format and available implementations such as in sycl header or llvm-spirv translator itself. 

Since this patch doesn't change the behavior of using i32 as representation type for OpenCL output, I'd suggest to open a new issue for possible i64 input. How do you think?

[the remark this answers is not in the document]
aratajew: Actually, you're right. Since SPIRV Friendly IR representation operates on boolean inputs then extending them to i32 vs extending them to i64 doesn't make any difference. Your code is correct.

I'm just wondering why OpenCL C allows for different vector types `char, charn, short, shortn, int, intn, long and longn.`, while SPIRV representation operates only on boolean vectors.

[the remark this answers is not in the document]
hewj03: yes, such difference exists, and SYCL builtin signatures are mostly aligned with OpenCL C builtin .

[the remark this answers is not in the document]
hewj03: I'm not aware of the rational either.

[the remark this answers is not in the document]
aratajew: OpenCL C representation allows for more vector types, but it actually care only for one bit anyway (`"Returns 1 if the most significant bit...is set"`), so it looks like SPIRV representation was just simplified, but it still represents the same functionallity.

aratajew (approved):

test/transcoding/OpAllAny.ll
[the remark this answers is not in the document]
Naghasan: > I'm just wondering why OpenCL C allows for different vector types char, charn, short, shortn, int, intn, long and longn., while SPIRV representation operates only on boolean vectors.

Simply because SPIR-V is not limited to represent OpenCL C (SPIR-V has more ties to Vulkan than to OpenCL).

test/relationals.ll
```diff
@@ -14,17 +14,17 @@ declare dso_local spir_func <4 x i8> @_Z16__spirv_IsFiniteIDv4_aDv4_fET_T0_(<4 x
...
-; CHECK-SPV-LLVM: call spir_func <4 x i32> @_Z13__spirv_IsNanDv4_f
-; CHECK-SPV-LLVM: call spir_func <4 x i32> @_Z13__spirv_IsInfDv4_f
-; CHECK-SPV-LLVM: call spir_func <4 x i32> @_Z16__spirv_IsFiniteDv4_f
-; CHECK-SPV-LLVM: call spir_func <4 x i32> @_Z16__spirv_IsNormalDv4_f
-; CHECK-SPV-LLVM: call spir_func <4 x i32> @_Z18__spirv_SignBitSetDv4_f
+; CHECK-SPV-LLVM: call spir_func <4 x i1> @_Z13__spirv_IsNanDv4_f
+; CHECK-SPV-LLVM: call spir_func <4 x i1> @_Z13__spirv_IsInfDv4_f
+; CHECK-SPV-LLVM: call spir_func <4 x i1> @_Z16__spirv_IsFiniteDv4_f
+; CHECK-SPV-LLVM: call spir_func <4 x i1> @_Z16__spirv_IsNormalDv4_f
+; CHECK-SPV-LLVM: call spir_func <4 x i1> @_Z18__spirv_SignBitSetDv4_f
```
Naghasan, line 21: Sorry jumping into this late, is `i1` the type we want for builtins returning bools ?

We need to factor in the capability of a frontend to emit the correct return type. While `i1` and vectors of `i1` could be used, it is not a given the source language will allow this expression. The frontend can always be taught this, but that means you need to change the compiler.

AFAIK, that should be OK from a opencl C or C++ point of view, but the translator is not limited to these input languages.

@AnastasiaStulova @bader do you have any thoughts on that ?

AnastasiaStulova: OpenCL frontend won't produce this IR because the functions return int type:
https://godbolt.org/z/bqz7T9cjv

bader: I agree with @Naghasan. @hewj03, could you clarify how do you see clang-based front-end to emit vector of i1? Anyway to store such values in memory i1 must be extended to i8 (at least).

Why it's only SPIR->OCL translation? We can argue that front-end issues are not relevant and these built-ins can be implemented in LLVM IR. But I don't think it's a good idea to put such restriction. In addition to that this change breaks code consistency after round-trip translation <n x i8> -> SPIR-V -> <n x i1>.

I suggest keeping existing return type for SPIR-V built-ins as update OpenCL built-ins only as described in the PR description.

hewj03: @Naghasan @bader thank you for catching this. This patch didn't intend to change spirv-friendly IR. Please review the new change.

i1 is from https://github.com/KhronosGroup/SPIRV-LLVM-Translator/blob/5f289bae97f2815199a555a0c8151d7e2f36305c/lib/SPIRV/SPIRVReader.cpp#L375

test/transcoding/relationals_double.ll
```diff
@@ -3,68 +3,279 @@
...
+; CHECK-SPV-IR: fcmp une double
+; CHECK-SPV-IR: fcmp ogt double
+; CHECK-SPV-IR: fcmp oge double
+; CHECK-SPV-IR: fcmp olt double
+; CHECK-SPV-IR: fcmp ole double
+; CHECK-SPV-IR: fcmp one double
+; CHECK-SPV-IR: fcmp ord double
+; CHECK-SPV-IR: fcmp uno double
+
+; CHECK-SPV-IR: call spir_func <2 x i64> @_Z16__spirv_IsFiniteDv2_d(<2 x double>
```
hewj03, line 64: not related to this patch, I think we shall unify the return type in spirv-friendly IR for vector output:
; CHECK-SPV-IR: call spir_func <2 x i64> @_Z16__spirv_IsFiniteDv2_d(<2 x double>
; CHECK-SPV-IR: call spir_func <2 x i32> @_Z16__spirv_IsFiniteDv2_f(<2 x float>
; CHECK-SPV-IR: call spir_func <2 x i16> @_Z16__spirv_IsFiniteDv2_Dh(<2 x half>

@Naghasan how do you think?

test/relationals.ll
hewj03: llvm based backend can consume spirv-friendly IR as input. If a builtin returning vector of i1 is not inlined, as I know, x86 backend won't has abi issue, but I don't know about other backends.

test/transcoding/relationals_double.ll
aratajew: Could we always represent boolean vector as a vector of i8 in SPIRV Friendly IR? I think that such rule should be applied to all opcodes operating on boolean vector and should also be documented in https://github.com/KhronosGroup/SPIRV-LLVM-Translator/blob/master/docs/SPIRVRepresentationInLLVM.rst.

test/relationals.ll
Naghasan: > OpenCL frontend won't produce this IR because the functions return int type:
https://godbolt.org/z/bqz7T9cjv

@AnastasiaStulova your example is about the opencl c builtins. My point is about the SPIR-V insn https://www.khronos.org/registry/SPIR-V/specs/unified1/SPIRV.html#OpIsNan

Specifically:
```
Result Type must be a scalar or vector of Boolean type.
```

```
In addition to that this change breaks code consistency after round-trip translation

 I suggest keeping existing return type for SPIR-V built-ins
```

I don't think it makes sense for SPIR-V builtin functions to not follow the SPIR-V spec...

------

My point is what should a boolean maps to when it is taken as input or output of a SPIR-V builtin function, and more specifically: would mapping this to `i1` cause be a problem for a frontend to emit these ? 

Seems clang still don't allow for vector of bools https://godbolt.org/z/8Ts1c68Ta, so i1 is a problem in general.

test/transcoding/relationals_double.ll
Naghasan: answered above, I agree with @aratajew, `i8` is probably the most portable type from a frontend perspective

hewj03:
clang-tidy error is not related to this patch. I get same fail on SPIRV-LLVM-Translator trunk:
```
SPIRV-LLVM-Translator/SPIRV-LLVM-Translator/lib/SPIRV/SPIRVReader.cpp:4245:27: error: function 'transOCLAllAny' is within a recursive call chain [misc-no-recursion,-warnings-as-errors]
SPIRV-LLVM-Translator/SPIRV-LLVM-Translator/lib/SPIRV/SPIRVReader.cpp:4245:27: error: function 'transOCLAllAny' is within a recursive call chain [misc-no-recursion,-warnings-as-errors]
```

test/transcoding/relationals_double.ll
aratajew: Intel GPU Backend also expects that boolean vectors will be represented as i8 vectors: https://github.com/intel/intel-graphics-compiler/blob/master/IGC/BiFModule/Headers/spirv.h#L153-L157 So it sounds reasonable to me to make it this way.

hewj03: if there is no objection, I can update this PR to change SPIRV Friendly IR return type of relational builtin to match with https://github.com/intel/llvm/blob/8c5b7017a925701ef4034056b5ed8e0fac2a0011/clang/lib/Sema/SPIRVBuiltins.td#L855

hewj03: > if there is no objection, I can update this PR to change SPIRV Friendly IR return type of relational builtin to match with https://github.com/intel/llvm/blob/8c5b7017a925701ef4034056b5ed8e0fac2a0011/clang/lib/Sema/SPIRVBuiltins.td#L855

done

aratajew (approved):

test/transcoding/relationals_double.ll
aratajew: @MrSidims, @bader what do you think about current solution? Do you agree to have it this way? If so, it would be great to merge it.

MrSidims: Let me revisit the PR by EO tomorrow.

bader: I think we should get approval from @Naghasan and resolve merge conflicts before we can merge it.

hewj03: merge conflict resolved. @Naghasan please review.

test/relationals.ll
MrSidims: Not sure if there was an argument, but just a note:
I believe, that SPIR-V spec is not violated here see in :
```
; CHECK-SPIRV: {{[0-9]+}} TypeBool [[TBool:[0-9]+]]
; CHECK-SPIRV: {{[0-9]+}} IsNan [[TBoolVec]] [[IsNanRes:[0-9]+]]
```
so since the SPIR-V module is valid -the SPIR-V consumer can do whatever reasonable operations with it, for this case transforming a vector of i1 to a vector of i8.

MrSidims (approved):
LGTM
Nit: can we consider renaming the patch from:
"Align relational builtin in SPIRV Friendly IR with spirv spec, move OpenCL relational builtin translation to SPIRVToOCL"
to just:
"Move OpenCL relational builtin translation to SPIRVToOCL"
to make it a one-liner name?

hewj03:
> LGTM Nit: can we consider renaming the patch from: "Align relational builtin in SPIRV Friendly IR with spirv spec, move OpenCL relational builtin translation to SPIRVToOCL" to just: "Move OpenCL relational builtin translation to SPIRVToOCL" to make it a one-liner name?

done

Naghasan (approved):
LGTM, thanks

PR #1322, opened 2021-12-08
Restrict special handling of sampler OpVariable only to UniformConstant

aratajew:
This is a fixup for aaa3cb328dfc1f56804da28ff949c7428678a00d.


aratajew:
@svenvh, @MrSidims could you please take a look? The faster we merge it the better, since this is a fix for the regression introduced in aaa3cb3.

MrSidims (approved):
LGTM
I don't have a strong opinion whether the test should be added

PR #1330, opened 2021-12-09
Fix handling of complex constant expression vectors inside metadata

vmaksimo:
Removed restriction for all the vector elements to be constant expressions as some of them might be `undef`.

### DIFF
--- a/lib/SPIRV/SPIRVLowerConstExpr.cpp
+++ b/lib/SPIRV/SPIRVLowerConstExpr.cpp
@@ -139,7 +139,7 @@ void SPIRVLowerConstExprBase::visit(Module *M) {
       auto II = WorkList.front();
 
       auto LowerOp = [&II, &FBegin, &I](Value *V) -> Value * {
-        if (isa<Function>(V))
+        if (isa<Function>(V) || !isa<ConstantExpr>(V))
           return V;
         auto *CE = cast<ConstantExpr>(V);
         SPIRVDBG(dbgs() << "[lowerConstantExpressions] " << *CE;)
@@ -170,7 +170,7 @@ void SPIRVLowerConstExprBase::visit(Module *M) {
       auto LowerConstantVec = [&II, &LowerOp, &WorkList,
                                &M](ConstantVector *Vec,
                                    unsigned NumOfOp) -> Value * {
-        if (std::all_of(Vec->op_begin(), Vec->op_end(), [](Value *V) {
+        if (std::any_of(Vec->op_begin(), Vec->op_end(), [](Value *V) {
               return isa<ConstantExpr>(V) || isa<Function>(V);
             })) {
           // Expand a vector of constexprs and construct it back with
--- a/test/vector-metadata-constexpr.ll
+++ b/test/vector-metadata-constexpr.ll
@@ -1,6 +1,6 @@
 ; RUN: llvm-as %s -o %t.bc
-; RUN: llvm-spirv -s %t.bc -o - | llvm-dis -o - | FileCheck %s
-; RUN: llvm-spirv %t.bc
+; RUN: llvm-spirv -s --spirv-ext=+SPV_INTEL_arbitrary_precision_integers,+SPV_INTEL_vector_compute %t.bc -o - | llvm-dis -o - | FileCheck %s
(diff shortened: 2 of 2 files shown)


vmaksimo:
Tagging @MrSidims @AlexeySotkin to take a look

test/vector-metadata-constexpr.ll
```diff
@@ -1,15 +1,25 @@
 ; RUN: llvm-as %s -o %t.bc
-; RUN: llvm-spirv -s %t.bc -o - | llvm-dis -o - | FileCheck %s
-; RUN: llvm-spirv %t.bc
+; RUN: llvm-spirv -s --spirv-ext=+SPV_INTEL_arbitrary_precision_integers,+SPV_INTEL_vector_compute %t.bc -o - | llvm-dis -o - | FileCheck %s
+; RUN: llvm-spirv --spirv-ext=+SPV_INTEL_arbitrary_precision_integers,+SPV_INTEL_vector_compute %t.bc
```
AlexeySotkin, line 3: Why did you add the extensions?

vmaksimo: I've added test case with the vector of size 1 which is supported only with the `SPV_INTEL_vector_compute` extension, and the second extension is not needed at all.
I'll change the test a bit to avoid using extensions.

AlexeySotkin: Thanks!

vmaksimo: done in 1bc8f8b

lib/SPIRV/SPIRVLowerConstExpr.cpp
```diff
@@ -170,7 +170,7 @@ void SPIRVLowerConstExprBase::visit(Module *M) {
       auto LowerConstantVec = [&II, &LowerOp, &WorkList,
                                &M](ConstantVector *Vec,
                                    unsigned NumOfOp) -> Value * {
-        if (std::all_of(Vec->op_begin(), Vec->op_end(), [](Value *V) {
+        if (std::any_of(Vec->op_begin(), Vec->op_end(), [](Value *V) {
```
AlexeySotkin, line 173: Do we really need this `if`? Is it possible to have a non-constant value in `ConstantVector`?

AlexeySotkin:
@Fznamznon could you please take a look?

AlexeySotkin:
In general this patch looks good to me.

However the whole approach with replacing constant expressions with non-constant instructions seems to be not quite efficient from size of generated module perspective. If we have multiple `llvm.dbg.value` with such complex constant expressions, we may increase size of the code significantly. Translating constant expressions to `OpSpecConstantOp` seems to be more efficient but perhaps harder solution.

lib/SPIRV/SPIRVLowerConstExpr.cpp
Fznamznon: From commit message and the test it seems `undef` if not a constant value, and can be met inside a `ConstantVector`.

Fznamznon (approved):
LGTM

lib/SPIRV/SPIRVLowerConstExpr.cpp
Fznamznon: I don't know what is the purpose of this `if` though.

vmaksimo:
> In general this patch looks good to me.
> 
> However the whole approach with replacing constant expressions with non-constant instructions seems to be not quite efficient from size of generated module perspective. If we have multiple `llvm.dbg.value` with such complex constant expressions, we may increase size of the code significantly. Translating constant expressions to `OpSpecConstantOp` seems to be more efficient but perhaps harder solution.

If we choose to go with the more efficient approach (less size of a generated module) then we'll need to drop the existing lowering of constant expressions.
I have a draft patch that enables such an approach, but it does not rework the existing lowering pass. @AlexeySotkin does it seems a good idea to just drop the existing pass?

vmaksimo:
>  UPD: I believe it'll be enough to drop the lowering of constant expression vectors, not all the pass. And this functionality will be covered with the translation to `OpSpecConstantOp`.

Per my experiments done, the approach from #1335 is not perfect and can't be finalized before the SPIR-V spec change is introduced (mentioned here: https://github.com/KhronosGroup/SPIRV-Tools/issues/4272).

I suggest proceeding with the solution offered by this PR.

PR #1331, opened 2021-12-10
Fix build of check-llvm-spirv target if SPIRV-Tools are not found

vmaksimo:
The problem is present after beb88bf.
If SPIRV-Tools are not found during the compilation, the build fails with the syntax error in the config file:
```
build/projects/SPIRV-LLVM-Translator/test/lit.site.cfg.py", line 26
    config.spirv_tools_found =
                              ^
SyntaxError: invalid syntax
```


svenvh (approved):
Sorry for the breakage, thanks for the fix!

Issue #1216, opened 2021-09-20
SPIR-V 1.4: Interface variables not listed in Interface of OpEntryPoint


StuartDBrady:
When run on SPIR-V 1.4 output produced by the translator, the SPIR-V Validator may report errors of the following form:

> error: line _n_: Interface variable id <_var_id_> is used by entry point '_name_' id <_kernel_id_>, but it is not listed as an interface
>   `%out = OpVariable %_ptr_UniformConstant_type UniformConstant %type`

This can be reproduced with a trivial program using a `__constant` variable (i.e. using the _UniformConstant_ storage class), such as:
```
__constant int var = 0;
__kernel void test()
{
    (void)(var * 2);
}
```

resulting in the LLVM IR:
```
  %1 = load i32, i32 addrspace(2)* @var.const, align 4
  %2 = mul nsw i32 %1, 2
```
It can also be reproduced with the _CrossWorkgroup_ and _Workgroup_ storage classes (i.e. using global and local variables, respectively). I am unsure whether this issue is applicable to the _Generic_ and _Input_ storage classes.

This appears to relate to the _OpEntryPoint_ instruction, which prior to SPIR-V 1.4 had an _Interface_ variable list that was limited to variables of the _Input_ and shader-specific _Output_ storage classes only (and so excluding the _UniformConstant_ storage class, for example). In SPIR-V 1.4 and later, variables of all storage classes must be included in _Interface_.

In the generated SPIR-V for the example above, the _Interface_ of the _OpEntryPoint_ instruction was empty.

Approximately 5850 out of 50,000 SPIR-V files generated from the OpenCL Conformance Test Suite reproduced this validation failure.

StuartDBrady:
Note, this was originally reported as issue #1205, but it turns out to be a more general problem than initially described. The use of OpenCL C code blocks is one specific case of this, which is affected because of the use of "block literal" globals in the implementation of OpenCL C code blocks. Once interface variables are correctly listed, the problem should be resolved for OpenCL C code blocks too with no further action needed.

I don't have any particular suggestions as to how this should be resolved, other than to say that it would seem to require an analysis pass to walk the call graph of LLVM functions as part of the translation.

It's not clear to me what the Interface list should contain for functions that call externally defined functions. How is it possible to know all of the globals used by an external function? Is this even required?

AlexeySotkin:
[...]

PR #1177, opened 2021-08-24
Add SPIR-V 1.4 checks

MrSidims:
Signed-off-by: Dmitry Sidorov <dmitry.sidorov@intel.com>

PR #1334, opened 2021-12-13
Fix the collection of entry point interfaces

KornevNikita:
This is a patch to expand the collection of entry point interfaces.
In SPIR-V 1.4 and later OpEntryPoint must list all global variables in the
interface. Also fix quoted string output in SPIRV text format.
See: https://github.com/KhronosGroup/SPIRV-LLVM-Translator/issues/1216

### DIFF
--- a/lib/SPIRV/SPIRVWriter.cpp
+++ b/lib/SPIRV/SPIRVWriter.cpp
@@ -3570,6 +3570,12 @@ bool LLVMToSPIRVBase::isAnyFunctionReachableFromFunction(
 void LLVMToSPIRVBase::collectEntryPointInterfaces(SPIRVFunction *SF,
                                                   Function *F) {
   for (auto &GV : M->globals()) {
+    if (SF->getModule()->getSPIRVVersion() <
+        static_cast<uint32_t>(VersionNumber::SPIRV_1_4)) {
+      const auto AS = GV.getAddressSpace();
+      if (AS != SPIRAS_Input && AS != SPIRAS_Output)
+        continue;
+    }
     std::unordered_set<const Function *> Funcs;
 
     for (const auto &U : GV.uses()) {
--- a/lib/SPIRV/SPIRVWriter.cpp
+++ b/lib/SPIRV/SPIRVWriter.cpp
@@ -3575,7 +3575,7 @@ LLVMToSPIRVBase::collectEntryPointInterfaces(SPIRVFunction *SF, Function *F) {
       if (AS != SPIRAS_Input && AS != SPIRAS_Output)
         continue;
     } else {
-      BM->setSPIRVVersion(static_cast<SPIRVWord>(VersionNumber::SPIRV_1_4));
+      BM->setMinSPIRVVersion(static_cast<SPIRVWord>(VersionNumber::SPIRV_1_4));
     }
     std::unordered_set<const Function *> Funcs;
 
--- a/lib/SPIRV/SPIRVWriter.cpp
+++ b/lib/SPIRV/SPIRVWriter.cpp
@@ -3587,7 +3587,7 @@ LLVMToSPIRVBase::collectEntryPointInterfaces(SPIRVFunction *SF, Function *F) {
     if (isAnyFunctionReachableFromFunction(F, Funcs)) {
       SPIRVWord ModuleVersion = static_cast<SPIRVWord>(BM->getSPIRVVersion());
       if (AS != SPIRAS_Input && AS != SPIRAS_Output &&
-           ModuleVersion < static_cast<SPIRVWord>(VersionNumber::SPIRV_1_4))
+          ModuleVersion < static_cast<SPIRVWord>(VersionNumber::SPIRV_1_4))
         BM->setMinSPIRVVersion(
             static_cast<SPIRVWord>(VersionNumber::SPIRV_1_4));
       Interface.push_back(ValueMap[&GV]->getId());
--- a/lib/SPIRV/SPIRVWriter.cpp
+++ b/lib/SPIRV/SPIRVWriter.cpp
@@ -3569,27 +3569,28 @@ std::vector<SPIRVId>
 LLVMToSPIRVBase::collectEntryPointInterfaces(SPIRVFunction *SF, Function *F) {
   std::vector<SPIRVId> Interface;
   for (auto &GV : M->globals()) {
+    const auto AS = GV.getAddressSpace();
     SPIRVModule *BM = SF->getModule();
-    if (!BM->isAllowedToUseVersion(VersionNumber::SPIRV_1_4)) {
-      const auto AS = GV.getAddressSpace();
+    if (!BM->isAllowedToUseVersion(VersionNumber::SPIRV_1_4))
       if (AS != SPIRAS_Input && AS != SPIRAS_Output)
         continue;
-    } else {
-      std::unordered_set<const Function *> Funcs;
 
-      for (const auto &U : GV.uses()) {
-        const Instruction *Inst = dyn_cast<Instruction>(U.getUser());
-        if (!Inst)
-          continue;
-        Funcs.insert(Inst->getFunction());
-      }
+    std::unordered_set<const Function *> Funcs;
 
-      if (isAnyFunctionReachableFromFunction(F, Funcs)) {
-        if (static_cast<SPIRVWord>(BM->getSPIRVVersion()) <
-            static_cast<SPIRVWord>(VersionNumber::SPIRV_1_4))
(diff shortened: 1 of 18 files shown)


lib/SPIRV/libSPIRV/SPIRVModule.cpp
```diff
@@ -1834,9 +1841,20 @@ spv_ostream &operator<<(spv_ostream &O, SPIRVModule &M) {
...
 
   for (auto &I : MI.EntryPointVec)
-    for (auto &II : I.second)
+    for (auto &II : I.second) {
+      // getVariables() will return a non-empty vector only when translating
+      // LLVM-IR -> SPIRV text, because the entry points was translated as
+      // SPIRVFunction and their interfaces are saved as variables by
+      // collectEntryPointInterfaces function. When translating SPIRV bitcode ->
+      // SPIRV text the entry points are decoded from the bitcode and the
+      // interfaces are stored to EntryPointInterfaces map to not to be lost.
```
AlexeySotkin, line 1850: Why do we need extra containers in `SPIRVModule` and `SPIRVFunction` classes to store interface ids? I mean can we store the ids in `SPIRVEntryPoint::Variables`?

test/entry-point-interfaces.ll
```diff
@@ -0,0 +1,52 @@
+; RUN: llvm-as %s -o %t.bc
+
+; RUN: llvm-spirv %t.bc -o %t.spv
+; RUN: spirv-val %t.spv
```
AlexeySotkin, line 4:
```suggestion
; RUN: spirv-val --target-env spv1.4 %t.spv
```

lib/SPIRV/libSPIRV/SPIRVEntry.cpp
```diff
@@ -541,9 +541,14 @@ void SPIRVEntryPoint::encode(spv_ostream &O) const {
 }
 
 void SPIRVEntryPoint::decode(std::istream &I) {
-  getDecoder(I) >> ExecModel >> Target >> Name >> Variables;
+  getDecoder(I) >> ExecModel >> Target >> Name;
+  // One fixed word is the name of the entry point. When we subtract FixedWC
```
AlexeySotkin, line 545: The name of the entry point can occupy more than one word. But for the `FixedWC` we assume that the name takes _at least_ one word. So your calculations are correct. The comment is just a bit confusing.

lib/SPIRV/libSPIRV/SPIRVModule.cpp
```diff
@@ -129,15 +129,15 @@ class SPIRVModuleImpl : public SPIRVModule {
...
   }
   SPIRVFunction *getEntryPoint(SPIRVExecutionModelKind EM,
                                unsigned I) const override {
-    auto Loc = EntryPointVec.find(EM);
-    if (Loc == EntryPointVec.end())
+    auto Loc = EntryPointMapVec.find(EM);
+    if (Loc == EntryPointMapVec.end())
       return nullptr;
     assert(I < Loc->second.size());
     return get<SPIRVFunction>(Loc->second[I]);
```
AlexeySotkin, line 143: AFAIK there are no usages for these 2 methods. Perhaps we can remove them.
Also I'm not sure we need `EntryPointMapVec` at all.

MrSidims:
Could you please uplift spirv-max version to 1.4 for test/transcoding/block_w_struct_return.cl and test/transcoding/global_block.cl tests removing the appropriate TODOs from there? (see also https://github.com/KhronosGroup/SPIRV-LLVM-Translator/pull/1177/files )

MrSidims (commented):
LGTM

lib/SPIRV/SPIRVWriter.cpp
```diff
@@ -3567,13 +3565,18 @@ bool LLVMToSPIRVBase::isAnyFunctionReachableFromFunction(
...
-    if (AS != SPIRAS_Input && AS != SPIRAS_Output)
-      continue;
-
+    SPIRVModule *BM = SF->getModule();
+    if (!BM->isAllowedToUseVersion(VersionNumber::SPIRV_1_4)) {
+      const auto AS = GV.getAddressSpace();
+      if (AS != SPIRAS_Input && AS != SPIRAS_Output)
+        continue;
+    } else {
+      BM->setSPIRVVersion(static_cast<SPIRVWord>(VersionNumber::SPIRV_1_4));
```
AlexeySotkin, line 3578:
```suggestion
      BM->setMinSPIRVVersion(static_cast<SPIRVWord>(VersionNumber::SPIRV_1_4));
```

test/transcoding/PipeStorage.ll
```diff
@@ -11,9 +11,6 @@
 ; CHECK-LLVM-DAG: [[CREATOR_NAME:[^ ]+]] = linkonce_odr addrspace(1) global %spirv.ConstantPipeStorage { i32 16, i32 16, i32 1 }, align 4
 ; CHECK-LLVM-DAG: @mygpipe = addrspace(1) global %"[[CL_PIPE_STORAGE_NAME]]" { %spirv.PipeStorage addrspace(1)* bitcast (%spirv.ConstantPipeStorage addrspace(1)* [[CREATOR_NAME]] to %spirv.PipeStorage addrspace(1)*) }, align 4
 
-; check for magic number followed by version 1.1
```
AlexeySotkin, line 14: Why your changes affected this test?

KornevNikita: Because I'm setting spirv1.4 version (if translator allowed to use it) if there are global variables with a storage class != input / output. So this test fails as it expexts version 1.1.

test/transcoding/block_w_struct_return.cl
```diff
@@ -1,17 +1,7 @@
 // RUN: %clang_cc1 -triple spir -cl-std=cl2.0 -disable-llvm-passes -fdeclare-opencl-builtins -finclude-default-header %s -emit-llvm-bc -o %t.bc
-// TODO: currently max version is limited to 1.1 for this test. Issues here
```
AlexeySotkin, line 2: Can we modify this test in a way that it will check both cases when max version is limited to 1.1 and when it is not limited at all?

test/transcoding/global_block.cl
```diff
@@ -4,18 +4,8 @@
 // removed
 
 // RUN: %clang_cc1 -O0 -triple spir-unknown-unknown -cl-std=CL2.0 -x cl %s -emit-llvm-bc -o %t.bc
-// TODO: currently max version is limited to 1.1 for this test. Issues here
```
AlexeySotkin, line 7: an we modify this test in a way that it will check both cases when max version is limited to 1.1 and when it is not limited at all?

lib/SPIRV/SPIRVWriter.cpp
AlexeySotkin: It seems like by default(unless `--spirv-max-version` option is used) we will generate SPIR-V version 1.4 every time there is a global variable used by a kernel, i.e. almost always. In case there are concerns about that, probably it makes sense to create a separate PR which will rise the version based on features in the module(not only for this case).

KornevNikita: We can rise version if translator allowed to use v1.4 & interface variables are not of input/output storage class. It should reduce the number of such cases.

KornevNikita: 0462754e8b1fe484a29ce21bf0d622eda7bc6460

KornevNikita: Applied in two last commits

test/transcoding/global_block.cl
```diff
@@ -4,27 +4,25 @@
...
+// RUN: llvm-spirv -r %t.spirv1.4.spv -o - | llvm-dis | FileCheck %s --check-prefix=CHECK-LLVM
 
 kernel void block_kernel(__global int* res) {
   typedef int (^block_t)(int);
   constant block_t b1 = ^(int i) { return i + 1; };
   *res = b1(5);
 }
 
+// CHECK-SPIRV1_4: EntryPoint 6 [[#]] "block_kernel" [[#InterdaceId:]]
+// CHECK-SPIRV1_4: Name [[#InterdaceId]] "__block_literal_global"
```
AlexeySotkin, line 25:
```suggestion
// CHECK-SPIRV1_4: EntryPoint 6 [[#]] "block_kernel" [[#InterfaceId:]]
// CHECK-SPIRV1_4: Name [[#InterfaceId]] "__block_literal_global"
```

AlexeySotkin (approved):
Other than typo in tests, LGTM.

PR #1335, opened 2021-12-13
Translate complex nested vector expressions instead of lowering them

vmaksimo:
Allow the translator to go deeper into complex nested instructions.
Enabled a possibility to translate expressions in `extractelement` and binary operator instructions (e.g., `fadd`, `fmul`).
Removed lowering of constant expression vectors as these two methods are mutually exclusive.

This patch was developed as an alternative to this fix: #1330


vmaksimo:
Tagging @AlexeySotkin @MrSidims @pauzinl @KornevNikita to take a look.

lib/SPIRV/libSPIRV/SPIRVInstruction.h
```diff
@@ -2757,11 +2757,28 @@ _SPIRV_OP(ImageQuerySamples, true, 4)
...
+  std::vector<SPIRVEntry *> getNonLiteralOperands() const override {
+    // If SpecConstant results from CompositeExtract/Insert operation, then all
+    // operands are expected to be literals.
+    SPIRVWord LiteralOpcode = Ops[0];
+    if (LiteralOpcode == OpCompositeExtract ||
+        LiteralOpcode == OpCompositeInsert)
+      return {};
+
+    return SPIRVInstTemplateBase::getNonLiteralOperands();
+  }
```
AlexeySotkin, line 2775: Still, we may call `isOperandLiteral()` which will return wrong answer. I suggest overriding `isOperandLiteral()` instead
```suggestion
  bool isOperandLiteral(unsigned I) const override {
    switch (Ops[0]) { // Opcode
    case OpCompositeExtract:
    case OpCompositeInsert:
      return true;
    default:
      return SPIRVInstTemplateBase::isOperandLiteral(I); // Will it always return false?
    }
  }
```

vmaksimo: Nice suggestion, thanks! Not sure if `SPIRVInstTemplateBase::isOperandLiteral(I)` will always return `false`, but it looks right to call the base class method to be safe for sure.

AlexeySotkin:
@vmaksimo could you please create a PR with this patch at intel/llvm repo and run tests there to make sure this patch will not bring any regressions when we integrate it to the SYCL compiler?

AlexeySotkin (approved):

AlexeySotkin:
@zuban32 FYI.

PR #1344, opened 2021-12-16
Preserve DISubrange's upperBound

KornevNikita:
DISubrange must contain count or upperBound. Prior to this patch
translator preserved only DISubrange's count, so if there is upperBound
insted of count, the reverse translation will fail, as resulting
DISubrange will contain nothing.


KornevNikita:
@AlexeySotkin @MrSidims @mlychkov @AlexeySachkov take a look please

AlexeySotkin (approved):

PR #1346, opened 2021-12-20
Implements spirv.Decorations and spirv.ParameterDecorations metadata translation

steffenlarsen:
This patch adds translation to and from `spirv.Decorations` and `spirv.ParameterDecorations` metadata. These metadata nodes represent global variable and function parameter decorations respectively, allowing explicit decoration in LLVM IR.

Documentation: https://github.com/KhronosGroup/SPIRV-LLVM-Translator/pull/1319

### DIFF
--- a/lib/SPIRV/SPIRVReader.cpp
+++ b/lib/SPIRV/SPIRVReader.cpp
@@ -160,7 +160,7 @@ static void addNamedMetadataStringSet(LLVMContext *Context, Module *M,
   NamedMD->addOperand(MDNode::get(*Context, ValueVec));
 }
 
-static void addOCLKernelArgumentMetadata(
+static void addKernelArgumentMetadata(
     LLVMContext *Context, const std::string &MDName, SPIRVFunction *BF,
     llvm::Function *Fn,
     std::function<Metadata *(SPIRVFunctionParameter *)> ForeachFnArg) {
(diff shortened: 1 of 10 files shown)


lib/SPIRV/SPIRVReader.cpp
```diff
@@ -3555,12 +3555,79 @@ void SPIRVToLLVM::transGlobalAnnotations() {
...
+}
+
+void SPIRVToLLVM::transVarDecorationsToMetadata(SPIRVValue *BV, Value *V) {
+  if (!BV->isVariable())
+    return;
+
+  if (auto *GV = dyn_cast<GlobalVariable>(V)) {
+    std::vector<SPIRVDecorate const *> Decorates = BV->getDecorations();
+    if (!Decorates.empty()) {
+      MDNode *MDList = transDecorationsToMetadataList(Context, Decorates);
```
MrSidims, line 3618: It does look like, that in the end after reverse translation we will have both decoration metadata and it's previously defined representation in LLVM IR module, correct? If so, it kind of looking incorrect, since SPIR-V friendly LLVM IR should be only emitted under SPV-IR flag (and decoration metadata is SPIR-V friendly form). But at the same time, I'm 100% sure, that none of SPIR-V centric LLVM IR device backends are ready to switch from usual representation to decoration metadata representation, so I believe, we should leave this code as is, but with some comment added, WDYT?

test/spirv_param_decorations.ll
```diff
@@ -0,0 +1,50 @@
...
+; RUN: llvm-spirv %t.bc -spirv-text -o - | FileCheck %s --check-prefix=CHECK-SPIRV
+; RUN: llvm-spirv %t.bc -o %t.spv
+; RUN: llvm-spirv -r %t.spv -o %t.rev.bc
+; RUN: llvm-dis < %t.rev.bc | FileCheck %s --check-prefix=CHECK-LLVM
+
+target datalayout = "e-p:32:32-i64:64-v16:16-v24:32-v32:32-v48:64-v96:128-v192:256-v256:256-v512:512-v1024:1024"
+target triple = "spir"
+
+; Function Attrs: convergent nounwind
+define spir_kernel void @k(float %a, float %b, float %c) #0 !kernel_arg_addr_space !4 !kernel_arg_access_qual !5 !kernel_arg_type !6 !kernel_arg_type_qual !7 !kernel_arg_base_type !6 !spirv.ParameterDecorations !14 {
```
MrSidims, line 11: Could you please a test case, when we have both decoration metadata attached to the kernel and restrict/volatile specifiers on the kernel arguments to check if the translation can handle interfering representations. This comment is applicable to another test.

lib/SPIRV/SPIRVReader.cpp
MrSidims: If I'm correct about duality of LLVM IR output, could you please reflect it in the tests?

MrSidims (commented):
Overall LGTM, thanks! I added some thoughts about reverse translation part and test suggestions.

test/spirv_param_decorations.ll
steffenlarsen: I have added a test test/spirv_param_decorations.ll that tests translation where `!kernel_arg_type_quals` specifies an argument to be volatile and `!spirv.ParameterDecorations` specifies noalias on the same argument. Note that `restrict` in `!kernel_arg_type_quals` infers noalias and not Restrict decoration. The test shows that the qualifiers are merged, i.e. the resulting `!kernel_arg_type_quals` has `!"volatile restrict"` and the re-translated decorator metadata list has both noalias and volatile decorator nodes. Is this in line with what you were thinking?

lib/SPIRV/SPIRVReader.cpp
steffenlarsen: You're absolutely right! It didn't seem like a lot of extra work to only generate the metadata if in SPIR-V friendly mode so I went with that instead. The tests have also been changed to check that the metadata is only there if SPV-IR mode is set. Is this okay or would you prefer I revert it and just add a comment?

test/spirv_param_decorations.ll
MrSidims: Yes, thanks!

lib/SPIRV/SPIRVReader.cpp
MrSidims: I'd gather opinions from @aratajew @svenvh and @AlexeySotkin 

Folks, like this in SPIR-V friendly IR after the translation !spirv.(Param)Decoration will appear and will replace previous representation of some features, for example kernel_arg_type_quals carrying volatile/restrict qualifiers will replaced with the mentioned SPIR-V friendly form. Even if I believe, that it's the correct way to go, I'm afraid, that right now it might break gathering of Arg type qualifiers in OpenCL runtime, since !spirv.(Param)Decoration metadata aren't supported in any devices' compilers yet. WDYT?

aratajew: In my opinion, `!spirv.(Param)Decoration` metadata representation should be used only when a user triggers translation with `--spirv-target-env=SPV-IR` parameter. When `--spirv-target-env=CL2.0` or `--spirv-target-env=CL1.2` is specified, then `SPIRVToOCL` pass should translate `!spirv.(Param)Decoration` to OpenCL representation (`kernel_arg_type_quals` etc.). As I can see, this PR doesn't include any changes in `SPIRVToOCL` pass, so I agree with @MrSidims that this change will break  OpenCL backends that are based `CL2.0` or `CL1.2` translation.

Also, just out of curiousity, @steffenlarsen what is the main motivation to use `!spirv.(Param)Decoration` for SPIR-V Friendly IR path instead of OpenCL representation? Is it because not all SPIRV decorations have their's corresponding OpenCL metadata?

aratajew: > In my opinion, !spirv.(Param)Decoration metadata representation should be used only when a user triggers translation with --spirv-target-env=SPV-IR parameter. When --spirv-target-env=CL2.0 or --spirv-target-env=CL1.2 is specified, then SPIRVToOCL pass should translate !spirv.(Param)Decoration to OpenCL representation (kernel_arg_type_quals etc.). As I can see, this PR doesn't include any changes in SPIRVToOCL pass, so I agree with @MrSidims that this change will break OpenCL backends that are based CL2.0 or CL1.2 translation.

I see that I misunderstood the concern. This change properly produces metadata based on `--spirv-target`, but the concern is whether OpenCL Runtimes based on `SPV-IR` translation path will be able to consume new metadata representation. In case of Intel OpenCL GPU backend, the compiler is in the middle of transition from internal SPIRVReader to Khronos one, so there will be a need to implement a support for new metadata.

steffenlarsen: > Also, just out of curiousity, @steffenlarsen what is the main motivation to use !spirv.(Param)Decoration for SPIR-V Friendly IR path instead of OpenCL representation? Is it because not all SPIRV decorations have their's corresponding OpenCL metadata?

Yes, that is definitely a driving factor. `!spirv.(Param)Decoration` offers more freedom at the cost of making it more SPIR-V specific. In principle any implementation could implement some extension decorator that the translator does not yet support and still be able to produce SPIR-V using that decorator by generating `!spirv.(Param)Decoration`, assuming the parameter space is simple. Additionally, I personally think there is value to the generality of `!spirv.(Param)Decoration`, i.e. the parsing of the metadata trees are overlapping for the two current variants and future versions would likely have the same structure.

steffenlarsen:
@MrSidims @aratajew @svenvh @AlexeySotkin - Friendly ping. Are there any outstanding concerns or comments for this?

lib/SPIRV/SPIRVReader.cpp
```diff
@@ -3627,7 +3627,10 @@ bool SPIRVToLLVM::transDecoration(SPIRVValue *BV, Value *V) {
 
   transIntelFPGADecorations(BV, V);
   transMemAliasingINTELDecorations(BV, V);
-  transVarDecorationsToMetadata(BV, V);
+
+  // Decoration metadata is only enabled in SPIR-V friendly mode
+  if (BM->getDesiredBIsRepresentation() == BIsRepresentation::SPIRVFriendlyIR)
+    transVarDecorationsToMetadata(BV, V);
```
MrSidims, line 3633: This 'if' might be  removed in the future, making the Decoration metadata be emitted by default and be transformed to OpenCL representation in SPIRVToOCL pass, but lets first see new features being developed atop of this infrastructure.

MrSidims (approved):

svenvh (approved):
LGTM, no concerns.

MrSidims:
Something went wrong after merging this patch:
Failed Tests (1):
  LLVM_SPIRV :: spirv_param_decorations_quals.ll
taking a look

PR #1347, opened 2021-12-21
Improved constant expression translation

DmitryBushev:
Added support for binary operators in constant expression translation.


DmitryBushev:
Actually this was already done by #1335. Closing

PR #1350, opened 2021-12-21
Support DISubrange DIExpression upperBound

KornevNikita:


KornevNikita:
I don't know if I should add the test for this case `!DISubrange(lowerBound: ..., upperBound: !DIExpression(...)` as there is FortranArray.ll, but it is turned off. @AlexeySotkin @MrSidims what do you think?

test/FortranArray.ll
```diff
@@ -10,7 +10,7 @@
 ; and back to .ll. This causes the LLVM IR verifier to fail as there
 ; are different rules for valid DISubRange depending on language ID.
 
-; CHECK-LLVM: !DISubrange(lowerBound: 1)
+; CHECK-LLVM: !DISubrange(lowerBound: 1, upperBound: !DIExpression(DW_OP_push_object_address, DW_OP_plus_uconst, 64, DW_OP_deref, DW_OP_push_object_address, DW_OP_plus_uconst, 48, DW_OP_deref, DW_OP_plus, DW_OP_constu, 1, DW_OP_minus))
```
AlexeySotkin, line 13: Can we preserve the expression for the lower bound too?

AlexeySotkin:
> I don't know if I should add the test for this case `!DISubrange(lowerBound: ..., upperBound: !DIExpression(...)` as there is FortranArray.ll, but it is turned off. @AlexeySotkin @MrSidims what do you think? I checked that the test is passing if we revert #1074

If FortranArray.ll passes now, I think we should turn it on (remove XFAIL).

test/FortranArray.ll
KornevNikita: Now - yes: 8d7e19c

KornevNikita:
@AlexeySotkin FortranArray.ll doesn't pass now, because it is necessary to revert #1074.
But I've added the new case `!DISubrange(lowerBound: !DIExpression(), upperBound: !DIExpression())` to make sure that everything works.

AlexeySotkin (approved):

AlexeySotkin:
LGTM, Thanks!
I think we can commit it as is.

PR #1357, opened 2021-12-22
Fix SPIRV Friendly IR for OpGenericPtrMemSemantics

aratajew:


MrSidims (approved):
LGTM, thanks!

BTW, how far we want to go fixing SPIR-V friendly IR generating? My guess is only instructions with OpenCL counterparts are affected, while we don't want to touch something like OpLoad/OpStore.

aratajew:
> LGTM, thanks!
> 
> BTW, how far we want to go fixing SPIR-V friendly IR generating? My guess is only instructions with OpenCL counterparts are affected, while we don't want to touch something like OpLoad/OpStore.

Yes, my intention is to fix all builtins that are translated to OpenCL representation instead of SPIRV Friendly IR. I don't want to touch OpLoad/OpStore etc., since these may be represented via llvm instructions.

PR #1374, opened 2022-01-10
Fix DISubrange translation

KornevNikita:
If there is count field in DISubrange, then lowerBound may not be equal
to zero.


AlexeySotkin (approved):

KornevNikita:
Retrigger checks

PR #1385, opened 2022-01-26
Handle translattion of DIArgList with single argument

AlexeySotkin:
In the case there are more than one argument we can't represent it with
the current SPIR-V debug info spec, therefore we drop DIArgList and
DIExpression.

### DIFF
--- a/lib/SPIRV/LLVMToSPIRVDbgTran.cpp
+++ b/lib/SPIRV/LLVMToSPIRVDbgTran.cpp
@@ -150,12 +150,16 @@ void LLVMToSPIRVDbgTran::finalizeDebugValue(
   SPIRVExtInst *DV = static_cast<SPIRVExtInst *>(V);
   SPIRVBasicBlock *BB = DV->getBasicBlock();
   Value *Val = DbgValue->getVariableLocationOp(0);
-
+  DIExpression *Expr = DbgValue->getExpression();
+  if (DbgValue->getNumVariableLocationOps() > 1) {
+    Val = UndefValue::get(Val->getType());
+    Expr = DIExpression::get(M->getContext(), {});
+  }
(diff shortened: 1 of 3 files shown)


AlexeySotkin:
@vmaksimo @KornevNikita please take a look.

MrSidims (approved):
LGTM

MrSidims:
Test failure should be fixed by https://github.com/KhronosGroup/SPIRV-LLVM-Translator/pull/1388

vmaksimo:
Suggestion for commit message: `In case there is more than one argument that we can't represent with the current SPIR-V debug info spec, we drop DIArgList and DIExpression.`

vmaksimo:
The implementation itself LGTM.
But it's not obvious to me why we drop `DIArgList` only if it has more than 1 arg. Why not drop it even there is only one unsupported argument?

KornevNikita (approved):
LGTM

AlexeySotkin:
> The implementation itself LGTM. But it's not obvious to me why we drop `DIArgList` only if it has more than 1 arg. Why not drop it even there is only one unsupported argument?

If there is one argument in the list we can preserve the debug info. Why should we drop it?

vmaksimo:
> > The implementation itself LGTM. But it's not obvious to me why we drop `DIArgList` only if it has more than 1 arg. Why not drop it even there is only one unsupported argument?
> 
> If there is one argument in the list we can preserve the debug info. Why should we drop it?

Sorry, just according to the modified test I thought that we actually do support `DIArgList` as the translator doesn't crash.
Now the purpose is clear to me - we "support" `DIArgList" only syntactically but can't properly handle the list of debug arguments for now.

vmaksimo (approved):

PR #1386, opened 2022-01-26
Update for PointerType::getElementType deprecation

svenvh:
Update for LLVM commit 184591aeeb5a ("[OpaquePtrs] Deprecate
PointerType::getElementType()", 2022-01-25).

This commit changes all affected calls to getPointerElementType, which
is a temporary solution.  Followup work around the
getPointerElementType calls will be required to align with LLVM's
opaque pointer changes.


MrSidims (approved):

AlexeySotkin (approved):
LGTM, Thanks @svenvh!

PR #1388, opened 2022-01-27
[NFC] Fix CHECK strings in the test

MrSidims:


MrSidims:
@steffenlarsen please take a look. For some reasons pre-commit didn't catch test issues.

steffenlarsen (approved):
My best guess is that there may have been a change to when `noalias` is added to parameters. For the sake of this test it is more important that `kernel_arg_type_qual` has the right qualifiers, so this should be fine.

PR #1389, opened 2022-01-27
Implement SPV_INTEL_global_variable_decorations extension

vmaksimo:
This change adds support of decorations applied to global variables through
`spirv.Decorations` metadata in LLVM IR.

These decorations are mostly intended to help code generation for FPGA devices:
* HostAccessINTEL
* InitModeINTEL
* ImplementInCSRINTEL

Spec:
https://github.com/intel/llvm/blob/sycl/sycl/doc/extensions/DeviceGlobal/SPV_INTEL_global_variable_decorations.asciidoc


vmaksimo:
Tagging @AlexeySotkin @MrSidims @KornevNikita @pauzinl to take a look

KornevNikita (approved):
Looks good for me

lib/SPIRV/SPIRVWriter.cpp
```diff
@@ -2151,6 +2153,20 @@ static void transMetadataDecorations(Metadata *MD, SPIRVEntry *Target) {
           Target, Name->getString().str(), TypeKind));
       break;
     }
+    case spv::internal::DecorationHostAccessINTEL: {
```
MrSidims, line 2156: According to spec: "Only valid on global (module scope) OpVariable". It seems like there is no appropriate check for this.

lib/SPIRV/libSPIRV/SPIRVDecorate.cpp
```diff
@@ -114,6 +114,9 @@ void SPIRVDecorate::encode(spv_ostream &O) const {
   case DecorationUserSemantic:
     SPIRVDecorateUserSemanticAttr::encodeLiterals(Encoder, Literals);
     break;
+  case spv::internal::DecorationHostAccessINTEL:
```
MrSidims, line 117: Seems like there should be
`switch (static_cast<uint32_t>(Dec)) {`

lib/SPIRV/SPIRVWriter.cpp
vmaksimo: I thought that `transMetadataDecorations()` is only called for GlobalVariable: https://github.com/KhronosGroup/SPIRV-LLVM-Translator/blob/a96f8da5942d175645b26f822929ea995fc31f26/lib/SPIRV/SPIRVWriter.cpp#L2268-L2271
But I found that it also can be called for kernel parameter decorations. 
Thanks for noticing, I'll add the appropriate check.

lib/SPIRV/SPIRVWriter.cpp
```diff
@@ -2151,6 +2153,20 @@ static void transMetadataDecorations(Metadata *MD, SPIRVEntry *Target) {
...
+      assert(NumOperands == 3 &&
+             "HostAccessINTEL requires exactly 3 extra operand");
```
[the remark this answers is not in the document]
vmaksimo: Yeah, here's a bit incorrect naming/numbering of operands - it actually means that we expect metadata to contain 3 operands like `!{i32 6147, i32 0, !{"Name"}}`, where the decoration itself contains exactly 2 operands.
It was done alike to DecorationLinkageAttributes handling (see above).
I suppose the warning in the assert should be changed to avoid confusion.

lib/SPIRV/libSPIRV/SPIRVDecorate.cpp
vmaksimo: Thanks! Found that the same fix was applied for `SPIRVDecorate::getRequiredExtension()`, will apply that.

test/transcoding/SPV_INTEL_global_variable_decorations/global_var_decorations.ll
```diff
@@ -0,0 +1,73 @@
...
+; CHECK-SPV-IR: ![[#MD_CSR_0]] = !{i32 6149, i32 0}
+
+
+; CHECK-LLVM-NOT: @int_var = {{.*}}, !spirv.Decorations ![[#]]
+; CHECK-LLVM-NOT: @float_var = {{.*}}, !spirv.Decorations ![[#]]
+; CHECK-LLVM-NOT: @bool_var = {{.*}}, !spirv.Decorations ![[#]]
+
+; CHECK-LLVM: @int_var = addrspace(1) global i32 42
+; CHECK-LLVM: @float_var = addrspace(1) global float 1.000000e+00
+; CHECK-LLVM: @bool_var = addrspace(1) global i1 false
```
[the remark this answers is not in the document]
vmaksimo: Now it would be the same as for default behavior (IR as in CHECK-LLVM).

What is expected for the output of `SPIRVToOCL` pass? Should we just drop the decorations or to transform them to global variable attributes like `attributes #0 = {"init_mode"="0" "host_access"="1"}`? Or maybe something else?

gmlueck:
FYI to @steffenlarsen since you were working in this area of the SPIRV-LLVM translator also.

test/transcoding/SPV_INTEL_global_variable_decorations/global_var_decorations.ll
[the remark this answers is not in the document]
AlexeySotkin: IMHO, I'd expect we do drop the metadata in OpenCL IR mode _until_ we have OpenCL extension for this feature. When/if we will have  the OpenCL extension, we will have it's representation in OpenCL C and clang will translate it into something in LLVM IR. And that LLVM IR representation, I think, should be generated as output of `SPIRVToOCL` pass.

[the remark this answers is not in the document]
AlexeySotkin: It looks like we have [one](https://github.com/intel/llvm/tree/sycl/sycl/doc/extensions/DeviceGlobal/cl_intel_global_variable_access.asciidoc). However it doesn't define OpenCL C representation. Probably it is not final draft.

MrSidims (approved):
LGTM

lib/SPIRV/SPIRVWriter.cpp
```diff
@@ -2151,6 +2164,52 @@ static void transMetadataDecorations(Metadata *MD, SPIRVEntry *Target) {
...
+          Target, AccessMode->getZExtValue(), Name->getString().str()));
+      break;
+    }
+    case spv::internal::DecorationInitModeINTEL: {
+      checkIsGlobalVar(Target, DecoKind);
+      assert(static_cast<SPIRVVariable *>(Target)->getInitializer() &&
+             "InitModeINTEL only be applied to a global (module scope) "
+             "variable which has an Initializer operand");
+
+      assert(NumOperands == 2 &&
```
AlexeySotkin, line 2190:
```suggestion
      assert(NumOperands == 1 &&
```

AlexeySotkin (approved):

PR #1390, opened 2022-01-28
Implement forward translation for llvm.fence

pauzinl:
This patch adds forward translation for fence instruction.
Reverse translation uses OCL builtin for OpMemoryBarrier.

Signed-off-by: Leonid Pauzin leonid.pauzin@intel.com

### DIFF
--- a/lib/SPIRV/SPIRVWriter.cpp
+++ b/lib/SPIRV/SPIRVWriter.cpp
@@ -1968,6 +1968,11 @@ LLVMToSPIRVBase::transValueWithoutDecoration(Value *V, SPIRVBasicBlock *BB,
     return BV ? mapValue(V, BV) : nullptr;
   }
 
+  if (FenceInst *FI = dyn_cast<FenceInst>(V)) {
+    SPIRVValue *BV = transFenceInst(FI, BB);
+    return BV ? mapValue(V, BV) : nullptr;
+  }
+
   if (InlineAsm *IA = dyn_cast<InlineAsm>(V))
     if (BM->isAllowedToUseExtension(ExtensionID::SPV_INTEL_inline_assembly))
       return mapValue(V, transAsmINTEL(IA));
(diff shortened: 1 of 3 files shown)


pauzinl:
@MrSidims @AlexeySotkin take a look please

lib/SPIRV/SPIRVWriter.cpp
```diff
@@ -3460,6 +3465,37 @@ SPIRVValue *LLVMToSPIRVBase::transIntrinsicInst(IntrinsicInst *II,
   return nullptr;
 }
 
+SPIRVValue *LLVMToSPIRVBase::transFenceInst(FenceInst *FI,
+                                            SPIRVBasicBlock *BB) {
+  SPIRVWord MemorySemantics = static_cast<SPIRVWord>(FI->getOrdering());
+  // Fence ordering may only be Acquire, Release, AcquireRelease, or
+  // SequentiallyConsistent
+  switch (MemorySemantics) {
+  case 4:
```
MrSidims, line 3474: Guess, Ordering starts with '3', assuming NotAtomic is '0' (at least https://llvm.org/doxygen/namespacellvm.html#a9bccbe67aaab722783ca4e7c504aaaa7 makes me thinking so)

pauzinl: https://llvm.org/doxygen/AtomicOrdering_8h_source.html#l00056 I guess 3 should be Consume and it is "Not specified yet"

MrSidims (approved):
LGTM
@svenvh could you please take a look at this PR as well?

lib/SPIRV/SPIRVWriter.cpp
```diff
@@ -3460,6 +3465,37 @@ SPIRVValue *LLVMToSPIRVBase::transIntrinsicInst(IntrinsicInst *II,
   return nullptr;
 }
 
+SPIRVValue *LLVMToSPIRVBase::transFenceInst(FenceInst *FI,
+                                            SPIRVBasicBlock *BB) {
+  SPIRVWord MemorySemantics = static_cast<SPIRVWord>(FI->getOrdering());
+  // Fence ordering may only be Acquire, Release, AcquireRelease, or
+  // SequentiallyConsistent
+  switch (MemorySemantics) {
```
svenvh, line 3473: Maybe switch based on `FI->getOrdering()` so that we can use the original `AtomicOrdering` enum value names in the `case` statements?

lib/SPIRV/SPIRVWriter.cpp
```diff
@@ -3460,6 +3465,37 @@ SPIRVValue *LLVMToSPIRVBase::transIntrinsicInst(IntrinsicInst *II,
...
+    MemorySemantics = 0x10;
+    break;
+  default:
+    assert(false && "Unexpected fence ordering");
+    MemorySemantics = SPIRVWORD_MAX;
+    break;
+  }
+
+  Module *M = FI->getParent()->getModule();
+  SPIRVValue *RetScope = transConstant(getUInt32(M, 0));
```
svenvh, line 3493: Should we call `FI->getSyncScopeID()` and map the return value to a suitable SPIR-V Scope?

lib/SPIRV/SPIRVWriter.cpp
```diff
@@ -3460,6 +3465,37 @@ SPIRVValue *LLVMToSPIRVBase::transIntrinsicInst(IntrinsicInst *II,
...
 }
 
+SPIRVValue *LLVMToSPIRVBase::transFenceInst(FenceInst *FI,
+                                            SPIRVBasicBlock *BB) {
+  SPIRVWord MemorySemantics = static_cast<SPIRVWord>(FI->getOrdering());
+  // Fence ordering may only be Acquire, Release, AcquireRelease, or
+  // SequentiallyConsistent
+  switch (MemorySemantics) {
+  case 4:
+    MemorySemantics = 0x2;
```
svenvh, line 3475:
```suggestion
    MemorySemantics = MemorySemanticsAcquireMask;
```

Same for the hardcoded constants below.

pauzinl: Thanks! Changed in [4ff8796](https://github.com/KhronosGroup/SPIRV-LLVM-Translator/pull/1390/commits/4ff879629a5bba97459cd450964789570ef5e5b3)

lib/SPIRV/SPIRVWriter.cpp
pauzinl: Probably I am wrong, but llvm.fence instruction doesn't contain information about SPIR-V Scope (workgroup/subgroup/etc).
SyncScope can only be SingleThread or System https://llvm.org/doxygen/LLVMContext_8h_source.html#l00046

svenvh: Yeah I'm not very confident about a suitable mapping either actually...  But we should make it more obvious that we're assuming `ScopeCrossDevice` then:
```suggestion
  // Treat all llvm.fence instructions as having CrossDevice scope:
  SPIRVValue *RetScope = transConstant(getUInt32(M, ScopeCrossDevice));
```

svenvh (approved):

PR #1392, opened 2022-02-01
Added option to choose program address space in emitted LLVM IR

DmitryBushev:
SPIRV does not specify storage class for functions, although llvm does.
This patch enables choise of address space for function in data layout
of emitted LLVM module.

### DIFF
--- a/lib/SPIRV/SPIRVReader.cpp
+++ b/lib/SPIRV/SPIRVReader.cpp
@@ -355,6 +355,18 @@ std::string SPIRVToLLVM::transVCTypeName(SPIRVTypeBufferSurfaceINTEL *PST) {
   return VectorComputeUtil::getVCBufferSurfaceName();
 }
 
+static unsigned transPointerStorageClass(SPIRVTypePointer const *PtrType) {
+  assert(PtrType && "Must be valid pointer type.");
+
+  if (PtrType->getPointerElementType()->getOpCode() == OpTypeFunction) {
+    auto *M = PtrType->getModule();
+    if (M->hasProgramAddressSpace())
+      return M->getProgramAddressSpace();
+  }
+
+  return SPIRSPIRVAddrSpaceMap::rmap(PtrType->getPointerStorageClass());
+}
+
 Type *SPIRVToLLVM::transType(SPIRVType *T, bool IsClassMember) {
   auto Loc = TypeMap.find(T);
   if (Loc != TypeMap.end())
@@ -385,7 +397,7 @@ Type *SPIRVToLLVM::transType(SPIRVType *T, bool IsClassMember) {
     return mapType(
         T, PointerType::get(
                transType(T->getPointerElementType(), IsClassMember),
-               SPIRSPIRVAddrSpaceMap::rmap(T->getPointerStorageClass())));
+               transPointerStorageClass(static_cast<SPIRVTypePointer *>(T))));
   case OpTypeVector:
     return mapType(T,
                    FixedVectorType::get(transType(T->getVectorComponentType()),
@@ -3244,15 +3256,23 @@ bool SPIRVToLLVM::translate() {
   return true;
 }
 
+static std::string transDataLayout(SPIRVModule *M,
+                                   std::string const &GeneralLayout) {
+  return M->hasProgramAddressSpace()
+             ? GeneralLayout + "-P" +
+                   std::to_string(M->getProgramAddressSpace())
+             : GeneralLayout;
+}
+
 bool SPIRVToLLVM::transAddressingModel() {
   switch (BM->getAddressingModel()) {
   case AddressingModelPhysical64:
     M->setTargetTriple(SPIR_TARGETTRIPLE64);
-    M->setDataLayout(SPIR_DATALAYOUT64);
+    M->setDataLayout(transDataLayout(BM, SPIR_DATALAYOUT64));
     break;
   case AddressingModelPhysical32:
     M->setTargetTriple(SPIR_TARGETTRIPLE32);
-    M->setDataLayout(SPIR_DATALAYOUT32);
+    M->setDataLayout(transDataLayout(BM, SPIR_DATALAYOUT32));
     break;
   case AddressingModelLogical:
     // Do not set target triple and data layout
--- a/lib/SPIRV/libSPIRV/SPIRVModule.h
+++ b/lib/SPIRV/libSPIRV/SPIRVModule.h
@@ -488,6 +488,14 @@ public:
     return TranslationOpts.isAllowedToUseExtension(RequestedExtension);
   }
 
+  bool hasProgramAddressSpace() const {
+    return TranslationOpts.hasProgramAddressSpace();
+  }
+
+  unsigned getProgramAddressSpace() const {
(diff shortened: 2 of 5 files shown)


lib/SPIRV/SPIRVReader.cpp
```diff
@@ -3224,11 +3224,17 @@ bool SPIRVToLLVM::transAddressingModel() {
   switch (BM->getAddressingModel()) {
   case AddressingModelPhysical64:
     M->setTargetTriple(SPIR_TARGETTRIPLE64);
-    M->setDataLayout(SPIR_DATALAYOUT64);
+    M->setDataLayout(BM->hasProgramAddressSpace()
+                         ? std::string(SPIR_DATALAYOUT64) + "-P" +
+                               std::to_string(BM->getProgramAddressSpace())
+                         : SPIR_DATALAYOUT64);
```
dmitryryintel, line 3230: It would be nice to put this logic into a separate function that takes `BM` and `SPIR_DATALAYOUT64`. This way you will avoid the copy-paste below.

test/program-addrspace-option.ll
```diff
@@ -0,0 +1,21 @@
...
+; CHECK: target datalayout
+; CHECK-SAME: P3
+
+target datalayout = "e-p:32:32:32-i1:8:8-i8:8:8-i16:16:16-i32:32:32-i64:64:64-f32:32:32-f64:64:64-v16:16:16-v24:32:32-v32:32:32-v48:64:64-v64:64:64-v96:128:128-v128:128:128-v192:256:256-v256:256:256-v512:512:512-v1024:1024:1024"
+target triple = "spir-unknown-unknown"
+
+; CHECK: @test
+; CHECK-SAME: addrspace(3)
+
+define spir_kernel void @test() {
```
dmitryryintel, line 17: Could you also check the case with function pointers? Will they end up in the correct namespace?

lib/SPIRV/libSPIRV/SPIRVModule.h
```diff
@@ -488,6 +488,14 @@ class SPIRVModule {
     return TranslationOpts.isAllowedToUseExtension(RequestedExtension);
   }
 
+  virtual bool hasProgramAddressSpace() const final {
```
dmitryryintel, line 491: Don't get the logic behind virtual final here. Why not just use non-virtual method?

dmitryryintel (changes requested):

lib/SPIRV/libSPIRV/SPIRVModule.h
DmitryBushev: I didn't get the logic either, I just copy-pasted method header and changed name. Did not even notice that first.

test/program-addrspace-option.ll
DmitryBushev: As I've inspected, pointers' address spaces are translated directly into StorageClass and back. So, this information is already preserved by translator

lib/SPIRV/SPIRVReader.cpp
```diff
@@ -354,6 +354,17 @@ std::string SPIRVToLLVM::transVCTypeName(SPIRVTypeBufferSurfaceINTEL *PST) {
     return VectorComputeUtil::getVCBufferSurfaceName(PST->getAccessQualifier());
   return VectorComputeUtil::getVCBufferSurfaceName();
 }
+static unsigned transPointerStorageClass(SPIRVTypePointer const *PtrType) {
```
dmitryryintel, line 357:
```suggestion

static unsigned transPointerStorageClass(SPIRVTypePointer const *PtrType) {
```

lib/SPIRV/SPIRVReader.cpp
```diff
@@ -3220,15 +3231,22 @@ bool SPIRVToLLVM::translate() {
   return true;
 }
 
+std::string SPIRVToLLVM::transDataLayout(std::string const &GeneralLayout) {
+  return BM->hasProgramAddressSpace()
+             ? GeneralLayout + "-P" +
+                   std::to_string(BM->getProgramAddressSpace())
+             : GeneralLayout;
```
dmitryryintel, line 3238:
```suggestion
  if (!BM->hasProgramAddressSpace())
    return GeneralLayout;
  return GeneralLayout + "-P" + std::to_string(BM->getProgramAddressSpace());
```

lib/SPIRV/SPIRVReader.cpp
```diff
@@ -3220,15 +3231,22 @@ bool SPIRVToLLVM::translate() {
   return true;
 }
 
+std::string SPIRVToLLVM::transDataLayout(std::string const &GeneralLayout) {
```
dmitryryintel, line 3234: I would just make it a static function and pass `PM`. Passing `this` is not required but costs the same.

lib/SPIRV/libSPIRV/SPIRVModule.h
dmitryryintel: So why adding something you don't understand. I would `virtual` and `final` here.

dmitryryintel (approved):

lib/SPIRV/SPIRVReader.cpp
```diff
@@ -354,6 +354,17 @@ std::string SPIRVToLLVM::transVCTypeName(SPIRVTypeBufferSurfaceINTEL *PST) {
     return VectorComputeUtil::getVCBufferSurfaceName(PST->getAccessQualifier());
   return VectorComputeUtil::getVCBufferSurfaceName();
 }
+static unsigned transPointerStorageClass(SPIRVTypePointer const *PtrType) {
+  assert(PtrType && "Must be valid pointer type.");
+
+  if (PtrType->getPointerElementType()->getOpCode() == OpTypeFunction) {
+    auto *M = PtrType->getModule();
+    if (M->hasProgramAddressSpace())
```
MrSidims, line 362: It seems to me, that any function in the module will have the set address space, which seems incorrect to me (just consider such artificial example:
```
%ptr = alloca i8
call addrspace(3) void @llvm.lifetime.start.p0i8(i64 1, i8* %ptr)
```
how backend will consider addrspace(3) ?
What if a function had address space before the translation? Not this AS is just being lost (and some BE might do some assumptions regarding what AS to pick), with this option turned on the AS won't do any assumptions, just picking the set AS, is it error-prune?

MrSidims (commented):
Can we instead of making this option modify [nction_pointers.asciidoc](https://github.com/intel/llvm/blob/sycl/sycl/doc/extensions/SPIRV/SPV_INTEL_function_pointers.asciidoc) , introducing storage class optional parameter to OpFunctionPointerINTEL
 ?

lib/SPIRV/SPIRVReader.cpp
dmitryryintel: If the function had addrspace before the translation the current translator will preserve AS on a pointer but will lose it on the function itself, which will result in an error during the compilation. To some extend this options saves us. Though fixing the current function pointer implementation first is a nice idea too. There's CodeSectionINTEL storage class but it is never used and it is not clear what LLVM IR addrspace should it correspond.

MrSidims:
Closing as stale

PR #1395, opened 2022-02-03
Update LLVM version from 14 to 15

svenvh:


AlexeySotkin:
Should we fork `llvm_release_140` _before_ merging this PR?

svenvh:
> Should we fork `llvm_release_140` branch _before_ merging this PR?

Good point, although forking could also be done after merging (just need to be more careful to pick the right fork point in that case).

Do you want me to create the `llvm_release_140` branch?

AlexeySotkin:
> > Should we fork `llvm_release_140` branch _before_ merging this PR?
> 
> Good point, although forking could also be done after merging (just need to be more careful to pick the right fork point in that case).
> 
> Do you want me to create the `llvm_release_140` branch?

I gave some thoughts to it. There is a pros in creating the branch later - we will need to backport less patches. So, I'd prefer to postpone creating the branch until the first LLVM API breakage. Does it make sense?

AlexeySotkin (approved):

svenvh:
> I gave some thoughts to it. There is a pros in creating the branch later - we will need to backport less patches. So, I'd prefer to postpone creating the branch until the first LLVM API breakage. Does it make sense?

That works for me.  We'll have to revert this commit on the new release branch when it's created, but that should be trivial.

PR #1396, opened 2022-02-03
Fix regularization of vector addrspacecasts

zuban32:


zuban32:
Well, I just forgot vectors of pointers are not supported. Beacuse of that this change is pending, no need to review now

zuban32:
@AlexeySotkin @MrSidims since this fixes LLVM regularization only could we merge this if I'd modify the test to run regularization since that has formally nothing to do with SPIR-V spec?

MrSidims:
Closing the PR as it's being added as a part of vector of pointers extension implementation

PR #1399, opened 2022-02-08
[Backport][120] cherry-pick for ConstExpr Binary operators

DmitryBushev:
This is partial cherry-pick of c3c3c68bd0438058b92abd00f0704afe1d7fb63e (https://github.com/KhronosGroup/SPIRV-LLVM-Translator/pull/1335),
intended to enable translation of binary operators inside constant expressions.

This is done due to original commit is too complex and depends on a chain of commits that need to be backported prior to fully backporting entire commit. For a short term, I suppose it is better to partially cherry-pick to unblock some needed functionality. It can be fully backported later on.

### DIFF
--- a/lib/SPIRV/SPIRVWriter.cpp
+++ b/lib/SPIRV/SPIRVWriter.cpp
@@ -888,7 +888,7 @@ SPIRVInstruction *LLVMToSPIRV::transBinaryInst(BinaryOperator *B,
       transBoolOpCode(Op0, OpCodeMap::map(LLVMOC)), transType(B->getType()),
       Op0, transValue(B->getOperand(1), BB), BB);
 
-// BinaryOperator can have no parent if it is handled as an expression inside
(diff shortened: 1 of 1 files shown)


lib/SPIRV/SPIRVWriter.cpp
```diff
@@ -888,7 +888,9 @@ SPIRVInstruction *LLVMToSPIRV::transBinaryInst(BinaryOperator *B,
       transBoolOpCode(Op0, OpCodeMap::map(LLVMOC)), transType(B->getType()),
       Op0, transValue(B->getOperand(1), BB), BB);
 
-  if (isUnfusedMulAdd(B)) {
+// BinaryOperator can have no parent if it is handled as an expression inside
```
AlexeySachkov, line 891:
```suggestion
  // BinaryOperator can have no parent if it is handled as an expression inside
```

PR #1400, opened 2022-02-09
Improve backward compatibility for DISubRange

KornevNikita:
If the SPIR-V file was generated by an older version of the translator, Ops may
not contain the LowerBound


KornevNikita:
@AlexeySotkin take a look please

AlexeySotkin (approved):

lib/SPIRV/SPIRVToLLVMDbgTran.cpp
```diff
@@ -225,9 +225,18 @@ SPIRVToLLVMDbgTran::transTypeArray(const SPIRVExtInst *DebugInst) {
     if (!getDbgInst<SPIRVDebug::DebugInfoNone>(Ops[I])) {
```
mlychkov, line 225: May we use `if .. else if ..` instead of `continue` at the end of each if block?

mlychkov, line 225: It also would be nice to have an example of expected TypeArray representation in SPIR-V for each if block.

KornevNikita: Aren't they represented here? https://github.com/KhronosGroup/SPIRV-LLVM-Translator/blob/master/test/DebugInfo/DebugInfoSubrange.ll

KornevNikita: I am afraid that in the future there may be too many different "else-if" blocks, it will be difficult to fit into 80 characters

mlychkov: Ok, this is just my opinion, but I think that a code is less error prone and easier to understand when `if .. else if` is used.
Fitting into 80 characters is not so important, IMHO.

mlychkov: Yes, thank you. But as I see DIExpression case is represented only for empty expression. It would also be nice to have non-empty expression case.

PR #1401, opened 2022-02-09
Restore asserts accidentally removed in 7452422

vmaksimo:
The original change (74759ef) introduced assertions in case `dyn_cast` returns `nullptr`.
Restore them back after being removed in 7452422.

https://github.com/KhronosGroup/SPIRV-LLVM-Translator/commit/74759ef334182a32fddf1defa6f3830a87ab1f2b#diff-f8b80fcd2438a87e6b6ec7219f0cd5240975c6eea3a111afa0a4e9332a01f55dR355-R356

### DIFF
--- a/lib/SPIRV/SPIRVRegularizeLLVM.cpp
+++ b/lib/SPIRV/SPIRVRegularizeLLVM.cpp
@@ -447,9 +447,13 @@ void SPIRVRegularizeLLVMBase::adaptStructTypes(StructType *ST) {
            "Expected a pointer to an array to represent joint matrix type");
     size_t TypeLayout[4] = {0, 0, 0, 0};
     ArrayType *ArrayTy = dyn_cast<ArrayType>(PtrTy->getPointerElementType());
+    assert(ArrayTy && "Expected a pointer element type of an array type to "
+                      "represent joint matrix type");
     TypeLayout[0] = ArrayTy->getNumElements();
     for (size_t I = 1; I != 4; ++I) {
       ArrayTy = dyn_cast<ArrayType>(ArrayTy->getElementType());
+      assert(ArrayTy &&
+             "Expected a element type to represent joint matrix type");
       TypeLayout[I] = ArrayTy->getNumElements();
     }
 


vmaksimo:
Tagging @svenvh @MrSidims @AlexeySotkin

svenvh (approved):
Sorry, I forgot to add the asserts back after rewriting the lambda.  Thanks for fixing!

AlexeySotkin (approved):

lib/SPIRV/SPIRVRegularizeLLVM.cpp
```diff
@@ -447,9 +447,13 @@ void SPIRVRegularizeLLVMBase::adaptStructTypes(StructType *ST) {
            "Expected a pointer to an array to represent joint matrix type");
     size_t TypeLayout[4] = {0, 0, 0, 0};
     ArrayType *ArrayTy = dyn_cast<ArrayType>(PtrTy->getPointerElementType());
+    assert(ArrayTy && "Expected a pointer element type of an array type to "
+                      "represent joint matrix type");
```
AlexeySachkov, line 451: If we are not checking an `ArrayTy` for `nullptr`, then we should use simple `cast` to avoid runtime `isa` check within `dyn_cast`

vmaksimo: Maybe the intention of using `dyn_cast` was to assert with a more meaningful message than `cast` would raise. And actually `cast ` does `isa` check, and asserts if it failed - `dyn_cast` seems to be more gentle in this sense. 
But agree that it's not the general use for `dyn_cast`. Tagging @MrSidims as an author to take a look.

svenvh: Keep in mind that the `getPointerElementType` call eventually has to be removed: the code cannot rely on obtaining the element type through `PtrTy` once LLVM has transitioned fully to opaque pointers, see https://llvm.org/docs/OpaquePointers.html for more info.  It might be worth trying to remove the `getPointerElementType` call now if anyone is going to look at improving this code.

AlexeySachkov: > And actually `cast` does `isa` check

It does so only within an `assert` meaning that it has no overhead on a runtime check. If we anyway ignore the result of a runtime check performed by `dyn_cast`, then why do it in the first place? I tend to think that you should be able to quickly understand what went wrong even with generic assert message from within `cast` by simply looking at backtrace and/or moving a few frames upwards in a debugger.

PR #1406, opened 2022-02-12
Support cl_bf16_conversions

KornevNikita:
OpenCL specification: https://github.com/KhronosGroup/OpenCL-Docs/pull/762
SPIR-V: https://github.com/intel/llvm/blob/sycl/sycl/doc/design/spirv-extensions/SPV_INTEL_bfloat16_conversion.asciidoc


KornevNikita:
@AlexeySotkin @MrSidims @vmaksimo @mlychkov friendly ping to take a look, please

mlychkov:
Is extension doc published somewhere in open source?

mlychkov:
May we add negative tests also?
Like trying to translate `call spir_func <4 x i16> @_Z34intel_convert_bfloat166_as_ushort6Dv6_f(<6 x float> zeroinitializer) #2` or `  %call3 = call spir_func <4 x i16> @_Z34intel_convert_bfloat164_as_ushort4Dv4_f(<8 x float> zeroinitializer) #2`?

KornevNikita:
@AlexeySotkin @MrSidims @vmaksimo @mlychkov there is public specification: https://github.com/KhronosGroup/OpenCL-Docs/pull/762

lib/SPIRV/OCLToSPIRV.cpp
```diff
@@ -1916,6 +1939,30 @@ void OCLToSPIRVBase::visitCallLdexp(CallInst *CI, StringRef MangledName,
   visitCallBuiltinSimple(CI, MangledName, DemangledName);
 }
 
+void OCLToSPIRVBase::visitConvertBFloat16AsUshort(CallInst *CI,
+                                                  StringRef DemangledName) {
+  assert(CI->getCalledFunction() && "Unexpected indirect call");
```
vmaksimo, line 1944: Maybe no need for this assert as this condition is checked in `visitCallInst`:
https://github.com/KhronosGroup/SPIRV-LLVM-Translator/blob/56cad67cb57e5b90011b3e8cecd3636b954d59d6/lib/SPIRV/OCLToSPIRV.cpp#L395-L399

vmaksimo (commented):
Overall LGTM, code is written according to the spec.

>  I think such LLVM-IR will be invalid, because these N-size functions take an appropriate N-size vector or scalar as an argument, so the compiler should fail with such a source kernel.

Agree about N-size functions, but maybe we still should try adding some negative test. 

What will happen if we translate function with unsupported size of vector (differ from existing OCL types), e.g. `call spir_func <5 x i16> @_Z34intel_convert_bfloat165_as_ushort5Dv5_f(<5 x float> zeroinitializer)`? Will it be silently ignored, ignored with warning, or even failed the translator? 
I expect the compiler should fail with such a source kernel even earlier than llvm-spirv translation, but not sure if we should rely on that expectation.

lib/SPIRV/OCLToSPIRV.cpp
KornevNikita: I agree, there is no need for this assert. About the negative test - I've checked, the translator fails with:
> virtual void SPIRV::SPIRVTypeVector::validate() const: Assertion `CompCount == 2 || CompCount == 3 || CompCount == 4 || CompCount == 8 || CompCount == 16' failed

According to Mikhail [comment](https://github.com/KhronosGroup/SPIRV-LLVM-Translator/pull/1406#issuecomment-1041392080) I think it will be enough to check that the function name matches the returning type and the argument type and add the appropriate test. I will add it a littlie bit later, thanks.

KornevNikita: Done in [9139158](https://github.com/KhronosGroup/SPIRV-LLVM-Translator/pull/1406/commits/9139158c5a3936529c38fb66d1157eabd9a3c62b) ... [c368af5](https://github.com/KhronosGroup/SPIRV-LLVM-Translator/pull/1406/commits/c368af564860f325bd09b5c8a1b9004cc6f303b2).

lib/SPIRV/OCLToSPIRV.cpp
```diff
@@ -1916,6 +1939,101 @@ void OCLToSPIRVBase::visitCallLdexp(CallInst *CI, StringRef MangledName,
...
+  if (DemangledName == kOCLBuiltinName::ConvertBFloat16AsUShort) {
+    if (!CI->getType()->isIntegerTy(16U) ||
+        !CI->getOperand(0)->getType()->isFloatTy())
+      report_fatal_error(
+          "OpConvertBFloat16AsUShort must be of i16 and take float");
+
+  } else {
+    FixedVectorType *RetTy = dyn_cast_or_null<FixedVectorType>(CI->getType());
+    FixedVectorType *ArgTy =
+        dyn_cast_or_null<FixedVectorType>(CI->getOperand(0)->getType());
```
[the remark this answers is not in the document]
KornevNikita: Anyway, I'll change it so I don't have to call `getType` multiple times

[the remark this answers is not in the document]
MrSidims: So by putting dyn_cast_or_null you are assuming, that CI->getType() can be null.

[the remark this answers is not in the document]
KornevNikita: Thanks, done in [c3a3d72](https://github.com/KhronosGroup/SPIRV-LLVM-Translator/pull/1406/commits/c3a3d7245063f9206782201bf3a399fadb00920a)

lib/SPIRV/OCLToSPIRV.cpp
```diff
@@ -1916,6 +1939,103 @@ void OCLToSPIRVBase::visitCallLdexp(CallInst *CI, StringRef MangledName,
...
+        !ArgTyVec || !ArgTyVec->getElementType()->isFloatTy())
+      report_fatal_error("OpConvertBFloat16NAsUShortN must be of <N x i16> and "
+                         "take <N x float>");
+    unsigned RetTyVecSize = RetTyVec->getNumElements();
+    unsigned ArgTyVecSize = ArgTyVec->getNumElements();
+    if (DemangledName == kOCLBuiltinName::ConvertBFloat162AsUShort2)
+      if (RetTyVecSize != 2 || ArgTyVecSize != 2)
+        report_fatal_error("ConvertBFloat162AsUShort2 must be of <2 x i16> and "
+                           "take <2 x float>");
+    if (DemangledName == kOCLBuiltinName::ConvertBFloat163AsUShort3)
```
mlychkov, line 1964: May we use `else if` here to reduce number of unnecessary string comparisons?

lib/SPIRV/OCLToSPIRV.cpp
```diff
@@ -1916,6 +1939,119 @@ void OCLToSPIRVBase::visitCallLdexp(CallInst *CI, StringRef MangledName,
...
+    if (DemangledName == kOCLBuiltinName::ConvertBFloat162AsUShort2) {
+      if (RetTyVecSize != 2 || ArgTyVecSize != 2)
+        report_fatal_error("ConvertBFloat162AsUShort2 must be of <2 x i16> and "
+                           "take <2 x float>");
+    } else {
+      if (DemangledName == kOCLBuiltinName::ConvertBFloat163AsUShort3) {
+        if (RetTyVecSize != 3 || ArgTyVecSize != 3)
+          report_fatal_error("ConvertBFloat163AsUShort3 must be of <3 x i16> "
+                             "and take <3 x float>");
+      } else {
```
mlychkov, line 1968:
```suggestion
    } else if (DemangledName == kOCLBuiltinName::ConvertBFloat163AsUShort3) {
      if (RetTyVecSize != 3 || ArgTyVecSize != 3)
        report_fatal_error("ConvertBFloat163AsUShort3 must be of <3 x i16> "
                           "and take <3 x float>");
    } else if ...```

KornevNikita: Thanks, replaced in 8394830

mlychkov (approved):
LGTM. But it would be nice if Sidorov Dmitry also approved it.

MrSidims (approved):

PR #1412, opened 2022-02-19
Fix argument attributes for SPIRV -> OCL group builtins

fzou1:
The first argument or first two arguments are not used as arguments in
new call instruction to OCL group builtins, so the related attributes
should be removed from attributes of new call instruction.

### DIFF
--- a/lib/SPIRV/SPIRVToOCL.cpp
+++ b/lib/SPIRV/SPIRVToOCL.cpp
@@ -539,6 +539,12 @@ void SPIRVToOCLBase::visitCallSPIRVGroupBuiltin(CallInst *CI, Op OC) {
 
   assert(CI->getCalledFunction() && "Unexpected indirect call");
   AttributeList Attrs = CI->getCalledFunction()->getAttributes();
+  SmallVector<AttributeSet, 2> ArgAttrs;
+  for (int I = (hasGroupOperation(OC) ? 2 : 1);
+       I < (int)Attrs.getNumAttrSets() - 2; I++)
+    ArgAttrs.push_back(Attrs.getParamAttrs(I));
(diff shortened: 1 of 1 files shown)


fzou1:
@MrSidims , @AlexeySotkin , could you help review? Thanks.

MrSidims (commented):
Functionally LGTM. Is there is a way to reflect this in the test? (may be by editing one of the existing tests)

fzou1:
> Functionally LGTM. Is there is a way to reflect this in the test? (may be by editing one of the existing tests)

Sorry. No, as it's same fix as PR #1407. issue #1425 is opened for this by @bashbaug .

fzou1:
Rebased the PR.

bashbaug:
Hi, I'd still like to understand the problem this change is intending to fix.  Can you please provide an example that was "failing" before this change and is "passing" after?

Do we need a similar change for other non-group builtins?

fzou1:
> Hi, I'd still like to understand the problem this change is intending to fix. Can you please provide an example that was "failing" before this change and is "passing" after?

It's failed as assertion in LLVM verifier: Attribute after last parameter!

Before SPIRVToOCL20Legacy pass:
%call2 = call spir_func noundef i32 @_Z22__spirv_GroupBroadcastjjm(i32 noundef 2, i32 noundef %call, i64 noundef %call1) 
After SPIRVToOCL20Legacy pass:
%call2 = call spir_func noundef i32 @_Z20work_group_broadcastjm(i32 noundef %call, i64 noundef %call1)

> Do we need a similar change for other non-group builtins?
Maybe.

fzou1:
@bashbaug , @MrSidims , any comments? Thanks

MrSidims (approved):

PR #1424, opened 2022-02-25
add support for split barriers extension SPV_INTEL_split_barrier

bashbaug:
This PR adds support for split barriers and the SPV_INTEL_split_barrier extension.

The related SPIR-V extension spec can be found here:

* https://github.com/KhronosGroup/SPIRV-Registry/pull/136

The related OpenCL C extension spec can be found here:

* https://github.com/KhronosGroup/OpenCL-Docs/pull/765


lib/SPIRV/SPIRVToOCL.h
```diff
@@ -215,6 +215,11 @@ class SPIRVToOCLBase : public InstVisitor<SPIRVToOCLBase> {
   /// - OCL1.2: barrier
   virtual void visitCallSPIRVControlBarrier(CallInst *CI) = 0;
 
+  /// Transform split __spirv_ControlBarrier barrier to:
```
MrSidims, line 218: Guess, it should be __spirv_ControlBarrierArriveINTEL and __spirv_ControlBarrierWaitINTEL

test/transcoding/SPV_INTEL_split_barrier/split_work_group_barrier_12.ll
```diff
@@ -0,0 +1,101 @@
...
+; CHECK-ERROR-NEXT: SPV_INTEL_split_barrier
+
+; ModuleID = 'split_barrier.cl'
+source_filename = "split_barrier.cl"
+target datalayout = "e-i64:64-v16:16-v24:32-v32:32-v48:64-v96:128-v192:256-v256:256-v512:512-v1024:1024"
+target triple = "spir64"
+
+; CHECK-SPIRV: Capability SplitBarrierINTEL
+; CHECK-SPIRV: Extension "SPV_INTEL_split_barrier"
+; CHECK-SPIRV: Name [[TEST_FUNC:[0-9]+]] "test"
```
MrSidims, line 32: nit: 
```suggestion
; CHECK-SPIRV: Name [[#TEST_FUNC:]] "test"
```

lib/SPIRV/OCLToSPIRV.cpp
```diff
@@ -1890,6 +1897,37 @@ void OCLToSPIRVBase::visitSubgroupAVCBuiltinCallWithSampler(
       &Attrs);
 }
 
+void OCLToSPIRVBase::visitCallSplitBarrierINTEL(CallInst *CI,
+                                                StringRef DemangledName) {
+  auto Lit = getBarrierLiterals(CI);
```
MrSidims, line 1902: nit:
Some static analyzers will be paranoid about nullptrness of CI. Even there is an appropriate check in the call stack before  `visitCallSplitBarrierINTEL` can we also put an assertion here?

MrSidims (commented):
LGTM, just few nits

lib/SPIRV/SPIRVToOCL.h
bashbaug: Yes, I've fixed these comments so now they are all a variant of:

```c
  /// Transform split __spirv_ControlBarrierArriveINTEL and
  /// __spirv_ControlBarrierWaitINTEL barrier to...
```

test/transcoding/SPV_INTEL_split_barrier/split_work_group_barrier_12.ll
bashbaug: This change didn't work for me, but it also doesn't look like other tests have a similar check, so I've simply removed them for now.  Please check and let me know if this will work for you.

lib/SPIRV/OCLToSPIRV.cpp
bashbaug: I'm willing to make a change here but I'm not sure what to type.  This particular split barrier addition is purposefully very similar to `visitCallBarrier`, which has no assertion.  Is this suggestion something that would be more appropriate for a separate PR?

MrSidims: Okay, there is actually no need for the assertion, since there is the appropriate check earlier in the stack.

MrSidims (approved):
LGTM, will merge the patch by EOD

PR #1434, opened 2022-03-05
Support complex recursive types that use an array of pointers

vmaksimo:
The use of pointer arrays inside the recursive type leads to the SPIR-V IR, where
the TypePointer is declared after TypeArray. In turn, TypeArray can't handle the
unknown element type, because the info from TypeForwardPointer is not used.

Now the translator stores forward declared IDs to use them in backward
translation. It actually means that the translator "knows" about the type but it
would be handled later.

### DIFF
--- a/lib/SPIRV/SPIRVReader.cpp
+++ b/lib/SPIRV/SPIRVReader.cpp
@@ -475,6 +475,12 @@ Type *SPIRVToLLVM::transType(SPIRVType *T, bool IsClassMember) {
         getSPIRVTypeName(kSPIRVTypeName::JointMatrixINTEL, SS.str());
     return mapType(T, getOrCreateOpaquePtrType(M, Name));
   }
+  case OpTypeForwardPointer: {
+    SPIRVTypeForwardPointer *FP =
+        static_cast<SPIRVTypeForwardPointer *>(static_cast<SPIRVEntry *>(T));
+    return mapType(T, transType(static_cast<SPIRVType *>(
+                          BM->getEntry(FP->getPointerId()))));
+  }
 
   default: {
     auto OC = T->getOpCode();
--- a/lib/SPIRV/libSPIRV/SPIRVModule.cpp
+++ b/lib/SPIRV/libSPIRV/SPIRVModule.cpp
@@ -694,7 +694,7 @@ SPIRVEntry *SPIRVModuleImpl::addEntry(SPIRVEntry *Entry) {
     } else
       IdEntryMap[Id] = Entry;
   } else {
-    // Collect entries with no ID to de-allocate them at end.
+    // Collect entries with no ID to de-allocate them at the end.
     // Entry of OpLine will be deleted by std::shared_ptr automatically.
     if (Entry->getOpCode() != OpLine)
       EntryNoId.insert(Entry);
@@ -1736,6 +1736,11 @@ class TopologicalSort {
       return true;
     State = Discovered;
     for (SPIRVEntry *Op : E->getNonLiteralOperands()) {
+      if (Op->getOpCode() == OpTypeForwardPointer) {
+        SPIRVEntry *FP = E->getModule()->getEntry(
+            static_cast<SPIRVTypeForwardPointer *>(Op)->getPointerId());
+        Op = FP;
+      }
       if (EntryStateMap[Op] == Visited)
         continue;
       if (visit(Op)) {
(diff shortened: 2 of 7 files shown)


vmaksimo:
Tagging @MrSidims @AlexeySachkov to take a look

lib/SPIRV/libSPIRV/SPIRVModule.cpp
```diff
@@ -693,9 +694,15 @@ SPIRVEntry *SPIRVModuleImpl::addEntry(SPIRVEntry *Entry) {
     } else
       IdEntryMap[Id] = Entry;
   } else {
+    // Collect entries with no ID to de-allocate them at end.
```
MrSidims, line 697:
```suggestion
    // Collect entries with no ID to de-allocate them at the end.
```

lib/SPIRV/SPIRVReader.cpp
```diff
@@ -475,6 +475,12 @@ Type *SPIRVToLLVM::transType(SPIRVType *T, bool IsClassMember) {
         getSPIRVTypeName(kSPIRVTypeName::JointMatrixINTEL, SS.str());
     return mapType(T, getOrCreateOpaquePtrType(M, Name));
   }
+  case OpTypeForwardPointer: {
+    SPIRVTypeForwardPointer *FP =
+        static_cast<SPIRVTypeForwardPointer *>(static_cast<SPIRVEntry *>(T));
+    return mapType(T, transType(static_cast<SPIRVType *>(
+                          BM->getEntry(FP->getPointerId()))));
```
MrSidims, line 482: What about
```suggestion
                          BM->getEntry(FP->getOperand(0)->getId()()))));
```
to not add extra class member?

MrSidims (commented):
Do --to-text work as well with this change?

lib/SPIRV/SPIRVReader.cpp
vmaksimo: @AlexeySachkov is right, we can't access the Pointer field here as it's not known yet.

lib/SPIRV/libSPIRV/SPIRVModule.cpp
```diff
@@ -749,8 +756,15 @@ SPIRVId SPIRVModuleImpl::getId(SPIRVId Id, unsigned Increment) {
...
-  assert(Loc != IdEntryMap.end() && "Id is not in map");
-  return Loc->second;
+  if (Loc != IdEntryMap.end()) {
+    return Loc->second;
+  }
+  SPIRVIdToEntryMap::const_iterator LocFwd = IdTypeForwardMap.find(Id);
+  if (LocFwd != IdTypeForwardMap.end()) {
+    return LocFwd->second;
+  }
+  assert("Id is not in map");
```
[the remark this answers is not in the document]
vmaksimo: Applied suggestion in 33049395b

lib/SPIRV/libSPIRV/SPIRVType.h
```diff
@@ -282,9 +282,11 @@ class SPIRVTypeForwardPointer : public SPIRVEntryNoId<OpTypeForwardPointer> {
       : Pointer(nullptr), SC(StorageClassUniformConstant) {}
 
   SPIRVTypePointer *getPointer() const { return Pointer; }
+  SPIRVId getPointerId() const { return PointerId; }
   _SPIRV_DCL_ENCDEC
 private:
   SPIRVTypePointer *Pointer;
+  SPIRVId PointerId;
```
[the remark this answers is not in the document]
vmaksimo: Fixed in 33049395b

lib/SPIRV/libSPIRV/SPIRVStream.h
```diff
@@ -193,6 +193,7 @@ template <typename T>
 const SPIRVEncoder &operator<<(const SPIRVEncoder &O, T *P) {
   return O << P->getId();
 }
+template <> const SPIRVEncoder &operator<<(const SPIRVEncoder &O, SPIRVType *P);
```
svenvh, line 196: Can `P` be const too?
```suggestion
template <> const SPIRVEncoder &operator<<(const SPIRVEncoder &O, const SPIRVType *P);
```

vmaksimo: I've checked, and the answer is no. The translator uses non-const SPIRVType in the general template, so the `const SPIRVType` specialization won't ever be used.

svenvh: I see, thanks for checking anyway!

svenvh (approved):

PR #1436, opened 2022-03-11
Fix annotation translation when the value is 0

pauzinl:
This patch fixes translation for annotations, when all values of annotation
are 0.

### DIFF
--- a/lib/SPIRV/SPIRVWriter.cpp
+++ b/lib/SPIRV/SPIRVWriter.cpp
@@ -2579,6 +2579,17 @@ void processOptionalAnnotationInfo(Constant *Const,
         AnnotationString += std::to_string(CInt->getSExtValue());
       }
     }
+  } else if (auto *ZeroStruct =
+                 dyn_cast<ConstantAggregateZero>(Const->getOperand(0))) {
+    // It covers case when when all elements of struct are 0 and they become
+    // zeroinitializer. It represents like: { i32 i32 ... } zeroinitializer
+    uint32_t NumOperands = ZeroStruct->getType()->getStructNumElements();
+    AnnotationString += ": ";
+    AnnotationString += "0";
+    for (uint32_t I = 1; I != NumOperands; ++I) {
+      AnnotationString += ", ";
+      AnnotationString += "0";
+    }
   }
 }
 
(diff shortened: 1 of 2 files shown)


pauzinl:
@MrSidims @KornevNikita take a look please

lib/SPIRV/SPIRVWriter.cpp
```diff
@@ -2579,6 +2579,17 @@ void processOptionalAnnotationInfo(Constant *Const,
         AnnotationString += std::to_string(CInt->getSExtValue());
       }
     }
+  } else if (auto *ZeroStruct =
+                 dyn_cast<ConstantAggregateZero>(Const->getOperand(0))) {
+    // It covers case when when all elements of struct are 0 and they become
```
KornevNikita, line 2584: Typo

test/transcoding/annotate_attribute.ll
```diff
@@ -44,8 +47,10 @@ target triple = "spir64-unknown-linux"
 @.str.5 = private unnamed_addr constant [4 x i8] c"128\00", section "llvm.metadata"
 @.str.6 = private unnamed_addr constant [4 x i8] c"qux\00", section "llvm.metadata"
 @.str.7 = private unnamed_addr constant [22 x i8] c"my_custom_annotations\00", section "llvm.metadata"
+@.str.8 = private unnamed_addr constant [31 x i8] c"annotation_with_zerointializer\00", section "llvm.metadata"
 @.args.0 = private unnamed_addr constant { i32, i32 } { i32 30, i32 60 }, section "llvm.metadata"
 @.args.1 = private unnamed_addr constant { i32, i32, i32 } { i32 20, i32 60, i32 80 }, section "llvm.metadata"
+@.args.2 = private unnamed_addr constant { i32, i32, i32 } zeroinitializer, section "llvm.metadata"
```
KornevNikita, line 53: Could you also add a check if the value is `false` (`[[clang::annotate("my_annotation", false)]]`). And add a case for a mix with 0 e.g. `{ i32 0, i32 1, i32 0 }` to make sure that the issue occures only if all values are zero?

lib/SPIRV/SPIRVWriter.cpp
pauzinl: Thanks! Fixed it

MrSidims (approved):
LGTM
Will merge after @KornevNikita approval

KornevNikita (approved):
Thanks! Also LGTM!

PR #1437, opened 2022-03-14
[Upstream] Rename TARGET_TRIPLE to LLVM_TARGET_TRIPLE

pvchupin:
Required after this change done in LLVM upstream:

    commit 0c0f6cfb7b9fab0ee69ef67ed95c194698c9d794
    Author: Petr Hosek <phosek@google.com>

    [CMake] Rename TARGET_TRIPLE to LLVM_TARGET_TRIPLE

    This clarifies that this is an LLVM specific variable and avoids
    potential conflicts with other projects.

    Differential Revision: https://reviews.llvm.org/D119918

Can be observed as:
lit.py: /usr/lib/llvm-15/build/utils/lit/lit/llvm/config.py:304: fatal:
Could not turn '' into Itanium ABI triple


pvchupin:
I believe 2 LIT fails are unrelated to this specific change, it just uncovers other problems

MrSidims (approved):
Thanks!
I believe, that some other LLVM change resulted in these test to fail, since locally with an 'old' LLVM and fresh translator they are passing.

PR #1438, opened 2022-03-14
Explicitly add 'sret' parameter attribute for builtins

MrSidims:
It's required after:
commit 2182665305d90ae7d0d2b573dced9ef0ef414a9a
Author: Nikita Popov <npopov@redhat.com>
Date:   Fri Mar 11 17:30:34 2022 +0100

    [Bitcode] Don't confuse type attributes on declaration and call

    We should not be using APIs here that try to fetch the attribute
    from both the call attributes and the function attributes. Otherwise
    we'll try to upgrade a non-existent sret attribute on the call using
    the attribute on the function.

The patch fixes following failures:
Failed Tests (2):
  LLVM_SPIRV :: transcoding/BuildNDRange.ll
  LLVM_SPIRV :: transcoding/BuildNDRange_2.ll


MrSidims:
@svenvh @vmaksimo @KornevNikita @pauzinl please take a look

vmaksimo (approved):

PR #1446, opened 2022-03-21
Expand annotation-carried decoration parsing

steffenlarsen:
The SPIR-V translator currently allows LLVM IR annotation intrinsics to carry a selection of INTEL extension decorations as part of the string in its second argument. These changes expand the parsing of these to allow generic parsing of decorations in annotation strings, identifying the decorations by their SPIR-V integer identifiers rather than reserved names.

It is legal to have decorations with reserved named specified by their SPIR-V integer identifiers, but to allow backwards compatibility the reverse translation will use the reserved names for all decorations with them.

### DIFF
--- a/lib/SPIRV/SPIRVWriter.cpp
+++ b/lib/SPIRV/SPIRVWriter.cpp
@@ -2639,13 +2639,17 @@ void processAnnotationString(IntrinsicInst *II, std::string &AnnotationString) {
 static bool tryParseAnnotationDecoValues(StringRef ValueStr,
                                          std::vector<std::string> &ParsedArgs) {
   unsigned ValueStart = 0;
-  bool ParsingStrLit = false;
+  bool IsParsingStringLiteral = false;
   for (unsigned I = 0; I < ValueStr.size(); ++I) {
     const char CurrentC = ValueStr[I];
-    if (ParsingStrLit) {
+    if (IsParsingStringLiteral) {
       if (CurrentC == '"') {
-        ParsingStrLit = false;
+        // We have reached the end of a string literal and have the arg string
+        // between this character and the start of the string literal.
+        IsParsingStringLiteral = false;
         ParsedArgs.push_back(ValueStr.substr(ValueStart, I - ValueStart).str());
+        // End of a string literal must either be at the end of the values or
+        // right before a comma.
         if (I + 1 != ValueStr.size() && ValueStr[I + 1] != ',')
           return false;
         // Skip the , delimiter and go directly to the start of next value.
@@ -2654,22 +2658,33 @@ static bool tryParseAnnotationDecoValues(StringRef ValueStr,
       continue;
     }
     if (CurrentC == ',') {
+      // Since we are not currently in a string literal, comma denotes a
+      // separation of decoration arguments and we can copy the substring we are
+      // currently parsing.
       ParsedArgs.push_back(ValueStr.substr(ValueStart, I - ValueStart).str());
       ValueStart = I + 1;
       continue;
     }
     if (CurrentC == '"') {
+      // We are entering a string literal. This must be either at the beginning
+      // of the values or right after a comma.
       if (I != 0 && ValueStr[I - 1] != ',')
         return false;
-      ParsingStrLit = true;
+      IsParsingStringLiteral = true;
       ValueStart = I + 1;
       continue;
     }
+    // Any other character will be consumed as part of the argument.
   }
+  // If we were still parsing a decoration argument when reaching the end of the
+  // parsed string, we must be at the end of the argument.
   if (ValueStart < ValueStr.size())
     ParsedArgs.push_back(
         ValueStr.substr(ValueStart, ValueStr.size() - ValueStart).str());
-  return !ParsingStrLit;
+
+  // At the end, the arguments parsed are valid if we were not parsing a string
+  // literal with no end.
+  return !IsParsingStringLiteral;
 }
 
 AnnotationDecorations tryParseAnnotationString(SPIRVModule *BM,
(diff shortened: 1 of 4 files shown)


lib/SPIRV/SPIRVWriter.cpp
```diff
@@ -2628,52 +2633,104 @@ void processAnnotationString(IntrinsicInst *II, std::string &AnnotationString) {
...
+    Decorates.MemoryAttributesVec.emplace_back(
+        DecorationUserSemantic, std::vector<std::string>{AnnotatedCode.str()});
+    return Decorates;
+  }
+
   const bool AllowFPGAMemAccesses =
       BM->isAllowedToUseExtension(ExtensionID::SPV_INTEL_fpga_memory_accesses);
   const bool AllowFPGAMemAttr = BM->isAllowedToUseExtension(
       ExtensionID::SPV_INTEL_fpga_memory_attributes);
-  if (!AllowFPGAMemAccesses && !AllowFPGAMemAttr) {
```
vmaksimo, line 2638: Why have we dropped this condition? I see that now we do `emplace_back` for all the cases, but not sure if it should have been removed.

lib/SPIRV/SPIRVWriter.cpp
```diff
@@ -2753,27 +2820,31 @@ void addAnnotationDecorations(SPIRVEntry *E, DecorationsInfoVec &Decorations) {
 
     switch (I.first) {
     case DecorationUserSemantic:
-      E->addDecorate(new SPIRVDecorateUserSemanticAttr(E, I.second));
+      assert(I.second.size() == 1);
```
vmaksimo, line 2823: Maybe it'd be better to use `checkError` here instead of assertions to fail in a more gentle way. However, it's a recommendation, not a request.

lib/SPIRV/SPIRVWriter.cpp
```diff
@@ -2753,27 +2820,31 @@ void addAnnotationDecorations(SPIRVEntry *E, DecorationsInfoVec &Decorations) {
...
     } break;
     case DecorationMergeINTEL: {
       if (M->isAllowedToUseExtension(
               ExtensionID::SPV_INTEL_fpga_memory_attributes)) {
-        StringRef Name = StringRef(I.second).split(':').first;
-        StringRef Direction = StringRef(I.second).split(':').second;
+        assert(I.second.size() == 2);
         E->addDecorate(
-            new SPIRVDecorateMergeINTELAttr(E, Name.str(), Direction.str()));
+            new SPIRVDecorateMergeINTELAttr(E, I.second[0], I.second[1]));
```
vmaksimo, line 2838: Nice to mention what are `I.second[N]` (to add a comment that these are Name and Direction)

lib/SPIRV/SPIRVWriter.cpp
```diff
@@ -2628,52 +2633,104 @@ void processAnnotationString(IntrinsicInst *II, std::string &AnnotationString) {
       processOptionalAnnotationInfo(C, AnnotationString);
 }
 
+// Try to parse the annotation decoration values in a string. These values must
+// be separated by a "," and must be either a word (including numbers) or a
+// quotation mark enclosed string.
+static bool tryParseAnnotationDecoValues(StringRef ValueStr,
+                                         std::vector<std::string> &ParsedArgs) {
+  unsigned ValueStart = 0;
+  bool ParsingStrLit = false;
```
vmaksimo, line 2642:
```suggestion
  bool IsParsingStringLiteral = false;
```

lib/SPIRV/SPIRVWriter.cpp
```diff
@@ -2628,52 +2633,104 @@ void processAnnotationString(IntrinsicInst *II, std::string &AnnotationString) {
...
+  for (unsigned I = 0; I < ValueStr.size(); ++I) {
+    const char CurrentC = ValueStr[I];
+    if (ParsingStrLit) {
+      if (CurrentC == '"') {
+        ParsingStrLit = false;
+        ParsedArgs.push_back(ValueStr.substr(ValueStart, I - ValueStart).str());
+        if (I + 1 != ValueStr.size() && ValueStr[I + 1] != ',')
```
vmaksimo, line 2649: I'd personally like to have some comments on exit conditions in this function - nice to mention why (maybe with examples?) the parsed string is not valid.

vmaksimo (commented):
Agree with @MrSidims, the design idea looks good.
I've left some minor comments and suggestions about the implementation itself. And maybe we need some negative tests if it's possible.

lib/SPIRV/SPIRVWriter.cpp
steffenlarsen: The intention was to allow member decorations outside `SPV_INTEL_fpga_memory_accesses` and `SPV_INTEL_fpga_memory_attributes` to be parsed by their identifier. When parsing them by reserved names later, these conditions will be checked and the parsing will be skipped if they are not allowed.

Granted, currently there is no support for any member decorations outside the ones with reserved names, so we could keep the check here for now, but I suspect it will have to go at some point.

lib/SPIRV/SPIRVWriter.cpp
steffenlarsen: I've added some comments documenting most of the parsing. Is this along the lines of what you were thinking?

lib/SPIRV/SPIRVWriter.cpp
```diff
@@ -2553,17 +2554,21 @@ struct IntelLSUControlsInfo {
     DecorationsInfoVec ResultVec;
     // Simple flags
     if (BurstCoalesce)
-      ResultVec.emplace_back(DecorationBurstCoalesceINTEL, "");
+      ResultVec.emplace_back(DecorationBurstCoalesceINTEL,
+                             std::vector<std::string>());
```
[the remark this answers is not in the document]
steffenlarsen: Sadly not. `emplace_back` is not happy about passing initializer lists on.

lib/SPIRV/SPIRVWriter.cpp
vmaksimo: Absolutely, thanks! It'll simplify the code support in future.

lib/SPIRV/SPIRVWriter.cpp
vmaksimo: Yes, please add the same comment right here too

lib/SPIRV/SPIRVWriter.cpp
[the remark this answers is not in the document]
vmaksimo: Okay, thanks for the clarification

lib/SPIRV/SPIRVWriter.cpp
vmaksimo: I see what's going on, thanks. I believe you made the right assumption.

vmaksimo:
> > And maybe we need some negative tests if it's possible.
> 
> Do we have some other negative tests I can draw inspiration from?

@steffenlarsen I've searched through the existing tests, and haven't found the implemented negative tests for annotations/decoration (although, there are fixme/todo marks).
If you'll decide to try doing some negative tests, this test can help with an inspiration https://github.com/KhronosGroup/SPIRV-LLVM-Translator/blob/master/test/transcoding/AllowIntrinsics.ll#L21. But for the current situation, I believe they are not necessary.

> clang-tidy seems to be choking SPIRVErrorEnum.h but not due to the changes in this PR. Is there anything I can do about that?

I don't know for sure, maybe @MrSidims can suggest something about that.
I personally think that we can use an existing `SPIRVEC_InvalidLlvmModule` error instead of new ones. I'd proceed with them to keep the generalization of error messages.

lib/SPIRV/SPIRVWriter.cpp
steffenlarsen: I thought I added a comment here, but it seems like it got lost somehow. It should be there now.

steffenlarsen:
> I personally think that we can use an existing `SPIRVEC_InvalidLlvmModule` error instead of new ones. I'd proceed with them to keep the generalization of error messages.

I'm okay with that. I have changed it so it uses `SPIRVEC_InvalidLlvmModule` instead and clang-tidy is happy again. I suspect that it could show up again in the future however.

MrSidims (approved):
LGTM, thanks!

PR #1447, opened 2022-03-21
Remove Bitcast -> llvm.memset optimization in reverse translation

vmaksimo:
The change is done in order to avoid `llvm.memcpy` -> `llvm.memset` transformation 
which can be an incorrect assumption.

Details:
SPIR-V does not have a memset instruction, there is a direct mapping only for
`llvm.memcpy` to `OpCopyMemory*`.
To handle `llvm.memset`, the input LLVM IR is lowered by `SPIRVRegularizeLLVM`
pass. The mentioned optimization tried to restore the memset instruction but it
surely does not cover all the possible cases (e.g., memcpy is not a memset inst).

Possibly, the number of follow-up fixes for different corner cases can prove
that the optimization is not absolutely correct, and it's better to rely on
regularize pass.


vmaksimo:
Tagging @MrSidims to take a look

MrSidims (approved):
LGTM

svenvh (approved):

PR #1448, opened 2022-03-22
Align group-instructions.ll test with the spec

KornevNikita:
According to https://github.com/KhronosGroup/SPIRV-Registry/blob/main/extensions/KHR/SPV_KHR_uniform_group_instructions.asciidoc
the OpGroupLogicalAnd/Or/Xor X argument and the result type must be
scalar/vector of bool.


MrSidims (approved):

KornevNikita:
@aratajew take a look please

aratajew (approved):

svenvh (approved):

PR #1450, opened 2022-03-23
Extend bool arg of shift operations

KornevNikita:
This is a patch to regularize lshr instruction with an i1 argument.
According to the SPIR-V specification OpShiftRightLogical operands
should be of integer type.

### DIFF
--- a/lib/SPIRV/SPIRVRegularizeLLVM.cpp
+++ b/lib/SPIRV/SPIRVRegularizeLLVM.cpp
@@ -128,7 +128,7 @@ public:
   // i1 operands, they are treated as a bool. We need to extend them to i32 to
   // comply with the specification. For example: "%shift = lshr i1 0, 1";
   // The bit instruction should be changed to the extended version
-  // "%shift = i32 0, 1" so the args are treated as int operands.
+  // "%shift = lshr i32 0, 1" so the args are treated as int operands.
   Value *extendBitInstBoolArg(Instruction *OldInst);
 
   static std::string lowerLLVMIntrinsicName(IntrinsicInst *II);
--- a/lib/SPIRV/SPIRVRegularizeLLVM.cpp
+++ b/lib/SPIRV/SPIRVRegularizeLLVM.cpp
@@ -123,6 +123,18 @@ public:
   void expandVEDWithSYCLTypeSRetArg(Function *F);
   void expandVIDWithSYCLTypeByValComp(Function *F);
 
+  // According to the specification, the operands of a shift instruction must be
+  // a scalar/vector of integer. When LLVM-IR contains a shift instruction with
+  // i1 operands, they are treated as a bool. We need to extend them to i32 to
+  // comply with the specification. For example:
+  // %21 = load i1, i1 addrspace(4)* %20
+  // %22 = load i32, i32 addrspace(4)* %9
+  // %23 = trunc i32 %22 to i1
+  // %24 = lshr i1 %21, %23
+  // %25 = zext i1 %24 to i32
+  // %24 should be changed to take (i32(%21), %22); %25 users should use %24
+  Value *extendBitInstBoolArg(Instruction *OldInst);
+
   static std::string lowerLLVMIntrinsicName(IntrinsicInst *II);
   void adaptStructTypes(StructType *ST);
   static char ID;
@@ -418,6 +430,13 @@ void SPIRVRegularizeLLVMBase::expandSYCLTypeUsing(Module *M) {
     expandVIDWithSYCLTypeByValComp(F);
 }
 
+Value *SPIRVRegularizeLLVMBase::extendBitInstBoolArg(Instruction *II) {
+  IRBuilder<> Builder(II);
+  auto *NewBase = Builder.CreateZExt(II->getOperand(0), Builder.getInt32Ty());
+  auto *ShiftInt32 = cast<User>(II->getOperand(1))->getOperand(0);
+  return Builder.CreateLShr(NewBase, ShiftInt32);
+}
+
 void SPIRVRegularizeLLVMBase::adaptStructTypes(StructType *ST) {
   if (!ST->hasName())
     return;
@@ -560,6 +579,19 @@ bool SPIRVRegularizeLLVMBase::regularize() {
               lowerFunnelShift(II);
             else if (II->getIntrinsicID() == Intrinsic::umul_with_overflow)
               lowerUMulWithOverflow(II);
+          }
+        }
+
+        if (II.isLogicalShift() &&
+            II.getOperand(0)->getType()->isIntegerTy(1)) {
+          if (II.getOpcode() == Instruction::LShr) {
+            auto *NewLSHR = extendBitInstBoolArg(&II);
+            for (auto U : II.users()) {
+              U->replaceAllUsesWith(NewLSHR);
+              ToErase.push_back(cast<Instruction>(U));
+            }
+            ToErase.push_back(&II);
+            ToErase.push_back(cast<Instruction>(II.getOperand(1)));
           }
         }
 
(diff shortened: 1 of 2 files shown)


KornevNikita:
@MrSidims @vmaksimo @mlychkov @AlexeySachkov take a look please

lib/SPIRV/SPIRVRegularizeLLVM.cpp
```diff
@@ -563,6 +582,19 @@ bool SPIRVRegularizeLLVMBase::regularize() {
           }
         }
 
+        if (II.isLogicalShift() &&
+            II.getOperand(0)->getType()->isIntegerTy(1)) {
+          if (II.getOpcode() == Instruction::LShr) {
```
vmaksimo, line 587: `isLogicalShift()` returns true for right-shift `shr` instruction too, which is mapped to `OpShiftLeftLogical` SPIR-V operation. Should we do the same extension for this instruction too?

lib/SPIRV/SPIRVRegularizeLLVM.cpp
```diff
@@ -123,6 +123,18 @@ class SPIRVRegularizeLLVMBase {
...
+  // According to the specification, the operands of a shift instruction must be
+  // a scalar/vector of integer. When LLVM-IR contains a shift instruction with
+  // i1 operands, they are treated as a bool. We need to extend them to i32 to
+  // comply with the specification. For example:
+  // %21 = load i1, i1 addrspace(4)* %20
+  // %22 = load i32, i32 addrspace(4)* %9
+  // %23 = trunc i32 %22 to i1
+  // %24 = lshr i1 %21, %23
+  // %25 = zext i1 %24 to i32
+  // %24 should be changed to take (i32(%21), %22); %25 users should use %24
```
vmaksimo, line 135: I believe there is no need for such a detailed IR. Let's simplify this in favor of readability, e.g.:
```suggestion
  // %shift = lshr i1 0, 1
  // The instruction should be changed to the extended version "%shift = i32 0, 1" so the args are treated as int operands;
```

lib/SPIRV/SPIRVRegularizeLLVM.cpp
```diff
@@ -563,6 +582,19 @@ bool SPIRVRegularizeLLVMBase::regularize() {
           }
         }
 
+        if (II.isLogicalShift() &&
```
vmaksimo, line 585: Please add some high-level comment here about why are we doing this regularization.

vmaksimo (commented):
Will the fix work for `lshr <2 x i1> < i1 1, i32 4>, < i1 0, i32 2>`? The spec says that `lshr` args can be the integer or vector of integer type.

lib/SPIRV/SPIRVRegularizeLLVM.cpp
KornevNikita: Thanks, all done in [343c988](https://github.com/KhronosGroup/SPIRV-LLVM-Translator/pull/1450/commits/343c9888c8e3b8e57046e51d364769047367fb13)

lib/SPIRV/SPIRVRegularizeLLVM.cpp
```diff
@@ -123,16 +123,9 @@ class SPIRVRegularizeLLVMBase {
...
-  // comply with the specification. For example:
-  // %21 = load i1, i1 addrspace(4)* %20
-  // %22 = load i32, i32 addrspace(4)* %9
-  // %23 = trunc i32 %22 to i1
-  // %24 = lshr i1 %21, %23
-  // %25 = zext i1 %24 to i32
-  // %24 should be changed to take (i32(%21), %22); %25 users should use %24
+  // %shift = lshr i1 0, 1
+  // The bit instruction should be changed to the extended version
+  // "%shift = i32 0, 1" so the args are treated as int operands
```
vmaksimo, line 128: Please add some comments about the specification requirement, like it was in the initial commit.
Also, the IR in the last line should be `%shift = lshr i32 0, 1`.

lib/SPIRV/SPIRVRegularizeLLVM.cpp
```diff
@@ -432,9 +425,24 @@ void SPIRVRegularizeLLVMBase::expandSYCLTypeUsing(Module *M) {
...
 Value *SPIRVRegularizeLLVMBase::extendBitInstBoolArg(Instruction *II) {
   IRBuilder<> Builder(II);
-  auto *NewBase = Builder.CreateZExt(II->getOperand(0), Builder.getInt32Ty());
-  auto *ShiftInt32 = cast<User>(II->getOperand(1))->getOperand(0);
-  return Builder.CreateLShr(NewBase, ShiftInt32);
+  auto *ArgTy = II->getOperand(0)->getType();
+  Type *NewArgType = nullptr;
+  if (ArgTy->isIntegerTy()) {
+    NewArgType = Builder.getInt32Ty();
+  } else {
```
vmaksimo, line 432: I see that the function is called only for integers and vectors of integers now, but it's not guaranteed later.
I suggest we check for the vector type here too, e.g. `else if (ArgTy->isIntVectorTy())`. This will require dealing with the `else` case (e.g. warning + `return II`, or something else like asserts if you find it right). It would be helpful if the function would be misused in the future.

vmaksimo (commented):
Overall LGTM, but have some new comments to handle.

lib/SPIRV/SPIRVRegularizeLLVM.cpp
KornevNikita: Thanks, [2a8c9e3](https://github.com/KhronosGroup/SPIRV-LLVM-Translator/pull/1450/commits/2a8c9e3ed30da49314b7cd464c45597f646674d6)

lib/SPIRV/SPIRVRegularizeLLVM.cpp
```diff
@@ -563,6 +596,20 @@ bool SPIRVRegularizeLLVMBase::regularize() {
...
         }
 
+        // Translator treats i1 as boolean, but bit instructions take
+        // a scalar/vector integers, so we have to extend such arguments
+        if (II.isLogicalShift() &&
+            II.getOperand(0)->getType()->isIntOrIntVectorTy(1)) {
+          auto *NewInst = extendBitInstBoolArg(&II);
+          for (auto *U : II.users()) {
+            if (cast<Instruction>(U)->getOpcode() == Instruction::ZExt) {
+              U->replaceAllUsesWith(NewInst);
```
MrSidims, line 606: Lest call `dropAllReferences` here as well

KornevNikita: Thanks, done 1597dff

vmaksimo (approved):
LGTM, thanks!

lib/SPIRV/SPIRVRegularizeLLVM.cpp
```diff
@@ -123,6 +123,14 @@ class SPIRVRegularizeLLVMBase {
   void expandVEDWithSYCLTypeSRetArg(Function *F);
   void expandVIDWithSYCLTypeByValComp(Function *F);
 
+  // According to the specification, the operands of a shift instruction must be
+  // a scalar/vector of integer. When LLVM-IR contains a shift instruction with
+  // i1 operands, they are treated as a bool. We need to extend them to i32 to
+  // comply with the specification. For example: "%shift = lshr i1 0, 1";
+  // The bit instruction should be changed to the extended version
+  // "%shift = i32 0, 1" so the args are treated as int operands.
```
AlexeySachkov, line 131:
```suggestion
  // "%shift = lshr i32 0, 1" so the args are treated as int operands.
```

PR #1451, opened 2022-03-24
Allow OpenCL builtins mangling for non-standard precision integer

KornevNikita:


KornevNikita:
@MrSidims @mlychkov take a look please

MrSidims (approved):

PR #1452, opened 2022-03-25
Fix wrong DICompositeType size

KornevNikita:
If count = -1 we shouldln't cast it to unsigned


KornevNikita:
@MrSidims @mlychkov take a look please

MrSidims (approved):

PR #1456, opened 2022-03-29
Add 'Use' parameter to TypeJointMatrixINTEL

MrSidims:
'Use' is an optional parameter that shows where in a math operation the matrix
is used. It must be the result of a constant instruction with scalar
'integer type'.

Spec: https://github.com/intel/llvm/pull/5944

Signed-off-by: Dmitry Sidorov [dmitry.sidorov@intel.com](mailto:dmitry.sidorov@intel.com)


MrSidims:
@KornevNikita @pauzinl please take a look. I'll post link to the spec a bit later. Pointer here is that we are adding an optional parameter to TypeJointMatrixINTEL instruction.

lib/SPIRV/SPIRVRegularizeLLVM.cpp
```diff
@@ -445,16 +445,26 @@ void SPIRVRegularizeLLVMBase::adaptStructTypes(StructType *ST) {
...
+    TypeLayout.push_back(ArrayTy->getNumElements());
     for (size_t I = 1; I != 4; ++I) {
       ArrayTy = dyn_cast<ArrayType>(ArrayTy->getElementType());
       assert(ArrayTy &&
              "Expected a element type to represent joint matrix type");
-      TypeLayout[I] = ArrayTy->getNumElements();
+      TypeLayout.push_back(ArrayTy->getNumElements());
+    }
+    // JointMatrixINTEL type can have optional 'Use' parameter, which is encoded
+    // as another array dimention. In case if it has default 'Unnecessary' (4)
```
KornevNikita, line 460: May we move this comment "In case if it has default 'Unnecessary' (4) ..." down to `if (UseInt != 4)`?

lib/SPIRV/libSPIRV/SPIRVType.h
```diff
@@ -1081,11 +1077,20 @@ class SPIRVTypeJointMatrixINTEL : public SPIRVType {
...
-  SPIRVValue *getLayout() const { return Layout; }
-  SPIRVValue *getRows() const { return Rows; }
-  SPIRVValue *getColumns() const { return Columns; }
-  SPIRVValue *getScope() const { return Scope; }
+  SPIRVValue *getRows() const { return Args[0]; }
+  SPIRVValue *getColumns() const { return Args[1]; }
+  SPIRVValue *getLayout() const { return Args[2]; }
+  SPIRVValue *getScope() const { return Args[3]; }
+  SPIRVValue *getUse() const {
+    if (Args.size() > 4)
```
KornevNikita, line 1090: You can ignore it. But do you mind using: `return Args.size() > 4 ?  return Args[4] : nullptr`

KornevNikita (approved):
LGTM 👍🏻

lib/SPIRV/SPIRVRegularizeLLVM.cpp
MrSidims: All of this block handles Use parameter, so no

lib/SPIRV/libSPIRV/SPIRVType.h
MrSidims: Thanks! Applied

lib/SPIRV/SPIRVRegularizeLLVM.cpp
MrSidims: Ah, I've misread the comment, it's not about all comment move, but about a particular part.

lib/SPIRV/SPIRVWriter.cpp
```diff
@@ -564,12 +564,14 @@ SPIRVType *LLVMToSPIRVBase::transSPIRVJointMatrixINTELType(
...
-  SPIRVValue *Columns = transConstant(ParseInteger(Postfixes[2]));
-  SPIRVValue *Layout = transConstant(ParseInteger(Postfixes[3]));
-  SPIRVValue *Scope = transConstant(ParseInteger(Postfixes[4]));
-  return mapType(T, BM->addJointMatrixINTELType(transType(ElemTy), Rows,
-                                                Columns, Layout, Scope));
+  std::vector<SPIRVValue *> Args;
+  Args.emplace_back(transConstant(ParseInteger(Postfixes[1])));
+  Args.emplace_back(transConstant(ParseInteger(Postfixes[2])));
+  Args.emplace_back(transConstant(ParseInteger(Postfixes[3])));
+  Args.emplace_back(transConstant(ParseInteger(Postfixes[4])));
```
pauzinl, line 571: Isn't it better to use loop for this?

pauzinl (approved):

PR #1460, opened 2022-04-01
Add assertation for FenceInst translation

pauzinl:


lib/SPIRV/SPIRVWriter.cpp
```diff
@@ -3727,6 +3727,7 @@ SPIRVValue *LLVMToSPIRVBase::transFenceInst(FenceInst *FI,
   // Treat all llvm.fence instructions as having CrossDevice scope:
   SPIRVValue *RetScope = transConstant(getUInt32(M, ScopeCrossDevice));
   SPIRVValue *Val = transConstant(getUInt32(M, MemorySemantics));
+  assert(RetScope && Val);
```
MrSidims, line 3730: Lets put some message here, like `RetScope and Value are not constants` .

pauzinl: Done

MrSidims (approved):

PR #1461, opened 2022-04-01
Migrate SPIRVReader to new PassManager

svenvh:
There is more preparatory work needed for the SPIRVWriter, but we can already migrate the SPIRVReader.

Co-authored-by: Pedro Olsen Ferreira


MrSidims (approved):
LGTM, thanks!

Issue #1444, opened 2022-03-18
How to deal with opaque pointers in the translator


MrSidims:
for reference: https://llvm.org/docs/OpaquePointers.html#migration-instructions
We have discussed some high-level plan about what to do in the translator. This issue is reflecting this plan leaving the field to discuss the details.

LLVM community is moving towards typeless pointers. There are several things to do in the translator to make sure typeless pointers are translatable:
0. Get rid of LLVM pointer-element-type dependent API in the translator code base, for example `getElementType` itself;
1. Represent typeless pointers somehow. Currently we don't have any extension for this and even if we had - still we would need to translate such LLVM IR to SPIR-V and backwards. As an option we can create pointers to a magically named opaque structure in SPIR-V. Note, that it would require to add extra casts to/from this new type for the pointer users. 

2. During reverse translation there is no good way to differ this opaque structure pointer from the true typeless pointer. So to make sure, that new typesless pointer is being emitted by SPIR-V consumer we probably need to introduce an option that forces _any_ pointer type to become opaque. This option is also enforcing the translator to eliminate casts that were inserted on the previous step (or that were presenting in the LLVM IR before). If the option is not passed during the reverse translation (default behavior at least for the first time), the translator should emit LLVM IR as there are no typeless pointers at all (aka emit pointers to opaque structure).

3. We actually need to have opaque/typeless pointer extension and make it properly working with the mentioned SPIR-V consumer option.

At minimum we would need to translate the following IR snippet, I assume the translation flow (without an extension) should be looking like:
LLVM IR before the translator:
```
define spir_kernel void @test(ptr addrspace(1) %p) {
  store i32 0, ptr %p
  %v = load i64, ptr %p
  ret void
}
```
Generated SPIR-V:
```
5 EntryPoint 6 19 "test"
3 Source 0 0
6 Name 3 "struct.Magic"
4 Name 6 "test"
3 Name 7 "p"
4 Name 8 "entry"
3 Name 18 "v"
3 Name 20 "p"
6 Decorate 6 LinkageAttributes "test" Export
4 TypeInt 9 32 0
4 TypeInt 15 64 0
4 Constant 9 14 0
2 TypeVoid 2
6 TypeOpaque 3 "struct.Magic"
4 TypePointer 4 5 3
4 TypeFunction 5 2 4
4 TypePointer 10 5 9
4 TypePointer 12 8 9
4 TypePointer 16 8 15


5 Function 2 6 0 5
[...]

PR #1463, opened 2022-04-13
Fix sret attribute to be valid LLVM IR

jcranmer-intel:
The sret attribute is supposed to indicate what would be the pointee type of the
parameter (to support the transition to opaque pointers). Somehow, the original
change to this code concluded it was supposed to be a pointer type instead,
however, for the sret attribute, while the byval attribute was correctly
handled.

### DIFF
--- a/lib/SPIRV/SPIRVReader.cpp
+++ b/lib/SPIRV/SPIRVReader.cpp
@@ -2774,7 +2774,7 @@ void SPIRVToLLVM::transFunctionAttrs(SPIRVFunction *BF, Function *F) {
         AttrTy = cast<PointerType>(I->getType())->getPointerElementType();
         break;
       case Attribute::AttrKind::StructRet:
-        AttrTy = I->getType();
+        AttrTy = cast<PointerType>(I->getType())->getPointerElementType();
         break;
       default:
         break; // do nothing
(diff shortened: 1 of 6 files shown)


lib/SPIRV/SPIRVReader.cpp
```diff
@@ -2774,7 +2774,7 @@ void SPIRVToLLVM::transFunctionAttrs(SPIRVFunction *BF, Function *F) {
         AttrTy = cast<PointerType>(I->getType())->getPointerElementType();
         break;
       case Attribute::AttrKind::StructRet:
-        AttrTy = I->getType();
+        AttrTy = cast<PointerType>(I->getType())->getPointerElementType();
```
YetAnotherCompilerEngineer, line 2777: Is it possible not to use `getPointerElementType` API here? As is it goes against the intentions of https://github.com/KhronosGroup/SPIRV-LLVM-Translator/issues/1444

jcranmer-intel: I haven't looked at replacing the uses of `getPointerElementType` in this context yet--the next patch I have is replacing a dozen-ish uses with types scavenged form places like `sret`/`byval`/other parameters, which is where I discovered that the `sret` types were being improperly generated.

I think it would be better to handle the replacement of the `getPointerElementType` here at the same time as the one a few lines above rather than as a component of this patch in particular.

YetAnotherCompilerEngineer: Okay, LGTM me then

YetAnotherCompilerEngineer (approved):

svenvh (approved):

PR #1464, opened 2022-04-19
Remove more uses of getPointerElementType (part of issue #1444).

jcranmer-intel:
This is the easiest tranche of changes: where the pointer element type is
usually retrieved by looking a little further afield for the element type, or
via scavenging from sret/byval parameters.

### DIFF
--- a/lib/SPIRV/OCLToSPIRV.cpp
+++ b/lib/SPIRV/OCLToSPIRV.cpp
@@ -70,6 +70,23 @@ static size_t getOCLCpp11AtomicMaxNumOps(StringRef Name) {
       .Default(0);
 }
 
+static Type *getBlockStructType(Value *Parameter) {
+  // In principle, this information should be passed to us from Clang via
+  // an elementtype attribute. However, said attribute requires that the
+  // function call be an intrinsic, which it is not. Instead, we rely on being
+  // able to trace this to the declaration of a variable: OpenCL C specification
+  // section 6.12.5 should guarantee that we can do this.
+  Value *UnderlyingObject = Parameter->stripPointerCasts();
+  Type *ParamType = nullptr;
+  if (auto *GV = dyn_cast<GlobalValue>(UnderlyingObject))
+    ParamType = GV->getValueType();
+  else if (auto *Alloca = dyn_cast<AllocaInst>(UnderlyingObject))
+    ParamType = Alloca->getAllocatedType();
+  else
+    llvm_unreachable("Blocks in OpenCL C must be traceable to allocation site");
+  return ParamType;
+}
+
 class OCLToSPIRVBase : public InstVisitor<OCLToSPIRVBase> {
 public:
   OCLToSPIRVBase() : M(nullptr), Ctx(nullptr), CLVer(0) {}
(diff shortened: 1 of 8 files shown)


lib/SPIRV/OCLToSPIRV.cpp
```diff
@@ -70,6 +70,23 @@ static size_t getOCLCpp11AtomicMaxNumOps(StringRef Name) {
...
+  // able to trace this to the declaration of a variable: OpenCL C specification
+  // section 6.12.5 should guarantee that we can do this.
+  Value *UnderlyingObject = getUnderlyingObject(Parameter);
+  Type *ParamType = nullptr;
+  if (auto *GV = dyn_cast<GlobalValue>(UnderlyingObject))
+    ParamType = GV->getValueType();
+  else if (auto *Alloca = dyn_cast<AllocaInst>(UnderlyingObject))
+    ParamType = Alloca->getAllocatedType();
+  else
+    assert(false && "Blocks in OpenCL C must be traceable to allocation site");
```
YetAnotherCompilerEngineer, line 86: Suggest to use `llvm_unreachable` here instead of an assert. Or the intetion is to hide the branch in release build? In this case `getBlockStructType` could return nullptr making static analyzer tools unhappy.

lib/SPIRV/OCLToSPIRV.cpp
```diff
@@ -70,6 +70,23 @@ static size_t getOCLCpp11AtomicMaxNumOps(StringRef Name) {
       .Default(0);
 }
 
+static Type *getBlockStructType(Value *Parameter) {
+  // In principle, this information should be passed to us from Clang via
+  // an elementtype attribute. However, said attribute requires that the
+  // function call be an intrinsic, which it is not. Instead, we rely on being
+  // able to trace this to the declaration of a variable: OpenCL C specification
+  // section 6.12.5 should guarantee that we can do this.
+  Value *UnderlyingObject = getUnderlyingObject(Parameter);
```
YetAnotherCompilerEngineer, line 79: `getUnderlyingObject` has default `MaxLookup` parameter of 6. Does it make sense to explicitly pass 0 to make sure, that we have no limitations for number of the instructions to strip off?

YetAnotherCompilerEngineer (approved):
That looks good, thanks! Two minor questions/suggestions though.

lib/SPIRV/OCLToSPIRV.cpp
jcranmer-intel: Will change.

lib/SPIRV/OCLToSPIRV.cpp
jcranmer-intel: I copied `getUnderlyingObject` from the original version of this code, but now that I think about it, `stripPointerCasts` is probably the better function anyways--it will look through any number of 0-offset geps and bitcasts, but will stop if there's a non-zero offset on the process.

svenvh:
LGTM, but waiting for @YetAnotherCompilerEngineer's suggestions to be addressed.

svenvh (approved):

PR #1465, opened 2022-04-19
Replace uses of getPointerElementType in SPIRVReader.

jcranmer-intel:
This should allow for SPIR-V->LLVM conversion when compiling with opaque
pointers enabled, though this has not yet been tested.

### DIFF
--- a/lib/SPIRV/SPIRVReader.cpp
+++ b/lib/SPIRV/SPIRVReader.cpp
@@ -1283,7 +1283,8 @@ void SPIRVToLLVM::addMemAliasMetadata(Instruction *I, SPIRVId AliasListId,
   I->setMetadata(AliasMDKind, MDAliasListMap[AliasListId]);
 }
 
-void transFunctionPointerCallArgumentAttributes(SPIRVValue *BV, CallInst *CI) {
+void SPIRVToLLVM::transFunctionPointerCallArgumentAttributes(
+    SPIRVValue *BV, CallInst *CI, SPIRVTypeFunction *CalledFnTy) {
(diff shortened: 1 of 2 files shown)


YetAnotherCompilerEngineer (approved):

svenvh (commented):
> This should allow for SPIR-V->LLVM conversion when compiling with opaque
> pointers enabled, though this has not yet been tested.

I'm a bit confused about the scope of this PR; you're only removing a few `getPointerElementType` calls while many remain.  Those remaining calls will assert when compiling with opaque pointers enabled.

If this is prework for further changes that's fine of course, but please clarify the scope in the commit message.

jcranmer-intel:
> I'm a bit confused about the scope of this PR; you're only removing a few `getPointerElementType` calls while many remain. Those remaining calls will assert when compiling with opaque pointers enabled.
> 
> If this is prework for further changes that's fine of course, but please clarify the scope in the commit message.

I'm spreading the work of removing `getPointerElementType` among several PRs. This PR by itself fixes only those uses in `SPIRVReader` (which turned out to be much fewer than I was expecting--unfortunately `SPIRVType::getPointerElementType` does obfuscate a little bit how many calls to `llvm::Type::getPointerElementType` remain to be converted) and does not seek to fix anything else--those are being handled in other PRs, including #1464 and more I have yet to make.

At this stage, I'm measuring progress solely by how many `getPointerElementType` calls remain; I haven't tried running anything with `-opaque-pointers` to see how things get or don't get. I'll therefore delete this claim to avoid confusion about how much is actually expected to work with this PR enabled.

svenvh (approved):

PR #1466, opened 2022-04-20
.clang-tidy: allow recursion

svenvh:
The code already uses recursion in various places and this check fails
whenever a change is made in the recursion chain.


YetAnotherCompilerEngineer (approved):
Yeah, we do have recursions with memoization as a design concept for some translation routes (and some of them impossible to rewrite via iteration others not worth it).

PR #1470, opened 2022-04-22
Remove more getPointerElementType queries.

jcranmer-intel:
Most of this tranche of calls were being used to disambiguate between different
built-ins that share the same or similar names.

The exception is the change in visitCallSPIRVPipeBuiltin, which is a convoluted
way of adding a cast to i8 addrspace(4)* if not already i8 addrspace(4)*, so it
is instead rewritten to check if a cast needs to be added without querying
getPointerElementType.


YetAnotherCompilerEngineer (approved):
LGTM

svenvh (approved):

PR #1471, opened 2022-05-04
Make the LowerBitCast pass support opaque pointers.

jcranmer-intel:
This is a relatively large change, as the original pass relied on being able to
track the initial bitcast <3 x i64>* to <6 x i32>* to know where to start
rewriting. Instead, this patch starts at the final invalid extractelement call
and works its way backwards as far as necessary to generate correct code.

### DIFF
--- a/lib/SPIRV/SPIRVLowerBitCastToNonStandardType.cpp
+++ b/lib/SPIRV/SPIRVLowerBitCastToNonStandardType.cpp
@@ -40,104 +40,83 @@
 // point types, 2/3/4/8/16-element vector of scalar types").
 //
 //===----------------------------------------------------------------------===//
-#define DEBUG_TYPE "spv-lower-bitcast-to-nonstandard-type"
-
 #include "SPIRVInternal.h"
 
 #include "llvm/IR/IRBuilder.h"
+#include "llvm/IR/NoFolder.h"
 #include "llvm/IR/PassManager.h"
 #include "llvm/Pass.h"
+#include "llvm/Transforms/Utils/Local.h"
 
 #include <utility>
+
+#define DEBUG_TYPE "spv-lower-bitcast-to-nonstandard-type"
 
 using namespace llvm;
 
 namespace SPIRV {
 
-static VectorType *getVectorType(Type *Ty) {
-  assert(Ty != nullptr && "Expected non-null type");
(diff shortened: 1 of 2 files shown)


test/lower-non-standard-types.ll
```diff
@@ -24,8 +24,8 @@ define dso_local spir_func void @vmult2() local_unnamed_addr #0 !sycl_explicit_s
 entry:
   %0 = load <6 x i32>, <6 x i32> addrspace(4)* addrspacecast (<6 x i32> addrspace(1)* bitcast (<3 x i64> addrspace(1)* @Id to <6 x i32> addrspace(1)*) to <6 x i32> addrspace(4)*), align 32
   %1 = load <6 x i32>, <6 x i32> addrspace(4)* addrspacecast (<6 x i32> addrspace(1)* bitcast (<3 x i64> addrspace(1)* @Id to <6 x i32> addrspace(1)*) to <6 x i32> addrspace(4)*), align 32
-  %2 = extractelement <6 x i32> %0, i32 1
-  %3 = extractelement <6 x i32> %1, i32 4
+  %2 = extractelement <6 x i32> %0, i32 0
+  %3 = extractelement <6 x i32> %1, i32 5
```
YetAnotherCompilerEngineer, line 28: I'd assume opaque pointers-related work at the current stage to be NFC. This change doesn't look NFC. Do you have insights on why the extracted index was changed?

jcranmer-intel: The original code was, from what I can figure, outright incorrect. The InstCombine transformation in question is here: https://github.com/llvm/llvm-project/blob/588ffdaf376a191ddcf106c29462f156fa8bdce5/llvm/lib/Transforms/InstCombine/InstCombineCasts.cpp#L922-L951. As the SPIR-V datalayout string is little-endian, this means that index 0 of `<6 x i32>` should be mapped to the low bits of index 0 of `<3 x i64>`, not the high bits.

So in this case, the change is intentionally NFC, but it should fix a bug in the existing conversion anyways.

svenvh (approved):

test/lower-non-standard-types.ll
YetAnotherCompilerEngineer: Got it, thanks!

YetAnotherCompilerEngineer (approved):

PR #1472, opened 2022-05-04
Remove a few more getPointerElementType() calls.

jcranmer-intel:
The change in getScalarOrArrayConstantInt() is not actually exercised by the
current test suite, and it may be impossible to exercise it given that the
caller of the function relies on the parameter being a GEP of array type.
However, this change should correctly deduce the type were it a correct call.

The call in SPIRVRegularizeLLVM is of course unnecessary with opaque pointers:
addrspacecast can never change the pointee type of a pointer type in such
scenarios.


lib/SPIRV/SPIRVUtil.cpp
```diff
@@ -1254,8 +1254,16 @@ Value *getScalarOrArrayConstantInt(Instruction *Pos, Type *T, unsigned Len,
     assert(Len == 1 && "Invalid length");
     return ConstantInt::get(IT, V, IsSigned);
   }
-  if (auto PT = dyn_cast<PointerType>(T)) {
-    auto ET = PT->getPointerElementType();
+  if (isa<PointerType>(T)) {
+    unsigned PointerSize = Pos->getParent()
+                               ->getParent()
+                               ->getParent()
```
YetAnotherCompilerEngineer, line 1260:
```suggestion
    unsigned PointerSize = Pos->getModule()
```

lib/SPIRV/SPIRVUtil.cpp
```diff
@@ -1254,8 +1254,16 @@ Value *getScalarOrArrayConstantInt(Instruction *Pos, Type *T, unsigned Len,
...
+                               ->getDataLayout()
+                               .getPointerTypeSizeInBits(T);
+    auto *ET = Type::getIntNTy(T->getContext(), PointerSize);
+    assert(
+        (T->isOpaquePointerTy() || T->getNonOpaquePointerElementType() == ET) &&
+        "Only call this method with pointer-to-size_t arguments");
```
YetAnotherCompilerEngineer, line 1266: Technically, not the most correct assert message, since the very first line of the function is about processing integers.
`if (auto IT = dyn_cast<IntegerType>(T)) {`

lib/SPIRV/SPIRVRegularizeLLVM.cpp
```diff
@@ -640,8 +640,10 @@ bool SPIRVRegularizeLLVMBase::regularize() {
         if (auto *ASCast = dyn_cast<AddrSpaceCastInst>(&II)) {
           Type *DestTy = ASCast->getDestTy();
           Type *SrcTy = ASCast->getSrcTy();
-          if (DestTy->getPointerElementType() !=
-              SrcTy->getPointerElementType()) {
+          if (!II.getContext().supportsTypedPointers())
+            continue;
+          if (DestTy->getNonOpaquePointerElementType() !=
+              SrcTy->getNonOpaquePointerElementType()) {
```
nikic, line 646: `if (!DestTy->hasSameElementTypeAs(SrcTy))` should avoid explicitly checking for typed pointers here.

lib/SPIRV/SPIRVUtil.cpp
```diff
@@ -1254,8 +1254,13 @@ Value *getScalarOrArrayConstantInt(Instruction *Pos, Type *T, unsigned Len,
...
     return ConstantInt::get(IT, V, IsSigned);
   }
-  if (auto PT = dyn_cast<PointerType>(T)) {
-    auto ET = PT->getPointerElementType();
+  if (isa<PointerType>(T)) {
+    unsigned PointerSize =
+        Pos->getModule()->getDataLayout().getPointerTypeSizeInBits(T);
+    auto *ET = Type::getIntNTy(T->getContext(), PointerSize);
+    assert(
+        (T->isOpaquePointerTy() || T->getNonOpaquePointerElementType() == ET) &&
```
nikic, line 1262: This is usually written as `T->isOpaqueOrPointeeTypeMatches(ET)`.

YetAnotherCompilerEngineer:
Btw, there is no need to force push changes in this repository - github allows to merge PRs with squash and merge strategy.

lib/SPIRV/SPIRVRegularizeLLVM.cpp
svenvh: @jcranmer-intel this comment hasn't been addressed yet, other than that this PR looks good to me.

svenvh (approved):

PR #1476, opened 2022-05-09
Remove LowerSPIRBlocks pass, as it appears to be obsolete.

jcranmer-intel:
This is nominally part of the opaque-pointer work.

As mentioned in https://github.com/KhronosGroup/SPIRV-LLVM-Translator/issues/1444#issuecomment-1119093601, this isn't actually exercised by the test suite, and I've locally been unable to put together a pipeline to cause this code to be generated. If this *isn't* actually obsolete, I'd be happy to actually fix the code to use opaque pointers, but I'd need a test case to make sure the change is correct in order to do that, and so far, I haven't found one.


svenvh (approved):
`test/global_block.ll` seems to have been one of the tests that was exercising this pass, but that test has been rewritten by commit 7fb1a7ba ("Fix translation of device execution built-ins", 2018-05-30). Since then, the test suite mirrors the IR that clang produces instead of following SPIR 2.0 for blocks.

So with this patch we're effectively removing SPIR 2.0 blocks support, which has been untested for almost 4 years now.  I'm okay with the removal, but I'll hold off merging this for a few more days in case there are any objections from anyone else.

YetAnotherCompilerEngineer (approved):

PR #1487, opened 2022-05-31
Disable test after LLVM change

svenvh:
An LLVM change in the range 1f06398e96d4..80b3dcc045f8 caused this
test to fail.  XFAIL the test while investigating.


svenvh:
```
FAIL: LLVM_SPIRV :: lower-non-standard-vec-with-ext.ll (542 of 631)
Exit Code: 1

Command Output (stderr):
--
LLVM ERROR: Unsupported vector type with 20 elements

--
```

@jcranmer-intel the failure might be somewhat related to your rewrite done in #1471 (though obviously not caused by it, since it only started failing recently after an LLVM commit).  Would you be able to have a quick look?

jcranmer-intel:
See #1490 for a fix.

PR #1490, opened 2022-05-31
Don't use not --crash to check for report_fatal_error messages.

jcranmer-intel:
In https://reviews.llvm.org/D126550, report_fatal_error was changed to not use
the abort function, which means it no longer registers as a crash by the
definition of not --crash.

### DIFF
--- a/test/lower-non-standard-vec-with-ext.ll
+++ b/test/lower-non-standard-vec-with-ext.ll
@@ -1,9 +1,6 @@
 ; RUN: llvm-as %s -o %t.bc
-; RUN: not --crash llvm-spirv -s %t.bc
+; RUN: not llvm-spirv -s %t.bc
 ; RUN: llvm-spirv --spirv-ext=+SPV_INTEL_vector_compute -s %t.bc
-
-; Temporarily disable test as it fails after an LLVM change.
-; XFAIL: *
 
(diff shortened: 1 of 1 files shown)


jcranmer-intel:
Oh, that's dandy.

`not --crash` will fail if it has D126550; solo `not` will fail if it doesn't have it, and there's no way to let it pass whether or not it crashes.

The other way to fix this is to use the second argument to `report_fatal_error` to generate crash diagnostics, which causes it to *always* crash (even in builds with D126550), but this doesn't feel like the right way to use `report_fatal_error` here.

Thoughts?

svenvh:
Thanks for the fix, I didn't realize that this test was deliberately crashing.  Your fix makes sense.

There was a fixup commit to D126550, which is `8c55de9ee7f6 ("fix tests after my commit 80b3dcc045f8ea6e5e532d8891bbf1305bce89e8  doesn't like exit code 126 I'm afraid", 2022-05-30)`.  The prebuilt llvm binaries used in the out-of-tree tests have D126550 but not the fixup commit; I suspect that is causing the failures.  The apt.llvm.org nightly binaries are currently stuck at 30 May; the next build should hopefully have both commits and then we should be able to merge this PR.

PR #1499, opened 2022-06-10
Prepare SPIRVWriter for type conversion without opaque pointers.

jcranmer-intel:
This changeset is, by itself, not yet enough to get most SPIR-V files to be
emitted when the input module is in opaque pointer mode. However, this does
remove all of the calls to `getPointerElementType` that SPIRVWriter makes
(directly or indirectly), except for the ones that directly correspond to
translating a pointer type.

A later changeset will add a type scavenger that will be used to find the
pointee type of a pointer. All calls to `getPointerElementType` that remain
after this one will be instead shifted to query the type scavenger instead. To
facilitate this change, several methods are added to avoid querying pointer
element types, and they have been added in several places where their need is
known.

The most basic of basic kernels, those that do not use pointer types (other than
declaring global values and functions) will work in opaque pointer mode with
this changeset.

### DIFF
--- a/lib/SPIRV/OCLToSPIRV.cpp
+++ b/lib/SPIRV/OCLToSPIRV.cpp
@@ -976,13 +976,18 @@ void OCLToSPIRVBase::visitCallReadImageWithSampler(CallInst *CI,
                                                    StringRef MangledName) {
   assert(MangledName.find(kMangledName::Sampler) != StringRef::npos);
   assert(CI->getCalledFunction() && "Unexpected indirect call");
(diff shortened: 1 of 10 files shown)


svenvh (commented):
Great progress!  I'll try and have a more detailed look soon.  Please don't forget to address the issues reported by the automated checks.

svenvh (approved):
LGTM.  I'll merge this early next week if nobody objects.

YetAnotherCompilerEngineer (approved):

PR #1498, opened 2022-06-10
[Backport][110] Translate atomic_fetch_sub to OpAtomicFAddEXT with ne…

haonanya:
…gative value operand

Backport for https://github.com/KhronosGroup/SPIRV-LLVM-Translator/pull/1492

Signed-off-by: haonanya <haonan.yang@intel.com>

PR #1492, opened 2022-06-07
Translate atomic_fetch_sub to OpAtomicFAddEXT with negative value ope…

haonanya:
…rand

Signed-off-by: Haonan Yang <haonan.yang@intel.com>

PR #1505, opened 2022-06-16
Translate atomic_fetch_sub to OpAtomicFAddEXT with negative value ope…

haonanya:
…rand (#1498)

Backport for https://github.com/KhronosGroup/SPIRV-LLVM-Translator/pull/1492

Signed-off-by: haonanya <haonan.yang@intel.com>

PR #1506, opened 2022-06-16
Translate atomic_fetch_sub to OpAtomicFAddEXT with negative value ope…

haonanya:
…rand (#1498)

Backport for https://github.com/KhronosGroup/SPIRV-LLVM-Translator/pull/1492

Signed-off-by: haonanya <haonan.yang@intel.com>


haonanya:
This is same Travis CI environment issue as https://github.com/KhronosGroup/SPIRV-LLVM-Translator/pull/1505, and Github Actions In-tree build has failures when install dependences which should also be an env issue. 
Hi, @svenvh, could you please have a look? Thanks very much.

svenvh (approved):

PR #1510, opened 2022-06-20
Fix build by reverting a ValueMap back to std::map

svenvh:
Commit 4a9c78ee ("Prepare SPIRVWriter for type conversion without
opaque pointers. (#1499)", 2022-06-20) changed
`OCLTypeToSPIRVBase::AdaptedTy` to a `ValueMap`, but that made
`OCLTypeToSPIRVBase` non-movable and non-copyable.  Landing eb8e9adc
("Migrate SPIRVWriter to new PassManager", 2022-06-20) on top caused a
build breakage as it requires `OCLTypeToSPIRVPass` to be copyable.

cc: @jcranmer-intel

### DIFF
--- a/lib/SPIRV/OCLTypeToSPIRV.h
+++ b/lib/SPIRV/OCLTypeToSPIRV.h
@@ -46,7 +46,6 @@
 #include "llvm/IR/Function.h"
 #include "llvm/IR/LLVMContext.h"
 #include "llvm/IR/PassManager.h"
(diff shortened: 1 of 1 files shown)


svenvh:
Merging this right away to fix the build.

@jcranmer-intel if you need the `ValueMap` back for your next patch in the series, it would be best to do that in a separate change I suppose.

PR #1512, opened 2022-06-21
Scavenge type parameters for use in translating opaque pointers.

jcranmer-intel:
This change is incomplete, and will not properly convert every single LLVM IR file to SPIR-V. However, it will convert many files. Areas that are known to require more investigation:
* ptr-valued extractvalue, especially from cmpxchg instructions
* return values of functions
* llvm.*.annotation support
* function pointer extensions
* pointer types in structs


lib/SPIRV/SPIRVTypeScavenger.h
```diff
@@ -0,0 +1,129 @@
...
+  void deduceDeclarationTypes(Function &F);
+
+  /// This assigns known pointer element types for parameters of LLVM
+  /// intrinsics.
+  void typeIntrinsicArgs(Function &F, Intrinsic::ID Id);
+
+public:
+  /// Compute pointer element types for all pertinent values in the module. This
+  /// method must be called before any other method on this class.
+  void typeModule(Module &M);
```
svenvh, line 109: Perhaps make `LLVMToSPIRVBase::Scavenger` an `std::unique_ptr` and pass its constructor an `llvm::Module`.  Then we won't need to be careful about calling this initialization method first, it will happen by construction.

lib/SPIRV/SPIRVTypeScavenger.cpp
```diff
@@ -0,0 +1,561 @@
...
+#include "llvm/IR/IntrinsicInst.h"
+#include "llvm/IR/IRBuilder.h"
+#include "llvm/IR/NoFolder.h"
+#include "llvm/IR/Operator.h"
+
+#define DEBUG_TYPE "type-scavenger"
+
+using namespace llvm;
+
+static Type *getPointerUseType(Function *F, Op Opcode, unsigned ArgNo);
```
svenvh, line 54: Remove this line?

lib/SPIRV/SPIRVTypeScavenger.h
```diff
@@ -0,0 +1,129 @@
...
+      return OS << "deferred type for " << *AsDeferred->Values[0];
+    else if (auto *AsValue = dyn_cast<Value *>(Ty))
+      return OS << "points to " << *AsValue;
+    else
+      return OS;
+  }
+
+  /// Compute the pointer element type of a value, solely based on its
+  /// definition. The value must be pointer-valued.
+  DeducedType getTypeInternal(Value *V);
```
svenvh, line 86: Maybe clarify the name:

```suggestion
  DeducedType computePointerElementType(Value *V);
```

lib/SPIRV/SPIRVTypeScavenger.h
```diff
@@ -0,0 +1,129 @@
...
+  /// Enforce that the pointer element types of all operands of the instruction
+  /// matches the type that the instruction itself requires. If a pointer
+  /// element type of one of the operands is deferred, this will type the use
+  /// correctly.
+  void correctUseTypes(Instruction &I);
+
+  /// This assigns known pointer element types for the parameters of a function.
+  /// This method should be called for all functions before doing any type
+  /// analysis on the module.
+  void deduceDeclarationTypes(Function &F);
```
svenvh, line 100:
```suggestion
  void deduceFunctionArgTypes(Function &F);
```

lib/SPIRV/SPIRVTypeScavenger.h
```diff
@@ -0,0 +1,129 @@
...
+
+  /// This assigns known pointer element types for parameters of LLVM
+  /// intrinsics.
+  void typeIntrinsicArgs(Function &F, Intrinsic::ID Id);
```
svenvh, line 104:
```suggestion
  void deduceIntrinsicTypes(Function &F, Intrinsic::ID Id);
```
as it also populates `DeducedTypes`

svenvh (commented):
Great progress!  Just a few high-level comments for now, as I haven't gone into the details yet.

lib/SPIRV/SPIRVTypeScavenger.cpp
```diff
@@ -0,0 +1,561 @@
...
+// Software is furnished to do so, subject to the following conditions:
+//
+// Redistributions of source code must retain the above copyright notice,
+// this list of conditions and the following disclaimers.
+// Redistributions in binary form must reproduce the above copyright notice,
+// this list of conditions and the following disclaimers in the documentation
+// and/or other materials provided with the distribution.
+// Neither the names of Intel Corporation, nor the names of its
+// contributors may be used to endorse or promote products derived from this
+// Software without specific prior written permission.
```
YetAnotherCompilerEngineer, line 24: Not sure, whether that is a correct license for the translator.

lib/SPIRV/SPIRVTypeScavenger.cpp
```diff
@@ -0,0 +1,561 @@
...
+  default:
+    return nullptr;
+  }
+}
+
+void SPIRVTypeScavenger::typeIntrinsicArgs(Function &F, Intrinsic::ID Id) {
+  static constexpr unsigned Return = ~0U;
+  auto AddParameter = [&](unsigned ArgNo, DeducedType Ty) {
+    if (ArgNo == Return) {
+    } else {
```
YetAnotherCompilerEngineer, line 125:
```suggestion
    if (ArgNo != Return) {
```

lib/SPIRV/SPIRVTypeScavenger.cpp
```diff
@@ -0,0 +1,561 @@
...
+      Argument *Arg = F.getArg(ArgNo);
+      LLVM_DEBUG(dbgs() << "Parameter " << *Arg << " of " << F.getName() << " has type " << Ty << "\n");
+      DeducedTypes[Arg] = Ty;
+    }
+  };
+  LLVMContext &Ctx = F.getContext();
+
+  switch (Id) {
+  // TODO: handle case Intrinsic::memset
```
YetAnotherCompilerEngineer, line 134: If I understand correctly, at this point we no longer have memsets in IR since they were transformed to memcpys

lib/SPIRV/SPIRVTypeScavenger.cpp
jcranmer-intel: I've noticed there are a few different license block headers in this repository, and I wasn't sure which one was the proper one to use the base. If you could point to a better one, I would happily use it.

lib/SPIRV/SPIRVTypeScavenger.cpp
jcranmer-intel: I was using the list of intrinsics in SPIRVWriter's `transIntrinsicInst` as the base, which does include `memset` as a handled intrinsic.

lib/SPIRV/SPIRVTypeScavenger.h
jcranmer-intel: I'm aware I didn't address this in my push--I need to do a touch more research before I commit to it.

lib/SPIRV/SPIRVTypeScavenger.cpp
svenvh: I think that eventually we want to move to the Khronos header; you can see an example [here](https://github.com/KhronosGroup/SPIRV-LLVM-Translator/pull/1483/files#diff-1efbc99e15b75571adb74b27faa47bbb4caf823309bb64a97888c6c8b1859a70)

lib/SPIRV/SPIRVTypeScavenger.cpp
YetAnotherCompilerEngineer: Yeah, I mean there is no need to handle memset, since the Regularize LLVM pass has already replaced memsets with memcpys and hence the comment can be removed :)

svenvh (approved):
LGTM, but I'll leave it open for a few more days in case @YetAnotherCompilerEngineer has any more feedback.

MrSidims (approved):
LGTM, thanks for taking care of the issue!

Issue #1509, opened 2022-06-17
Use exact SPIR-V version from the target triple


AnastasiaStulova:
Clang now permits setting the exact SPIR-V version via target triple https://llvm.org/docs/SPIRVUsage.html#target-triples. It would be good to teach the translator to respect the version set in the triple.

PR #1516, opened 2022-06-24
Use SPIR-V subarch version if present in triple

svenvh:
~We cannot emit an exact version yet, so at least warn when the user
provides an LLVM module with an exact SPIR-V version specified.~

Ensure that the SPIR-V version specified as the subarch is taken into account by llvm-spirv.

### DIFF
--- a/lib/SPIRV/SPIRVWriter.cpp
+++ b/lib/SPIRV/SPIRVWriter.cpp
@@ -5176,6 +5176,24 @@ bool isValidLLVMModule(Module *M, SPIRVErrorLog &ErrorLog) {
 
 namespace {
 
+VersionNumber getVersionFromTriple(const Triple &TT, SPIRVErrorLog &ErrorLog) {
+  switch (TT.getSubArch()) {
+  case Triple::SPIRVSubArch_v10:
+    return VersionNumber::SPIRV_1_0;
+  case Triple::SPIRVSubArch_v11:
+    return VersionNumber::SPIRV_1_1;
+  case Triple::SPIRVSubArch_v12:
+    return VersionNumber::SPIRV_1_2;
+  case Triple::SPIRVSubArch_v13:
+    return VersionNumber::SPIRV_1_3;
+  case Triple::SPIRVSubArch_v14:
+    return VersionNumber::SPIRV_1_4;
+  default:
+    ErrorLog.checkError(false, SPIRVEC_InvalidSubArch, TT.getArchName().str());
+    return VersionNumber::MaximumVersion;
+  }
+}
+
 bool runSpirvWriterPasses(Module *M, std::ostream *OS, std::string &ErrMsg,
                           const SPIRV::TranslatorOpts &Opts) {
   // Perform the conversion and write the resulting SPIR-V if an ostream has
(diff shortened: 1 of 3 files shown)


lib/SPIRV/SPIRVWriter.cpp
```diff
@@ -5182,6 +5182,14 @@ bool runSpirvWriterPasses(Module *M, std::ostream *OS, std::string &ErrMsg,
   // been given; otherwise only perform regularization.
   bool WriteSpirv = OS != nullptr;
 
+  Triple TargetTriple(M->getTargetTriple());
+  if ((TargetTriple.getArch() == Triple::spirv32 ||
+       TargetTriple.getArch() == Triple::spirv64) &&
+      TargetTriple.getSubArch() != Triple::NoSubArch) {
+    // We cannot set the exact SPIR-V version the translator will emit, so warn.
+    errs() << "warning: SPIR-V version in target triple will be ignored\n";
```
AnastasiaStulova, line 5190: Ok, I wonder though if we could also recommend to use --max-version in this diagnostic? I assume this could be some sort of the workaround?

svenvh: Sure, I can add an extra line.

AnastasiaStulova (approved):
LGTM! Thanks

YetAnotherCompilerEngineer:
What is blocking the translator to emit the exact version now? From what I see `SubArchType` enum already contains SPIRV version values (at least in LangRef) and hence we can call `setMinSPIRVVersion` here with the appropriate parameter.

svenvh:
> What is blocking the translator to emit the exact version now?

I didn't realize we essentially have all the functionality to do so, thanks for the suggestion!  I have reworked the patch.

YetAnotherCompilerEngineer:
> I didn't realize we essentially have all the functionality to do so, thanks for the suggestion! I have reworked the patch.

Thanks! One thought for an enhancement. We have several `setMinSPIRVVersion` in the forward translator depending on the instructions found in the module. I believe, that if LLVM IR was compiled with one SPIR-V version (which is reflected in the tripple) and then some check raises the version - the translator should emit a warning. But guess, it can be done later.

lib/SPIRV/SPIRVWriter.cpp
```diff
@@ -5176,6 +5176,24 @@ bool isValidLLVMModule(Module *M, SPIRVErrorLog &ErrorLog) {
...
+  case Triple::SPIRVSubArch_v11:
+    return VersionNumber::SPIRV_1_1;
+  case Triple::SPIRVSubArch_v12:
+    return VersionNumber::SPIRV_1_2;
+  case Triple::SPIRVSubArch_v13:
+    return VersionNumber::SPIRV_1_3;
+  case Triple::SPIRVSubArch_v14:
+    return VersionNumber::SPIRV_1_4;
+  default:
+    ErrorLog.checkError(false, SPIRVEC_InvalidSubArch, TT.getArchName().str());
```
YetAnotherCompilerEngineer, line 5192: This diagnostic is not tested

YetAnotherCompilerEngineer (commented):
LGTM with one nit

svenvh:
> We have several `setMinSPIRVVersion` in the forward translator depending on the instructions found in the module. I believe, that if LLVM IR was compiled with one SPIR-V version (which is reflected in the tripple) and then some check raises the version - the translator should emit a warning.

Good observation!  I was also not convinced that this would already be working perfectly in all cases, but haven't had the time to write extensive tests.  I've added this point to #1509 and will leave that issue open after merging this PR.

YetAnotherCompilerEngineer (approved):

PR #1517, opened 2022-06-24
Modify ptr annotation handling to emit both memory attribute and memory access annotations for user strings

asudarsa:


steffenlarsen (approved):

svenvh:
I suppose this is a functional change, so it would be good to extend a test case to check for the additional annotations.

asudarsa:
This change needs more extensive testing. Converting to a draft till then.

PR #1518, opened 2022-06-25
Support multiple annotations to single variable

asudarsa:

### DIFF
--- a/lib/SPIRV/SPIRVWriter.cpp
+++ b/lib/SPIRV/SPIRVWriter.cpp
@@ -2935,11 +2935,6 @@ getBankBitsFromStrings(const std::vector<std::string> &BitsStrings) {
 void addAnnotationDecorations(SPIRVEntry *E, DecorationsInfoVec &Decorations) {
   SPIRVModule *M = E->getModule();
   for (const auto &I : Decorations) {
-    // Such decoration already exists on a type, skip it
-    if (E->hasDecorate(I.first, /*Index=*/0, /*Result=*/nullptr)) {
-      continue;
-    }
-
     switch (I.first) {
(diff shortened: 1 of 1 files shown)


lib/SPIRV/SPIRVWriter.cpp
```diff
@@ -2935,11 +2935,6 @@ getBankBitsFromStrings(const std::vector<std::string> &BitsStrings) {
 void addAnnotationDecorations(SPIRVEntry *E, DecorationsInfoVec &Decorations) {
   SPIRVModule *M = E->getModule();
   for (const auto &I : Decorations) {
-    // Such decoration already exists on a type, skip it
```
asudarsa, line 2938: A type (or a pointer variable) can have multiple annotations and this needs to be preserved during the LLVM to SPIR-V translation stage.

YetAnotherCompilerEngineer:
Please hold with a merge of this patch (unless it's urgent). If I remember correctly it might cause several decorations with incompatible arguments to appear, but I need to double check it.

asudarsa:
Hi @YetAnotherCompilerEngineer 

Thanks for the comment. The user in this case wanted to add multiple annotations each of which will be lowered toUserSemantic decoration. I think we should atleast allow that.
Example:
struct MyIP {
[[clang::annotate("abc",1,2,3)]]
[[clang::annotate("xyz",5,6,7)]] 
int *A;
...
}
In this case, we can generate two UserSemantic Decorations.

Please let m ekow if that is agreeable.

Thanks

asudarsa:
Modified the implementation to allow multiple decorations only for UserSemantics. Will add a test now. Thanks

PR #1103, opened 2021-07-06
Add support for SPV_INTEL_bfloat16_conversion extension

mlychkov:
This extension provides instructions to convert single-precision 32-bit
floating-point value to bfloat16 format and vice versa.
It doesn't introduce bfloat16 type in SPIR-V, instead instructions below
use 16-bit integer type whose bit pattern represents (bitcasted from)
a bfloat16 value.

Spec: https://github.com/intel/llvm/blob/f587bbfb8a742bd55c80b12e01505ed085a0748f/sycl/doc/extensions/SPIRV/SPV_INTEL_bf16_convert.asciidoc

Signed-off-by: Mikhail Lychkov <mikhail.lychkov@intel.com>

PR #1521, opened 2022-06-27
 Add support for SPV_INTEL_bfloat16_conversion extension

haonanya:
This backports https://github.com/KhronosGroup/SPIRV-LLVM-Translator/pull/1103
This extension provides instructions to convert single-precision 32-bit
floating-point value to bfloat16 format and vice versa.
It doesn't introduce bfloat16 type in SPIR-V, instead instructions below
use 16-bit integer type whose bit pattern represents (bitcasted from)
a bfloat16 value.

Spec: https://github.com/intel/llvm/blob/f587bbfb8a742bd55c80b12e01505ed085a0748f/sycl/doc/extensions/SPIRV/SPV_INTEL_bf16_convert.asciidoc

Signed-off-by: haonanya <haonan.yang@intel.com>


haonanya:
Hi, @svenvh , could you please have a look? Thanks very much.

svenvh:
Any reason why you squashed the 3 commits into one?  Ideally we should keep backported commits as separate commits (though they can be combined in one pull request of course).

haonanya:
Some negative tests with .spt format are not added. There are some failures: Invalid SPIR-V module: invalid magic number. It seems that they are not added directly.

haonanya:
Hi, @svenvh , could you please have a look? Thanks very much.

svenvh (approved):

PR #1524, opened 2022-06-27
Implement support for @llvm.sadd.sat intrinsic.

jcranmer-intel:


lib/SPIRV/SPIRVLowerSaddWithOverflow.cpp
```diff
@@ -46,8 +46,8 @@
 #include "SPIRVError.h"
 #include "libSPIRV/SPIRVDebug.h"
 
+#include "llvm/IR/IRBuilder.h"
 #include "llvm/IR/InstVisitor.h"
```
svenvh, line 50: We can drop this include too I guess?

lib/SPIRV/SPIRVLowerSaddWithOverflow.h
```diff
@@ -35,22 +35,21 @@
 #ifndef SPIRV_SPIRVLOWERSADDWITHOVERFLOW_H
 #define SPIRV_SPIRVLOWERSADDWITHOVERFLOW_H
 
-#include "llvm/IR/InstVisitor.h"
 #include "llvm/IR/PassManager.h"
 #include "llvm/Pass.h"
 
 namespace SPIRV {
 
-class SPIRVLowerSaddWithOverflowBase
```
svenvh, line 44: I guess we should drop the "WithOverflow" part from the class name now (and maybe rename the file too).  But perhaps it would be better to do that in a followup commit/PR.

lib/SPIRV/SPIRVLowerSaddWithOverflow.cpp
```diff
@@ -46,8 +46,8 @@
...
-#include "llvm/IR/IntrinsicInst.h"
```
svenvh, line 50: Actually did you intend to keep this include and remove the one above?

lib/SPIRV/SPIRVLowerSaddWithOverflow.cpp
```diff
@@ -81,39 +77,85 @@ void SPIRVLowerSaddWithOverflowBase::visitIntrinsicInst(CallInst &I) {
...
+    Value *Result = Builder.CreateSelect(Overflow, OverflowedRes, Sum);
+    II->replaceAllUsesWith(Result);
+    II->eraseFromParent();
+  }
+
+  // Now replace the sadd_with_overflow intrinsic itself.
+  replaceSaddOverflow(*SaddO);
 }
 
 bool SPIRVLowerSaddWithOverflowBase::runLowerSaddWithOverflow(Module &M) {
```
asudarsa, line 149: Hi @jcranmer-intel 
Is there a reason why we cannot create a new set of llvm_sadd_sat_* functions into the loaded module? And then use that function here? Do you expect that the size of the loaded module might explode?
Thanks

lib/SPIRV/SPIRVLowerSaddWithOverflow.cpp
```diff
@@ -81,39 +77,85 @@ void SPIRVLowerSaddWithOverflowBase::visitIntrinsicInst(CallInst &I) {
...
+  // Replace all uses of the intrinsic with equivalent code relying on
+  // sadd_with_overflow
+  IRBuilder<> Builder(F.getContext());
+  unsigned BitWidth = IntTy->getIntegerBitWidth();
+  Value *IntMin = Builder.getInt(APInt::getSignedMinValue(BitWidth));
+  Value *ShiftWidth = Builder.getIntN(BitWidth, BitWidth - 1);
+
+  for (IntrinsicInst *II : Intrinsics) {
+    Builder.SetInsertPoint(II);
+    // sadd_sat(a, b) => overflow ? (res >> bitwidth) ^ intmin : res;
```
asudarsa, line 133: Can you please add a line to show where res and overflow are coming from? Thanks

lib/SPIRV/SPIRVLowerSaddWithOverflow.cpp
```diff
@@ -81,39 +77,85 @@ void SPIRVLowerSaddWithOverflowBase::visitIntrinsicInst(CallInst &I) {
...
+
+  // Now replace the sadd_with_overflow intrinsic itself.
+  replaceSaddOverflow(*SaddO);
 }
 
 bool SPIRVLowerSaddWithOverflowBase::runLowerSaddWithOverflow(Module &M) {
   Context = &M.getContext();
   Mod = &M;
-  visit(M);
+  for (Function &F : M) {
```
asudarsa, line 152: Can you please specify why you adopted this way rather than calling replaceSaddOverflow or replaceSaddSat inside visitIntrinsicInst? Thanks

lib/SPIRV/SPIRVLowerSaddWithOverflow.h
asudarsa: +1

lib/SPIRV/SPIRVLowerSaddWithOverflow.h
```diff
@@ -35,22 +35,21 @@
...
+class SPIRVLowerSaddWithOverflowBase {
 public:
   SPIRVLowerSaddWithOverflowBase() : Context(nullptr), Mod(nullptr) {}
-  virtual ~SPIRVLowerSaddWithOverflowBase() {}
-  virtual void visitIntrinsicInst(llvm::CallInst &I);
 
   bool runLowerSaddWithOverflow(llvm::Module &M);
 
 private:
+  void replaceSaddOverflow(llvm::Function &F);
```
asudarsa, line 50: THinking if lowerSaddOverflow/lowerSaddSat will be better names?

jcranmer-intel: Were the parameter an `llvm::IntrinsicInst`, I would agree with you, but the function declarations are replacing the function definition, so I think "replace" is a better verb here than "lower."

lib/SPIRV/SPIRVLowerSaddWithOverflow.h
jcranmer-intel: The main reason I kept the name was I didn't have a particularly better name for the pass, given that not all of the `*_with_overflow` or `*_sat` intrinsics are lowered here.

Maybe `LowerSaddIntrinsics`?

svenvh: Either `LowerSadd` or `LowerSaddIntrinsics` sound good to me.

asudarsa (approved):
Thanks for the explanations. LGTM.

svenvh (approved):

PR #1526, opened 2022-06-29
[NFC] Skip SPIRV -> LLVM metadata translation for entry point wrappers

AGindinson:
Within `SPIRVToLLVM::transFunction()`, SPIR-V's entry point kernel
wrappers and the actual kernels they reference get translated into
the same `llvm::Function *`. Skip the kernel metadata generation
routine for LLVM functions that are mapped onto entry point wrappers,
when similar metadata entries has been generated based on the SPIR-V
kernel function itself.

Additionally, improve code readability through some cosmetic helper
renaming.

### DIFF
--- a/lib/SPIRV/SPIRVReader.cpp
+++ b/lib/SPIRV/SPIRVReader.cpp
@@ -111,7 +111,7 @@ const static char *Restrict = "restrict";
 const static char *Pipe = "pipe";
 } // namespace kOCLTypeQualifierName
 
-static bool isKernel(SPIRVFunction *BF) {
+static bool isEntryPointKernel(SPIRVFunction *BF) {
   return BF->getModule()->isEntryPoint(ExecutionModelKernel, BF->getId());
 }
 
@@ -2810,7 +2810,7 @@ Function *SPIRVToLLVM::transFunction(SPIRVFunction *BF) {
   if (Loc != FuncMap.end())
     return Loc->second;
 
-  auto IsKernel = isKernel(BF);
+  bool IsKernel = isEntryPointKernel(BF);
 
   if (IsKernel) {
     // search for a previous function with the same name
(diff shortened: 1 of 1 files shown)


AGindinson:
@airlied @svenvh Could you please take a look?

lib/SPIRV/SPIRVReader.cpp
```diff
@@ -2810,11 +2810,16 @@ Function *SPIRVToLLVM::transFunction(SPIRVFunction *BF) {
...
+  bool IsKernel = isEntryPointKernel(BF);
 
   if (IsKernel) {
     // search for a previous function with the same name
     // upgrade it to a kernel and drop this if it's found
+    // TODO: Right now we're mapping the same LLVM kernel
+    // onto 2 different SPIR-V functions. Should we completely
+    // erase the entry point function from the SPIRVModule
+    // instead? E.g. to avoid duplicate translation when
+    // iterating over SPIRVModule's functions.
```
YetAnotherCompilerEngineer, line 2822: Not sure if I got the comment right. I'd recommend not to remove the entry point, as it's being used by OpenCL runtime as the actual kernel function.

AGindinson: Never mind - I've removed the comment altogether precisely because there any many cases that don't match the said condition:
> if the original kernel has already been translated within this module (as the below algorithm connects both SPIR-V functions to the same LLVM function)

foutrelis:
Seems the `spirv_param_decorations_quals.ll` test might need to be adjusted for this change (unless it's indicating an actual issue).

AGindinson:
> Seems the `spirv_param_decorations_quals.ll` test might need to be adjusted for this change (unless it's indicating an actual issue).

It is, in fact, indicating an issue with my approach - the checks I've tried to implement are turning out incompatible with SPIR-V parameter decorations functionality (kernel argument qualifiers don't get translated for certain cases like this failing LIT test).

Closing the PR for now - the actual functional issue with [Intel Graphics Compiler](https://github.com/intel/intel-graphics-compiler) that I've been trying to address by this change is specific to IGC's internal copy of SPIR-V Reader (since we're generating additional module metadata that collects kernel argument metadata nodes for all kernels). Personally, I'll try to investigate the entry point wrapper functionality deeper, and should I still have some refactoring propositions regarding the SPIR-V Reader part of https://github.com/KhronosGroup/SPIRV-LLVM-Translator/commit/85815e725ce5bdc970b812b4bbff73d4b2a44046, I'll come up with a new change.

PR #1535, opened 2022-07-07
Remove redundant call-sites adding user-semantic Decorations

asudarsa:


lib/SPIRV/SPIRVWriter.cpp
```diff
@@ -2808,10 +2808,8 @@ AnnotationDecorations tryParseAnnotationString(SPIRVModule *BM,
   RegexIterT DecorationsEnd;
```
asudarsa, line 2808: The functions that call tryParseAnnotationString do check for empty list of decorations and add appropriate UserSemantic Decorations. We can remove the sites that add UserSemantic Decorations inside the callee here.

PR #1089, opened 2021-06-23
Implement SPV_INTEL_debug_module extension

Fznamznon:
Spec https://github.com/intel/llvm/pull/3976

PR #1539, opened 2022-07-07
[Backport] Implement SPV_INTEL_debug_module extension (#1089) (#1520)

mateuszchudyk:
Backport of implementation of SPV_INTEL_debug_module extension (https://github.com/KhronosGroup/SPIRV-LLVM-Translator/pull/1089). Few small changes compared to the origin patch (https://github.com/KhronosGroup/SPIRV-LLVM-Translator/commit/2b3a2f38e2b768e0036440e6f66e07f98043006f) due to differences between llvm11 and llvm13+.


mateuszchudyk:
All required builds passed but one. In this case the problem is in the build command not in the patch. I described it here: https://github.com/KhronosGroup/SPIRV-LLVM-Translator/issues/1546.

svenvh (approved):
> All required builds passed but one. In this case the problem is in the build command not in the patch. I described it here: https://github.com/KhronosGroup/SPIRV-LLVM-Translator/issues/1546.

Thanks, I've seen that failing before but not quite sure what's causing it.  Shouldn't block merging this PR indeed.

PR #1541, opened 2022-07-07
Draft approach for handling transTypeDesc

jcranmer-intel:
This patch is nowhere near ready enough for full review, but I wanted to solicit some feedback on the approach before I delve too much further into implementation:

The basic underlying goal is to have everyone who generates mangled function names identify what the pointer element types of their parameters should be. (This is basically the inverse of the `getParameterTypes` function I added earlier in #1477). As far as getting this information into the guts of the mangler, I've settled on stuffing it in `BuiltinArgTypeMangleInfo` (and this comes with some concomitant changes to `BuiltinTypeMangleInfo` to make doing this easier). But there's a few different ways of getting the type information to that point, and I'm somewhat less certain as to how to do that.

One option is to use `elementtype` attributes and then port the information from attribute lists to the mangling info structs. Except it turns out that actually repeatedly specifying `elementtype` attributes in code isn't very ergonomic, and the code has to follow up by stripping `elementtype` from the attribute lists because it's not (presently) legal in LLVM IR.

The other main option I've considered is to add in an extra `ArrayRef<Type *> PointerElementTypes` argument. This is easier to work with when you're doing something like `addCallInstSPIRV`, but it doesn't quite handle the case where we have a `ocl_event *` parameter, which needs an extra level of indirection. (The indirection in general is somewhat hackily specified, but it does work well enough for now.)

Whatever method is done, all of the `mutateCallInst` calls become more painful. Now you have to either make sure that the attribute sets are moved properly when you add or delete arguments, or you have to concomitantly modify a second array of pointer element types. I've basically decided to give up on keeping the current interface of `mutateCallInst` call and instead resort to a new way of specifying it that makes the modifications generally more obvious:

```c++
mutateCallInst(CI, NewFuncName)
  .addArg(SomeValue)
  .removeArg(3)
  .moveArg(5, 2)
  .mapArg(3, [](Value *V) { return Builder.CreateSplat(10, V); });
```

I've replaced all of the calls to `mutateCallInstSPIRV` to get a feeling for how sufficient or not sufficient this interface is. There's still a lot of cleanup that would need to be done before I'd consider this ready for full review (in particular, `Info.PostProc` needs to operate on a mutator object, and not the `std::vector<Value*> Args` that it does now), and there's even more scope for replacing more code with `IRBuilder` calls. Documentation of the new interface is of course woefully lacking. But I'd like to get some feedback from @svenvh, @YetAnotherCompilerEngineer, and possibly others on if this interface is a desirable approach, or if I should pursue some other way of handling all of the `mutateCallInst` transformations.

### DIFF
--- a/include/LLVMSPIRVLib.h
+++ b/include/LLVMSPIRVLib.h
@@ -157,7 +157,8 @@ bool regularizeLlvmForSpirv(Module *M, std::string &ErrMsg,
 
 /// \brief Mangle OpenCL builtin function function name.
 void mangleOpenClBuiltin(const std::string &UnmangledName,
-                         ArrayRef<Type *> ArgTypes, std::string &MangledName);
+                         ArrayRef<Type *> ArgTypes, AttributeList &Attrs,
+                         std::string &MangledName);
 
 /// Create a pass for translating LLVM to SPIR-V.
 ModulePass *createLLVMToSPIRVLegacy(SPIRV::SPIRVModule *);
--- a/lib/SPIRV/CMakeLists.txt
+++ b/lib/SPIRV/CMakeLists.txt
@@ -9,6 +9,7 @@ add_llvm_library(LLVMSPIRVLib
   OCLTypeToSPIRV.cpp
   OCLUtil.cpp
   VectorComputeUtil.cpp
+  SPIRVBuiltinHelper.cpp
   SPIRVLowerBitCastToNonStandardType.cpp
   SPIRVLowerBool.cpp
   SPIRVLowerConstExpr.cpp
--- a/lib/SPIRV/OCLToSPIRV.cpp
+++ b/lib/SPIRV/OCLToSPIRV.cpp
@@ -146,7 +146,7 @@ void OCLToSPIRVBase::transVecLoadStoreName(std::string &DemangledName,
 char OCLToSPIRVLegacy::ID = 0;
 
 bool OCLToSPIRVBase::runOCLToSPIRV(Module &Module) {
-  M = &Module;
+  initialize(Module);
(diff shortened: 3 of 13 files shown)


svenvh:
> One option is to use elementtype attributes

As you already mention, that isn't legal LLVM IR so relying on that feels a bit fragile (and LLVM may at any point stick an assert in `addAttribute` to enforce that `elementtype` is only used on pointer arguments to intrinsic calls).  So I'm naturally inclined towards your second option, the extra `ArrayRef<Type *> PointerElementTypes` argument.  Perhaps it will end up having to be some kind of `ArrayRef<struct {Type *type; bool IsPointer; }> PointerElementTypes` argument to capture the optional extra level of indirection?

> I've basically decided to give up on keeping the current interface of mutateCallInst call and instead resort to a new way of specifying it that makes the modifications generally more obvious:

That looks like a great improvement to me!  It makes any `mutateCallInst`* calls more readable so I support changing the interface.  It also looks flexible enough, though I do not know the trickier corner cases that you may have come across, if any.  I suppose you'll be opening a separate PR for the `mutateCallInst` work, as this can be done independently from the opaque pointer work (assuming no objections from others).

jcranmer-intel:
I've pulled the changes to `transTypeDesc` itself into a separate PR for review, this will be #1549, which handles everything but the `mutateCallInst` logic. Unfortunately, fixing `mutateCallInst` is ultimately necessary for getting opaque pointer work, since several (I daresay most) of the calls have at least one pointer argument that needs to be mangled. I'll still try to split out as much as I can, however.

jcranmer-intel:
This PR is no longer necessary.

PR #1543, opened 2022-07-08
Add support for toolchain compilation with LLVM_LINK_LLVM_DYLIB option

fineg74:


MrSidims (approved):

svenvh (approved):

aaronpuchert:
Paradoxically, our builds with `LLVM_LINK_LLVM_DYLIB=ON` worked before this change, and this change breaks them. We use `LLVM_LINK_LLVM_DYLIB` because we don't want the static libraries to be used, in fact we don't even ship them. So the `DISABLE_LLVM_LINK_LLVM_DYLIB` is one problem. Another is that we want shared libraries, but this hardcodes the library as static.

The change seems like a workaround for an issue in intel/llvm that causes regressions for other downstream users. Looking at intel/llvm#6413, I think the patch should have been applied there, or at least limited to that particular build scenario and not all builds with `LLVM_LINK_LLVM_DYLIB`. Something like

```cmake
if(INTEL_LLVM AND LLVM_LINK_LLVM_DYLIB)
  add_llvm_library(LLVMSPIRVLib STATIC DISABLE_LLVM_LINK_LLVM_DYLIB ...)
else()
  add_llvm_library(LLVMSPIRVLib ...)
endif()
```

PR #1544, opened 2022-07-11
Update test after LLVM ConstantExpr change

svenvh:
LLVM commit 8edb9c3c56e8 ("[ConstantExpr] Don't create float binop
expressions", 2022-07-08) caused this test to fail.  XFAIL the test to
get CI green again.

### DIFF
--- a/test/complex-constexpr-vector.ll
+++ b/test/complex-constexpr-vector.ll
@@ -6,6 +6,9 @@
 ; RUN: llvm-dis %t.rev.bc
 ; RUN: FileCheck < %t.rev.ll %s --check-prefix=CHECK-LLVM
 
+; TODO: update or remove test following deprecation of float binop constant expressions.
+; XFAIL: *
+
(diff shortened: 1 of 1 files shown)


svenvh:
@bader yeah I noticed the new failure too, seems to be due to LLVM 4bb7b6fae3be ("[IR] Remove support for float binop constant expressions", 2022-07-12).

I haven't investigated yet what the [removal of most constant expressions](https://discourse.llvm.org/t/rfc-remove-most-constant-expressions/63179) means for the translator.  Both the test that I disabled and the newly failing test seem to stem from SYCL, perhaps you have a suggestion going forward?  In the meantime we can of course disable `vector-metadata-constexpr.ll` too, or we can remove the unsupported construct for now.

bader:
It looks like clang emits different expressions for C++ language now and most likely we just need to update LLVM IR in these tests.
@asudarsa, FYI.

svenvh:
I suggest we XFAIL for now with https://github.com/KhronosGroup/SPIRV-LLVM-Translator/pull/1547 to get CI back to green, unless anyone is already working on patching the LLVM IR in the tests.

PR #1545, opened 2022-07-11
Add -emit-opaque-pointers flag

bader:
This option is useful for testing opaque pointers support and controled
transition of lit tests.

### DIFF
--- a/tools/llvm-spirv/llvm-spirv.cpp
+++ b/tools/llvm-spirv/llvm-spirv.cpp
@@ -153,6 +153,11 @@ static cl::opt<bool>
     SPIRVToolsDis("spirv-tools-dis", cl::init(false),
                   cl::desc("Emit textual assembly using SPIRV-Tools"));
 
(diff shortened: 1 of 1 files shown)


tools/llvm-spirv/llvm-spirv.cpp
```diff
@@ -153,6 +153,11 @@ static cl::opt<bool>
     SPIRVToolsDis("spirv-tools-dis", cl::init(false),
                   cl::desc("Emit textual assembly using SPIRV-Tools"));
 
+static cl::opt<bool> EmitOpaquePointers(
+    "emit-opaque-pointers", cl::init(false),
```
svenvh, line 157: I guess this flag might have to be removed soon again, when LLVM drops non-opaque pointer support.  So perhaps it makes sense to make it `cl::Hidden`?

svenvh (commented):
cc: @jcranmer-intel

svenvh (approved):

PR #1549, opened 2022-07-12
[OpaquePtr] Pass pointer type information into mangleBuiltin.

jcranmer-intel:
This includes a rejigging of the BuiltinFuncMangleInfo to actually act as the
source of information to pass this information into.

Only some of the call sites of mangleBuiltin have been properly updated to pass
in the pointer type element information. The various mutateCallInst* callers
have yet to be updated, so those methods are temporarily relying on querying
getPointerElementType() to generate this information.

### DIFF
--- a/lib/SPIRV/SPIRVInternal.h
+++ b/lib/SPIRV/SPIRVInternal.h
@@ -447,6 +447,10 @@ enum Spir2SamplerKind {
   CLK_FILTER_LINEAR = 0x0020,
 };
 
+/// This contains a pair of the pointer element type and an indirection
+/// parameter (to capture cases where an array of OpenCL types is used).
+typedef llvm::PointerIntPair<llvm::Type *, 1, bool> PointerIndirectPair;
+
 /// Additional information for mangling a function argument type.
 struct BuiltinArgTypeMangleInfo {
   bool IsSigned;
@@ -457,10 +461,11 @@ struct BuiltinArgTypeMangleInfo {
   bool IsLocalArgBlock;
   SPIR::TypePrimitiveEnum Enum;
   unsigned Attr;
+  PointerIndirectPair PointerElementType;
   BuiltinArgTypeMangleInfo()
       : IsSigned(true), IsVoidPtr(false), IsEnum(false), IsSampler(false),
         IsAtomic(false), IsLocalArgBlock(false), Enum(SPIR::PRIMITIVE_NONE),
-        Attr(0) {}
+        Attr(0), PointerElementType(nullptr, false) {}
 };
 
 /// Information for mangling builtin function.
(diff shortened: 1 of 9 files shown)


jcranmer-intel:
Note: with this PR, it's possible to get several tests converting SPIR-V to LLVM IR with opaque pointers enabled. Due to not handling the `SPIRVToOCL*` calls to `mutateCallInst`, there are still a lot of assertion failures, and even more tests will fail due to IR checks looking for typed pointer IR in the `CHECK` lines. Between this and #1545, though, it should be possible to start migrating those checks to opaque pointers, however.

lib/SPIRV/SPIRVReader.cpp
```diff
@@ -582,6 +582,98 @@ std::string SPIRVToLLVM::transTypeToOCLTypeName(SPIRVType *T, bool IsSigned) {
...
+        SS << kSPIRVTypeName::PostfixDelim
+           << static_cast<SPIRVConstant *>(Use)->getZExtIntValue();
+      std::string Name =
+          getSPIRVTypeName(kSPIRVTypeName::JointMatrixINTEL, SS.str());
+      PtrElemTy.setPointer(getOrCreateOpaqueStructType(M, Name));
+      break;
+    }
+    case OpTypeFunction: {
+      // A function parameter will get converted into a function pointer later,
+      // so make sure elementtype() attribute reflects the actual function type.
```
svenvh, line 660: Does this comment about `elementtype` still apply?

lib/SPIRV/SPIRVReader.h
```diff
@@ -41,10 +41,12 @@
 #ifndef SPIRVREADER_H
 #define SPIRVREADER_H
 
+#include "SPIRVInternal.h"
 #include "SPIRVModule.h"
 
 #include "llvm/ADT/DenseMap.h"
 #include "llvm/ADT/StringSet.h"
+#include "llvm/IR/Attributes.h"
```
svenvh, line 49: Doesn't seem needed.

lib/SPIRV/SPIRVInternal.h
```diff
@@ -447,6 +447,10 @@ enum Spir2SamplerKind {
   CLK_FILTER_LINEAR = 0x0020,
 };
 
+/// This contains a pair of the pointer element type and an indirection
+/// parameter (to capture cases where an array of OpenCL types is used).
+typedef llvm::PointerIntPair<llvm::Type *, 1, bool> PointerIndirectPair;
```
svenvh, line 452: Do we need this duplicate definition?  There is a `#include "LLVMSPIRVLib.h"` at the top already.

jcranmer-intel: I don't believe we do. (As historical note, I added it here first, and only added it to `LLVMSPIRVLib.h` after discovering that I needed support for it in `mangleOpenClBuiltin` as well.)

svenvh (approved):

PR #1550, opened 2022-07-13
Fix out of bounds access in SPIRVToOCLBase::mutateArgsForImageOperands

dwoodwor-intel:
When called from visitCallSPIRVImageReadBuiltIn, Args only has three
elements and ImOpArgIndex is 2. This has been causing this function to
"update" an argument past the end of the vector, which caused a crash
when running transcoding/image_signedness.ll on Windows because the
debug build in Windows has bounds checking on std::vector.

With this change, it should update the correct (and not past the end)
argument instead and satisfy the Windows bounds checks.


asudarsa (approved):
LGTM. Thanks

bader (approved):

svenvh (approved):
LGTM, thanks for fixing!

svenvh:
Since the commit introducing the issue is also on the `llvm_release_140` branch, I've prepared a backport of the fix: https://github.com/KhronosGroup/SPIRV-LLVM-Translator/pull/1551

PR #1556, opened 2022-07-27
Rework the demangler to support recovering more pointer element types.

jcranmer-intel:
This branch doesn't quite work yet, because it turns out the demangler changes I relied on were in https://github.com/intel/llvm/pull/4207 and not in LLVM itself, so I have to go upstream that first.

### DIFF
--- a/lib/SPIRV/SPIRVUtil.cpp
+++ b/lib/SPIRV/SPIRVUtil.cpp
@@ -38,6 +38,10 @@
 ///
 //===----------------------------------------------------------------------===//
 
+// This file needs to be included before anything that declares
+// llvm::PointerType to avoid a compilation bug on MSVC.
+#include "llvm/Demangle/ItaniumDemangle.h"
+
 #include "FunctionDescriptor.h"
 #include "ManglingUtils.h"
 #include "NameMangleAPI.h"
@@ -50,7 +54,6 @@
 
 #include "llvm/ADT/StringSwitch.h"
 #include "llvm/Bitcode/BitcodeWriter.h"
-#include "llvm/Demangle/ItaniumDemangle.h"
 #include "llvm/IR/IRBuilder.h"
 #include "llvm/IR/IntrinsicInst.h"
 #include "llvm/IR/Metadata.h"
--- a/lib/SPIRV/SPIRVUtil.cpp
+++ b/lib/SPIRV/SPIRVUtil.cpp
@@ -50,7 +50,7 @@
 
 #include "llvm/ADT/StringSwitch.h"
 #include "llvm/Bitcode/BitcodeWriter.h"
-#include "llvm/Demangle/Demangle.h"
+#include "llvm/Demangle/ItaniumDemangle.h"
 #include "llvm/IR/IRBuilder.h"
 #include "llvm/IR/IntrinsicInst.h"
 #include "llvm/IR/Metadata.h"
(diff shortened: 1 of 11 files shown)


jcranmer-intel:
(Also, I might as well wait for https://reviews.llvm.org/D130592 to go in, which will let me get rid of `PointerIndirectPair` as well.)

jcranmer-intel:
The build failures at this point appear to be the base compilers somehow being unable to compile `ItaniumDemangle.h` properly, which is somewhat mystifying given that it's part of the LLVM tree.

jcranmer-intel:
Looking at the build failures and godbolt, it's a gcc warning that's fixed in gcc 10.1 (or 9.3/8.4), but no fix on the 7.x branch, so gcc 7.5 makes it hard fail. Additionally, there's no way to disable this warning without disabling `-Werror` entirely, it seems.

So the options to make the out-of-tree builds not fail are to either disable `-Werror` entirely (at least for out-of-tree), install a newer version of `gcc`, or switch to `clang` as the compiler for building. Thoughts on which is the most preferable?

svenvh:
> Looking at the build failures and godbolt, it's a gcc warning that's fixed in gcc 10.1 (or 9.3/8.4), but no fix on the 7.x branch, so gcc 7.5 makes it hard fail. Additionally, there's no way to disable this warning without disabling `-Werror` entirely, it seems.
> 
> So the options to make the out-of-tree builds not fail are to either disable `-Werror` entirely (at least for out-of-tree), install a newer version of `gcc`, or switch to `clang` as the compiler for building. Thoughts on which is the most preferable?

We could perhaps switch the job to ubuntu-20.04 which comes with gcc 9 if I'm not mistaken.  We'll have to upgrade anyway in less than a year from now as ubuntu-18.04 will be EOL in April 2023.

svenvh:
I've opened https://github.com/KhronosGroup/SPIRV-LLVM-Translator/pull/1579 to upgrade the CI jobs.

lib/SPIRV/SPIRVUtil.cpp
```diff
@@ -701,61 +882,37 @@ void getParameterTypes(Function *F, SmallVectorImpl<StructType *> &ArgTys) {
...
-          Pointee = getOrCreateOpaqueStructType(M, MangledStructName);
-        }
-      } else if (!Arg.contains(' ') && Arg.startswith("ocl_")) {
-        std::string StructName = demangleBuiltinOpenCLTypeName(Arg);
-        Pointee = getOrCreateOpaqueStructType(M, StructName);
-      }
-      *ArgIter++ = Pointee;
+  // Sanity check that the name mangling matches up to the expected number of
+  // arguments.
+  if (RootNode->getParams().size() > (size_t)(ArgTys.end() - ArgIter)) {
```
MrSidims, line 896:
```suggestion
  if (RootNode->getParams().size() != (size_t)(ArgTys.end() - ArgIter)) {
```

lib/SPIRV/SPIRVUtil.cpp
```diff
@@ -701,61 +882,37 @@ void getParameterTypes(Function *F, SmallVectorImpl<StructType *> &ArgTys) {
...
+    LLVM_DEBUG(dbgs() << "[getParameterTypes] function " << MangledName
+                      << " appears to have " << RootNode->getParams().size()
+                      << " arguments but has " << (ArgTys.end() - ArgIter)
+                      << "\n");
```
MrSidims, line 900: Here we should probably notify llvm-spirv user and not llvm-spirv developer, that something went very wrong. Should it be `unreachable` or any other error reporting mechanism instead of LLVM_DEBUG?

MrSidims (commented):
Looks good, just 1 nit/suggestion

lib/SPIRV/SPIRVUtil.cpp
jcranmer-intel: This function will, in certain cases, be called on user-defined functions (as opposed to just SPIR-V/OpenCL builtins), so I'm reluctant to make it be a hard fail, in case someone decides to name their OpenCL kernel function something which is a valid Itanium name mangling that doesn't match their function signature. While reporting to the llvm-spirv user in lieu of the developer would be useful, I'm not sure if there's an appropriate mechanism for that.

MrSidims: Got it, thanks!

>> While reporting to the llvm-spirv user in lieu of the developer would be useful, I'm not sure if there's an appropriate mechanism for that.

Yeah, I remember having some internal tracker for that :)

svenvh (approved):
LGTM, thanks!

Issue #1562, opened 2022-07-29
incorrect translation of sampler/image kernel arguments


franz:
I have bitcode ( [sample.ll.gz](https://github.com/KhronosGroup/SPIRV-LLVM-Translator/files/9220700/sample.ll.gz) ) with a function that forwards image+sampler arguments to another, and is correctly translated to SPIR-V by the `llvm_release_140` branch but incorrectly by the main (@ 70ebd9b82edb5d8aca04a3e8bbc) branch.

```llvm
%opencl.image2d_ro_t = type opaque
%opencl.sampler_t = type opaque
....

; Function Attrs: convergent mustprogress norecurse nounwind
define hidden spir_kernel void @_Z11tex2DKernelPfP13__hip_textureii(float addrspace(1)* nocapture noundef writeonly %0, %opencl.image2d_ro_t addrspace(1)* %1, %opencl.sampler_t addrspace(2)* %2, i32 noundef %3, i32 noundef %4) local_unnamed_addr #5
```

translated with llvm 14 branch:

```
....
         %20 = OpTypeImage %void 2D 0 0 0 0 Unknown ReadOnly
         %21 = OpTypeSampler
....
%_Z11tex2DKernelPfP13__hip_textureii = OpFunction %void None %61
         %63 = OpFunctionParameter %_ptr_CrossWorkgroup_float
         %64 = OpFunctionParameter %20
         %65 = OpFunctionParameter %21
         %66 = OpFunctionParameter %uint
         %67 = OpFunctionParameter %uint
....
```

translated with main branch:

```
....
         %20 = OpTypeImage %void 2D 0 0 0 0 Unknown ReadOnly
         %21 = OpTypeSampler
....
%_Z11tex2DKernelPfP13__hip_textureii = OpFunction %void None %61
         %63 = OpFunctionParameter %_ptr_CrossWorkgroup_float
         %64 = OpFunctionParameter %21
         %65 = OpFunctionParameter %21
         %66 = OpFunctionParameter %uint
         %67 = OpFunctionParameter %uint
...
```

The 2nd argument ends up being sampler instead of image. It also fails validation:

```
error: line 187: OpFunctionCall Argument <id> '64[%64]'s type does not match Function <id> '20[%20]'s parameter type.
  %100 = OpFunctionCall %v4float %_chip_tex2df_impl %64 %65 %99
```

svenvh:
[...]

PR #1567, opened 2022-08-05
Restore getArgIndex in OCLTypeToSPIRV

svenvh:
Commit 4a9c78ee ("Prepare SPIRVWriter for type conversion without
opaque pointers. (#1499)", 2022-06-20) removed `getArgIndex` in favor
of passing `Idx`, but this leads to incorrect adaptation of argument
types if sampler arguments are not in the same position between
different functions.

We might be able to drop `adaptArgumentsBySamplerUse` entirely, as we
don't represent samplers as i32 anymore; but for now just fix the
regression.

Fixes https://github.com/KhronosGroup/SPIRV-LLVM-Translator/issues/1562

### DIFF
--- a/lib/SPIRV/LLVMSPIRVOpts.cpp
+++ b/lib/SPIRV/LLVMSPIRVOpts.cpp
@@ -50,9 +50,9 @@ using namespace SPIRV;
 
 bool TranslatorOpts::isUnknownIntrinsicAllowed(IntrinsicInst *II) const
     noexcept {
-  if (!SPIRVAllowUnknownIntrinsics.hasValue())
+  if (!SPIRVAllowUnknownIntrinsics.has_value())
     return false;
-  const auto &IntrinsicPrefixList = SPIRVAllowUnknownIntrinsics.getValue();
+  const auto &IntrinsicPrefixList = SPIRVAllowUnknownIntrinsics.value();
   StringRef IntrinsicName = II->getCalledOperand()->getName();
   for (const auto Prefix : IntrinsicPrefixList) {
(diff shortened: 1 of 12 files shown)


svenvh:
@jcranmer-intel just wondering if the removal of `getArgIndex` was intentional?

jcranmer-intel:
> @jcranmer-intel just wondering if the removal of `getArgIndex` was intentional?

It was. Rather than iterating over the function, it should be possible to just `cast<Argument>(SamplerArg)->getArgNo()` instead.

svenvh:
> It was. Rather than iterating over the function, it should be possible to just cast<Argument>(SamplerArg)->getArgNo() instead.

That's a nice improvement in itself, but that doesn't explain [this change](https://github.com/KhronosGroup/SPIRV-LLVM-Translator/pull/1499/files#diff-ce1d0dba5c92bf9cd4c03cef699f218b66153363b26bfdba389a6fb443d2f335L215) in #1499:

```
@@ -206,11 +181,6 @@ void OCLTypeToSPIRVBase::adaptArgumentsBySamplerUse(Module &M) {
[...]
       addWork(Caller);
-      TraceArg(Caller, getArgIndex(Caller, SamplerArg));
+      TraceArg(Caller, Idx);
     }
```

I doubt this is in the way for opaque pointers as we're tracing Values here, but perhaps I'm missing something?

jcranmer-intel:
> I doubt this is in the way for opaque pointers as we're tracing Values here, but perhaps I'm missing something?

It was done as part of the work to convince myself that the if statement that used `getPointerElementType` was unnecessary. Strictly speaking, the change wasn't necessary.

jcranmer-intel (approved):

PR #1568, opened 2022-08-05
Use TypedPointerType in lieu of PointerIndirectPair in mangleBuiltin.

jcranmer-intel:


lib/SPIRV/SPIRVReader.cpp
```diff
@@ -355,11 +356,19 @@ std::string SPIRVToLLVM::transVCTypeName(SPIRVTypeBufferSurfaceINTEL *PST) {
   return VectorComputeUtil::getVCBufferSurfaceName();
 }
 
-Type *SPIRVToLLVM::transType(SPIRVType *T, bool IsClassMember) {
+Type *SPIRVToLLVM::transType(SPIRVType *T, bool UseTPT) {
   auto Loc = TypeMap.find(T);
-  if (Loc != TypeMap.end())
+  if (Loc != TypeMap.end() && !UseTPT)
```
AlexeySachkov, line 361: I'm a bit confused here: if `UseTPT == true`, does it mean that even if `TypeMap` contains the requested `T`, we will still ignore it and generate a new opaque type?

Won't it create types duplicates due to names collision (like `struct.name`, `struct.name.1`, etc.)?

lib/SPIRV/SPIRVReader.cpp
```diff
@@ -381,11 +390,13 @@ Type *SPIRVToLLVM::transType(SPIRVType *T, bool IsClassMember) {
...
-  case OpTypePointer:
-    return mapType(
-        T, PointerType::get(
-               transType(T->getPointerElementType(), IsClassMember),
-               SPIRSPIRVAddrSpaceMap::rmap(T->getPointerStorageClass())));
+  case OpTypePointer: {
+    unsigned AS = SPIRSPIRVAddrSpaceMap::rmap(T->getPointerStorageClass());
+    Type *ElementTy = transType(T->getPointerElementType(), UseTPT);
+    if (UseTPT)
+      return TypedPointerType::get(ElementTy, AS);
```
AlexeySachkov, line 397: Why can't we use `transType` result as-is in this case?

lib/SPIRV/SPIRVReader.cpp
jcranmer-intel: (see below for answer)

lib/SPIRV/SPIRVReader.cpp
jcranmer-intel: This is part of the opaque pointers work. When reading SPIR-V IR, we need to keep track of what the typed pointer equivalent of all of the SPIR-V types are for eventual feeding into the name mangler. When `transType` normally returns a `PointerType` (whether special SPIR-V opaque type or a SPIR-V pointer type), we need to keep track elsewhere than the resulting LLVM IR type of what the type is.

Previously, this was done with `PointerIndirectPair`, but this patch replaces that representation with `TypedPointerType`. This also removes the duplication of `transType` just to compute `PointerIndirectPair`/`TypedPointerType`, but in doing so, it requires more care: we don't want to cache a result of `TypedPointerType` in the type map (since `TypedPointerType` cannot be used in LLVM IR), and if we want to compute a `TypedPointerType`, we can't reuse the cached `PointerType` in the map, since it doesn't have the info we need.

As for worries about type duplication: LLVM uniques most types already, with struct types being the case where it's possible to get type duplication. However, most of the struct types go through the `getOrCreateOpaqueStructType` helper, which will ensure that there's no duplication of names. The ones that don't go through `OpTypeStruct`, where there is an explicit check to reuse the type map, even if `UseTPT` is true (as struct types don't pass `UseTPT` into recursive type construction--we never create structs whose elements may be `TypedPointerType`, as we don't need them for mangling purposes).

I can clarify this in source code comments for the next revision.

MrSidims (commented):
The patch seems legit, but I would need to take a look at other `mangleBuiltin`-related patches first, since I've almost entirely missed them.

lib/SPIRV/SPIRVReader.cpp
AlexeySachkov: Thanks for the explanation, I now get it: in some code paths we at first produce actual `typed` pointers to generate proper mangling and then we opaquify them in order to emit correct LLVM IR

lib/SPIRV/SPIRVUtil.cpp
```diff
@@ -1634,6 +1631,10 @@ std::string mangleBuiltin(StringRef UniqName, ArrayRef<Type *> ArgTypes,
                                               : (unsigned)BtnInfo->getVarArg();
          I != E; ++I) {
       auto T = ArgTypes[I];
+      auto MangleInfo = BtnInfo->getTypeMangleInfo(I);
+      if (MangleInfo.PointerTy && T->isPointerTy()) {
```
AlexeySachkov, line 1635: Could there be situations where we encountered an opaque pointer type for which we don't have corresponding actual type within `MangleInfo`?

AlexeySachkov (approved):
Overall LGTM, but I wonder what is the testing plan for the change (or changes related to opaque pointers in general)?

lib/SPIRV/SPIRVUtil.cpp
jcranmer-intel: Yes. `transTypeDesc` has an `if (Ty->isPointerTy())` block that treats such pointers as if they were `i8*`.

MrSidims (approved):

PR #1570, opened 2022-08-08
Drop the JointMatrixINTEL struct-renaming pass when opaque pointers are enabled.

jcranmer-intel:
The frontend is being changed to lower the struct name to the correct LLVM name
directly, obviating the need for this check. See
https://github.com/intel/llvm/pull/6535 for this change.

This marks the removal of the final call to the deprecated method
Type::getPointerElementType, although there remains some code that is not fully
working with opaque pointers enabled.

### DIFF
--- a/lib/SPIRV/SPIRVRegularizeLLVM.cpp
+++ b/lib/SPIRV/SPIRVRegularizeLLVM.cpp
@@ -375,7 +375,7 @@ void SPIRVRegularizeLLVMBase::adaptStructTypes(StructType *ST) {
   // register by OpCompositeConstruct. And we can't claim, that the Result type
   // of OpCompositeConstruct instruction is always the joint matrix type, it's
   // simply not true.
-  if (MangledName == "__spirv_JointMatrixINTEL" && !PtrTy->isOpaque()) {
+  if (MangledName == "__spirv_JointMatrixINTEL" && !ST->isOpaquePointerTy()) {
     auto *PtrTy = dyn_cast<PointerType>(ST->getElementType(0));
     assert(PtrTy &&
            "Expected a pointer to an array to represent joint matrix type");
--- a/lib/SPIRV/SPIRVRegularizeLLVM.cpp
+++ b/lib/SPIRV/SPIRVRegularizeLLVM.cpp
@@ -355,6 +355,10 @@ void SPIRVRegularizeLLVMBase::adaptStructTypes(StructType *ST) {
   STName.consume_front("__spv::");
   StringRef MangledName = STName.substr(0, STName.find('.'));
 
+  // Older versions of clang will generate JointMatrixINTEL types using this
+  // representation. Newer versions will generate the correct struct name
+  // "%spirv.JointMatrixINTEL._{parameters}" directly, obviating the need for
+  // this check.
   // Representation in LLVM IR before the translator is a pointer array wrapped
   // in a structure:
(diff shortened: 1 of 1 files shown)


lib/SPIRV/SPIRVRegularizeLLVM.cpp
```diff
@@ -373,10 +377,11 @@ void SPIRVRegularizeLLVMBase::adaptStructTypes(StructType *ST) {
   // simply not true.
   if (MangledName == "__spirv_JointMatrixINTEL") {
```
MrSidims, line 378: Guess, it should be looking like this then, SPIRVWriter part is supposed to parse the type name:

```suggestion
  if (MangledName == "__spirv_JointMatrixINTEL" && !PtrTy->isOpaque()) {
```

lib/SPIRV/SPIRVRegularizeLLVM.cpp
```diff
@@ -373,10 +377,11 @@ void SPIRVRegularizeLLVMBase::adaptStructTypes(StructType *ST) {
...
     auto *PtrTy = dyn_cast<PointerType>(ST->getElementType(0));
-    assert(PtrTy &&
+    assert(PtrTy && !PtrTy->isOpaque() &&
```
MrSidims, line 380:
```suggestion
    assert(PtrTy &&
```

lib/SPIRV/SPIRVRegularizeLLVM.cpp
```diff
@@ -355,6 +355,10 @@ void SPIRVRegularizeLLVMBase::adaptStructTypes(StructType *ST) {
   STName.consume_front("__spv::");
   StringRef MangledName = STName.substr(0, STName.find('.'));
 
+  // Older versions of clang will generate JointMatrixINTEL types using this
+  // representation. Newer versions will generate the correct struct name
+  // "%spirv.JointMatrixINTEL._{parameters}" directly, obviating the need for
+  // this check.
```
AlexeySachkov, line 361: I suppose we should add a test with new LLVM IR representation?

BTW, do we need to remove the old one or we still consider that we should be/are able to accept the old IR as well?

jcranmer-intel: My plan is to maintain compatibility with the old IR for a short while (perhaps until LLVM rips out support for `PointerType::getPointerElementType` entirely).

As for tests, I'm somewhat holding off on that until I (or somebody else) starts going through and moving everything to use opaque pointers in lieu of typed pointers.

AlexeySachkov: Considering that we don't test non-opaque pointers path at all (if I understand correctly), I guess it is ok to merge it as-is for now and push some fixes if they would be necessary later.

I will approve this patch, but it would be good if someone else confirmed that it is ok for such patches to come without tests for now. (Pardon my ignorance, I haven't been following the development in the translator for the past few months)

AlexeySachkov (approved):

MrSidims (approved):

lib/SPIRV/SPIRVRegularizeLLVM.cpp
svenvh: > it would be good if someone else confirmed that it is ok for such patches to come without tests for now

I think that's the most practical thing we can do while we are in this transitional state, and it is pretty much how we've handled the transition so far.  The expectation is indeed that further fixes will be needed when we start converting the test suite to opaque pointers.

PR #1578, opened 2022-08-16
Dot product bugfix to include more floating point types

Quetzonarch:
This is a bugfix for my previous change ( #1174 ) regarding the integer dot product translation. 

Slightly changed the check for Dot call in OCLToSPIRV to include more floating point types, as some were omitted by mistake.

### DIFF
--- a/lib/SPIRV/OCLToSPIRV.cpp
+++ b/lib/SPIRV/OCLToSPIRV.cpp
@@ -323,8 +323,8 @@ void OCLToSPIRVBase::visitCallInst(CallInst &CI) {
     return;
   }
   if (DemangledName == kOCLBuiltinName::Dot &&
-      (CI.getOperand(0)->getType()->isFloatTy() ||
-       CI.getOperand(1)->getType()->isDoubleTy())) {
+      (!CI.getOperand(0)->getType()->isVectorTy() &&
+       !CI.getOperand(0)->getType()->isIntegerTy())) {
     visitCallDot(&CI);
     return;
   }
--- a/lib/SPIRV/OCLToSPIRV.cpp
+++ b/lib/SPIRV/OCLToSPIRV.cpp
@@ -323,8 +323,7 @@ void OCLToSPIRVBase::visitCallInst(CallInst &CI) {
     return;
   }
   if (DemangledName == kOCLBuiltinName::Dot &&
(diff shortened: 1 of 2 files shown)


lib/SPIRV/OCLToSPIRV.cpp
```diff
@@ -323,8 +323,8 @@ void OCLToSPIRVBase::visitCallInst(CallInst &CI) {
     return;
   }
   if (DemangledName == kOCLBuiltinName::Dot &&
-      (CI.getOperand(0)->getType()->isFloatTy() ||
-       CI.getOperand(1)->getType()->isDoubleTy())) {
+      (!CI.getOperand(0)->getType()->isVectorTy() &&
+       !CI.getOperand(0)->getType()->isIntegerTy())) {
```
svenvh, line 327: Wondering if just `CI.getOperand(0)->getType()->isFPOrFPVectorTy()` is sufficient?

svenvh (commented):
> Slightly changed the check for Dot call in OCLToSPIRV to include more floating point types, as some were omitted by mistake.

I suppose you mean that any floating point type should be handled as before?  It would be great if you could add a test case that was translated incorrectly after #1174 and is fixed with this change.

lib/SPIRV/OCLToSPIRV.cpp
Quetzonarch: From what I understand this visitCallDot is meant to translate dot with scalar floating point arguments (but not vectors) into FMul. Because of that FPVectors shouldn't pass this check but instead stay as calls to dot.

Quetzonarch:
I've added a test case which uses the dot call with scalar half arguments. This should be translated into FMul but because of my previous change it would be translated into an integer dot (DotKHR). This is what I was mainly aiming to fix with this update.

lib/SPIRV/OCLToSPIRV.cpp
jcranmer-intel: `isFloatingPointTy()`?

Quetzonarch: Yeah, that might be a good idea

svenvh (approved):

MrSidims (approved):

PR #1583, opened 2022-08-19
[Backport to 11] Add an entry point wrapper around functions (llvm pass) (#1149)

stanleygambarin:
SPIR-V spec states:

In order to satisfy SPIR-V that entrypoints and functions
must be different, this introduces an entrypoint wrapper around
functions at the LLVM IR level, then fixes up a few things like
naming at the SPIRV translation.


MrSidims:
Interesting that CLA message presents here and not in other PRs. @stanleygambarin  please sign it.

MrSidims (approved):

PR #1587, opened 2022-08-19
Mass add -emit-opaque-pointers for tests that require some changes.

jcranmer-intel:


jcranmer-intel:
Note that the vast, vast majority of the changes actually required consists of changing `i8*` (or other pointer types as appropriate) to `ptr` or LLVM intrinsic mangling changes (e.g. `llvm.memcpy.p0i8.p0i8.i64` becoming `llvm.memcpy.p0.p0.i64`). There is a case or two of a `bitcast` getting dropped, and another one where another check line gets added because otherwise the `alloca ptr` matches the wrong line (in test/transcoding/SPV_INTEL_function_pointers/function-pointer-as-function-arg.ll).

Between this and #1586, most of the `llvm-spirv -r` tests will have been modified to check opaque pointer IR rather than typed pointer IR. Those that do not are largely for one of three reasons: either the tool will crash (generally because `SPIRVToOCL` has a call to `mutateCallInst` that currently queries pointer element type), the IR output gets fed back into SPIR-V that doesn't match the typed IR input, or the output opaque pointer IR is more radically different compared to the checks--particularly those that match struct definitions or `%foo = type opaque` lines. There is around 80 tests that I opted not to convert to checking opaque IR output.

svenvh (approved):
LGTM for the non-vendor-specific tests, thanks!

There were some that slipped through the cracks of #1586 it seems?  For example `test/transcoding/NoSignedUnsignedWrap.ll` does not seem to have any changes to its CHECKs.

MrSidims (approved):
I'm slightly worried, that with these changes we are starting to lack testing of the translator without `-emit-opaque-pointers` option. I acknowledge that `-emit-opaque-pointers` should become a default behavior, but I would expect to see few tests with an opposite behavior. Though those tests might be added later. WDYT?

svenvh:
Yes that's a fair point; it depends how much we want to support non-opaque pointers I suppose (and that depends on how long we will be in the transition phase of supporting both opaque and non-opaque mode at the same time).

We could duplicate the current test suite without `-emit-opaque-pointers` into `test/non-opaque` perhaps, but that has the downside of massive duplication.  Or we could only have a handful of smaller non-opaque tests just to verify that some basic typed pointer support keeps working.  Any preferences?

jcranmer-intel:
> There were some that slipped through the cracks of #1586 it seems? For example `test/transcoding/NoSignedUnsignedWrap.ll` does not seem to have any changes to its CHECKs.

A couple of the files slipped through the cracks because I assumed that every file with >2 lines changed had a change to CHECK lines. (#1586 and #1587 were originally one patch, but I split it in two to help avoid reviewers' eyes glazing over.)

> I'm slightly worried, that with these changes we are starting to lack testing of the translator without -emit-opaque-pointers option. I acknowledge that -emit-opaque-pointers should become a default behavior, but I would expect to see few tests with an opposite behavior. Though those tests might be added later. WDYT?

This is definitely a concern. For the tests in this batch of changes, I didn't want to change any tests where getting the pointer element type correct seemed to be the point of the test--for example, the JointMatrixINTEL tests which check for opaque struct names. For future tests, particularly the tests around image opaque types, duplicating them to opaque and non-opaque tests makes sense. The only tests in this batch that I see that might be valuable to duplicate to non-opaque tests would be the function pointer tests.

> Yes that's a fair point; it depends how much we want to support non-opaque pointers I suppose (and that depends on how long we will be in the transition phase of supporting both opaque and non-opaque mode at the same time).

My understanding is that LLVM is looking to rip out support for non-opaque pointers before the end of this year.

MrSidims:
>> My understanding is that LLVM is looking to rip out support for non-opaque pointers before the end of this year.

The thing is that out of the current version of main branch we will still have one more concurrent with LLVM release, which is supposed to support non-opaque pointers.

I up for merging this PR. I'll post my thoughts of the lists of tests to add in the issue. Or even better - I'll create a PR with those tests by EOW. For sure it won't be the whole duplication of the existing tests, but I would definitely add a test for pointers to all basic types and as Joshua has mentioned pointers to special OpenCL types and the appropriate objects initialization (for example PipeStorages thought bitcast initialization) and function pointers tests. If would expect in total of 20 tests to be duplicated.

With all these said, @svenvh feel free to merge the PR unless the discussion above raised some concerns.

svenvh:
Thanks, that sounds good to me @MrSidims.  I'll merge this PR.

PR #1604, opened 2022-08-23
[NFC] Remove ExecutionModeNamedBarrierCountINTEL from spirv_internal

MrSidims:
It was actually upstreamed to SPIR-V Headers already.


bashbaug (approved):
Thanks for the quick update!

PR #1607, opened 2022-08-31
Make Use JointMatrix mandatory

MrSidims:


MrSidims:
TODO:
1. Add description;
2. Fix the failing tests. For some reasons postfix for the SPIR-V-friendly instructions  names changed from:
._short_2_2_0_3_0 to __short_2_2_0_3_0 , which is unexpected.

MrSidims:
@yubingex007-a11y could you please take a look, since you are familiar with SPIR-V matrix processing

yubingex007-a11y:
would you provide some testcases?

lib/SPIRV/SPIRVWriter.cpp
```diff
@@ -626,17 +626,74 @@ SPIRVType *LLVMToSPIRVBase::transSPIRVJointMatrixINTELType(
...
+// { row_major, col_major, packed, unused }
+// while the SPIR-V spec states, that 'Layout' can be:
+// { RowMajor, ColMajor, PackedA, PackedB }
+// So we need to transform packed to PackedA or PackedB depending on
+// 'Use' parameter of the JointMatrix type.
+template<uint32_t LayoutPos = 0>
+SPIRVValue *LLVMToSPIRVBase::transSPIRVJointMatrixINTELMemoryLayout(
+    CallInst *CI, SPIRVBasicBlock *BB, SPIRVTypeJointMatrixINTEL *MatTy) {
+  auto *Use = MatTy->getUse();
+  if (!Use)
```
yubingex007-a11y, line 662: if we don't have use and layout is packed, layout is still "packed" after spv-writer?
@MrSidims

lib/SPIRV/SPIRVWriter.cpp
```diff
@@ -626,17 +626,74 @@ SPIRVType *LLVMToSPIRVBase::transSPIRVJointMatrixINTELType(
...
+  case 1:
+    return transValue(CI->getArgOperand(LayoutPos), BB);
+  case 2: {
+    if (UseVal == internal::InternalJointMatrixUse::MatrixA)
+      return transValue(
+          ConstantInt::get(Type::getInt32Ty(CI->getContext()),
+                           internal::InternalJointMatrixLayout::PackedA), BB);
+    if (UseVal == internal::InternalJointMatrixUse::MatrixB)
+      return transValue(
+          ConstantInt::get(Type::getInt32Ty(M->getContext()),
```
yubingex007-a11y, line 679: maybe align with line675? "M->getContext()"=>"CI->getContext()"

lib/SPIRV/SPIRVWriter.cpp
```diff
@@ -626,17 +626,74 @@ SPIRVType *LLVMToSPIRVBase::transSPIRVJointMatrixINTELType(
...
     unsigned long long N = 0;
     consumeUnsignedInteger(Postfix, 10, N);
-    return getUInt32(M, N);
+    return N;
   };
-  std::vector<SPIRVValue *> Args;
+  std::vector<uint32_t> ProcessedPostfixes;
   for (size_t I = 1; I != Postfixes.size(); ++I)
-    Args.emplace_back(transConstant(ParseInteger(Postfixes[I])));
+    ProcessedPostfixes.emplace_back(ParseInteger(Postfixes[I]));
```
yubingex007-a11y, line 636: So, here we are parsing the  "10_2_0_0" of 
%spirv.JointMatrixINTEL._bfloat16_10_2_0_0, right? instead of parsing each dimension of arrays?

lib/SPIRV/SPIRVWriter.cpp
MrSidims: Yes. I consider, that for previous code cases we still have PackedA and PackedB as the layout. It will remain the same.

lib/SPIRV/SPIRVWriter.cpp
MrSidims: The flow was the following:
We have a multidimensional array to process by RegularizeLLVM pass (it's W/A). It was changing a matrix name to have a postfix like you've mentioned.
Then SPIRVWriter looks at this postfix and parse it to obtain rows, cols etc.
So this patch doesn't change this flow.

Note, that in a close future we will remove this nasty array processing, since now DPCPP FE can create an opaque matrix structure with the correct name (see https://github.com/intel/llvm/pull/6535 )

lib/SPIRV/SPIRVWriter.cpp
MrSidims: Sure

MrSidims:
No need for this PR anymore

PR #1608, opened 2022-09-01
Translate function pointer from global variable as pointer, not as declaration

vmaksimo:
This patch helps to avoid invalid SPV generation.

When global variable contains a pointer to a function, translator tries to translate it as declaration. Then it translates this function the second time when going through the function list. This leads to double translation of the same function and to the usage of the same IDs in SPIR-V file.

### DIFF
--- a/lib/SPIRV/SPIRVWriter.cpp
+++ b/lib/SPIRV/SPIRVWriter.cpp
@@ -1683,7 +1683,7 @@ LLVMToSPIRVBase::transValueWithoutDecoration(Value *V, SPIRVBasicBlock *BB,
           }
         }
       }
-      BVarInit = transValue(Init, nullptr);
+      BVarInit = transValue(Init, nullptr, CreateForward, FuncTrans);
     }
 
     SPIRVStorageClassKind StorageClass;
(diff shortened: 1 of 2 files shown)


vmaksimo:
@MrSidims @jcranmer-intel @asudarsa @vmaksimo @jcranmer-intel please take a look

lib/SPIRV/SPIRVWriter.cpp
```diff
@@ -1683,7 +1683,7 @@ LLVMToSPIRVBase::transValueWithoutDecoration(Value *V, SPIRVBasicBlock *BB,
           }
         }
       }
-      BVarInit = transValue(Init, nullptr);
+      BVarInit = transValue(Init, nullptr, CreateForward, FuncTrans);
```
jcranmer-intel, line 1686: I think that, rather than passing through the `CreateForward` and `FuncTrans` parameters, this should just always call it with `true` and `FuncTransMode::Pointer` here...

lib/SPIRV/SPIRVWriter.cpp
```diff
@@ -4159,7 +4159,11 @@ bool LLVMToSPIRVBase::transGlobalVariables() {
       continue;
     } else if (MDNode *IO = ((*I).getMetadata("io_pipe_id")))
       transGlobalIOPipeStorage(&(*I), IO);
-    else if (!transValue(&(*I), nullptr))
+    // As global variables define a pointer to their "content" type, we should
+    // translate here only pointer without declaration even it is a function
+    // pointer.
+    else if (!transValue(&(*I), nullptr, true /*CreateForward - default value*/,
+                         FuncTransMode::Pointer))
```
jcranmer-intel, line 4166: ... which means this change would be unnecessary. It feels wrong to me that the code in the writer responsible for causing the global value's definition to get emitted is telling the value emission code to *not* emit definitions (even if that mode only affects functions, not global values in general).

jcranmer-intel (changes requested):

lib/SPIRV/SPIRVWriter.cpp
vmaksimo: Thanks! You are right, it'd be better to call the function with the needed options directly rather than passing through.
I'll upload the fixing change shortly.

MrSidims (approved):
LGTM

jcranmer-intel (approved):

PR #1613, opened 2022-09-09
[Draft] tests for full use of new mutator API

jcranmer-intel:
This is just a draft PR to run all of the tests after moving all of the tests with the new mutator API added in #1611. The actual PRs will be much smaller to make it much easier to review.


jcranmer-intel:
This PR is no longer necessary, as all of the pieces have been broken out into independent PRs.

PR #1614, opened 2022-09-13
[Mutator] Migrate atomics builtins to the mutator interface.

jcranmer-intel:
The old mutateCallInst* functions and the new mutator doesn't quite get some of the value names correct, which necessitated some test changes that were looking for particular value names.


MrSidims (approved):
LGTM one nit/suggestion (not really related to the patch)

PR #1630, opened 2022-09-28
Fix parsing of --spec-const option

franz:
While running OpenCL-CTS SPIR-V subtests, i found that launching llvm-spirv with arguments:

`-r --spec-const=101:i64:4609589727908835759`

results in error message:

`Error: Invalid value for '-spec-const' option! In "101:i64:4609589727908835759": can't convert "4609589727908835759" to 64-bit integer number`

However, 4609589727908835759 fits into 64bits (0x3ff88d6f544bb1af). The problem seems to be that getNumWords() returns 2 so i changed it to use getActiveBits(), but i'm not 100% sure this is the proper fix.

### DIFF
--- a/test/transcoding/spec_const.ll
+++ b/test/transcoding/spec_const.ll
@@ -4,7 +4,7 @@
 ; RUN: llvm-spirv %t.spv -to-text -o - | FileCheck %s --check-prefix=CHECK-SPIRV
 ; RUN: llvm-spirv -r -emit-opaque-pointers %t.spv -o %t.rev.bc
 ; RUN: llvm-dis < %t.rev.bc | FileCheck %s --check-prefix=CHECK-LLVM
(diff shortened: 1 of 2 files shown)


MrSidims:
Thanks for the report and the PR, that is very appreciated! Could you please add a small test (may be modify transcoding/spec_const.ll ) so we can ensure, that the issue is gone? TBH I'm very surprised by the fact, that getNumWords() returns 2 for 4609589727908835759 .

svenvh (approved):
> TBH I'm very surprised by the fact, that getNumWords() returns 2 for 4609589727908835759 .

Me too, so I did a little investigation.  It turns out `StringRef::getAsInteger` deliberately overestimates the required number of bits based on the radix and the string length.

So `getActiveWords` is the right method to use here.  Thanks for the fix!

MrSidims (approved):

PR #1633, opened 2022-09-30
Integer dot product 4x8 packed translation

Quetzonarch:
This proposed bugfix updates the integer dot translation to use the correct function names from the OpenCL extension specification for packed format with scalar integer arguments (i.e. dot_4x8packed or dot_acc_sat_4x8packed instead of dot or dot_acc_sat with scalar integer arguments) to translate them into proper OpCodes. This is a fix for incorrect function names introduced in #1174 
Additionally removed unused variables from visitCallDot to clean up the code.


MrSidims (approved):
LGTM

svenvh (approved):

PR #1635, opened 2022-10-06
Fix reverse translation addrspace annotation assert

dwoodwor-intel:
When annotating values, SPIRVToLLVM::transIntelFPGADecorations scans the preceding instructions to see if they're based on an alloca. llvm.var.annotation is used to mark that value if there is a base alloca; otherwise, llvm.ptr.annotation is used. In some cases an annotated value based on an alloca has a non-zero address space, which is a problem because llvm.var.annotation (unlike llvm.ptr.annotation) only accepts address space zero pointers. This change is a simple fix to add another addrspacecast to return the pointer to address space zero so it can be annotated.

Some misleading variable names are also updated to more clearly convey the behavior of this code.

### DIFF
--- a/test/transcoding/annotate_attribute.ll
+++ b/test/transcoding/annotate_attribute.ll
@@ -221,8 +221,7 @@ define weak_odr dso_local spir_kernel void @_ZTSZ11TestKernelAvE4MyIP(i32 addrsp
   ; CHECK-LLVM: %[[ALLOCA:.*]] = alloca %struct.MyIP, align 8
   ; CHECK-LLVM: %[[BITCAST:.*]] = bitcast ptr %[[ALLOCA]] to ptr
   ; CHECK-LLVM: %[[ADDRSPACECAST:.*]] = addrspacecast ptr %[[BITCAST]] to ptr addrspace(4)
-  ; CHECK-LLVM: %[[UNADDRSPACECAST:.*]] = addrspacecast ptr addrspace(4) %[[ADDRSPACECAST]] to ptr
-  ; CHECK-LLVM: call void @llvm.var.annotation(ptr %[[UNADDRSPACECAST]], ptr [[STR12]], ptr undef, i32 undef, ptr undef)
+  ; CHECK-LLVM: call ptr addrspace(4) @llvm.ptr.annotation.p4(ptr addrspace(4) %[[ADDRSPACECAST]], ptr [[STR12]], ptr undef, i32 undef, ptr undef)
   %6 = bitcast i8 addrspace(4)* %5 to i32 addrspace(4)* addrspace(4)*
   %7 = addrspacecast i32 addrspace(1)* %0 to i32 addrspace(4)*
   store i32 addrspace(4)* %7, i32 addrspace(4)* addrspace(4)* %6, align 8, !tbaa !17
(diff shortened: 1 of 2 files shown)


MrSidims:
Could you please point out where can I learn about this restriction? Can't find it in https://llvm.org/docs/LangRef.html#llvm-var-annotation-intrinsic

dwoodwor-intel:
> Could you please point out where can I learn about this restriction? Can't find it in https://llvm.org/docs/LangRef.html#llvm-var-annotation-intrinsic

It's there in the signature:

```
declare void @llvm.var.annotation(ptr <val>, ptr <str>, ptr <str>, i32  <int>)
```

`ptr` is an address space 0 pointer; pointer types with other address spaces (like `ptr addrspace(1)` in the `llvm.ptr.annotation.*` syntax example) are distinct and incompatible types.

test/transcoding/annotate_attribute.ll
```diff
@@ -205,6 +211,30 @@ entry:
...
+  %2 = alloca %struct.MyIP, align 8
+  %3 = bitcast %struct.MyIP* %2 to i8*
+  call void @llvm.lifetime.start.p0i8(i64 8, i8* nonnull %3) #4
+  %4 = addrspacecast i8* %3 to i8 addrspace(4)*
+  %5 = call i8 addrspace(4)* @llvm.ptr.annotation.p4i8(i8 addrspace(4)* %4, i8* getelementptr inbounds ([4 x i8], [4 x i8]* @.str.11, i64 0, i64 0), i8* getelementptr inbounds ([9 x i8], [9 x i8]* @.str.1.12, i64 0, i64 0), i32 13, i8* bitcast ({ i32, i32, i32 }* @.args to i8*))
+  ; CHECK-LLVM: %[[ALLOCA:.*]] = alloca %struct.MyIP, align 8
+  ; CHECK-LLVM: %[[BITCAST:.*]] = bitcast ptr %[[ALLOCA]] to ptr
+  ; CHECK-LLVM: %[[ADDRSPACECAST:.*]] = addrspacecast ptr %[[BITCAST]] to ptr addrspace(4)
+  ; CHECK-LLVM: %[[UNADDRSPACECAST:.*]] = addrspacecast ptr addrspace(4) %[[ADDRSPACECAST]] to ptr
+  ; CHECK-LLVM: call void @llvm.var.annotation(ptr %[[UNADDRSPACECAST]], ptr [[STR12]], ptr undef, i32 undef, ptr undef)
```
MrSidims, line 225: In the original LLVM IR here is `llvm.ptr.annotation` and not `llvm.var.annotation`.

I believe the issue is that to access 0th member of the structure clang doesn't generate GEP(0, 0) and clang is correct in doing this, because GEP(0, 0) is quite useless instruction. So the algorithm in the reader should be adjusted - once we reach alloca it should check whether it's alloca for struct or not.

MrSidims:
Yeah, but `load` and `store` instructions don't have any mentions of addrspace in both syntax and example as well, yet we can definitely load from and store to various address spaces.

I've asked the question: https://github.com/llvm/llvm-project/issues/58227

test/transcoding/annotate_attribute.ll
dwoodwor-intel: Ah, I've run across this 0th struct member problem in other transforms before. I've updated the loop to only set `isStaticMemoryAttribute` if the base alloca is not a struct type.

dwoodwor-intel:
#1646 does seem like a better way of fixing this, so this PR can be closed

PR #1641, opened 2022-10-07
[Backport to 10] Fix wrong DICompositeType size (#1452)

stanleygambarin:
If count = -1 we shouldln't cast it to unsigned


MrSidims:
> Could you please also cherry-pick the tests or there are some additional dependencies not worth to resolve?

I'm OK to merge as is if there are extra dependencies.

stanleygambarin:
> > Could you please also cherry-pick the tests or there are some additional dependencies not worth to resolve?
> 
the test that came with the cherry-pick tests lower/upper bound definitions.  However, the concept of lower/upper bound definitions is not introduced until llvm_release_140 branch; implementing lower/upper bound in all previous branches would be  a much bigger task (and i am not sure it's warranted).

MrSidims:
I understand, thanks!

PR #1642, opened 2022-10-07
Correctly identify mangling of varargs in the demangler.

jcranmer-intel:


lib/SPIRV/SPIRVUtil.cpp
```diff
@@ -894,17 +894,37 @@ void getParameterTypes(Function *F, SmallVectorImpl<TypedPointerType *> &ArgTys,
...
+      if (auto *Name = dyn_cast<NameType>(Params[Params.size() - 1])) {
+        if (stringify(Name) == "...")
+          HasVarArgParam = true;
+      }
+    }
+    if (HasVarArgParam) {
+      Params = NodeArray(Params.begin(), Params.size() - 1);
+    } else {
+      LLVM_DEBUG(dbgs() << "[getParameterTypes] function " << MangledName
+                        << " was expected to have a varargs function type\n");
```
svenvh, line 912:
```suggestion
                        << " was expected to have a varargs parameter\n");
```

as we're inside the `if (F->isVarArg())` here.

Have you considered making this an `llvm_unreachable` instead of just a debug print?

MrSidims (commented):
LGTM

lib/SPIRV/SPIRVUtil.cpp
jcranmer-intel: Given that `getParameterTypes` will get called on potential user functions, I'm wary of making a path that causes aborts or `llvm_unreachable` on incorrectly mangled functions. (Hell, one of the things that made me discover this bug was someone generating an incorrect mangled name here).

svenvh: Sounds reasonable to leave it as is then, thanks!

svenvh (approved):

Issue #1548, opened 2022-07-12
Update tests containing unsupported constant expressions


svenvh:
LLVM is removing most constant expressions, see https://discourse.llvm.org/t/rfc-remove-most-constant-expressions/63179 , which is causing some test failures.  For now, failing tests have been XFAILed:

- https://github.com/KhronosGroup/SPIRV-LLVM-Translator/pull/1544
- https://github.com/KhronosGroup/SPIRV-LLVM-Translator/pull/1547

cc: @bader @asudarsa

MrSidims:
@vmaksimo could I please ask you to take a look at this and update/remove the tests when you have time

MrSidims:
Fixed by https://github.com/KhronosGroup/SPIRV-LLVM-Translator/pull/1644

PR #1644, opened 2022-10-11
Address issue #1548 after removal most of constant expressions in LLVM

vmaksimo:
- Rewrote test cases that were using float operations (`fadd`) to the supported instructions.
- Removed vector-metadata-constexpr.ll in favor of complex-constexpr-vector.ll, because it has both LLVM and SPIRV code checks.


vmaksimo:
@MrSidims @svenvh please take a look. Thanks!

svenvh (approved):

PR #1646, opened 2022-10-13
Emit OpMemberDecorate for member 0 annotations

dwoodwor-intel:
LLVMToSPIRVBase::transIntrinsicInst translates ptr.annotation intrinsics marking struct members by looking for struct GEPs. However, a GEPs is not necessary for the first member, and so in some cases when the first member is annotated SPIRVWriter has been emitting OpDecorate for these instead which leads to a crash in some cases on reverse translation. This change enables emitting OpMemberDecorate for the first member even if there isn't a GEP and the pointer value is based on a struct alloca instead.

This change also moves the handling of FPGARegIntel annotations so that they're translated to OpFPGARegINTEL before attempting to translate struct member annotations. This ensures that they don't get incorrectly translated with OpMemberDecorate.

### DIFF
--- a/lib/SPIRV/SPIRVWriter.cpp
+++ b/lib/SPIRV/SPIRVWriter.cpp
@@ -3761,8 +3761,8 @@ SPIRVValue *LLVMToSPIRVBase::transIntrinsicInst(IntrinsicInst *II,
     // to the first member of the struct; if so, attempt to get a struct type
     // from an alloca instead.
     SPIRVType *StructTy = nullptr;
-    SPIRVValue *ResPtr;
-    SPIRVWord MemberNumber;
+    [[maybe_unused]] SPIRVValue *ResPtr = nullptr;
+    [[maybe_unused]] SPIRVWord MemberNumber = 0;
     if (auto *const GI = dyn_cast<GetElementPtrInst>(AnnotSubj)) {
       if (auto *const STy = dyn_cast<StructType>(GI->getSourceElementType())) {
         StructTy = transType(STy);
--- a/lib/SPIRV/SPIRVWriter.cpp
+++ b/lib/SPIRV/SPIRVWriter.cpp
@@ -3745,15 +3745,38 @@ SPIRVValue *LLVMToSPIRVBase::transIntrinsicInst(IntrinsicInst *II,
     AnnotationDecorations Decorations =
         tryParseAnnotationString(BM, AnnotationString);
 
+    // Translate FPGARegIntel annotations to OpFPGARegINTEL.
+    if (AnnotationString == kOCLBuiltinName::FPGARegIntel) {
+      // TODO: Check for opaque pointer requirements.
+      auto *Ty = transType(II->getType());
+      auto *BI = dyn_cast<BitCastInst>(II->getOperand(0));
+      if (BM->isAllowedToUseExtension(ExtensionID::SPV_INTEL_fpga_reg))
+        return BM->addFPGARegINTELInst(Ty, transValue(BI, BB), BB);
+      return transValue(BI, BB);
+    }
+
     // If the pointer is a GEP on a struct, then we have to emit a member
     // decoration for the GEP-accessed struct, or a memory access decoration
-    // for the GEP itself.
-    auto *GI = dyn_cast<GetElementPtrInst>(AnnotSubj);
-    if (GI && isa<StructType>(GI->getSourceElementType())) {
-      auto *Ty = transType(GI->getSourceElementType());
-      auto *ResPtr = transValue(GI, BB);
-      SPIRVWord MemberNumber =
-          dyn_cast<ConstantInt>(GI->getOperand(2))->getZExtValue();
+    // for the GEP itself. There may not be a GEP in this case if the access is
+    // to the first member of the struct; if so, attempt to get a struct type
+    // from an alloca instead.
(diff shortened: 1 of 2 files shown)


dwoodwor-intel:
This is another approach to the problem in #1635 which seems to be a more proper fix

lib/SPIRV/SPIRVWriter.cpp
```diff
@@ -3745,15 +3745,38 @@ SPIRVValue *LLVMToSPIRVBase::transIntrinsicInst(IntrinsicInst *II,
...
-      auto *Ty = transType(GI->getSourceElementType());
-      auto *ResPtr = transValue(GI, BB);
-      SPIRVWord MemberNumber =
-          dyn_cast<ConstantInt>(GI->getOperand(2))->getZExtValue();
+    // for the GEP itself. There may not be a GEP in this case if the access is
+    // to the first member of the struct; if so, attempt to get a struct type
+    // from an alloca instead.
+    SPIRVType *StructTy = nullptr;
+    SPIRVValue *ResPtr;
+    SPIRVWord MemberNumber;
```
MrSidims, line 3765:
```suggestion
    [[maybe_unused]] SPIRVValue *ResPtr = nullptr;
    [[maybe_unused]] SPIRVWord MemberNumber = 0;
```

MrSidims (commented):
LGTM just one nit

MrSidims:
@asudarsa @vmaksimo could you please also take a look?

lib/SPIRV/SPIRVWriter.cpp
```diff
@@ -3745,15 +3745,38 @@ SPIRVValue *LLVMToSPIRVBase::transIntrinsicInst(IntrinsicInst *II,
     AnnotationDecorations Decorations =
         tryParseAnnotationString(BM, AnnotationString);
 
+    // Translate FPGARegIntel annotations to OpFPGARegINTEL.
```
asudarsa, line 3748: This change seems to be unrelated to the issue being addressed. Can you please comment? Thanks

lib/SPIRV/SPIRVWriter.cpp
```diff
@@ -3778,27 +3801,18 @@ SPIRVValue *LLVMToSPIRVBase::transIntrinsicInst(IntrinsicInst *II,
...
-              DecSubj, AnnotationString.c_str()));
-        else
-          // Apply the LSU parameter decoration to the pointer result of an
-          // instruction. Note it's the address to the accessed memory that's
-          // loaded from the original pointer variable, and not the value
-          // accessed by the latter.
-          addAnnotationDecorations(DecSubj, Decorations.MemoryAccessesVec);
-        II->replaceAllUsesWith(II->getOperand(0));
-      }
+      // Memory accesses to a standalone pointer variable
```
asudarsa, line 3804: Can you please specify if the scope of this change goes beyond addressing the first member of the struct variable? This change also seems unrelated. Thanks

lib/SPIRV/SPIRVWriter.cpp
dwoodwor-intel: This was something I discovered when running the tests: there are a few cases where this FPGARegIntel annotation is used on struct values, and my change was incorrectly trying to translate those as OpMemberDecorate instead. It seems like in cases where we have an FPGARegIntel annotation we want to translate it to OpFPGARegINTEL regardless of whether we find a struct type, so I thought the cleanest way of doing that would be moving this check up here.

@vmaksimo, from the git log it looks like you originally added this code; is putting it here correct?

lib/SPIRV/SPIRVWriter.cpp
dwoodwor-intel: I've just removed the OpFPGARegINTEL code from here; the rest of the changes in this part are just formatting

lib/SPIRV/SPIRVWriter.cpp
dwoodwor-intel: Done

lib/SPIRV/SPIRVWriter.cpp
asudarsa: Looks good. Thanks

MrSidims (approved):
CI issues don't relate with the patch
Checking them

asudarsa (approved):
LGTM. Please validate SPIR-V generated using spirv-val. Thanks

dwoodwor-intel:
> LGTM. Please validate SPIR-V generated using spirv-val. Thanks

I ran spirv-val on this test and did find an error:

```
$ spirv-val annotate_attribute.spv
error: line 236: Variables can not have a function[7] storage class outside of a function
  %_str = OpVariable %_ptr_Function__arr_uchar_ulong_3 Function %9
```

I also tried without this change on the main branch and I see the same error there too. It seems to be a preexisting issue unrelated to this change.

lib/SPIRV/SPIRVWriter.cpp
vmaksimo: @dwoodwor-intel Oh, that has been some long time ago, and I missed a big part of changes related to this feature transformation :) 

But I do think that now it's a correct place - it looks logical and has not failed any tests.

MrSidims:
llvm-bolt is a known issue reported in https://github.com/llvm/llvm-project/issues/58317 and several other places. Merging.

PR #1654, opened 2022-10-14
[Backport to 15] Integer dot product 4x8 packed translation

Quetzonarch:
Backport of #1633 with adjustments for llvm_release_150

This proposed bugfix updates the integer dot translation to use the correct function names from the OpenCL extension specification for packed format with scalar integer arguments (i.e. dot_4x8packed or dot_acc_sat_4x8packed instead of dot or dot_acc_sat with scalar integer arguments) to translate them into proper OpCodes. This is a fix for incorrect function names introduced in #1174 
Additionally removed unused variables from visitCallDot to clean up the code.
Link to OpenCL spec:
https://registry.khronos.org/OpenCL/specs/3.0-unified/html/OpenCL_Ext.html#cl_khr_integer_dot_product


MrSidims (approved):
clang-tidy error with "IsAccSat can be const"
Yet https://github.com/KhronosGroup/SPIRV-LLVM-Translator/pull/1633 was merged without such change and without such error. Hence merging this PR as is.

PR #1655, opened 2022-10-14
Reorder SPIR-V blocks to emit dominators first

### LINKED ISSUE #205: Incorrect block ordering in generated SPIR-V
The SPIR-V produced for the following module:

```
source_filename = "reproducer.cl"
target datalayout = "e-i64:64-v16:16-v24:32-v32:32-v48:64-v96:128-v192:256-v256:256-v512:512-v1024:1024"

define spir_kernel void @test(i8 addrspace(1)* %arg) local_unnamed_addr #0 !kernel_arg_addr_space !1 !kernel_arg_access_qual !2 !kernel_arg_type !3 !kernel_arg_base_type !3 !kernel_arg_type_qual !4 {
entry:
  br label %for.body

for.body137.lr.ph:                                ; preds = %for.body
  ret void

for.body:                                         ; preds = %for.body, %entry
  br i1 undef, label %for.body, label %for.body137.lr.ph
}

attributes #0 = { "use-soft-float"="false" }
attributes #1 = { nounwind readnone speculatable }

!llvm.ident = !{!0}

!0 = !{!"clang version 9.0.0 "}
!1 = !{i32 1}
!2 = !{!"none"}
!3 = !{!"uchar*"}
!4 = !{!""}
```

doesn't satisfy the block ordering rules (2.16.1 in the 1.4 unified spec):

```
The order of blocks in a function must satisfy the rule that blocks appear before all blocks they dominate.
```

The SPIR-V generated is:
```
; SPIR-V                                                                                                                             
; Version: 1.0                                                                                                                       
; Generator: Khronos LLVM/SPIR-V Translator; 14                                                                                      
; Bound: 14
; Schema: 0
               OpCapability Addresses
               OpCapability Kernel
               OpCapability Int8
          %1 = OpExtInstImport "OpenCL.std"
               OpEntryPoint Kernel %6 "test"
         %13 = OpString "kernel_arg_type.test.uchar*,"
               OpSource Unknown 0
               OpName %arg "arg"
               OpName %entry "entry"
               OpName %for_body137_lr_ph "for.body137.lr.ph"
               OpName %for_body "for.body"
      %uchar = OpTypeInt 8 0
       %void = OpTypeVoid
%_ptr_CrossWorkgroup_uchar = OpTypePointer CrossWorkgroup %uchar
          %5 = OpTypeFunction %void %_ptr_CrossWorkgroup_uchar
       %bool = OpTypeBool
         %12 = OpUndef %bool
          %6 = OpFunction %void None %5
        %arg = OpFunctionParameter %_ptr_CrossWorkgroup_uchar
      %entry = OpLabel
               OpBranch %for_body
%for_body137_lr_ph = OpLabel
               OpReturn
   %for_body = OpLabel
               OpBranchConditional %12 %for_body %for_body137_lr_ph
               OpFunctionEnd
```

which the validator rejects:

```
error: line 25: Block 9[%for_body137_lr_ph] appears in the binary before its dominator 10[%for_body]
  %for_body = OpLabel
```

dwoodwor-intel:
This implements the block ordering rule in section 2.16.1 of the SPIR-V specification:

> The order of blocks in a function must satisfy the rule that blocks appear
before all blocks they dominate.

This fixes #205.

### DIFF
--- a/lib/SPIRV/SPIRVWriter.cpp
+++ b/lib/SPIRV/SPIRVWriter.cpp
@@ -4387,34 +4387,38 @@ void LLVMToSPIRVBase::fpContractUpdateRecursive(Function *F, FPContract FPC) {
 /// them to visit dominators ahead of their dominated blocks when needed. \p DT
 /// is not copied by this function and needs to outlive any iterators created
 /// from this range.
-static auto stablePreDominatorTraversal(Function& F, const DominatorTree& DT) {
+static auto stablePreDominatorTraversal(Function &F, const DominatorTree &DT) {
 
   // A local iterator type for traversing a function in the desired order.
-  class StablePreDominatorIterator : public iterator_facade_base<StablePreDominatorIterator, std::forward_iterator_tag, BasicBlock> {
+  class StablePreDominatorIterator
+      : public iterator_facade_base<StablePreDominatorIterator,
+                                    std::forward_iterator_tag, BasicBlock> {
 
     // The passed DominatorTree; may be unset for end iterators.
-    const DominatorTree* DT;
+    const DominatorTree *DT;
 
     // The set of basic blocks already visited in this traversal.
-    DenseSet<const BasicBlock*> VisitedBBs;
+    DenseSet<const BasicBlock *> VisitedBBs;
 
     // The next basic block in original function order, or nullptr if the
     // traversal is over.
-    BasicBlock* NextBB = nullptr;
+    BasicBlock *NextBB = nullptr;
 
     // The current basic block in the traversal, or nullptr for end iterators.
-    BasicBlock* CurBB = nullptr;
+    BasicBlock *CurBB = nullptr;
 
     // Returns the most immediate dominator of \p BB which does not have an
     // unvisited dominator and so can be visited in this traversal.
-    BasicBlock* visitableDominator(BasicBlock* BB) const {
+    BasicBlock *visitableDominator(BasicBlock *BB) const {
 
       // Find BB's dominator; if there is none, BB can be visited immediately.
-      const auto*const BBNode = DT->getNode(BB);
-      if (!BBNode) return BB;
-      const auto*const DomNode = BBNode->getIDom();
-      if (!DomNode) return BB;
-      BasicBlock*const Dominator = DomNode->getBlock();
+      const auto *const BBNode = DT->getNode(BB);
+      if (!BBNode)
+        return BB;
+      const auto *const DomNode = BBNode->getIDom();
+      if (!DomNode)
+        return BB;
+      BasicBlock *const Dominator = DomNode->getBlock();
 
       // If the dominator's been visited, BB can now be visited.
       if (VisitedBBs.count(Dominator))
@@ -4426,10 +4430,11 @@ static auto stablePreDominatorTraversal(Function& F, const DominatorTree& DT) {
 
     // Advances the iterator and returns the next basic block to be visited in
     // the traversal.
-    BasicBlock* next() {
+    BasicBlock *next() {
 
       // If NextBB is nullptr, the end of the traversal has been reached.
-      if (!NextBB) return nullptr;
+      if (!NextBB)
+        return nullptr;
 
       // Check if NextBB has already been visited; if so, advance past it.
       if (VisitedBBs.count(NextBB)) {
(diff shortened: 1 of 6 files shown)


test/create-placeholders-for-phi-operands.ll
```diff
@@ -31,7 +31,7 @@ BB.0:
   br i1 %or.cond, label %BB.2, label %BB.3
 
 BB.1:                                             ; preds = %BB.12, %BB.7
-  %savedstack.sink = phi i8* [ %savedstack, %BB.12 ], [ %savedstack.us, %BB.7 ]
+  %savedstack.sink = phi i8* [ %savedstack, %BB.12.loopexit ], [ %savedstack.us, %BB.11.loopexit ]
```
MrSidims, line 34: Could you please explain this change? TBH I wouldn't expect source IR for test to be changed - only IR after the translation.
This comment applies to every other IR change in the tests.

dwoodwor-intel: My assumption from the name of this test is that the intention is to make sure there aren't problems with round-tripping phi nodes that appear before their non-dominating operands in the function. The problem that I ran into is that the translator adds these two loopexit blocks ahead of `BB.1` and dominator ordering pulls all of the loopexit blocks' dominators ahead of them (including the ones that define `savedstack` and `savedstack.us`). The result is that this phi no longer appears before the non-dominating operands it references, which seems to defeat the purpose of this test. My fix here is to create these loopexit blocks manually before the translator can and to place them later in the function so that they don't pull the phi operands ahead of the phi.

test/transcoding/SPV_INTEL_variable_length_array/complex-cfg.ll
```diff
@@ -21,6 +21,25 @@ newFuncRoot:
   %temp = alloca i32, align 4
   br label %fallthru
 
+; CHECK-LABEL: fallthru
+; CHECK-LLVM: %"ascastB$val41" = alloca i32, i64 %div.3
+fallthru:                          ; preds = %newFuncRoot
```
dwoodwor-intel, line 26: This one's a simpler case: this block which dominates most of the function was originally near the end, so there isn't any way to get around moving it. The test is written in a style where CHECK lines are next to the corresponding input code. Since I don't think this test is intended to test dominator ordering, it seems clearer to just move this block up so that it doesn't need to be reordered rather than moving just the CHECK lines by themselves.

lib/SPIRV/SPIRVWriter.cpp
```diff
@@ -4378,10 +4378,108 @@ void LLVMToSPIRVBase::fpContractUpdateRecursive(Function *F, FPContract FPC) {
...
+      const auto *const BBNode = DT->getNode(BB);
+      if (!BBNode)
+        return BB;
+      const auto *const DomNode = BBNode->getIDom();
+      if (!DomNode)
+        return BB;
+      BasicBlock *const Dominator = DomNode->getBlock();
+
+      // If the dominator's been visited, BB can now be visited.
+      if (VisitedBBs.count(Dominator))
```
svenvh, line 4424:
```suggestion
      if (VisitedBBs.contains(Dominator))
```

lib/SPIRV/SPIRVWriter.cpp
```diff
@@ -4378,10 +4378,108 @@ void LLVMToSPIRVBase::fpContractUpdateRecursive(Function *F, FPContract FPC) {
...
+  // A local iterator type for traversing a function in the desired order.
+  class StablePreDominatorIterator
+      : public iterator_facade_base<StablePreDominatorIterator,
+                                    std::forward_iterator_tag, BasicBlock> {
+
+    // The passed DominatorTree; may be unset for end iterators.
+    const DominatorTree *DT;
+
+    // The set of basic blocks already visited in this traversal.
+    DenseSet<const BasicBlock *> VisitedBBs;
```
svenvh, line 4401: Have you considered using a `SmallPtrSet` instead of a `DenseSet`?

lib/SPIRV/SPIRVWriter.cpp
svenvh: Same for the `next()` method.

dwoodwor-intel: Done

lib/SPIRV/SPIRVWriter.cpp
dwoodwor-intel: `SmallPtrSet` is probably a better choice here; I've switched to using that

MrSidims (approved):
LGTM

svenvh (approved):

dwoodwor-intel:
The out-of-tree build failures don't look related to this change; are they part of a known testing issue?

MrSidims:
> The out-of-tree build failures don't look related to this change; are they part of a known testing issue?

It's known packaging issue in llvm, it was fixed for 15.0.3 but not for 16.0.0. I was hoping it could be resolved by EOW so we can run clang-format, but it looks like it's not the case.

PR #1656, opened 2022-10-17
Implement SPV_INTEL_tensor_float32_conversion extension

MrSidims:
This extension adds conversion instruction from float to tensor float (TF32) data format. TF32 uses 1 bit for a sign, 8 bits for an exponent and 10 bits for a fraction. This extension doesn’t introduce TF32 type in SPIR-V, instead instruction below uses 32-bit float type to represent TF32 value.

Spec: https://github.com/intel/llvm/pull/6990


MrSidims:
@asudarsa please take a look

test/transcoding/SPV_INTEL_tensor_float32_conversion/convert_tensor_float32.ll
```diff
@@ -0,0 +1,50 @@
+; RUN: llvm-as %s -o %t.bc
```
MrSidims, line 1: the test should be moved to new dir: extensions

lib/SPIRV/libSPIRV/SPIRVInstruction.h
```diff
@@ -3561,6 +3561,64 @@ class SPIRVMaskedScatterINTELInst
...
+      InCompTy = InCompTy->getVectorComponentType();
+    }
+
+    auto InstName = OpCodeNameMap::map(OC);
+    SPIRVErrorLog &SPVErrLog = this->getModule()->getErrorLog();
+
+    SPVErrLog.checkError(
+        ResCompTy->isTypeFloat(32), SPIRVEC_InvalidInstruction,
+        InstName + "\nResult value must be a scalar or vector of integer "
+                   "16-bit type\n");
```
vmaksimo, line 3606: Does the result value need to be a 16-bit integer? I believe we check for 32-bit float here, please fix the message.

vmaksimo (commented):
LGTM.
I'm not sure how to fix unrelated clang-tidy fix, but at least please resolve clang-format hit.

lib/SPIRV/libSPIRV/SPIRVInstruction.h
MrSidims: Thanks!

MrSidims:
incidentally amended the change, sorry

MrSidims:
LIT failures are related to commit 304f1d59ca41872c094def3aee0a8689df6aa398
Author: Nikita Popov <npopov@redhat.com>
Date:   Wed Nov 2 16:46:56 2022 +0100

    [IR] Switch everything to use memory attribute

    This switches everything to use the memory attribute proposed in
    https://discourse.llvm.org/t/rfc-unify-memory-effect-attributes/65579.
    The old argmemonly, inaccessiblememonly and inaccessiblemem_or_argmemonly
    attributes are dropped. The readnone, readonly and writeonly attributes
    are restricted to parameters only.

PR #1657, opened 2022-10-17
[NFC] Remove Intel Op codes that are being upstreamed

MrSidims:
Follows https://github.com/KhronosGroup/SPIRV-Headers/pull/298
Had to keep internal::DecorationRuntimeAlignedINTEL since the extension
implementation was done none conformant to the spec.


MrSidims:
@asudarsa @vmaksimo please take a look

MrSidims:
Found out, that runtime aligned feature is not conformant to the spec, had to keep the appropriate decoration in `internal `namespace

MrSidims:
@asudarsa @vmaksimo friendly ping

lib/SPIRV/libSPIRV/SPIRVNameMapEnum.h
```diff
@@ -182,10 +182,10 @@ template <> inline void SPIRVMap<Decoration, std::string>::init() {
...
   // From spirv_internal.hpp
   add(internal::DecorationCallableFunctionINTEL, "CallableFunctionINTEL");
-  add(internal::DecorationMathOpDSPModeINTEL, "MathOpDSPModeINTEL");
-  add(internal::DecorationInitiationIntervalINTEL, "InitiationIntervalINTEL");
-  add(internal::DecorationMaxConcurrencyINTEL, "MaxConcurrencyINTEL");
-  add(internal::DecorationPipelineEnableINTEL, "PipelineEnableINTEL");
+  add(DecorationMathOpDSPModeINTEL, "MathOpDSPModeINTEL");
+  add(DecorationInitiationIntervalINTEL, "InitiationIntervalINTEL");
+  add(DecorationMaxConcurrencyINTEL, "MaxConcurrencyINTEL");
+  add(DecorationPipelineEnableINTEL, "PipelineEnableINTEL");
```
vmaksimo, line 188: It says "from spirv_internal.hpp", but as I understood the decorations are now in spirv.hpp header. Should we move this to other place?
(same for capabilities)

lib/SPIRV/libSPIRV/spirv_internal.hpp
```diff
@@ -91,10 +84,7 @@ enum InternalExecutionMode {
   IExecModeStreamingInterfaceINTEL = 6154
 };
 
-enum InternalLoopControlMask {
-  ILoopControlLoopCountINTELMask = 0x1000000,
-  ILoopControlMaxReinvocationDelayINTELMask = 0x2000000
-};
+enum InternalLoopControlMask {};
```
vmaksimo, line 87: Could you please clarify why we still need this trick?

MrSidims: We don't need it, but we might add new Masks here in the future, so I don't want to remove enum

lib/SPIRV/libSPIRV/SPIRVNameMapEnum.h
MrSidims: Thanks! I'll move the code

lib/SPIRV/libSPIRV/spirv_internal.hpp
vmaksimo: I see your point, but it still looks strange to me. There is a chance that this line would not be found by a developer and a new enum will be created.

vmaksimo (commented):
LGTM, but possibly need to resolve the discussion first.

lib/SPIRV/libSPIRV/spirv_internal.hpp
vmaksimo: I see your point, but it still looks strange to me. There is a chance that this line would not be found by a developer and a new enum will be created.

MrSidims: Agree, removed the enum

vmaksimo (approved):

PR #1659, opened 2022-10-18
[NFC] Add opening parenthesis to call checks

svenvh:
The affected CHECK lines ended with a function name.  Any spurious trailing characters in those function names would not be detected by the test.  Append the opening parenthesis of the call expression to catch any undesired trailing characters.

This commit only affects tests that do not require any changes besides the trailing parenthesis.


MrSidims (approved):

MrSidims:
Seems like [llvm-bolt](https://github.com/llvm/llvm-project/issues/58317) issue is not yet resolved

PR #1666, opened 2022-10-20
Translate constant global variable references correctly in opaque pointer mode.

jcranmer-intel:


lib/SPIRV/SPIRVWriter.cpp
```diff
@@ -1034,6 +1034,32 @@ void LLVMToSPIRVBase::transFPGAFunctionMetadata(SPIRVFunction *BF,
...
+
+  assert(C->getType()->isPointerTy() &&
+         "Only pointer type mismatches should be possible");
+  // In the common case of strings ([N x i8] GVs), see if we can emit a GEP
+  // instruction.
+  if (auto *GV = dyn_cast<GlobalVariable>(C)) {
+    if (GV->getValueType()->isArrayTy() &&
+        GV->getValueType()->getArrayElementType()->isIntegerTy(8)) {
+      SPIRVValue *Offset = transValue(getUInt32(M, 0), nullptr);
+      return BM->addPtrAccessChainInst(ExpectedType, Trans, {Offset, Offset},
```
MrSidims, line 1054: Please add a test case for this part of the code, aka add check for SPIR-V's PtrAcessChain in your test. But I have a feeling, that this check won't trigger.

jcranmer-intel: The new test that is added does check this code--if I add an `assert(false);` before this line, transcoding/constant-vars.ll fails.

test/transcoding/constant-vars.ll
```diff
@@ -0,0 +1,53 @@
...
+
+@astr = internal addrspace(2) constant [7 x i8] c"string\00", align 4
+; CHECK-SPIRV: 5 Variable {{[0-9]+}} [[ASTR:[0-9]+]] 0
+
+@i64arr = addrspace(1) constant [3 x i64] [i64 0, i64 1, i64 2]
+; CHECK-SPIRV: 5 Variable {{[0-9]+}} [[I64ARR:[0-9]+]] 5
+
+@struct = addrspace(1) global {ptr addrspace(2), ptr addrspace(1)} { ptr addrspace(2) @astr, ptr addrspace(1) @i64arr }
+
+; CHECK-SPIRV: 7 SpecConstantOp [[AS2]] [[ASTRC:[0-9]+]] 70 [[ASTR]] [[I320]] [[I320]]
```
jcranmer-intel, line 30: Specifically, this line is a `SpecConstantOp PtrInBoundsAccessChain`. The latter has an opcode of 70, and the llvm-spirv --to-text output doesn't do enough prettifying of names to recognize that this parameter is a SPIR-V opcode.

MrSidims (approved):

lib/SPIRV/SPIRVWriter.cpp
MrSidims: Got it, thanks!

svenvh (approved):

PR #1669, opened 2022-10-24
Restore GEP instruction translation for vector of pointers

MrSidims:
Like now PtrAccessChain has a pointer as a result type, but should have a vector of
pointers result type.


MrSidims:
@jcranmer-intel please take a look

MrSidims:
FYI @stanleygambarin

lib/SPIRV/SPIRVWriter.cpp
```diff
@@ -2072,9 +2072,11 @@ LLVMToSPIRVBase::transValueWithoutDecoration(Value *V, SPIRVBasicBlock *BB,
         IndexGroupArrayMap[ContainedIndexGroup].insert(AccessedArrayId);
       }
     }
-
-    SPIRVType *TranslatedTy = transPointerType(
-        GEP->getResultElementType(), GEP->getType()->getPointerAddressSpace());
+    SPIRVType *TranslatedTy =
+        GEP->getType()->isVectorTy()
+            ? transType(GEP->getType())
```
jcranmer-intel, line 2077: This isn't right, though, in opaque-pointer mode, right? `GEP->getType()` would be `<4 x ptr>`, which would get translated as if it were `<4 x i8*>`. Instead, for vector types, you'd want to wrap the result of `transPointerType(...)` into a vector type--`transType(VectorType::get(TypedPointerType::get()))` is probably the cleanest way to get it. (Admittedly, vector-of-pointer types is likely to break elsewhere in opaque pointer mode anyways--the type scavenger for sure can't handle it right now).

MrSidims: Thanks! Probably the latest code is the one we need.
There is some inefficiency though in opaque pointers mode: I see
```
    %1 = bitcast <4 x ptr addrspace(1)> %tmp2 to <4 x ptr addrspace(1)>
    %2 = bitcast <4 x ptr addrspace(1)> %1 to <4 x ptr addrspace(1)>
```
after reverse translation. The second bitcast relates with this patch, while the first one does not. May be it's worth to run DCE pass after the translation, since TBH I don't see a good way to eliminate these useless bitcasts.

MrSidims:
The failing test is looking like to be a bug in LLVM:
```
FixedVectorType *FixedVectorType::get(Type *ElementType, unsigned NumElts) {
assert(NumElts > 0 && "#Elements of a VectorType must be greater than 0");
assert(isValidElementType(ElementType) && "Element type of a VectorType must " // here we fail


GDB output
(gdb) p ElementType->dump()
typedptr(i32, 1)
$7 = void
(gdb) p ElementType->isPointerTy()
$8 = false
(gdb) p ElementType->getTypeID()
$9 = llvm::Type::TypedPointerTyID
```

lib/SPIRV/SPIRVWriter.cpp
MrSidims: Fixed. The tests are now passing in debug as well. @jcranmer-intel please take another look when you have time.

test/extensions/INTEL/intel-basic-vector-pointers-opaque.ll
```diff
@@ -0,0 +1,80 @@
+; RUN: llvm-as %s -o %t.bc
+; RUN: llvm-spirv %t.bc --spirv-ext=+SPV_INTEL_masked_gather_scatter -o %t.spv
+; RUN: llvm-spirv %t.spv --to-text -o %t.spt
+; RUN: FileCheck < %t.spt %s --check-prefix=CHECK-SPIRV
+
+; RUN: llvm-spirv -r %t.spv -o %t.rev.bc
```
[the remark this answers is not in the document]
MrSidims: Added

PR #1673, opened 2022-10-26
[Backport to 15] Translate llvm.loop.unroll.full metadata

MrSidims:
It can be generated via #pragma clang unroll(full) pragma. llvm.loop.unroll.full means attempt to do full unroll of the loop and disable the unrolling if the trip count is not known at compile time.

Unroll mask to which it was previously mapped doesn't much the description.

The way the patch represents it in SPIR-V is:
 Unroll mask + PartialCount mask with '1' parameter

This patch also removes some overtesting for unroll metadata.

This backports: https://github.com/KhronosGroup/SPIRV-LLVM-Translator/pull/1664


MrSidims:
@svenvh may I ask you for help?
The pre-commit is failing with:
"E: The repository 'https://packages.lunarg.com/vulkan bionic Release' does not have a Release file."
I took a look and see, that a new vulkan package 1.3.231 was added on 25-Oct-2022 14:58 . And this version actually present in https://packages.lunarg.com/vulkan/releases.json , so for me the error seems to be a bit odd. Do you know what is happening here?

svenvh:
I am suspecting that they are no longer providing packages for Bionic (i.e. Ubuntu 18.04).  https://packages.lunarg.com/vulkan/dists.json only lists Focal and Jammy.  I suppose we'll have to cherry-pick 20a1fd783817b5399a1349eeedaac539e4619b00 to the release branches?

MrSidims:
Oh, I see, thanks! Since I'm the most interest in this at this point of time - I will create cherry-picks on my own.

PR #1683, opened 2022-10-27
[NFCI] Use TypedPointerType in more places.

jcranmer-intel:
One of the reasons behind this change is to make code more easily deal with the future prospect of opaque types, so that helper methods (like adjusting image types) can handle both pointer type and opaque type representations simply by querying if the input type is a TypedPointerType or an OpaqueType [name for the latter still pending].

The set of changes are:
* OCLTypeToSPIRV now uses TypedPointerType internally
* adaptSPIRVImageType and getSPIRVStructTypeByChangeBaseTypeName are collapsed into one method (adjustImageType) that works with TypedPointerTypes.
* A few is*StructType methods have been reverted back to is*Type methods, taking a TypedPointerType parameter instead.
* BuiltinCallHelper::addSPIRVCall{Pair} allows for the creation of SPIR-V calls that can use TypedPointerType or actual type for parameters and return value.
* BuiltinCallHelper::getCallValue{Type} is a simple helper that hides many of the uses of getParameterTypes.
* The Type* parameter of the callback in BuiltinCallMutator::mapArg now provides a TypedPointerType or the actual type, instead of the pointer element type.
* BuiltinCallMutator::ValueTypePair similarly takes a TypedPointerType or the actual type.
* getParameterTypes (when passed a SmallVector<Type *>) does a similar thing. (The SmallVector<TypedPointerType *> variant has been removed in favor of only using the other one.)

Note that the last few changes do change the semantics of function parameters without changing the function name or signature.

### DIFF
--- a/lib/SPIRV/SPIRVBuiltinHelper.cpp
+++ b/lib/SPIRV/SPIRVBuiltinHelper.cpp
@@ -266,7 +266,7 @@ Type *BuiltinCallHelper::adjustImageType(Type *T, StringRef OldImageKind,
 BuiltinCallMutator::ValueTypePair
 BuiltinCallHelper::getCallValue(CallInst *CI, unsigned ArgNo) {
   Function *CalledFunc = CI->getCalledFunction();
-  assert(CalledFunc && "Call must be to a builtin function");
+  assert(CalledFunc && "Unexpected indirect call");
   if (CalledFunc != CachedFunc) {
     CachedFunc = CalledFunc;
     [[maybe_unused]] bool DidDemangle =
--- a/lib/SPIRV/SPIRVBuiltinHelper.cpp
+++ b/lib/SPIRV/SPIRVBuiltinHelper.cpp
@@ -64,9 +64,15 @@ BuiltinCallMutator::BuiltinCallMutator(
     : CI(CI), FuncName(FuncName),
       Attrs(CI->getCalledFunction()->getAttributes()), ReturnTy(CI->getType()),
       Args(CI->args()), Rules(Rules), Builder(CI) {
-  getParameterTypes(CI->getCalledFunction(), PointerTypes,
-                    std::move(NameMapFn));
-  PointerTypes.resize(Args.size(), nullptr);
+  bool DidDemangle = getParameterTypes(CI->getCalledFunction(), PointerTypes,
+                                       std::move(NameMapFn));
+  if (!DidDemangle) {
+    // TODO: PipeBlocking.ll causes demangling failures.
+    // assert(isNonMangledOCLBuiltin(CI->getCalledFunction()->getName()) &&
+    //    "SPIR-V builtin functions should be mangled");
+    for (Value *Arg : Args)
+      PointerTypes.push_back(Arg->getType());
+  }
 }
 
 BuiltinCallMutator::BuiltinCallMutator(BuiltinCallMutator &&Other)
@@ -84,9 +90,15 @@ Value *BuiltinCallMutator::doConversion() {
   assert(CI && "Need to have a call instruction to do the conversion");
   auto Mangler = makeMangler(CI, Rules);
   for (unsigned I = 0; I < Args.size(); I++) {
-    Mangler->getTypeMangleInfo(I).PointerTy = PointerTypes[I];
+    Mangler->getTypeMangleInfo(I).PointerTy =
+        dyn_cast<TypedPointerType>(PointerTypes[I]);
   }
   assert(Attrs.getNumAttrSets() <= Args.size() + 2 && "Too many attributes?");
+
+  // Sanitize the return type, in case it's a TypedPointerType.
+  if (auto *TPT = dyn_cast<TypedPointerType>(ReturnTy))
+    ReturnTy = PointerType::get(TPT->getElementType(), TPT->getAddressSpace());
+
   CallInst *NewCall =
       Builder.Insert(addCallInst(CI->getModule(), FuncName, ReturnTy, Args,
                                  &Attrs, nullptr, Mangler.get()));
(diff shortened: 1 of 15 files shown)


lib/SPIRV/SPIRVBuiltinHelper.cpp
```diff
@@ -64,9 +64,15 @@ BuiltinCallMutator::BuiltinCallMutator(
...
       Attrs(CI->getCalledFunction()->getAttributes()), ReturnTy(CI->getType()),
       Args(CI->args()), Rules(Rules), Builder(CI) {
-  getParameterTypes(CI->getCalledFunction(), PointerTypes,
-                    std::move(NameMapFn));
-  PointerTypes.resize(Args.size(), nullptr);
+  bool DidDemangle = getParameterTypes(CI->getCalledFunction(), PointerTypes,
+                                       std::move(NameMapFn));
+  if (!DidDemangle) {
+    // TODO: PipeBlocking.ll causes demangling failures.
+    // assert(isNonMangledOCLBuiltin(CI->getCalledFunction()->getName()) &&
```
MrSidims, line 71: It does look like that if a function name has `.1` in the end (for example 
`declare spir_func void @_Z30__spirv_WritePipeBlockingINTELPU3AS115__spirv_Pipe__1PU3AS4iii.1(%opencl.pipe_wo_t addrspace(1)* %0, i9 addrspace(4)* %1, i32 %2, i32 %3)`
RootNode will be `nullptr`
`auto *RootNode = dyn_cast_or_null<FunctionEncoding>(Demangler.parse());`
unsure if it's expected

lib/SPIRV/SPIRVBuiltinHelper.cpp
```diff
@@ -211,3 +210,71 @@ BuiltinCallMutator BuiltinCallHelper::mutateCallInst(CallInst *CI,
...
+    StringRef Postfixes;
+    if (isSPIRVStructType(StructTy, OldImageKind, &Postfixes))
+      StructTy = getOrCreateOpaqueStructType(
+          M, getSPIRVTypeName(NewImageKind, Postfixes));
+    else {
+      report_fatal_error("Type did not have expected image kind");
+    }
+    return TypedPointerType::get(StructTy, TypedPtrTy->getAddressSpace());
+  }
+  report_fatal_error("Expected type to be a SPIRV image type");
```
MrSidims, line 263: nit: May be worth to move to an early exit, like:
```
if (!isa<TypedPointerType>(T))
  report_fatal_error("Expected type to be a SPIRV image type");
if (auto *TypedPtrTy = cast<TypedPointerType>(T)) {
}
```

lib/SPIRV/SPIRVBuiltinHelper.cpp
```diff
@@ -211,3 +210,71 @@ BuiltinCallMutator BuiltinCallHelper::mutateCallInst(CallInst *CI,
...
+}
+
+BuiltinCallMutator::ValueTypePair
+BuiltinCallHelper::getCallValue(CallInst *CI, unsigned ArgNo) {
+  Function *CalledFunc = CI->getCalledFunction();
+  assert(CalledFunc && "Call must be to a builtin function");
```
MrSidims, line 269: what about 
```suggestion
  assert(CalledFunc && "Unexpected indirect call");
```
?

lib/SPIRV/SPIRVUtil.cpp
```diff
@@ -836,14 +827,14 @@ parseNode(Module *M, const llvm::itanium_demangle::Node *ParamType,
...
-void getParameterTypes(Function *F, SmallVectorImpl<TypedPointerType *> &ArgTys,
+bool getParameterTypes(Function *F, SmallVectorImpl<Type *> &ArgTys,
                        std::function<std::string(StringRef)> NameMapFn) {
   using namespace llvm::itanium_demangle;
   // If there's no mangled name, we can't do anything. Also, if there's no
   // parameters, do nothing.
   StringRef Name = F->getName();
   if (!Name.startswith("_Z") || F->arg_empty())
-    return;
+    return Name.startswith("_Z");
```
MrSidims, line 837: Not sure if I understand, this line. We don't want to just return `false `in case if `F->arg_empty()`, but why?

lib/SPIRV/SPIRVUtil.cpp
```diff
@@ -892,7 +883,7 @@ void getParameterTypes(Function *F, SmallVectorImpl<TypedPointerType *> &ArgTys,
   // name encoding, bail out.
   auto *RootNode = dyn_cast_or_null<FunctionEncoding>(Demangler.parse());
```
MrSidims, line 884: So there is from where we get false to trigger the assert.

lib/SPIRV/SPIRVBuiltinHelper.cpp
jcranmer-intel: I intend for there to be a future
```c++
if (auto *OpaqueTy = dyn_cast<OpaqueType>(T)) {
  // Whee, do more stuff!
}
```
if statement, so I'm averse to the early exit here.

lib/SPIRV/SPIRVUtil.cpp
jcranmer-intel: Some builtins take no arguments, so if the function was a `_Z3foov`, we don't want to say it was unsuccessfully demangled.

lib/SPIRV/SPIRVBuiltinHelper.cpp
jcranmer-intel: If the `.1` is added, then the function name isn't a valid mangled name which is... probably problematic, but that requires some more investigation.

MrSidims: Agree

lib/SPIRV/SPIRVBuiltinHelper.cpp
MrSidims: Got it, thanks!

MrSidims (approved):

lib/SPIRV/SPIRVUtil.cpp
MrSidims: Thanks for the explanation!

jcranmer-intel:
Friendly review ping?

svenvh:
> Friendly review ping?

Sorry for the delay, I was already working through it, but as it has a lot of scattered changes I'll need some more time.

svenvh (approved):

PR #1684, opened 2022-10-31
Disable 2 tests that FAIL with clang-8

stanleygambarin:


test/transcoding/spec_const.ll
```diff
@@ -1,3 +1,7 @@
+; clang-8 does not generate 'OpCapability Float16', leading to failure in 'spirv-val' step
```
svenvh, line 1: > clang-8 does not generate 'OpCapability Float16'

This confuses me. This is a `.ll` test, so clang is not involved here as far as I can see?

Perhaps we're missing a fix on the llvm_release_80 branch; the branch is quite old, but if you still care about it, then it might be worth identifying how this was fixed in newer branches, instead of just xfailing the test.

stanleygambarin: https://github.com/KhronosGroup/SPIRV-LLVM-Translator/pull/1686 created with updated backport.

PR #1698, opened 2022-11-07
[CI] Upgrade to Ubuntu 20.04

MrSidims:
The Ubuntu 18.04 image is marked deprecated [1], so move to a newer image.

[1] https://github.com/actions/runner-images


svenvh (approved):
LGTM

We should probably fix the build error, I guess that appeared due to using a newer compiler version.

PR #1540, opened 2022-07-07
Emit 0/1 for annotations with boolean value

asudarsa:
Signed-off-by: Arvind Sudarsanam <arvind.sudarsanam@intel.com>

PR #1714, opened 2022-11-09
Emit 0/1 for annotations with boolean value

vmaksimo:
This continues #1540 by adding a test for the fix.


vmaksimo:
@svenvh @MrSidims could you please take a look? You've already approved the code change in the previous PR, I just added a required test for it.

PR #1718, opened 2022-11-11
[Backport to 15] Restore GEP instruction translation for vector of pointers (#1669)

MrSidims:
Make GEP returning vector of pointers in this case


MrSidims:
No need for this

PR #1721, opened 2022-11-16
[Backport to 8] Fix several loop metadata translation issues

MrSidims:
Has several patches:
1. Replace assumptions during llvm.loop translation with checks
2. Fix loop metadata layout in reverse translation
3. Fix loss of loop metadata during reverse translation
4. Translate llvm.loop.unroll.full metadata


MrSidims:
All tests have passed, fixed clang-format issue

PR #1724, opened 2022-11-18
Relax OpenCL extended instruction restrictions

vmustya:
When VectorAnyINTEL capability is enabled, OpenCL extended instructions
are able to work on vector types with any number of components greater
then or equal to 2.


MrSidims (commented):
Probably some OpenCL environmental spec is required, but as a start the code is LGTM.

Please add some tests for this.

MrSidims (approved):

PR #1726, opened 2022-11-21
Allow variable number of operands in work-group metadata

steffenlarsen:
The `reqd_work_group_size` and `work_group_size_hint` metadata nodes are currently expected to have exactly 3 operands which are in turn translated to the 3 operands of their corresponding SPIR-V execution modes. This commit relaxes this requirement by allowing the metadata to have 1-3 operands, where the missing dimensions are padded with 1's in the corresponding execution mode. Note that the information about the dimensionality of the original metadata is lost after translation, so reverse translation will always result in the metadata having all three operands.

Motivation comes from https://github.com/intel/llvm/pull/7450 where instead of doing the implicit padding during LLVM IR generation we want to simplify the metadata to only have the operands specified. Not only does it relax the requirements, it also lets us preserve information about dimensionality up until it is translated into SPIR-V.

### DIFF
--- a/lib/SPIRV/OCLUtil.cpp
+++ b/lib/SPIRV/OCLUtil.cpp
@@ -799,12 +799,15 @@ unsigned getOCLVersion(Module *M, bool AllowMulti) {
   return encodeOCLVer(Ver.first, Ver.second, 0);
 }
 
-void decodeMDNode(MDNode *N, unsigned &X, unsigned &Y, unsigned &Z) {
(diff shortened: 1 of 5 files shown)


MrSidims (approved):
LGTM
But please specify, what FE is capable to produce 1-dimentional WG metadata, if I understand correctly, it's not just llvm.org's clang compiling vanilla OpenCL code.

svenvh (approved):

PR #1727, opened 2022-11-21
Support FPGA function !spirv.Decoration metadata

mendell27:
Supporting the new FPGA kernel attribute pipelined needs the translator to handle !spirv.Decoration metadata on functions as well as on global variables.  If the MD is found on the function, call the existing translation of metadata.


lib/SPIRV/SPIRVWriter.cpp
```diff
@@ -974,6 +974,8 @@ void LLVMToSPIRVBase::transVectorComputeMetadata(Function *F) {
   }
 }
 
+static void transMetadataDecorations(Metadata *MD, SPIRVEntry *Target) {
```
mendell27, line 977: Is this the right way to handle this?  Or should I move LLVMToSPIRVBase::transFPGAFunctionMetadata down to after the declaration of transMetadataDecorations?

MrSidims (commented):
Hm, it does look like that I missed SPV_INTEL_fpga_invocation_pipelining_attributes during restructuring test folder. Anyway, could you please to either move the test to test/extensions/INTEL/SPV_INTEL_kernel_attributes/ or to test/extensions/INTEL/SPV_INTEL_fpga_invocation_pipelining_attributes/ (which should be created)?

mendell27:
> Hm, it does look like that I missed `SPV_INTEL_fpga_invocation_pipelining_attributes` during restructuring test folder. Anyway, could you please to move the test to either `test/extensions/INTEL/SPV_INTEL_kernel_attributes/` or `test/extensions/INTEL/SPV_INTEL_fpga_invocation_pipelining_attributes/` (which should be created)?
> 
> Other than a test location the patch is LGTM

Moved to test/extensions/INTEL/SPV_INTEL_kernel_attributes/

MrSidims (approved):

PR #1729, opened 2022-11-23
Translate LLVM intrinsics into native_* OpenCL instructions

vmustya:
When an intrinsic function is called with `afn` flag, it's allowed to
substitute an approximate calculations. So the translator can emit
native versions of OpenCL extended instructions.


MrSidims (approved):
LGTM
@svenvh could you please take a look as well?

svenvh (approved):

PR #1737, opened 2022-11-24
[WIP] Attempt to translate non-standart integers < 64 if they are not a part of a type of global symbol

MrSidims:
This is an incomplete patch, but I'd like to gather opinions whether we can legalize some non-standard integers in such 'cheap' manner.  

### DIFF
--- a/lib/SPIRV/SPIRVWriter.cpp
+++ b/lib/SPIRV/SPIRVWriter.cpp
@@ -297,7 +297,7 @@ static bool recursiveType(const StructType *ST, const Type *Ty) {
   return Run(Ty);
 }
 
-SPIRVType *LLVMToSPIRVBase::transType(Type *T) {
+SPIRVType *LLVMToSPIRVBase::transType(Type *T, bool IsGlobalSymbol) {
   LLVMToSPIRVTypeMap::iterator Loc = TypeMap.find(T);
   if (Loc != TypeMap.end())
     return Loc->second;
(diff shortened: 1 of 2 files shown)


MrSidims:
@asudarsa please take a look

lib/SPIRV/SPIRVWriter.cpp
```diff
@@ -316,11 +316,16 @@ SPIRVType *LLVMToSPIRVBase::transType(Type *T) {
...
             ExtensionID::SPV_INTEL_arbitrary_precision_integers) ||
-        BM->getErrorLog().checkError(
-            BitWidth == 8 || BitWidth == 16 || BitWidth == 32 || BitWidth == 64,
-            SPIRVEC_InvalidBitWidth, std::to_string(BitWidth))) {
-      return mapType(T, BM->addIntegerType(T->getIntegerBitWidth()));
-    }
+            BitWidth == 8 || BitWidth == 16 || BitWidth == 32 || BitWidth == 64)
+      return mapType(T, BM->addIntegerType(BitWidth));
+    // Map non-standart integers to either int32 or int64 in case if they are
+    // not a part of a function type.
```
MrSidims, line 322:
```suggestion
    // not a part of a function type or global variable
```

lib/SPIRV/SPIRVWriter.cpp
```diff
@@ -297,7 +297,7 @@ static bool recursiveType(const StructType *ST, const Type *Ty) {
   return Run(Ty);
 }
 
-SPIRVType *LLVMToSPIRVBase::transType(Type *T) {
+SPIRVType *LLVMToSPIRVBase::transType(Type *T, bool IsGlobalSymbol) {
```
bader, line 300: `IsGlobalSymbol` should be renamed to something type related. For instance, `IsTypeExtendAllowed`?

`transType` is supposed to translate types, so having `IsGlobalSymbol` parameter is confusing.

bader (commented):
I'm not sure this is safe transformation.
There are bit-manipulating instructions might be confused by such type changes. Right?

PR #1738, opened 2022-11-27
Fix builtin vars translation

MrSidims:
The translator was crashing in case if builin GV was accessed via GEP without AS cast due to incorrect assumption.


MrSidims:
@Fznamznon @asudarsa please take a look

svenvh (approved):

Issue #1752, opened 2022-12-04
relational operators mistranslated? (e.g. signbit)


airlied:
I'm tracking down a failure on signbit with llvmpipe backed rustictl. I think it's due to how llvmpipe encodes booleans as 32-bits internally.

Result is:
           ==> ERROR: signbit: 2 ulp error at -0x1p+113: *1 vs. -1

The problem is the signbit spec says for scalars it returns 1, and vectors it returns -1 now the translator converts signbit to SPIR-V SignBitSet which is a boolean operation, but if I have 32-bit booleans I end up with -1 for the scalar case.

I think the translator should be more explicit here, it should return integer values for these somehow instead of bools. I'm not 100% sure this is the translator's bug, but I figured here is a good place to start.

PR #1753, opened 2022-12-05
fix scalar relational operator translation.

airlied:
The CL spec for relational operators says that for scalar values the return value is 0/1 not 0/-1 which is for vectors.

Fix the translator to use 0/1 when it's a scalar operator


airlied:
Fixes #1752

MrSidims (approved):
Hi, thanks for the report and the PR!

Unlike OpenCL spec - SPIR-V says nothing about, what bits should be set, but I believe that the fix is correct. Not sure if it worth to add a test.
@svenvh  please take a look.

MrSidims:
Closing and reopening PR to restart tests

svenvh:
> @svenvh any pointers on what sort of test could be used here? I'm not really sure where the current code is tested if at all, so how would I test the return value here?

The current pattern doesn't seem to be tested indeed (otherwise some tests would have failed with your change).  I think `test/transcoding/relationals_{half,float,double}.ll` exercises this code path, so I would try adding CHECK lines for the new IR after the corresponding `call`s and `Is*/SignBit` ops.

MrSidims (commented):
LGTM, thanks for adding the test!

PR #1757, opened 2022-12-06
Translate atomicrmw fadd into AtomicFAddEXT

sarnex:
Some frontends can generate atomicrmw fadd instead of __spirv_AtomicFAddEXT, so we should support this too

### DIFF
--- a/test/atomicrmw_fadd.ll
+++ b/test/atomicrmw_fadd.ll
@@ -1,7 +1,7 @@
 ; RUN: llvm-as < %s -o %t.bc
 ; RUN: llvm-spirv --spirv-ext=+SPV_EXT_shader_atomic_float_add %t.bc -o %t.spv
 ; RUN: spirv-val %t.spv
-; RUN: llvm-spirv --spirv-ext=+SPV_EXT_shader_atomic_float_add -to-text %t.spv -o - | FileCheck %s
+; RUN: llvm-spirv -to-text %t.spv -o - | FileCheck %s
 
 ; CHECK: TypeInt [[Int:[0-9]+]] 32 0
 ; CHECK-DAG: Constant [[Int]] [[Scope_Device:[0-9]+]] 1 {{$}}
--- a/test/atomicrmw_fadd.ll
+++ b/test/atomicrmw_fadd.ll
@@ -1,0 +1,31 @@
+; RUN: llvm-as < %s -o %t.bc
+; RUN: llvm-spirv --spirv-ext=+SPV_EXT_shader_atomic_float_add %t.bc -o %t.spv
+; RUN: spirv-val %t.spv
+; RUN: llvm-spirv --spirv-ext=+SPV_EXT_shader_atomic_float_add -to-text %t.spv -o - | FileCheck %s
+
+; CHECK: TypeInt [[Int:[0-9]+]] 32 0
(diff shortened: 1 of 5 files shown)


test/atomicrmw_fadd.ll
```diff
@@ -0,0 +1,31 @@
+; RUN: llvm-as < %s -o %t.bc
+; RUN: llvm-spirv --spirv-ext=+SPV_EXT_shader_atomic_float_add %t.bc -o %t.spv
+; RUN: spirv-val %t.spv
+; RUN: llvm-spirv --spirv-ext=+SPV_EXT_shader_atomic_float_add -to-text %t.spv -o - | FileCheck %s
```
AlexeySachkov, line 4:
```suggestion
; RUN: llvm-spirv -to-text %t.spv -o - | FileCheck %s
```

No need to specify extensions when doing text-binary SPIR-V conversion.

MrSidims (approved):
LGTM with Alexey's comment applied

BTW there are also fmin/fmax atomic instructions that have a representation in SPIR-V in a different extension: SPV_EXT_shader_atomic_float_min_max . Do you know if we have a request to support them as well?

test/atomicrmw_fadd.ll
sarnex: applied, thanks

sarnex:
> BTW there are also fmin/fmax atomic instructions that have a representation in SPIR-V in a different extension: SPV_EXT_shader_atomic_float_min_max . Do you know if we have a request to support them as well?

The request I have is specifically for fadd. I wasn't able to modify the frontend repro case for fadd to emit fmin/fmax, it looks like we never use atomicrmw for that right now.

svenvh (approved):

sarnex:
> > The request I have is specifically for fadd. I wasn't able to modify the frontend repro case for fadd to emit fmin/fmax, it looks like we never use atomicrmw for that right now, however I'm happy to implement it here if you think it's valuable
> 
> It's good to add, but probably not in this PR, since it's a different SPIR-V extension. I'll merge once testing is complete.

Thanks, I will work on a separate PR for fmin/fmax

test/atomicrmw_fadd.ll
```diff
@@ -0,0 +1,31 @@
+; RUN: llvm-as < %s -o %t.bc
+; RUN: llvm-spirv --spirv-ext=+SPV_EXT_shader_atomic_float_add %t.bc -o %t.spv
```
MrSidims, line 2: @sarnex 
Sorry, I have to withdraw my approve. Please add checks in the test like:
CHECK-DAG: Extension "SPV_EXT_shader_atomic_float_add"
CHECK-DAG: Capability AtomicFloat32AddEXT

I believe, that this checks will fail, since we probably need to add them explicitly. Also there should be a separate test for double to emit AtomicFloat64AddEXT capability.

Please also move this (or these two tests) into test/extensions/EXT/SPV_EXT_shader_atomic_float_/ directory

MrSidims (changes requested):

test/atomicrmw_fadd.ll
sarnex: @MrSidims Capability stuff seems to work. I believe I addressed all of this in my latest commit

MrSidims (approved):

PR #1758, opened 2022-12-07
Report spec constant types when they are queried with --spec-const-info

aarongreig:
Also allow binary float spec constant input by passing a hex value with 0x prefix.

This contribution is being made by Codeplay on behalf of Samsung

### DIFF
--- a/include/LLVMSPIRVLib.h
+++ b/include/LLVMSPIRVLib.h
@@ -132,7 +132,11 @@ bool readSpirv(LLVMContext &C, const SPIRV::TranslatorOpts &Opts,
 /// \brief Partially load SPIR-V from the stream and decode only instructions
 /// needed to get information about specialization constants.
 /// \returns true if succeeds.
-using SpecConstInfoTy = std::pair<uint32_t, uint32_t>;
+struct SpecConstInfoTy {
+  uint32_t ID;
+  uint32_t Size;
+  std::string Type;
+};
 bool getSpecConstInfo(std::istream &IS,
                       std::vector<SpecConstInfoTy> &SpecConstInfo);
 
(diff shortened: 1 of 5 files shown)


MrSidims:
@aarongreig thanks for your contribution!
It's mandatory for this project to sign CLA to have the patches merged. Please also consider adding a test for the patch.

aarongreig:
Thanks for taking a look at it, I'm waiting for internal confirmation before I sign the CLA but I don't anticipate any problems with it. I've updated the spec constant test to check the new functionality works as expected. I also made a minor reivision to the lit config file as I was having an issue where it used a stale llvm-spirv binary I had lying around in my llvm build directory rather than the one built from this repo, I think this change should be otherwise harmless.

aarongreig:
I don't think it's my changes tripping the CI

MrSidims:
>> I don't think it's my changes tripping the CI

Indeed, I'll close and re-open PR to restart all of the tests, they should be fixed ATM.

MrSidims:
>> I'm waiting for internal confirmation before I sign the CLA

Sure! AFAIK it's not the first Codeplay's contribution to this project, you may ask @Naghasan for examples.

MrSidims (approved):
The patch looks good, thanks!
I just have may be a silly question, more for my own curiosity, since I'm no so strong with spec constants. Why we use "f16, f32, f64" instead of "half, float, double" (just like in LLVM IR?)

svenvh (approved):
LGTM.

> Why we use "f16, f32, f64" instead of "half, float, double" (just like in LLVM IR?)

I don't think it's mandated anywhere, so we could pick either. But since we've already used "f<N>" for a while I'd be reluctant to change it.

aarongreig:
I went with the llvm convention kind of out of habit so I'm not particularly attached to it, although it is easier to parse on the other side (e.g. `if(type[0] == 'f') {parseFloatValue()} else if (type[0] == 'i') {parseIntValue()}`)

PR #1765, opened 2022-12-08
Catch up with llvm/llvm-project@9114ac67a986400155b8b82013d09a9e4f48e534

arichardson:
The ptr_annotation/var_annotation/annotation intrinsics gained additional overloads which means we have to update calls to Intrinsic::getDeclaration()/Builder.CreateIntrinsic().


arichardson:
I believe a follow-up change would be to use the Constant address space for the global strings, but since I don't understand the required semantics here I did not include that in this PR.

MrSidims (approved):
LGTM
@arichardson please apply clang tidy suggestion

arichardson:
> LGTM @arichardson please apply clang tidy suggestion

Done. I guess the CI failures are caused by the builders using an older LLVM snapshot?

PR #1775, opened 2022-12-12
[OpaquePointers] Improvements to the type scavenger.

jcranmer-intel:
The main improvement is to enable intrinsics with more complex typing rules to be described. For example, memcpy's requirement that its two pointer arguments have the same type.

There is still a larger rewrite of the type scavenger to use TypedPointerType coming down the line. However, this rewrite also needs opaque types to represent the deferred type concept, which is why it hasn't been done yet. This is an intermediate step in the rewrite which better supports some of the other intrinsics with current issues.


MrSidims:

MrSidims:
@jcranmer-intel it seems like CI failures are related to the patch, could you please take a look?

svenvh (approved):

MrSidims (approved):

PR #1776, opened 2022-12-12
Fix build and tests for LLVM changes

svenvh:
Update test after llvm-project commit 6b2829dd8772 ("Allow epilogue_begin to be emitted when generating DWARF", 2022-12-09).

Update after llvm-project commit f7dffc28b3f8 ("Don't include None.h (NFC)", 2022-12-10).


AlexeySachkov (approved):
Thanks!

PR #1777, opened 2022-12-13
Use std::optional instead of llvm::Optional

svenvh:
The LLVM project is migrating away from `llvm::Optional` to `std::optional`.  Update the translator to use `std::optional`.


AlexeySachkov (approved):
Thanks!

@Fznamznon: FYI

PR #1778, opened 2022-12-13
Fix issue due to ignoring return value

asudarsa:
This is a minor fix to resolve some issues reported by static analysis:
1. Use 'assert' to NOT ignore return values from getAsInteger and consumeUnsignedInteger.
2. Use 'assert' and other means to NOT ignore return value from getConstantStringInfo.

Thanks

### DIFF
--- a/lib/SPIRV/SPIRVUtil.cpp
+++ b/lib/SPIRV/SPIRVUtil.cpp
@@ -471,7 +471,7 @@ bool oclIsBuiltin(StringRef Name, StringRef &DemangledName, bool IsCpp) {
     size_t Start = Name.find_first_not_of("0123456789", DemangledNameLenStart);
     size_t Len = 0;
     if (Name.substr(DemangledNameLenStart, Start - DemangledNameLenStart)
-        .getAsInteger(10, Len))
+            .getAsInteger(10, Len))
       assert(0 && "Error in extracting integer value");
     DemangledName = Name.substr(Start, Len);
   } else {
--- a/lib/SPIRV/SPIRVUtil.cpp
+++ b/lib/SPIRV/SPIRVUtil.cpp
@@ -471,8 +471,10 @@ bool oclIsBuiltin(StringRef Name, StringRef &DemangledName, bool IsCpp) {
     size_t Start = Name.find_first_not_of("0123456789", DemangledNameLenStart);
     size_t Len = 0;
     if (Name.substr(DemangledNameLenStart, Start - DemangledNameLenStart)
-            .getAsInteger(10, Len))
-      assert(0 && "Error in extracting integer value");
+            .getAsInteger(10, Len)) {
+      SPIRVDBG(errs() << "Error in extracting integer value");
+      return false;
+    }
     DemangledName = Name.substr(Start, Len);
   } else {
     size_t Start = Name.find_first_not_of("0123456789", 2);
--- a/lib/SPIRV/SPIRVUtil.cpp
+++ b/lib/SPIRV/SPIRVUtil.cpp
@@ -470,8 +470,9 @@ bool oclIsBuiltin(StringRef Name, StringRef &DemangledName, bool IsCpp) {
     size_t DemangledNameLenStart = NameSpaceStart + 11;
     size_t Start = Name.find_first_not_of("0123456789", DemangledNameLenStart);
     size_t Len = 0;
-    Name.substr(DemangledNameLenStart, Start - DemangledNameLenStart)
-        .getAsInteger(10, Len);
+    if (Name.substr(DemangledNameLenStart, Start - DemangledNameLenStart)
+        .getAsInteger(10, Len))
+      assert(0 && "Error in extracting integer value");
     DemangledName = Name.substr(Start, Len);
   } else {
     size_t Start = Name.find_first_not_of("0123456789", 2);
--- a/lib/SPIRV/SPIRVWriter.cpp
+++ b/lib/SPIRV/SPIRVWriter.cpp
@@ -650,7 +650,7 @@ SPIRVType *LLVMToSPIRVBase::transSPIRVJointMatrixINTELType(
   auto ParseInteger = [this](StringRef Postfix) -> ConstantInt * {
     unsigned long long N = 0;
     if (consumeUnsignedInteger(Postfix, 10, N)) {
-      SPIRVDBG(errs() << "Error in extracting integer value");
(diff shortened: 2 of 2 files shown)


lib/SPIRV/SPIRVWriter.cpp
```diff
@@ -4284,7 +4285,8 @@ void LLVMToSPIRVBase::transGlobalAnnotation(GlobalVariable *V) {
         cast<GlobalVariable>(CS->getOperand(1)->stripPointerCasts());
 
     StringRef AnnotationString;
-    getConstantStringInfo(GV, AnnotationString);
+    bool HasAnno = getConstantStringInfo(GV, AnnotationString);
+    assert(HasAnno && "Annotation string missing");
```
AlexeySachkov, line 4289:
```suggestion
    assert(HasAnno && "Annotation string missing");
    (void)HasAnno;
```

Or non-assert `-Werror` build will fail. Alternatively, `std::ignore = HasAnno;` can be used

asudarsa: Fixed it in a different way. Thanks

lib/SPIRV/SPIRVWriter.cpp
```diff
@@ -649,7 +649,8 @@ SPIRVType *LLVMToSPIRVBase::transSPIRVJointMatrixINTELType(
 
   auto ParseInteger = [this](StringRef Postfix) -> ConstantInt * {
     unsigned long long N = 0;
-    consumeUnsignedInteger(Postfix, 10, N);
+    if (consumeUnsignedInteger(Postfix, 10, N))
+      assert(0 && "Error in extracting integer value");
```
MrSidims, line 653: Should probably be a SPIR-V module error in this case to see such scenario in release build as well (and I believe we can't put a contract here that it would never happen).

lib/SPIRV/SPIRVUtil.cpp
```diff
@@ -470,8 +470,9 @@ bool oclIsBuiltin(StringRef Name, StringRef &DemangledName, bool IsCpp) {
     size_t DemangledNameLenStart = NameSpaceStart + 11;
     size_t Start = Name.find_first_not_of("0123456789", DemangledNameLenStart);
     size_t Len = 0;
-    Name.substr(DemangledNameLenStart, Start - DemangledNameLenStart)
-        .getAsInteger(10, Len);
+    if (Name.substr(DemangledNameLenStart, Start - DemangledNameLenStart)
+            .getAsInteger(10, Len))
+      assert(0 && "Error in extracting integer value");
```
MrSidims, line 475: Should probably be an error as well.

lib/SPIRV/SPIRVWriter.cpp
```diff
@@ -4284,7 +4286,8 @@ void LLVMToSPIRVBase::transGlobalAnnotation(GlobalVariable *V) {
         cast<GlobalVariable>(CS->getOperand(1)->stripPointerCasts());
 
     StringRef AnnotationString;
-    getConstantStringInfo(GV, AnnotationString);
+    if (!getConstantStringInfo(GV, AnnotationString))
+      assert(0 && "Annotation string missing");
```
MrSidims, line 4290:
```suggestion
    if (!getConstantStringInfo(GV, AnnotationString)) {
      assert(!"Annotation string missing");
      return;
    }
```
Or an error. Though I believe it's no fatal enough to abort translation, at some point we should have warnings added.

MrSidims:
I agree, that we should reuse diagnostics return value of StringRef functions, thanks for fixing this!

asudarsa:
Hi @MrSidims 

I have addressed your comment here. Can you please take a look when convenient?

Thanks

lib/SPIRV/SPIRVWriter.cpp
```diff
@@ -649,7 +649,10 @@ SPIRVType *LLVMToSPIRVBase::transSPIRVJointMatrixINTELType(
 
   auto ParseInteger = [this](StringRef Postfix) -> ConstantInt * {
     unsigned long long N = 0;
-    consumeUnsignedInteger(Postfix, 10, N);
+    if (consumeUnsignedInteger(Postfix, 10, N)) {
+      SPIRVDBG(errs() << "Error in extracting integer value");
```
MrSidims, line 653:
```suggestion
      BM->getErrorLog().checkError(false, InvalidLlvmModule, "TypeJointMatrixINTEL expects integer parameters");
```

lib/SPIRV/SPIRVUtil.cpp
```diff
@@ -470,8 +470,11 @@ bool oclIsBuiltin(StringRef Name, StringRef &DemangledName, bool IsCpp) {
     size_t DemangledNameLenStart = NameSpaceStart + 11;
     size_t Start = Name.find_first_not_of("0123456789", DemangledNameLenStart);
     size_t Len = 0;
-    Name.substr(DemangledNameLenStart, Start - DemangledNameLenStart)
-        .getAsInteger(10, Len);
+    if (Name.substr(DemangledNameLenStart, Start - DemangledNameLenStart)
+            .getAsInteger(10, Len)) {
+      SPIRVDBG(errs() << "Error in extracting integer value");
```
MrSidims, line 475: I meant the same error as the suggested in the Writer, but from what I see in this scope we can not access SPIRVModule object, hence SPIRVDBG or assert are both LGTM.

lib/SPIRV/SPIRVWriter.cpp
```diff
@@ -649,7 +649,10 @@ SPIRVType *LLVMToSPIRVBase::transSPIRVJointMatrixINTELType(
 
   auto ParseInteger = [this](StringRef Postfix) -> ConstantInt * {
     unsigned long long N = 0;
-    consumeUnsignedInteger(Postfix, 10, N);
+    if (consumeUnsignedInteger(Postfix, 10, N)) {
+      BM->getErrorLog().checkError(false, InvalidLlvmModule, "TypeJointMatrixINTEL expects integer parameters");
+      return 0;
+    }
```
MrSidims, line 655:
```suggestion
    if (consumeUnsignedInteger(Postfix, 10, N)) {
      BM->getErrorLog().checkError(
          false, SPIRVEC_InvalidLlvmModule,
          "TypeJointMatrixINTEL expects integer parameters");
      return 0;
    }
```

MrSidims (approved):

PR #1780, opened 2022-12-15
Translate atomicrmw fsub into FNegate and AtomicFAddEXT

sarnex:
We can support atomic fsub by just negating the operand with atomic fadd.

### DIFF
--- a/lib/SPIRV/OCLUtil.cpp
+++ b/lib/SPIRV/OCLUtil.cpp
@@ -647,6 +647,7 @@ template <> void LLVMSPIRVAtomicRmwOpCodeMap::init() {
   add(llvm::AtomicRMWInst::UMax, OpAtomicUMax);
   add(llvm::AtomicRMWInst::UMin, OpAtomicUMin);
   add(llvm::AtomicRMWInst::FAdd, OpAtomicFAddEXT);
+  add(llvm::AtomicRMWInst::FSub, OpAtomicFAddEXT);
   add(llvm::AtomicRMWInst::FMin, OpAtomicFMinEXT);
   add(llvm::AtomicRMWInst::FMax, OpAtomicFMaxEXT);
 }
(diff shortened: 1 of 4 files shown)


sarnex:
Failures seem unrelated

lib/SPIRV/OCLUtil.cpp
```diff
@@ -647,6 +647,7 @@ template <> void LLVMSPIRVAtomicRmwOpCodeMap::init() {
   add(llvm::AtomicRMWInst::UMax, OpAtomicUMax);
   add(llvm::AtomicRMWInst::UMin, OpAtomicUMin);
   add(llvm::AtomicRMWInst::FAdd, OpAtomicFAddEXT);
+  add(llvm::AtomicRMWInst::FSub, OpAtomicFAddEXT);
```
MrSidims, line 650: I beleive we shouln't add such entry to the map making fsub being implicitly replaced by fadd (it can probably lead to bugs in the future) - instead FNegative and AtomicFAddEXT should be both generated explicitly.

test/extensions/EXT/SPV_EXT_shader_atomic_float_/atomicrmw_fsub_double.ll
```diff
@@ -0,0 +1,34 @@
+; RUN: llvm-as < %s -o %t.bc
+; RUN: llvm-spirv --spirv-ext=+SPV_EXT_shader_atomic_float_add %t.bc -o %t.spv
```
MrSidims, line 2: Please also test -r here - what LLVM IR would be generated after the translation?

lib/SPIRV/OCLUtil.cpp
sarnex: Thanks for the review, good point, I'll special case fsub instead of relying on the existing path.

test/extensions/EXT/SPV_EXT_shader_atomic_float_/atomicrmw_fsub_double.ll
sarnex: Will add, thanks

lib/SPIRV/OCLUtil.cpp
sarnex: Please check the latest commit when you get a chance, I think I addressed this feedback while reusing existing code but maybe I misunderstood

MrSidims:

MrSidims (approved):
LGTM, thanks!
@svenvh please also take a look if you are OK with such solution

svenvh (approved):

PR #1781, opened 2022-12-16
[OpaquePointers] Avoid infinite cycles in scavenging the types of PHIs.

jcranmer-intel:
PHI instructions have their type scavenged from one of their incoming values. Originally, this used solely the first one, but it is possible for this to generate an infinite cycle (consider a pointer-swapping loop). This fixes the logic to detect such infinite cycles and to consider other incoming values instead.


MrSidims:

MrSidims (approved):
LGTM, thanks!

svenvh (approved):

PR #1782, opened 2022-12-19
Change OpFMod and OpFRem implementations to be in line with CTS.

aarongreig:
The CTS expects these instructions to be implemented in terms of the fmod builtin.

See the CTS implementations [here](https://github.com/KhronosGroup/OpenCL-CTS/blob/main/test_conformance/spirv_new/test_op_fmath.cpp#L69).

This contribution is being made by Codeplay on behalf of Samsung

### DIFF
--- a/lib/SPIRV/SPIRVReader.cpp
+++ b/lib/SPIRV/SPIRVReader.cpp
@@ -2300,21 +2300,36 @@ Value *SPIRVToLLVM::transValueWithoutDecoration(SPIRVValue *BV, Function *F,
 
   case OpFMod: {
     // translate OpFMod(a, b) to:
-    //   r = frem(a, b)
-    //   c = copysign(r, b)
-    //   needs_fixing = islessgreater(r, c)
-    //   result = needs_fixing ? r + b : c
+    //   r = fmod(a, b)
+    //   result = copysign(r, b)
+    // this is how the CTS expects it to be implemented
     IRBuilder<> Builder(BB);
     SPIRVFMod *FMod = static_cast<SPIRVFMod *>(BV);
-    auto Dividend = transValue(FMod->getOperand(0), F, BB);
     auto Divisor = transValue(FMod->getOperand(1), F, BB);
-    auto FRem = Builder.CreateFRem(Dividend, Divisor, "frem.res");
-    auto CopySign = Builder.CreateBinaryIntrinsic(
-        llvm::Intrinsic::copysign, FRem, Divisor, nullptr, "copysign.res");
-    auto FAdd = Builder.CreateFAdd(FRem, Divisor, "fadd.res");
-    auto Cmp = Builder.CreateFCmpONE(FRem, CopySign, "cmp.res");
-    auto Select = Builder.CreateSelect(Cmp, FAdd, CopySign);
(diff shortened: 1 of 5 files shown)


MrSidims:

MrSidims (approved):
LGTM, thanks!

svenvh:
Please allow @StuartDBrady to have a look at this before merging.

lib/SPIRV/SPIRVReader.cpp
```diff
@@ -2300,21 +2300,36 @@ Value *SPIRVToLLVM::transValueWithoutDecoration(SPIRVValue *BV, Function *F,
...
+    auto *FModBuiltinCall = transBuiltinFromInst("fmod", FMod, BB);
+    auto *CopySign = Builder.CreateBinaryIntrinsic(llvm::Intrinsic::copysign,
+                                                   FModBuiltinCall, Divisor,
+                                                   nullptr, "copysign.res");
+    return mapValue(BV, CopySign);
+  }
+
+  case OpFRem: {
+    // translate OpFRem(a, b) to:
+    //   result = fmod(a, b)
```
AlexeySachkov, line 2318: This doesn't seem right, because result of [`OpFRem`](https://registry.khronos.org/SPIR-V/specs/unified1/SPIRV.html#OpFRem) should not be always the same as [`OpFMod`](https://registry.khronos.org/SPIR-V/specs/unified1/SPIRV.html#OpFMod): they both compute [Reminder](https://registry.khronos.org/SPIR-V/specs/unified1/SPIRV.html#Remainder), but `OpFRem` uses the sign of `Operand 1` and `OpFMod` uses the sign of `Operand 2`.

The comment says that results of those two are equal, which is clearly may not be the case per SPIR-V spec.

StuartDBrady (changes requested):
The CTS implementation of **OpFMod** in this review is completely incorrect, in that the result returned _is not a remainder_.  Please change the CTS implementation to match the implementation here, instead.  (Note that the CTS implementation was not exercised with any negative operands, which is why it would fail to catch incorrect implementations of **OpFMod** such as this one but still passes with a correct implementation.)

StuartDBrady:
To clarify, these are examples of calculations using remainder operations:

* 9 REM 10 = 9
* 27 REM 10 = 7
* -4 REM 10 = -4
* -33 REM 10 = -3
* -4 MOD 10 = 6 (**NOT 4**)
* -33 REM 10 = 7 (**NOT 3**)

StuartDBrady:
See KhronosGroup/OpenCL-CTS/issues/1548 and PR #296 for more explanation regarding **OpFMod**.

StuartDBrady:
Note that the SPIR-V specification is unambiguous about this — in [§ 2.2.3. _Computation_](https://registry.khronos.org/SPIR-V/specs/unified1/SPIRV.html#_computation) (in § 2.2. _Terms_):

> Remainder: When dividing a by b, a remainder r is defined to be a value that satisfies r + q × b = a where q is a whole number and |r| < |b|.

aarongreig:
Ok my bad, thought I had an easy fix to a surprise CTS fail and failed to check if there were any issues on the cts repo about it.

PR #1783, opened 2022-12-19
Replace internal StreamingInterfaceINTEL uses with official ones

mendell27:
Now that StreamingInterfaceINTEL  is in the official headers, replace all the internal uses with the official ones


mendell27:
Trying with a fresh  branch

PR #1784, opened 2022-12-20
[NFC] Use the official StreamingInterfaceINTEL enums

mendell27:
Now that the official enums have been added, replace the internal uses with the official ones.
Update the commit for the Headers to get the enums defined


mendell27:
check-llvm-spirv passes if I revert the change in lib/Bitcode/Reader/BitcodeReader.cpp that went in to LLVM last week

MrSidims (approved):
LGTM
Please let us fix CI issues first, I'll be back from sick leave soon to help with it.

PR #1785, opened 2022-12-20
Run llvm-as and llvm-dis with -opaque-pointers=0

svenvh:
Update some of the tests that started failing after llvm-project commit e45cf479231f ("[Bitcode] Remove auto-detection for typed pointers", 2022-12-14).

This commit only updates tests that pass again after mechanically adding `-opaque-pointers=0` to their `llvm-as` and `llvm-dis` commands.  Many more tests are still failing due to the opaque-pointer auto-upgrade happening when llvm-spirv reads the LLVM IR which won't be fixed by this commit.


svenvh:
Expecting all CI jobs to still fail, however within each job there should be fewer tests failing now.

@jcranmer-intel please let me know if this collides with any unpushed opaque-ptr changes you may have.

MrSidims (approved):

PR #1789, opened 2022-12-22
[NFC] Fix some static analyzer issues

asudarsa:
In this change set, we avoid local copy of strings.


MrSidims (approved):
LGTM

PR #1791, opened 2022-12-22
Add lowering for llvm.uadd.sat.* intrinsics

maksimsab:


maksimsab:
@jcranmer-intel Hi! Could you please take a look on it?

lib/SPIRV/SPIRVLowerSaddIntrinsics.cpp
```diff
@@ -147,17 +147,94 @@ void SPIRVLowerSaddIntrinsicsBase::replaceSaddSat(Function &F) {
   replaceSaddOverflow(*SaddO);
 }
 
+namespace {
+
+// This function for a given @IntegerBitWidth generates the following IR:
```
AlexeySachkov, line 152: We have [u_add_sat](https://registry.khronos.org/SPIR-V/specs/1.0/OpenCL.ExtendedInstructionSet.100.html#u_add_sat) OpenCL extended instruction, why are we not using it and instead emitting its implementation?

lib/SPIRV/SPIRVLowerSaddIntrinsics.cpp
```diff
@@ -147,17 +147,94 @@ void SPIRVLowerSaddIntrinsicsBase::replaceSaddSat(Function &F) {
...
+//     return static_cast<unsigned T>(-1);
+//   return a + b;
+// }
+Function *generateUaddSatReplacement(Module *M, const std::string &FuncName,
+                                     int IntegerBitWidth) {
+  Function *F;
+  if ((F = M->getFunction(FuncName)))
+    return F;
+
+  Type *ReturnType = IntegerType::get(M->getContext(), IntegerBitWidth);
```
AlexeySachkov, line 175: The intrinsic can accept vectors of integers as arguments.

AlexeySachkov (changes requested):

lib/SPIRV/SPIRVLowerSaddIntrinsics.cpp
maksimsab: There is a unclear question whether we can rely on having this extended instruction set and what should we do if it isn't available.

lib/SPIRV/SPIRVLowerSaddIntrinsics.cpp
```diff
@@ -37,21 +37,21 @@
 // function calls
 //
 //===----------------------------------------------------------------------===//
-#define DEBUG_TYPE "spv-lower-llvm_sadd_intrinsics"
```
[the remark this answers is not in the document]
maksimsab: This is not used and brings warnings which are treated as errors
https://github.com/KhronosGroup/SPIRV-LLVM-Translator/pull/1791/checks?check_run_id=10256417841

lib/SPIRV/SPIRVLowerSaddIntrinsics.cpp
AlexeySachkov: The translator project is 99% focused on `Compute` capabilities of SPIR-V, which are used by OpenCL. Any conformant OpenCL implementation must support `OpenCL.std` extended instruction set, see [The OpenCL™ SPIR-V Environment Specification](https://registry.khronos.org/OpenCL/specs/2.2/html/OpenCL_Env.html#_extended_instruction_sets)

Moreover, that extended instruction set is already extensively used by the translator and I doubt that you are targeting SPIR-V consumer which doesn't support it, because you would have already faced some issues.

Therefore, I would like us to continue to use it whenever possible and not to manually implement all LLVM intrinsics instead.

lib/SPIRV/SPIRVLowerSaddIntrinsics.cpp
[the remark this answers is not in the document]
AlexeySachkov: Fair enough

lib/SPIRV/SPIRVLowerSaddIntrinsics.cpp
maksimsab: Here is the new PR: https://github.com/KhronosGroup/SPIRV-LLVM-Translator/pull/1797

MrSidims:
Should be closed in favor of https://github.com/KhronosGroup/SPIRV-LLVM-Translator/pull/1797 ?

maksimsab:
Abandoned in favor of https://github.com/KhronosGroup/SPIRV-LLVM-Translator/pull/1797

PR #1792, opened 2022-12-23
Link to SPIRV-Tools via CMake package when pkgconfig is not available

dimula73:
pkgconfig is usually not available on Windows, so try to use normal CMake modules to locate the binaries for SPIRV-Tools.

The patch doesn't change the existing behavior when the pkgconfig is available, only when it is missing.

The CMake path is available only with this patch that has already been accepted upstream:
https://github.com/KhronosGroup/SPIRV-Tools/pull/5034

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -112,11 +112,24 @@ endif()
 
 set(LLVM_SPIRV_INCLUDE_DIRS ${CMAKE_CURRENT_SOURCE_DIR}/include)
 
+# first try locating SPIRV-Tools via pkgconfig (the old way)
 pkg_search_module(SPIRV_TOOLS SPIRV-Tools)
-if(NOT SPIRV_TOOLS_FOUND)
+if (NOT SPIRV_TOOLS_FOUND)
+  # then try locating SPIRV-Tools via cmake (the new way)
+  find_package(SPIRV-Tools)
+  find_package(SPIRV-Tools-tools)
+  if (SPIRV-Tools_FOUND AND SPIRV-Tools-tools_FOUND)
+    set(SPIRV_TOOLS_FOUND TRUE)
+    set(SPIRV_TOOLS_LDFLAGS SPIRV-Tools-shared)
+    get_target_property(SPIRV_TOOLS_INCLUDE_DIRS SPIRV-Tools-shared INTERFACE_INCLUDE_DIRECTORIES)
+  endif()
+endif()
+
+if (NOT SPIRV_TOOLS_FOUND)
   message(STATUS "SPIRV-Tools not found; project will be built without "
           "--spirv-tools-dis support.")
-endif(NOT SPIRV_TOOLS_FOUND)
+endif()
+
 
 add_subdirectory(lib/SPIRV)
 add_subdirectory(tools/llvm-spirv)
--- a/test/CMakeLists.txt
+++ b/test/CMakeLists.txt
@@ -5,7 +5,8 @@ llvm_canonicalize_cmake_booleans(SPIRV_SKIP_DEBUG_INFO_TESTS)
 get_target_property(LLVM_SPIRV_DIR llvm-spirv BINARY_DIR)
 set(LLVM_SPIRV_TEST_SOURCE_DIR ${CMAKE_CURRENT_SOURCE_DIR})
 
-if(SPIRV_TOOLS_FOUND)
+if(SPIRV_TOOLS_FOUND AND NOT SPIRV-Tools-tools_FOUND)
+  # we found SPIRV-Tools via pkgconfig
   find_program(SPIRV_TOOLS_SPIRV_AS NAMES spirv-as PATHS ${SPIRV_TOOLS_PREFIX}/bin)
   if(SPIRV_TOOLS_SPIRV_AS)
     set(SPIRV_TOOLS_SPIRV_AS_FOUND True)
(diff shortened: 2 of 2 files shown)


CMakeLists.txt
```diff
@@ -112,11 +112,24 @@ endif()
...
 
+# first try locating SPIRV-Tools via pkgconfig (the old way)
 pkg_search_module(SPIRV_TOOLS SPIRV-Tools)
-if(NOT SPIRV_TOOLS_FOUND)
-  message(STATUS "SPIRV-Tools not found; project will be built without "
+if (NOT SPIRV_TOOLS_FOUND)
+  # then try locating SPIRV-Tools via cmake (the new way)
+  find_package(SPIRV-Tools)
+  find_package(SPIRV-Tools-tools)
+  if (SPIRV-Tools_FOUND AND SPIRV-Tools-tools_FOUND)
```
MrSidims, line 121: It's a first time I see spelling like SPIRV-Tools-tools and googling doesn't help. When I can learn more about this package? If it's the same as SPIRV-Tools, shouldn't we have `OR` instead of `AND` in `if` statement?

CMakeLists.txt
```diff
@@ -112,11 +112,24 @@ endif()
...
+    set(SPIRV_TOOLS_FOUND TRUE)
+    set(SPIRV_TOOLS_LDFLAGS SPIRV-Tools-shared)
+    get_target_property(SPIRV_TOOLS_INCLUDE_DIRS SPIRV-Tools-shared INTERFACE_INCLUDE_DIRECTORIES)
+  endif()
+endif()
+
+if (NOT SPIRV_TOOLS_FOUND)
+  message(STATUS "SPIRV-Tools-tools not found; project will be built without "
```
MrSidims, line 129: Original message was just about `SPIRV-Tools` and not about `SPIRV-Tools-tools`

test/CMakeLists.txt
```diff
@@ -5,7 +5,8 @@ llvm_canonicalize_cmake_booleans(SPIRV_SKIP_DEBUG_INFO_TESTS)
 get_target_property(LLVM_SPIRV_DIR llvm-spirv BINARY_DIR)
 set(LLVM_SPIRV_TEST_SOURCE_DIR ${CMAKE_CURRENT_SOURCE_DIR})
 
-if(SPIRV_TOOLS_FOUND)
+if (SPIRV_TOOLS_FOUND AND NOT SPIRV-Tools-tools_FOUND)
```
MrSidims, line 8: Okay, I probably have figured that out - they are the same, but SPIRV-Tools-tools is available only for `find_package` and you use it to distinguish how tools package was found, correct? If yes, may be we shouldn't call `find_package(SPIRV-Tools)` leaving only `find_package(SPIRV-Tools-tools)` ?

test/CMakeLists.txt
```diff
@@ -22,19 +23,44 @@ if(SPIRV_TOOLS_FOUND)
   endif()
 
   set(SPIRV_TOOLS_BINDIR "${SPIRV_TOOLS_PREFIX}/bin")
+elseif(SPIRV-Tools-tools_FOUND)
```
MrSidims, line 26: What about `SPIRV-Tools_FOUND` ?

CMakeLists.txt
dimula73: Hi, @MrSidims!

This is a separate package for the binary executables of the standard SPIRV-Tools. I have added the patch upstream for that just yesterday, so there is no google results for that: https://github.com/KhronosGroup/SPIRV-Tools/pull/5034

Basically, SPIRV-Tools define a separate package for each ot their subfolders. This one is for the subfolder `./tools/`

MrSidims: Basically all my comments can be replaced with: do we need both `find_package(SPIRV-Tools)` and `find_package(SPIRV-Tools-tools)` and if yes, how `..._FOUND` should be used in the checks?

MrSidims: @dimula73 got it, thanks! Feel free to ignore the comments below

test/CMakeLists.txt
MrSidims: Please ignore the comment above.

@dimula73 Yet what about:
```suggestion
if (SPIRV_TOOLS_FOUND)
```
?
Because if SPIRV_TOOLS_FOUND = TRUE, then we never try to execute `find_package(SPIRV-Tools-tools)`

CMakeLists.txt
```diff
@@ -112,11 +112,24 @@ endif()
...
+  find_package(SPIRV-Tools)
+  find_package(SPIRV-Tools-tools)
+  if (SPIRV-Tools_FOUND AND SPIRV-Tools-tools_FOUND)
+    set(SPIRV_TOOLS_FOUND TRUE)
+    set(SPIRV_TOOLS_LDFLAGS SPIRV-Tools-shared)
+    get_target_property(SPIRV_TOOLS_INCLUDE_DIRS SPIRV-Tools-shared INTERFACE_INCLUDE_DIRECTORIES)
+  endif()
+endif()
+
+if (NOT SPIRV_TOOLS_FOUND)
```
MrSidims, line 128: Shouldn't it be like
```suggestion
if (NOT SPIRV_TOOLS_FOUND AND NOT SPIRV-Tools_FOUND AND NOT SPIRV-Tools-tools_FOUND)
```
?

test/CMakeLists.txt
dimula73: 'find_package(SPIRV-Tools)' is used to add targets for shared libraries. Basically, they provide us `SPIRV_TOOLS_LDFLAGS` that is used here: https://github.com/KhronosGroup/SPIRV-LLVM-Translator/blob/main/tools/llvm-spirv/CMakeLists.txt#L32

And this particular line is used to distinguish between the pkgconfig and cmake paths.

test/CMakeLists.txt
dimula73: This particular branch doesn't use anything from `SPIRV-Tools` package...

CMakeLists.txt
dimula73: Fix pushed, thanks :)

CMakeLists.txt
dimula73: Well, it doesn't change anything in this particular case, but might be good for the safe of consistency

MrSidims (approved):
LGTM

test/CMakeLists.txt
```diff
@@ -22,19 +23,44 @@ if(SPIRV_TOOLS_FOUND)
...
+    set(SPIRV_TOOLS_SPIRV_LINK_FOUND True)
+  endif()
+
+  if (TARGET spirv-val)
+    set(SPIRV_TOOLS_SPIRV_VAL_FOUND True)
+  endif()
+
+  get_target_property(SPIRV_AS_PATH spirv-as IMPORTED_LOCATION_RELEASE)
+  if (NOT SPIRV_AS_PATH)
+	  get_target_property(SPIRV_AS_PATH spirv-as IMPORTED_LOCATION_DEBUG)
```
svenvh, line 43: nit: the indent here looks inconsistent.

test/CMakeLists.txt
```diff
@@ -22,19 +23,44 @@ if(SPIRV_TOOLS_FOUND)
...
+  endif()
+
+  if (SPIRV_AS_PATH)
+	  cmake_path(GET SPIRV_AS_PATH PARENT_PATH SPIRV_AS_PATH)
```
svenvh, line 47: Another inconsistent indent.

MrSidims:
@dimula73 please address comments from Sven before merging.

test/CMakeLists.txt
dimula73: fixed

svenvh (approved):

PR #1795, opened 2022-12-27
Do not use AttributeList::removeParamAttributes

haonanya:
AttributeList::removeParamAttributes function sets attribute at specified
index empty so that return value of AttributeList::getNumAttrSet() keeps
unchanged after that call. When call BuiltinCallMutator::removeArg function,
there is assert failure on BuiltinCallMutator::doConversion() since new CallInst
removed arg but still holds attribute of that removed arg.

### DIFF
--- a/lib/SPIRV/SPIRVBuiltinHelper.cpp
+++ b/lib/SPIRV/SPIRVBuiltinHelper.cpp
@@ -184,6 +184,15 @@ BuiltinCallMutator &BuiltinCallMutator::removeArg(unsigned Index) {
   // If the argument being dropped is the last one, there is nothing to move, so
   // just remove the attributes.
   if (Index == Args.size() - 1) {
+    // TODO: Remove this workaround when LLVM fixes
+    // https://github.com/llvm/llvm-project/issues/59746 on
+    // AttributeList::removeParamAttributes function.
+    // AttributeList::removeParamAttributes function sets attribute at
+    // specified index empty so that return value of
+    // AttributeList::getNumAttrSet() keeps unchanged after that call. When call
+    // BuiltinCallMutator::removeArg function, there is assert failure on
+    // BuiltinCallMutator::doConversion() since new CallInst removed arg but
+    // still holds attribute of that removed arg.
     SmallVector<AttributeSet, 4> ArgAttrs;
     for (unsigned I = 0; I < Index; ++I)
       ArgAttrs.push_back(Attrs.getParamAttrs(I));
--- a/lib/SPIRV/SPIRVBuiltinHelper.cpp
+++ b/lib/SPIRV/SPIRVBuiltinHelper.cpp
@@ -183,9 +183,13 @@ BuiltinCallMutator &BuiltinCallMutator::replaceArg(unsigned Index,
 BuiltinCallMutator &BuiltinCallMutator::removeArg(unsigned Index) {
   // If the argument being dropped is the last one, there is nothing to move, so
   // just remove the attributes.
-  if (Index == Args.size() - 1)
-    Attrs = Attrs.removeParamAttributes(CI->getContext(), Index);
-  else
+  if (Index == Args.size() - 1) {
+    SmallVector<AttributeSet, 4> ArgAttrs;
+    for (unsigned I = 0; I < Index; ++I)
+      ArgAttrs.push_back(Attrs.getParamAttrs(I));
+    Attrs = AttributeList::get(CI->getContext(), Attrs.getFnAttrs(),
+                               Attrs.getRetAttrs(), ArgAttrs);
+  } else
     moveAttributes(CI->getContext(), Attrs, Index + 1, Args.size() - Index - 1,
                    Index);
   Args.erase(Args.begin() + Index);


haonanya:
@MrSidims , could you please take a look? Thanks very much.

lib/SPIRV/SPIRVToOCL.cpp
```diff
@@ -642,6 +642,12 @@ void SPIRVToOCLBase::visitCallGenericCastToPtrExplicitBuiltIn(CallInst *CI,
   default:
     llvm_unreachable("Invalid address space");
   }
+
+  AttributeList Attrs = CI->getCalledFunction()->getAttributes();
+  SmallVector<AttributeSet, 1> ArgAttrs = {Attrs.getParamAttrs(0)};
+  AttributeList NewAttrs = AttributeList::get(*Ctx, Attrs.getFnAttrs(),
+                                              Attrs.getRetAttrs(), ArgAttrs);
```
AlexeySachkov, line 649: Why not use [`AttributeList::removeParamAttributes`](https://llvm.org/doxygen/classllvm_1_1AttributeList.html#a14fe429b8e7e3af4cab47a826ae6918f)?

AlexeySachkov:
> visitCallGenericCastToPtrExplicitBuiltIn removes 1th argument but its corresponding attribute isn't removed which cause assert failures in SPIRVBuiltinHelper.cpp.

It seems like we have a missing LIT test, could you please add it?

lib/SPIRV/SPIRVToOCL.cpp
haonanya: Actually https://github.com/KhronosGroup/SPIRV-LLVM-Translator/blob/main/lib/SPIRV/SPIRVToOCL.cpp#L648 will call [AttributeList::removeParamAttributes](https://llvm.org/doxygen/classllvm_1_1AttributeList.html#a14fe429b8e7e3af4cab47a826ae6918f), but it don't work. We still have 2 function parameter attribute although the 1th argument is removed.

AlexeySachkov: I feel like there is a bug in `removeArg` implementation, because it only modifies internal field `Attrs` of `BuiltinCallMutator` class instance, but I don't see it applying new attributes for the resulting instruction anywhere.

If we already try to handle attributes in that helper, I would prefer us to fix that helper instead of adding such quickfixes everywhere: I suppose that this is not the only place where we leave extra attributes after removing call arguments.

haonanya: @AlexeySachkov, New attributes is applied on https://github.com/KhronosGroup/SPIRV-LLVM-Translator/blob/main/lib/SPIRV/SPIRVBuiltinHelper.cpp#L102,L104, bug is from https://github.com/llvm/llvm-project/blob/main/llvm/lib/IR/Attributes.cpp#L1422 which cause after call removeParamAttributes function, Attrs.getNumAttrSets() still unchanged.  Maybe I can submit patch to LLVM, SPIRV don't need any change.
+@jcranmer-intel

AlexeySachkov: If there is a bug in the upstream LLVM, we should definitely report it.

If you have a strong case with this particular instruction: I think we can merge this as a temporary workaround, but we should left a comment referencing upstream issue and a test for this case. 

However, my preference would be to fix this in the upstream first, because we are likely have more instructions for which we have the same bug with extra attributes: ideally, we should fix them all at once and not to insert a workaround for each.

haonanya: Filed an issue in LLVM https://github.com/llvm/llvm-project/issues/59746

haonanya:
> > visitCallGenericCastToPtrExplicitBuiltIn removes 1th argument but its corresponding attribute isn't removed which cause assert failures in SPIRVBuiltinHelper.cpp.
> 
> It seems like we have a missing LIT test, could you please add it?

I'm not able to provide a lit test because currently noundef attribute is not preserved in spirv.

lib/SPIRV/SPIRVToOCL.cpp
haonanya: @AlexeySachkov, I don't know if the AttributeList::removeParamAttributes has bug or not so I filed an issue in LLVM. And the patch removes AttributeList::removeParamAttributes call in SPIRVBuiltinHelper.cpp so we don't need add corner case for every instruction that need remove arg. Thanks very much for your review.

MrSidims: I'm OK with the fix with TODO placed in the code and in the translator's issues.

MrSidims:
> > > visitCallGenericCastToPtrExplicitBuiltIn removes 1th argument but its corresponding attribute isn't removed which cause assert failures in SPIRVBuiltinHelper.cpp.
> > 
> > 
> > It seems like we have a missing LIT test, could you please add it?
> 
> I'm not able to provide a lit test because currently noundef attribute is not preserved in spirv.

The problem seems to be a bit more general, than just about `GenericCastToPtrExplicit` instruction and `noundef `attribute. It can be emulated by writing LLVM IR with, for example, `__spirv_AtomicStore` call with `zeroext` attribute put on the 2nd parameter and then calling just SPIRVToOCL12 pass (if I read the code correctly, we will call `.removeArg(2)` which should end up in assertion you observe). Don't have a strong opinion though if it's worth an effort to add such test.

lib/SPIRV/SPIRVToOCL.cpp
haonanya: done.

MrSidims (approved):
LGTM but would like to get an approval from @AlexeySachkov and @svenvh as well

lib/SPIRV/SPIRVBuiltinHelper.cpp
```diff
@@ -183,9 +183,21 @@ BuiltinCallMutator &BuiltinCallMutator::replaceArg(unsigned Index,
 BuiltinCallMutator &BuiltinCallMutator::removeArg(unsigned Index) {
   // If the argument being dropped is the last one, there is nothing to move, so
   // just remove the attributes.
-  if (Index == Args.size() - 1)
-    Attrs = Attrs.removeParamAttributes(CI->getContext(), Index);
-  else
+  if (Index == Args.size() - 1) {
+    // TODO: Remove this workaround when LLVM fixes potential bug on
```
svenvh, line 187: Perhaps it would be useful to include a link to the issue.
```suggestion
    // TODO: Remove this workaround when LLVM fixes https://github.com/llvm/llvm-project/issues/59746 on
```

svenvh (approved):

AlexeySachkov (approved):

PR #1796, opened 2022-12-28
Fix SPV_INTEL_runtime_aligned implementation part 1

MrSidims:
It was implemented via new decoration, which is not correct. Instead it should be Function Parameter Attribute decoration.

In this commit starts fixing this in step-by-step manner.

This step only adds reverse translation to be reused later by SPIR-V consumers. 

Link to spec: https://github.com/KhronosGroup/SPIRV-Registry/blob/main/extensions/INTEL/SPV_INTEL_runtime_aligned.asciidoc


MrSidims:
@broxigarchen may I ask you to take a look?

MrSidims:
@broxigarchen @tiwaria1  @asudarsa @vmaksimo @jgstarIntel @maksimsab please take a look

lib/SPIRV/SPIRVReader.cpp
```diff
@@ -2798,6 +2798,10 @@ void SPIRVToLLVM::transFunctionAttrs(SPIRVFunction *BF, Function *F) {
     mapValue(BA, &(*I));
     setName(&(*I), BA);
     BA->foreachAttr([&](SPIRVFuncParamAttrKind Kind) {
+      // Skip this function parameter attribute as it will translated among
```
vmaksimo, line 2801: Just to clarify, why wasn't it skipped before?

vmaksimo (commented):
Only from the code I understood that somehow we want to save the non-conformant implementation for now. Please update the description about what this step do. Also the link to the spec could be useful.

Am I right that `test/transcoding/SPV_INTEL_runtime_aligned/RuntimeAligned.ll` is not changed for now because we preserve the old implementation?

MrSidims:
> Only from the code I understood that somehow we want to save the non-conformant implementation for now. Please update the description about what this step do. Also the link to the spec could be useful.
> 
> Am I right that `test/transcoding/SPV_INTEL_runtime_aligned/RuntimeAligned.ll` is not changed for now because we preserve the old implementation?

Added in the description. Yeap, the test is unchanged intentionally

lib/SPIRV/SPIRVReader.cpp
MrSidims: Because it wasn't a function parameter attribute (since it was implemented incorrectly)

tiwaria1:
> > Hey Dmitry, what did this look like in IR after reverse-translation before this change? What will the IR be after this change? (I am guessing kernel parameter attribute?)
> 
> @tiwaria1 It remains to be the same - a function metadata looking like: !kernel_arg_runtime_aligned !N !N = !{i1 true, i1 false, i1 true, i1 false, i1 false} (see both CHECK-LLVM in the test in this PR and https://github.com/KhronosGroup/SPIRV-LLVM-Translator/blob/main/test/extensions/INTEL/SPV_INTEL_runtime_aligned/RuntimeAligned.ll for reference


Thanks, I asked to understand if FPGA backend would need to be updated after this fix to handle IR changes. Looks like it will not be needed.

lib/SPIRV/SPIRVReader.cpp
```diff
@@ -197,19 +197,19 @@ static void addRuntimeAlignedMetadata(
     LLVMContext *Context, SPIRVFunction *BF, llvm::Function *Fn,
     std::function<Metadata *(SPIRVFunctionParameter *)> ForeachFnArg) {
   std::vector<Metadata *> ValueVec;
-  bool DecorationFound = false;
+  bool RuntimeAlignedFound = false;
   BF->foreachArgument([&](SPIRVFunctionParameter *Arg) {
-    if (Arg->getType()->isTypePointer() &&
+    if (Arg->hasAttr(FunctionParameterAttributeRuntimeAlignedINTEL) ||
         Arg->hasDecorate(internal::DecorationRuntimeAlignedINTEL)) {
```
broxigarchen, line 203: we finally will remove this line right?

broxigarchen:
Sorry I guess your previous email get buried in my mail box and I just saw it now.

PR looks good to me. I guess the SPIRV consumer might need to merge the patch later after it's finalized. Just curious why we need to do this step by step?

lib/SPIRV/SPIRVReader.cpp
```diff
@@ -197,19 +197,19 @@ static void addRuntimeAlignedMetadata(
...
-      DecorationFound = true;
+      RuntimeAlignedFound = true;
       ValueVec.push_back(ForeachFnArg(Arg));
```
asudarsa, line 205: ForeachFnArg seems unnecessary here. We can just create a ConstantInt with 1 and push it here.

lib/SPIRV/libSPIRV/SPIRVIsValidEnum.h
```diff
@@ -167,6 +167,7 @@ inline bool isValid(spv::FunctionParameterAttribute V) {
   case FunctionParameterAttributeNoCapture:
   case FunctionParameterAttributeNoWrite:
   case FunctionParameterAttributeNoReadWrite:
+  case FunctionParameterAttributeRuntimeAlignedINTEL:
```
asudarsa, line 170: I suppose this change will be useful when SPIRVWriter.cpp is updated.

lib/SPIRV/SPIRVReader.cpp
```diff
@@ -197,19 +197,19 @@ static void addRuntimeAlignedMetadata(
...
     } else {
       llvm::Metadata *DefaultNode = ConstantAsMetadata::get(
```
asudarsa, line 207: One nit: Instead of creating DefaultNode for each Argument, we can create a single static global DefaultNode and push it when required.

asudarsa (commented):
Few nits. Are not PR blocking.

MrSidims:
> Just curious why we need to do this step by step?

Without consumers update such SPIR-V changes will result in a missing feature. And once it's updated - it takes time until new driver is being propagated in different CIs (and SPIR-V generator change might reach CI at this point). Also I don't have much control over every SPIR-V consumer, hence prefer to go with conservative approach anyway.

asudarsa:
Closing and reopening to run tests.

asudarsa:
Hi @MrSidims,

Is this change ready to be merged? Also, IIUC, the forward translation is the next and final step in this effort?

Thanks

lib/SPIRV/SPIRVReader.cpp
vmaksimo: It actually does exactly what you're mentioned, see https://github.com/KhronosGroup/SPIRV-LLVM-Translator/pull/1796/files#diff-7990d99acb2f8325d245dfecdab758eae5aebae2b54c04fd8e906212eb80ae10R4472-R4473

lib/SPIRV/SPIRVReader.cpp
vmaksimo: Right, I believe that is the intention - but only after consumers are ready to work with the correct implementation.

lib/SPIRV/SPIRVReader.cpp
vmaksimo: Guess no need to have it as a global, moved it out of lambda to have it initialized only once in the function.

asudarsa (approved):
Translator change Looks good. Thanks

broxigarchen (approved):

PR #1797, opened 2022-12-28
Translate saturation addition and substraction LLVM intrinsics into OpenCL instructions

maksimsab:


maksimsab:
Hi @jcranmer-intel!
Can you remind please why do we need a lowering for @llvm.sadd.sat in this [PR](https://github.com/KhronosGroup/SPIRV-LLVM-Translator/pull/1524)?
And can we replace it with s_add_sat instruction?
And do we need a lowering for overflow version of intrinsic?

lib/SPIRV/SPIRVLowerSaddWithOverflow.cpp
```diff
@@ -0,0 +1,149 @@
+//===- SPIRVLowerSaddWithOverflow.cpp - Lower llvm.sadd.with.overflow -----===//
```
MrSidims, line 1: Copyright from lib/SPIRV/SPIRVLowerSaddIntrinsics.cpp and hence the one here is incorrect. Please use the one from https://github.com/KhronosGroup/SPIRV-LLVM-Translator/blob/main/lib/SPIRV/SPIRVTypeScavenger.cpp (shouldn't be Intel, but Khronos in the first place)

MrSidims (commented):
LGTM with one comment
This was missing in the previous implementation and is missing in this one as well. When we substitute intrinsic with a function call metadata attached to the intrinsic is not being copied. I suggest to fix it.

lib/SPIRV/SPIRVLowerSaddWithOverflow.cpp
```diff
@@ -0,0 +1,149 @@
...
+  }
+
+  // Redirect @llvm.sadd.with.overflow.* call to the function we have in
+  // the loaded module @llvm_sadd_with_overflow_*
+  Function *F = Mod->getFunction(FuncName);
+  if (F) { // This function is already linked in.
+    I.setCalledFunction(F);
+    return;
+  }
+  FunctionCallee FC = Mod->getOrInsertFunction(FuncName, I.getFunctionType());
```
MrSidims, line 89: Here should also apply metadata that was previously put on llvm intrinsic call.

jcranmer-intel:
> And do we need a lowering for overflow version of intrinsic?

Removing compilation support for an existing LLVM intrinsic is generally a bad idea. In principle, we should support all non-target-specific LLVM intrinsics, but obviously, this is subject to the realities of limited developer resources.

lib/SPIRV/SPIRVLowerSaddWithOverflow.cpp
maksimsab: Here an intrinsic instruction stays the same and metadata does not change. Only called function is being changed what does not influence a metadata. Nevertheless, I added a change in test for metadata.

lib/SPIRV/SPIRVLowerSaddWithOverflow.cpp
maksimsab: fixed it.

MrSidims (approved):
LGTM
But it requires approval from @jcranmer-intel

PR #1799, opened 2023-01-03
[OpaquePointers] Add support for translating target extension types.

jcranmer-intel:
The target extension type for SPIR-V is essentially target("spirv.TypeName", <image type>, <int params>).

Most of the work to support translation of these types has already happened beforehand, so the primary step here is to enable translation work in SPIRVWriter as well as making the SPIRVBuiltinHelpers work with target types as well.

Constructing LLVM IR from SPIR-V using these types is not yet supported, mainly out of uncertainty of the proper interface to let the resultant consumers indicate that they wish to support these types.


jcranmer-intel:
The clang patch I wrote to generate these types is at https://reviews.llvm.org/D141008.

svenvh (approved):

MrSidims (approved):
LGTM, but I suggest to wait for clang patch first to make sure that IR format is approved.
I also believe, that some of the .cl tests might require to be updated when clang patch is merged.

PR #1800, opened 2023-01-03
Translate Intel-specific RegisterAllocMode metadata

sarnex:
Frontends can use the RegisterAllocMode metadata can be used to configure the register allocation mode. Here we translate into a form that Intel Graphics Compiler can use to actually apply the configuration.

This will only be used by Intel and won't affect anything else.

This is only a temporary solution, we intend to introduce an extension in the future.

### DIFF
--- a/lib/SPIRV/SPIRVWriter.cpp
+++ b/lib/SPIRV/SPIRVWriter.cpp
@@ -918,7 +918,7 @@ SPIRVFunction *LLVMToSPIRVBase::transFunctionDecl(Function *F) {
       // means small grf mode
       std::string NumThreads = RegisterAllocNodeMDOp == 2 ? "4" : "8";
       BF->addDecorate(new SPIRVDecorateUserSemanticAttr(
-                          BF, "num-thread-per-eu " + NumThreads));
+          BF, "num-thread-per-eu " + NumThreads));
     }
   }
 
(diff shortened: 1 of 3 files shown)


sarnex:
@MrSidims Hey this is the one we discussed over email

lib/SPIRV/SPIRVWriter.cpp
```diff
@@ -901,6 +901,26 @@ SPIRVFunction *LLVMToSPIRVBase::transFunctionDecl(Function *F) {
 
   transFPGAFunctionMetadata(BF, F);
 
+  if (auto *RegisterAllocModeMD = F->getMetadata("RegisterAllocMode")) {
```
MrSidims, line 904: Probably should move this code into a separate function like:
`transFunctionMetadataAsUserSemanticDecoration` to separate this a bit

test/transcoding/registerallocmode.ll
```diff
@@ -0,0 +1,74 @@
+; RUN: llvm-as %s -o %t.bc
+; RUN: llvm-spirv -spirv-text %t.bc -o - | FileCheck %s --check-prefix=CHECK-SPIRV
+; RUN: llvm-spirv %t.bc -o %t.spv
+; RUN: spirv-val %t.spv
```
MrSidims, line 4: Okay, seems like the validator doesn't handle UserSemantic properly, good for us :)

MrSidims (commented):
Just 1 nit, other than that LGTM

lib/SPIRV/SPIRVWriter.cpp
sarnex: good idea, done in latest commit

test/transcoding/registerallocmode.ll
sarnex: :)

MrSidims (approved):

PR #1801, opened 2023-01-04
Decorate the pointer argument of OpLoad and Opstore with BufferLocationINTEL

broxigarchen:
This PR is regarding spec https://github.com/KhronosGroup/SPIRV-Registry/pull/185

This patch parse the string in a "ptr.annotation" call, extracting buffer location information and create a SPIRV decoration on the pointer using the buffer location information. This applies only on the pointer argument of OpLoad/Opstore

### DIFF
--- a/lib/SPIRV/SPIRVWriter.cpp
+++ b/lib/SPIRV/SPIRVWriter.cpp
@@ -2829,6 +2829,12 @@ using DecorationsInfoVec =
 struct AnnotationDecorations {
   DecorationsInfoVec MemoryAttributesVec;
   DecorationsInfoVec MemoryAccessesVec;
+  DecorationsInfoVec BufferLocationVec;
+
+  bool empty() {
+    return (MemoryAttributesVec.empty() && MemoryAccessesVec.empty() &&
+            BufferLocationVec.empty());
+  }
 };
 
 struct IntelLSUControlsInfo {
@@ -3019,6 +3025,8 @@ AnnotationDecorations tryParseAnnotationString(SPIRVModule *BM,
       BM->isAllowedToUseExtension(ExtensionID::SPV_INTEL_fpga_memory_accesses);
   const bool AllowFPGAMemAttr = BM->isAllowedToUseExtension(
       ExtensionID::SPV_INTEL_fpga_memory_attributes);
+  const bool AllowFPGABufLoc =
+      BM->isAllowedToUseExtension(ExtensionID::SPV_INTEL_fpga_buffer_location);
 
   bool ValidDecorationFound = false;
   DecorationsInfoVec DecorationsVec;
@@ -3037,8 +3045,15 @@ AnnotationDecorations tryParseAnnotationString(SPIRVModule *BM,
       std::vector<std::string> DecValues;
       if (tryParseAnnotationDecoValues(ValueStr, DecValues)) {
         ValidDecorationFound = true;
-        DecorationsVec.emplace_back(static_cast<Decoration>(DecorationKind),
-                                    std::move(DecValues));
+
+        if (AllowFPGABufLoc &&
+            DecorationKind == DecorationBufferLocationINTEL) {
+          Decorates.BufferLocationVec.emplace_back(
+              static_cast<Decoration>(DecorationKind), std::move(DecValues));
+        } else {
+          DecorationsVec.emplace_back(static_cast<Decoration>(DecorationKind),
+                                      std::move(DecValues));
+        }
       }
       continue;
     }
@@ -3109,6 +3124,7 @@ AnnotationDecorations tryParseAnnotationString(SPIRVModule *BM,
       DecorationsVec.emplace_back(Dec, std::move(DecValues));
     }
   }
+
   // Even if there is an annotation string that is split in blocks like Intel
   // FPGA annotation, it's not necessarily an FPGA annotation. Translate the
   // whole string as UserSemantic decoration in this case.
@@ -3225,6 +3241,18 @@ void addAnnotationDecorations(SPIRVEntry *E, DecorationsInfoVec &Decorations) {
         E->addDecorate(I.first, Result);
       }
     } break;
+    case DecorationBufferLocationINTEL: {
+      if (M->isAllowedToUseExtension(
+              ExtensionID::SPV_INTEL_fpga_buffer_location)) {
+        M->getErrorLog().checkError(I.second.size() == 1,
+                                    SPIRVEC_InvalidLlvmModule,
+                                    "Decoration requires a single argument.");
+        SPIRVWord Result = 0;
+        StringRef(I.second[0]).getAsInteger(10, Result);
+        E->addDecorate(I.first, Result);
+      }
+    } break;
+
     default:
       // Other decorations are either not supported by the translator or
       // handled in other places.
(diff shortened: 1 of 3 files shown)


MrSidims (changes requested):
OpStore doesn't have an ID to decorate. Probably the design should be reconsidered.

broxigarchen:
> OpStore doesn't have an ID to decorate. Probably the design should be reconsidered.

Sorry this is not decorating the OpStore directly. I'll change the title

MrSidims (commented):
The patch is LGTM, but test results says otherwise. Awaiting for spec review and test fix.

broxigarchen:
> The patch is LGTM, but test results says otherwise. Awaiting for spec review and test fix.

It seems there is an opaque pointer change happened recently on llvm.ptr.annotation intrinsic. Updated the unit test file

MrSidims:
Restarting CI with a new HEAD

lib/SPIRV/SPIRVWriter.cpp
```diff
@@ -3490,6 +3518,26 @@ static bool allowsApproxFunction(IntrinsicInst *II) {
...
+      for (auto *Sussessor : Cast->users())
+        UserList.push_back(Sussessor);
+    } else {
+      UserList.push_back(Inst);
+    }
+  }
+
+  for (auto &Inst : UserList)
+    if (isa<LoadInst>(Inst) || isa<StoreInst>(Inst))
+      return true;
```
MrSidims, line 3536: Lets imagine a situation, when we have a pointer to a composite object. The memory this pointer points to has a specific memory location and we annotate it appropriately. Now we access this composite via GEP. I would assume, that the result pointer should point to the same specific memory and hence should be decorated, but it's seems like it won't happen. Is it intentional or it wasn't not considered?

lib/SPIRV/SPIRVWriter.cpp
```diff
@@ -3490,6 +3518,26 @@ static bool allowsApproxFunction(IntrinsicInst *II) {
            cast<VectorType>(Ty)->getElementType()->isFloatTy()));
 }
 
+bool allowDecorateWithBufferLocationINTEL(IntrinsicInst *II) {
+  SmallVector<Value *, 8> UserList;
+
+  for (auto *Inst : II->users()) {
```
MrSidims, line 3524: Should we use [stripPointerCasts](https://llvm.org/doxygen/classllvm_1_1Value.html#a966eb231e7d4e572874d2cb49b18faea) here?

MrSidims (commented):
@vmaksimo please also take a look

lib/SPIRV/SPIRVWriter.cpp
vmaksimo: The PR title says "decorating a scalar pointer", so maybe this case is not covered intentionally.

test/extensions/INTEL/SPV_INTEL_fpga_buffer_location/FPGABufferLocation.ll
```diff
@@ -30,20 +31,75 @@
...
+  %arg_a.addr = alloca i32 addrspace(1)*, align 8
+  %MyIP = alloca %struct.MyIP, align 8
+  %arg_a.addr.ascast = addrspacecast i32 addrspace(1)** %arg_a.addr to i32 addrspace(1)* addrspace(4)*
+  %MyIP.ascast = addrspacecast %struct.MyIP* %MyIP to %struct.MyIP addrspace(4)*
+  store i32 addrspace(1)* %arg_a, i32 addrspace(1)* addrspace(4)* %arg_a.addr.ascast, align 8
+  %0 = bitcast %struct.MyIP* %MyIP to i8*
+  %a = getelementptr inbounds %struct.MyIP, %struct.MyIP addrspace(4)* %MyIP.ascast, i32 0, i32 0
+  %1 = bitcast i32 addrspace(4)* addrspace(4)* %a to i8 addrspace(4)*
+  %2 = call i8 addrspace(4)* @llvm.ptr.annotation.p4i8(i8 addrspace(4)* %1, i8* getelementptr inbounds ([19 x i8], [19 x i8]* @.str.4, i32 0, i32 0), i8* getelementptr inbounds ([9 x i8], [9 x i8]* @.str.1, i32 0, i32 0), i32 7, i8* null)
+; CHEKC-LLVM: call i32 addrspace(4)* addrspace(4)* @llvm.ptr.annotation.p4p4i32.p0i8(i32 addrspace(4)* addrspace(4)* %a, i8* getelementptr inbounds ([33 x i8], [33 x i8]* @0, i32 0, i32 0), i8* undef, i32 undef, i8* undef)
```
vmaksimo, line 79: Would be nice to check the content of generated annotation strings (e.g., `@0`)

test/extensions/INTEL/SPV_INTEL_fpga_buffer_location/FPGABufferLocation.ll
```diff
@@ -1,13 +1,13 @@
-; RUN: llvm-as %s -o %t.bc
-; RUN: llvm-spirv %t.bc --spirv-ext=+SPV_INTEL_fpga_buffer_location -o %t.spv
+; RUN: llvm-as -opaque-pointers=0 %s -o %t.bc
```
vmaksimo, line 1: I think we'd better stay with the opaque pointers enabled approach whenever possible

lib/SPIRV/SPIRVWriter.cpp
```diff
@@ -3109,9 +3124,10 @@ AnnotationDecorations tryParseAnnotationString(SPIRVModule *BM,
       DecorationsVec.emplace_back(Dec, std::move(DecValues));
     }
   }
+
   // Even if there is an annotation string that is split in blocks like Intel
   // FPGA annotation, it's not necessarily an FPGA annotation. Translate the
-  // whole string as UserSemantic decoration in this case.
+  // whole string as UserSemantic decoration in this case
```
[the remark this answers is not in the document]
broxigarchen: LOL. This is a change by mistake. I can revert it.

lib/SPIRV/SPIRVWriter.cpp
broxigarchen: The title is a bit misleading. Let me update it.

I think the confusion is coming from the title that we are trying to decorate scalar pointers. If we want to do this, Like Dimitry said, we will want to find all alias pointers and apply the Bufferlocation ID on all of them. This is the previous approach we tried, such that by applying Bufferlocation ID on a kernel arg and trying to find all alias pointer of it. But it seems we find a lot of difficulties in this implementation.

This new implementation is to decorate the pointer argument of OpLoad and Opstore with BufferLocationINTEL that given in its predecessor llvm.ptr.annotation intrinsic. The form of the IR is secured by a header file and builtin functions so we can keep the llvm.ptr.annotation to Opload/Opstore as close as possible in all occurrences.

MrSidims: Okay, implementation-wise it makes sense. Should we change the description in the spec from "Indicates that the pointer must only point into ..." to "Indicates that the pointer and all of its aliases must only point into ..."? I'll duplicate this question in https://github.com/KhronosGroup/SPIRV-Registry/pull/185

broxigarchen: I double-checked with co-workers and spec. We don't need to handle GEP cases. Just need the BufferLocation ID on the annotated pointer. Also added comments in the spec PR

lib/SPIRV/SPIRVWriter.cpp
broxigarchen: Probably do not need to. The reason is the same as above.

lib/SPIRV/SPIRVWriter.cpp
MrSidims: Resolving this comment. Only nits remain to fix.

lib/SPIRV/SPIRVWriter.cpp
```diff
@@ -3490,6 +3518,26 @@ static bool allowsApproxFunction(IntrinsicInst *II) {
            cast<VectorType>(Ty)->getElementType()->isFloatTy()));
 }
 
+bool allowDecorateWithBufferLocationINTEL(IntrinsicInst *II) {
+  SmallVector<Value *, 8> UserList;
+
+  for (auto *Inst : II->users()) {
+    // if castInst, push Sussessors
```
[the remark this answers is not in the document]
broxigarchen: done

test/extensions/INTEL/SPV_INTEL_fpga_buffer_location/FPGABufferLocation.ll
broxigarchen: added.

test/extensions/INTEL/SPV_INTEL_fpga_buffer_location/FPGABufferLocation.ll
broxigarchen: Tried to replace"-opaque-pointers=0" with "-emit-opaque-pointers", it seems it does not work anymore. Do I need to change all pointers with "ptr" to get it work?

broxigarchen: done

PR #1809, opened 2023-01-12
Translate aliases to aliasee value in more contexts

dwoodwor-intel:
Currently llvm-spirv crashes with an assert when translating uses of aliases which are not the callee of a function call instruction. This change remedies this by handling these uses of aliases the same way
function call callees are handled: by using the value of their aliasee.

### DIFF
--- a/test/extensions/INTEL/SPV_INTEL_function_pointers/alias.ll
+++ b/test/extensions/INTEL/SPV_INTEL_function_pointers/alias.ll
@@ -9,27 +9,27 @@ target triple = "spir64-unknown-unknown"
 ; Check that aliases are dereferenced and translated to their aliasee values
 ; when used since they can't be translated directly.
 
-; CHECK-SPIRV-DAG: 3 Name [[#FOO:]] "foo"
-; CHECK-SPIRV-DAG: 3 Name [[#BAR:]] "bar"
-; CHECK-SPIRV-DAG: 3 Name [[#Y:]] "y"
-; CHECK-SPIRV-DAG: 5 Name [[#FOOPTR:]] "foo.alias"
-; CHECK-SPIRV-DAG: 5 Decorate [[#FOO]] LinkageAttributes "foo" Export
-; CHECK-SPIRV-DAG: 5 Decorate [[#BAR]] LinkageAttributes "bar" Export
-; CHECK-SPIRV-DAG: 4 TypeInt [[#I32:]] 32 0
-; CHECK-SPIRV-DAG: 4 TypeInt [[#I64:]] 64 0
-; CHECK-SPIRV-DAG: 4 TypeFunction [[#FOO_TYPE:]] [[#I32]] [[#I32]]
-; CHECK-SPIRV-DAG: 2 TypeVoid [[#VOID:]]
-; CHECK-SPIRV-DAG: 4 TypePointer [[#I64PTR:]] 7 [[#I64]]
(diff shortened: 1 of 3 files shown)


dwoodwor-intel:
Let me know if this is the right approach or if there's a different way these should be handled.

Also, I am seeing these five tests fail with this but I'm seeing them fail without my change too so they seem to be existing failures unrelated to this change:

Failed Tests (5):
  LLVM_SPIRV :: matrix_times_matrix.spt
  LLVM_SPIRV :: matrix_times_scalar.spt
  LLVM_SPIRV :: matrix_times_vector.spt
  LLVM_SPIRV :: vector_times_matrix.spt
  LLVM_SPIRV :: vector_times_scalar.spt

dwoodwor-intel:
It seems like spirv-val is not happy with this, so it seems like something else is needed:

```
 ******************** TEST 'LLVM_SPIRV :: transcoding/alias.ll' FAILED ********************
Script:
--
# ...
: 'RUN: at line 4';   /usr/bin/spirv-val /home/runner/work/SPIRV-LLVM-Translator/SPIRV-LLVM-Translator/build/test/test_output/transcoding/Output/alias.ll.tmp.spv
# ...
--
Exit Code: 1

Command Output (stderr):
--
error: line 30: Invalid use of function result id '4[%foo_alias]'.
  %14 = OpConvertPtrToU %ulong %foo_alias
```

I don't think this is specific to aliases; it seems to be a larger issue because the code I'm dealing with involves function pointers. Would I need to use SPV_INTEL_function_pointers to get around this?

dwoodwor-intel:
I've updated this to use SPV_INTEL_function_pointers; it looks like all of the tests are passing now, but my understanding is that spirv-val doesn't implement that extension and so I have to remove it from the test. Let me know if this looks right or if it should be translated differently.

test/extensions/INTEL/SPV_INTEL_function_pointers/alias.ll
```diff
@@ -0,0 +1,49 @@
...
+; CHECK-SPIRV-DAG: 5 Decorate [[#FOO]] LinkageAttributes "foo" Export
+; CHECK-SPIRV-DAG: 5 Decorate [[#BAR]] LinkageAttributes "bar" Export
+; CHECK-SPIRV-DAG: 4 TypeInt [[#I32:]] 32 0
+; CHECK-SPIRV-DAG: 4 TypeInt [[#I64:]] 64 0
+; CHECK-SPIRV-DAG: 4 TypeFunction [[#FOO_TYPE:]] [[#I32]] [[#I32]]
+; CHECK-SPIRV-DAG: 2 TypeVoid [[#VOID:]]
+; CHECK-SPIRV-DAG: 4 TypePointer [[#I64PTR:]] 7 [[#I64]]
+; CHECK-SPIRV-DAG: 4 TypeFunction [[#BAR_TYPE:]] [[#VOID]] [[#I64PTR]]
+; CHECK-SPIRV-DAG: 4 TypePointer [[#FOOPTR_TYPE:]] 7 [[#FOO_TYPE]]
+; CHECK-SPIRV-DAG: 4 ConstantFunctionPointerINTEL [[#FOOPTR_TYPE]] [[#FOOPTR]] [[#FOO]]
```
MrSidims, line 25: This seems the right way to go.

MrSidims (approved):
nit: no need to check number of SPIR-V words of instruction in test(s) (unless these tests contains something like a very big constant with a width > 64 bits, and you want to check if the parameters were processed correctly).

PR #1810, opened 2023-01-17
Update for another parseIR LLVM API change

dwoodwor-intel:
The parseIR API has been updated again in b56df190b013 to require a ParserCallbacks object which is not implicitly convertible from a lambda. This change adds an explicit conversion to fix the build failure.


dwoodwor-intel:
(out-of-tree builds are going to fail for this too until they're updated)

MrSidims (approved):
LGTM, thanks!

svenvh (approved):

PR #1826, opened 2023-01-25
Update LLVM version from 16 to 17

svenvh:


MrSidims (approved):

CMakeLists.txt
```diff
@@ -1,7 +1,7 @@
 cmake_minimum_required(VERSION 3.13.4)
```
MrSidims, line 2: TBH I don't have a strong opinion about this, but shouldn't we mimic LLVM here adding:

```
if ("${CMAKE_VERSION}" VERSION_LESS "3.20.0")
  message(WARNING
    "Your CMake version is ${CMAKE_VERSION}. Starting with LLVM 17.0.0, the "
    "minimum version of CMake required to build SPIRV-LLVM-Translator will become 3.20.0, and "
    "using an older CMake will become an error. Please upgrade your CMake to "
    "at least 3.20.0 now to avoid issues in the future!")
endif()

```
?

svenvh: Once LLVM bumps the minimum version we can update our minimum version too I guess.  I wouldn't bother too much otherwise, for the reasons you already mentioned. :)

svenvh:
The CI builds still fail as the `-17` packages are not on apt.llvm.org yet.

CMakeLists.txt
MrSidims: Sure

PR #1827, opened 2023-01-25
Fix static analysis issues

asudarsa:
In some cases, return value of consume_front is ignored. This change fixes the issue.

### DIFF
--- a/lib/SPIRV/OCLToSPIRV.cpp
+++ b/lib/SPIRV/OCLToSPIRV.cpp
@@ -812,7 +812,8 @@ void OCLToSPIRVBase::visitCallGroupBuiltin(CallInst *CI,
                                   .Case("ballot_exclusive_scan", "add")
                                   .Default(FuncName.take_back(
                                       3)); // assumes op is three characters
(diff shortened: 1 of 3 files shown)


lib/SPIRV/OCLToSPIRV.cpp
```diff
@@ -812,7 +812,8 @@ void OCLToSPIRVBase::visitCallGroupBuiltin(CallInst *CI,
                                   .Case("ballot_exclusive_scan", "add")
                                   .Default(FuncName.take_back(
                                       3)); // assumes op is three characters
-          GroupOp.consume_front("_");      // when op is two characters
+          if (!GroupOp.consume_front("_")) // when op is two characters
+            return true;                   // continue
```
MrSidims, line 816: This seems incorrect. Shouldn't it be:
```suggestion
          (void)GroupOp.consume_front("_")); // when op is two characters
```

asudarsa: Ah...I think you are right. We can ignore the result here. Good catch.

Thanks

MrSidims (approved):

PR #1828, opened 2023-01-26
[NFC] Remove -no-opaque-pointers flag from 2 .cl tests and XFAIL them

MrSidims:
Non-opaque pointers are (almost) not supported in LLVM passes, so compilation with higher than O0 optimization level with -no-opaque-pointers flag will result in assertions firing.

To avoid this there are 2 options:
1. Pass -O0 to the tests and update checks;
2. Remove the flag and XFAIL the tests.

Short term option 1. is better, long term - 2. is better.

By this patch I vote for 2.

### DIFF
--- a/test/transcoding/clk_event_t.cl
+++ b/test/transcoding/clk_event_t.cl
@@ -1,12 +1,14 @@
-// RUN: %clang_cc1 -triple spir-unknown-unknown -O1 -cl-std=CL2.0 -fdeclare-opencl-builtins -finclude-default-header -emit-llvm-bc %s -o %t.bc -no-opaque-pointers
-// RUN: llvm-spirv -opaque-pointers=0 %t.bc -spirv-text -o - | FileCheck %s --check-prefix=CHECK-SPIRV
-// RUN: llvm-spirv -opaque-pointers=0 %t.bc -o %t.spv
+// RUN: %clang_cc1 -triple spir-unknown-unknown -O1 -cl-std=CL2.0 -fdeclare-opencl-builtins -finclude-default-header -emit-llvm-bc %s -o %t.bc
+// RUN: llvm-spirv %t.bc -spirv-text -o - | FileCheck %s --check-prefix=CHECK-SPIRV
+// RUN: llvm-spirv %t.bc -o %t.spv
 // RUN: spirv-val %t.spv
-// RUN: llvm-spirv -opaque-pointers=0 -r %t.spv -o %t.rev.bc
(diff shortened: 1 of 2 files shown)


MrSidims:
It depends on https://github.com/KhronosGroup/SPIRV-LLVM-Translator/pull/1826

svenvh:
> so compilation with higher than O0 optimization level with -no-opaque-pointers flag will result in assertions firing.

Just for my understanding: is this an actual breakage?  If so it must be very recent, as I didn't see those tests failing this morning yet.

MrSidims:
> 

My head is on:
```
commit cdeaf5f28c3dca7fa313bdc5657cd07fb08dec5b (HEAD -> main, origin/main, origin/HEAD)
Author: Florian Hahn <flo@fhahn.com>
Date:   Thu Jan 26 11:45:10 2023 +0000

    Recommit "[SCCP] Use range info to prove AddInst has NUW flag."
```

The assert is looking like:
```
clang: LLVM/llvm-project/llvm/lib/Transforms/Scalar/SROA.cpp:1662: llvm::Value* getAdjustedPtr({anonymous}::IRBuilderTy&, const llvm::DataLayout&, llvm::Value*, llvm::APInt, llvm::Type*, const llvm::Twine&): Assertion `Ptr->getType()->isOpaquePointerTy() && "Only opaque pointers upported"' failed.
PLEASE submit a bug report to https://github.com/llvm/llvm-project/issues/ and include the crash backtrace, preprocessed source, and associated run script.
Stack dump:
0.      Program arguments: LLVM/llvm-project/build/bin/clang -cc1 -internal-isystem LLVM/llvm-project/build/lib/clang/17/include -nostdsysteminc -x cl -cl-std=CL2.0 -triple spir64-unknown-unknown -emit-llvm-bc -fdeclare-opencl-builtins -finclude-default-header -Dcl_khr_subgroups Translator/SPIRV-LLVM-Translator/test/transcoding/pipe_builtins.cl -o Translator/build/test/test_output/transcoding/Output/pipe_builtins.cl.tmp.bc -no-opaque-pointers
```

svenvh:
Ah I see now, llvm-project commit `e7fbb381c847 ("[SROA] Break typed pointer support", 2023-01-26)`.

I agree that XFAILing is the better option.

svenvh (approved):

MrSidims:
I no longer observe this failure, so closing the PR (temporary?)

MrSidims:
Actually, still see this locally

svenvh (approved):

Issue #1230, opened 2021-10-04
OpAtomicCompareExchangeWeak is emitted but was removed in SPIR-V 1.4


StuartDBrady:
_OpAtomicCompareExchangeWeak_ is emitted by the translator but it was removed in SPIR-V 1.4. The op was deprecated by SPIR-V 1.2 revision 4 for _all_ versions of SPIR-V, but has identical semantics to _OpAtomicCompareExchange_, which is available in every version of SPIR-V.

_OpAtomicCompareExchangeWeak_ does not directly indicate the success of the compare/exchange operation, and has been mistranslated on consumption to a weak compare/exchange giving a false indication of the operation's success (as fixed for the Khronos SPIR-V/LLVM Translator in pull request #1221). It is generally valid to promote a weak compare/exchange to a strong one that does not require a separate indication of the operation's success. It in generally invalid to remove the indication of a weak compare/exchange's success _without_ promoting it to a strong compare/exchange.

StuartDBrady:
Care should be taken when addressing this to ensure that our test coverage for translation of _OpAtomicCompareExchangeWeak_ to LLVM IR remains adequate. There are certain transcoding tests that currently emit _OpAtomicCompareExchangeWeak_, which is then translated back to LLVM IR. These transcoding tests will no longer generate _OpAtomicCompareExchangeWeak_ once this issue is resolved, which may affect test coverage.

If the coverage of translation of _OpAtomicCompareExchangeWeak_ into IR in _non_-transcoding tests is adequate, then there _might_ not be any need to add any extra non-transcoding testing for this.

MrSidims:
Fixed with https://github.com/KhronosGroup/SPIRV-LLVM-Translator/pull/1830

PR #1830, opened 2023-01-30
updates spirv instructions generated for atomic_compare_exhange_weak

jgstarIntel:
Updating atomic compare exchange weak SPIRV instructions given OpAtomicCompareExchangeWeak is deprecated and has the same semantics as OpAtomicCompareExchange


test/transcoding/atomics.spt
```diff
@@ -40,7 +40,7 @@
 7 AtomicXor 3 24 7 13 14 13 
 7 AtomicAnd 3 25 7 13 14 13 
 9 AtomicCompareExchange 3 26 7 13 14 14 13 17 
-9 AtomicCompareExchangeWeak 3 27 7 13 14 14 13 17 
+9 AtomicCompareExchange 3 27 7 13 14 14 13 17
```
MrSidims, line 43: Why do we need this change? While stopping generation of `OpAtomicCompareExchangeWeak` during LLVM IR to SPIR-V translation, we shouldn't break backward compatibility for previously generated SPIR-V modules.

svenvh:
Relates to #1230 I suppose.

test/transcoding/atomics.spt
asudarsa: I agree that we need to emit OpAtomicCompareExchangeWeak, but selectively for SPIR-V 1.3 or earlier. 
Thanks

MrSidims: Well, we may stop emitting the instruction, it's not a problem. The problem is, that once emitted - it should be supported by the translator during SPIR-V consumption. Just imagine being a user, who compiled once a valid OpenCL application in JIT mode one day with drivers update find out, that the application crashes during execution (I believe, that is what happening with the test now).
So, my suggestion is:
for LLVM IR to SPIR-V translation have
```
compare_exchange_strong_explicit -> AtomicCompareExchange
compare_exchange_weak_explicit -> AtomicCompareExchange
```

For reverse translation:
```
AtomicCompareExchange ->compare_exchange_strong_explicit
AtomicCompareExchangeWeak -> compare_exchange_weak_explicit
```

jgstarIntel: I've read this #1230. The backward compatibility argument makes sense & broadens test coverage - so that's fine with me. I can update.

lib/SPIRV/OCLUtil.cpp
```diff
@@ -269,7 +269,9 @@ template <> void SPIRVMap<std::string, Op, SPIRVInstruction>::init() {
   _SPIRV_OP(store_explicit, AtomicStore)
   _SPIRV_OP(exchange_explicit, AtomicExchange)
   _SPIRV_OP(compare_exchange_strong_explicit, AtomicCompareExchange)
-  _SPIRV_OP(compare_exchange_weak_explicit, AtomicCompareExchangeWeak)
+  // OpAtomicCompareExchangeWeak is not "weak" at all, but instead has the same
```
asudarsa, line 272: Please also mention that OpAtomicCompareExchangeWeak has been deprecated.

lib/SPIRV/SPIRVWriter.cpp
```diff
@@ -5134,6 +5134,15 @@ Op LLVMToSPIRVBase::transBoolOpCode(SPIRVValue *Opn, Op OC) {
...
 LLVMToSPIRVBase::transBuiltinToInstWithoutDecoration(Op OC, CallInst *CI,
                                                      SPIRVBasicBlock *BB) {
+  if (BM->isAllowedToUseVersion(VersionNumber::SPIRV_1_4)) {
+    // OpAtomicCompareExchangeWeak is not "weak" at all,
+    // but instead has the same semantics as OpAtomicCompareExchange.
+    // Moreover, OpAtomicCompareExchangeWeak has been deprecated for
+    // SPIR-V version 1.4 or later.
+    if (OC == OpAtomicCompareExchangeWeak) {
+      OC = OpAtomicCompareExchange;
+    }
```
MrSidims, line 5144: nit: LLVM's coding guidelines suggest not to separate one-liners with brackets
```suggestion
    if (OC == OpAtomicCompareExchangeWeak)
      OC = OpAtomicCompareExchange;
```

MrSidims (approved):
LGTM

@StuartDBrady please take a look

lib/SPIRV/SPIRVWriter.cpp
jgstarIntel: updated

test/transcoding/atomics.spt
StuartDBrady: _AtomicCompareExchangeWeak_ must _not_ be translated to _compare_exchange_weak_explicit_.

It can _only_ be safely translated to _compare_exchange_strong_explicit._

MrSidims: @StuartDBrady yes, it was my misinterpretation. Current version of the patch doesn't follow it, see test/transcoding/atomics.spt test remained to be unchanged .

MrSidims: @StuartDBrady friendly ping. The patch doesn't add this mapping to compare_exchange_weak_explicit

MrSidims: @StuartDBrady I was talking about this test: https://github.com/KhronosGroup/SPIRV-LLVM-Translator/blob/main/test/transcoding/atomics.spt
As you may see it has both `AtomicCompareExchange` and `AtomicCompareExchangeWeak` in pre-generated SPIR-V and then on line [92 and 93 ](https://github.com/KhronosGroup/SPIRV-LLVM-Translator/blob/main/test/transcoding/atomics.spt#L92) we have a check for "strong" OpenCL call. So given that this patch doesn't modify this test, it seems like the behavior is correct.

lib/SPIRV/SPIRVWriter.cpp
```diff
@@ -5134,6 +5134,14 @@ Op LLVMToSPIRVBase::transBoolOpCode(SPIRVValue *Opn, Op OC) {
 SPIRVInstruction *
 LLVMToSPIRVBase::transBuiltinToInstWithoutDecoration(Op OC, CallInst *CI,
                                                      SPIRVBasicBlock *BB) {
+  if (BM->isAllowedToUseVersion(VersionNumber::SPIRV_1_4)) {
```
StuartDBrady, line 5137: Why is this conditional?  We should _never_ generate OpAtomicCompareExchangeWeak for _any_ SPIR-V version.

(Can we map the OpenCL C atomic_compare_exchange_weak builtin to OpAtomicCompareExchange before we even get here, and remove this code?)

jgstarIntel: @MrSidims @StuartDBrady. I've updated to remove the conditional so it's always the case the OpAtomicCompareExchange is used for any SPIR-V version. Please let me know if this is enough to resolve the issue.

jgstarIntel: Regarding a mapping, an earlier patch used that approach & had a range of issues pop-up (e.g. failing transcoding testing - atomics.spt) where atomic_compare_exchange_strong_explicit(..) would be generated with unexpected input parameter types and so that approach might require deeper investigation of that issue.

asudarsa (approved):
LGTM. Thanks

lib/SPIRV/SPIRVWriter.cpp
MrSidims: @StuartDBrady should be addressed by this point

svenvh (approved):
LGTM; approving on behalf of @StuartDBrady too.

PR #1834, opened 2023-02-07
Add JointMatrixGetElementCoordINTEL instruction

MrSidims:
The instruction returns (Row, Column) coordinate of dynamically selected element of a matrix

Updated version of the spec is here

Instruction correctness checks will be added later among non-backward compatible changes.

### DIFF
--- a/lib/SPIRV/libSPIRV/SPIRVInstruction.h
+++ b/lib/SPIRV/libSPIRV/SPIRVInstruction.h
@@ -3324,7 +3324,22 @@ class SPIRVJointMatrixINTELInst : public SPIRVJointMatrixINTELInstBase {
 _SPIRV_OP(JointMatrixLoad, true, 6, true)
 _SPIRV_OP(JointMatrixStore, false, 5, true)
 _SPIRV_OP(JointMatrixMad, true, 7)
+// TODO: move to SPIRVJointMatrixINTELWorkItemInst
 _SPIRV_OP(JointMatrixWorkItemLength, true, 4)
+#undef _SPIRV_OP
+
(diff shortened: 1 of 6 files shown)


MrSidims:
@asudarsa @vmaksimo please take a look

lib/SPIRV/libSPIRV/SPIRVInstruction.h
```diff
@@ -3324,9 +3324,24 @@ class SPIRVJointMatrixINTELInst : public SPIRVJointMatrixINTELInstBase {
 _SPIRV_OP(JointMatrixLoad, true, 6, true)
 _SPIRV_OP(JointMatrixStore, false, 5, true)
 _SPIRV_OP(JointMatrixMad, true, 7)
+// TODO: move to SPIRVJointMatrixINTELWorkItemInst
```
vmaksimo, line 3327: Why can't we move this right away in this PR? To preserve backward compatibility by not requiring support of `JointMatrixWIInstructions` capability for now?

MrSidims: Usage of `OpJointMatrixWorkItemLength` will currently emit just `CapabilityJointMatrixINTEL`. Which is supported in our downstream. Given the fact, that it's SYCL's counterpart is already implemented and used in tests and user code making the SPIR-V instruction be bound with `CapabilityJointMatrixWIInstructionsINTEL` would make these tests/user code not working, since this capability is not yet introduced in our downstream.

With this patch I'm introducing it, and then, when our drivers adopt it - the TODO can be resolved.

vmaksimo (approved):
I guess we can proceed with the specification inconsistency if it'd be addressed soon, and if there is no other simple way to introduce backward-incompatible changes.

PR #1835, opened 2023-02-07
Add ComponentTypeInterpretation for joint matrix type

MrSidims:
It specifies how to interpret 'Component Type' when components of a joint matrix are storages for values of different types, for example float for TF32, unsigned short for bfloat16.

At this point only tf32 type interpretation is added during SPIR-V generation. Adding it to bf16 is a breaking change and
requires adaptation across drivers.

Spec update:

### DIFF
--- a/lib/SPIRV/SPIRVWriter.cpp
+++ b/lib/SPIRV/SPIRVWriter.cpp
@@ -610,7 +610,7 @@ SPIRVType *LLVMToSPIRVBase::transPointerType(SPIRVType *ET, unsigned AddrSpc) {
 
 // Representation in LLVM IR before the translator is a pointer to an opaque
 // structure:
-// %spirv.JointMatrixINTEL._%element_type%_%rows%_%cols%_%scope%_%use%
+// %spirv.JointMatrixINTEL._%element_type%_%rows%_%cols%_%layout%_%scope%_%use%
 // Here we check the structure name yet again. Another option would be to
 // check SPIR-V friendly function calls (by their name) and obtain return
 // or their parameter types, assuming, that the appropriate types are Matrix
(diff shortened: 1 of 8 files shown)


MrSidims:
@asudarsa @vmaksimo @jcranmer-intel  please take a look

Most of it should be re-written during switch to opaque pointers (see https://github.com/intel/llvm/pull/8343/commits/7707bb0e7c2cde05e7c27026d6cc95d27e5f00e5 ), but the switch has several dependencies still, so to unblock BE development I'd prefer to merge it as is and to create opaque pointers adjustments on top of this patch.

MrSidims:
Actually, it's a breaking change already. Need to think how to avoid it. So merge together with opaque pointers should be the way to go.

MrSidims:
@asudarsa @vmaksimo @jcranmer-intel please take a look

lib/SPIRV/SPIRVWriter.cpp
```diff
@@ -621,6 +621,20 @@ SPIRVType *LLVMToSPIRVBase::transPointerType(SPIRVType *ET, unsigned AddrSpc) {
 // simply not true.
 SPIRVType *LLVMToSPIRVBase::transSPIRVJointMatrixINTELType(
     SmallVector<std::string, 8> Postfixes) {
+  auto ParseInteger = [this](StringRef Postfix) -> ConstantInt * {
+    unsigned long long N = 0;
+    if (consumeUnsignedInteger(Postfix, 10, N)) {
+      BM->getErrorLog().checkError(
```
vmaksimo, line 627: Can't comment the description above the function.
I guess we need to update the LLVM IR representation description as there are less arguments than in the added test:
```
// Representation in LLVM IR before the translator is a pointer to an opaque
// structure:
// %spirv.JointMatrixINTEL._%element_type%_%rows%_%cols%_%scope%_%use%
```

lib/SPIRV/SPIRVWriter.cpp
```diff
@@ -621,6 +621,20 @@ SPIRVType *LLVMToSPIRVBase::transPointerType(SPIRVType *ET, unsigned AddrSpc) {
...
+          false, SPIRVEC_InvalidLlvmModule,
+          "TypeJointMatrixINTEL expects integer parameters");
+      return 0;
```
vmaksimo, line 630: My understanding is that `checkError()` exits the program when error occured, so no return is needed here. Guess it's also correct for lamdas

lib/SPIRV/SPIRVWriter.cpp
MrSidims: The idea was that %layout% will be removed (bot now, since it's a breaking change). But I agree - let's return it here and update the description once we are ready.

lib/SPIRV/SPIRVWriter.cpp
MrSidims: Applied, thanks!

MrSidims:
@asudarsa @vmaksimo friendly reminder to take a look

MrSidims:
@asudarsa @vmaksimo please review the patch.

vmaksimo (approved):

PR #1697, opened 2022-11-07
Translate readnone attribute as function parameter attribute

MrSidims:
Community restricted readnone, readonly and writeonly attributes to be only function parameter attributes (commit 304f1d59ca4).
This patch aligns the translator with llvm.org.

It also fixes a bug, when readnone attribute is being mapped to NoWrite SPIR-V function parameter attribute.

Signed-off-by: Sidorov, Dmitry <dmitry.sidorov@intel.com>

PR #1836, opened 2023-02-07
[Backport to 15] Translate readnone attribute as function parameter attribute (#1697)

stanleygambarin:
Community restricted readnone, readonly and writeonly attributes
to be only function parameter attributes. This patch aligns
the translator with llvm.org.

It also fixes a bug, when readnone attribute is being mapped
to NoWrite SPIR-V function parameter attribute.

### DIFF
--- a/lib/SPIRV/SPIRVInternal.h
+++ b/lib/SPIRV/SPIRVInternal.h
@@ -240,6 +240,7 @@ inline void SPIRVMap<Attribute::AttrKind, SPIRVFuncParamAttrKind>::init() {
   add(Attribute::NoAlias, FunctionParameterAttributeNoAlias);
   add(Attribute::NoCapture, FunctionParameterAttributeNoCapture);
   add(Attribute::ReadOnly, FunctionParameterAttributeNoWrite);
+  add(Attribute::ReadNone, FunctionParameterAttributeNoReadWrite);
 }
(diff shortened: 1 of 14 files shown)


svenvh:
> Community restricted readnone, readonly and writeonly attributes
> to be only function parameter attributes. This patch aligns
> the translator with llvm.org.

This restriction only applies to LLVM 16 I believe, so not sure if this patch should be cherry-picked in its entirety to older branches?

stanleygambarin:
@MrSidims: i will follow up with b95227e980fd632e478c1faa2fc1544d42b4c355 cherry-pick (https://github.com/KhronosGroup/SPIRV-LLVM-Translator/pull/1746), but can you elaborate on why SPIRVFunctionControlMaskKind should not be changed (and which older LLVM branches) ?

MrSidims:
> @MrSidims: i will follow up with [b95227e](https://github.com/KhronosGroup/SPIRV-LLVM-Translator/commit/b95227e980fd632e478c1faa2fc1544d42b4c355) cherry-pick (#1746), but can you elaborate on why SPIRVFunctionControlMaskKind should not be changed (and which older LLVM branches) ?

Because in LLVM 15 and earlies it's not restricted to have these 2 attributes as function attributes. So the mapping should remain.

MrSidims (approved):
LGTM thanks! Will merge after @svenvh approves

svenvh (approved):

PR #1846, opened 2023-02-13
Initial support NonSemantic.Kernel.DebugInfo.100

vmaksimo:
This patch implements the initial support for the new debug specification NonSemantic.Kernel.DebugInfo.100.
It also introduces support for the new debug instruction DISubrange.

Spec: https://github.com/KhronosGroup/SPIRV-Registry/pull/186


MrSidims:
@svenvh may I ask you to take a look if you (ARM) would be interested in new debug info extended instruction set?

test/DebugInfo/DebugInfoSubrange.ll
```diff
@@ -7,38 +7,51 @@
...
-; CHECK-LLVM: [[#SubrangeExpr1]] = !{![[#SubrangeExpr2:]]}
-; CHECK-LLVM: ![[#SubrangeExpr2]] = !DISubrange(lowerBound: !DIExpression(), upperBound: !DIExpression())
-
-; CHECK-LLVM: !DISubrange(count: 1000, lowerBound: 1)
-
-; CHECK-LLVM: !DICompositeType(tag: DW_TAG_array_type, baseType: ![[#BaseType:]], elements: ![[#Subrage:]])
-; CHECK-LLVM: ![[#BaseType]] = !DIBasicType(name: "int", size: 32, encoding: DW_ATE_signed)
-; CHECK-LLVM: ![[#Subrage]] = !{![[#Subrage:]]}
-; CHECK-LLVM: ![[#Subrage]] = !DISubrange(count: -1
+; According to the OpenCL.DebugInfo.100 specification only Count should be
```
MrSidims, line 10: I suggest to outline old version of this test into new DebugInfo/OpenCL100/ directory adding a new test in DebugInfo/NonSemanticKernel100/ (one can also imagine a better naming). Same applicable to all tests that will be added in a scope of this feature.

Also I kind of expecting existing workarounds to translate some semantics in a scope of old debug info spec. I suggest to preserve these workarounds for OpenCL.100 debug info. It means, that 'old' version of this tests should remain unchanged.

lib/SPIRV/LLVMToSPIRVDbgTran.cpp
```diff
@@ -277,6 +277,9 @@ SPIRVEntry *LLVMToSPIRVDbgTran::transDbgEntryImpl(const MDNode *MDN) {
     case dwarf::DW_TAG_array_type:
       return transDbgArrayType(cast<DICompositeType>(DIEntry));
 
+    case dwarf::DW_TAG_subrange_type:
+      return transDbgSubrangeType(cast<DISubrange>(DIEntry));
```
MrSidims, line 281: Suggest to add a check for NonSemantic_Kernel_DebugInfo_100 and don't translate the dwarf type if the instruction set is not enabled

MrSidims (commented):
Follow up to the comment to the test:
I see some W/As to preserve subranges in existing `DebugTypeArray` translation. I suggest not to remove these W/As: instead we may have (temporary) two version of translate DebugTypeArray functions:
- 1st should be the same implementation as we have now without the patch, and they should be run if OpenCL debug info instruction set is enabled;
- 2nd should be the changed functions as they are being added in this patch, and they should be run if nonsemantic debug info instruction set is enabled

lib/SPIRV/LLVMToSPIRVDbgTran.cpp
```diff
@@ -594,6 +608,92 @@ SPIRVEntry *LLVMToSPIRVDbgTran::transDbgArrayType(const DICompositeType *AT) {
...
+      // Constant or DIVariable. No other operands must be preserved.
+      if (Count)
+        Ops[ComponentCountIdx + I] =
+            SPIRVWriter->transValue(Count, nullptr)->getId();
+      else if (auto *CountNode =
+                   dyn_cast_or_null<MDNode>(SR->getRawCountNode()))
+        Ops[ComponentCountIdx + I] = transDbgEntry(CountNode)->getId();
+      else
+        Ops[ComponentCountIdx + I] = getDebugInfoNoneId();
+    }
```
MrSidims, line 643: This if - else is redundant, we already did it in `transDbgArrayType`

lib/SPIRV/SPIRVToLLVMDbgTran.cpp
```diff
@@ -246,6 +255,53 @@ SPIRVToLLVMDbgTran::transTypeArray(const SPIRVExtInst *DebugInst) {
...
+DICompositeType *
+SPIRVToLLVMDbgTran::transTypeArrayNonSemantic(const SPIRVExtInst *DebugInst) {
+  using namespace SPIRVDebug::Operand::TypeArray;
+  const SPIRVWordVec &Ops = DebugInst->getArguments();
+  assert(Ops.size() >= MinOperandCount && "Invalid number of operands");
+  DIType *BaseTy =
+      transDebugInst<DIType>(BM->get<SPIRVExtInst>(Ops[BaseTypeIdx]));
+  size_t TotalCount = 1;
+  SmallVector<llvm::Metadata *, 8> Subscripts;
+  if (DebugInst->isExtInst(SPIRVEIS_NonSemantic_Kernel_DebugInfo_100,
```
MrSidims, line 267: Same

test/DebugInfo/NonSemanticKernel100/DebugInfoSubrange.ll
```diff
@@ -0,0 +1,133 @@
+; According to the OpenCL.DebugInfo.100 specification only Count should be
```
MrSidims, line 1: The comment is outdated

test/FortranArray.ll
```diff
@@ -10,7 +10,13 @@
 ; and back to .ll. This causes the LLVM IR verifier to fail as there
 ; are different rules for valid DISubRange depending on language ID.
 
+; RUN: llvm-spirv %t.bc -spirv-text --spirv-debug-info-version=nonsemantic-kernel-100
```
MrSidims, line 13: Lets don't change this test in _this_ patch

MrSidims (commented):
Overall LGTM, several minor comments

test/FortranArray.ll
vmaksimo: Why not? We just add the run/check lines for new behavior for DISubrange. Or do we plan to preserve _only one_ version of DISubrange translation?

MrSidims: The problem here is with the Fortran language. As we discussed offline, the lower bound can start from either 0 or 1 depending on the language. So for the older spec, we won't have Fortran language, which means that after the translation, we will have OpenCL encoded in dwarf. In this test, the lower bound seems to be set as 1, causing the LLVM verifier to error out.

MrSidims (approved):
LGTM

MrSidims:
Merging do unblock further development. Name of the instruction set and option can be changed.

PR #1851, opened 2023-02-17
Fix FortranArray test

vmaksimo:
The test was originally XFAIL-ed in 9e234d9
The problem described in the comment is fixed by adding Fortran95 language.

Another problem was that DISubrange belongs to extended instruction set. Added the appropriate key to the RUN lines to make test passed.


MrSidims:
Closing as no longer needed

PR #1853, opened 2023-02-22
[DebugInfo] Add Target Function optional parameter to DebugFunction

MrSidims:
It's being added in
https://github.com/KhronosGroup/SPIRV-Registry/pull/186

In DWARF it's used in 'trampoline' functions generated for Fortran external function calls.

### DIFF
--- a/lib/SPIRV/LLVMToSPIRVDbgTran.cpp
+++ b/lib/SPIRV/LLVMToSPIRVDbgTran.cpp
@@ -988,8 +988,18 @@ SPIRVEntry *LLVMToSPIRVDbgTran::transDbgFunction(const DISubprogram *Func) {
 
     if (DISubprogram *FuncDecl = Func->getDeclaration())
       Ops.push_back(transDbgEntry(FuncDecl)->getId());
-    else
+    else {
       Ops.push_back(getDebugInfoNoneId());
+      if (BM->getDebugInfoEIS() == SPIRVEIS_NonSemantic_Kernel_DebugInfo_100) {
+        // Translate targetFuncName mostly for Fortran trampoline function if it
+        // is the case
+        StringRef TargetFunc = Func->getTargetFuncName();
+        if (!TargetFunc.empty()) {
+          Ops.resize(MinOperandCount + 2);
+          Ops[TargetFunctionNameIdx] = BM->getString(TargetFunc.str())->getId();
+        }
+      }
+    }
 
     DebugFunc = BM->addDebugInfo(SPIRVDebug::Function, getVoidTy(), Ops);
     MDMap.insert(std::make_pair(Func, DebugFunc));
--- a/lib/SPIRV/LLVMToSPIRVDbgTran.cpp
+++ b/lib/SPIRV/LLVMToSPIRVDbgTran.cpp
@@ -994,10 +994,8 @@ SPIRVEntry *LLVMToSPIRVDbgTran::transDbgFunction(const DISubprogram *Func) {
         // Translate targetFuncName mostly for Fortran trampoline function if it
         // is the case
         StringRef TargetFunc = Func->getTargetFuncName();
-        if (!TargetFunc.empty()) {
-          Ops.resize(MinOperandCount + 2);
-          Ops[TargetFunctionNameIdx] = BM->getString(TargetFunc.str())->getId();
-        }
+        if (!TargetFunc.empty())
+          Ops.push_back(BM->getString(TargetFunc.str())->getId());
       }
     }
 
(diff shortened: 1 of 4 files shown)


MrSidims:
@vmaksimo please take a look

lib/SPIRV/SPIRVToLLVMDbgTran.cpp
```diff
@@ -608,9 +608,17 @@ DINode *SPIRVToLLVMDbgTran::transFunction(const SPIRVExtInst *DebugInst) {
...
+    // Create targetFuncName mostly for Fortran trampoline function if it is
+    // the case
+    StringRef TargetFunction;
+    if (Ops.size() > TargetFunctionNameIdx) {
+      TargetFunction = getString(Ops[TargetFunctionNameIdx]);
+    }
     DIS = Builder.createFunction(Scope, Name, LinkageName, File, LineNo, Ty,
-                                 ScopeLine, Flags, SPFlags, TParamsArray, FD);
+                                 ScopeLine, Flags, SPFlags, TParamsArray, FD,
+                                 nullptr, nullptr, TargetFunction);
```
vmaksimo, line 620: Maybe would be nice to left comments about what `nullptr` means

lib/SPIRV/LLVMToSPIRVDbgTran.cpp
```diff
@@ -988,8 +988,18 @@ SPIRVEntry *LLVMToSPIRVDbgTran::transDbgFunction(const DISubprogram *Func) {
...
-    else
+    else {
       Ops.push_back(getDebugInfoNoneId());
+      if (BM->getDebugInfoEIS() == SPIRVEIS_NonSemantic_Kernel_DebugInfo_100) {
+        // Translate targetFuncName mostly for Fortran trampoline function if it
+        // is the case
+        StringRef TargetFunc = Func->getTargetFuncName();
+        if (!TargetFunc.empty()) {
+          Ops.resize(MinOperandCount + 2);
+          Ops[TargetFunctionNameIdx] = BM->getString(TargetFunc.str())->getId();
```
vmaksimo, line 999: Why not to use just `push_back` here? Not obvious why the `resize` is needed here

MrSidims: `push_back` can be used here indeed. I was considering it and logic to prefer resize was the following:
We need to add `TargetFunctionNameIdx` in `lib/SPIRV/libSPIRV/SPIRV.debug.h` anyway to reuse it during reverse translation. So for the consistency with the code around it make sense to reuse it in forward translation as well. Also `resize` was also used on line 976 as well in similar example.

lib/SPIRV/SPIRVToLLVMDbgTran.cpp
MrSidims: Sure, I'll add it!

MrSidims: Added, but please don't ask me, what these parameter mean :D

lib/SPIRV/LLVMToSPIRVDbgTran.cpp
vmaksimo: Sorry, i'm not following it fully - seems that resize to `MinOperandCount+1` or `TargetFunctionNameIdx` would be enough.

Also, I found that we didn't define `Ops[DeclarationIdx]` here, but the spec says:

> When Target function present and is not an <id> of [DebugInfoNone](https://github.com/KhronosGroup/SPIRV-Registry/blob/4fab24de6c61ae302fe5c3ba900fcddeae349a4a/nonsemantic/NonSemantic.Kernel.DebugInfo.100.asciidoc#DebugInfoNone) - Declaration must be an <id> of [DebugInfoNone](https://github.com/KhronosGroup/SPIRV-Registry/blob/4fab24de6c61ae302fe5c3ba900fcddeae349a4a/nonsemantic/NonSemantic.Kernel.DebugInfo.100.asciidoc#DebugInfoNone).

Am I right that we should fill this Declaration Id any time we translate Target Id?

MrSidims: >> Sorry, i'm not following it fully - seems that resize to MinOperandCount+1 or TargetFunctionNameIdx would be enough.

Nope, because there is actually one `push_back` earlier in the code :) See line 992 `Ops.push_back(getDebugInfoNoneId());` , it address both of the questions.

MrSidims: replaced resize with push_back

vmaksimo (approved):

PR #1854, opened 2023-02-22
[DebugInfo] Add new Source Languages table used in DebugCompilationUnit 

vmaksimo:
This extended source language table is used by DebugCompilationUnit instruction when the extension is enabled. It enables support for more languages than exists in a core specification.
Spec: https://github.com/KhronosGroup/SPIRV-Registry/pull/186/

Enabling of Fortran language also allowed to fix FortranArray test that was originally XFAIL-ed in 9e234d9.

### DIFF
--- a/lib/SPIRV/libSPIRV/SPIRV.debug.h
+++ b/lib/SPIRV/libSPIRV/SPIRV.debug.h
@@ -2,6 +2,7 @@
 #define SPIRV_DEBUG_H
 #include "SPIRVUtil.h"
 #include "spirv/unified1/spirv.hpp"
+#include "spirv_internal.hpp"
 #include "llvm/BinaryFormat/Dwarf.h"
 
 namespace SPIRVDebug {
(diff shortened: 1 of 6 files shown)


lib/SPIRV/libSPIRV/SPIRV.debug.h
```diff
@@ -809,9 +810,7 @@ inline spv::SourceLanguage convertDWARFSourceLangToSPIRV(dwarf::SourceLanguage D
   switch (DwarfLang) {
   // When updating this function, make sure to also
   // update convertSPIRVSourceLangToDWARF()
-
-  // LLVM does not yet define DW_LANG_C_plus_plus_17
-  // case dwarf::SourceLanguage::DW_LANG_C_plus_plus_17:
+  case dwarf::SourceLanguage::DW_LANG_C_plus_plus_17:
   case dwarf::SourceLanguage::DW_LANG_C_plus_plus_14:
   case dwarf::SourceLanguage::DW_LANG_C_plus_plus:
     return spv::SourceLanguage::SourceLanguageCPP_for_OpenCL;
```
MrSidims, line 816: @svenvh do I understand correctly, that mapping of C++ to OpenCL here was done as a workaround in the first place?
For the context, result of this function is used to create `DebugCompilationUnit` and not for `OpSource`.

MrSidims (commented):
The patch is LGTM
I'd suggest to also find an existing test for `DebugCompilationUnit`, that has a mapping of C++ to OpenCL C++ and create a new debug info test as a copy of the one found adding a check, that it's no longer OpenCL C++, but C++ itself.

MrSidims:
I have rephrased the header a bit, since probably the instruction set will be renamed to something like NonSemantic.Shader.DebugInfo.200 and hence if it happens the commit header will be confusing for developers to come

MrSidims:
Out-of-tree build failures are expected due to
```
commit efae3174f09560353fb0f3d528bcbffe060d5438
Author: Petr Hosek <phosek@google.com>
Date:   Wed Feb 1 08:25:35 2023 +0000

    [CMake] Unify llvm_check_linker_flag and llvm_check_compiler_linker_flag
```

which were later reverted in the community

MrSidims (approved):
LGTM

lib/SPIRV/libSPIRV/SPIRV.debug.h
svenvh: Yes I think so.

PR #1857, opened 2023-02-24
[OpaquePointers] Fix translation of EnqueueKernel and block functions

dwoodwor-intel:
This specifies the right parameter types for EnqueueKernel and for block
functions (with mangled names of the form {__*,}_block_invoke{_*,}{_kernel,}),
so that SPIR-V code is generated for these as expected. This fixes a
crash in the Intel OpenCL CPU backend, and likely other SPIR-V
consumers.

### DIFF
--- a/lib/SPIRV/SPIRVTypeScavenger.cpp
+++ b/lib/SPIRV/SPIRVTypeScavenger.cpp
@@ -233,7 +233,8 @@ void SPIRVTypeScavenger::deduceFunctionType(Function &F) {
 
   // The first non-sret argument of block_invoke functions is the block capture
   // struct, which should be passed as an i8*.
-  static Regex BlockInvokeRegex("^(__.+)?_block_invoke(_[0-9]+)?(_kernel)?$");
+  static const Regex BlockInvokeRegex(
+      "^(__.+)?_block_invoke(_[0-9]+)?(_kernel)?$");
   if (BlockInvokeRegex.match(F.getName())) {
     for (Argument *Arg : PointerArgs) {
       if (!Arg->hasAttribute(Attribute::StructRet)) {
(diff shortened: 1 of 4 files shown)


dwoodwor-intel:
This patch is ready for review, but depends on #1799 (which also depends on [D141008](https://reviews.llvm.org/D141008)). I am seeing some test failures which don't seem to be specific to this change; they probably need to be fixed in #1799 when that is updated to be merged:

```
Failed Tests (2):
  LLVM_SPIRV :: extensions/INTEL/SPV_INTEL_device_side_avc_motion_esimation/subgroup_avc_intel_vme_image.cl
  LLVM_SPIRV :: image-unoptimized.cl

********************
Unexpectedly Passed Tests (1):
  LLVM_SPIRV :: transcoding/pipe_builtins.cl
```

MrSidims (commented):
LGTM, thanks!

dwoodwor-intel:
I've rebased this on the current main branch now that the relevant fixes (and the test fix yesterday for #1885) are there. @jcranmer-intel also pointed out a few more block-related tests that need to be updated for opaque pointers, so I've updated those here too and fixed some bugs in my first implementation of the block part of this.

lib/SPIRV/SPIRVTypeScavenger.cpp
```diff
@@ -226,6 +231,20 @@ void SPIRVTypeScavenger::deduceFunctionType(Function &F) {
       DeducedTypes[Arg] = Ty;
   }
 
+  // The first non-sret argument of block_invoke functions is the block capture
+  // struct, which should be passed as an i8*.
+  static Regex BlockInvokeRegex("^(__.+)?_block_invoke(_[0-9]+)?(_kernel)?$");
```
MrSidims, line 236: To clarify, why do we need it be static?

svenvh (approved):

lib/SPIRV/SPIRVTypeScavenger.cpp
dwoodwor-intel: This doesn't need to be static, but it helps if it is static because re-building the regex each time the function is called can be wasteful.

It looks like I missed marking this const though; I've added a small update for that.

MrSidims (approved):

PR #1863, opened 2023-03-01
Fix static analyzer issues

asudarsa:
This change fixes the following issues:
1. Unchecked return value
2. Explicit NULL dereferenced

Thanks

### DIFF
--- a/lib/SPIRV/SPIRVUtil.cpp
+++ b/lib/SPIRV/SPIRVUtil.cpp
@@ -481,7 +481,10 @@ bool oclIsBuiltin(StringRef Name, StringRef &DemangledName, bool IsCpp) {
   } else {
     size_t Start = Name.find_first_not_of("0123456789", 2);
     size_t Len = 0;
-    Name.substr(2, Start - 2).getAsInteger(10, Len);
+    if (Name.substr(2, Start - 2).getAsInteger(10, Len)) {
+      SPIRVDBG(errs() << "Error in extracting integer value");
+      return false;
+    }
     DemangledName = Name.substr(Start, Len);
   }
   return true;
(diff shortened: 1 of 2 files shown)


lib/SPIRV/libSPIRV/SPIRVModule.cpp
```diff
@@ -547,7 +547,10 @@ SPIRVModuleImpl::getCurrentLine() const {
 
 void SPIRVModuleImpl::setCurrentLine(
     const std::shared_ptr<const SPIRVLine> &Line) {
-  CurrentLine = Line;
+  if (Line)
```
MrSidims, line 550: I assume this relates to a code analyzer report for 'explicit null dereference'. If so, can you please elaborate on how this could happen? Also, can the std::shared_ptr constructor even return null?

lib/SPIRV/SPIRVUtil.cpp
```diff
@@ -481,7 +481,10 @@ bool oclIsBuiltin(StringRef Name, StringRef &DemangledName, bool IsCpp) {
   } else {
     size_t Start = Name.find_first_not_of("0123456789", 2);
     size_t Len = 0;
-    Name.substr(2, Start - 2).getAsInteger(10, Len);
+    if (Name.substr(2, Start - 2).getAsInteger(10, Len)) {
+      SPIRVDBG(errs() << "Error in extracting integer value");
+      return false;
+    }
     DemangledName = Name.substr(Start, Len);
```
MrSidims, line 488: This is a micro-optimization that avoids an extra jump in scenarios where the condition is always true.
```suggestion
    if (Name.substr(2, Start - 2).getAsInteger(10, Len)) {
        DemangledName = Name.substr(Start, Len);
        return true;

    }
    SPIRVDBG(errs() << "Error in extracting integer value");
    return false;
```

PR #1864, opened 2023-03-01
[NFC] Fix static analyzer issues

asudarsa:
This change fixes the following issue reported by static analyzer:

Unchecked return value
Thanks


MrSidims (approved):
LGTM

PR #1865, opened 2023-03-01
Fix static analyzer issues

asudarsa:
This change fixes the following issues:

Unchecked return value
Explicit NULL dereferenced
Thanks


lib/SPIRV/libSPIRV/SPIRVModule.cpp
```diff
@@ -547,7 +547,10 @@ SPIRVModuleImpl::getCurrentLine() const {
 
 void SPIRVModuleImpl::setCurrentLine(
     const std::shared_ptr<const SPIRVLine> &Line) {
-  CurrentLine = Line;
+  if (Line)
```
MrSidims, line 550: I assume this relates to a code analyzer report for 'explicit null dereference'. If so, can you please elaborate on how this could happen? Also, can the std::shared_ptr constructor even return null?

lib/SPIRV/SPIRVUtil.cpp
```diff
@@ -481,7 +481,10 @@ bool oclIsBuiltin(StringRef Name, StringRef &DemangledName, bool IsCpp) {
   } else {
     size_t Start = Name.find_first_not_of("0123456789", 2);
     size_t Len = 0;
-    Name.substr(2, Start - 2).getAsInteger(10, Len);
+    if (Name.substr(2, Start - 2).getAsInteger(10, Len)) {
+      SPIRVDBG(errs() << "Error in extracting integer value");
+      return false;
+    }
```
MrSidims, line 487: This is a micro-optimization that avoids an extra jump in scenarios where the condition is always true.
```suggestion
    if (!Name.substr(2, Start - 2).getAsInteger(10, Len)) {
        DemangledName = Name.substr(Start, Len);
        return true;
    }
    SPIRVDBG(errs() << "Error in extracting integer value");
    return false;
```
Additionally, is not using the return value from a function a real issue? According to the specifications of C and C++, it is allowed. From a logical standpoint, what problems could arise if we fall through this check

asudarsa:
@MrSidims 

here is your earlier comment:
I assume this relates to a code analyzer report for 'explicit null dereference'. If so, can you please elaborate on how this could happen? Also, can the std::shared_ptr constructor even return null?

Here, we have a scenario where we are trying to set null value to a shared_ptr. Code analyzer complains that assignment statement with nullptr on the RHS causes this issue. I am trying to avoid this issue by emptying the shared_ptr.

lib/SPIRV/SPIRVUtil.cpp
asudarsa: If we fall through, we return 'false' for isOCLBuiltin check and I think the caller will handle this correctly.

Thanks

asudarsa: Also, about the optimization, I was just trying to keep it similar to what is being done a few lines above. I do not have any issues either ways. Can you please take a look at the code above this change and let me know what you think?

Thanks

MrSidims:
> @MrSidims
> 
> here is your earlier comment: I assume this relates to a code analyzer report for 'explicit null dereference'. If so, can you please elaborate on how this could happen? Also, can the std::shared_ptr constructor even return null?
> 
> Here, we have a scenario where we are trying to set null value to a shared_ptr. Code analyzer complains that assignment statement with nullptr on the RHS causes this issue. I am trying to avoid this issue by emptying the shared_ptr.

Got it, thanks! So the issue appears in `getEntry` function in the following code
```
if (Entry->isEndOfBlock() || OpCode == OpNoLine)
M.setCurrentLine(nullptr);
```
Now I see, why the analyzer complains about it. But if we pass this check: `(Entry->isEndOfBlock() || OpCode == OpNoLine)` wouldn't it mean, that `use_count` should be zero and if it's not zero, then it's a bug? Aka shouldn't we instead of emptying ownership of the shared pointer via reset add an assertion that there are no other shared_ptr instances that still hold ownership of the object that CurrentLine points to? Still explicit null assignment would be OK if there are no owners left.

lib/SPIRV/SPIRVUtil.cpp
MrSidims: I just suggest not to worsen already not great performance in favor of satisfaction of some static analyzer tool reporting not critical issues.

MrSidims: Yeah, I should probably be more consistent in my reviews, and for some reasons didn't complain in https://github.com/KhronosGroup/SPIRV-LLVM-Translator/pull/1778/

asudarsa:
> 

I think we can ignore the issues reported as 'explicit null dereferencing'. Thanks

lib/SPIRV/SPIRVUtil.cpp
```diff
@@ -472,19 +472,23 @@ bool oclIsBuiltin(StringRef Name, StringRef &DemangledName, bool IsCpp) {
...
     size_t Len = 0;
-    Name.substr(2, Start - 2).getAsInteger(10, Len);
-    DemangledName = Name.substr(Start, Len);
+    if (!Name.substr(2, Start - 2).getAsInteger(10, Len)) {
+      DemangledName = Name.substr(Start, Len);
+      return true;
+    }
+    SPIRVDBG(errs() << "Error in extracting integer value");
+    return false;
   }
```
MrSidims, line 491: (can't apply suggestion 'on deleted lines):
```
 size_t Start = Name.find_first_not_of("0123456789", 2);
 size_t Len = 0;
 if (!Name.substr(2, Start - 2).getAsInteger(10, Len)) {
   return true;
 }
 SPIRVDBG(errs() << "Error in extracting integer value");
 return false;
```

asudarsa: Addressed. Thanks

MrSidims (approved):

PR #1879, opened 2023-03-13
Add ability to preserve all function metadata and attributes from LLVM-IR

sarnex:
In Intel's SYCL compiler, we have a use case where we need to convert to SPIR-V and then back to LLVM-IR and preserve all function metadata and attributes. The metadata and attributes we need to preserve have no value outside an internal stage of Intel's SYCL compiler and are not good fits for SPIR-V extensions. The list of what is needed is very volatile as well.

To implement generic support in the translator, the general approach is to use `SPV_KHR_non_semantic_info` to add a new nonsemantic EIS, named `NonSemantic.AuxData`

This new instruction set has two instructions:
`NonSemanticAuxDataFunctionMetadata`
and
`NonSemanticAuxDataFunctionAttribute`

Both instructions will be placed outside of the function CFG and take in what function to target as an operand, similar to debuginfo. We do this to support function declarations.

The operands after the function to target are either strings or values describing the metadata or attribute. Specific information can be found in the SPIRVWriter.cpp changes.

I also added a new option, `--spirv-preserve-auxdata` to preserve these.

If a flag is passed for translation to SPIR-V the respective information will be preserved as described above. The flag is also required for reverse translation. If it not provided, it will be dropped as allowed in the spec.

In order to handle the case of some metadata/attributes being handled manually (maybe we have metadata "foo=bar" in LLVM-IR, but we translate to "foo=oof" for some reason in SPIR-V (and thus reverse translation)), we process these new instructions very late in reverse translation, and if the metadata or attribute already exists, we skip it.

As future work, I plan to make passing either of the new flags implicitly enable SPV_KHR_non_semantic_info.

I plan to discuss this work in an upcoming OpenCL tooling meeting. 

Thanks to @MrSidims for help with the design and early initial review of this PR.

### DIFF
--- a/lib/SPIRV/SPIRVReader.cpp
+++ b/lib/SPIRV/SPIRVReader.cpp
@@ -3334,9 +3334,8 @@ bool SPIRVToLLVM::translate() {
   DbgTran->addDbgInfoVersion();
   DbgTran->finalize();
 
-  for (SPIRVExtInst *EI :
-       BM->getPreservedFunctionMetadataAndAttributesInstVec()) {
-    transIntelPreserveInst(EI);
+  for (SPIRVExtInst *EI : BM->getAuxDataInstVec()) {
+    transAuxDataInst(EI);
   }
 
   return true;
@@ -4468,8 +4467,8 @@ Instruction *SPIRVToLLVM::transOCLBuiltinFromExtInst(SPIRVExtInst *BC,
   return CI;
 }
 
-void SPIRVToLLVM::transIntelPreserveInst(SPIRVExtInst *BC) {
-  assert(BC->getExtSetKind() == SPIRV::SPIRVEIS_Intel_Preserve);
+void SPIRVToLLVM::transAuxDataInst(SPIRVExtInst *BC) {
+  assert(BC->getExtSetKind() == SPIRV::SPIRVEIS_Aux_Data);
   if (!BC->getModule()->preserveAllFunctionAttributesAndMetadata())
     return;
   auto Args = BC->getArguments();
@@ -4480,7 +4479,7 @@ void SPIRVToLLVM::transIntelPreserveInst(SPIRVExtInst *BC) {
   assert(F && "Function should already have been translated!");
   auto AttrOrMDName = BC->getModule()->get<SPIRVString>(Args[1])->getStr();
   switch (BC->getExtOp()) {
-  case NonSemanticIntelPreserve::FunctionAttribute: {
+  case NonSemanticAuxData::FunctionAttribute: {
     assert(Args.size() < 4 && "Unexpected FunctionAttribute Args");
     // If this attr was specially handled and added elsewhere, skip it.
     if (F->hasFnAttribute(AttrOrMDName))
@@ -4495,7 +4494,7 @@ void SPIRVToLLVM::transIntelPreserveInst(SPIRVExtInst *BC) {
     }
     break;
   }
-  case NonSemanticIntelPreserve::FunctionMetadata: {
+  case NonSemanticAuxData::FunctionMetadata: {
     // If this metadata was specially handled and added elsewhere, skip it.
     if (F->hasMetadata(AttrOrMDName))
       break;
--- a/lib/SPIRV/SPIRVReader.cpp
+++ b/lib/SPIRV/SPIRVReader.cpp
@@ -3333,6 +3333,12 @@ bool SPIRVToLLVM::translate() {
 
   DbgTran->addDbgInfoVersion();
   DbgTran->finalize();
+
+  for (SPIRVExtInst *EI :
+       BM->getPreservedFunctionMetadataAndAttributesInstVec()) {
+    transIntelPreserveInst(EI);
+  }
+
   return true;
 }
 
@@ -4460,6 +4466,60 @@ Instruction *SPIRVToLLVM::transOCLBuiltinFromExtInst(SPIRVExtInst *BC,
   setCallingConv(CI);
   addFnAttr(CI, Attribute::NoUnwind);
   return CI;
+}
+
+void SPIRVToLLVM::transIntelPreserveInst(SPIRVExtInst *BC) {
+  assert(BC->getExtSetKind() == SPIRV::SPIRVEIS_Intel_Preserve);
+  if (!BC->getModule()->preserveAllFunctionAttributesAndMetadata())
+    return;
+  auto Args = BC->getArguments();
+  // Args 0 and 1 are common between attributes and metadata.
+  // 0 is the function, 1 is the name of the attribute/metadata as a string
+  auto *SpvFcn = BC->getModule()->getValue(Args[0]);
+  auto *F = static_cast<Function *>(getTranslatedValue(SpvFcn));
+  assert(F && "Function should already have been translated!");
+  auto AttrOrMDName = BC->getModule()->get<SPIRVString>(Args[1])->getStr();
+  switch (BC->getExtOp()) {
+  case NonSemanticIntelPreserve::FunctionAttribute: {
+    assert(Args.size() < 4 && "Unexpected FunctionAttribute Args");
+    // If this attr was specially handled and added elsewhere, skip it.
+    if (F->hasFnAttribute(AttrOrMDName))
+      break;
+    // For attributes, arg 2 is the attribute value as a string, which may not
+    // exist.
+    if (Args.size() == 3) {
+      auto AttrValue = BC->getModule()->get<SPIRVString>(Args[2])->getStr();
+      F->addFnAttr(AttrOrMDName, AttrValue);
+    } else {
+      F->addFnAttr(AttrOrMDName);
+    }
+    break;
+  }
+  case NonSemanticIntelPreserve::FunctionMetadata: {
+    // If this metadata was specially handled and added elsewhere, skip it.
+    if (F->hasMetadata(AttrOrMDName))
+      break;
+    SmallVector<Metadata *> MetadataArgs;
+    // Process the metadata values.
+    for (size_t CurArg = 2; CurArg < Args.size(); CurArg++) {
+      auto *Arg = BC->getModule()->get<SPIRVEntry>(Args[CurArg]);
+      // For metadata, the metadata values can be either values or strings.
+      if (Arg->getOpCode() == OpString) {
+        auto *ArgAsStr = static_cast<SPIRVString *>(Arg);
+        MetadataArgs.push_back(
+            MDString::get(F->getContext(), ArgAsStr->getStr()));
+      } else {
+        auto *ArgAsVal = static_cast<SPIRVValue *>(Arg);
+        auto *TranslatedMD = transValue(ArgAsVal, F, nullptr);
+        MetadataArgs.push_back(ValueAsMetadata::get(TranslatedMD));
+      }
+    }
+    F->setMetadata(AttrOrMDName, MDNode::get(*Context, MetadataArgs));
+    break;
+  }
+  default:
+    llvm_unreachable("Invalid op");
+  }
 }
 
 // SPIR-V only contains language version. Use OpenCL language version as
--- a/lib/SPIRV/SPIRVReader.cpp
+++ b/lib/SPIRV/SPIRVReader.cpp
@@ -4468,8 +4468,8 @@ Instruction *SPIRVToLLVM::transOCLBuiltinFromExtInst(SPIRVExtInst *BC,
 }
 
 void SPIRVToLLVM::transAuxDataInst(SPIRVExtInst *BC) {
-  assert(BC->getExtSetKind() == SPIRV::SPIRVEIS_Aux_Data);
-  if (!BC->getModule()->preserveAllFunctionAttributesAndMetadata())
+  assert(BC->getExtSetKind() == SPIRV::SPIRVEIS_NonSemantic_AuxData);
+  if (!BC->getModule()->preserveAuxData())
     return;
   auto Args = BC->getArguments();
   // Args 0 and 1 are common between attributes and metadata.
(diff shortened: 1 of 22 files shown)


lib/SPIRV/libSPIRV/SPIRVInstruction.h
```diff
@@ -42,6 +42,7 @@
 
 #include "SPIRVBasicBlock.h"
 #include "SPIRVEnum.h"
+#include "SPIRVFunction.h"
```
sarnex, line 45: I don't know why my change caused clang tidy to fail, but we need this include because of https://github.com/KhronosGroup/SPIRV-LLVM-Translator/blob/11f2f38a829eb57f95ccb7558a8da5ab43c9462c/lib/SPIRV/libSPIRV/SPIRVInstruction.h#L1694 which clang tidy says should have defn of cast target

sarnex:
Feedback from Sven: Combine into a single option

I'll do this

svenvh:
If we were to generalize this into a vendor-agnostic extension, would that perhaps allow us to deprecate options such as `--spirv-gen-kernel-arg-name-md` and `--preserve-ocl-kernel-arg-type-metadata-through-string`?

svenvh:
> I still think we'd need the options, we could use this EIS to implement those, but with this proposed option we preserve everything for all fcns, not just kernel arg name/argtypes metadata, so we'd need a way to restrict what we preserve to that metadata and only for kernel fcns.  The only other path I see would be to add an option that allows the user to specify what functions to save stuff for (maybe you can choose everything vs kernels only)

Good point, we would indeed produce more metadata for all functions, which may not be what the user wants.  An option to select between all functions or only kernels seems reasonable.  Perhaps all of this is beyond this PR's scope, but it would be good to consider we may do so in the future.  With that in mind, would it be an option to make the EIS vendor-agnostic?

sarnex:
@svenvh Yeah sure I only made it Intel specific because I didn't expect outside interest. Can you and @MrSidims help me on a name?

Maybe `NonSemantic.Preserve`? I have no opinion.

svenvh:
> @svenvh Yeah sure I only made it Intel specific because I didn't expect outside interest. Can you and @MrSidims help me on a name?
> 
> Maybe `NonSemantic.Preserve`? I have no strong opinion.

Naming, that's probably the most difficult part! :)  Not sure if "Preserve" conveys the intent of the EIS well enough, so here are some more ideas:
- `NonSemantic.Annotations` (could be confused with `!annotation` metadata)
- `NonSemantic.CompilerAnnotations` (as above)
- `NonSemantic.Metadata` ("metadata" in the general non-LLVM-IR specific sense, though I admit it could be confusing as we also handle attributes)
- `NonSemantic.AuxData` or `NonSemantic.AuxiliaryData`

sarnex:
`NonSemantic.AuxData` seems good to me, and it also leaves it open for other instructions besides just the attr/metadata ones I added.

@MrSidims Any opinion?

MrSidims:
NonSemantic.AuxData or NonSemantic.AuxiliaryData seem good

I hope to do proper review tomorrow

lib/SPIRV/SPIRVReader.cpp
```diff
@@ -4452,6 +4457,60 @@ Instruction *SPIRVToLLVM::transOCLBuiltinFromExtInst(SPIRVExtInst *BC,
...
+void SPIRVToLLVM::transAuxDataInst(SPIRVExtInst *BC) {
+  assert(BC->getExtSetKind() == SPIRV::SPIRVEIS_Aux_Data);
+  if (!BC->getModule()->preserveAllFunctionAttributesAndMetadata())
+    return;
+  auto Args = BC->getArguments();
+  // Args 0 and 1 are common between attributes and metadata.
+  // 0 is the function, 1 is the name of the attribute/metadata as a string
+  auto *SpvFcn = BC->getModule()->getValue(Args[0]);
+  auto *F = static_cast<Function *>(getTranslatedValue(SpvFcn));
+  assert(F && "Function should already have been translated!");
```
MrSidims, line 4469: Should we really assert that or may be return void is better?

lib/SPIRV/SPIRVReader.cpp
```diff
@@ -4452,6 +4457,60 @@ Instruction *SPIRVToLLVM::transOCLBuiltinFromExtInst(SPIRVExtInst *BC,
...
+  auto AttrOrMDName = BC->getModule()->get<SPIRVString>(Args[1])->getStr();
+  switch (BC->getExtOp()) {
+  case NonSemanticAuxData::FunctionAttribute: {
+    assert(Args.size() < 4 && "Unexpected FunctionAttribute Args");
+    // If this attr was specially handled and added elsewhere, skip it.
+    if (F->hasFnAttribute(AttrOrMDName))
+      break;
```
MrSidims, line 4476: nit here and after
```suggestion
      return;
```

test/DebugInfo/NonSemantic/Shader200/DebugInfoStringType.ll
```diff
@@ -1,5 +1,5 @@
 ; RUN: llvm-as %s -o %t.bc
-; RUN: llvm-spirv -spirv-text %t.bc -o %t.spt --spirv-debug-info-version=nonsemantic-shader-200
+; RUN: llvm-spirv -spirv-text %t.bc -o %t.spt --spirv-debug-info-version=nonsemantic-shader-200 --spirv-ext=+SPV_KHR_non_semantic_info
```
MrSidims, line 2: No need to enable SPV_KHR_non_semantic_info explicitly for Debug tests - the extension is being enabled implicitly

lib/SPIRV/SPIRVWriter.cpp
```diff
@@ -2761,6 +2828,11 @@ bool LLVMToSPIRVBase::transBuiltinSet() {
             SPIRVBuiltinSetNameMap::map(BM->getDebugInfoEIS()), &EISId))
       return false;
   }
+  if (BM->preserveAllFunctionAttributesAndMetadata()) {
```
MrSidims, line 2831: What will happen if the 'preserve all' flag is set to true while `SPV_KHR_non_semantic_info` is explicitly disabled by `--spirv-ext=-SPV_KHR_non_semantic_info` . I'd expect it will result in an error. But do we want it to be an error? Should we also add a check here like `isAllowedToUseExtension(ExtensionID::SPV_KHR_non_semantic_info)`. Or should we enabled `SPV_KHR_non_semantic_info` implicitly by the flag?

MrSidims (commented):
Overall LGTM
May I ask you to run experiments how test/transcoding/KernelArgTypeInOpString2.ll would act with this option?
Please also update the description of the PR like Intel naming -> current (description will be used as the commit message)

lib/SPIRV/libSPIRV/NonSemantic.AuxData.h
```diff
@@ -0,0 +1,37 @@
...
+** OR IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+** FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL
+** THE AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+** LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING
+** FROM,OUT OF OR IN CONNECTION WITH THE MATERIALS OR THE USE OR OTHER DEALINGS
+** IN THE MATERIALS.
+*/
+
+//
+// Author: Nick Sarnie, Intel
```
svenvh, line 28: We tend to follow https://llvm.org/docs/DeveloperPolicy.html#attribution-of-changes I believe, so please avoid author information in the source code.

lib/SPIRV/libSPIRV/SPIRVEnum.h
```diff
@@ -135,6 +136,7 @@ template <> inline void SPIRVMap<SPIRVExtInstSetKind, std::string>::init() {
       "NonSemantic.Shader.DebugInfo.100");
   add(SPIRVEIS_NonSemantic_Shader_DebugInfo_200,
       "NonSemantic.Shader.DebugInfo.200");
+  add(SPIRVEIS_Aux_Data, "NonSemantic.AuxData");
```
svenvh, line 139: It would be good to be consistent on spelling everywhere: either `AuxData` or `Aux_Data` (personally I prefer the one without `_`).

svenvh, line 139: Since this is a NonSemantic EIS:
```suggestion
  add(SPIRVEIS_NonSemantic_AuxData, "NonSemantic.AuxData");
```

tools/llvm-spirv/llvm-spirv.cpp
```diff
@@ -193,6 +193,10 @@ static cl::opt<bool>
     SPIRVMemToReg("spirv-mem2reg", cl::init(false),
                   cl::desc("LLVM/SPIR-V translation enable mem2reg"));
 
+static cl::opt<bool> SPIRVPreserveAllFunctionAttributesAndMetadata(
+    "spirv-preserve-all-function-attributes-and-metadata", cl::init(false),
```
svenvh, line 197: This is a very long option. How about something shorter like `--spirv-preserve-auxdata`?

lib/SPIRV/SPIRVReader.cpp
sarnex: imo we should assert because i have some code that skips processing the inst if the metadata/attr already exists, if the fcn wasnt translated that logic doesn't work

lib/SPIRV/libSPIRV/NonSemantic.AuxData.h
sarnex: ah sorry, i copied another file that had it

lib/SPIRV/libSPIRV/SPIRVEnum.h
sarnex: good point, will do

tools/llvm-spirv/llvm-spirv.cpp
sarnex: sure, thanks

test/DebugInfo/NonSemantic/Shader200/DebugInfoStringType.ll
sarnex: ah nice that landed, fixed, thanks

lib/SPIRV/SPIRVReader.cpp
sarnex: done for both places, thanks

lib/SPIRV/SPIRVWriter.cpp
sarnex: Just tried this:

```
build/tools/llvm-spirv/llvm-spirv --spirv-preserve-auxdata --spirv-ext=-SPV_KHR_non_semantic_info test/extensions/KHR/SPV_KHR_non_semantic_info/preserve-all-function-attributes.bc

RequiresExtension: Feature requires the following SPIR-V extension:
 SPV_KHR_non_semantic_info

```

imo implicitly enabling it makes sense, let me implement that, thanks

sarnex:
>Please also update the description of the PR like Intel naming -> current (description will be used as the commit message)

Done

lib/SPIRV/SPIRVWriter.cpp
sarnex: ok i believe i implemented this in my latest commit, but im noticing something weird. if you pass `--spirv-preserve-auxdata` AND `---spirv-ext=-SPV_KHR_non_semantic_info`, it doesn't error and uses the extension anyway. is this what we want? i based my change on https://github.com/KhronosGroup/SPIRV-LLVM-Translator/commit/3fe439333601c24d5e8f07862874a369854469be, maybe i did something dumb

lib/SPIRV/SPIRVReader.cpp
```diff
@@ -4462,6 +4467,60 @@ Instruction *SPIRVToLLVM::transOCLBuiltinFromExtInst(SPIRVExtInst *BC,
   return CI;
 }
 
+void SPIRVToLLVM::transAuxDataInst(SPIRVExtInst *BC) {
+  assert(BC->getExtSetKind() == SPIRV::SPIRVEIS_NonSemantic_AuxData);
+  if (!BC->getModule()->preserveAuxData())
```
asudarsa, line 4472: This check can be done at the call-site (outside the loop). Thanks

asudarsa (approved):
Couple of minor nits. Looks good overall. Is there any documentation that will be added to support this? Thanks

lib/SPIRV/SPIRVReader.cpp
sarnex: replied to this in other reply, thanks

asudarsa: In this case, we will be unnecessarily running the loop if the user did not want to preserve AuxData. It is not PR blocking. So pls feel free to ignore.

Thanks

lib/SPIRV/SPIRVWriter.cpp
sarnex: @MrSidims any take on the above? is the behavior incorrect?

MrSidims: I didn't consider this corner case. I think if the extension is explicitly disallowed the translator should respect it. Hence setAllowedToUseExtension should be either modified or wrapped within another function, with finding the extension in the map and not changing Allowed status if the extension is found with 'false' flag. @svenvh WDYT?

svenvh: > I think if the extension is explicitly disallowed the translator should respect it. Hence setAllowedToUseExtension should be either modified or wrapped within another function, with finding the extension in the map and not changing Allowed status if the extension is found with 'false' flag. @svenvh WDYT?

I agree, that's the behavior I'd expect when the extension is explicitly disallowed.

sarnex:
```
Failed Tests (2):
  LLVM_SPIRV :: GroupAndSubgroupInstructions.spvasm
  LLVM_SPIRV :: spirv-tools-dis.ll
```

And in both cases the error is

```
error: exactly one input file must be specified.
```
so it doesn't seem related.

@svenvh @MrSidims Any final feedback? Thanks

sarnex:
Close + Open for CI

MrSidims (approved):

sarnex:
@svenvh Any final feedback or can we merge? Thanks!

svenvh (approved):
LGTM, thanks!

PR #1885, opened 2023-03-16
[NFC] Update complex-constexpr.ll

MrSidims:
bbfb13a5ffbc "[ConstExpr] Remove select constant expression" forbids select constant expression


MrSidims:
I'm not 100% if it's the correct fix, but it seems like existence of inttoptr instruction inside of the expression in the test's IR should be enough to test the functionality

svenvh (approved):
LGTM

PR #1886, opened 2023-03-16
[NFC] Remove #include "LLVMSPIRVOpts.h" from SPIRVEntry.h

MrSidims:


MrSidims:
> This is confusing me, we need the include for `VersionNumber` as far as I can tell. But with the transitive include through `SPIRVEnum.h` it just works fine?

It 'just works' but I didn't notice VersionNumber dependency. So, the fix seems to be incorrect, let me think, how to fix clang-tidy from another direction.

PR #1888, opened 2023-03-17
[CI] Upgrade to actions/checkout@v3

svenvh:
Node.js 12 actions are deprecated, so update to v3 which uses Node 16.

See also https://github.blog/changelog/2022-09-22-github-actions-all-actions-will-begin-running-on-node16-instead-of-node12/


svenvh:
I have raised https://github.com/llvm/llvm-project/issues/61473 for the LLVMExports.cmake issue.

MrSidims (approved):

Issue #1868, opened 2023-03-04
Linking of lib/SPIRV/libLLVMSPIRVLib.so.16 fails


frantisekz:
I've tried to build spirv-llvm-translator using Fedora build infra as a prep for llvm 16 rebase, but it unfortunately failed on linking of lib/SPIRV/libLLVMSPIRVLib.so.16 . I've tried both gcc and clang, with and without lto. Build of llvm 15 tag in this repository works fine with the same build args.

I am using llvm 16.0-rc3.

Full build logs:

[build.log](https://github.com/KhronosGroup/SPIRV-LLVM-Translator/files/10888079/build.log)

MrSidims:
@frantisekz thanks for the report! I will try to find Fedora machine to reproduce it. So far on RHEL9.0 machine with following GCC: gcc (GCC) 11.2.1 20220127 (Red Hat 11.2.1-9) it worked just fine. I was building LLVM myself from release/16.x branch with following HEAD: eb9440b3d8b4 and Translator was on https://github.com/KhronosGroup/SPIRV-LLVM-Translator/tree/llvm_release_160 branch with the following HEAD [5fdc47b1a94db](https://github.com/KhronosGroup/SPIRV-LLVM-Translator/commit/5fdc47b1a94dbf2ad19b3a807736ddde5173e3db)

kraj:
Using yocto, I am on https://github.com/llvm/llvm-project/commit/f8514f04eef7c819f955746b0e6534798dfab741 on llvm 16 branch and latest on llvm_release_160 branch of spirv-llvm-translator
and I also see same errors like @frantisekz

frantisekz:
> @frantisekz thanks for the report! I will try to find Fedora machine to reproduce it. 

You should be able to use RHEL 9.x to replicate the Fedora build process via mock from the EPEL repository if you want. 
The diff applied on top of spirv-llvm-translator dist-git ( https://src.fedoraproject.org/rpms/spirv-llvm-translator ) would be the following:

```
diff --git a/spirv-llvm-translator.spec b/spirv-llvm-translator.spec
index c0076a3..9a57cb8 100644
--- a/spirv-llvm-translator.spec
+++ b/spirv-llvm-translator.spec
@@ -1,10 +1,10 @@
 
-%global commit 78ad93b9952cb78889b86e9aa390d013c49dca0b
+%global commit 296cb645e184864afd28136e453ca161a35728ae
[...]

PR #1889, opened 2023-03-17
Fix standalone builds with LLVM_LINK_LLVM_DYLIB=ON

tuliom:
Move the LLVM components to LINK_COMPONENTS because the DEPENDS list has the same semantics as add_dependencies(). In this
case it doesn't include the LLVM components when calling the linker.

It's almost complete revert of https://github.com/KhronosGroup/SPIRV-LLVM-Translator/pull/1543

### DIFF
--- a/lib/SPIRV/CMakeLists.txt
+++ b/lib/SPIRV/CMakeLists.txt
@@ -38,42 +38,23 @@ set(SRC_LIST
   libSPIRV/SPIRVType.cpp
   libSPIRV/SPIRVValue.cpp
 )
-if(LLVM_LINK_LLVM_DYLIB)
-  add_llvm_library(LLVMSPIRVLib STATIC DISABLE_LLVM_LINK_LLVM_DYLIB
-    ${SRC_LIST}
-    DEPENDS
-      intrinsics_gen
-      LLVMAnalysis
-      LLVMBitWriter
-      LLVMCodeGen
-      LLVMCore
-      LLVMDemangle
-      LLVMIRReader
-      LLVMLinker
-      LLVMPasses
-      LLVMSupport
-      LLVMTargetParser
-      LLVMTransformUtils
-  )
-else()
-  add_llvm_library(LLVMSPIRVLib
-    ${SRC_LIST}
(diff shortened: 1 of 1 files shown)


frantisekz:
This fixes https://github.com/KhronosGroup/SPIRV-LLVM-Translator/issues/1868 .

MrSidims:
Test failure is unrelated

MrSidims:
@fineg74 please take a look

MrSidims:
@tuliom please don't forget to sign CLA

tuliom:
Heads up: I'm still working on the CLA stuff and will keep you updated.
Thanks for your patience!

MrSidims:
>> Unexpectedly Passed Tests (1):
  LLVM_SPIRV :: transcoding/pipe_builtins.cl
  
Unrelated failure. I'm restarting CI with new head to get fresh results.

tuliom:
I've just signed the CLA.
Thanks again for your patience!

MrSidims:
@tuliom thanks a lot for your patch!
btw, would you mind applying https://github.com/KhronosGroup/SPIRV-LLVM-Translator/issues/1936#issuecomment-1496563307 ?

tuliom:
> @tuliom thanks a lot for your patch! btw, would you mind applying [#1936 (comment)](https://github.com/KhronosGroup/SPIRV-LLVM-Translator/issues/1936#issuecomment-1496563307) ?

Done.

MrSidims (approved):
LGTM, thanks!
@svenvh may I ask you to take a look?

aaronpuchert:
This seems almost like a revert of #1543. I don't think there is anything wrong with that, it's a workaround for a specific downstream build that perhaps shouldn't be here at all. But maybe the workaround can be limited to that specific downstream build, so that it doesn't bother anyone else. I wrote such a comment on the original change.

Otherwise, if we like this change as it is, it should perhaps be officially declared as a revert in the interest of transparency.

MrSidims:
I've merged the patch adding the revert comment in the message.

PR #1895, opened 2023-03-22
Translate annotation string with the braces for UserSemantic decoration

vmaksimo:
UserSemantic's handling is done alike to other decorations. However, we should preserve the braces so we save the full input string during the forward and reverse translation.


asudarsa (approved):
LGTM. Thanks

MrSidims (approved):

PR #1896, opened 2023-03-22
[NFC][DebugInfo] Align Instruction names to the specification

vmaksimo:
This patch fixes the discovered typos in Debug Instruction names, so we can generate spec-conformant SPIR-V module.
https://registry.khronos.org/SPIR-V/specs/unified1/DebugInfo.html

### DIFF
--- a/lib/SPIRV/LLVMToSPIRVDbgTran.cpp
+++ b/lib/SPIRV/LLVMToSPIRVDbgTran.cpp
@@ -344,7 +344,7 @@ SPIRVEntry *LLVMToSPIRVDbgTran::transDbgEntryImpl(const MDNode *MDN) {
 
     // Compilation unit
     case dwarf::DW_TAG_compile_unit:
-      return transDbgCompilationUnit(cast<DICompileUnit>(DIEntry));
+      return transDbgCompileUnit(cast<DICompileUnit>(DIEntry));
 
     // Templates
     case dwarf::DW_TAG_template_type_parameter:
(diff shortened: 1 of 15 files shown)


vmaksimo:
Please take a look @MrSidims @asudarsa @jgstarIntel

MrSidims (approved):
LGTM, thanks!

lib/SPIRV/LLVMToSPIRVDbgTran.cpp
```diff
@@ -344,7 +344,7 @@ SPIRVEntry *LLVMToSPIRVDbgTran::transDbgEntryImpl(const MDNode *MDN) {
 
     // Compilation unit
     case dwarf::DW_TAG_compile_unit:
-      return transDbgCompilationUnit(cast<DICompileUnit>(DIEntry));
+      return transDbgCompileUnit(cast<DICompileUnit>(DIEntry));
```
asudarsa, line 347: I see this name is DebugCompilationUnit in the Spec. Can you please check? Thanks

lib/SPIRV/LLVMToSPIRVDbgTran.h
```diff
@@ -99,7 +99,7 @@ class LLVMToSPIRVDbgTran {
   SPIRVId getDebugInfoNoneId();
 
   // Compilation unit
-  SPIRVEntry *transDbgCompilationUnit(const DICompileUnit *CU);
+  SPIRVEntry *transDbgCompileUnit(const DICompileUnit *CU);
```
asudarsa, line 102: Same as above comment.

lib/SPIRV/LLVMToSPIRVDbgTran.cpp
vmaksimo: Yes, I used the logic Dmitry described. So no more changes are expected here

asudarsa: It might be ideal to have the same name in both. But I think that is for a different PR. Thanks

asudarsa (approved):
LGTM. Thanks

PR #1899, opened 2023-03-23
fix the ptr.annotation generation in reverse translator

broxigarchen:
Fix the `ptr.annotation` in reverse translator such that the generated pointer annotation call is being used properly in the IR.

The IR generated from the previous code base is:
```
%0 = xxx
%1 = ptr.annotation %0 ...  (left unused)
%2 = load %0 ...
```
while the expected form is:
```
%0 = xxx
%1 = ptr.annotation %0 ...
%2 = load %1 ...
```

### DIFF
--- a/lib/SPIRV/SPIRVReader.cpp
+++ b/lib/SPIRV/SPIRVReader.cpp
@@ -1280,7 +1280,7 @@ static void replaceOperandWithAnnotationIntrinsicCallResult(Function *F,
     }
   }
 
-  // overwite operand with intrinsic call result
+  // overwrite operand with intrinsic call result
   if (CallResult) {
     SPIRVDBG(dbgs() << "BaseValue: " << *BaseValue << '\n'
                     << "CallResult: " << *CallResult << '\n');
--- a/lib/SPIRV/SPIRVReader.cpp
+++ b/lib/SPIRV/SPIRVReader.cpp
@@ -1236,6 +1236,8 @@ Value *SPIRVToLLVM::oclTransConstantPipeStorage(
                             GlobalValue::NotThreadLocal, SPIRAS_Global);
 }
 
+namespace {
+
 // A pointer annotation may have been generated for the operand. If the operand
 // is used further in IR, it should be replaced with the intrinsic call result.
 // Otherwise, the generated pointer annotation call is left unused.
@@ -1292,6 +1294,8 @@ static void replaceOperandWithAnnotationIntrinsicCallResult(Function *F,
       V = CallResult;
   }
 }
+
+} // namespace
 
 // Translate aliasing memory access masks for SPIRVLoad and SPIRVStore
 // instructions. These masks are mapped on alias.scope and noalias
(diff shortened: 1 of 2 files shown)


MrSidims (commented):
I though that there was/is a patch from @vmaksimo to fix annotations on dead code path

broxigarchen:
> I though that there was/is a patch from @vmaksimo to fix annotations on dead code path

Hi @vmaksimo can you comment? I am posting this PR since the annotated_ptr feature that I am working on is pending on it

vmaksimo:
Hi @broxigarchen. I did the change with the similar impact in January https://github.com/KhronosGroup/SPIRV-LLVM-Translator/pull/1723
But I guess your patch could be more general, I somehow tested my PR only for clang::annotate (UserSemantic decorations).

Could you please add/modify some test so we can see which case is not covered by the current implementation?

broxigarchen:
> Hi @broxigarchen. I did the change with the similar impact in January #1723 But I guess your patch could be more general, I somehow tested my PR only for clang::annotate (UserSemantic decorations).
> 
> Could you please add/modify some test so we can see which case is not covered by the current implementation?

Sure. Let me update the lit test that was failing without this patch

broxigarchen:
> Hi @broxigarchen. I did the change with the similar impact in January #1723 But I guess your patch could be more general, I somehow tested my PR only for clang::annotate (UserSemantic decorations).
> 
> Could you please add/modify some test so we can see which case is not covered by the current implementation?

Done. In the updated lit test, there are multiple duplicated bit-cast auto-generated from the reverse translator, and I guess that could be why it failed to recognize the ptr.annotation before

lib/SPIRV/SPIRVReader.cpp
```diff
@@ -1236,20 +1236,67 @@ Value *SPIRVToLLVM::oclTransConstantPipeStorage(
...
+    while (Inst && (isa<BitCastInst>(Inst) || isa<AddrSpaceCastInst>(Inst))) {
+      if ((Inst = dyn_cast<Instruction>(Inst->getOperand(0))) &&
+          SearchPtrAnn(Inst, CallResult)) {
+        BaseValue = Inst;
+        break;
+      }
     }
   }
+
+  // overwite operand with intrinsic call result
```
MrSidims, line 1283:
```suggestion
  // overwrite operand with intrinsic call result
```

lib/SPIRV/SPIRVReader.cpp
```diff
@@ -1236,20 +1236,67 @@ Value *SPIRVToLLVM::oclTransConstantPipeStorage(
...
+  if (CallResult) {
+    SPIRVDBG(dbgs() << "BaseValue: " << *BaseValue << '\n'
+                    << "CallResult: " << *CallResult << '\n');
+    DominatorTree DT(*F);
+    BaseValue->replaceUsesWithIf(CallResult, [&DT, &CallResult](Use &U) {
+      return DT.dominates(CallResult, U);
+    });
+
+    // overwite V
```
MrSidims, line 1292:
```suggestion
    // overwrite V
```

MrSidims (approved):

MrSidims:
Restarting CI with new head

PR #1901, opened 2023-03-24
Improve DbgArrayType compatibility

KorovinVlad:
llvm11+ has UpperBound and LowerBound as metadata. This commit implements a loop the same way it is implemented in new translator versions(llvm14+)


MrSidims (approved):
LGTM, but would love to hear from @vmaksimo

PR #1902, opened 2023-03-27
[Backmerge to 11] [NFC] Optimize adding of decorations (#1233)

aratajew:
* [NFC] Optimize adding of decorations

Use std::undordered_set instead of std::multiset with custom comparator. It is a bit unclear why std::multiset was required since there is no order requirement for OpDecorate instructions in SPIR-V spec, in addition old data structure was highly inefficient.


aratajew:
@MrSidims, @svenvh could we please merge this backport ASAP? :) Some kernels (especially these compiled in O0) translates even 50x slower without this change.

MrSidims (approved):
LGTM, was a bit scared to see 65k changes looking at the weekends :D

PR #1918, opened 2023-03-29
[Backmerge to 11] Improve DbgArrayType compatibility (#1901)

KorovinVlad:
Backport of https://github.com/KhronosGroup/SPIRV-LLVM-Translator/pull/1901


stanleygambarin:
OpenSource version is currently based on LLVM11, so this patch would fix few tests in a short time.  We are moving to LLVM14, which would make this patch obsolete.  I do *not* see any issue with having this patch enabled though.

PR #1922, opened 2023-03-30
Add translation of fpclass llvm intrinsic

MrSidims:
Optimized LLVM IR started to contain this intrinsic recently after InstCombine updates, like:
"Fold and/or of fcmp into class" ( 08f03887 )

There is no direct counterpart for it in SPIR-V, so testing is mapped on a sequence of SPIR-V instructions:
1. test for both qnan and snan -> OpIsNan
2. test for either of qnan and snan -> isquiet(V) ==> abs(V) >= (unsigned(Inf) | quiet_bit)
                                                             issignaling(V) ==> isnan(V) && !isquiet(V)
3. test for neg/pos inf -> OpIsInf + OpSignBitSet
4. test for neg/pos normal -> OpIsNormal + OpSignBitSet
5. test for neg/pos subnormal -> issubnormal(V) ==> unsigned(abs(V) - 1) < (all mantissa bits set)
6. test for neg/pos zero -> compare bitcasted to int float with 0 + OpSignBitSet

- [x] add test

### DIFF
--- a/lib/SPIRV/SPIRVWriter.cpp
+++ b/lib/SPIRV/SPIRVWriter.cpp
@@ -4245,6 +4245,96 @@ SPIRVValue *LLVMToSPIRVBase::transIntrinsicInst(IntrinsicInst *II,
     return BM->addInstTemplate(internal::OpMaskedScatterINTEL, Ops, BB,
                                nullptr);
   }
+  case Intrinsic::is_fpclass: {
+    // There is no direct counterpart for the intrinsic in SPIR-V, hence
+    // we need to emulate it's work by sequence of other instructions
+    SPIRVType *Ty = transType(II->getType());
+    auto *InputFloat = transValue(II->getArgOperand(0), BB);
+    uint64_t FPClass =
+        cast<ConstantInt>(II->getArgOperand(1))->getZExtValue();
+    std::vector<SPIRVValue *> ResultVec;
+    std::vector<SPIRVId> Ops = {InputFloat->getId()};
+    // Adds test for Negative/Positive values
+    auto GetNegPosInstTest = [&](SPIRVValue *TestInst, bool IsNegative) {
+      auto *SignBitTest =
+          BM->addInstTemplate(OpSignBitSet, {TestInst->getId()}, BB, Ty);
+      if (IsNegative)
+        return SignBitTest;
+      return BM->addInstTemplate(OpLogicalNot, {SignBitTest->getId()}, BB, Ty);
+    };
+    // Integer paramter of the intrinsic is combined from several bit masks
+    // referenced in FPClassTest enum from FloatingPointMode.h in LLVM.
+    // Since a single intrinsic can provide multiple tests - here we might end
+    // up adding several sequences of SPIR-V instructions
+    if (FPClass & fcSNan || FPClass & fcQNan) {
+      // There are no signaling and quite NaN instructions in SPIR-V, so map
+      // both of the tests on OpIsNan
+      ResultVec.emplace_back(BM->addInstTemplate(OpIsNan, Ops, BB, Ty));
+    }
+    if (FPClass & fcNegInf || FPClass & fcPosInf) {
+      auto *TestIsInf = BM->addInstTemplate(OpIsInf, Ops, BB, Ty);
+      if (FPClass & fcNegInf && FPClass & fcPosInf)
+        // Map on OpIsInf if we have both Inf test bits set
+        ResultVec.emplace_back(TestIsInf);
+      else
+        // Map on OpIsInf with following check for sign bit
+        ResultVec.emplace_back(GetNegPosInstTest(TestIsInf,
+                                                 FPClass & fcNegInf));
+    }
+    if (FPClass & fcNegNormal || FPClass & fcPosNormal) {
+      auto *TestIsNormal = BM->addInstTemplate(OpIsNormal, Ops, BB, Ty);
+      if (FPClass & fcNegNormal && FPClass & fcPosNormal)
+        // Map on OpIsNormal if we have both Normal test bits set
+        ResultVec.emplace_back(TestIsNormal);
+      else
+        // Map on OpIsNormal with following check for sign bit
+        ResultVec.emplace_back(GetNegPosInstTest(TestIsNormal,
+                                                 FPClass & fcNegNormal));
+    }
+    if (FPClass & fcNegSubnormal || FPClass & fcPosSubnormal) {
+      // If value is both not NaN and Normal, then it's subnormal
+      // TODO: it's quite silly check, can we figure out something better?
+      auto *TestIsNan = BM->addInstTemplate(OpIsNan, Ops, BB, Ty);
+      auto *TestIsNormal = BM->addInstTemplate(OpIsNormal, Ops, BB, Ty);
+      auto *TestIsNotSubnormal =
+          BM->addInstTemplate(OpLogicalOr, {TestIsNan->getId(),
+                                            TestIsNormal->getId()}, BB, Ty);
+      auto *TestIsSubnormal =
+          BM->addInstTemplate(OpLogicalNot, {TestIsNotSubnormal->getId()}, BB,
+                              Ty);
+      if (FPClass & fcNegSubnormal && FPClass & fcPosSubnormal)
+        ResultVec.emplace_back(TestIsSubnormal);
+      else
+        ResultVec.emplace_back(GetNegPosInstTest(TestIsSubnormal,
+                                                 FPClass & fcNegZero));
+    }
+    if (FPClass & fcNegZero || FPClass & fcPosZero) {
+      // Map on OpFOrdEqual compare to 0.0 with following check for sign bit
+      Constant *Zero = ConstantFP::getZero(II->getArgOperand(0)->getType());
+      auto *TestIsZero = BM->addCmpInst(OpFOrdEqual, Ty, InputFloat,
+                                        transValue(Zero, BB), BB);
+      if (FPClass & fcNegZero && FPClass & fcPosZero)
+        // Map on OpFOrdEqual compare to 0.0
+        ResultVec.emplace_back(TestIsZero);
+      else
+        // Map on OpFOrdEqual compare to 0.0 with following check for sign bit
+        ResultVec.emplace_back(GetNegPosInstTest(TestIsZero,
+                                                 FPClass & fcNegZero));
+    }
+    // if no tests are provided - return true
+    if (ResultVec.empty())
+      return BM->addConstant(Ty, 1);
+    if (ResultVec.size() == 1)
+      return ResultVec.back();
+    SPIRVValue *Result = ResultVec.front();
+    for (size_t I = 1; I != ResultVec.size(); ++I) {
+      // Create a sequence of LogicalOr instructions from ResultVec to get
+      // the overall test result
+      std::vector<SPIRVId> LogicOps = {Result->getId(), ResultVec[I]->getId()};
+      Result = BM->addInstTemplate(OpLogicalOr, Ops, BB, Ty);
+    }
+    return Result;
+  }
 
   default:
     if (BM->isUnknownIntrinsicAllowed(II))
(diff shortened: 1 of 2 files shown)


lib/SPIRV/SPIRVWriter.cpp
```diff
@@ -4245,6 +4245,96 @@ SPIRVValue *LLVMToSPIRVBase::transIntrinsicInst(IntrinsicInst *II,
...
+        // Map on OpIsNormal if we have both Normal test bits set
+        ResultVec.emplace_back(TestIsNormal);
+      else
+        // Map on OpIsNormal with following check for sign bit
+        ResultVec.emplace_back(GetNegPosInstTest(TestIsNormal,
+                                                 FPClass & fcNegNormal));
+    }
+    if (FPClass & fcNegSubnormal || FPClass & fcPosSubnormal) {
+      // If value is both not NaN and Normal, then it's subnormal
+      // TODO: it's quite silly check, can we figure out something better?
```
MrSidims, line 4296: @svenvh @asudarsa @bashbaug may I ask you to take a look? I assumed, that subnormal test should be test for not Nan and not Normal, though may be you see better ways to perform that.

lib/SPIRV/SPIRVWriter.cpp
```diff
@@ -4245,6 +4245,96 @@ SPIRVValue *LLVMToSPIRVBase::transIntrinsicInst(IntrinsicInst *II,
...
+      if (FPClass & fcNegSubnormal && FPClass & fcPosSubnormal)
+        ResultVec.emplace_back(TestIsSubnormal);
+      else
+        ResultVec.emplace_back(GetNegPosInstTest(TestIsSubnormal,
+                                                 FPClass & fcNegZero));
+    }
+    if (FPClass & fcNegZero || FPClass & fcPosZero) {
+      // Map on OpFOrdEqual compare to 0.0 with following check for sign bit
+      Constant *Zero = ConstantFP::getZero(II->getArgOperand(0)->getType());
+      auto *TestIsZero = BM->addCmpInst(OpFOrdEqual, Ty, InputFloat,
```
MrSidims, line 4314: Adding OpFOrdEqual is also a bit questionable, since it's instruction from SPIR-V 1.6 and we officially support only SPIR-V 1.4 in the translator.

lib/SPIRV/SPIRVWriter.cpp
```diff
@@ -4245,6 +4245,96 @@ SPIRVValue *LLVMToSPIRVBase::transIntrinsicInst(IntrinsicInst *II,
...
+      return BM->addInstTemplate(OpLogicalNot, {SignBitTest->getId()}, BB, Ty);
+    };
+    // Integer parameter of the intrinsic is combined from several bit masks
+    // referenced in FPClassTest enum from FloatingPointMode.h in LLVM.
+    // Since a single intrinsic can provide multiple tests - here we might end
+    // up adding several sequences of SPIR-V instructions
+    if (FPClass & fcSNan || FPClass & fcQNan) {
+      // There are no signaling and quite NaN instructions in SPIR-V, so map
+      // both of the tests on OpIsNan
+      ResultVec.emplace_back(BM->addInstTemplate(OpIsNan, Ops, BB, Ty));
```
jcranmer-intel, line 4272: It's not correct to consider the tests for sNaN or qNaN as equivalent, because the definition is based on the bit pattern (and InstCombine may canonicalize such a test to an fpclass check!). Distinguishing between qNaN and sNaN requires checking if the quiet bit (MSB of the mantissa) is set or not.

lib/SPIRV/SPIRVWriter.cpp
jcranmer-intel: Simplest check here is to check if the exponent bits are minimum exponent, and mantissa is nonzero. Looking at the TargetLowering variant, this can be reduced further to `unsigned(mask_sign(V) - 1) < all_ones_mantissa`

jcranmer-intel (commented):
Since I've been reviewing several of the upstream patches related here, I felt I should give some drive-by comments.

The "correct" lowering of is_fpclass is given by TargetLowering::expandIS_FPCLASS  (https://github.com/llvm/llvm-project/blob/main/llvm/lib/CodeGen/SelectionDAG/TargetLowering.cpp#L8019). Instcombine should be canonicalizing the tests that are equivalent to a single comparison into a single instruction, so checking for the really simple cases may not be all that beneficial.

The general strategy for the implementation should not to be to check with floating-point operations, but integer operations on bitcast-to-int. This avoids issues where floating-point operations may have underspecified issues (e.g., comparison to zero returning different results depending on denormals-are-zero setting).

MrSidims:
@jcranmer-intel thanks! I've applied the comments. Not QNan, SNan, PosSubnormal, NegSubnormal, PosZero and NegZero are checked after bitcast to int. Checks for normal and inf remained to be the same - relying on existing SPIR-V instructions.

lib/SPIRV/SPIRVWriter.cpp
```diff
@@ -4245,6 +4245,151 @@ SPIRVValue *LLVMToSPIRVBase::transIntrinsicInst(IntrinsicInst *II,
...
+    const APInt Inf = APFloat::getInf(Semantics).bitcastToAPInt();
+    const APInt AllOneMantissa =
+        APFloat::getLargest(Semantics).bitcastToAPInt() & ~Inf;
+
+    // Integer parameter of the intrinsic is combined from several bit masks
+    // referenced in FPClassTest enum from FloatingPointMode.h in LLVM.
+    // Since a single intrinsic can provide multiple tests - here we might end
+    // up adding several sequences of SPIR-V instructions
+    if (FPClass & fcNan) {
+      auto *TestIsNan = BM->addInstTemplate(OpIsNan, {InputFloat->getId()}, BB,
```
MrSidims, line 4293: To move it a bit later in the code

lib/SPIRV/SPIRVWriter.cpp
```diff
@@ -4245,6 +4245,151 @@ SPIRVValue *LLVMToSPIRVBase::transIntrinsicInst(IntrinsicInst *II,
...
+    if (FPClass & fcInf) {
+      auto *TestIsInf = BM->addInstTemplate(OpIsInf, {InputFloat->getId()}, BB,
+                                            ResTy);
+      if (FPClass & fcNegInf && FPClass & fcPosInf)
+        // Map on OpIsInf if we have both Inf test bits set
+        ResultVec.emplace_back(TestIsInf);
+      else
+        // Map on OpIsInf with following check for sign bit
+        ResultVec.emplace_back(GetNegPosInstTest(TestIsInf,
+                                                 FPClass & fcNegInf));
```
MrSidims, line 4335: Here and after logical error: check for bit should be done on float, not on IsTest result

MrSidims:
@jcranmer-intel @asudarsa @svenvh @maksimsab @jgstarIntel please take a look

lib/SPIRV/SPIRVWriter.cpp
```diff
@@ -4245,6 +4245,153 @@ SPIRVValue *LLVMToSPIRVBase::transIntrinsicInst(IntrinsicInst *II,
...
+    // if all tests are provided - return true
+    if (FPClass == fcAllFlags)
+      return BM->addConstant(ResTy, true);
+
+    Type *OpLLVMTy = II->getArgOperand(0)->getType();
+    SPIRVValue *InputFloat = transValue(II->getArgOperand(0), BB);
+    std::vector<SPIRVValue *> ResultVec;
+
+    // Adds test for Negative/Positive values
+    auto GetNegPosInstTest = [&](SPIRVValue *TestInst, bool IsNegative) {
```
MrSidims, line 4265: Hm, I feel like there is room for optimization(s): for example test for sign can be cached at first entry and its ID can be reused by other tests.

test/llvm-intrinsics/fpclass.ll
```diff
@@ -0,0 +1,451 @@
+; RUN: llvm-as %s -o %t.bc
+; RUN: llvm-spirv %t.bc -spirv-text -o - | FileCheck %s --check-prefix=CHECK-SPIRV
+; RUN: llvm-spirv %t.bc -o %t.spv
+; RUN: spirv-val %t.spv
+; RUN: llvm-spirv -r %t.spv -o %t.rev.bc
+
+; Just check that reverse translation works
```
asudarsa, line 7: Can we have a quick check to ensure that fpclass is not generated here?

Thanks

lib/SPIRV/SPIRVWriter.cpp
```diff
@@ -4245,6 +4245,153 @@ SPIRVValue *LLVMToSPIRVBase::transIntrinsicInst(IntrinsicInst *II,
...
+          BM->addInstTemplate(OpIsInf, {InputFloat->getId()}, BB, ResTy);
+      if (FPClass & fcNegInf && FPClass & fcPosInf)
+        // Map on OpIsInf if we have both Inf test bits set
+        ResultVec.emplace_back(TestIsInf);
+      else
+        // Map on OpIsInf with following check for sign bit
+        ResultVec.emplace_back(
+            GetNegPosInstTest(TestIsInf, FPClass & fcNegInf));
+    }
+    if (FPClass & fcNormal) {
```
asudarsa, line 4339: Is there a possibility that fcNormal bit is set and both fcPosNormal and fcNegNormal are not set? In such such, what is the expected behavior? Thanks

asudarsa (approved):
Overall LGTM. Just a couple of minor questions/comments.

Thanks

test/llvm-intrinsics/fpclass.ll
MrSidims: I guess it's being checked implicitly by the fact, that forward translation doesn't error out due to unknown intrinsic

MrSidims:
@asudarsa may I ask to take another look? I've added 2 enhancements: added caching for sign test results to be reused => less IsSignBit instructions generated and added check for inverted test aka (is inf, is normal, is subnormal is zero) = (is not nan)

MrSidims:
@jcranmer-intel may I ask you to revisit the PR?

jcranmer-intel (approved):
In theory, the tests could be a little more comprehensive, but without an `update-test-checks`-like script, it is going to be too painful to make such a test for the limited extra value it would bring.

lib/SPIRV/SPIRVWriter.cpp
```diff
@@ -4318,6 +4245,204 @@ SPIRVValue *LLVMToSPIRVBase::transIntrinsicInst(IntrinsicInst *II,
...
+        return InvertedTest;
+      }
+      return fcNone;
+    };
+    bool IsInverted = false;
+    if (llvm::FPClassTest InvertedCheck = GetInvertedFPClassTest(FPClass)) {
+      IsInverted = true;
+      FPClass = InvertedCheck;
+    }
+    auto GetInvertedTestIfNeeded = [&](SPIRVValue *TestInst) -> SPIRVValue * {
```
karolzwolak, line 4335: This is an old PR right now but I believe this was ported incorrectly. If we invert the test, we should negate the end result, not the individual checks.

Inverting the individual class tests makes no sense.

Suppose we have 5 classes ABCDE, and we have a check for 3 of them (ABC), which we decide to invert.
Inverted, that's 2 checks (DE).
Right now, we invert each check, so we will produce:
!D || !E <- oops, returns true for E!
instead, we should invert the end result:
!(D || E) <- neither D nor E, so A or B or C

This mistake probably comes from the fact that in the LLVM implementation, they have an optimization when there's only a singular check, in which case, we can invert the test as it's also the result.
Inversion of each case, but there's no more tests, they return immediately:
https://github.com/llvm/llvm-project/blob/2aab31a94e7b9eb0c66d3a9639e987a257efc444/llvm/lib/CodeGen/SelectionDAG/TargetLowering.cpp#L9158-L9162
https://github.com/llvm/llvm-project/blob/2aab31a94e7b9eb0c66d3a9639e987a257efc444/llvm/lib/CodeGen/SelectionDAG/TargetLowering.cpp#L9126-L9130

In the regular case with more than a few tests, they invert only the result:
https://github.com/llvm/llvm-project/blob/2aab31a94e7b9eb0c66d3a9639e987a257efc444/llvm/lib/CodeGen/SelectionDAG/TargetLowering.cpp#L9430-L9434
No inversion in append result:
https://github.com/llvm/llvm-project/blob/2aab31a94e7b9eb0c66d3a9639e987a257efc444/llvm/lib/CodeGen/SelectionDAG/TargetLowering.cpp#L9280-L9287

MrSidims: Thanks! I'll open an issue and investigate this concern a bit later.

MrSidims: (note, there were some follow up PRs, may be they resolve the issue, may be not, requires me to double check it :) )

Created https://github.com/KhronosGroup/SPIRV-LLVM-Translator/issues/3696

PR #1926, opened 2023-03-31
Update spirv-as invocations

svenvh:
After SPIRV-Tools commit 2e0f4b52 ("tools: refactorize tools flags parsing. (#5111)", 2023-02-27), spirv-as needs to be told explicitly when reading from stdin.


MrSidims (approved):
Thanks! I can cherry-pick it to the release branches.

PR #1935, opened 2023-03-31
Start preparing for TypeJointMatrixINTEL switch

MrSidims:
The patch adds TypeJointMatrixINTELv2 which maps to new type OpCode
6184. Under new OpCode matrix type no longer has Layout parameter. The patch also moved 'scope' to optional matrix muladd instruction.

The changes are done only in the consumer part to prepare the switch and make E2E switch backward compatible by preparing consumers ahead of time.

Unfortunately there is no way to add a test foe this unless it's binary test, but it seems to be a bit unsafe to add this, so the patch was tested localy.

Spec change:


MrSidims:
@asudarsa @maksimsab @jgstarIntel please take a look

lib/SPIRV/libSPIRV/SPIRVType.h
```diff
@@ -1085,11 +1090,25 @@ class SPIRVTypeJointMatrixINTEL : public SPIRVType {
...
   SPIRVValue *getColumns() const { return Args[1]; }
-  SPIRVValue *getLayout() const { return Args[2]; }
-  SPIRVValue *getScope() const { return Args[3]; }
-  SPIRVValue *getUse() const { return Args.size() > 4 ? Args[4] : nullptr; }
+  SPIRVValue *getLayout() const {
+    if (this->getOpCode() == internal::OpTypeJointMatrixINTEL)
+      return Args[2];
+    return nullptr;
+  }
+  SPIRVValue *getScope() const {
```
maksimsab, line 1098: Could you please separate this methods by blank lines?

maksimsab (changes requested):
I see in the patch description that `scope is moved to optional matrix muladd instruction`.

Could you please clarify where is it done? Unfortunately, not obvious for me :(

lib/SPIRV/libSPIRV/SPIRVInstruction.h
```diff
@@ -3329,10 +3330,10 @@ class SPIRVJointMatrixINTELInst : public SPIRVJointMatrixINTELInstBase {
...
 _SPIRV_OP(JointMatrixLoad, true, 6, true)
 _SPIRV_OP(JointMatrixStore, false, 5, true)
-_SPIRV_OP(JointMatrixMad, true, 7)
-_SPIRV_OP(JointMatrixSUMad, true, 7)
-_SPIRV_OP(JointMatrixUSMad, true, 7)
-_SPIRV_OP(JointMatrixUUMad, true, 7)
+_SPIRV_OP(JointMatrixMad, true, 6, true)
+_SPIRV_OP(JointMatrixSUMad, true, 6, true)
+_SPIRV_OP(JointMatrixUSMad, true, 6, true)
+_SPIRV_OP(JointMatrixUUMad, true, 6, true)
```
MrSidims, line 3336: @maksimsab move of scope parameter happens here (see the second 'true').

Note, that 'scope' according to old spec is mandatory parameter and per new spec is removed at all. So to balance it here making both representations translatable we have to add is as optional for a while.

lib/SPIRV/libSPIRV/SPIRVType.cpp
```diff
@@ -279,13 +280,19 @@ void SPIRVTypeForwardPointer::decode(std::istream &I) {
...
+    SPIRVModule *M, SPIRVId TheId, Op OC, SPIRVType *CompType,
     std::vector<SPIRVValue *> Args)
     : SPIRVType(M, FixedWC + Args.size(), OC, TheId), CompType(CompType),
       Args(Args) {}
 
+SPIRVTypeJointMatrixINTEL::SPIRVTypeJointMatrixINTEL(
+    SPIRVModule *M, SPIRVId TheId, SPIRVType *CompType,
+    std::vector<SPIRVValue *> Args)
+    : SPIRVType(M, FixedWC + Args.size(), internal::OpTypeJointMatrixINTEL,
+                TheId), CompType(CompType), Args(Args) {}
```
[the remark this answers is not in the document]
MrSidims: Thanks! We probably should also scan other places in the translator to embrace this optimization

maksimsab (approved):
LGTM! Thank you

PR #1947, opened 2023-04-04
Support 0 value of Intel-specific RegisterAllocMode MD

sarnex:
We are planning on using this in the backend.

### DIFF
--- a/test/transcoding/registerallocmode.ll
+++ b/test/transcoding/registerallocmode.ll
@@ -6,16 +6,18 @@
 
 ; CHECK-SPIRV: 8 Decorate 4 UserSemantic "num-thread-per-eu 4"
 ; CHECK-SPIRV: 8 Decorate 6 UserSemantic "num-thread-per-eu 8"
-; CHECK-SPIRV-NOT: 8 Decorate 8
+; CHECK-SPIRV  8 Decorate 8 UserSemantic "num-thread-per-eu 0"
 ; CHECK-SPIRV-NOT: 8 Decorate 10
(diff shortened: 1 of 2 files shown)


sarnex:
@MrSidims Please review when you get a chance, I can't set the reviewers on here. Should be pretty non-interesting (hopefully)

test/transcoding/registerallocmode.ll
```diff
@@ -6,16 +6,18 @@
 
 ; CHECK-SPIRV: 8 Decorate 4 UserSemantic "num-thread-per-eu 4"
 ; CHECK-SPIRV: 8 Decorate 6 UserSemantic "num-thread-per-eu 8"
-; CHECK-SPIRV-NOT: 8 Decorate 8
+; CHECK-SPIRV  8 Decorate 8 UserSemantic "num-thread-per-eu 0"
```
MrSidims, line 9:
```suggestion
; CHECK-SPIRV:  8 Decorate 8 UserSemantic "num-thread-per-eu 0"
```

MrSidims: Also, it seems like I've missed this during review of the initial patch.
Please do not add word count check of the instruction aka "8 Decorate 8" -> "Decorate 8". Also in remaining "Decorate 8" '8' is an ID of something, please explicitly add a check for this something and reuse its ID as a CHECK variable.

sarnex: whoops, all fixed, thanks

MrSidims (approved):

PR #1948, opened 2023-04-04
Allow ICmpInst in transValue assert

LU-JOHN:
Fix assertion failure in transValue when ICmpInst is hit.


test/transValueAssert.ll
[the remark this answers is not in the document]
MrSidims: Yeah, we've discussed it offline. The test will be reduced

asudarsa:
Hi @LU-JOHN 

Please convert the test to use opaque pointers. Also, please address the test fail. I think it's failing in the spirv-validator (spirv-val) test.

Thanks

test/constexpr_icmp.ll
```diff
@@ -0,0 +1,48 @@
+; RUN: llvm-as -opaque-pointers=0 %s -o %t.bc
+; RUN: llvm-spirv %t.bc  -spirv-ext=+SPV_INTEL_vector_compute -opaque-pointers=0 -o %t.spv
```
MrSidims, line 2:
```suggestion
; RUN: llvm-as %s -o %t.bc
; RUN: llvm-spirv %t.bc  -spirv-ext=+SPV_INTEL_vector_compute -o %t.spv
```

btw, why do we need SPV_INTEL_vector_compute here?

MrSidims (commented):
Do we have insights why spirv-val failed with
error: line 29: ID '32[%32]' has not been defined
  %[33](https://github.com/KhronosGroup/SPIRV-LLVM-Translator/actions/runs/4622113314/jobs/8174381839#step:8:34) = OpSpecConstantOp %uchar CompositeExtract %32 0
  
  ?
That's look a bit strange though - there are no changes in input/output llvm.dbg.value

test/constexpr_icmp.ll
LU-JOHN: I think the failure:

error: line 29: ID '32[%32]' has not been defined

is because llvm.dbg.value is declared and not defined.   Other tests that declare llvm.dbg.value (complex-constexpr-vector and lshr-constexpr.ll) do not validate with spirv-val.

If the option SPV_INTEL_vector_compute is left out, there is an assertion failure:

llvm-spirv: /iusers/lujohn/src/SPIRV-Team/SPIRV-LLVM-Translator/lib/SPIRV/libSPIRV/SPIRVType.h:334: virtual void SPIRV::SPIRVTypeVector::validate() const: Assertion `CompCount == 2 || CompCount == 3 || CompCount == 4 || CompCount == 8 || CompCount == 16' failed.

MrSidims (approved):

PR #1950, opened 2023-04-11
[DebugInfo] Support multiple CompileUnits

MrSidims:
It's possible for LLVM IR module to contain multiple CU in case if this module is a result
of llvm-link between two modules compiled for different languages and or compiled with
different options.

This patch introduces handling of such modules.
std::unordered_map<SPIRVId, std::unique_ptr<DIBuilder>> BuilderMap was introduced
to SPIR-V consumption part and
std::unordered_map<const DICompileUnit *, SPIRVExtInst *> SPIRVCUMap was introduced
to SPIR-V generation part to preserve Scope relations between DI metadata in SPIR-V and vice versa.

Note, that DIBuilder has a single CU field and this class is not trivially copiable.

### DIFF
--- a/lib/SPIRV/LLVMToSPIRVDbgTran.cpp
+++ b/lib/SPIRV/LLVMToSPIRVDbgTran.cpp
@@ -52,11 +52,11 @@ void LLVMToSPIRVDbgTran::transDebugMetadata() {
   if (DIF.compile_unit_count() == 0)
     return;
 
-  DICompileUnit *CU = *DIF.compile_units().begin();
-  transDbgEntry(CU);
-
-  for (DIImportedEntity *IE : CU->getImportedEntities())
-    transDbgEntry(IE);
+  for (DICompileUnit *CU : DIF.compile_units()) {
+    transDbgEntry(CU);
+    for (DIImportedEntity *IE : CU->getImportedEntities())
+      transDbgEntry(IE);
+  }
 
   for (const DIType *T : DIF.types())
     transDbgEntry(T);
@@ -404,10 +404,9 @@ SPIRVType *LLVMToSPIRVDbgTran::getInt32Ty() {
 SPIRVEntry *LLVMToSPIRVDbgTran::getScope(DIScope *S) {
   if (S)
     return transDbgEntry(S);
-  else {
-    assert(SPIRVCU && "Compile unit is expected to be already translated");
-    return SPIRVCU;
-  }
+  assert(!SPIRVCUMap.empty() &&
+         "Compile units are expected to be already translated");
+  return SPIRVCUMap.begin()->second;
 }
 
 SPIRVEntry *LLVMToSPIRVDbgTran::getGlobalVariable(const DIGlobalVariable *GV) {
@@ -556,9 +555,9 @@ SPIRVEntry *LLVMToSPIRVDbgTran::transDbgCompileUnit(const DICompileUnit *CU) {
         Ops, {SPIRVDebugInfoVersionIdx, DWARFVersionIdx, LanguageIdx});
   BM->addModuleProcessed(SPIRVDebug::ProducerPrefix + CU->getProducer().str());
   // Cache CU in a member.
-  SPIRVCU = static_cast<SPIRVExtInst *>(
+  SPIRVCUMap[CU] = static_cast<SPIRVExtInst *>(
       BM->addDebugInfo(SPIRVDebug::CompilationUnit, getVoidTy(), Ops));
-  return SPIRVCU;
+  return SPIRVCUMap[CU];
 }
 
 // Types
@@ -985,7 +984,7 @@ SPIRVEntry *LLVMToSPIRVDbgTran::transDbgInheritance(const DIDerivedType *DT) {
 }
 
 SPIRVEntry *LLVMToSPIRVDbgTran::transDbgPtrToMember(const DIDerivedType *DT) {
-  using namespace SPIRVDebug::Operand::PtrToMember;
+  using namespace SPIRVDebug::Operand::TypePtrToMember;
   SPIRVWordVec Ops(OperandCount);
   Ops[MemberTypeIdx] = transDbgEntry(DT->getBaseType())->getId();
   Ops[ParentIdx] = transDbgEntry(DT->getClassType())->getId();
@@ -1084,7 +1083,7 @@ LLVMToSPIRVDbgTran::transDbgGlobalVariable(const DIGlobalVariable *GV) {
 
   // Parent scope
   DIScope *Context = GV->getScope();
-  SPIRVEntry *Parent = SPIRVCU;
+  SPIRVEntry *Parent = SPIRVCUMap.begin()->second;
   // Global variable may be declared in scope of a namespace or imported module,
   // it may also be a static variable declared in scope of a function.
   if (Context && (isa<DINamespace>(Context) || isa<DISubprogram>(Context) ||
@@ -1119,10 +1118,15 @@ SPIRVEntry *LLVMToSPIRVDbgTran::transDbgFunction(const DISubprogram *Func) {
   Ops[LineIdx] = Func->getLine();
   Ops[ColumnIdx] = 0; // This version of DISubprogram has no column number
   auto Scope = Func->getScope();
-  if (Scope && isa<DIFile>(Scope))
-    Ops[ParentIdx] = SPIRVCU->getId();
-  else
+  if (Scope && !isa<DIFile>(Scope)) {
     Ops[ParentIdx] = getScope(Scope)->getId();
+  } else {
+    if (auto *Unit = Func->getUnit())
+      Ops[ParentIdx] = SPIRVCUMap[Unit]->getId();
+    else
+      // it might so happen, that DISubprogram is missing Unit parameter
+      Ops[ParentIdx] = SPIRVCUMap.begin()->second->getId();
+  }
   Ops[LinkageNameIdx] = BM->getString(Func->getLinkageName().str())->getId();
   Ops[FlagsIdx] = adjustAccessFlags(Scope, transDebugFlags(Func));
   if (isNonSemanticDebugInfo())
--- a/lib/SPIRV/SPIRVToLLVMDbgTran.cpp
+++ b/lib/SPIRV/SPIRVToLLVMDbgTran.cpp
@@ -64,7 +64,7 @@ static uint64_t getDerivedSizeInBits(const DIType *Ty) {
 }
 SPIRVToLLVMDbgTran::SPIRVToLLVMDbgTran(SPIRVModule *TBM, Module *TM,
                                        SPIRVToLLVM *Reader)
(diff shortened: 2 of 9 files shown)


lib/SPIRV/SPIRVToLLVMDbgTran.cpp
```diff
@@ -80,8 +80,9 @@ DIFile *SPIRVToLLVMDbgTran::getDIFile(
     std::optional<DIFile::ChecksumInfo<StringRef>> CS) {
   return getOrInsert(FileMap, FileName, [=]() {
     SplitFileName Split(FileName);
+    // FIXME
     if (!Split.BaseName.empty())
-      return Builder.createFile(Split.BaseName, Split.Path, CS);
+      return BuilderMap.begin()->second->createFile(Split.BaseName, Split.Path, CS);
```
MrSidims, line 85: TODO: need to set correct CU

lib/SPIRV/SPIRVToLLVMDbgTran.cpp
```diff
@@ -154,17 +155,20 @@ SPIRVToLLVMDbgTran::transCompilationUnit(const SPIRVExtInst *DebugInst) {
 
   using namespace SPIRVDebug::Operand::CompilationUnit;
   assert(Ops.size() == OperandCount && "Invalid number of operands");
-  SPIRVWord DWARFVersion = getConstantValueOrLiteral(
-      Ops, DWARFVersionIdx, DebugInst->getExtSetKind());
-  M->addModuleFlag(llvm::Module::Max, "Dwarf Version", DWARFVersion);
+  if (!M->getModuleFlag("Dwarf Version")) {
```
MrSidims, line 158: TODO: figure out, what metadata generate here

lib/SPIRV/SPIRVToLLVMDbgTran.cpp
```diff
@@ -1326,6 +1331,21 @@ DIFile *SPIRVToLLVMDbgTran::getFile(const SPIRVId SourceId) {
...
+  assert(BuilderMap.size() != 0 && "No debug compile units");
+  if (BuilderMap.size() == 1)
+    return *BuilderMap.begin()->second;
+  while (DebugInst->getExtOp() != SPIRVDebug::CompilationUnit) {
+    if (DebugInst->getExtOp() == SPIRVDebug::DebugInfoNone)
+      return *BuilderMap.begin()->second;
+    if (SPIRVEntry *Scope = DebugInst->getDebugScope())
+      DebugInst = static_cast<SPIRVExtInst *>(Scope);
+    else
+      return *BuilderMap.begin()->second;
```
MrSidims, line 1344:
```suggestion
    if (SPIRVEntry *Scope = DebugInst->getDebugScope()) {
      DebugInst = static_cast<SPIRVExtInst *>(Scope);
      continue;
    }
    return *BuilderMap.begin()->second;
```
or something like this

maksimsab (changes requested):

lib/SPIRV/SPIRVToLLVMDbgTran.cpp
MrSidims: This Builder/CU is fine

lib/SPIRV/SPIRVToLLVMDbgTran.cpp
MrSidims: Embraced same behavior as LLVM

lib/SPIRV/libSPIRV/SPIRV.debug.h
```diff
@@ -830,6 +830,59 @@ enum {
...
+  case SPIRVDebug::ModuleINTEL:
+    ParentScopeIdx = ModuleINTEL::ParentIdx;
+    return true;
+  case SPIRVDebug::Module:
+    ParentScopeIdx = ModuleINTEL::ParentIdx;
+    return true;
+  default:
+    return false;
+  }
+  return false;
```
vmaksimo, line 883: Guess this return is unneeded since we have one in `default` case.

lib/SPIRV/SPIRVToLLVMDbgTran.cpp
```diff
@@ -1326,6 +1365,26 @@ DIFile *SPIRVToLLVMDbgTran::getFile(const SPIRVId SourceId) {
...
+  assert(BuilderMap.size() != 0 && "No debug compile units");
+  if (BuilderMap.size() == 1)
+    return *BuilderMap.begin()->second;
+  while (DebugInst->getExtOp() != SPIRVDebug::CompilationUnit) {
+    if (DebugInst->getExtOp() == SPIRVDebug::DebugInfoNone)
+      return *BuilderMap.begin()->second;
+    const SPIRVWordVec &Ops = DebugInst->getArguments();
+    SPIRVWord ParentScopeIdx = 0;
+    if (!hasDbgInstParentScopeIdx(DebugInst->getExtOp(), ParentScopeIdx))
+      return *BuilderMap.begin()->second;
```
vmaksimo, line 1378: Just to understand, do we always take first Builder by default because Instructions with no scope are related to the "default" Builder?

lib/SPIRV/SPIRVToLLVMDbgTran.cpp
```diff
@@ -1326,6 +1365,26 @@ DIFile *SPIRVToLLVMDbgTran::getFile(const SPIRVId SourceId) {
...
+    if (SPIRVEntry *Scope = BM->getEntry(Ops[ParentScopeIdx])) {
+      DebugInst = static_cast<SPIRVExtInst *>(Scope);
+      continue;
+    }
+    return *BuilderMap.begin()->second;
+  }
+  return *BuilderMap[DebugInst->getId()];
```
vmaksimo, line 1385: So, eventually we do map CompilationUnit Id (not general DebugInst) to the Builder pointer, right?

lib/SPIRV/SPIRVToLLVMDbgTran.cpp
```diff
@@ -1044,24 +1074,32 @@ DINode *SPIRVToLLVMDbgTran::transImportedEntry(const SPIRVExtInst *DebugInst) {
...
-      return Builder.createImportedDeclaration(Scope, GVE->getVariable(), File,
-                                               Line, Name);
-    return Builder.createImportedDeclaration(Scope, Entity, File, Line, Name);
+      return getDIBuilder(DebugInst).createImportedDeclaration(
+          Scope, GVE->getVariable(), File, Line, Name);
+    return getDIBuilder(DebugInst).createImportedDeclaration(Scope, Entity,
+                                                             File, Line, Name);
   }
-  llvm_unreachable("Unexpected kind of imported entity!");
+  // FIXME: uncomment and fix following line, with existing bugs it's reachable.
```
vmaksimo, line 1098: Should we create an issue/tracker for fixing the imported entity translation?

lib/SPIRV/LLVMToSPIRVDbgTran.cpp
```diff
@@ -404,10 +404,9 @@ SPIRVType *LLVMToSPIRVDbgTran::getInt32Ty() {
 SPIRVEntry *LLVMToSPIRVDbgTran::getScope(DIScope *S) {
   if (S)
     return transDbgEntry(S);
-  else {
-    assert(SPIRVCU && "Compile unit is expected to be already translated");
-    return SPIRVCU;
-  }
+  assert(!SPIRVCUMap.empty() &&
+         "Compile units are expected to be already translated");
+  return SPIRVCUMap.begin()->second;
```
vmaksimo, line 409: The same question for my understanding - why do we take the first CU by default for scope?

test/DebugInfo/Generic/c-and-cpp-mixed.ll
```diff
@@ -0,0 +1,116 @@
...
+; RUN: llvm-spirv -r -emit-opaque-pointers %t.spv -o - | llvm-dis -o %t.ll
+; RUN: FileCheck < %t.ll %s --check-prefix=CHECK-LLVM
+
+; CHECK-SPIRV: String [[#Foo:]] "foo"
+; CHECK-SPIRV: String [[#Main:]] "main"
+; CHECK-SPIRV: ExtInst [[#]] [[#CU1:]] [[#]] DebugCompilationUnit
+; CHECK-SPIRV: ExtInst [[#]] [[#CU2:]] [[#]] DebugCompilationUnit
+; CHECK-SPIRV: ExtInst [[#]] [[#Func1:]] [[#]] DebugFunction [[#Foo]] [[#]] [[#]] [[#]] [[#]] [[#CU1]]
+; CHECK-SPIRV: ExtInst [[#]] [[#Func2:]] [[#]] DebugFunction [[#Main]] [[#]] [[#]] [[#]] [[#]] [[#CU2]]
+; CHECK-SPIRV: ExtInst [[#]] [[#]] [[#]] DebugLexicalBlock [[#]] [[#]] [[#]] [[#Func1]]
```
vmaksimo, line 13: Do we really need to check for `DebugLexicalBlock`? Can't quite understand its semantics.

MrSidims: I'd prefer to add this check just to make sure, that we have a clean dependencies check from top layer Debug CU through Debug Function to bottom layer Debug info. Mainly such checks are done in CHECK-LLVM, but would like to have some in CHECK-SPIRV

lib/SPIRV/SPIRVToLLVMDbgTran.cpp
MrSidims: Builder here is being used to create dependencies between scope lowering them to CU scope. If we have no scope information - we translate such debug info with the best effort possible. If scope is just File - we actually don't really care, what builder to use. If CU is linked with a particular file, then debugger can link Debug info with the appropriate CU with the same File scope.

lib/SPIRV/SPIRVToLLVMDbgTran.cpp
MrSidims: Yes and then we try to get CU by looking at parent scope of a particular instruction

lib/SPIRV/SPIRVToLLVMDbgTran.cpp
MrSidims: > Should we create an issue/tracker for fixing the imported entity translation?

It's already created internally. It's called something like "ImportedEntity is missing 2nd operand". Can share it upon request.

vmaksimo (approved):

test/DebugInfo/Generic/c-and-cpp-mixed.ll
```diff
@@ -0,0 +1,116 @@
+; RUN: llvm-as < %s -o %t.bc
```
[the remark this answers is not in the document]
MrSidims: Done

lib/SPIRV/libSPIRV/SPIRV.debug.h
MrSidims: Applied

PR #1951, opened 2023-04-12
Preserve invalid SPIRV source lang literal in module metadata

michalpaszkowski:

I will talk about this change during the next OpenCL Tooling call.

### DIFF
--- a/lib/SPIRV/SPIRVToLLVMDbgTran.cpp
+++ b/lib/SPIRV/SPIRVToLLVMDbgTran.cpp
@@ -168,10 +168,17 @@ SPIRVToLLVMDbgTran::transCompilationUnit(const SPIRVExtInst *DebugInst) {
   }
   SPIRVWord SourceLang =
       getConstantValueOrLiteral(Ops, LanguageIdx, DebugInst->getExtSetKind());
-  if (DebugInst->getExtSetKind() == SPIRVEIS_NonSemantic_Shader_DebugInfo_200)
+  if (DebugInst->getExtSetKind() == SPIRVEIS_NonSemantic_Shader_DebugInfo_200) {
     SourceLang = convertSPIRVSourceLangToDWARFNonSemanticDbgInfo(SourceLang);
-  else
+  } else if (isSPIRVSourceLangValid(SourceLang)) {
     SourceLang = convertSPIRVSourceLangToDWARF(SourceLang);
+  } else {
+    // Some SPIR-V producers generate invalid source language value. In such
+    // case the original value should be preserved in "Source Lang Literal"
+    // module flag for later use by LLVM IR consumers.
+    M->addModuleFlag(llvm::Module::Warning, "Source Lang Literal", SourceLang);
+    SourceLang = dwarf::DW_LANG_OpenCL;
+  }
 
   BuilderMap[DebugInst->getId()] = std::make_unique<DIBuilder>(*M);
 
--- a/lib/SPIRV/libSPIRV/SPIRV.debug.h
+++ b/lib/SPIRV/libSPIRV/SPIRV.debug.h
@@ -918,6 +918,23 @@ inline spv::SourceLanguage convertDWARFSourceLangToSPIRV(dwarf::SourceLanguage D
   }
 }
 
+inline bool isSPIRVSourceLangValid(unsigned SourceLang) {
+  switch (SourceLang) {
+  // When updating this function, make sure to also
+  // update convertSPIRVSourceLangToDWARF()
+  case spv::SourceLanguage::SourceLanguageOpenCL_CPP:
+  case spv::SourceLanguage::SourceLanguageCPP_for_OpenCL:
+  case spv::SourceLanguage::SourceLanguageOpenCL_C:
+  case spv::SourceLanguage::SourceLanguageESSL:
+  case spv::SourceLanguage::SourceLanguageGLSL:
+  case spv::SourceLanguage::SourceLanguageHLSL:
+  case spv::SourceLanguage::SourceLanguageUnknown:
+    return true;
+  default:
+    return false;
+  }
+}
+
 inline dwarf::SourceLanguage convertSPIRVSourceLangToDWARF(unsigned SourceLang) {
   switch (SourceLang) {
   // When updating this function, make sure to also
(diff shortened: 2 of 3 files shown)


lib/SPIRV/SPIRVToLLVMDbgTran.cpp
```diff
@@ -159,10 +159,14 @@ SPIRVToLLVMDbgTran::transCompilationUnit(const SPIRVExtInst *DebugInst) {
   M->addModuleFlag(llvm::Module::Max, "Dwarf Version", DWARFVersion);
   SPIRVWord SourceLang =
       getConstantValueOrLiteral(Ops, LanguageIdx, DebugInst->getExtSetKind());
-  if (DebugInst->getExtSetKind() == SPIRVEIS_NonSemantic_Shader_DebugInfo_200)
+  if (DebugInst->getExtSetKind() == SPIRVEIS_NonSemantic_Shader_DebugInfo_200) {
     SourceLang = convertSPIRVSourceLangToDWARFNonSemanticDbgInfo(SourceLang);
-  else
+  } else {
+    if (!isSPIRVSourceLangValid(SourceLang))
```
MrSidims, line 165: like this we call `convertSPIRVSourceLangToDWARF` anyway, shouldn't we not call it if isSPIRVSourceLangValid = false?

MrSidims (commented):
I don't have strong opinion about the patch. Ideally we should stick to the OpenCL.DebugInfo.100 spec for such cases and don't generate undocumented languages, but we already have multiple workarounds.

lib/SPIRV/SPIRVToLLVMDbgTran.cpp
michalpaszkowski: > it seems that like this we call `convertSPIRVSourceLangToDWARF` anyway, shouldn't we not call it if isSPIRVSourceLangValid = false?

Right, then the call to `convertSPIRVSourceLangToDWARF` is really redundant. Thanks, fixed!

lib/SPIRV/SPIRVToLLVMDbgTran.cpp
```diff
@@ -159,10 +159,14 @@ SPIRVToLLVMDbgTran::transCompilationUnit(const SPIRVExtInst *DebugInst) {
...
   SPIRVWord SourceLang =
       getConstantValueOrLiteral(Ops, LanguageIdx, DebugInst->getExtSetKind());
-  if (DebugInst->getExtSetKind() == SPIRVEIS_NonSemantic_Shader_DebugInfo_200)
+  if (DebugInst->getExtSetKind() == SPIRVEIS_NonSemantic_Shader_DebugInfo_200) {
     SourceLang = convertSPIRVSourceLangToDWARFNonSemanticDbgInfo(SourceLang);
-  else
+  } else if (isSPIRVSourceLangValid(SourceLang)) {
     SourceLang = convertSPIRVSourceLangToDWARF(SourceLang);
+  } else {
+    M->addModuleFlag(llvm::Module::Warning, "Source Lang Literal", SourceLang);
```
asudarsa, line 167: Can we think about unconditionally adding this info? In that case, we do not need to worry about adding and maintaining changes in spirv.debug.h.

This is a non-blocking comment.

Thanks

michalpaszkowski: > Can we think about unconditionally adding this info? In that case, we do not need to worry about adding and maintaining changes in spirv.debug.h.
> 
> This is a non-blocking comment.
> 
> Thanks

Do you mean whether we could always generate the "Source Lang Literal" metadata node without checking `isSPIRVSourceLangValid(SourceLang)`? 
I personally think that reducing the amount of metadata when possible makes most sense. I will try to rework this and get rid of the check once we have the new debug info spec. Ideally this could go in another pull request.

michalpaszkowski:
@svenvh @MrSidims I have added a test as we discussed during the OpenCL Tooling call. Do you think the pull request could be merged as is?

MrSidims:
> @MrSidims Could you let me know here/offline where I can find some info about the new spec that you discussed during the call? I will try to see what we could do here from IGC side and hopefully address Arvind's comment in the next pull request.

It's placed here https://github.com/KhronosGroup/SPIRV-Registry/pull/186
But language table from there is already added and adopted by IGC

lib/SPIRV/libSPIRV/SPIRV.debug.h
```diff
@@ -850,10 +850,27 @@ inline spv::SourceLanguage convertDWARFSourceLangToSPIRV(dwarf::SourceLanguage D
...
+  default:
+    return false;
+  }
+}
+
 inline dwarf::SourceLanguage convertSPIRVSourceLangToDWARF(unsigned SourceLang) {
   switch (SourceLang) {
   // When updating this function, make sure to also
-  // update convertDWARFSourceLangToSPIRV()
+  // update convertDWARFSourceLangToSPIRV() and isSPIRVSourceLangValid()
```
MrSidims, line 873: I hope it's temporary
```suggestion
  // update convertDWARFSourceLangToSPIRV()
```

MrSidims (commented):
I'm OK with the patch but it's W/A to solve Intel-specific problem. So would prefer to hear from @StuartDBrady or @svenvh

lib/SPIRV/SPIRVToLLVMDbgTran.cpp
```diff
@@ -159,10 +159,14 @@ SPIRVToLLVMDbgTran::transCompilationUnit(const SPIRVExtInst *DebugInst) {
   M->addModuleFlag(llvm::Module::Max, "Dwarf Version", DWARFVersion);
   SPIRVWord SourceLang =
       getConstantValueOrLiteral(Ops, LanguageIdx, DebugInst->getExtSetKind());
-  if (DebugInst->getExtSetKind() == SPIRVEIS_NonSemantic_Shader_DebugInfo_200)
+  if (DebugInst->getExtSetKind() == SPIRVEIS_NonSemantic_Shader_DebugInfo_200) {
     SourceLang = convertSPIRVSourceLangToDWARFNonSemanticDbgInfo(SourceLang);
-  else
+  } else if (isSPIRVSourceLangValid(SourceLang)) {
     SourceLang = convertSPIRVSourceLangToDWARF(SourceLang);
+  } else {
```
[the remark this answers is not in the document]
michalpaszkowski: Agree, added the comment!

MrSidims (approved):
LGTM

svenvh (approved):

MrSidims:
@michalpaszkowski please resolve merge conflict

PR #1952, opened 2023-04-12
Remove assert statement to allow for multiple decorations on a given pointer

asudarsa:
This change tries to address corner cases where a single pointer can have multiple decorations associated with it.

Thanks

### DIFF
--- a/lib/SPIRV/SPIRVReader.cpp
+++ b/lib/SPIRV/SPIRVReader.cpp
@@ -1257,10 +1257,8 @@ static void replaceOperandWithAnnotationIntrinsicCallResult(Function *F,
     for (auto *Use : BV->users()) {
       if (auto *II = dyn_cast<IntrinsicInst>(Use)) {
         if (II->getIntrinsicID() == Intrinsic::ptr_annotation &&
-            II->getType() == BV->getType()) {
-          assert(CR == nullptr && "Multiple annotation created for same value");
+            II->getType() == BV->getType())
           CR = II;
-        }
       }
     }
     return CR ? true : false;


asudarsa:
Hi @broxigarchen and @MrSidims 

Can you please take a look at this?

Thanks

lib/SPIRV/SPIRVReader.cpp
```diff
@@ -1257,10 +1257,8 @@ static void replaceOperandWithAnnotationIntrinsicCallResult(Function *F,
     for (auto *Use : BV->users()) {
       if (auto *II = dyn_cast<IntrinsicInst>(Use)) {
         if (II->getIntrinsicID() == Intrinsic::ptr_annotation &&
-            II->getType() == BV->getType()) {
```
asudarsa, line 1260: The SPEC says: It is invalid to decorate any given <id> or structure member more than one time with the same [decoration](https://registry.khronos.org/SPIR-V/specs/unified1/SPIRV.html#Decoration), unless explicitly allowed below for a specific decoration
However, there is no restriction about decorating any given <id> with two different decorations. So, I think we can remove this assert.
Please provide feedback if the SPEC has not been interpreted correctly.
Thanks

MrSidims (approved):
Patch and explanation looks legit.

broxigarchen (approved):

lib/SPIRV/SPIRVReader.cpp
MrSidims: @broxigarchen @asudarsa is this thread resolved?

asudarsa: Yes. We agreed that this is something we may address in the future. But it does not block the current PR.

Thanks

PR #1956, opened 2023-04-13
[DebugInfo] Handle null value of DW_TAG_template_value_parameter

MrSidims:
It might be set as null in case if a function pointer is passed as auto template parameter.

The patch also adds a test for a 'good' DW_TAG_template_value_parameter value for function pointers, just because this case was untested.

### DIFF
--- a/test/DebugInfo/Generic/function-ptr-templ-null.ll
+++ b/test/DebugInfo/Generic/function-ptr-templ-null.ll
@@ -1,0 +1,107 @@
+;; compiled from (null added manually):
+;; template <auto Func>
+;; int foo() {
+;;   int result = Func();
+;;   return result;
+;; };
+;;
+;; long get() { return 42; }
+;;
+;; void boo() {
+;;   int val = foo<get>();
+;; }
+
+; REQUIRES: object-emission
+
+; RUN: llvm-as < %s -o %t.bc
+; RUN: llvm-spirv %t.bc -o %t.spv
+; RUN: llvm-spirv -r -emit-opaque-pointers %t.spv -o - | llvm-dis -o %t.ll
+; RUN: FileCheck < %t.ll %s --check-prefix=CHECK-LLVM
+
+; RUN: llc -mtriple=x86_64-linux -O0 -filetype=obj < %t.ll | llvm-dwarfdump -v -debug-info - | FileCheck %s --check-prefix=CHECK-DWARF
+
+; CHECK-LLVM: ![[#]] = !DITemplateValueParameter(name: "Func", type: ![[#Type:]], value: ptr null)
+; CHECK-LLVM: ![[#Type]] = !DIDerivedType(tag: DW_TAG_pointer_type, baseType: ![[#]], size: 64)
+
+; CHECK-DWARF: DW_TAG_subprogram
+; CHECK-DWARF: DW_AT_name{{.*}}"foo<&get>"
+
+; CHECK-DWARF: DW_TAG_template_value_parameter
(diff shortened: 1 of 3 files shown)


test/DebugInfo/Generic/function-ptr-templ-null.ll
```diff
@@ -0,0 +1,107 @@
...
+; REQUIRES: object-emission
+
+; RUN: llvm-as < %s -o %t.bc
+; RUN: llvm-spirv %t.bc -o %t.spv
+; RUN: llvm-spirv -r -emit-opaque-pointers %t.spv -o - | llvm-dis -o %t.ll
+; RUN: FileCheck < %t.ll %s --check-prefix=CHECK-LLVM
+
+; RUN: llc -mtriple=x86_64-linux -O0 -filetype=obj < %t.ll | llvm-dwarfdump -v -debug-info - | FileCheck %s --check-prefix=CHECK-DWARF
+
+; CHECK-LLVM: ![[#]] = !DITemplateValueParameter(name: "Func", type: ![[#Type:]], value: ptr null)
```
asudarsa, line 23: Input LLVM has: !26 = !DITemplateValueParameter(name: "Func", type: !27, value: null) and this is different from what is found in resulting LLVM. Do we need some changes in SPIRVReader.cpp to generate the same code?

Thanks

MrSidims: It's a good question and it's indeed a weakness of this patch. We could make a change in `transTypeTemplateParameter` from SPIRVToLLVMDbgTran.cpp . For example we can add a null constant check here https://github.com/KhronosGroup/SPIRV-LLVM-Translator/blob/main/lib/SPIRV/SPIRVToLLVMDbgTran.cpp#L972 and under this condition call
    return Builder.createTemplateValueParameter(Context, Name, Ty, false, nullptr);
to emit proper null.

Unfortunately it would break a legit case described in DebugInfo/Generic/constant-pointers.ll test.

Originally I was thinking to put id of DebugInfoNone instruction instead of ConstantNull, but it would mean, that (because of how `transTypeTemplateParameter`  works) `DITemplateValueParameter` would be transformed to `DITemplateTypeParameter`, which is less correct then current result.

Overall the current patch represents the closest we can get to 'the correct result for quite invalid debug info case' without breaking other features.

MrSidims: Just to clarify, that fact that we have 'null' as value for function pointers if FE bug in the first place.

MrSidims: also on dwarf level ptr null vs just null has no diff, see
```
DW_TAG_template_value_parameter [6]   (0x0000006f)
DW_AT_type [DW_FORM_ref4]     (cu + 0x00ab => {0x000000ab} "int (*)()")
DW_AT_name [DW_FORM_strp]     ( .debug_str[0x00000096] = "Func")
```
vs
```
DW_TAG_template_value_parameter [6]   (0x0000006f)
DW_AT_type [DW_FORM_ref4]     (cu + 0x00ab => {0x000000ab} "int (*)()")
DW_AT_name [DW_FORM_strp]     ( .debug_str[0x0000009b] = "Func")
```

so the only diff is in string address which can be expected

asudarsa: LGTM. Thanks

asudarsa (approved):
LGTM overall. Thanks Dmitry.

PR #1957, opened 2023-04-13
[OpaquePointers] Rewrite the type scavenger.

jcranmer-intel:
This is a massive change, which consists of the following pieces that are not so easily seperable (why this is a large, single commit instead of many smaller ones):
* Use typed pointer types in the external interface of the scavenger.
* Use target extension types in lieu of DeferredType to represent types not yet known. These types are now target("typevar", N), where N is some integer.
* Use TypedPointerType in lieu of DeducedType. This makes multi-level pointer types simpler to use.
* Replace std::pair<unsigned, DeducedType> with a more generalized TypeRule class, which simplifies some of the rules for the typing procedure.
* Make computePointerElementType and correctUseTypes both rely on the same list of type rules, removing duplication.
* Add support for vectors of pointers, arrays of pointers, and function pointers in the type scavenger.
* Allow global variables to have scavenged types.
* Allow return types to not be i8*.
* Add more documentation of the type scavenger process.
* More tests!
* Miscellaneous type rule fixes.

Overall, the logic of the type scavenger largely remains unchanged (albeit with increased functionality), but the way that logic is expressed has changed very significantly, and the easier functionality for handling a few cases causes shifting of expected bitcast placements in existing tests.

### DIFF
--- a/lib/SPIRV/SPIRVTypeScavenger.cpp
+++ b/lib/SPIRV/SPIRVTypeScavenger.cpp
@@ -33,8 +33,85 @@
 //===----------------------------------------------------------------------===//
 //
 // This file implements the necessary logic to recover pointer types from LLVM
-// IR for the output SPIR-V file after LLVM IR completes its transition to
-// opaque pointers.
+// IR for the output SPIR-V file from the opaque pointers in LLVM IR.
+//
+// The core algorithm being implemented is rather simple, although there are
+// several complications that make its implementation more difficult. At its
+// core, the algorithm work like this:
+//
+// 1. Replace every instance of an opaque pointer type with a typed pointer that
+//    points to an unknown type variable.
+// 2. Convert each instruction into a series of typing rules. For example,
+//    load i8, ptr %ptr implies that %ptr must be typedptr(i8, 0).
+// 3. Based on the typing rules, resolve the type variables to concrete types.
+// 4. If the typing rules produce a contradiction (e.g., i8 == i32), insert a
+//    synthetic bitcast to represent the bitcast that would have been present in
+//    a typed pointer IR.
+// 5. If any type variables are unresolved at the end of the typing process,
+//    assign i8 to them instead.
+//
+// Typed pointers are represented with the TypedPointerType. Type variables are
+// represented as target("typevar", i), where i is an integer.
+//
+// Step 3 of the above algorithm is represented by unifyType, which implements a
+// unification-based type algorithm. This means there exists essentially just
+// four cases that need to be considered:
+// * unify(type var, concrete type):
+//   In this case, the concrete type is assigned to the type variable
+//   Note: It is possible for concrete type to contain nested type variables,
+//         e.g., typedptr(target("typevar", 3), 4)
+// * unify(type var, concrete type containing type var):
+//   Unification fails in this case. This can come up if you have code like
+//   this:
+//     %ptr = alloca ptr
+//     store ptr %ptr, ptr %ptr
+// * unify(type var, type var):
+//   In this case, the two type variables are unified into one so that they get
+//   the same concrete type. This uses IntEqClasses as the implementation of the
+//   union-find data structure.
+// * unify(concrete type, same concrete type):
+//   There is nothing to do in this case
+// * unify(concrete type, different concrete type):
+//   Unification fails in this case, and a bitcast needs to be generated.
+//
+// Note that this algorithm does not attempt to seek a minimal set of bitcasts
+// that need to be added to produce a correctly-typed program.
+//
+// Type rules are represented by the SPIRVTypeScavenger::TypeRule class, and
+// should be constructed using the provided static methods (which are easier to
+// understand than the constructor itself). Type rules boil down to the
+// following categories:
+// * operand I has type T
+// * operand I has the same type as operand J
+// * the return value has type T
+// * the return value has the same type as operand I
+// with any of the above operands, types, or return values potentially having a
+// level of indirection. For example, the rule for an addrspacecast is that the
+// return value points to the same type that its sole operand points to. The
+// indirection effectively means T->getScalarType()->getPointerElementType().
+// Note: When constructing type rules fixing an operand or the return to a
+//       particular type T, the type must be a type using typed pointers and/or
+//       type variables in lieu of ptr.
+// Note: getTypeRules may be called twice on an instruction, so if a new type
+//       variable needs to be created for type rules, it needs to be saved in
+//       the AssociatedTypeVariables method to ensure proper functioning. This
+//       is particularly important if you need to use the type variable both in
+//       constraining the return value and an operand.
+//
+// Now for the complications to the above algorithm:
+//
+// The most notable issue is that LLVM does not allow no-op constant
+// expressions to be created. This means that we have to be very careful about
+// the types of constant values. As the SPIR-V translator expands most constant
+// expressions into instructions, this isn't much of an issue, but where it
+// really comes into play is with global variable initializers, which don't have
+// that luxury. Global variable typing therefore pays careful attention to the
+// type of the initializer.
+//
+// Constructing type variables is a slightly expensive step, so before
+// attempting to create a type variable for the return of an instruction, we
+// instead look through the type rules to see if we can get the type of the
+// return value from one of its input operands.
 //
 //===----------------------------------------------------------------------===//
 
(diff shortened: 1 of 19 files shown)


test/extensions/INTEL/SPV_INTEL_fpga_buffer_location/FPGABufferLocation.ll
```diff
@@ -79,7 +79,7 @@ entry:
 ; CHECK-LLVM: %[[INTRINSIC_CALL:[[:alnum:].]+]] = call ptr addrspace(4) @llvm.ptr.annotation.p4.p0(ptr addrspace(4) %a, ptr @[[ANN_STR]], ptr undef, i32 undef, ptr undef)
 ; CHECK-LLVM: %[[BITCAST_CALL1:[[:alnum:].]+]] = bitcast ptr addrspace(4) %[[INTRINSIC_CALL]] to ptr addrspace(4)
 ; CHECK-LLVM: %[[BITCAST_CALL2:[[:alnum:].]+]] = bitcast ptr addrspace(4) %[[BITCAST_CALL1]] to ptr addrspace(4)
-; CHECK-LLVM: store ptr addrspace(4) %4, ptr addrspace(4) %[[BITCAST_CALL2]], align 8
+; CHECK-LLVM: store ptr addrspace(4) %2, ptr addrspace(4) %[[BITCAST_CALL2]], align 8
```
MrSidims, line 82: just to not change the test every time. I assume %2 is not checked later on, right?
```suggestion
; CHECK-LLVM: store ptr addrspace(4) %[[#]], ptr addrspace(4) %[[BITCAST_CALL2]], align 8
```

test/transcoding/OpSwitch32.ll
```diff
@@ -30,7 +30,7 @@ target triple = "spir64-unknown-unknown"
 
 ;CHECK-LLVM: test_32
 ;CHECK-LLVM: entry
-;CHECK-LLVM: switch i32 %7, label %sw.epilog
+;CHECK-LLVM: switch i32 %6, label %sw.epilog
```
MrSidims, line 33: While we are here:, %6 is not checked anywhere afterwards 
```suggestion
;CHECK-LLVM: switch i32 %[[#]], label %sw.epilog
```

test/transcoding/OpSwitch64.ll
```diff
@@ -33,7 +33,7 @@ target triple = "spir64-unknown-unknown"
 
 ;CHECK-LLVM: test_64
 ;CHECK-LLVM: entry
-;CHECK-LLVM: switch i64 %7, label %sw.epilog [
+;CHECK-LLVM: switch i64 %6, label %sw.epilog [
```
MrSidims, line 36: While we are here:, %6 is not checked anywhere afterwards 
```suggestion
;CHECK-LLVM: switch i64 %[[#]], label %sw.epilog [
```

test/transcoding/OpSwitchChar.ll
```diff
@@ -29,7 +29,7 @@ target datalayout = "e-p:32:32-i64:64-v16:16-v24:32-v32:32-v48:64-v96:128-v192:2
 target triple = "spir-unknown-unknown"
 
 ;CHECK-LLVM-LABEL: @test_switch
-;CHECK-LLVM: switch i8 %1, label %sw.epilog
+;CHECK-LLVM: switch i8 %0, label %sw.epilog
```
MrSidims, line 32:
```suggestion
;CHECK-LLVM: switch i8 %[[#]], label %sw.epilog
```

MrSidims (commented):
Description and the tests look reasonable. I didn't really understand the following bullet though:
`Use target extension types in lieu of DeferredType to represent types not yet known. These types are now target("typevar", N), where N is some integer.`

But I do believe I'll understand it while looking at the code, which will do through the week.

Some nits to the tests added (not related to the patch - just tests were originally written not in the best way).

dwoodwor-intel (approved):
The code here looks good besides the test brittleness; I just have a few very minor comments:

lib/SPIRV/SPIRVTypeScavenger.cpp
```diff
@@ -33,8 +33,85 @@
...
+//    load i8, ptr %ptr implies that %ptr must be typedptr(i8, 0).
+// 3. Based on the typing rules, resolve the type variables to concrete types.
+// 4. If the typing rules produce a contradiction (e.g., i8 == i32), insert a
+//    synthetic bitcast to represent the bitcast that would have been present in
+//    a typed pointer IR.
+// 5. If any type variables are unresolved at the end of the typing process,
+//    assign i8 to them instead.
+//
+// Typed pointers are represented with the TypedPointerType. Type variables are
+// represented as target("typevar", i), where i is an integer.
```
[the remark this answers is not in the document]
MrSidims: I have the same question actually. Is it address space?

lib/SPIRV/SPIRVTypeScavenger.cpp
```diff
@@ -53,156 +130,498 @@
 
 using namespace llvm;
 
+namespace {
+static bool isTypeVariable(Type *T, unsigned &TypeVarNum) {
```
MrSidims, line 134: nit: seems like a good function to add `inline`

lib/SPIRV/SPIRVTypeScavenger.cpp
```diff
@@ -53,156 +130,498 @@
...
+}
+
+/// Get a type where all internal pointer types are replaced with i8*.
+Type *getUnknownTyped(Type *T) {
+  Type *Int8Ty = Type::getInt8Ty(T->getContext());
+  return mutateType(
+      T, [=](unsigned AS) { return TypedPointerType::get(Int8Ty, AS); });
+}
+
+bool hasTypeVariable(Type *T, unsigned TypeVarNum) {
```
MrSidims, line 208: More for readers then for the compiler, since for example isTypeVariable modifies the CheckNum which is used in the same context.
```suggestion
bool hasTypeVariable(Type *T, const unsigned TypeVarNum) {
```

test/type-scavenger/call.ll
```diff
@@ -0,0 +1,38 @@
+; RUN: llvm-as %s -o %t.bc
+; RUN: llvm-spirv %t.bc -spirv-text -o - | FileCheck %s
+; RUN: llvm-spirv %t.bc -o %t.spv
+; RUN: spirv-val %t.spv
+
+target datalayout = "e-p:32:32-i64:64-v16:16-v24:32-v32:32-v48:64-v96:128-v192:256-v256:256-v512:512-v1024:1024"
+target triple = "spir-unknown-unknown"
+
+; CHECK: Name [[CALL:[0-9]+]] "call"
+; CHECK: 4 TypeInt [[INT:[0-9]+]] 32 0
```
MrSidims, line 10: nit here and below: no need to check for number of SPIRV words
```suggestion
; CHECK: TypeInt [[INT:[0-9]+]] 32 0
```

lib/SPIRV/SPIRVTypeScavenger.cpp
```diff
@@ -53,156 +130,498 @@
...
+    ActualTy = allocateTypeVariable(ActualTy);
+
+    // Check for recursion in type variables. Such recursive types generally
+    // cannot be correctly typed.
+    if (hasTypeVariable(ActualTy, Leader))
+      return false;
+
+    LLVM_DEBUG(dbgs() << "Type variable " << TypeVarNum << " is " << *ActualTy
+                      << "\n");
+    assert(!TypeVariables[Leader] && "Type was already fixed?");
```
MrSidims, line 285: Did we resize `TypeVariables` somewhere? If not, doesn't it lead to our-of-bound access? I'm not sure if it will be nullptr in such case.

MrSidims (commented):
Very nice explanation in the comments, many thanks for this!
There is first set of comments. May be some of them are false alarms. I'd also like to get answer for the role of 'i' for `typevar`, it should help with understanding.

lib/SPIRV/SPIRVTypeScavenger.cpp
```diff
@@ -113,6 +115,11 @@
 // instead look through the type rules to see if we can get the type of the
 // return value from one of its input operands.
 //
+// Recursive pointer types are not possible to express, as we make no attempt
```
MrSidims, line 118: Got it, thanks!

lib/SPIRV/SPIRVTypeScavenger.cpp
```diff
@@ -238,57 +683,32 @@ void SPIRVTypeScavenger::deduceFunctionType(Function &F) {
...
-          DeducedTypes[Arg] = PointeeTy;
-          LLVM_DEBUG(dbgs()
-                     << "Arg " << Arg->getArgNo() << " of " << F.getName()
-                     << " has type " << *PointeeTy << "\n");
+  // If the function is a mangled name, try to recover types from the Itanium
+  // name mangling. Do this only for function types that without bodies, where
+  // existing code can propagate types to the parameters.
+  // TODO: Investigate if target extension types and the specially-handled
+  // SPIR-V intrinsics renders this code unnecessary.
+  if (F.isDeclaration() && F.getName().startswith("_Z")) {
```
MrSidims, line 697: Do we have a test for this code?

lib/SPIRV/SPIRVTypeScavenger.cpp
```diff
@@ -53,156 +137,497 @@
...
+    case OpAtomicSMin:
+    case OpAtomicUMin:
+    case OpAtomicFMinEXT:
+    case OpAtomicSMax:
+    case OpAtomicUMax:
+    case OpAtomicFMaxEXT:
+    case OpAtomicAnd:
+    case OpAtomicOr:
+    case OpAtomicXor:
+      TypeRules.push_back(TypeRule::pointsTo(CB, 0, CB.getType()));
```
MrSidims, line 463: nit here and after for TypeRule::pointsTo because I'm not sure if it's 'easy' to copy PointerUnion without overhead (please share if you have another data)
```suggestion
      TypeRules.emplace_back(TypeRule::pointsTo(CB, 0, CB.getType()));
```

MrSidims:
In general LGTM

MrSidims (approved):

PR #1961, opened 2023-04-19
[DebugInfo] Support translation of DebugFunctionDefinition instruction

vmaksimo:
`DebugFunction` does not have an `Function Id` operand in NonSemantic.Shader debug info specification. It's been replaced by the whole new `DebugFunctionDefinition` instruction to avoid forward references.

This instruction must appear in the entry basic block of an OpFunction.

Specification:
https://github.com/KhronosGroup/SPIRV-Registry/blob/main/nonsemantic/NonSemantic.Shader.DebugInfo.100.asciidoc#DebugFunctionDefinition

### DIFF
--- a/lib/SPIRV/LLVMToSPIRVDbgTran.cpp
+++ b/lib/SPIRV/LLVMToSPIRVDbgTran.cpp
@@ -1133,6 +1133,7 @@ SPIRVEntry *LLVMToSPIRVDbgTran::transDbgFunction(const DISubprogram *Func) {
     transformToConstant(Ops, {LineIdx, ColumnIdx, FlagsIdx});
 
   SPIRVEntry *DebugFunc = nullptr;
+  SPIRVValue *FuncDef = nullptr;
   if (!Func->isDefinition()) {
     DebugFunc =
         BM->addDebugInfo(SPIRVDebug::FunctionDeclaration, getVoidTy(), Ops);
@@ -1150,9 +1151,14 @@ SPIRVEntry *LLVMToSPIRVDbgTran::transDbgFunction(const DISubprogram *Func) {
         SPIRVValue *SPIRVFunc = SPIRVWriter->getTranslatedValue(&F);
         assert(SPIRVFunc && "All function must be already translated");
         Ops[FunctionIdIdx] = SPIRVFunc->getId();
+        FuncDef = SPIRVFunc;
         break;
       }
     }
+    // For NonSemantic.Shader.DebugInfo we store Function Id index as a
+    // separate DebugFunctionDefinition instruction.
+    if (isNonSemanticDebugInfo())
+      Ops.pop_back();
 
     if (DISubprogram *FuncDecl = Func->getDeclaration())
       Ops.push_back(transDbgEntry(FuncDecl)->getId());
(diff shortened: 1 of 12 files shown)


vmaksimo:
There has been discovered a bug, that after "second" translation to SPIR-V (e.g. first one is `*.bc -o *.spv` and the second one is `*.spv -to-text -o *.spt`), we get the same `DebugFunctionDefinition` two times in a module. 
One is the correct one, inside the function basic block; and the second one is incorrect - the instruction appears in the list of Debug instructions section.

The issue is seen both for `-to-text` and for `-to-binary` translation, but the translation to binary is failing with the assertion.
Trying to investigate the possible solutions to the problem. Possibly could be addressed in a separate PR.

lib/SPIRV/LLVMToSPIRVDbgTran.cpp
```diff
@@ -1150,9 +1151,14 @@ SPIRVEntry *LLVMToSPIRVDbgTran::transDbgFunction(const DISubprogram *Func) {
...
         assert(SPIRVFunc && "All function must be already translated");
         Ops[FunctionIdIdx] = SPIRVFunc->getId();
+        FuncDef = SPIRVFunc;
         break;
       }
     }
+    // For NonSemantic.Shader.DebugInfo we store Function Id index as a
+    // separate DebugFunctionDefinition instruction.
+    if (isNonSemanticDebugInfo())
+      Ops.pop_back();
```
MrSidims, line 1161: I guess, because of SPIRVToLLVMDbgTran a cleaner solution of removing FunctionIdIdx from SPIRV.debug.h is quite hard. Yet, I'd suggest to a little improvement here: move resize of the vector to line 1147 and do `for (const llvm::Function &F : M->functions()) {` loop is else branch.

test/DebugInfo/NonSemantic/DebugFunction.cl
```diff
@@ -4,7 +4,7 @@
 // - Parent operand of DebugFunction is DebugCompilationUnit, not an OpString,
 //   even if in LLVM IR it points to a DIFile instead of DICompileUnit.
 
-// RUN: %clang_cc1 %s -cl-std=clc++ -emit-llvm-bc -triple spir -debug-info-kind=line-tables-only -O0 -o - | llvm-spirv -o %t.spv
+// RUN: %clang_cc1 %s -cl-std=clc++ -emit-llvm-bc -triple spir -debug-info-kind=line-tables-only -O0 -o - | llvm-spirv --spirv-debug-info-version=nonsemantic-shader-100 -o %t.spv
```
MrSidims, line 7: Please add new RUN line with --spirv-debug-info-version=nonsemantic-shader-200

lib/SPIRV/LLVMToSPIRVDbgTran.cpp
MrSidims: I'm wrong

test/DebugInfo/NonSemantic/DebugFunction.cl
MrSidims: ah, it's my bad that I didn't add it earlier in the first place. anyway, please add it :)

MrSidims (approved):
LGTM with the test nit
to-text / to-binary can be resolved later if you want

MrSidims (approved):

PR #1965, opened 2023-04-20
[DebugInfo] Add Flag parameter to DebugTypeBasic

MrSidims:
It can only be FlagUnknownPhysicalLayout. There is no way we can generate it LLVM in environment and get use of it, hence the patch just ignores it if it come from another SPIR-V generator.

In general, there are following possible debug flags for DIBasicType: BigEndian, LittleEndian and Artificial. There is no way that clang will ever generate them, but they can be produced by manually writing assembly and transforming it to LLVM IR. While it can be potential improvement for the future - I don't see it useful to add to the spec and implementation right now.

Spec:
https://github.com/KhronosGroup/SPIRV-Registry/blob/main/nonsemantic/NonSemantic.Shader.DebugInfo.100.asciidoc


MrSidims:

lib/SPIRV/SPIRVToLLVMDbgTran.cpp
```diff
@@ -181,7 +181,9 @@ SPIRVToLLVMDbgTran::transCompilationUnit(const SPIRVExtInst *DebugInst) {
 DIBasicType *SPIRVToLLVMDbgTran::transTypeBasic(const SPIRVExtInst *DebugInst) {
   using namespace SPIRVDebug::Operand::TypeBasic;
   const SPIRVWordVec &Ops = DebugInst->getArguments();
-  assert(Ops.size() == OperandCount && "Invalid number of operands");
+  assert((Ops.size() == OperandCountOCL ||
+          Ops.size() == OperandCountNonSemantic)
+         && "Invalid number of operands");
```
vmaksimo, line 186: Ok, even if we do nothing for these flags in forward translation (can we produce some default value to be spec conformant?), I believe we still should add handling of flags in reverse, especially knowing that `createBasicType` does have Flags argument. Does it make sense?

MrSidims: Unfortunately - no. Default flag is no flag aka DINode::FlagZero (it's already default parameter of createBasicType in DIBuilder).

There is no 'unknown physical layout' base type entity in C, C++, Fortran and GLSL. There we can only have structures, that usually have hardware dependent physical layout (aka unknown for this compilation stage). Also there is a close thing in Rust called 'unsafe', but TBH I didn't check what Rust FE with -g would generate for it.

Also one create 'pseudo base type' with unknown layout in assembly (not our case) and may be in shaders, but I didn't try it (anyway it's not LLVM IR).

test/DebugInfo/NonSemantic/basic-type-with-flag.spt
```diff
@@ -0,0 +1,96 @@
...
+4 Constant 3 28 6
+4 Constant 3 31 32
+4 Constant 3 32 4
+4 Constant 3 37 1
+4 Constant 3 38 2
+4 Constant 3 41 4294967295
+4 Constant 3 46 3
+4 Constant 3 47 136
+4 Constant 3 49 16
+4 Constant 3 56 131072
```
vmaksimo, line 51: Nit: maybe to explicitly point out that this line is `FlagUnknownPhysicalLayout`

lib/SPIRV/SPIRVToLLVMDbgTran.cpp
vmaksimo: Please fix clang-format here

vmaksimo (approved):
Sounds a bit complicated to investigate the purpose of this change in future, but I'm OK with merging it (after fixing clang-format).

vmaksimo (approved):

PR #1966, opened 2023-04-21
Update for llvm::StringView removal

svenvh:
Fix the build after llvm-project commit 7c59e8001a3b ("Reland: [Demangle] replace use of llvm::StringView w/ std::string_view", 2023-04-20).

PR #1967, opened 2023-04-21
Fix build after llvm-project@7c59e80

vmaksimo:
`llvm::StringView` is replaced with `std::string_view` in LLVM lib


vmaksimo:
Closing in favor of #1966 (@svenvh is faster again :))

svenvh:
> @svenvh is faster again :))

Haha, only by a minute though! ;-)

PR #1968, opened 2023-04-21
[DebugInfo] Add module producer info for NonSemantic.Shader.DebugInfo.200

vmaksimo:
Keep existing W/A (emit `ModuleProcessed` instruction - introduced by 601e4a5) for other debug info specs, except NonSemantic.Shader.DebugInfo.100

### DIFF
--- a/lib/SPIRV/LLVMToSPIRVDbgTran.cpp
+++ b/lib/SPIRV/LLVMToSPIRVDbgTran.cpp
@@ -540,7 +540,7 @@ SPIRVId LLVMToSPIRVDbgTran::getDebugInfoNoneId() {
 
 SPIRVEntry *LLVMToSPIRVDbgTran::transDbgCompileUnit(const DICompileUnit *CU) {
   using namespace SPIRVDebug::Operand::CompilationUnit;
-  SPIRVWordVec Ops(OperandCount);
+  SPIRVWordVec Ops(MinOperandCount);
   Ops[SPIRVDebugInfoVersionIdx] = SPIRVDebug::DebugInfoVersion;
   Ops[DWARFVersionIdx] = M->getDwarfVersion();
   Ops[SourceIdx] = getSource(CU)->getId();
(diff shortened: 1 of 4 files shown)


lib/SPIRV/LLVMToSPIRVDbgTran.cpp
```diff
@@ -553,7 +553,14 @@ SPIRVEntry *LLVMToSPIRVDbgTran::transDbgCompileUnit(const DICompileUnit *CU) {
...
         Ops, {SPIRVDebugInfoVersionIdx, DWARFVersionIdx, LanguageIdx});
-  BM->addModuleProcessed(SPIRVDebug::ProducerPrefix + CU->getProducer().str());
+
+  if (BM->getDebugInfoEIS() == SPIRVEIS_NonSemantic_Shader_DebugInfo_200) {
+    Ops.push_back(BM->getString(CU->getProducer().str())->getId());
+  } else if (BM->getDebugInfoEIS() !=
+             SPIRVEIS_NonSemantic_Shader_DebugInfo_100) {
+    BM->addModuleProcessed(SPIRVDebug::ProducerPrefix +
+                           CU->getProducer().str());
+  }
```
MrSidims, line 563: This can bring unexpected problems for theoretical SPIRVEIS_NonSemantic_Shader_DebugInfo_300 built on top of 100 and not inheriting features from 200. I suggest to check for OpenCL.DebugInfo.100 explicitly instead of BM->getDebugInfoEIS() !=
             SPIRVEIS_NonSemantic_Shader_DebugInfo_100
             
please also leave TODO to remove old W/A

MrSidims (commented):
LGTM just one nit

lib/SPIRV/LLVMToSPIRVDbgTran.cpp
vmaksimo: We also have "SPIRV.debug" and "OpenCL.std" debug info. Maybe we should check that the debug info is not NonSemantic (version 200 will be checked twice, but it'd help to avoid issues with the next versions of non-semantic spec)

MrSidims: >> Maybe we should check that the debug info is not NonSemantic

This should also work, actually
if (NonSemantic) {
if (200) {
// add producer
} // 200
} // NonSemantic else {
 add W/A
 }

vmaksimo: Exactly what I meant :) will add shortly

MrSidims (approved):

PR #1969, opened 2023-04-21
[NFC] Use typed pointers for null function pointers debug info

MrSidims:
`PointerType::get(Context` will return typeless/opaque pointer. While it is OK for KHR translator since it's being built on top of LLVM trunk, where opaque pointers are enabled by default - it won't work for intel/llvm, where their generation is disabled.

Upstream of https://github.com/intel/llvm/pull/9118


MrSidims:
@vmaksimo @maksimsab please take a look

vmaksimo (approved):
Guess we don't need a test for such change?

MrSidims:
> Guess we don't need a test for such change?

we can't added until we want to create new CI action to build customized LLVM without opaque pointers.

PR #1973, opened 2023-04-25
[DebugInfo] Support translation of DebugEntryPoint instruction

vmaksimo:
This instruction is generated for DWARF `DISPFlagMainSubprogram` flag of function as well as for `spir_kernel` functions.

Spec:
https://github.com/KhronosGroup/SPIRV-Registry/blob/main/nonsemantic/NonSemantic.Shader.DebugInfo.100.asciidoc#DebugEntryPoint

### DIFF
--- a/lib/SPIRV/LLVMToSPIRVDbgTran.cpp
+++ b/lib/SPIRV/LLVMToSPIRVDbgTran.cpp
@@ -1218,19 +1218,16 @@ SPIRVEntry *LLVMToSPIRVDbgTran::transDbgFunction(const DISubprogram *Func) {
 
   if (isNonSemanticDebugInfo() &&
       (Func->isMainSubprogram() || IsEntryPointKernel))
-    transDbgEntryPoint(Func, DebugFunc);
+    [[maybe_unused]] SPIRVEntry *Inst = transDbgEntryPoint(Func, DebugFunc);
 
-  if (isNonSemanticDebugInfo())
-    transDbgFuncDefinition(FuncDef, DebugFunc);
+  if (isNonSemanticDebugInfo() && FuncDef)
+    [[maybe_unused]] SPIRVEntry *Inst = transDbgFuncDefinition(FuncDef, DebugFunc);
 
   return DebugFunc;
 }
 
 SPIRVEntry *LLVMToSPIRVDbgTran::transDbgFuncDefinition(SPIRVValue *FuncDef,
                                                        SPIRVEntry *DbgFunc) {
-  if (!isNonSemanticDebugInfo() || !FuncDef)
-    return nullptr;
-
   using namespace SPIRVDebug::Operand::FunctionDefinition;
   SPIRVWordVec Ops(OperandCount);
   Ops[FunctionIdx] = DbgFunc->getId();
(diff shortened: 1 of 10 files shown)


vmaksimo:

lib/SPIRV/LLVMToSPIRVDbgTran.cpp
```diff
@@ -1196,6 +1216,10 @@ SPIRVEntry *LLVMToSPIRVDbgTran::transDbgFunction(const DISubprogram *Func) {
     DebugFunc = transDbgTemplateParams(TPA, DebugFunc);
   }
 
+  if (isNonSemanticDebugInfo() &&
+      (Func->isMainSubprogram() || IsEntryPointKernel))
+    transDbgEntryPoint(Func, DebugFunc);
```
MrSidims, line 1221: I just realized, that for this case and for the case below - transDbgFuncDefinition, static analyzer tools will complain, that result of the functions is unused. Can we make these two functions returning void? (can be done it this PR)

lib/SPIRV/LLVMToSPIRVDbgTran.cpp
```diff
@@ -1219,6 +1243,28 @@ SPIRVEntry *LLVMToSPIRVDbgTran::transDbgFuncDefinition(SPIRVValue *FuncDef,
                         Ops, BB, BB->getInst(0));
 }
 
+SPIRVEntry *LLVMToSPIRVDbgTran::transDbgEntryPoint(const DISubprogram *Func,
+                                                   SPIRVEntry *DbgFunc) {
+  if (!isNonSemanticDebugInfo())
```
MrSidims, line 1248: We've already checked isNonSemanticDebugInfo before

lib/SPIRV/LLVMToSPIRVDbgTran.cpp
```diff
@@ -1219,6 +1243,28 @@ SPIRVEntry *LLVMToSPIRVDbgTran::transDbgFuncDefinition(SPIRVValue *FuncDef,
...
+  SPIRVWordVec Ops(SPIRVDebug::Operand::EntryPoint::OperandCount);
+
+  DICompileUnit *CU = Func->getUnit();
+  StringRef Producer = CU->getProducer();
+  StringRef Flags = CU->getFlags();
+
+  auto It = MDMap.find(CU);
+  SPIRVEntry *CUVal = It != MDMap.end() ? It->second : getDebugInfoNone();
+
+  using namespace SPIRVDebug::Operand::EntryPoint;
```
MrSidims, line 1260: nit:
can be moved before
`SPIRVWordVec Ops(SPIRVDebug::Operand::EntryPoint::OperandCount);`

lib/SPIRV/LLVMToSPIRVDbgTran.cpp
```diff
@@ -1219,6 +1243,28 @@ SPIRVEntry *LLVMToSPIRVDbgTran::transDbgFuncDefinition(SPIRVValue *FuncDef,
...
+    return nullptr;
+
+  SPIRVWordVec Ops(SPIRVDebug::Operand::EntryPoint::OperandCount);
+
+  DICompileUnit *CU = Func->getUnit();
+  StringRef Producer = CU->getProducer();
+  StringRef Flags = CU->getFlags();
+
+  auto It = MDMap.find(CU);
+  SPIRVEntry *CUVal = It != MDMap.end() ? It->second : getDebugInfoNone();
```
MrSidims, line 1258: We can probably get use of brand-new SPIRVCUMap

lib/SPIRV/SPIRVToLLVMDbgTran.cpp
```diff
@@ -909,6 +912,18 @@ DINode *SPIRVToLLVMDbgTran::transFunctionDecl(const SPIRVExtInst *DebugInst) {
   return DIS;
 }
 
+MDNode *SPIRVToLLVMDbgTran::transEntryPoint(const SPIRVExtInst *DebugInst) {
+  if (!isNonSemanticDebugInfo(DebugInst->getExtSetKind()))
```
MrSidims, line 916: Is it a possible scenario? If we believe it's not - better to convert to assertion.

lib/SPIRV/libSPIRV/SPIRV.debug.h
```diff
@@ -55,6 +55,7 @@ enum Instruction {
   ModuleINTEL                   = 36,
   InstCount                     = 37,
   FunctionDefinition            = 101,
+  EntryPoint                    = 107,
```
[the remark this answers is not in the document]
vmaksimo: Should I have done this for FunctionDefinition also?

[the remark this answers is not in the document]
MrSidims: No, FunctionDefinition has no parent scope, aka it's linked with CU implicitly thorough the appropriate DebugFunction

lib/SPIRV/LLVMToSPIRVDbgTran.cpp
vmaksimo: It's possible to make these two functions return void. However they call `BM->addDebugInfo` inside, which does return the value - so the problem for static analysis will still exist. Not sure how to resolve this properly

MrSidims: okay, lets do not change result type to void. Instead we can add
 [[maybe_unused]] auto *Inst = transDbgEntryPoint

MrSidims (approved):
LGTM, thanks!
Please fix clang-format

PR #1974, opened 2023-04-25
[NFC] Add gather/scatter intrinsics to 'known'

MrSidims:
It prevents their declaration from generation in case if SPV_INTEL_masked_gather_scatter and -spirv-allow-unknown-intrinsics are both enabled.


MrSidims:
@asudarsa @vmaksimo @maksimsab @jgstarIntel @LU-JOHN please take a look

MrSidims:
> Could you please clarify, why do we always translate declarations of unknown intrinsics? I guess the solution is correct, but want to be sure that it won't cause any problems when we translate such intrinsic (or is it not possible without extension enabled?)

I don't know. At this moment this code is too old to touch and may require some refactoring.

vmaksimo (approved):

PR #1975, opened 2023-04-25
[DebugInfo] Fix Composite type translation for NonSemantic spec

MrSidims:
Tag should be OpConstant, not Literal.

### DIFF
--- a/test/DebugInfo/TransTypeCompositeCaseClass_.ll
+++ b/test/DebugInfo/TransTypeCompositeCaseClass_.ll
@@ -13,6 +13,10 @@
 ; RUN: llvm-spirv -r -emit-opaque-pointers %t.spv -o %t.rev.bc
 ; RUN: llvm-dis < %t.rev.bc | FileCheck %s --check-prefix=CHECK-LLVM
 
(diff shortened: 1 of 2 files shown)


MrSidims:
@asudarsa @vmaksimo @maksimsab @jgstarIntel @LU-JOHN please take a look

test/DebugInfo/TransTypeCompositeCaseClass_.ll
```diff
@@ -13,6 +13,10 @@
 ; RUN: llvm-spirv -r -emit-opaque-pointers %t.spv -o %t.rev.bc
 ; RUN: llvm-dis < %t.rev.bc | FileCheck %s --check-prefix=CHECK-LLVM
 
+; RUN: llvm-spirv --spirv-debug-info-version=nonsemantic-shader-100 %t.bc -o %t.spv
```
vmaksimo, line 16: Should we test something like that in DebugInfo/NonSemantic tests?

MrSidims: This test has checks for both OpenCL and NonSemantic versions of the debug info. So IMHO leaving it in the common directory is good enough.

PR #1978, opened 2023-04-26
[DebugInfo] Translate inline namespace debug info

vmaksimo:
In NonSemantic.Shader.DebugInfo.200 `DebugLexicalBlock` has Inline namespace parameter, which specifies if C++ namespace is inline or not. 
It allows us to save `exportSymbols` field of `DINamespace` metadata.

Spec:
https://github.com/KhronosGroup/SPIRV-Registry/pull/186/


vmaksimo:

MrSidims (approved):
LGTM, thanks! I'll wait till tomorrow if any other feedback shows up

PR #1979, opened 2023-04-27
[Backport to 14] Preserve invalid SPIRV source lang literal in module…

mateuszchudyk:
… metadata (#1951)


aratajew:
Since @mateuszchudyk is OoO, I've created a separate PR with the same backport but with failed test fixed: https://github.com/KhronosGroup/SPIRV-LLVM-Translator/pull/1992 @MrSidims please take a look.
This one can be closed.

PR #1980, opened 2023-04-27
[Backport to 15] Preserve invalid SPIRV source lang literal in module metadata

mateuszchudyk:


MrSidims (approved):
Do we need this in 16 branch as well?

PR #1982, opened 2023-04-27
[DebugInfo] Fix translation of Target Function operand

vmaksimo:
Before reading Target function name operand of `DebugFunction` we need to make sure it's the translation of appropriate debug extension. Otherwise, we can get in a situation, where we do have 11 operands, but the last one is not a `String` with name, but the `DebugFunctionDeclaration`.

Also added more checks for differing operand numbers in different debug specs.

This is a follow-up/alignment fix for: 55ce223 and 02d37bb


vmaksimo:

MrSidims (approved):

PR #1988, opened 2023-05-03
Fix DebugFunction test

vmaksimo:
It fails in out-of-tree build because the debug info producer looks different for pre-installed LLVM compiler, e.g.:
`Ubuntu clang version 17.0.0 (++20230502031405+3e850a6`


MrSidims (approved):

svenvh (approved):
Thanks! I was having this issue locally too but didn't look into it yet; your patch fixes it for my setup.

Issue #1989, opened 2023-05-03
llvm-spirv failing with SIGSEGV or abort() from glibc when built with `-UNDEBUG` (via LLVM)


mgorny:
If LLVM is built with `-DCMAKE_BUILD_TYPE=RelWithDebInfo -DLLVM_ENABLE_ASSERTIONS=ON`, it causes `-UNDEBUG` to be implicitly appended to SPIRV-LLVM-Translator compiler flags. This in turn causes `llvm-spirv` to crash a lot. I've originally hit it while attempting to build libclc but I can also reproduce the problem with SPIRV-LLVM-Translator's test suite:

```
Testing Time: 29.34s
  Unsupported      :  43
  Passed           :  12
  Expectedly Failed:   3
  Failed           : 612
```

Example assertion failures I'm seeing are:

```
+ /tmp/portage/dev-util/spirv-llvm-translator-16.0.0/work/SPIRV-LLVM-Translator-16.0.0_build/tools/llvm-spirv/llvm-spirv -r -emit-opaque-pointers /tmp/portage/dev-util/spirv-llvm-translator-16.0.0/work/SPIRV-LLVM-Translator-16.0.0_build/test/test_output/DebugInfo/Output/DebugInfoSubrangeWithOnlyCount.spt.tmp.spv -o /tmp/portage/dev-util/spirv-llvm-translator-16.0.0/work/SPIRV-LLVM-Translator-16.0.0_build/test/test_output/DebugInfo/Output/DebugInfoSubrangeWithOnlyCount.spt.tmp.rev.bc
Fatal glibc error: malloc.c:2593 (sysmalloc): assertion failed: (old_top == initial_top (av) && old_size == 0) || ((unsigned long) (old_size) >= MINSIZE && prev_inuse (old_top) && ((unsigned long) old_end & (pagesize - 1)) == 0)
PLEASE submit a bug report to https://github.com/llvm/llvm-project/issues/ and include the crash backtrace.
Stack dump:
[...]

PR #1990, opened 2023-05-04
[test] Use LD_LIBRARY_PATH to ensure that a new lib is used (#1989)

mgorny:
Add the library `BINARY_DIR` to `LD_LIBRARY_PATH` to ensure that the freshly built `libLLVMSPIRVLib.so` is tested.  Otherwise, llvm-spirv spawned by the test suite may use the previously installed `libLLVMSPIRVLib.so`.

I have noticed the problem after rebuilding LLVM with `-DLLVM_ENABLE_ASSSERTIONS=ON`.  This meant that the previous version of `libLLVMSPIRVLib.so` now crashed, effectively causing the test suite to fail incorrectly.


MrSidims (approved):
Thanks!

svenvh (approved):

PR #1992, opened 2023-05-04
[Backport to 14] Preserve invalid SPIRV source lang literal in module…

aratajew:
… metadata (#1951)


MrSidims (approved):

MrSidims:
@aratajew thanks! We probably need this patch in LLVM 16 branch as well

PR #1993, opened 2023-05-04
[DebugInfo] Support translation of DebugSourceContinued

vmaksimo:
Max length of SPIRV instruction is 65535 words by specification.
`DebugSourceContinued` helps to overcome the limitation and specify full source code text by continuing the string from the previous `DebugSource(Continued)`instruction.

Spec:
https://github.com/KhronosGroup/SPIRV-Registry/blob/main/nonsemantic/NonSemantic.Shader.DebugInfo.100.asciidoc#DebugSourceContinued


test/DebugInfo/NonSemantic/DebugSourceContinued.ll
```diff
@@ -0,0 +1,46 @@
...
+; RUN: llvm-spirv %t.bc  --spirv-debug-info-version=nonsemantic-shader-100 -o %t.spv
+; RUN: llvm-spirv -r -emit-opaque-pointers %t.spv -o %t.rev.bc
+; RUN: llvm-dis %t.rev.bc -o %t.rev.ll
+; RUN: FileCheck %s --input-file %t.rev.ll --check-prefix CHECK-LLVM
+
+; CHECK-LLVM: !DIFile(filename: "t.c", directory: "/test", source: "A
+; CHECK-LLVM-COUNT-600000: A
+
+; CHECK-SPIRV: String [[#Str:]] "A
+; CHECK-SPIRV-COUNT-262130: A
```
MrSidims, line 13: COUNT is something new for me. Naïve calculation 262130 * 2 + 75737 = 599997 != 600000
Does COUNT counts from 1 or from 0? I'm trying to find missing "AAA".

vmaksimo: You can see that the "first A" is checked in the line above. So, somehow it is already counted, and the correct calculation will be: `(262130 + 1) * 2 + (75737 + 1) = 600000`.

For LLVM checks however it does the counting "once again" and `600000` is a correct check

MrSidims (approved):
The patch looks good, thanks!

PR #1995, opened 2023-05-08
Support the spirv.BufferSurfaceINTEL target extension type

sarnex:
This target extension type is created here: https://github.com/intel/vc-intrinsics/blob/master/GenXIntrinsics/lib/GenXIntrinsics/GenXSPIRVWriterAdaptor.cpp#L245

As with other target extension types, reverse translation is not yet supported.


MrSidims:
restarting ci

MrSidims (approved):

PR #1996, opened 2023-05-08
[DebugInfo] Translate checksum info inside DebugSource instruction

vmaksimo:
It's done in scope of NonSemantic.Shader.200.DebugInfo spec to have a proper solution for translation of checksum info (instead of the W/A done for OpenCL DebugInfo spec in #936).

DebugSource now have two more optional operands (will be reflected in spec PR later), and the operand list for instruction looks like:

| | Optional | Optional | Optional|
|--------|--------|--------|--------|
| \<id\> File| \<id\> Checksum Kind| \<id\> Checksum Value| \<id\> Text| 

Meaning that if we do have `Text` field, the previous two should be also filled; and if we do have `Checksum Kind`, then it should come in pair with `Checksum Value`.

### DIFF
--- a/lib/SPIRV/LLVMToSPIRVDbgTran.cpp
+++ b/lib/SPIRV/LLVMToSPIRVDbgTran.cpp
@@ -1347,11 +1347,18 @@ SPIRVExtInst *LLVMToSPIRVDbgTran::getSource(const T *DIEntry) {
   Ops[FileIdx] = BM->getString(FileName)->getId();
   DIFile *F = DIEntry ? DIEntry->getFile() : nullptr;
 
-  if (F && F->getRawChecksum() && !isNonSemanticDebugInfo()) {
+  if (F && F->getRawChecksum()) {
     auto CheckSum = F->getChecksum().value();
-    Ops.push_back(BM->getString("//__" + CheckSum.getKindAsString().str() +
-                                ":" + CheckSum.Value.str())
-                      ->getId());
+
+    if (!isNonSemanticDebugInfo())
+      Ops.push_back(BM->getString("//__" + CheckSum.getKindAsString().str() +
+                                  ":" + CheckSum.Value.str())
+                        ->getId());
+    else if (BM->getDebugInfoEIS() ==
+             SPIRVEIS_NonSemantic_Shader_DebugInfo_200) {
+      Ops.push_back(BM->getString(CheckSum.getKindAsString().str())->getId());
+      Ops.push_back(BM->getString(CheckSum.Value.str())->getId());
+    }
   }
 
   if (F && F->getRawSource() && isNonSemanticDebugInfo()) {
@@ -1361,6 +1368,11 @@ SPIRVExtInst *LLVMToSPIRVDbgTran::getSource(const T *DIEntry) {
     constexpr size_t MaxStrSize = MaxNumWords * 4 - 1;
     const size_t NumWords = getSizeInWords(Str);
 
+    if (BM->getDebugInfoEIS() == SPIRVEIS_NonSemantic_Shader_DebugInfo_200 &&
+        Ops.size() == MinOperandCount) {
+      Ops.push_back(getDebugInfoNoneId());
+      Ops.push_back(getDebugInfoNoneId());
+    }
     Ops.push_back(BM->getString(Str.substr(0, MaxStrSize))->getId());
     SPIRVExtInst *Source = static_cast<SPIRVExtInst *>(
         BM->addDebugInfo(SPIRVDebug::Source, getVoidTy(), Ops));
--- a/lib/SPIRV/SPIRVToLLVMDbgTran.cpp
+++ b/lib/SPIRV/SPIRVToLLVMDbgTran.cpp
@@ -136,8 +136,10 @@ const std::string &SPIRVToLLVMDbgTran::getString(const SPIRVId Id) {
 }
 
 const std::string
-SPIRVToLLVMDbgTran::getStringContinued(const SPIRVId Id,
-                                       SPIRVExtInst *DebugInst) {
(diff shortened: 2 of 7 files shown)


vmaksimo:

lib/SPIRV/LLVMToSPIRVDbgTran.cpp
```diff
@@ -1361,6 +1374,11 @@ SPIRVExtInst *LLVMToSPIRVDbgTran::getSource(const T *DIEntry) {
     constexpr size_t MaxStrSize = MaxNumWords * 4 - 1;
     const size_t NumWords = getSizeInWords(Str);
 
+    if (BM->getDebugInfoEIS() == SPIRVEIS_NonSemantic_Shader_DebugInfo_200 &&
+        Ops.size() == MinOperandCount) {
+      Ops.push_back(getDebugInfoNoneId());
+      Ops.push_back(getDebugInfoNoneId());
+    }
```
MrSidims, line 1381: That is not necessary, but, guess it simplifies translation, right?

lib/SPIRV/SPIRVToLLVMDbgTran.cpp
```diff
@@ -136,9 +136,9 @@ const std::string &SPIRVToLLVMDbgTran::getString(const SPIRVId Id) {
 }
 
 const std::string
-SPIRVToLLVMDbgTran::getStringContinued(const SPIRVId Id,
-                                       SPIRVExtInst *DebugInst) {
-  if (getDbgInst<SPIRVDebug::DebugInfoNone>(Id))
+SPIRVToLLVMDbgTran::getStringSourceContinued(const SPIRVId Id,
+                                             SPIRVExtInst *DebugInst) {
+  if (!isValidId(Id) || getDbgInst<SPIRVDebug::DebugInfoNone>(Id))
```
MrSidims, line 141: That looks fishy, could you please explain what you have observed during testing?

lib/SPIRV/SPIRVToLLVMDbgTran.cpp
```diff
@@ -1488,8 +1488,28 @@ DIFile *SPIRVToLLVMDbgTran::getFile(const SPIRVId SourceId) {
                      ParseChecksum(ChecksumStr));
   }
 
-  return getDIFile(getString(SourceArgs[FileIdx]), std::nullopt,
-                   getStringContinued(SourceArgs[TextIdx], Source));
+  std::optional<DIFile::ChecksumInfo<StringRef>> CS;
+  SPIRVWord StrIdx = SourceArgs[TextIdx];
+  if (Source->getExtSetKind() == SPIRVEIS_NonSemantic_Shader_DebugInfo_200) {
+    if (!getDbgInst<SPIRVDebug::DebugInfoNone>(SourceArgs[ChecksumKind]) &&
+        !getDbgInst<SPIRVDebug::DebugInfoNone>(SourceArgs[ChecksumValue])) {
```
MrSidims, line 1495: nit: check only for ChecksumKind or ChecksumValue and put assertion which infonone check for its companion

lib/SPIRV/SPIRVToLLVMDbgTran.cpp
```diff
@@ -1488,8 +1488,28 @@ DIFile *SPIRVToLLVMDbgTran::getFile(const SPIRVId SourceId) {
...
+                  ->getZExtIntValue()));
+      StringRef Checksum = getString(SourceArgs[ChecksumValue]);
+      size_t ChecksumEndPos = Checksum.find_if_not(llvm::isHexDigit);
+      CS.emplace(Kind, Checksum.substr(0, ChecksumEndPos));
+    }
+
+    if (SourceArgs.size() == MaxOperandCount)
+      StrIdx = SourceArgs[TextNonSemIdx];
+    else
+      StrIdx = SPIRVID_INVALID;
```
MrSidims, line 1508: Mmm, I'm starting to like putting debug info none(s) less. If we have another SPIR-V generator it would mean, that reader will end up not processing this instruction correctly. The spec defines number of operands for each optional parameters combination. If we have 1 optional parameter - it's always just Source. If 2 - it's always Kind and Checksum. If 3 - it's Kind followed by Checksum followed by Source.

MrSidims: Otherwise we should explicitly say in the spec, that Kind and Checksum are not optional, but mandatory and they can be DebugNoneId

lib/SPIRV/LLVMToSPIRVDbgTran.cpp
vmaksimo: Yes, it does, but I believe it should be necessary in this case. 
This branch relates to translation of Source code, which is saved in Text operand - I think we should fill previous arguments with DebugInfoNone to avoid confusion in reverse translation (is it Checksum Kind or Text operand)

lib/SPIRV/SPIRVToLLVMDbgTran.cpp
vmaksimo: Ok, maybe you're right and it would be better. I just understood the "optional" meaning differently - that they could be not set, but should if there is a text operand afterwards.

lib/SPIRV/SPIRVToLLVMDbgTran.cpp
vmaksimo: This check was added because of code on line 1509 where I explicitly set Id to invalid value to mark that we don't have Source operand in some cases.

asudarsa:
Will take a look before EOD today. Thanks

MrSidims:
I'm merging the patch as is and will follow up with my comments in different PR since can't update this PR

PR #2010, opened 2023-05-19
[DebugInfo] Fix CU translation when GV goes before CU

MrSidims:
Translation of DebugInfo compilation units and entry points moved before translation of GVs.

In other case we might end up in a situation when while quering for CUs we find none of them.


MrSidims:
@vmaksimo  @asudarsa @maksimsab @jgstarIntel @LU-JOHN please take a look

maksimsab (approved):
LGTM

clang-tidy complain looks like a complain about deprecated CI setting.

PR #2012, opened 2023-05-22
Implement DebugLine and DebugNoLine

LU-JOHN:
Implement DebugLine/DebugNoLine for NonSemantic.Shader.DebugInfo.100 and NonSemantic.Shader.DebugInfo.200.
Updated test/DebugInfo/NonSemantic/Shader200/DebugInfoStringType.ll  to test these changes.

### DIFF
--- a/lib/SPIRV/LLVMToSPIRVDbgTran.cpp
+++ b/lib/SPIRV/LLVMToSPIRVDbgTran.cpp
@@ -235,8 +235,13 @@ void LLVMToSPIRVDbgTran::transLocationInfo() {
               V = VPrev;
             }
           }
-          BM->addLine(V, File ? File->getId() : getDebugInfoNone()->getId(),
-                      LineNo, Col);
+          if (BM->getDebugInfoEIS() == SPIRVEIS_NonSemantic_Shader_DebugInfo_100 ||
+              BM->getDebugInfoEIS() == SPIRVEIS_NonSemantic_Shader_DebugInfo_200)
+            BM->addDebugLine(V, getVoidTy(), File ? File->getId() : getDebugInfoNone()->getId(),
+                             LineNo, Col);
+          else
+            BM->addLine(V, File ? File->getId() : getDebugInfoNone()->getId(),
+                        LineNo, Col);
         }
       } // Instructions
     }   // Basic Blocks
--- a/lib/SPIRV/LLVMToSPIRVDbgTran.cpp
+++ b/lib/SPIRV/LLVMToSPIRVDbgTran.cpp
@@ -243,7 +243,7 @@ void LLVMToSPIRVDbgTran::transLocationInfo() {
                              File ? File->getId() : getDebugInfoNoneId(),
                              LineNo, LineNo, Col, Col + 1);
           else
-            BM->addLine(V, File ? File->getId() : getDebugInfoNoneId(), LineNo, 
+            BM->addLine(V, File ? File->getId() : getDebugInfoNoneId(), LineNo,
                         Col);
         }
       } // Instructions
--- a/lib/SPIRV/LLVMToSPIRVDbgTran.cpp
+++ b/lib/SPIRV/LLVMToSPIRVDbgTran.cpp
@@ -241,7 +241,7 @@ void LLVMToSPIRVDbgTran::transLocationInfo() {
                   SPIRVEIS_NonSemantic_Shader_DebugInfo_200)
             BM->addDebugLine(V, getVoidTy(),
                              File ? File->getId() : getDebugInfoNone()->getId(),
-                             LineNo, Col);
+                             LineNo, LineNo, Col, Col + 1);
           else
             BM->addLine(V, File ? File->getId() : getDebugInfoNone()->getId(),
                         LineNo, Col);
--- a/lib/SPIRV/LLVMToSPIRVDbgTran.cpp
+++ b/lib/SPIRV/LLVMToSPIRVDbgTran.cpp
@@ -243,8 +243,8 @@ void LLVMToSPIRVDbgTran::transLocationInfo() {
                              File ? File->getId() : getDebugInfoNoneId(),
                              LineNo, LineNo, Col, Col + 1);
           else
-            BM->addLine(V, File ? File->getId() : getDebugInfoNoneId(),
-                        LineNo, Col);
+            BM->addLine(V, File ? File->getId() : getDebugInfoNoneId(), LineNo, 
+                        Col);
         }
       } // Instructions
     }   // Basic Blocks
--- a/lib/SPIRV/LLVMToSPIRVDbgTran.cpp
+++ b/lib/SPIRV/LLVMToSPIRVDbgTran.cpp
@@ -235,9 +235,12 @@ void LLVMToSPIRVDbgTran::transLocationInfo() {
               V = VPrev;
             }
           }
-          if (BM->getDebugInfoEIS() == SPIRVEIS_NonSemantic_Shader_DebugInfo_100 ||
-              BM->getDebugInfoEIS() == SPIRVEIS_NonSemantic_Shader_DebugInfo_200)
-            BM->addDebugLine(V, getVoidTy(), File ? File->getId() : getDebugInfoNone()->getId(),
+          if (BM->getDebugInfoEIS() ==
+                  SPIRVEIS_NonSemantic_Shader_DebugInfo_100 ||
+              BM->getDebugInfoEIS() ==
+                  SPIRVEIS_NonSemantic_Shader_DebugInfo_200)
+            BM->addDebugLine(V, getVoidTy(),
+                             File ? File->getId() : getDebugInfoNone()->getId(),
                              LineNo, Col);
           else
             BM->addLine(V, File ? File->getId() : getDebugInfoNone()->getId(),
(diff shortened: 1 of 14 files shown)


lib/SPIRV/SPIRVToLLVMDbgTran.cpp
```diff
@@ -1443,9 +1443,17 @@ DebugLoc SPIRVToLLVMDbgTran::transDebugScope(const SPIRVInstruction *Inst) {
   unsigned Col = 0;
   MDNode *Scope = nullptr;
   MDNode *InlinedAt = nullptr;
+
+  assert(!(Inst->getLine() && Inst->getDebugLine()) && "OpLine and DebugLine must not both be used");
```
MrSidims, line 1447: TBH I doubt that - there is no such rule in the spec.

LU-JOHN: What should be done if we have SPIRV that has both.  The spec does not show how to resolve a collision:

DebugLine
Specify source-level line and column information. This information applies to all following instructions, up to the first occurrence of any of the following: **the next end of block, the next DebugLine instruction, or the next DebugNoLine instruction**.

OpLine
Add source-level location information. This has no semantic impact and can safely be removed from a module.
This location information applies to the instructions physically following this instruction, up to the first occurrence of any of the following: **the next end of block, the next OpLine instruction, or the next [OpNoLine]**(https://registry.khronos.org/SPIR-V/specs/unified1/SPIRV.html#OpNoLine) instruction.

LU-JOHN: After discussion with MrSidims, we decided that OpLine/OpNoLine should be ignored if OpExtInstImport for non-semantic debug information is encountered.  These imports are necessary to use DebugLine/DebugNoLine so the meaning is clear.

LU-JOHN: Decided to not disable OpLine/OpNoLine if OpExtInstImport for non-semantic debug information is encountered.  We do not want a module that uses OpLine/OpNoLine to lose line information if it is linked with another module that has an OpExtInstImport for non-semantic debug information.   If OpLine and DebugLine are both active give DebugLine priority.

MrSidims:
@vmaksimo @asudarsa @maksimsab @jgstarIntel please take a look

test/DebugInfo/NonSemantic/Shader200/DebugInfoSubrange.ll
```diff
@@ -10,9 +10,17 @@
 ; CHECK-SPIRV: ExtInstImport [[#EISId:]] "NonSemantic.Shader.DebugInfo.200"
 
 ; CHECK-SPIRV: String [[#LocalVarNameId:]] "A$1$upperbound"
-; CHECK-SPIRV: TypeInt [[#TyInt64Id:]] 64 0
-; CHECK-SPIRV: Constant [[#TyInt64Id]] [[#Constant1Id:]] 1 0
-; CHECK-SPIRV: Constant [[#TyInt64Id]] [[#Constant1000Id:]] 1000 0
+; CHECK-SPIRV-DAG: TypeInt [[#TyInt32Id:]] 32 0
+; CHECK-SPIRV-DAG: TypeInt [[#TyInt64Id:]] 64 0
+; CHECK-NOT: THIS LINE IS USED TO SEPARATE DAGs
```
MrSidims, line 15: @vmaksimo can probably suggest a more graceful solution for such LIT check

MrSidims (commented):
Will take a look by EOD, but after quick scanning it looks great

The one last comment that may be worth considering, as we've discussed, Shader.100 spec says, that ColumnEnd > ColumnStart and to make it working in LLVM compilation flow we need to modify Shader.200 (and, guess, 100) saying that ColumnEnd >= ColumnStart.

We can't get use of ColumnEnd in the translator, but what about to satisfy the spec by generating ColumnEnd = ColumnStart + 1? In the end of the day no language construct can be  less then 1 char in length. And by doing that we won't break our implementation and will help a bit any other SPIR-V consumer that can work with NonSemantic.DebugInfo.Shader.100 spec

LU-JOHN:
> Will take a look by EOD, but after quick scanning it looks great
> 
> The one last comment that may be worth considering, as we've discussed, Shader.100 spec says, that ColumnEnd > ColumnStart and to make it working in LLVM compilation flow we need to modify Shader.200 (and, guess, 100) saying that ColumnEnd >= ColumnStart.
> 
> We can't get use of ColumnEnd in the translator, but what about to satisfy the spec by generating ColumnEnd = ColumnStart + 1? In the end of the day no language construct can be less then 1 char in length. And by doing that we won't break our implementation and will help a bit any other SPIR-V consumer that can work with NonSemantic.DebugInfo.Shader.100 spec

test/DebugInfo/NonSemantic/Shader200/DebugInfoSubrange.ll
LU-JOHN: The CHECK-NOT is used to make the test stricter.  Without the CHECK-NOT the test could pass if the TypeInt's come after the Constant's since the DAG's could be re-ordered.  

https://llvm.org/docs/CommandGuide/FileCheck.html recommends using CHECK-NOT to separate DAGs:

CHECK-NOT: directives could be mixed with CHECK-DAG: directives to exclude strings between the surrounding CHECK-DAG: directives. As a result, the surrounding CHECK-DAG: directives cannot be reordered, i.e. all occurrences matching CHECK-DAG: before CHECK-NOT: must not fall behind occurrences matching CHECK-DAG: after CHECK-NOT:. For example,

lib/SPIRV/libSPIRV/SPIRVFunction.cpp
```diff
@@ -143,7 +143,8 @@ bool SPIRVFunction::decodeBB(SPIRVDecoder &Decoder) {
     SPIRVEntry *Entry = Decoder.getEntry();
 
     if (Decoder.OpCode == OpLine) {
-      Module->add(Entry);
+      std::shared_ptr<const SPIRVLine> L(static_cast<SPIRVLine *>(Entry));
+      Module->setCurrentLine(L);
```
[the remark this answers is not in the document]
LU-JOHN: I feel this is a cleaner location.  We are setting the current line inside a loop that is processing other instructions that need the current line.   The other reason is that this makes the OpLine and DebugLine processing more similar, and I could not find an analogous decode location for DebugLine.

lib/SPIRV/LLVMToSPIRVDbgTran.cpp
```diff
@@ -235,8 +235,16 @@ void LLVMToSPIRVDbgTran::transLocationInfo() {
...
             }
           }
-          BM->addLine(V, File ? File->getId() : getDebugInfoNone()->getId(),
-                      LineNo, Col);
+          if (BM->getDebugInfoEIS() ==
+                  SPIRVEIS_NonSemantic_Shader_DebugInfo_100 ||
+              BM->getDebugInfoEIS() ==
+                  SPIRVEIS_NonSemantic_Shader_DebugInfo_200)
+            BM->addDebugLine(V, getVoidTy(),
+                             File ? File->getId() : getDebugInfoNone()->getId(),
```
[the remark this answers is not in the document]
LU-JOHN: I made this change.

test/DebugInfo/NonSemantic/Shader200/DebugInfoSubrange.ll
vmaksimo: I though that it should not happen that Types are defined after the Constants, but I double checked the spec and you're right here - it is possible. Although I doubt that we can ever generate type definition before its actual use. 
I believe we can leave this check as is if you prefer

lib/SPIRV/LLVMToSPIRVDbgTran.cpp
```diff
@@ -235,8 +235,16 @@ void LLVMToSPIRVDbgTran::transLocationInfo() {
...
+          if (BM->getDebugInfoEIS() ==
+                  SPIRVEIS_NonSemantic_Shader_DebugInfo_100 ||
+              BM->getDebugInfoEIS() ==
+                  SPIRVEIS_NonSemantic_Shader_DebugInfo_200)
+            BM->addDebugLine(V, getVoidTy(),
+                             File ? File->getId() : getDebugInfoNoneId(),
+                             LineNo, LineNo, Col, Col + 1);
+          else
+            BM->addLine(V, File ? File->getId() : getDebugInfoNoneId(),
+                        LineNo, Col);
```
MrSidims, line 247:
```suggestion
            BM->addLine(V, File ? File->getId() : getDebugInfoNoneId(), LineNo, 
                        Col);
```

lib/SPIRV/LLVMToSPIRVDbgTran.cpp
```diff
@@ -235,8 +235,16 @@ void LLVMToSPIRVDbgTran::transLocationInfo() {
...
+          if (BM->getDebugInfoEIS() ==
+                  SPIRVEIS_NonSemantic_Shader_DebugInfo_100 ||
+              BM->getDebugInfoEIS() ==
+                  SPIRVEIS_NonSemantic_Shader_DebugInfo_200)
+            BM->addDebugLine(V, getVoidTy(),
+                             File ? File->getId() : getDebugInfoNoneId(),
+                             LineNo, LineNo, Col, Col + 1);
+          else
+            BM->addLine(V, File ? File->getId() : getDebugInfoNoneId(), LineNo, 
+                        Col);
```
MrSidims, line 247:
```suggestion
            BM->addLine(V, File ? File->getId() : getDebugInfoNoneId(), LineNo,
                        Col);
```

PR #2013, opened 2023-05-23
Skip spirv decoration metadata with --spirv-preserve-auxdata

sarnex:
It's already explicitly handled for forward and reverse translation, and it's a bit complicated to handle MDNode metadata. Just skip it so we don't assert.

If I see this come up in more cases I will add support for MDNode metadata.


sarnex:
clang tidy failure seems bogus

MrSidims (approved):

PR #2027, opened 2023-05-29
Use Intel fast math ext when the capability is used

pjaaskel:


pjaaskel:
Wrong destination branch, sorry.

PR #2032, opened 2023-05-30
fix ptr.annotation in reverse translator for OpCopyMemorySized

broxigarchen:
For a memcpy call, there also could be `ptr.annotation` generated before the src and dest pointer. Update the code to check `ptr.annotation` for OpCopyMemorySized and update lit test


MrSidims:

LU-JOHN (approved):
LGTM

PR #2033, opened 2023-05-31
[DebugInfo] Adjust TypeMember for NonSemantic spec

MrSidims:
It no longer has a Scope (parent) parameter. It results in several changes including how to determine DIBuilder to use for debug info generation.

The patch also fixes a bug of incorrect debug info assignment in case of recursion DebugInfo inst generation.

### DIFF
--- a/lib/SPIRV/SPIRVToLLVMDbgTran.cpp
+++ b/lib/SPIRV/SPIRVToLLVMDbgTran.cpp
@@ -532,8 +532,8 @@ SPIRVToLLVMDbgTran::transTypeComposite(const SPIRVExtInst *DebugInst) {
     auto *MemberInst = BM->get<SPIRVExtInst>(Ops[I]);
     if (MemberInst->getExtOp() == SPIRVDebug::TypeMember) {
       auto *SPVMemberInst = BM->get<SPIRVExtInst>(Ops[I]);
-      DINode *MemberMD = transTypeMember(SPVMemberInst, DebugInst,
-                                         cast<DIScope>(CT));
+      DINode *MemberMD =
+          transTypeMember(SPVMemberInst, DebugInst, cast<DIScope>(CT));
       EltTys.push_back(MemberMD);
       DebugInstCache[SPVMemberInst] = MemberMD;
     } else {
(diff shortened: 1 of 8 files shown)


test/DebugInfo/OpenCL100/DebugInfoSubrange.ll
```diff
@@ -14,8 +14,8 @@
 ; CHECK-SPIRV: Constant [[#TypeInt64Id]] [[#NegativeCount:]] 4294967295 4294967295
 
 ; CHECK-SPIRV: [[#DbgFuncId:]] [[#]] DebugFunction [[#FuncNameId]]
+; CHECK-SPIRV: [[#]] [[#DbgLocVarId:]] [[#]] DebugLocalVariable [[#VarNameId]] [[#]] [[#]] [[#]] [[#]] [[#DbgFuncId]]
```
MrSidims, line 17: Basically here is one of the effects of https://github.com/KhronosGroup/SPIRV-LLVM-Translator/pull/2033/files#diff-ce5c478b99f08cea7d78a07829f628f520b4ab8746eb64c37874ad4f88fbccabR266 - DbgTemplateId is changed to DbgFuncId which is actually the correct behavior

MrSidims:

lib/SPIRV/SPIRVToLLVMDbgTran.h
```diff
@@ -77,10 +77,11 @@ class SPIRVToLLVMDbgTran {
...
            "Unexpected extended instruction set");
     auto It = DebugInstCache.find(DebugInst);
-    if (It != DebugInstCache.end())
+    if (It != DebugInstCache.end() && It->second != nullptr)
       return static_cast<T *>(It->second);
     MDNode *Res = transDebugInstImpl(DebugInst);
-    DebugInstCache[DebugInst] = Res;
+    if (Res)
+      DebugInstCache[DebugInst] = Res;
     return static_cast<T *>(Res);
```
vmaksimo, line 85: Why so that we check `It->second` for nullptr before casting, but we are ok with `Res` can be nullptr as a return value? Or is it the subtle logic of debug instructions nested translation?

lib/SPIRV/SPIRVToLLVMDbgTran.cpp
```diff
@@ -611,8 +620,19 @@ SPIRVToLLVMDbgTran::transTypeString(const SPIRVExtInst *DebugInst) {
                            0 /*AlignInBits*/, Encoding);
 }
 
-DINode *SPIRVToLLVMDbgTran::transTypeMember(const SPIRVExtInst *DebugInst) {
-  using namespace SPIRVDebug::Operand::TypeMember;
+DINode *
+SPIRVToLLVMDbgTran::transTypeMember(const SPIRVExtInst *DebugInst,
+                                    const SPIRVExtInst *ParentInst = nullptr,
```
vmaksimo, line 625: Maybe put default values in header file.

lib/SPIRV/SPIRVToLLVMDbgTran.h
MrSidims: It's a leftover from a previous version of the implementation. Removed.

lib/SPIRV/SPIRVToLLVMDbgTran.cpp
MrSidims: Applied, thanks!

vmaksimo (approved):

PR #2034, opened 2023-06-01
[DebugInfo] Add Flags operand for DebugTypeBasic for NonSemantic spec

vmaksimo:
Flags operand is not optional, fill it with DebugInfoNone value (see #1965 for clarifications why it could not be generated by clang or by LLVM environment).


MrSidims:

MrSidims:
@vmaksimo looks like we are missing a stand-alone LIT for DebugBasicType. May I ask you add a simple one, like this: https://godbolt.org/z/8Yx9efhKT in test/DebugInfo (or test/DebugInfo/Generic) and add checks for OpenCL and NonSemantic debug infos?

MrSidims (approved):

PR #2035, opened 2023-06-02
[OpaquePointers] Fix @llvm.annotation translation with opaque pointers enabled.

jcranmer-intel:


jcranmer-intel:
When putting this patch together, I noted that there's an issue with the handling of `@llvm.ptr.annotation`. If I change the type rules to specifically propagate the type of the first operand to the output operand instead of forcing both to be `i8*`, then I can avoid the test check changes needed here. However, this adjusts the output of several existing opaque pointer annotation tests, and there's a SPIRVReader crash on handling the resulting code for the `multiple_user_semantic_on_struct` change.

@vmaksimo, you seem to be the one mostly responsible for handling these annotations, can you shed some light on which translation should be preferred?

MrSidims:
@vmaksimo @broxigarchen please take a look

jcranmer-intel:
Friendly review ping?

MrSidims (approved):

MrSidims:
@vmaksimo @broxigarchen would still prefer to have your review

PR #2039, opened 2023-06-05
[DebugInfo] Adjust TypeInheritance for NonSemantic spec

vmaksimo:
It does not have `Child` parameter comparing to OpenCL specification.
https://github.com/KhronosGroup/SPIRV-Registry/blob/main/nonsemantic/NonSemantic.Shader.DebugInfo.100.asciidoc#DebugTypeInheritance

### DIFF
--- a/lib/SPIRV/SPIRVToLLVMDbgTran.cpp
+++ b/lib/SPIRV/SPIRVToLLVMDbgTran.cpp
@@ -536,6 +536,11 @@ SPIRVToLLVMDbgTran::transTypeComposite(const SPIRVExtInst *DebugInst) {
           transTypeMember(SPVMemberInst, DebugInst, cast<DIScope>(CT));
       EltTys.push_back(MemberMD);
       DebugInstCache[SPVMemberInst] = MemberMD;
+    } else if (MemberInst->getExtOp() == SPIRVDebug::TypeInheritance) {
+      auto *SPVMemberInst = BM->get<SPIRVExtInst>(Ops[I]);
+      DINode *MemberMD = transTypeInheritance(SPVMemberInst, cast<DIType>(CT));
+      EltTys.push_back(MemberMD);
+      DebugInstCache[SPVMemberInst] = MemberMD;
     } else {
       EltTys.emplace_back(transDebugInst(BM->get<SPIRVExtInst>(Ops[I])));
     }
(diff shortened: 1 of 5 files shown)


vmaksimo:

lib/SPIRV/SPIRVToLLVMDbgTran.cpp
```diff
@@ -536,6 +536,11 @@ SPIRVToLLVMDbgTran::transTypeComposite(const SPIRVExtInst *DebugInst) {
           transTypeMember(SPVMemberInst, DebugInst, cast<DIScope>(CT));
       EltTys.push_back(MemberMD);
       DebugInstCache[SPVMemberInst] = MemberMD;
+    } else if (MemberInst->getExtOp() == SPIRVDebug::TypeInheritance) {
```
LU-JOHN, line 539: Should this "else if" code also be guarded by a isNonSemanticDebugInfo() ?

vmaksimo: I don't think so, but I'm not really sure. Maybe @MrSidims can justify this as I just followed his change above for TypeMember

asudarsa: We do not need to check here. The check will be found inside transTypeInheritance when we are processing the operands.

lib/SPIRV/libSPIRV/SPIRV.debug.h
```diff
@@ -499,12 +499,14 @@ enum {
 
 namespace TypeInheritance {
 enum {
-  ChildIdx     = 0,
-  ParentIdx    = 1,
-  OffsetIdx    = 2,
-  SizeIdx      = 3,
-  FlagsIdx     = 4,
-  OperandCount = 5
+  ChildIdx        = 0,
```
asudarsa, line 502: I thought there was no 'child'?

asudarsa (approved):
LGTM. Just one question. But it should not block the review.

Thanks

MrSidims (approved):
LGTM

lib/SPIRV/libSPIRV/SPIRV.debug.h
MrSidims: Ultimately we want to do a refactoring splitting a single DebugInfo class to several:
```
DebugInfoCommon -> DebugInfoOpenCL
                                 -> DebugInfoNonSemantic -> Shader100
                                                                              -> Shader200
```
it will simplify future support

vmaksimo: Unification sounds good, although it would increase the amount of duplicating code (if we talk only about namespaces).
As I remember, it's not the first change that uses offset instead of separate namespace. But if we are about to unify and refactor such things, I believe it can be fixed in a separate PR (I can also help with that)

PR #2042, opened 2023-06-07
[DebugInfo] Add Column parameter to DebugInlinedAt instruction

vmaksimo:
The change is done as a part of NonSemantic.Shader.200 spec, and the new arguments for the `DebugInlinedAt` instruction will look like:

| \<id\> Line| \<id\> Column| \<id\> Scope| Optional\<id\> Inlined|
|--------|--------|--------|--------|

### DIFF
--- a/lib/SPIRV/LLVMToSPIRVDbgTran.cpp
+++ b/lib/SPIRV/LLVMToSPIRVDbgTran.cpp
@@ -1378,7 +1378,10 @@ SPIRVEntry *LLVMToSPIRVDbgTran::transDebugLoc(const DebugLoc &Loc,
 }
 
 SPIRVEntry *LLVMToSPIRVDbgTran::transDbgInlinedAt(const DILocation *Loc) {
-  using namespace SPIRVDebug::Operand::InlinedAt;
+  // There is a Column operand in NonSemantic.Shader.200 spec
+  if (BM->getDebugInfoEIS() == SPIRVEIS_NonSemantic_Shader_DebugInfo_200)
+    return transDbgInlinedAtNonSemanticShader200(Loc);
+  using namespace SPIRVDebug::Operand::InlinedAt::OpenCL;
   SPIRVWordVec Ops(MinOperandCount);
   Ops[LineIdx] = Loc->getLine();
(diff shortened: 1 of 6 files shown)


vmaksimo:

lib/SPIRV/LLVMToSPIRVDbgTran.cpp
```diff
@@ -1378,7 +1378,10 @@ SPIRVEntry *LLVMToSPIRVDbgTran::transDebugLoc(const DebugLoc &Loc,
 }
 
 SPIRVEntry *LLVMToSPIRVDbgTran::transDbgInlinedAt(const DILocation *Loc) {
-  using namespace SPIRVDebug::Operand::InlinedAt;
+  // There is a Column operand in NonSemantic.Shader.200 spec
+  if (BM->getDebugInfoEIS() == SPIRVEIS_NonSemantic_Shader_DebugInfo_200)
+    return transDbgInlinedAtNonSemanticShader200(Loc);
```
asudarsa, line 1383: Why do we need a separate function? Can we just not use an offset variable here? Juts a nit though. Thanks

lib/SPIRV/libSPIRV/SPIRV.debug.h
```diff
@@ -942,7 +955,8 @@ enum {
 // helper function to get parent scope of debug instruction, to be used
 // to determine with which compile unit the particular instruction relates
 inline bool hasDbgInstParentScopeIdx(const uint32_t Kind,
```
asudarsa, line 957: This change can be avoided if we change the operand order to be line,scope,column. Is there a reason why we need to have column before scope?

asudarsa (approved):
Looks good. Thanks

lib/SPIRV/libSPIRV/SPIRV.debug.h
vmaksimo: I understand this, but decided that we should not align specification for some implementation simplifications.
The order `line, column, scope` is common/"unified" throughout the specification, so I think we should put the arguments in alike way.

MrSidims (approved):

PR #2052, opened 2023-06-13
Fix crash with --spirv-preserve-auxdata and removed function

sarnex:
In `SPIRVReader`, we erase unused functions pretty late, but we were applying the ExtInst for `--spirv-preserve-auxdata` after that, causing us to try to work on a deleted function which lead to a crash.

This isn't a problem for the `SPIRVWriter` case because the unused function deletion happens way earlier.


sarnex:
The in-tree failures do not seem related.

asudarsa:
@MrSidims, @vmaksimo @LU-JOHN 

Please take a look.

Thanks

MrSidims (approved):

PR #2063, opened 2023-06-25
[Backport to 9] Adjust builtin variable tracking to support i8 geps

lsatanov:
The existing logic for the replacement of builtin variables with calls to functions relies on relatively brittle tracking that is broken when opaque pointers is turned on, and will be even more thoroughly broken if/when typed geps are replaced with i8 geps or ptradd. This patch replaces that logic with a less brittle variant that is able to handle any sequence of bitcast, gep, or addrspacecast instructions between the global variable and the ultimate load instruction.

It still will error out if the variable is used in too insane of a fashion (say, trying to load an i32 out of the i64, or a misaligned vector type).

------------------------ Backport notes ---------------------
- This patch is backported as a fix to scalar reads from vector global variables.
- Replaced LLVM::APInt::isZero() usage with operator!(), since isZero() API is not available in LLVM-9
- test/builtin-vars-gep.ll: @foo_i8gep converted not to use opaque pointers.

### DIFF
--- a/lib/SPIRV/SPIRVInternal.h
+++ b/lib/SPIRV/SPIRVInternal.h
@@ -1027,7 +1027,7 @@ std::string decodeSPIRVTypeName(StringRef Name,
                                 SmallVectorImpl<std::string> &Strs);
 
 // Copy attributes from function to call site.
-void setAttrByCalledFunc(CallInst *Call);
+CallInst *setAttrByCalledFunc(CallInst *Call);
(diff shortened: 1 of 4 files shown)


vmaksimo (approved):
I guess you can remove 'opaque pointers' tag from the title as long as we have a reference number to the original PR (we do)

lsatanov:
> I guess you can remove 'opaque pointers' tag from the title as long as we have a reference number to the original PR (we do)

Removed the tag.

MrSidims (approved):

PR #2066, opened 2023-06-27
[Backport to 14] Adjust builtin variable tracking to support i8 geps

lsatanov:
The existing logic for the replacement of builtin variables with calls to functions relies on relatively brittle tracking that is broken when opaque pointers is turned on, and will be even more thoroughly broken if/when typed geps are replaced with i8 geps or ptradd. This patch replaces that logic with a less brittle variant that is able to handle any sequence of bitcast, gep, or addrspacecast instructions between the global variable and the ultimate load instruction.

It still will error out if the variable is used in too insane of a fashion (say, trying to load an i32 out of the i64, or a misaligned vector type).

------------------------- Backport notes -------------------------
- This patch is backported as a fix to scalar reads from vector global variables.
- Original test from backported (LLVM-17) commit (requires opaque pointers support)
changed to contain no opaque pointers until-if the required opaque pointers support 
gets backported for LLVM->SPIRV stage.

### DIFF
--- a/test/builtin-vars-gep.ll
+++ b/test/builtin-vars-gep.ll
@@ -17,25 +17,33 @@ entry:
   %0 = addrspacecast <3 x i64> addrspace(1)* @__spirv_BuiltInWorkgroupSize to <3 x i64> addrspace(4)*
   %1 = getelementptr <3 x i64>, <3 x i64> addrspace(4)* %0, i64 0, i64 0
 ; CHECK-LLVM: %[[GLocalSize0:[0-9]+]] = call spir_func i64 @_Z14get_local_sizej(i32 0) #1
-; CHECK-LLVM: %[[Ins0:[0-9]+]] = insertelement <3 x i64> undef, i64 %[[GLocalSize0]], i32 0
-; CHECK-LLVM: %[[GLocalSize1:[0-9]+]] = call spir_func i64 @_Z14get_local_sizej(i32 1) #1
-; CHECK-LLVM: %[[Ins1:[0-9]+]] = insertelement <3 x i64> %[[Ins0]], i64 %[[GLocalSize1]], i32 1
-; CHECK-LLVM: %[[GLocalSize2:[0-9]+]] = call spir_func i64 @_Z14get_local_sizej(i32 2) #1
-; CHECK-LLVM: %[[Ins2:[0-9]+]] = insertelement <3 x i64> %[[Ins1]], i64 %[[GLocalSize2]], i32 2
-; CHECK-LLVM: %[[Extract:[0-9]+]] = extractelement <3 x i64> %[[Ins2]], i64 0
   %2 = addrspacecast <3 x i64> addrspace(1)* @__spirv_BuiltInWorkgroupSize to <3 x i64> addrspace(4)*
   %3 = getelementptr <3 x i64>, <3 x i64> addrspace(4)* %2, i64 0, i64 2
   %4 = load i64, i64 addrspace(4)* %1, align 32
-  %5 = load i64, i64 addrspace(4)* %3, align 8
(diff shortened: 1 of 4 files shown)


lsatanov:
@MrSidims , @vmaksimo , please tell how do we run workflows for this?

vmaksimo (approved):

lsatanov:
Checked locally, test is failing with -opaque-pointers option supplied 

```
llvm-spirv: /usr/lib/llvm-14/include/llvm/IR/Instructions.h:923: llvm::Type* llvm::checkGEPType(llvm::Type*): Assertion `Ty && "Invalid GetElementPtrInst indices for type!"' failed.                                                
PLEASE submit a bug report to https://github.com/llvm/llvm-project/issues/ and include the crash backtrace.
 
```

test/builtin-vars-gep.ll
```diff
@@ -13,29 +23,61 @@ target triple = "spir64"
...
+; ORIG:   %1 = getelementptr <3 x i64>, ptr addrspace(4) %0, i64 0, i64 0
+; ORIG: ; CHECK-LLVM %[[GLocalSize0:[0-9]+]] = call spir_func i64 @_Z14get_local_sizej(i32 0) #1
+; ORIG:   %2 = addrspacecast ptr addrspace(1) @__spirv_BuiltInWorkgroupSize to ptr addrspace(4)
+; ORIG:   %3 = getelementptr <3 x i64>, ptr addrspace(4) %2, i64 0, i64 2
+; ORIG:   %4 = load i64, ptr addrspace(4) %1, align 32
+; ORIG:   %5 = load i64, ptr addrspace(4) %3, align 8
+; ORIG: ; CHECK-LLVM %[[GLocalSize2:[0-9]+]] = call spir_func i64 @_Z14get_local_sizej(i32 2) #1
+; ORIG: ; CHECK-LLVM  mul i64 %[[GLocalSize0]], %[[GLocalSize2]]
+; ORIG:   %mul = mul i64 %4, %5
+; ORIG:   ret void
```
MrSidims, line 37: that's confusing, shouldn't we leave just CHECK-LLVM ?

MrSidims (approved):
Please consider providing justification of the backport in the comment like: improve of builtin GV translation removing unnecessary vector manipulation etc as LLVM 14 (and other branches) don't use opaque pointers by default

test/builtin-vars-gep.ll
lsatanov: This is the test as it was in original commit: with opaque pointers.

There are two ways to backport this change: 
1. Either find and port the other changes related to opaque pointers support on LLVM -> SPIRV translation stage, then we can use this original test with -opaque-pointers options added to RUN: commands. 
2. Backport this commit only, but with the test changed not to use opaque pointers (which is currently happening in this pull request).

MrSidims: I suggest to leave only used in the current patch on the LLVM 14 (15, 16, 9) CHECK-LLVMs removing ORIG lines.

lsatanov: Sorry, do you suggest removing 
1. All ORIG lines
2.  Only those containing ORIG: ; CHECK-LLVM
 ? 
 Please notice that  CHECK-LLVM are not triggered as they have no ":" at the end of the token

MrSidims: I suggest to remove all ORIG lines as this patch is not about opaque pointers and ORIG lines contains them, it's a minor comment - this very test on this patch is read only anyway. Feel free to not apply it. Yet would be nice to update the description for the PR itself saying what is needed from the original patch.

lsatanov:
> 

Added 

> ------------------------- Backport notes -------------------------
> * This patch is backported as a fix to scalar reads from vector global variables.

PR #1878, opened 2023-03-10
[DebugInfo] Support translation of DIModule

vmaksimo:
This entity represents a module in the programming language, for example a Fortran module.
Spec:
KhronosGroup/SPIRV-Registry#186

The implementation is the same as for SPV_INTEL_debug_module extension. Spec for extension:
https://github.com/intel/llvm/blob/sycl/sycl/doc/design/spirv-extensions/SPV_INTEL_debug_module.asciidoc

PR #2068, opened 2023-06-28
[DebugInfo] Fix translation of SPV_INTEL_debug_module extension

vmaksimo:
We support `DebugModuleINTEL` instruction as a part of [SPV_INTEL_debug_module extension](https://github.com/intel/llvm/blob/sycl/sycl/doc/design/spirv-extensions/SPV_INTEL_debug_module.asciidoc).
We also support translation of `DebugModule` instruction as a part of NonSemantic.Shader.200 specification, which is basically the same as for DebugModuleINTEL (introduced in #1878).

There was a problem in reverse translation when the extension and NonSemantic.Shader.100 debug info is enabled, which is fixed by this patch.
The related tests cases are enabled back.


vmaksimo:

vmaksimo:
Guess clang-tidy warnings can be ignored

maksimsab (approved):
LGTM

MrSidims (approved):

PR #2069, opened 2023-07-01
llvm-spirv: Set Opaque Pointers support also when compiling LLVM to SPIR-V

karolherbst:
LLVM implicitly upgrades modules to opaque pointers, so we have to do it also when translating LLVM to SPIR-V.

This should also get backported to branches containing 0fe726f5a6d3617b3122f8d000b00b9c90df7719 (and backports).


karolherbst:
Seems like a lot of tests are failing, because they expect `-emit-opaque-pointers` to be set. Maybe it's better to only do this for the llvm-16 based branch for now and hope everything is going to be fixed with llvm-17.

But maybe fixing the translator is the better approach here, though for llvm-16 I think it's better to fallback to keep opaque pointers disabled.

karolherbst:
maybe we even want to disable it by default as this might break users? Not quite sure. Any thoughts on this?

MrSidims:
AFAIK opaque pointers are not default for LLVM 16, so I'd vote not to enable them here. @jcranmer-intel WDYT?

karolherbst:
> AFAIK opaque pointers are not default for LLVM 16, so I'd vote not to enable them here. @jcranmer-intel WDYT?

they are on by default though, so that's why I need this change so the `LLVMContext` knows the translator wants them disabled.

LLVM tooling implicitly upgrades to opaque pointers in its tools as well (e.g. `link` or `opt`) unless you state you want them disabled, but that's a different problem.

It's already set for the "SPIRV to LLVM" path, just not in the "LLVM to SPIRV" one

jcranmer-intel:
I've created https://github.com/KhronosGroup/SPIRV-LLVM-Translator/pull/2074 for flipping the default of -emit-opaque-pointers on trunk.

The proper way to set the flag for LLVM-to-SPIR-V translation is to use the `-opaque-pointers=0` flag for versions that don't autodetect LLVM IR on output (which is 16 and newer). I don't see a need to add a redundant `-emit-opaque-pointers=0` for the forward translation, especially given that the name is inherently wrong.

karolherbst:
> I've created #2074 for flipping the default of -emit-opaque-pointers on trunk.
> 
> The proper way to set the flag for LLVM-to-SPIR-V translation is to use the `-opaque-pointers=0` flag for versions that don't autodetect LLVM IR on output (which is 16 and newer). I don't see a need to add a redundant `-emit-opaque-pointers=0` for the forward translation, especially given that the name is inherently wrong.

but that doesn't work? `llvm-spirv` consumes the LLVM IR (even with opaque pointers disabled) and upgrades it to opaque pointers breaking various things translating to SPIR-V without this change.

PR #2072, opened 2023-07-05
[DebugInfo] Cache CU value translated outside of usual flow

vmaksimo:
Compilation unit can be translated earlier, e.g. in `transEntryPoint()`. We should save the translated LLVM value to be used further.


vmaksimo:

MrSidims (approved):

PR #2077, opened 2023-07-07
.clang-tidy: disable misc-include-cleaner

svenvh:
This check was recently added to clang-tidy, but the code base doesn't quite comply to it, so disable it for now.


asudarsa:
Hi @svenvh 

Thanks for putting this up for review. I also ran into some issues due to this. In the long run, do we intend to ignore this, or do we intend to turn this back on after making the code base compliant? 
In case of the latter option, it will be good to add a TODO comment.

Thanks

svenvh:
> In the long run, do we intend to ignore this

Probably, for the foreseeable future at least.

MrSidims (approved):

PR #2084, opened 2023-07-14
[NFC] Convert tests to use -emit-opaque-pointers

jcranmer-intel:
This fixes most of the tests to use -emit-opaque-pointers. There are a few tests not yet converted for various reasons:

- Several tests only have check lines for opaque struct declarations. These are not emitted in opaque pointer mode, so the tests need deeper rewrites.
- A few tests check both typed and opaque pointer output.
- A few tests only make sense with typed pointer output (ForwardPtr.ll and RecursiveType.ll in particular).
- One test has a crash in reverse translation.

Outside of these cases, all tests should now be using opaque pointer for testing reverse translation.


svenvh (approved):
Thanks!  Clearly there's more work to do, but merging this brings us a step closer so LGTM.

MrSidims (approved):

PR #2087, opened 2023-07-19
Workaround unsupported freeze insn

LU-JOHN:
Workaround unsupported freeze insn by:
1. replacing uses of freeze's result with freeze's source or a random (but compilation reproducible) constant if freeze's source is undef/poison
2. deleting freeze insn.

Long term solution is to add a freeze instruction extension in SPIR-V.  Issue is tracked in (https://github.com/KhronosGroup/SPIRV-LLVM-Translator/issues/1140)

### DIFF
--- a/lib/SPIRV/SPIRVRegularizeLLVM.cpp
+++ b/lib/SPIRV/SPIRVRegularizeLLVM.cpp
@@ -416,7 +416,14 @@ bool SPIRVRegularizeLLVMBase::regularize() {
 
         // FIXME: This is not valid handling for freeze instruction
         if (auto *FI = dyn_cast<FreezeInst>(&II)) {
-          FI->replaceAllUsesWith(FI->getOperand(0));
+          auto *V = FI->getOperand(0);
+          if (isa<UndefValue>(V)) {
+            if (V->getType()->isIntegerTy())
+              V = ConstantInt::get(V->getType(), 0);
+            else if (V->getType()->isFPOrFPVectorTy())
+              V = ConstantFP::get(V->getType(), 0.0);
+          }
+          FI->replaceAllUsesWith(V);
           FI->dropAllReferences();
           ToErase.push_back(FI);
         }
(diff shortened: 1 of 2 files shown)


lib/SPIRV/SPIRVRegularizeLLVM.cpp
```diff
@@ -413,6 +413,14 @@ bool SPIRVRegularizeLLVMBase::regularize() {
           if (isa<PossiblyExactOperator>(BO) && BO->isExact())
             BO->setIsExact(false);
         }
+
+        // FIXME: This is not valid handling for freeze instruction
+        if (auto *FI = dyn_cast<FreezeInst>(&II)) {
```
MrSidims, line 418: Please also check for undef/poison values and materialize them

asudarsa:
Please mention the issue (https://github.com/KhronosGroup/SPIRV-LLVM-Translator/issues/1140) and the long term solution in your PR description.

Thanks

lib/SPIRV/SPIRVRegularizeLLVM.cpp
LU-JOHN: Updated PR to materialize frozen undef/poison with a random constant.

MrSidims (approved):

lib/SPIRV/SPIRVRegularizeLLVM.cpp
```diff
@@ -413,6 +413,21 @@ bool SPIRVRegularizeLLVMBase::regularize() {
...
+          auto *V = FI->getOperand(0);
+          if (isa<UndefValue>(V)) {
+            if (V->getType()->isIntegerTy())
+              V = ConstantInt::get(V->getType(), 0);
+            else if (V->getType()->isFPOrFPVectorTy())
+              V = ConstantFP::get(V->getType(), 0.0);
```
svenvh, line 424: Would it be better to use `Constant::getNullValue` here, which would cover any type?

LU-JOHN: Updated to use Constant::getNullValue.  Added testing for ptr type which needs this change.

svenvh (approved):

MrSidims:
Restarting CI

PR #2088, opened 2023-07-20
[OpaquePointers] Rewrite joint_matrix tests

MrSidims:
This patch adds joint_matrix reverse translation to target extension type and starts rewriting all of the tests.

Some tests are being removed as outdated

Remaining tests to add after the patch:
1. tf32 test
2. element wise operations test


MrSidims:
@jcranmer-intel @vmaksimo please take a look

lib/SPIRV/SPIRVReader.cpp
```diff
@@ -450,26 +450,28 @@ Type *SPIRVToLLVM::transType(SPIRVType *T, bool UseTPT) {
...
-                                     transTypeToOCLTypeName(MT->getCompType()),
-                                     Params, !UseTPT));
-    std::string ComponentTypeName;
+      return mapType(
+          T, llvm::TargetExtType::get(*Context, "spirv.JointMatrixINTEL",
+                                      transType(MT->getCompType()), Params));
     switch (static_cast<SPIRVConstant *>(CTI)->getZExtIntValue()) {
     case internal::InternalJointMatrixCTI::TF32:
-      ComponentTypeName = "tf32";
+      Params.push_back(1);
```
vmaksimo, line 458: Do we have some enum for these values?

MrSidims: Oh, we do have them :) it is "internal::InternalJointMatrixCTI::"

asudarsa:
Ah. I see that we are running into 'opaque-pointers=0' not available errors in our CI. How do we intend to work around this or tackle this?

Thanks

jcranmer-intel (approved):

PR #2090, opened 2023-07-20
Several more opaque test fixes

jcranmer-intel:
This does have a couple of code changes, with the riskiest being the PipeStorage changes, so it is not an NFC change.


MrSidims (approved):
Thanks!

svenvh (approved):

PR #2089, opened 2023-07-20
[OpaquePointers][NFC] Remove 3 calls of getNonOpaquePointerElementType

MrSidims:
getNonOpaquePointerElementType is deprecated. This patch removes 3 calls to it where it's no longer require functional changes.

There are several more calls to this API, but it requires some follow changes at this moment.

PR #2091, opened 2023-07-21
[OpaquePointers][NFC] Replace use of isOpaquePointerTy with isPointerTy

asudarsa:
'bool llvm::Type::isOpaquePointerTy() const' has been removed.
We need to use isPointerTy() instead.

This PR handles all uses except one (lib/SPIRV/SPIRVWriter.cpp:279) which is handled in a different PR (https://github.com/KhronosGroup/SPIRV-LLVM-Translator/pull/2089)

Thanks


lib/SPIRV/OCLUtil.cpp
```diff
@@ -1335,26 +1335,6 @@ static StringRef getStructName(Type *Ty) {
   return "";
 }
 
-Value *unwrapSpecialTypeInitializer(Value *V) {
```
asudarsa, line 1338: This function is specifically use for non opaque pointers.
Removing it completely and replacing the return value in call-sites with nullptr.

lib/SPIRV/SPIRVWriter.cpp
```diff
@@ -802,7 +801,7 @@ SPIRVType *LLVMToSPIRVBase::transSPIRVOpaqueType(StringRef STName,
 
 SPIRVType *LLVMToSPIRVBase::transScavengedType(Value *V) {
   if (auto *F = dyn_cast<Function>(V)) {
-    FunctionType *FnTy = F->getType()->isOpaquePointerTy()
+    FunctionType *FnTy = F->getType()->isPointerTy()
                              ? Scavenger->getFunctionType(F)
                              : F->getFunctionType();
```
asudarsa, line 806: It seems like this can be replaced as follows:

```suggestion
    FunctionType *FnTy = Scavenger->getFunctionType(F);
```

lib/SPIRV/SPIRVLowerBitCastToNonStandardType.cpp
```diff
@@ -77,7 +77,7 @@ static Value *removeBitCasts(Value *OldValue, Type *NewTy, NFIRBuilder &Builder,
   if (auto *LI = dyn_cast<LoadInst>(OldValue)) {
     Builder.SetInsertPoint(LI);
     Value *Pointer = LI->getPointerOperand();
-    if (!Pointer->getType()->isOpaquePointerTy()) {
+    if (!Pointer->getType()->isPointerTy()) {
```
MrSidims, line 80: `!Pointer->getType()->isPointerTy()` is always false, as Pointer is obtained from Load->getPointerOperand

lib/SPIRV/SPIRVWriter.cpp
jcranmer-intel: Indeed.

lib/SPIRV/SPIRVWriter.cpp
```diff
@@ -2009,7 +2003,7 @@ LLVMToSPIRVBase::transValueWithoutDecoration(Value *V, SPIRVBasicBlock *BB,
     assert(BV);
     // Don't store opaque pointer constants in the map--we might reuse the wrong
```
LU-JOHN, line 2004: Perhaps comment should be updated to:

//  Don't store pointer constants in the map -- they are opaque and thus we might reuse the wrong

MrSidims (approved):
LGTM

svenvh (commented):
LGTM, but there are still some comments to be addressed.

MrSidims (approved):

PR #2101, opened 2023-07-26
[NFC] Remove support for legacy pass manager

LU-JOHN:
Switch to new LLVM pass manager was started on Nov 20, 2020, with old functionality moved to a base class and two derived classes were added (one for new PM and one for legacy PM).  Legacy is no longer needed for SPIRVRegularizeLLVM.  Move all functionality into one class for SPIRVRegularizeLLVM.


svenvh:
Removal of the legacy PassManager isn't scheduled yet as far as I'm aware.  There may still be downstream users relying on it, so I would vote against removal of legacy PM support.

LU-JOHN:
Legacy pass manager may still be needed by downstream users.  Do not delete yet.

PR #2107, opened 2023-07-28
Fix AttrKind attributes with --spirv-preserve-auxdata

sarnex:
For AttrKind attributes, we need to convert them to the AttrKind enum before checking if it exists or adding it, otherwise it gets added as a string.


LU-JOHN (approved):
LGTM

MrSidims (approved):

PR #2109, opened 2023-07-31
Fix delete of functions that becomes unused

hewj03:
After the first loop of deleting instructions in ValuesToDelete, deleted
instructions in ValuesToDelete are in an unstable state. Then in the
second loop of deleting, dyn_cast to GlobalValue could return true for
an instruction and double eraseFromParent causes crash. The crash is
observed in a multi-threaded test case.

Global values in ValuesToDelete are functions. Unused functions are
deleted by eraseUselessFunctions anyway.

PR #2114, opened 2023-08-04
[Backport to 14] Fix delete of functions that becomes unused

hewj03:
After the first loop of deleting instructions in ValuesToDelete, deleted
instructions in ValuesToDelete are in an unstable state. Then in the
second loop of deleting, dyn_cast to GlobalValue could return true for
an instruction and double eraseFromParent causes crash.

Global values in ValuesToDelete are functions. Unused functions are
deleted by eraseUselessFunctions anyway.

(cherry picked from commit aea1ac74ed4301df12adbffff535f72f37d0bb13)

### DIFF
--- a/lib/SPIRV/OCLToSPIRV.cpp
+++ b/lib/SPIRV/OCLToSPIRV.cpp
@@ -159,12 +159,8 @@ bool OCLToSPIRVBase::runOCLToSPIRV(Module &Module) {
 
   visit(*M);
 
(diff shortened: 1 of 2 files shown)


hewj03:
Please see https://github.com/KhronosGroup/SPIRV-LLVM-Translator/pull/2109#issuecomment-1663247423 for why this is need for 14 branch

MrSidims (approved):

PR #2116, opened 2023-08-04
Add assert for kernel arg metadata not matching kernel signature

sarnex:
There was a case where Intel's SYCL compiler was optimizing out a kernel argument but not updating the metadata. This meant the metadata had more operands than the number of kernel arguments.

We use the number of metadata operands to iterate, so we did an out of bounds access.

We can't fix this by instead using the function arguments to iterate because we still don't know which argument was removed, so just assert the metadata is valid.


sarnex:
clang-tidy CI issue seems not related.

asudarsa:
> Code style CI issue seems not related

Hi @sarnex 

Thanks for the change. I did see the same issue in another PR I reviewed. Can you please open an issue for this, if convenient?

asudarsa (changes requested):
Hi @sarnex 

LGTM. Two minor changes requested. 

Thanks

lib/SPIRV/SPIRVWriter.cpp
```diff
@@ -112,6 +112,8 @@ static void foreachKernelArgMD(
     MDNode *MD, SPIRVFunction *BF,
     std::function<void(const std::string &Str, SPIRVFunctionParameter *BA)>
         Func) {
+  assert(BF->getNumArguments() == MD->getNumOperands() &&
+         "Invalid kernel metadata");
```
asudarsa, line 116:
```suggestion
         "Invalid kernel metadata. Number of metadata operands and Number of kernel arguments");
```

test/DebugInfo/NonSemantic/Shader200/DebugInfoSubrange.ll
```diff
@@ -97,10 +97,10 @@ attributes #1 = { nofree nosync nounwind readnone speculatable willreturn }
 !6 = !{i32 2, i32 0}
 !7 = !{}
 !8 = !{i16 6, i16 14}
-!9 = !{i32 0}
-!10 = !{!"none"}
-!11 = !{!"structtype"}
-!12 = !{!""}
+!9 = !{i32 0, i32 0, i32 0}
```
asudarsa, line 100: Can we also add a test which checks for the error when invalid metadata is found?

Thanks

svenvh: Keep in mind that this is currently only checked with an assert; such a test wouldn't work when asserts are compiled out, it would have to be reworked to use `checkError` instead.  Not sure that's really worth it, just asserting that the input is valid (as done in the current patch) is fine with me.

svenvh (approved):
LGTM as is.

MrSidims (approved):
LGTM, awaiting for @asudarsa approval

lib/SPIRV/SPIRVWriter.cpp
sarnex: Applied in latest commit, thanks!

test/DebugInfo/NonSemantic/Shader200/DebugInfoSubrange.ll
sarnex: I agree with Sven on this one, I think just an assert is enough, let me know if you disagree.

sarnex:
@asudarsa I think I addressed/responded to all feedback, please take a second look when you get a chance, thanks!

MrSidims:
Arvind is on vacation, will merge the patch

PR #2119, opened 2023-08-08
Don't wrap kernels that are not being called in the module

MrSidims:
This patch is a result of a reflection about previously merged PR https://github.com/KhronosGroup/SPIRV-LLVM-Translator/pull/1149 "add an entry point wrapper around functions (llvm pass)" and is enspired by various reported translator, clang (OpenCL) and Intel GPU drivers issues (see
https://github.com/KhronosGroup/SPIRV-LLVM-Translator/issues/2029 for reference).

While SPIR-V spec states:
===
*OpName*
--//--. This has nosemantic impact and can safely be removed from a module. ===
yet having EntryPoint function and a function that shares the name via OpName might be confusing by both (old) drivers and programmers, who read the SPIR-V file.

This patch prevents generation of the wrapper function when it's not necessary to generate it aka if a kernel function is not called by other kernel.

We can do better in other cases as well, for example I have experiments of renaming a wrapped function adding a previx, so it could be distinguished from the actual kernel/entry point, but for now it doesn't pass validation for E2E OpenCL tests.

### DIFF
--- a/lib/SPIRV/SPIRVRegularizeLLVM.cpp
+++ b/lib/SPIRV/SPIRVRegularizeLLVM.cpp
@@ -363,9 +363,10 @@ bool SPIRVRegularizeLLVMBase::runRegularizeLLVM(Module &Module) {
 /// Remove entities not representable by SPIR-V
 bool SPIRVRegularizeLLVMBase::regularize() {
   eraseUselessFunctions(M);
-  addKernelEntryPoint(M);
   expandSYCLTypeUsing(M);
 
+  // Store kernels called by other kernels
+  std::vector<Function *> CalledKernels;
   for (auto I = M->begin(), E = M->end(); I != E;) {
     Function *F = &(*I++);
     if (F->isDeclaration() && F->use_empty()) {
@@ -379,7 +380,9 @@ bool SPIRVRegularizeLLVMBase::regularize() {
         if (auto Call = dyn_cast<CallInst>(&II)) {
           Call->setTailCall(false);
           Function *CF = Call->getCalledFunction();
-          if (CF && CF->isIntrinsic()) {
+          if (CF && CF->getCallingConv() == CallingConv::SPIR_KERNEL) {
+            CalledKernels.push_back(CF);
+          } else if (CF && CF->isIntrinsic()) {
             removeFnAttr(Call, Attribute::NoUnwind);
             auto *II = cast<IntrinsicInst>(Call);
             if (II->getIntrinsicID() == Intrinsic::memset ||
@@ -497,19 +500,14 @@ bool SPIRVRegularizeLLVMBase::regularize() {
     }
   }
 
+  addKernelEntryPoint(CalledKernels);
   if (SPIRVDbgSaveRegularizedModule)
     saveLLVMModule(M, RegularizedModuleTmpFile);
   return true;
 }
 
-void SPIRVRegularizeLLVMBase::addKernelEntryPoint(Module *M) {
-  std::vector<Function *> Work;
-
-  // Get a list of all functions that have SPIR kernel calling conv
-  for (auto &F : *M) {
-    if (F.getCallingConv() == CallingConv::SPIR_KERNEL)
-      Work.push_back(&F);
-  }
+void
+SPIRVRegularizeLLVMBase::addKernelEntryPoint(std::vector<Function *> Work) {
   for (auto &F : Work) {
     // for declarations just make them into SPIR functions.
     F->setCallingConv(CallingConv::SPIR_FUNC);
--- a/lib/SPIRV/SPIRVRegularizeLLVM.h
+++ b/lib/SPIRV/SPIRVRegularizeLLVM.h
@@ -51,10 +51,10 @@ public:
   // Lower functions
   bool regularize();
 
-  // SPIR-V disallows functions being entrypoints and called
-  // LLVM doesn't. This adds a wrapper around the entry point
-  // that later SPIR-V writer renames.
-  void addKernelEntryPoint(llvm::Module *M);
+  // SPIR-V disallows functions being entrypoints/kernels and called
+  // OpenCL doesn't. This adds a wrapper around the entry point if it's called
+  // by other entry point that later SPIR-V writer renames.
+  void addKernelEntryPoint(std::vector<Function *> CalledKernels);
 
   /// Some LLVM intrinsics that have no SPIR-V counterpart may be wrapped in
   /// @spirv.llvm_intrinsic_* function. During reverse translation from SPIR-V
--- a/lib/SPIRV/SPIRVWriter.cpp
+++ b/lib/SPIRV/SPIRVWriter.cpp
@@ -782,10 +782,11 @@ SPIRVFunction *LLVMToSPIRVBase::transFunctionDecl(Function *F) {
   BF->setFunctionControlMask(transFunctionControlMask(F));
   if (F->hasName()) {
     if (isKernel(F)) {
-      /* strip the prefix as the runtime will be looking for this name */
-      std::string Prefix = kSPIRVName::EntrypointPrefix;
-      std::string Name = F->getName().str();
-      BM->setName(BF, Name.substr(Prefix.size()));
+      // Strip the entry point wrapper prefix as the runtime will be looking for
+      // this name if found if there is the prefix
+      StringRef Name = F->getName();
+      (void) Name.consume_front(kSPIRVName::EntrypointPrefix);
+      BM->setName(BF, Name.str());
     } else {
       if (isUniformGroupOperation(F))
         BM->getErrorLog().checkError(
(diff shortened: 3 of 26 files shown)


MrSidims:
@airlied @vmaksimo @asudarsa @LU-JOHN @maksimsab please take a look

lib/SPIRV/SPIRVWriter.cpp
```diff
@@ -5323,8 +5324,9 @@ bool LLVMToSPIRVBase::transMetadata() {
 // Work around to translate kernel_arg_type and kernel_arg_type_qual metadata
 static void transKernelArgTypeMD(SPIRVModule *BM, Function *F, MDNode *MD,
                                  std::string MDName) {
-  std::string Prefix = kSPIRVName::EntrypointPrefix;
-  std::string Name = F->getName().str().substr(Prefix.size());
+  StringRef FunName = F->getName();
+  (void) FunName.consume_front(kSPIRVName::EntrypointPrefix);
```
vmaksimo, line 5328: Same here

test/transcoding/kernel_arg_type_qual.ll
```diff
@@ -13,10 +13,10 @@ source_filename = "test.cl"
 target datalayout = "e-m:e-p270:32:32-p271:32:32-p272:64:64-i64:64-f80:128-n8:16:32:64-S128"
 target triple = "spir64-unknown-unknown."
 
-; CHECK-SPIRV: String 18 "kernel_arg_type_qual.test.volatile,const,,"
-; CHECK-SPIRV: Name [[ARG:1[0-9]+]] "g"
+; CHECK-SPIRV: String [[#]] "kernel_arg_type_qual.test.volatile,const,,"
```
vmaksimo, line 16: Should we check the uses of these string IDs instead of just ignoring?

test/entry_point_func.ll
```diff
@@ -12,10 +12,47 @@ define spir_kernel void @testfunction() {
...
-; CHECK-SPIRV: Name [[FUNC:[0-9]+]] "testfunction"
-; CHECK-SPIRV: Decorate [[FUNC]] LinkageAttributes "testfunction" Export
-; CHECK-SPIRV: Function 2 [[FUNC]] 0 3
-; CHECK-SPIRV: Function 2 [[EP]] 0 3
-; CHECK-SPIRV: FunctionCall 2 8 [[FUNC]]
\ No newline at end of file
+; CHECK-SPIRV: EntryPoint 6 [[#CallerEn:]] "callerfunction"
+; CHECK-SPIRV: EntryPoint 6 [[#TestEn:]] "testfunction"
+; CHECK-SPIRV: Name [[#TestDecl:]] "testdeclaration"
+; CHECK-SPIRV: Name [[#TestFn:]] "testfunction"
```
vmaksimo, line 27: `having EntryPoint function and a function that shares the name via OpName might be confusing` - why do we have 2 checks for `testfunction` here? Am I missing something?

lib/SPIRV/SPIRVWriter.cpp
```diff
@@ -782,10 +782,11 @@ SPIRVFunction *LLVMToSPIRVBase::transFunctionDecl(Function *F) {
...
   if (F->hasName()) {
     if (isKernel(F)) {
-      /* strip the prefix as the runtime will be looking for this name */
-      std::string Prefix = kSPIRVName::EntrypointPrefix;
-      std::string Name = F->getName().str();
-      BM->setName(BF, Name.substr(Prefix.size()));
+      // Strip the entry point wrapper prefix as the runtime will be looking for
+      // this name if found if there is the prefix
+      StringRef Name = F->getName();
+      (void) Name.consume_front(kSPIRVName::EntrypointPrefix);
```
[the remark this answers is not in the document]
MrSidims: cppreference says:
While the behavior of std::ignore outside of [std::tie](https://en.cppreference.com/w/cpp/utility/tuple/tie) **is not formally specified**, some code guides recommend using std::ignore to avoid warnings from unused return values of [[[nodiscard](https://en.cppreference.com/w/cpp/language/attributes/nodiscard)]] functions.

TBH I'd prefer not to rely on 'not formally specified' behavior.

test/transcoding/kernel_arg_type_qual.ll
MrSidims: If I remember correctly - these IDs are unused anywhere in the module. W/A to store argument types qualifiers relies in a kernel name stored in the string to bound it with the actual kernel.

lib/SPIRV/SPIRVWriter.cpp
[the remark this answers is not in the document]
vmaksimo: OK, agree

test/transcoding/kernel_arg_type_qual.ll
vmaksimo: Good! Please double-check if they are actually unused, and mark this conversation as resolved

test/entry_point_func.ll
vmaksimo: Thanks for the clarification

lib/SPIRV/SPIRVRegularizeLLVM.cpp
```diff
@@ -363,9 +363,10 @@ bool SPIRVRegularizeLLVMBase::runRegularizeLLVM(Module &Module) {
 /// Remove entities not representable by SPIR-V
 bool SPIRVRegularizeLLVMBase::regularize() {
   eraseUselessFunctions(M);
-  addKernelEntryPoint(M);
   expandSYCLTypeUsing(M);
 
+  // Store kernels called by other kernels
```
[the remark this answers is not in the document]
svenvh: Or just:

```suggestion
  // Kernels called by other kernels
```

lib/SPIRV/SPIRVRegularizeLLVM.h
```diff
@@ -51,10 +51,10 @@ class SPIRVRegularizeLLVMBase {
   // Lower functions
   bool regularize();
 
-  // SPIR-V disallows functions being entrypoints and called
-  // LLVM doesn't. This adds a wrapper around the entry point
-  // that later SPIR-V writer renames.
-  void addKernelEntryPoint(llvm::Module *M);
+  // SPIR-V disallows functions being entrypoints/kernels and called
+  // OpenCL doesn't. This adds a wrapper around the entry point if it's called
```
svenvh, line 55: Strictly speaking the part "if it's called by other entry point" isn't handled by this function anymore?

lib/SPIRV/SPIRVRegularizeLLVM.cpp
```diff
@@ -379,7 +380,9 @@ bool SPIRVRegularizeLLVMBase::regularize() {
         if (auto Call = dyn_cast<CallInst>(&II)) {
           Call->setTailCall(false);
           Function *CF = Call->getCalledFunction();
-          if (CF && CF->isIntrinsic()) {
+          if (CF && CF->getCallingConv() == CallingConv::SPIR_KERNEL) {
+            CalledKernels.push_back(CF);
```
svenvh, line 384: I assume this doesn't detect kernels launched through `enqueue_kernel` (aka device execution/device enqueue).  Do we need to handle that here too?

test/transcoding/kernel_arg_type_qual.ll
MrSidims: Double checked, here is the example (from this test):
  ```
1 119734787 65536 393230 13 0
  2 2 Capability Addresses
  3 2 Capability Kernel
  4 2 Capability Int8
  5 5 ExtInstImport 1 "OpenCL.std"
  6 3 MemoryModel 2 2
  7 5 EntryPoint 6 6 "test"
  8 11 String 11 "kernel_arg_type.test.float*,float*,"
  9 13 String 12 "kernel_arg_type_qual.test.volatile,const,,"
 10 3 Source 3 100000
 11 3 Name 7 "g"
 12 3 Name 8 "c"
 13 3 Name 9 "asd"
 14 4 Name 10 "entry"
 15
 16 3 Decorate 7 Volatile
 17 4 TypeInt 3 8 0
 18 2 TypeVoid 2
 19 4 TypePointer 4 5 3
 20 6 TypeFunction 5 2 4 4 4
 21
 22
 23
 24 5 Function 2 6 0 5
 25 3 FunctionParameter 4 7
 26 3 FunctionParameter 4 8
 27 3 FunctionParameter 4 9
 28
 29 2 Label 10
 30 1 Return
 31
 32 1 FunctionEnd
```

11 and 12 are not used anywhere

File: lib/SPIRV/SPIRVRegularizeLLVM.cpp (hunk 5 above)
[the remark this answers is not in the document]
Comment (MrSidims) [reply lib/SPIRV/SPIRVRegularizeLLVM.cpp]: Applied, thanks!

File: lib/SPIRV/SPIRVWriter.cpp [hunk 8]
```diff
@@ -782,10 +782,11 @@ SPIRVFunction *LLVMToSPIRVBase::transFunctionDecl(Function *F) {
   BF->setFunctionControlMask(transFunctionControlMask(F));
   if (F->hasName()) {
     if (isKernel(F)) {
-      /* strip the prefix as the runtime will be looking for this name */
-      std::string Prefix = kSPIRVName::EntrypointPrefix;
-      std::string Name = F->getName().str();
-      BM->setName(BF, Name.substr(Prefix.size()));
+      // Strip the entry point wrapper prefix as the runtime will be looking for
+      // this name if found if there is the prefix
```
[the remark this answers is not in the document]
MrSidims: Rephrased it a bit, please check a new version

lib/SPIRV/SPIRVRegularizeLLVM.h
MrSidims: Yeah, I've rephrased it a bit, I still want to keep 'if it's called by another entry point text here'

vmaksimo (approved):

lib/SPIRV/SPIRVRegularizeLLVM.cpp
svenvh: Sounds reasonable, thanks for checking!

svenvh (approved):

karolherbst:
sadly this breaks function linking as kernel declared in one translation unit can be called from another, so we can't know if a kernel is being called or not.

karolherbst:
https://github.com/KhronosGroup/SPIRV-LLVM-Translator/pull/2301

PR #2122, opened 2023-08-11
[NFC] Update stacksave and stackrestore after 25bc999d

MrSidims:
Intrinsics: Add type overload to stacksave and stackstore

    This allows use with non-0 address space stacks. llvm_ptr_ty should
    never be used. This could use some more percolation up through mlir,
    but this is enough to fix existing tests.

    https://reviews.llvm.org/D156666


MrSidims:
@vmaksimo please take a look as it's mostly for Intel extension

svenvh (approved):
LGTM, but would be good to have a review from @vmaksimo indeed.

PR #2123, opened 2023-08-11
[CI] Explicitly set clang-format binary

svenvh:
Ensure the correct clang-format binary is picked up, as `clang-format-diff.py` currently seems to look for clang-format-17 while we install clang-format-18.


svenvh:
I've added some formatting changes to `SPIRVReader.cpp` to force the check to run.

svenvh:
The formatting check seems to work again with this change.

MrSidims (approved):

PR #2126, opened 2023-08-15
Convert DebugInfo tests to opaque pointers

svenvh:
This is a bulk conversion done using the [migration script](https://gist.github.com/nikic/98357b71fd67756b0f064c9517b62a34).


MrSidims (approved):
Thanks!

PR #2127, opened 2023-08-16
Convert transcoding tests to opaque pointers

svenvh:
This is a bulk conversion done using the migration script.

There are more transcoding tests to be converted; however they will need manual fixups, so leave them out of this bulk conversion.


MrSidims (approved):
Thanks for doing this!

PR #2130, opened 2023-08-18
Fix is.fpclass intrinsic emulation

MrSidims:
After bitcast float to integer we should use not SAbs but UAbs call. But since UAbs is a bit useless - we shouldn't generate *Abs at all.


MrSidims:
@asudarsa @LU-JOHN @maksimsab please take a look

PR #2132, opened 2023-08-21
Add SPV_EXT_image_raw10_raw12 writer support

svenvh:
Add basic LLVM-to-SPIR-V support for the SPV_EXT_image_raw10_raw12 extension [1], enabling the extension if constants from the extension are found in the LLVM IR.

The extension adds 2 new return values for `OpImageQueryFormat`, which are integer constants that may appear anywhere in LLVM IR. Distinguishing between the extension's constants and arbitrary constants that happen to have the same value as the extension's constants is not possible in general.  Hence this patch only covers some common use cases, where the result of `OpImageQueryFormat` is directly used in an integer comparison or switch instruction.

[1] https://github.com/KhronosGroup/SPIRV-Registry/blob/main/extensions/EXT/SPV_EXT_image_raw10_raw12.asciidoc


test/extensions/EXT/SPV_EXT_image_raw10_raw12/constant-use.cl
```diff
@@ -0,0 +1,19 @@
...
+kernel void test_raw1012(global int *dst, int value) {
+  switch (value) {
+  case 0x10E3: // same value as CLK_UNSIGNED_INT_RAW10_EXT
+    *dst = 10;
+    break;
+  case 0x10E4: // same value as CLK_UNSIGNED_INT_RAW12_EXT
+    *dst = 12;
+    break;
+  }
+}
```
LU-JOHN, line 19: For completeness also check that:
if (value==0x10E3 || value==0x10E4)
does not trigger extension.

svenvh: Added.

lib/SPIRV/OCLToSPIRV.cpp
```diff
@@ -1355,10 +1357,56 @@ void OCLToSPIRVBase::visitCallScalToVec(CallInst *CI, StringRef MangledName,
...
+  return false;
+}
+} // anonymous namespace
+
 void OCLToSPIRVBase::visitCallGetImageChannel(CallInst *CI,
                                               StringRef DemangledName,
                                               unsigned int Offset) {
   assert(CI->getCalledFunction() && "Unexpected indirect call");
+
+  if (Offset == OCLImageChannelDataTypeOffset) {
```
MrSidims, line 1395: It seems like we don't check if the extension was enabled by the frontend driver. Should we do this check and report an error if it's not enabled?

MrSidims (approved):
LGTM with one nit

PR #2134, opened 2023-08-23
Convert tests to opaque pointers

svenvh:
This completes the initial bulk conversion of tests using the migration script.

There are more tests to be converted; however they will need manual fixups, so leave them out of this bulk conversion.

### DIFF
--- a/test/EnqueueEmptyKernel.ll
+++ b/test/EnqueueEmptyKernel.ll
@@ -36,9 +36,9 @@ target triple = "spir64-unknown-unknown"
 define spir_kernel void @test_enqueue_empty() #0 !kernel_arg_addr_space !0 !kernel_arg_access_qual !0 !kernel_arg_type !0 !kernel_arg_base_type !0 !kernel_arg_type_qual !0 {
 entry:
   %tmp = alloca %struct.ndrange_t, align 8
-  %call = call spir_func %opencl.queue_t* @_Z17get_default_queuev() #4
-  call spir_func void @_Z10ndrange_1Dm(%struct.ndrange_t* sret(%struct.ndrange_t) %tmp, i64 1) #4
-  %0 = call i32 @__enqueue_kernel_basic_events(%opencl.queue_t* %call, i32 1, %struct.ndrange_t* %tmp, i32 0, %opencl.clk_event_t* addrspace(4)* null, %opencl.clk_event_t* addrspace(4)* null, i8 addrspace(4)* addrspacecast (i8* bitcast (void (i8 addrspace(4)*)* @__test_enqueue_empty_block_invoke_kernel to i8*) to i8 addrspace(4)*), i8 addrspace(4)* addrspacecast (i8 addrspace(1)* bitcast ({ i32, i32 } addrspace(1)* @__block_literal_global to i8 addrspace(1)*) to i8 addrspace(4)*))
+  %call = call spir_func ptr @_Z17get_default_queuev() #4
+  call spir_func void @_Z10ndrange_1Dm(ptr sret(%struct.ndrange_t) %tmp, i64 1) #4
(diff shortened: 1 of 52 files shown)


test/layout.ll
```diff
@@ -157,7 +157,7 @@ attributes #1 = { "less-precise-fpmad"="false" "no-frame-pointer-elim"="false" "
 !opencl.used.optional.core.features = !{!8}
 !opencl.compiler.options = !{!8}
 
-!0 = !{void (<3 x i32> addrspace(1)*)* @foo, !1, !2, !3, !4, !5}
+!0 = !{ptr @foo, !1, !2, !3, !4, !5}
```
vmaksimo, line 160: same

test/AtomicBuiltinsFloat.ll
```diff
@@ -72,6 +72,6 @@ attributes #2 = { convergent nounwind }
 !1 = !{i32 2, i32 0}
 !2 = !{i32 3}
 !3 = !{!"none"}
-!4 = !{!"atomic_float*"}
+!4 = !{!"atomic_ptr"}
```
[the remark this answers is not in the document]
svenvh: Good catch, seems the script was over-enthusiastic here.  I'll revert it.

test/simple.ll
```diff
@@ -146,8 +143,8 @@ attributes #2 = { nounwind readnone }
 !4 = !{!"int*", !"int*", !"int"}
```
vmaksimo, line 143: I guess these should be changed as well as !8 and !9 below

vmaksimo (commented):
One more finding :)

test/simple.ll
svenvh: Good find! This belongs to the kernel arg metadata so shouldn't be changed. And neither should `!8` and `!9` so I've taken this file out now.

I also found a few more, which I've also taken out.

test/EnqueueEmptyKernel.ll
```diff
@@ -36,38 +36,37 @@ target triple = "spir64-unknown-unknown"
...
-define internal spir_kernel void @__test_enqueue_empty_block_invoke_kernel(i8 addrspace(4)*) #3 {
+define internal spir_kernel void @__test_enqueue_empty_block_invoke_kernel(ptr addrspace(4)) #3 {
 entry:
-  call void @__test_enqueue_empty_block_invoke(i8 addrspace(4)* %0)
+  call void @__test_enqueue_empty_block_invoke(ptr addrspace(4) %0)
   ret void
 }
 
-declare i32 @__enqueue_kernel_basic_events(%opencl.queue_t*, i32, %struct.ndrange_t*, i32, %opencl.clk_event_t* addrspace(4)*, %opencl.clk_event_t* addrspace(4)*, i8 addrspace(4)*, i8 addrspace(4)*)
+declare i32 @__enqueue_kernel_basic_events(ptr, i32, ptr, i32, ptr addrspace(4), ptr addrspace(4), ptr addrspace(4), ptr addrspace(4))
```
MrSidims, line 69: Shouldn't it be target types? Guess script conversion doesn't handle such case.

svenvh: I suppose it doesn't matter here, as translation of `__enqueue_kernel_basic_events` into `OpEnqueueKernel` is known by the translator and it mainly needs to forward the pointer arguments.

vmaksimo (approved):

svenvh:
> @svenvh if it's not a problem, could you please use separate commits for changes instead of force-push in future? :) It'd be much easier to review, mostly for such big changes

It may be a bit of a hidden GitHub feature, but you can see the diff between two force-pushes by clicking on the word "force-pushed" in "svenvh force-pushed the opaquify-tests branch ...".  But if incremental commits are preferred I can do that.

PR #2135, opened 2023-08-24
Test OpCooperativeMatrixLengthKHR

vmaksimo:
For now we are vialating the specification by passing Matrix Id instead of Matrix Type Id.

Place for improvement - to come up with the way to translate only the type.


MrSidims (approved):

test/extensions/KHR/SPV_KHR_cooperative_matrix/cooperative_matrix.ll
```diff
@@ -22,14 +22,16 @@
 ; CHECK-SPIRV-DAG: TypeCooperativeMatrixKHR [[#MatTy2:]] [[#Int8Ty]] [[#Const0]] [[#Const12]] [[#Const48]] [[#Const3]]
 ; CHECK-SPIRV-DAG: TypeCooperativeMatrixKHR [[#MatTy3:]] [[#Int8Ty]] [[#Const2]] [[#Const48]] [[#Const12]] [[#Const3]]
 ; CHECK-SPIRV: CompositeConstruct [[#MatTy1]]
-; CHECK-SPIRV: CooperativeMatrixLoadKHR [[#MatTy2]]
+; CHECK-SPIRV: CooperativeMatrixLoadKHR [[#MatTy2]] [[#Load1:]]
+; CHECK-SPIRV: CooperativeMatrixLengthKHR [[#Int32Ty]] [[#]] [[#Load1]]
```
svenvh, line 26: > For now we are vialating the specification by passing Matrix Id instead of Matrix Type Id.

Optional: maybe worth adding a TODO comment here in the test, so that we don't forget that this is currently wrong?

vmaksimo: Good idea, will add

svenvh (approved):

PR #2136, opened 2023-08-24
Fix kernel arg metadata in tests after opaque pointer conversion

svenvh:
During previous bulk conversion of tests, occurrences of `half*`, `float*`, and `double*` in kernel metadata were inadvertently updated to `ptr`.  Restore the original types.


svenvh:
@vmaksimo following your previous review comments I went through the other changes produced by the migration script and found a few more incorrect changes. Hopefully this should fix all of them.

vmaksimo (approved):

PR #2138, opened 2023-08-25
Add emulation of {s/u)_(add/sub)_sat intrinsics if needed

MrSidims:
While the translator was built around OpenCL - other targets that support SPIR-V and don't support OpenCL might use it.

This patch adds --spirv-use-ocl-math-for-llvm-intrinsic option to control whether we should translate them as math intrinsics to OpenCL ext math instructions or emulate. Default is true aka translate as math instructions.

I don't really want to end up implementing Quake's sqrt algorithm, but it's a possible scenario as well.

Plans are:
1. merge existing --spirv-replace-fmuladd-with-ocl-mad with the new option;
2. optionally introduce InstCombine pass in reverse translation.

### DIFF
--- a/include/LLVMSPIRVOpts.h
+++ b/include/LLVMSPIRVOpts.h
@@ -193,6 +193,14 @@ public:
     ReplaceLLVMFmulAddWithOpenCLMad = Value;
   }
 
+  void setUseOpenCLExtInstructionsForLLVMIntrinsic(bool Value) noexcept {
+    UseOpenCLExtInstructionsForLLVMIntrinsic = Value;
+  }
+
+  bool shouldUseOpenCLExtInstructionsForLLVMIntrinsic() const noexcept {
+    return UseOpenCLExtInstructionsForLLVMIntrinsic;
+  }
+
   bool shouldPreserveOCLKernelArgTypeMetadataThroughString() const noexcept {
     return PreserveOCLKernelArgTypeMetadataThroughString;
   }
@@ -242,6 +250,11 @@ private:
   // Controls whether llvm.fmuladd.* should be replaced with mad from OpenCL
   // extended instruction set or with a simple fmul + fadd
   bool ReplaceLLVMFmulAddWithOpenCLMad = true;
+
+  // Controls whether llvm math intrinsics should be replaced with instructions
+  // from OpenCL extended instruction set or emulated by native SPIR-V
+  // instructions
+  bool UseOpenCLExtInstructionsForLLVMIntrinsic = true;
 
   // Add a workaround to preserve OpenCL kernel_arg_type and
   // kernel_arg_type_qual metadata through OpString
--- a/lib/SPIRV/SPIRVWriter.cpp
+++ b/lib/SPIRV/SPIRVWriter.cpp
@@ -3860,21 +3860,127 @@ SPIRVValue *LLVMToSPIRVBase::transIntrinsicInst(IntrinsicInst *II,
   case Intrinsic::usub_sat:
   case Intrinsic::sadd_sat:
   case Intrinsic::ssub_sat: {
-    SPIRVWord ExtOp;
-    if (IID == Intrinsic::uadd_sat)
-      ExtOp = OpenCLLIB::UAdd_sat;
-    else if (IID == Intrinsic::usub_sat)
-      ExtOp = OpenCLLIB::USub_sat;
-    else if (IID == Intrinsic::sadd_sat)
-      ExtOp = OpenCLLIB::SAdd_sat;
-    else
-      ExtOp = OpenCLLIB::SSub_sat;
+    if (BM->shouldUseOpenCLExtInstructionsForLLVMIntrinsic()) {
+      SPIRVWord ExtOp;
+      if (IID == Intrinsic::uadd_sat)
+        ExtOp = OpenCLLIB::UAdd_sat;
+      else if (IID == Intrinsic::usub_sat)
+        ExtOp = OpenCLLIB::USub_sat;
+      else if (IID == Intrinsic::sadd_sat)
+        ExtOp = OpenCLLIB::SAdd_sat;
+      else
+        ExtOp = OpenCLLIB::SSub_sat;
 
-    SPIRVType *Ty = transType(II->getType());
-    std::vector<SPIRVValue *> Operands = {transValue(II->getArgOperand(0), BB),
-                                          transValue(II->getArgOperand(1), BB)};
-    return BM->addExtInst(Ty, BM->getExtInstSetId(SPIRVEIS_OpenCL), ExtOp,
-                          std::move(Operands), BB);
+      SPIRVType *Ty = transType(II->getType());
+      std::vector<SPIRVValue *> Operands = {
+        transValue(II->getArgOperand(0), BB),
+        transValue(II->getArgOperand(1), BB)};
+      return BM->addExtInst(Ty, BM->getExtInstSetId(SPIRVEIS_OpenCL), ExtOp,
+                            std::move(Operands), BB);
+    }
+    Type *LLVMTy = II->getType();
+    SPIRVType *Ty = transType(LLVMTy);
+    Type *BoolTy = IntegerType::getInt1Ty(M->getContext());
+    if (auto *VecTy = dyn_cast<VectorType>(LLVMTy))
+      BoolTy = VectorType::get(BoolTy, VecTy->getElementCount());
+    SPIRVType *SPVBoolTy = transType(BoolTy);
+    SPIRVValue *FirstArgVal = transValue(II->getArgOperand(0), BB);
+    SPIRVValue *SecondArgVal = transValue(II->getArgOperand(1), BB);
+    if (IID == Intrinsic::uadd_sat) {
+      // uadd.sat(a, b) -> res = a + b, (res > a) ? res : MAX
+      SPIRVValue *Max =
+          transValue(Constant::getAllOnesValue(II->getType()), BB);
+      SPIRVValue *Add =
+          BM->addBinaryInst(OpIAdd, Ty, FirstArgVal, SecondArgVal, BB);
+      SPIRVValue *Cmp = BM->addCmpInst(OpUGreaterThan, SPVBoolTy,
+                                       Add, FirstArgVal, BB);
+      return BM->addSelectInst(Cmp, Add, Max, BB);
+    }
+    if (IID == Intrinsic::usub_sat) {
+      // usub.sat(a, b) -> (a > b) ? a - b : 0
+      SPIRVValue *Sub =
+          BM->addBinaryInst(OpISub, Ty, FirstArgVal, SecondArgVal, BB);
+      SPIRVValue *Cmp = BM->addCmpInst(OpUGreaterThan, SPVBoolTy,
+                                       FirstArgVal, SecondArgVal, BB);
+      SPIRVValue *Zero = transValue(Constant::getNullValue(II->getType()), BB);
+      return BM->addSelectInst(Cmp, Sub, Zero, BB);
+    }
+
+    uint64_t NumBits = LLVMTy->getScalarSizeInBits();
+    SPIRVValue *Zero = transValue(Constant::getNullValue(II->getType()), BB);
+    SPIRVValue *Max =
+        transValue(Constant::getIntegerValue(
+              LLVMTy, APInt::getSignedMaxValue(NumBits)), BB);
+    SPIRVValue *Min =
+        transValue(Constant::getIntegerValue(
+              LLVMTy, APInt::getSignedMinValue(NumBits)), BB);
+    SPIRVValue *IsPositive =
+        BM->addCmpInst(OpSGreaterThan, SPVBoolTy, SecondArgVal, Zero, BB);
+    SPIRVValue *IsNegative =
+        BM->addCmpInst(OpSLessThan, SPVBoolTy, SecondArgVal, Zero, BB);
+    if (IID == Intrinsic::sadd_sat) {
+      // sadd.sat(a, b) -> if (b > 0) && a > MAX - b => overflow -> MAX
+      //                -> else if (b < 0) && a < MIN - b => overflow -> MIN
+      //                -> else a + b
+      // There two ways to represent such sequence in IR:
+      // 1. Via 2 branch instructions plus 'phi' instruction;
+      // 2. Via set of select instructions.
+      // The later was chosed because of the following consideration:
+      // speculative branch prediction will most likely consider 'if' statements
+      // here as always false falling through to 'a + b', and reversing it will
+      // case performance degradation.
+      SPIRVValue *Add =
+          BM->addBinaryInst(OpIAdd, Ty, FirstArgVal, SecondArgVal, BB);
+
+      // check if (b > 0) && a > MAX - b condition
+      SPIRVValue *MaxSubB =
+          BM->addBinaryInst(OpISub, Ty, Max, SecondArgVal, BB);
+      SPIRVValue *CanPosOverflow = BM->addCmpInst(OpSGreaterThan, SPVBoolTy,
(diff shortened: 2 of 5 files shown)


MrSidims:
@LU-JOHN @asudarsa @maksimsab please take a look

lib/SPIRV/SPIRVWriter.cpp
```diff
@@ -3860,21 +3860,127 @@ SPIRVValue *LLVMToSPIRVBase::transIntrinsicInst(IntrinsicInst *II,
...
+                            std::move(Operands), BB);
+    }
+    Type *LLVMTy = II->getType();
+    SPIRVType *Ty = transType(LLVMTy);
+    Type *BoolTy = IntegerType::getInt1Ty(M->getContext());
+    if (auto *VecTy = dyn_cast<VectorType>(LLVMTy))
+      BoolTy = VectorType::get(BoolTy, VecTy->getElementCount());
+    SPIRVType *SPVBoolTy = transType(BoolTy);
+    SPIRVValue *FirstArgVal = transValue(II->getArgOperand(0), BB);
+    SPIRVValue *SecondArgVal = transValue(II->getArgOperand(1), BB);
```
MrSidims, line 3888: should add an empty line to improve readability

test/llvm-intrinsics/add_sub.sat.ll
```diff
@@ -1,6 +1,16 @@
 ; RUN: llvm-as < %s -o %t.bc
 ; RUN: llvm-spirv %t.bc -o %t.spv
-; RUN: llvm-spirv %t.spv -to-text -o - | FileCheck %s
+; RUN: llvm-spirv %t.spv -to-text -o - | FileCheck %s --check-prefixes=COMMON,OPENCL
+; RUN: spirv-val %t.spv
+
+; RUN: llvm-as < %s -o %t.bc
```
vmaksimo, line 6: excessive - we've already assembled the *.ll file, no need to assemble again (here and on line 11)

lib/SPIRV/SPIRVWriter.cpp
```diff
@@ -3860,21 +3860,130 @@ SPIRVValue *LLVMToSPIRVBase::transIntrinsicInst(IntrinsicInst *II,
...
+
+    if (IID == Intrinsic::uadd_sat) {
+      // uadd.sat(a, b) -> res = a + b, (res > a) ? res : MAX
```
vmaksimo, line 3892: Is `res` will anyway be bigger or equal than `a`? I thought maybe the operation is between 2 unsigned integers

MrSidims: If we have an integer overflow - then no. For unsigned integers it's the defined behavior.

test/llvm-intrinsics/add_sub.sat.ll
MrSidims: oh, indeed, thanks!

lib/SPIRV/SPIRVWriter.cpp
vmaksimo: I understood the intrinsic description differently, here is info from LLVM lang ref (https://llvm.org/docs/LangRef.html#llvm-uadd-sat-intrinsics):

> The maximum value this operation can clamp to is the largest unsigned value representable by the bit width of the arguments. Because this is an unsigned operation, the result will never saturate towards zero.
`%res = call i4 @llvm.uadd.sat.i4(i4 8, i4 8)  ; %res = 15`

MrSidims: Okay, it seems like I misunderstood the question. I answer like it was a general question, but it's related to the intrinsic. Intrinsic acts like it's described in the comment: picking between 'res' and 'MAX' depending if overflow has happen. So emulation should also pick between them. But detecting overflow is being done by the check if `a + b > a`, which can only happen during overflow.

MrSidims: Aahh, I see, oops, the code and the comment do a bit an opposite.

asudarsa: In cases where b = 0; res > a will return false and hence MAX will be selected. I think that is not the correct answer. I think the checking condition needs to be res >= a.
Please revisit.

Thanks

lib/SPIRV/SPIRVWriter.cpp
```diff
@@ -3860,21 +3860,130 @@ SPIRVValue *LLVMToSPIRVBase::transIntrinsicInst(IntrinsicInst *II,
...
+      SPIRVValue *Max =
+          transValue(Constant::getAllOnesValue(II->getType()), BB);
+      SPIRVValue *Add =
+          BM->addBinaryInst(OpIAdd, Ty, FirstArgVal, SecondArgVal, BB);
+      SPIRVValue *Cmp =
+          BM->addCmpInst(OpUGreaterThan, SPVBoolTy, Add, FirstArgVal, BB);
```
asudarsa, line 3898: I think OpUGreaterThan is the right op. But, I am not able to reason out with an example. Can you please provide one if possible? Not a PR blocker.

Thanks

lib/SPIRV/SPIRVWriter.cpp
```diff
@@ -3860,21 +3860,130 @@ SPIRVValue *LLVMToSPIRVBase::transIntrinsicInst(IntrinsicInst *II,
...
+      // sadd.sat(a, b) -> if (b > 0) && a > MAX - b => overflow -> MAX
+      //                -> else if (b < 0) && a < MIN - b => overflow -> MIN
+      //                -> else a + b
+      // There two ways to represent such sequence in IR:
+      // 1. Via 2 branch instructions plus 'phi' instruction;
+      // 2. Via set of select instructions.
+      // The later was chosed because of the following consideration:
+      // speculative branch prediction will most likely consider 'if' statements
+      // here as always false falling through to 'a + b', and reversing it will
+      // case performance degradation.
```
asudarsa, line 3934:
```suggestion
      // cause performance degradation.
```

lib/SPIRV/SPIRVWriter.cpp
```diff
@@ -3860,21 +3860,130 @@ SPIRVValue *LLVMToSPIRVBase::transIntrinsicInst(IntrinsicInst *II,
...
+      SPIRVValue *Add =
+          BM->addBinaryInst(OpIAdd, Ty, FirstArgVal, SecondArgVal, BB);
+
+      // check if (b > 0) && a > MAX - b condition
+      SPIRVValue *MaxSubB =
+          BM->addBinaryInst(OpISub, Ty, Max, SecondArgVal, BB);
+      SPIRVValue *CanPosOverflow =
+          BM->addCmpInst(OpSGreaterThan, SPVBoolTy, FirstArgVal, MaxSubB, BB);
+      SPIRVValue *PosOverflow = BM->addInstTemplate(
```
asudarsa, line 3943: Not sure if the order of the operand could impact performance here. Again just a request for quick check. Not a PR blocker.

Thanks

test/llvm-intrinsics/add_sub.sat.ll
```diff
@@ -1,6 +1,16 @@
...
+; RUN: llvm-spirv %t.bc --spirv-use-ocl-math-for-llvm-intrinsic=true -o %t.spv
+; RUN: llvm-spirv %t.spv -to-text -o - | FileCheck %s --check-prefixes=COMMON,OPENCL
+; RUN: spirv-val %t.spv
+
+; RUN: llvm-as < %s -o %t.bc
+; RUN: llvm-spirv %t.bc --spirv-use-ocl-math-for-llvm-intrinsic=false -o %t.spv
+; RUN: llvm-spirv %t.spv -to-text -o - | FileCheck %s --check-prefixes=COMMON,EMULATION
 ; RUN: spirv-val %t.spv
 
 ; Test checks that saturation addition and substraction llvm intrinsics
```
asudarsa, line 16: Out of the scope of this change. But just pointing out a typo:
```suggestion
; Test checks that saturation addition and subtraction llvm intrinsics
```

asudarsa (changes requested):
One possible correction required. Please address. Thanks for this change.

asudarsa:
Title can be: Improve translation support for (s/u)_(add/sub)_sat intrinsics

tools/llvm-spirv/llvm-spirv.cpp
```diff
@@ -258,6 +258,12 @@ static cl::opt<SPIRV::BuiltinFormat> SPIRVBuiltinFormat(
         clEnumValN(SPIRV::BuiltinFormat::Global, "global",
                    "Use globals to represent SPIR-V builtin variables")));
 
+static cl::opt<bool> SPIRVUseOpenCLExtInstructionsForLLVMIntrinsic(
```
asudarsa, line 261: I feel that this option sounds very generic. It sounds as if every llvm intrinsic is covered under this option. Can you please elaborate?

Thanks

lib/SPIRV/SPIRVWriter.cpp
MrSidims: It should be `res < a` as overflow condition.

MrSidims: I was a bit confused answering, but basically @asudarsa is right, thanks!

lib/SPIRV/SPIRVWriter.cpp
MrSidims: Should be OpUGreaterThanEqual according to your comment from above
I've also checked generated IR after the translation by running inst combine - it's able to restore uadd_sat intrinsic

tools/llvm-spirv/llvm-spirv.cpp
MrSidims: Renamed to SPIRVUseOpenCLExtInstructionsForLLVMMargIntrinsic

lib/SPIRV/SPIRVWriter.cpp
MrSidims: TBH I'm not expecting this implementation be performant at all. Ideally we should generate such code that is collapsing to sadd/sub_sat intrinsic after we run InstCombine , but it's not happening. That is because InstCombine checks overflow by extending integer to a larger one and comparing the result with MAX/MIN of the previous type. We can't do this in general case, because we can't cast integer to a wider one in SPIR-V without extensions.

tools/llvm-spirv/llvm-spirv.cpp
```diff
@@ -258,6 +258,12 @@ static cl::opt<SPIRV::BuiltinFormat> SPIRVBuiltinFormat(
         clEnumValN(SPIRV::BuiltinFormat::Global, "global",
                    "Use globals to represent SPIR-V builtin variables")));
 
+static cl::opt<bool> SPIRVUseOpenCLExtInstructionsForLLVMMathIntrinsic(
+    "spirv-use-ocl-for-llvm-math-intrinsic", cl::init(true),
```
MrSidims, line 262: @svenvh may I ask you to take a look if this option is OK?

svenvh: I wonder if we can generalize the option?  If I understand correctly, you essentially want to control whether the translator is allowed to emit `OpenCL.std OpExtInst` instructions?

svenvh:
> I don't really want to end up implementing Quake's sqrt algorithm, but it's a possible scenario as well.

The prospect of growing a math library inside llvm-spirv doesn't sound exciting indeed... ;-)

tools/llvm-spirv/llvm-spirv.cpp
MrSidims: Yeah, exactly. So is your suggesting to rename it or something different? We kinda can find counterparts of https://registry.khronos.org/SPIR-V/specs/unified1/OpenCL.ExtendedInstructionSet.100.html in https://llvm.org/docs/LangRef.html#standard-c-c-library-intrinsics , so would it be `spirv-use-std-ocl-for-llvm-std-intrinsics` ?

MrSidims: Mmm, I see. I will spend some time thinking about it, I'd prefer not to add an option (and enable it in some compiler's frontend driver), only to remove it in a couple of weeks, replacing with a better solution :) So we can keep PR open for a while

lib/SPIRV/SPIRVWriter.cpp
```diff
@@ -3860,21 +3860,130 @@ SPIRVValue *LLVMToSPIRVBase::transIntrinsicInst(IntrinsicInst *II,
...
+    SPIRVValue *Min = transValue(
+        Constant::getIntegerValue(LLVMTy, APInt::getSignedMinValue(NumBits)),
+        BB);
+    SPIRVValue *IsPositive =
+        BM->addCmpInst(OpSGreaterThan, SPVBoolTy, SecondArgVal, Zero, BB);
+    SPIRVValue *IsNegative =
+        BM->addCmpInst(OpSLessThan, SPVBoolTy, SecondArgVal, Zero, BB);
+
+    if (IID == Intrinsic::sadd_sat) {
+      // sadd.sat(a, b) -> if (b > 0) && a > MAX - b => overflow -> MAX
```
LU-JOHN, line 3925: Can we use this sequence:

res = a+b
if (b>0 && a>res) res = MAX
if (b<0 && a>res) res = MIN

This makes it unnecessary to calculate "MAX - b" and "MIN - b"

LU-JOHN:
> 2\. introduce InstCombine pass in reverse translation under an option.

Some of the emulation sequences are long.  Could they be modified (e.g. if we know an addend is always positive) or re-ordered so that InstCombine cannot recognize the sequence anymore?

lib/SPIRV/SPIRVWriter.cpp
```diff
@@ -3860,21 +3860,130 @@ SPIRVValue *LLVMToSPIRVBase::transIntrinsicInst(IntrinsicInst *II,
...
+          BM->addCmpInst(OpSGreaterThan, SPVBoolTy, MinSubB, FirstArgVal, BB);
+      SPIRVValue *NegOverflow = BM->addInstTemplate(
+          OpLogicalAnd, {CanNegOverflow->getId(), IsNegative->getId()}, BB,
+          SPVBoolTy);
+
+      // do selects
+      SPIRVValue *FirstSelect = BM->addSelectInst(PosOverflow, Max, Add, BB);
+      return BM->addSelectInst(NegOverflow, Min, FirstSelect, BB);
+    }
+    // ssub.sat(a, b) -> if (b > 0) && a < MIN + b => overflow -> MIN
```
LU-JOHN, line 3960: This can be:

res = a-b
if (b<0 && a>res) res = MAX
if (b>0 && a<res) res = MIN

This makes it unnecessary to calculate "MAX + b" and "MIN + b"

lib/SPIRV/SPIRVWriter.cpp
LU-JOHN: Another implementation with a single comparison:

int sat_add(int a, int b) {
  int sum = a+b;

  if ((// same sign
       ~(a^b)   &
       //different sign
       (sum^a))
      // test MSB
      < 0) {
    sum = (0x7fffffff^(a>>31));
  }

  return sum;
}

include/LLVMSPIRVOpts.h
```diff
@@ -193,6 +193,14 @@ class TranslatorOpts {
     ReplaceLLVMFmulAddWithOpenCLMad = Value;
   }
 
+  void setUseOpenCLExtInstructionsForLLVMMathIntrinsic(bool Value) noexcept {
```
maksimsab, line 196: nit regarding `noexcept`:
LLVM itself prohibits [C++ exceptions](https://llvm.org/docs/CodingStandards.html#do-not-use-rtti-or-exceptions) and they are turned off in CMakeLists by compiler flag `-fno-exceptions`.
If we use exceptions in translator, then it would be better to stop doing it and start using the compiler flag `-fno-exceptions` as well.

Issue #2139, opened 2023-08-25
For `OpBitReverse`, `Shader` capability is used even if `SPV_KHR_bit_instructions` is enabled


fxzjshm:
Related discussion: <https://gitlab.freedesktop.org/mesa/mesa/-/issues/9690>

A piece of LLVM IR like
```llvm
%10 = call i32 @llvm.bitreverse.i32(i32 %1)
```
is translated into
```spirv
OpCapability Shader
...
OpCapability Kernel
...
%25 = OpBitReverse %uint %11
```

but with `SPV_KHR_bit_instructions` extension enabled,
```spirv
OpCapability Shader
...
OpCapability Kernel
...
OpCapability BitInstructions
OpExtension "SPV_KHR_bit_instructions"
...
%25 = OpBitReverse %uint %11
```
the `Shader` capability is still there, which seems not suitable for OpenCL.
Is this expected behavior?

<details><summary>Complete example code</summary>
<p>

```llvm
; ModuleID = 'bit_reverse-spirv.bc'
target datalayout = "e-p:64:64:64-i1:8:8-i8:8:8-i16:16:16-i32:32:32-i64:64:64-f32:32:32-f64:64:64-v16:16:16-v24:32:32-v32:32:32-v48:64:64-v64:64:64-v96:128:128-v128:128:128-v192:256:256-v256:256:256-v512:512:512-v1024:1024:1024"
target triple = "spir64-unknown-unknown"

%__hipsycl_sscp_pointer_wrapper.1 = type { ptr addrspace(1) }

; Function Attrs: nounwind
define spir_kernel void @"_Z21__hipsycl_sscp_kernelIN7hipsycl4glue15__sscp_dispatch11single_taskIZ4mainE3$_0EEEvRKT_"(ptr nocapture readonly byval(%__hipsycl_sscp_pointer_wrapper.1) %0, i32 %1, i8 %2, i8 %3, i8 %4, i8 %5) #0 !kernel_arg_addr_space !5 !kernel_arg_access_qual !6 !kernel_arg_type !7 !kernel_arg_type_qual !8 !kernel_arg_base_type !7 !spirv.ParameterDecorations !9 {
  %7 = bitcast ptr %0 to ptr
  %8 = load ptr addrspace(1), ptr %7, align 8
  %9 = addrspacecast ptr addrspace(1) %8 to ptr addrspace(4)
  %10 = call i32 @llvm.bitreverse.i32(i32 %1)
[...]

PR #2141, opened 2023-08-30
Only require Shader capability if unavoidable

svenvh:
Requiring the `Shader` capability for any module that uses `llvm.bitreverse` is problematic for OpenCL execution environments. Only require the `Shader` capability if we cannot use the `BitInstructions` capability.

Fixes https://github.com/KhronosGroup/SPIRV-LLVM-Translator/issues/2139

### DIFF
--- a/lib/SPIRV/SPIRVWriter.cpp
+++ b/lib/SPIRV/SPIRVWriter.cpp
@@ -3615,7 +3615,9 @@ SPIRVValue *LLVMToSPIRVBase::transIntrinsicInst(IntrinsicInst *II,
     return nullptr;
   }
   case Intrinsic::bitreverse: {
-    BM->addCapability(CapabilityShader);
+    if (!BM->isAllowedToUseExtension(ExtensionID::SPV_KHR_bit_instructions)) {
+      BM->addCapability(CapabilityShader);
+    }
     SPIRVType *Ty = transType(II->getType());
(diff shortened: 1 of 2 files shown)


MrSidims (approved):

lib/SPIRV/SPIRVWriter.cpp
```diff
@@ -3615,7 +3615,9 @@ SPIRVValue *LLVMToSPIRVBase::transIntrinsicInst(IntrinsicInst *II,
     return nullptr;
   }
   case Intrinsic::bitreverse: {
-    BM->addCapability(CapabilityShader);
+    if (!BM->isAllowedToUseExtension(ExtensionID::SPV_KHR_bit_instructions)) {
+      BM->addCapability(CapabilityShader);
```
MrSidims, line 3619: @svenvh should we remove this handling after https://github.com/KhronosGroup/SPIRV-LLVM-Translator/pull/2270 ?

svenvh: This seems to be the only (remaining) case where we emit `CapabilityShader` currently, so it makes sense to remove it indeed.  I'll open a PR shortly.

svenvh: https://github.com/KhronosGroup/SPIRV-LLVM-Translator/pull/2287

PR #2151, opened 2023-09-08
Fix check for (neg) zero for fpclass emulation

MrSidims:
We should compare not only to zero integer, but also 'negated' zero.

### DIFF
--- a/lib/SPIRV/SPIRVWriter.cpp
+++ b/lib/SPIRV/SPIRVWriter.cpp
@@ -4432,14 +4432,14 @@ SPIRVValue *LLVMToSPIRVBase::transIntrinsicInst(IntrinsicInst *II,
             GetNegPosInstTest(TestIsSubnormal, FPClass & fcNegSubnormal)));
     }
     if (FPClass & fcZero) {
-      // Create zero integer constant and check for equality with bitcasted to
+      // Create zero integer constant and check for equality with converted to
       // int float value
-      auto *BitCastToInt =
-          BM->addUnaryInst(OpBitcast, OpSPIRVTy, InputFloat, BB);
+      auto *Convert =
+          BM->addUnaryInst(OpConvertFToU, OpSPIRVTy, InputFloat, BB);
       auto *ZeroConst = transValue(
           Constant::getIntegerValue(IntOpLLVMTy, APInt::getZero(BitSize)), BB);
       auto *TestIsZero =
-          BM->addCmpInst(OpIEqual, ResTy, BitCastToInt, ZeroConst, BB);
+          BM->addCmpInst(OpIEqual, ResTy, Convert, ZeroConst, BB);
       if (FPClass & fcPosZero && FPClass & fcNegZero)
         ResultVec.emplace_back(GetInvertedTestIfNeeded(TestIsZero));
       else
--- a/lib/SPIRV/SPIRVWriter.cpp
+++ b/lib/SPIRV/SPIRVWriter.cpp
@@ -4434,8 +4434,8 @@ SPIRVValue *LLVMToSPIRVBase::transIntrinsicInst(IntrinsicInst *II,
     if (FPClass & fcZero) {
       // Create zero integer constant and check for equality with bitcasted to
       // int float value
-      auto SetUpCMPToZero = [&](SPIRVValue *BitCastToInt, bool IsPositive)
-        -> SPIRVValue * {
+      auto SetUpCMPToZero = [&](SPIRVValue *BitCastToInt,
+                                bool IsPositive) -> SPIRVValue * {
         APInt ZeroInt = APInt::getZero(BitSize);
         if (IsPositive) {
           auto *ZeroConst =
--- a/lib/SPIRV/SPIRVWriter.cpp
+++ b/lib/SPIRV/SPIRVWriter.cpp
@@ -4432,19 +4432,35 @@ SPIRVValue *LLVMToSPIRVBase::transIntrinsicInst(IntrinsicInst *II,
             GetNegPosInstTest(TestIsSubnormal, FPClass & fcNegSubnormal)));
     }
     if (FPClass & fcZero) {
-      // Create zero integer constant and check for equality with converted to
+      // Create zero integer constant and check for equality with bitcasted to
       // int float value
-      auto *Convert =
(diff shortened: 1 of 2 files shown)


MrSidims:
@LU-JOHN please take a look

lib/SPIRV/SPIRVWriter.cpp
```diff
@@ -4432,14 +4432,14 @@ SPIRVValue *LLVMToSPIRVBase::transIntrinsicInst(IntrinsicInst *II,
             GetNegPosInstTest(TestIsSubnormal, FPClass & fcNegSubnormal)));
     }
     if (FPClass & fcZero) {
-      // Create zero integer constant and check for equality with bitcasted to
+      // Create zero integer constant and check for equality with converted to
       // int float value
```
svenvh, line 4436: Not sure if that works correctly for e.g. an input value `0.123`, as `OpConvertFToU` converts that to `i32 0` so that the compare with 0 succeeds.  Effectively making the emulated fpclass report that `0.123` is zero.  Or did I miss something?

MrSidims: You are right, back to the drawing board

lib/SPIRV/SPIRVWriter.cpp
```diff
@@ -4436,15 +4436,30 @@ SPIRVValue *LLVMToSPIRVBase::transIntrinsicInst(IntrinsicInst *II,
...
       auto *BitCastToInt =
           BM->addUnaryInst(OpBitcast, OpSPIRVTy, InputFloat, BB);
-      auto *ZeroConst = transValue(
-          Constant::getIntegerValue(IntOpLLVMTy, APInt::getZero(BitSize)), BB);
-      auto *TestIsZero =
-          BM->addCmpInst(OpIEqual, ResTy, BitCastToInt, ZeroConst, BB);
-      if (FPClass & fcPosZero && FPClass & fcNegZero)
+      APInt ZeroInt = APInt::getZero(BitSize);
+      auto *ZeroConst =
```
MrSidims, line 4440: mmm, actually, should also move the under ifs

lib/SPIRV/SPIRVWriter.cpp
```diff
@@ -4434,17 +4434,40 @@ SPIRVValue *LLVMToSPIRVBase::transIntrinsicInst(IntrinsicInst *II,
...
+      auto SetUpCMPToZero = [&](SPIRVValue *BitCastToInt,
+                                bool IsPositive) -> SPIRVValue * {
+        APInt ZeroInt = APInt::getZero(BitSize);
+        if (IsPositive) {
+          auto *ZeroConst =
+              transValue(Constant::getIntegerValue(IntOpLLVMTy, ZeroInt), BB);
+          return BM->addCmpInst(OpIEqual, ResTy, BitCastToInt, ZeroConst, BB);
+        }
+        // Created 'negated' zero
+        ZeroInt.setSignBit();
```
svenvh, line 4446: As I understood APInt uses 2's complement, which does not have negative zero.  Are you sure this is giving the correct result?

lib/SPIRV/SPIRVWriter.cpp
```diff
@@ -4434,17 +4434,40 @@ SPIRVValue *LLVMToSPIRVBase::transIntrinsicInst(IntrinsicInst *II,
...
+        return BM->addCmpInst(OpIEqual, ResTy, BitCastToInt, NegZeroConst, BB);
+      };
       auto *BitCastToInt =
           BM->addUnaryInst(OpBitcast, OpSPIRVTy, InputFloat, BB);
-      auto *ZeroConst = transValue(
-          Constant::getIntegerValue(IntOpLLVMTy, APInt::getZero(BitSize)), BB);
-      auto *TestIsZero =
-          BM->addCmpInst(OpIEqual, ResTy, BitCastToInt, ZeroConst, BB);
-      if (FPClass & fcPosZero && FPClass & fcNegZero)
+      if (FPClass & fcPosZero && FPClass & fcNegZero) {
```
LU-JOHN, line 4453: Instead of generating:

if (val==+0 || val==-0)

we might get better performance with:

if ((val&0x7fff...)==0)

lib/SPIRV/SPIRVWriter.cpp
MrSidims: [this line in the test](https://github.com/KhronosGroup/SPIRV-LLVM-Translator/pull/2151/files#diff-6e15e8a39c298484201aa544ea77f67acd2acae0c091213a6e2c09089d2e0ea2R24) added after setting signed bit seems about right

svenvh: Right, it works because you're not doing any (2's complement) operations on the APInt.

asudarsa: Hi @svenvh 

Thanks for the review.
Is the use of APInt agreeable in this scenario, or do we have to change the logic here?

Thanks

lib/SPIRV/SPIRVWriter.cpp
asudarsa: We need both val == 0 and val == -0 results. So, it is worthwhile to generate both.

Thanks

lib/SPIRV/SPIRVWriter.cpp
svenvh: This becomes a moot point once the suggestion from @LU-JOHN below has been applied.

lib/SPIRV/SPIRVWriter.cpp
asudarsa: I think I might be missing something here. How will val & 0x&7fffffff give result for TestIsPosZero(..) and TestIsNegZero(..)? Is the idea here to replace val == 0 with (val & 0xFFFFFFFFF == 0) and val == -0 with (val && 0xFFFFFFFF != 0) && (val && 0x7FFFFFFF == 0)?
Can @svenvh or @LU-JOHN, please clarify? Thanks

LU-JOHN: +0.0 is represented by 0x0000.0000 (for floats)
-0.0 is represented by 0x8000.0000 (for floats)
Instead of doing two tests we can check for both by doing:
(val & 0x7FFF.FFFF)==0x0000.0000

svenvh:
Closing as this was superseded by #2154

PR #2152, opened 2023-09-08
Manually convert tests to opaque pointers

svenvh:
Manually convert some tests to opaque pointers that the migration script did not handle correctly.

### DIFF
--- a/test/DebugInfo/Generic/c-and-cpp-mixed.ll
+++ b/test/DebugInfo/Generic/c-and-cpp-mixed.ll
@@ -70,18 +70,18 @@ target triple = "spir64-unknown-unknown"
 
 define void @foo() nounwind !dbg !5 {
 entry:
-  %puts = tail call i32 @puts(i8* getelementptr inbounds ([4 x i8], [4 x i8]* @str, i32 0, i32 0)), !dbg !23
+  %puts = tail call i32 @puts(ptr @str), !dbg !23
   ret void, !dbg !25
 }
 
-declare i32 @puts(i8* nocapture) nounwind
+declare i32 @puts(ptr nocapture) nounwind
 
-define i32 @main(i32 %argc, i8** nocapture %argv) nounwind !dbg !12 {
+define i32 @main(i32 %argc, ptr nocapture %argv) nounwind !dbg !12 {
 entry:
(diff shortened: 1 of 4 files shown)


test/transcoding/spirv-private-array-initialization.ll
```diff
@@ -43,15 +43,15 @@ define spir_func void @test() #0 {
 entry:
   %arr = alloca [3 x i32], align 4
   %arr2 = alloca [3 x i32], align 4
-  %0 = bitcast [3 x i32]* %arr to i8*
-  call void @llvm.memcpy.p0i8.p2i8.i32(i8* align 4 %0, i8 addrspace(2)* align 4 bitcast ([3 x i32] addrspace(2)* @__const.test.arr to i8 addrspace(2)*), i32 12, i1 false)
-  %1 = bitcast [3 x i32]* %arr2 to i8*
-  call void @llvm.memcpy.p0i8.p2i8.i32(i8* align 4 %1, i8 addrspace(2)* align 4 bitcast ([3 x i32] addrspace(2)* @__const.test.arr2 to i8 addrspace(2)*), i32 12, i1 false)
+  %0 = bitcast ptr %arr to ptr
```
vmaksimo, line 46: do we need this bitcast?

test/DebugInfo/Generic/c-and-cpp-mixed.ll
```diff
@@ -70,18 +70,18 @@ target triple = "spir64-unknown-unknown"
 
 define void @foo() nounwind !dbg !5 {
 entry:
-  %puts = tail call i32 @puts(i8* getelementptr inbounds ([4 x i8], [4 x i8]* @str, i32 0, i32 0)), !dbg !23
+  %puts = tail call i32 @puts(ptr @str), !dbg !23
```
vmaksimo, line 73: just curious, in other tests `getelementptr` instructions are just modified, not totally removed like here. why so?

test/transcoding/spirv-private-array-initialization.ll
svenvh: The test is explicitly checking for the bitcasts (see lines 22 & 25), so I figured we might want to keep these bitcasts as it tests that we recover the correct types.

vmaksimo (approved):
Thanks for the explanations, seems good to go :)

PR #2154, opened 2023-09-12
Fix check for (neg) zero for fpclass emulation

asudarsa:
We should compare not only to zero integer, but also 'negated' zero.

This PR is built in top of: Fix check for (neg) zero for fpclass emulation #2151

### DIFF
--- a/lib/SPIRV/SPIRVWriter.cpp
+++ b/lib/SPIRV/SPIRVWriter.cpp
@@ -4432,14 +4432,14 @@ SPIRVValue *LLVMToSPIRVBase::transIntrinsicInst(IntrinsicInst *II,
             GetNegPosInstTest(TestIsSubnormal, FPClass & fcNegSubnormal)));
     }
     if (FPClass & fcZero) {
-      // Create zero integer constant and check for equality with bitcasted to
+      // Create zero integer constant and check for equality with converted to
       // int float value
-      auto *BitCastToInt =
-          BM->addUnaryInst(OpBitcast, OpSPIRVTy, InputFloat, BB);
+      auto *Convert =
+          BM->addUnaryInst(OpConvertFToU, OpSPIRVTy, InputFloat, BB);
       auto *ZeroConst = transValue(
           Constant::getIntegerValue(IntOpLLVMTy, APInt::getZero(BitSize)), BB);
       auto *TestIsZero =
-          BM->addCmpInst(OpIEqual, ResTy, BitCastToInt, ZeroConst, BB);
+          BM->addCmpInst(OpIEqual, ResTy, Convert, ZeroConst, BB);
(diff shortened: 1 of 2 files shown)


asudarsa:
Hi @svenvh and @LU-JOHN 

Sorry for inconvenience. Can we please continue the review of PR #2151 here? My last commit introduces a bitwise and as suggested by John.

Thanks

lib/SPIRV/SPIRVWriter.cpp
```diff
@@ -4434,17 +4434,42 @@ SPIRVValue *LLVMToSPIRVBase::transIntrinsicInst(IntrinsicInst *II,
...
-      auto *ZeroConst = transValue(
-          Constant::getIntegerValue(IntOpLLVMTy, APInt::getZero(BitSize)), BB);
-      auto *TestIsZero =
-          BM->addCmpInst(OpIEqual, ResTy, BitCastToInt, ZeroConst, BB);
-      if (FPClass & fcPosZero && FPClass & fcNegZero)
+      if (FPClass & fcPosZero && FPClass & fcNegZero) {
+        APInt ZeroInt = APInt::getZero(BitSize);
+        auto *ZeroConst =
+            transValue(Constant::getIntegerValue(IntOpLLVMTy, ZeroInt), BB);
+        APInt SignedMaxInt = APInt::getSignedMaxValue(BitSize);
```
LU-JOHN, line 4457: Rename variable and/or add comment that this is used as a mask to mask-off floating-point sign-bit.   It is coincidentally signed-max but that is not the essential property in this code.

test/llvm-intrinsics/fpclass.ll
```diff
@@ -21,6 +21,8 @@
 ; CHECK-SPIRV-DAG: Constant [[#Int32Ty]] [[#QNanBitConst:]] 2143289344
 ; CHECK-SPIRV-DAG: Constant [[#Int32Ty]] [[#MantissaConst:]] 8388607
 ; CHECK-SPIRV-DAG: Constant [[#Int32Ty]] [[#ZeroConst:]] 0
+; CHECK-SPIRV-DAG: Constant [[#Int32Ty]] [[#MaxConst:]] 2147483647
```
LU-JOHN, line 24: Perhaps a better name than MaxConst would be MaskToClearFPSign.

lib/SPIRV/SPIRVWriter.cpp
asudarsa: Thanks. I changed the variable names.

vmaksimo (approved):

svenvh (approved):

LU-JOHN (approved):

asudarsa:
Thanks everyone for review.

PR #2158, opened 2023-09-18
Revert "Allow cooperative matrix types for OpMatrixTimesScalar"

vmaksimo:
Per discussion in Khronos and the spec itself, `OpMatrixTimesScalar` should not be used in OpenCL environment, as long as the same can be achieved either via an elementwise multiplication with a cooperative matrix object created from the scalar using `OpConstantComposite` or by iterating over the elements of the cooperative matrix to multiply each element by the scalar.

Spec: https://github.com/KhronosGroup/SPIRV-Registry/blob/main/extensions/KHR/SPV_KHR_cooperative_matrix.asciidoc


MrSidims (approved):

MrSidims:
Withdrawing my approval. Arvind has a good comment. Realistically we do not target non-Kernel SPIR-V by the translator. But i doesn't mean we should hold the translation of such instruction. Ideally we should be checking if we have Kernel capability in the module and error out if it's true for MatrixTimesScalar, but we don't have such API yet. So probably we should revert the patch anyway.

MrSidims:
> Realistically we do not target non-Kernel SPIR-V by the translator. 

Now with https://github.com/KhronosGroup/SPIRV-LLVM-Translator/pull/2270 we have such restriction set in stone. @vmaksimo please rebase the PR :)

MrSidims (approved):

vmaksimo:
Continue in #2276, I have no rights to push to this PR

PR #2160, opened 2023-09-19
Fix SPIR-V global to function replacement for differing load types

sarnex:
In some cases, we will see IR with the following

```
@__spirv_BuiltInGlobalInvocationId = external dso_local local_unnamed_addr addrspace(1) constant <3 x i64>, align 32

...

%0 = load <6 x i32>, ptr addrspace(1) @__spirv_BuiltInGlobalInvocationId, align 32
%1 = extractelement <6 x i32> %0, i64 0
```

Note the global type and load type are different. Change the handling of vector loads from vector globals to reconstruct the global vector type and then bitcast to the load type.

Thanks to @jcranmer-intel for helping me find the simplest solution.

### DIFF
--- a/lib/SPIRV/SPIRVUtil.cpp
+++ b/lib/SPIRV/SPIRVUtil.cpp
@@ -2029,7 +2029,8 @@ static void replaceUsesOfBuiltinVar(Value *V, const APInt &AccumulatedOffset,
           // the load type may not match.
           // global <3 x i64>, load <6 x i32>
           VecTy = cast<FixedVectorType>(GV->getValueType());
-          if (!Index.isZero())
+          if (!Index.isZero() || DL.getTypeSizeInBits(VecTy) !=
+                                     DL.getTypeSizeInBits(Load->getType()))
             llvm_unreachable("Illegal use of a SPIR-V builtin variable");
           Replacement = UndefValue::get(VecTy);
           for (unsigned I = 0; I < VecTy->getNumElements(); I++) {
--- a/lib/SPIRV/SPIRVUtil.cpp
+++ b/lib/SPIRV/SPIRVUtil.cpp
@@ -2040,8 +2040,9 @@ static void replaceUsesOfBuiltinVar(Value *V, const APInt &AccumulatedOffset,
                 Builder.getInt32(I));
           }
           // Insert a bitcast from the reconstructed vector to the load vector
-          // type in case they are different. %2 = insertelement <3 x i64> %0,
-          // i64 %1, i32 2 bitcast <3 x i64> %2 to <6 x i32>
+          // type in case they are different.
+	  // %2 = insertelement <3 x i64> %0, i64 %1, i32 2
+	  // bitcast <3 x i64> %2 to <6 x i32>
           Replacement = Builder.CreateBitCast(Replacement, Load->getType());
         } else if (Load->getType() == ScalarTy) {
           Replacement = setAttrByCalledFunc(Builder.CreateCall(
--- a/lib/SPIRV/SPIRVUtil.cpp
+++ b/lib/SPIRV/SPIRVUtil.cpp
@@ -1989,19 +1989,20 @@ bool isSPIRVBuiltinVariable(GlobalVariable *GV,
 /// are accumulated in the AccumulatedOffset parameter, which will eventually be
 /// used to figure out which index of a variable is being used.
 static void replaceUsesOfBuiltinVar(Value *V, const APInt &AccumulatedOffset,
-                                    Function *ReplacementFunc) {
+                                    Function *ReplacementFunc,
+                                    GlobalVariable *GV) {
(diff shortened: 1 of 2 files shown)


sarnex:
@jcranmer-intel Looks like I don't have permission to add reviewers, but here's the PR.

lib/SPIRV/SPIRVUtil.cpp
```diff
@@ -2024,6 +2025,10 @@ static void replaceUsesOfBuiltinVar(Value *V, const APInt &AccumulatedOffset,
       } else {
         // The function has an index parameter.
         if (auto *VecTy = dyn_cast<FixedVectorType>(Load->getType())) {
+          // Reconstruct the original global variable vector because
+          // the load type may not match.
+          // global <3 x i64>, load <6 x i32>
+          VecTy = cast<FixedVectorType>(GV->getValueType());
```
jcranmer-intel, line 2031: I would also add in an `llvm_unreachable` call here if the Load and GV types aren't the same size.

sarnex: Added in latest commit, thanks!

lib/SPIRV/SPIRVUtil.cpp
```diff
@@ -2029,7 +2029,8 @@ static void replaceUsesOfBuiltinVar(Value *V, const APInt &AccumulatedOffset,
           // the load type may not match.
           // global <3 x i64>, load <6 x i32>
           VecTy = cast<FixedVectorType>(GV->getValueType());
-          if (!Index.isZero())
+          if (!Index.isZero() || DL.getTypeSizeInBits(VecTy) !=
+                                     DL.getTypeSizeInBits(Load->getType()))
```
jcranmer-intel, line 2033: Can you run clang-format on your patch? Pretty sure this isn't how clang-format would format this if statement...

sarnex: I did run clang format and also the CI said it's fine

```
 Check code style / clang-format & clang-tidy (pull_request) Successful in 1m 
```

I just ran it again and it seems right:
```
$ wget -q https://patch-diff.githubusercontent.com/raw/KhronosGroup/SPIRV-LLVM-Translator/pull/2160.patch
$ patch -p1 < 2160.patch 
patching file lib/SPIRV/SPIRVUtil.cpp
patching file test/transcoding/builtin_vars_different_type.ll
patching file lib/SPIRV/SPIRVUtil.cpp
patching file lib/SPIRV/SPIRVUtil.cpp
patching file lib/SPIRV/SPIRVUtil.cpp
$ git add lib/SPIRV/SPIRVUtil.cpp test/transcoding/builtin_vars_different_type.ll
$ git clang-format
clang-format did not modify any files
```

jcranmer-intel: That's... surprising, but if it's what clang-format is happy with...

jcranmer-intel (approved):

asudarsa:
Hi @sarnex 

I am wondering why the input LLVM IR would try to extract a 32-bit value from a 64-bit vector. Can you please point out the scenario in which such code is generated and is valid?

Thanks

lib/SPIRV/SPIRVUtil.cpp
```diff
@@ -2034,6 +2040,11 @@ static void replaceUsesOfBuiltinVar(Value *V, const APInt &AccumulatedOffset,
                     Builder.CreateCall(ReplacementFunc, {Builder.getInt32(I)})),
                 Builder.getInt32(I));
           }
+          // Insert a bitcast from the reconstructed vector to the load vector
+          // type in case they are different.
+          // %2 = insertelement <3 x i64> %0, i64 %1, i32 2
+          // bitcast <3 x i64> %2 to <6 x i32>
```
asudarsa, line 2046:
```suggestion
          // Input:
          // %2 = insertelement <3 x i64> %0, i64 %1, i32 2
          // %3 = extractelement <6 x i32> %2, i32 0
          // Modified:
          // %2 = insertelement <3 x i64> %0, i64 %1, i32 2
          // %3 = bitcast <3 x i64> %2 to <6 x i32>
          // %4 = extractelement <6 x i32> %3, i32 0
          
```

sarnex: The suggested comment doesn't exactly match what we are seeing, I took this idea and updated the comment.

sarnex:
Hi @asudarsa,
Yes, the input IR is valid. In this case it is the result of an optimization inside InstCombine. It's in the `combineLoadToOperationType` function where it changes the load type to be the use type in some cases.

asudarsa (approved):
Thanks for answers, LGTM

sarnex:
@MrSidims I think we can merge this bad boy, thanks

PR #2165, opened 2023-09-27
Reverse translation of access + load/store operations for cooperative matrix

vmaksimo:
Implement translation via SPIR-V friendly calls, as:
1. the LLVM instructions are not capable to accept target extension types;
2. cooperative matrix is an opaque object and accessing elements is implementation defined, hence we can't use GEP to which AccessChain naturally maps, since GEP has a different meaning.

As for now some BE would need to recognize and define what to do with a call to __spirv_AccessChain(matrix, index). Better option is to map such SPIR-V to an intrinsic or define an appropriate type in LLVM (hence defining rules for GEP and other instructions) , but it's off the table now.


vmaksimo:
Please take a look @MrSidims @asudarsa @LU-JOHN

MrSidims:
Same comment about justification, GEP not only can't work with target ext types, but also make a little sense in terms of accessing an element of a slice of a matrix owned by the current WI - which is purely implementation defined thing.

MrSidims (approved):

PR #2167, opened 2023-09-28
Add support for IAddCarry SPIRV instruction

bwlodarcz:
This commit implements bidirectional translation of the llvm.uadd.with.overflow and the IAddCarry intrinsic.
Intrinsic llvm.uadd.with.overflow returns struct which second element have a type of i1.
The llvm type i1 is, in llvm-spirv, directly translated to BoolType.
SPIRV specification requires that the composite which returns from IAddCarry needs to have both elements of the same type.
In result, current implementation is not compliant and should be considered temporary.


lib/SPIRV/libSPIRV/SPIRVOpCode.h
```diff
@@ -71,7 +71,7 @@ inline bool isAtomicOpCode(Op OpCode) {
 }
 inline bool isBinaryOpCode(Op OpCode) {
   return ((unsigned)OpCode >= OpIAdd && (unsigned)OpCode <= OpFMod) ||
```
asudarsa, line 73: I am wondering why this cannot be: (unsigned)OpCode >= OpIAdd && (unsigned)OpCode <= OpSMulExtended)
Ref: https://github.com/KhronosGroup/SPIRV-Headers/blob/79743b899fde5c954897b2694291002626358fac/include/spirv/unified1/spirv.h#L1377

test/llvm-intrinsics/uadd.with.overflow.ll
```diff
@@ -0,0 +1,43 @@
+; RUN: llvm-as %s -o %t.bc
+; RUN: llvm-spirv %t.bc -spirv-text -o - | FileCheck --check-prefix CHECK-SPIRV %s
+; RUN: llvm-spirv %t.bc -o %t.spv
+; RUN: spirv-val %t.spv
+; RUN: llvm-spirv -r %t.spv -o %t.rev.bc
+; RUN: llvm-dis %t.rev.bc -o - | FileCheck --check-prefix CHECK-LLVM %s
+
+target triple = "spir64-unknown-unknown"
+
+; CHECK-SPIRV: IAddCarry
```
asudarsa, line 10: May be we can add tests to make sure the SPIR-V types are correctly generated?
https://registry.khronos.org/SPIR-V/specs/unified1/SPIRV.html#OpIAddCarry

lib/SPIRV/SPIRVWriter.cpp
```diff
@@ -3969,6 +3970,11 @@ SPIRVValue *LLVMToSPIRVBase::transIntrinsicInst(IntrinsicInst *II,
     return BM->addExtInst(Ty, BM->getExtInstSetId(SPIRVEIS_OpenCL), ExtOp,
                           std::move(Operands), BB);
   }
+  case Intrinsic::uadd_with_overflow: {
```
asudarsa, line 3973: Normal transType will not work for result here.
For example: if we have LLVM result type {i16, i1}, we are going to get {TypeInt, TypeBool} for SPIR-V. Instead, we need {TypeInt, TypeInt} according to spec: https://registry.khronos.org/SPIR-V/specs/unified1/SPIRV.html#OpIAddCarry

The test is also failing in CI (spirv-val) because of this. Please take a look.

Thanks

lib/SPIRV/SPIRVReader.cpp
```diff
@@ -2492,6 +2492,14 @@ Value *SPIRVToLLVM::transValueWithoutDecoration(SPIRVValue *BV, Function *F,
   case OpSignBitSet:
     return mapValue(BV,
                     transRelational(static_cast<SPIRVInstruction *>(BV), BB));
+  case OpIAddCarry: {
+    IRBuilder Builder(BB);
+    auto *BC = static_cast<SPIRVBinary *>(BV);
+    return mapValue(BV, Builder.CreateBinaryIntrinsic(
```
asudarsa, line 2498: Please look at comment in SPIRVWriter.cpp. In the reverse translation, the second LLVM type will always be i1.

Thanks

lib/SPIRV/libSPIRV/SPIRVOpCode.h
bwlodarcz: The problem here is that I'm not sure what this function name actually means.
From context we can assume that binary in this case means two registers operator. If this is true than it means that this function doesn't cover all possible ops like e.g OpFOrdNotEqual.

test/llvm-intrinsics/uadd.with.overflow.ll
bwlodarcz: Done

lib/SPIRV/SPIRVWriter.cpp
bwlodarcz: Done

lib/SPIRV/SPIRVReader.cpp
bwlodarcz: Done

MrSidims (approved):
LGTM, will merge after @asudarsa approval

LU-JOHN (approved):
LGTM

lib/SPIRV/SPIRVReader.cpp
```diff
@@ -2492,6 +2492,14 @@ Value *SPIRVToLLVM::transValueWithoutDecoration(SPIRVValue *BV, Function *F,
...
+                            Intrinsic::uadd_with_overflow,
+                            transValue(BC->getOperand(0), F, BB),
+                            transValue(BC->getOperand(1), F, BB)));
```
asudarsa, line 2501: This type needs to be i1 in the resulting LLVM IR. 
Can you please check?

Thanks

bwlodarcz: This conversion already works by implicit conversion somewhere inside CreateBinaryIntrinsic. I'm not very fond of this kind of implicit behaviors and imo this should by default explode but it isn't. And now is a problem because without going inside CreateBinaryIntrinsic and finding out how conversion works there is no way to prove that forcing type of getOperand(1) works.
```
case OpIAddCarry: {
    IRBuilder Builder(BB);
    SPIRVTypeBool BOC;
    BC->getOperand(1)->setType(&BOC);
    return mapValue(BV, Builder.CreateBinaryIntrinsic(
                            Intrinsic::uadd_with_overflow,
                            transValue(BC->getOperand(0), F, BB),
                            transValue(BC->getOperand(1), F, BB)));
  }
```
This would be a possible way to force type change.

asudarsa (approved):
LGTM. Please track this non-conformance with Spec by opening a github issue.

Thanks

PR #2168, opened 2023-09-28
Add support for ISubBorrow SPIRV instruction

bwlodarcz:
This commit implements bidirectional translation of the llvm.usub.with.overflow and the ISubBorrow intrinsic.
Intrinsic llvm.usub.with.overflow returns struct which second element have a type of i1.
The llvm type i1 is, in llvm-spirv, directly translated to BoolType.
SPIRV specification requires that the composite which returns from ISubBorrow needs to have both elements of the same type.
In result, current implementation is not compliant and should be considered temporary.


MrSidims:
@bwlodarcz please comment spirv-val in the test

test/llvm-intrinsics/usub.with.overflow.ll
```diff
@@ -0,0 +1,43 @@
...
+; RUN: llvm-spirv -r %t.spv -o %t.rev.bc
+; RUN: llvm-dis %t.rev.bc -o - | FileCheck --check-prefix CHECK-LLVM %s
+
+target triple = "spir64-unknown-unknown"
+
+; CHECK-SPIRV: ISubBorrow
+; CHECK-LLVM: call { i16, i1 } @llvm.usub.with.overflow.i16
+; CHECK-LLVM: call { i32, i1 } @llvm.usub.with.overflow.i32
+; CHECK-LLVM: call { i64, i1 } @llvm.usub.with.overflow.i64
+; CHECK-LLVM: call { <4 x i32>, <4 x i1> } @llvm.usub.with.overflow.v4i32
```
MrSidims, line 14: would be nice to extend the test adding checks like in IAddCarry

MrSidims (approved):

PR #2169, opened 2023-10-02
[NFC] Remove unnecessary prefix check in LLVMToSPIRVBase::getFPBuiltinType

bwlodarcz:
Definition of StringRef.consume_front also performs prefix check. It makes additional check redundant.


asudarsa (approved):
LGTM. Thanks

Issue #2164, opened 2023-09-26
[llvm_release_170] Problem linking with LLVM static build


pal1000:

pal1000:
The problem seams to be `llvm-spirv` tool not being able to static link LLVM libraries.

PR #2175, opened 2023-10-10
[NO MERGE][ci/windows:/llvm_release_170: Tweak ci to reproduce #2164

pal1000:


pal1000:
It looks like #2164 is caused by LLVM switching CMP0091 to NEW while SPIRV-LLVM-Translator stayed on OLD. In tree builds get switched to new policy behavior automatically, but it can cause trouble with out of tree builds and explains why CI unexpectedly failed to reproduce.

PR #2190, opened 2023-10-19
Fix arguments asserts for InitModeINTEL

jsji:
sycl-cts is failing with
InvalidLlvmModule: Invalid LLVM module: InitModeINTEL requires exactly 1
extra operand


jsji:
The failure is because we haven't update the sycl with new decoration, the assert is correct here.

PR #2192, opened 2023-10-22
[Builtin] Fix issue with attribute list update

asudarsa:
When adding changes to support fpbuiltins in the translator, there was a bug that overwrote some attributes.
This PR aims to fix that issue.

Thanks

### DIFF
--- a/lib/SPIRV/SPIRVBuiltinHelper.cpp
+++ b/lib/SPIRV/SPIRVBuiltinHelper.cpp
@@ -103,7 +103,10 @@ Value *BuiltinCallMutator::doConversion() {
       Builder.Insert(addCallInst(CI->getModule(), FuncName, ReturnTy, Args,
                                  &Attrs, nullptr, Mangler.get()));
   NewCall->copyMetadata(*CI);
-  NewCall->setAttributes(CI->getAttributes());
+  if (CI->hasFnAttr("fpbuiltin-max-error")) {
+    auto Attr = CI->getFnAttr("fpbuiltin-max-error");
+    NewCall->addFnAttr(Attr);
+  }
   Value *Result = MutateRet ? MutateRet(Builder, NewCall) : NewCall;
   Result->takeName(CI);
   if (!CI->getType()->isVoidTy())


lib/SPIRV/SPIRVBuiltinHelper.cpp
```diff
@@ -103,7 +103,10 @@ Value *BuiltinCallMutator::doConversion() {
       Builder.Insert(addCallInst(CI->getModule(), FuncName, ReturnTy, Args,
                                  &Attrs, nullptr, Mangler.get()));
   NewCall->copyMetadata(*CI);
-  NewCall->setAttributes(CI->getAttributes());
+  if (CI->hasFnAttr("fpbuiltin-max-error")) {
+    auto Attr = CI->getFnAttr("fpbuiltin-max-error");
+    NewCall->addFnAttr(Attr);
```
vmaksimo, line 108: Should we also call `setAttributes()` to copy all the other attributes, not just `fpbuiltin-max-error`?

vmaksimo (commented):
Please also take a look at CI failure, the test or the fix may need modifications

lib/SPIRV/SPIRVBuiltinHelper.cpp
MrSidims: In addition to Viktoria's comment, like this we also discard calling convention if it was not a default one.

asudarsa: I am trying to NOT introduce any new logic here. This PR aims to only correct the bug introduced by PR #2056

LU-JOHN:
> When adding changes to support fpbuiltins in the translator, there was a bug that overwrote some attributes. This PR aims to fix that issue.

Does the test case have attributes that will be overwritten without this change?   It was not clear to me what part of the test case would hit this bug.

lib/SPIRV/SPIRVBuiltinHelper.cpp
asudarsa: I do not think copyMetadata overwrites existing metadata. So we should be fine here.

Thanks

vmaksimo (approved):

PR #235, opened 2019-06-24
Translate Intel FPGA merge, max_replicates and simple_dual_port memor…

vmaksimo:
…y attributes

http://htmlpreview.github.io/?https://github.com/KhronosGroup/SPIRV-Registry/blob/master/extensions/INTEL/SPV_INTEL_fpga_memory_attributes.html

PR #2196, opened 2023-10-31
Support for new SPV_INTEL_fpga_memory_attributes

bwlodarcz:
This commit adds StridesizeINTEL, WordsizeINTEL and TrueDualPortINTEL support to translator.


MrSidims:
restarting CI from the fresh HEAD

lib/SPIRV/SPIRVReader.cpp
```diff
@@ -3513,6 +3513,12 @@ void generateIntelFPGAAnnotation(
   }
   if (E->hasDecorate(DecorationForcePow2DepthINTEL, 0, &Result))
     Out << "{force_pow2_depth:" << Result << '}';
+  if (E->hasDecorate(DecorationStridesizeINTEL, 0, &Result))
+    Out << "{stride_size:" << Result << "}";
+  if (E->hasDecorate(DecorationWordsizeINTEL, 0, &Result))
+    Out << "{word_size:" << Result << "}";
+  if (E->hasDecorate(DecorationTrueDualPortINTEL))
+    Out << "{true_dual_port:1}";
```
artemrad, line 3521: it should be `{true_dual_port}` this annotation has no value associated with it

lib/SPIRV/SPIRVReader.cpp
```diff
@@ -3610,6 +3616,12 @@ void generateIntelFPGAAnnotationForStructMember(
   if (E->hasMemberDecorate(DecorationForcePow2DepthINTEL, 0, MemberNumber,
                            &Result))
     Out << "{force_pow2_depth:" << Result << '}';
+  if (E->hasMemberDecorate(DecorationStridesizeINTEL, 0, MemberNumber, &Result))
+    Out << "{stride_size:" << Result << "}";
+  if (E->hasMemberDecorate(DecorationWordsizeINTEL, 0, MemberNumber, &Result))
+    Out << "{word_size:" << Result << "}";
+  if (E->hasMemberDecorate(DecorationTrueDualPortINTEL, 0, MemberNumber))
+    Out << "{true_dual_port:1}";
```
artemrad, line 3624: ditto

lib/SPIRV/SPIRVReader.cpp
bwlodarcz: The value part of attribute is following convention from RegistryINTEL and SimpleDualPortINTEL which also doesn't have value. Value of 1 in this case means enabled.

MrSidims (approved):
LGTM

lib/SPIRV/SPIRVReader.cpp
MrSidims: @artemrad I agree with Bertrand, that the format that was set quite some time ago, you can see  [old translator PR](https://github.com/KhronosGroup/SPIRV-LLVM-Translator/pull/235) and [still existing test in clang](https://github.com/intel/llvm/blob/sycl/clang/test/CodeGenSYCL/intel-fpga-local.cpp#L21) for reference. Please reach us out if you would like to reconsider this format and refactor it.

artemrad: I disagree with the original request, since there is no scenario where `simple_dual_port:0` will get generated. This is redundant and confusing  translation of a SPIRV value-less decoration. The same logic applies to the new `true_dual_port`.

They should both have no value, to match with the SPIRV spec.

bwlodarcz: @artemrad Done

lib/SPIRV/SPIRVReader.cpp
bwlodarcz: Done.

PR #2198, opened 2023-11-06
Support for llvm.vector.reduce.* intrinsics

bwlodarcz:
A set of llvm.vector.reduce.* intrinsics doesn't have straight forward operation equivalent on the SPIRV side. The commit needed be split on three sections: Integer, Floating-point and Comparison reductions.
On every category at first we need to extract each value from vector.  
For Integer reductions we perform an operation on each pair of vector elements and repeat until there is only one value.
This way emitted code has the least number of dependencies between each operation and maximally utilizes instruction parallelism.
For FP reductions because of non associativity (due to rounding errors) we needed to emit sequential code in which every next operation is based on previous result. This solution is slow but correct. Potential way of improvement Floating-point is to use reassoc flag from -ffast-math.
For Comparison reductions we can utilize similar algorithm like in Integer section with the difference that now we need to emit Comp operation and OpSelect on which result is based like in cpp ternary operation. The rest is the same.


lib/SPIRV/SPIRVWriter.cpp
```diff
@@ -4023,6 +4024,90 @@ SPIRVValue *LLVMToSPIRVBase::transIntrinsicInst(IntrinsicInst *II,
                              transValue(II->getArgOperand(0), BB),
                              transValue(II->getArgOperand(1), BB), BB);
   }
+  case Intrinsic::vector_reduce_add:
+  case Intrinsic::vector_reduce_mul:
+  case Intrinsic::vector_reduce_and:
+  case Intrinsic::vector_reduce_or:
+  case Intrinsic::vector_reduce_xor: {
+    Op Op;
+    if (IID == Intrinsic::vector_reduce_add) {
```
asudarsa, line 4033: nit: switch statement may be better?

lib/SPIRV/SPIRVWriter.cpp
```diff
@@ -4023,6 +4024,90 @@ SPIRVValue *LLVMToSPIRVBase::transIntrinsicInst(IntrinsicInst *II,
...
+      Op = OpBitwiseAnd;
+    } else if (IID == Intrinsic::vector_reduce_or) {
+      Op = OpBitwiseOr;
+    } else {
+      Op = OpBitwiseXor;
+    }
+    VectorType *VT = cast<VectorType>(II->getArgOperand(0)->getType());
+    SPIRVValue *SV = transValue(II->getArgOperand(0), BB);
+    SPIRVTypeInt *ResultType =
+        BM->addIntegerType(VT->getElementType()->getIntegerBitWidth());
```
asudarsa, line 4047: Nit: 
```suggestion
        transType(VT->getElementType());
```

lib/SPIRV/SPIRVWriter.cpp
```diff
@@ -4023,6 +4024,90 @@ SPIRVValue *LLVMToSPIRVBase::transIntrinsicInst(IntrinsicInst *II,
                              transValue(II->getArgOperand(0), BB),
                              transValue(II->getArgOperand(1), BB), BB);
   }
+  case Intrinsic::vector_reduce_add:
```
asudarsa, line 4027: Please add some details in a comment about the implementation here.

Thanks

test/llvm-intrinsics/llvm-vector-reduce/add.ll
```diff
@@ -0,0 +1,552 @@
+; RUN: llvm-as %s -o %t.bc
+; RUN: llvm-spirv %t.bc -spirv-text -o - | FileCheck --check-prefix CHECK-SPIRV %s
+; RUN: llvm-spirv %t.bc -o %t.spv
+; RUN: spirv-val %t.spv
+
+target triple = "spir64-unknown-unknown"
+
+; CHECK-SPIRV-DAG: TypeInt [[#I8TYPE:]] 8
```
asudarsa, line 8: I do appreciate the thoroughness of this test. But I am not sure if it is really necessary to check for all data types here. If the implementation passes/fails for a particular data type, is there a reason why it will have a different result for some other data type?

test/llvm-intrinsics/llvm-vector-reduce/add.ll
```diff
@@ -0,0 +1,552 @@
+; RUN: llvm-as %s -o %t.bc
```
asudarsa, line 1: Given the amount of effort that has already gone into writing this test, I will categorize this comment as 'feel free to ignore' and 'non-blocking'. Do we want to test the reverse translated LLVM code here?

Thanks

asudarsa (dismissed):
A few nits. Overall change looks good. If possible, it will be a good idea to verify the functionality using E2E tests though it is not required for this PR to be approved.
Also, for future reference, I see two places where the complexity of the testing can be reduced: (1) I think it's sufficient to test for a single data type. If the implementation failed in one data type, i suspect it's going to fail for all (2) I also think testing for add/and/mul/or/xor can be represented using a single test as the implementation is same for all these ops. Similarly for smax/smin/umax/umin.

Good job on writing the tests. 

Thanks

lib/SPIRV/SPIRVWriter.cpp
bwlodarcz: Hmm... for me it looks more hacky and slower than just adding a IntegerType. BM->addIntegerType explictly is telling what's is doing - adding a necessary type (imo. should be something like getOrCreateIntegerType).
transType here is going around  by passing through whole translation logic to achieve the same goal without explicitly telling intention.
In addition to that transType is returning SPIRVType * but addIntegerType returns SPIRVIntegerType * which have more information than it's parent type. 
We can of course remove addXXXType api and establish that's transType it the way - it would be good to have discussion with community about that.

lib/SPIRV/SPIRVWriter.cpp
bwlodarcz: If we aren't using fallthrough (which will be just jumps) in switch clause than it's just an equivalent of if clauses so the resulting code will be (should be) the same - branches or cmovs.

test/llvm-intrinsics/llvm-vector-reduce/add.ll
bwlodarcz: IMO there is. The major necessity here comes from the fact that translator during translation is a stateful machine. This actually requires from the test to contain at least some kind of combination of types to be sure that at least in some ways this stateful behavior is covered by tests. The second reason is to be 100% sure that the types aren't changing of the implementation behavior. And third (imo. weakest) is that the api calls which are used by implementation are also tested for correctness in that context.

test/llvm-intrinsics/llvm-vector-reduce/add.ll
bwlodarcz: IMO. no. These are covered by appropriate instruction tests written in the past - such effort will be redundant when such tests are in place.

lib/SPIRV/SPIRVWriter.cpp
```diff
@@ -91,6 +91,7 @@
 #include "llvm/Transforms/Utils/LoopSimplify.h"
 #include "llvm/Transforms/Utils/Mem2Reg.h"
 
+#include <cstdio>
```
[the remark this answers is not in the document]
bwlodarcz: True. Sometimes my editor put these. Will be removed in final commit after cleanup.

lib/SPIRV/SPIRVWriter.cpp
asudarsa: Not a blocker. So marking this resolved.

test/llvm-intrinsics/llvm-vector-reduce/add.ll
asudarsa: I am not sure about correctness as we are not actually running anything and creating results in the LIT tests. My argument is this. LIT tests that are being added here test specifically the sequence of SPIR-V instructions generated by 'your' implementation and if your implementation breaks one test that checks SPIR-V code generation for one type, it is going to break for all types. 

Also, I am ok to have this discussion offline and not block this PR.

Thanks

test/llvm-intrinsics/llvm-vector-reduce/add.ll
asudarsa: Hmm. I do not think we have instruction level tests anywhere. Please feel free to point me to such tests.

Thanks

MrSidims (commented):
Integer add/mul, or/and/xor and float add/mul are LGTM. Tomorrow will take a look at min/max

lib/SPIRV/SPIRVWriter.cpp
```diff
@@ -4023,6 +4023,122 @@ SPIRVValue *LLVMToSPIRVBase::transIntrinsicInst(IntrinsicInst *II,
...
+        Extracts[Idx] = BM->addBinaryInst(Op, ResultSType, Extracts[Idx << 1],
+                                          Extracts[(Idx << 1) + 1], BB);
+      }
+      Counter >>= 1;
+    }
+    if ((VecSize & 1) != 0) {
+      Extracts[0] = BM->addBinaryInst(Op, ResultSType, Extracts[0],
+                                      Extracts[VecSize - 1], BB);
+    }
+    return Extracts[0];
```
MrSidims, line 4066: Here and later for max/min I don't oppose such optimization, but kind of have a feeling, that BE can this permutation as well. Anyway, may I ask you to add a short comment describing what's happening here and why?

MrSidims (approved):

PR #2209, opened 2023-11-08
[NFC] Remove Type::getInt8PtrTy after 7b9d73c2f90

MrSidims:
Follow-up to the opaque pointer transition llvm/llvm-project@7b9d73c2f90


vmaksimo (commented):
Can we use `getUnqual` as in llvm-project change?
Please also apply clang-format.

MrSidims:
> Can we use getUnqual as in llvm-project change?

Nope, `getUnqual` creates a pointer in default address space.

> Please also apply clang-format.

Sure :) I was just having some unrelated test failures in the local branch, which I would like to address first, but don't see them anymore.

vmaksimo (approved):

PR #2210, opened 2023-11-08
add initial f16 type support for atomicrmw in llvm-spirv translator

VyacheslavLevytskyy:
This PR aims to add f16 type support for atomicrmw in llvm-spirv translator, with the reference to the extension documented in [1].
There are two concerns related to the subject:
1. `SPIRVAtomicFAddEXTInst::getRequiredExtension()` should return a list of required extension to support the requirement to list both `SPV_EXT_shader_atomic_float16_add` and `SPV_EXT_shader_atomic_float_add` extensions in the module (see "Extension Name" section of the ref [1]). However, the return type is `std::optional<ExtensionID>` and returning a vector would need a bigger rework.
2. Including `SPV_EXT_shader_atomic_float16_add` into `--spirv-ext` argument of `llvm-spirv` doesn't result in producing the correspondent capability (`AtomicFloat16AddEXT`) and extension in a SPIRV output.
```
$ llvm-spirv AtomicFAddEXT.ll.tmp.bc --spirv-ext=+SPV_EXT_shader_atomic_float_add,+SPV_EXT_shader_atomic_float16_add -o AtomicFAddEXT.ll.tmp.spv
$ llvm-spirv -to-text AtomicFAddEXT.ll.tmp.spv -o /dev/stdout
...
2 Capability AtomicFloat32AddEXT
2 Capability AtomicFloat64AddEXT
9 Extension "SPV_EXT_shader_atomic_float_add"
...
```
This prevents extending the test case of AtomicFAddEXT.ll in EXT/SPV_EXT_shader_atomic_float.

References:
[1] https://github.com/KhronosGroup/SPIRV-Registry/blob/main/extensions/EXT/SPV_EXT_shader_atomic_float16_add.asciidoc

### DIFF
--- a/test/extensions/EXT/SPV_EXT_shader_atomic_float_/atomicrmw_fadd_half.ll
+++ b/test/extensions/EXT/SPV_EXT_shader_atomic_float_/atomicrmw_fadd_half.ll
@@ -1,0 +1,33 @@
+; RUN: llvm-as < %s -o %t.bc
+; RUN: llvm-spirv --spirv-ext=+SPV_EXT_shader_atomic_float16_add %t.bc -o %t.spv
+; RUN: spirv-val %t.spv
+; RUN: llvm-spirv -to-text %t.spv -o - | FileCheck %s
+
+; CHECK-DAG: Extension "SPV_EXT_shader_atomic_float16_add"
+; CHECK-DAG: Capability AtomicFloat16AddEXT
+; CHECK: TypeInt [[Int:[0-9]+]] 32 0
+; CHECK-DAG: Constant [[Int]] [[Scope_Device:[0-9]+]] 1 {{$}}
+; CHECK-DAG: Constant [[Int]] [[MemSem_SequentiallyConsistent:[0-9]+]] 16
+; CHECK: TypeFloat [[Half:[0-9]+]] 16
+; CHECK: Variable {{[0-9]+}} [[HalfPointer:[0-9]+]]
+; CHECK: Constant [[Half]] [[HalfValue:[0-9]+]] 20800
+
+target datalayout = "e-i64:64-v16:16-v24:32-v32:32-v48:64-v96:128-v192:256-v256:256-v512:512-v1024:1024"
+target triple = "spir64"
+
+@f = common dso_local local_unnamed_addr addrspace(1) global half 0.000000e+00, align 4
+
+; Function Attrs: nounwind
+define dso_local spir_func void @test_atomicrmw_fadd() local_unnamed_addr #0 {
+entry:
+ %0 = atomicrmw fadd ptr addrspace(1) @f, half 42.000000e+00 seq_cst
+; CHECK: AtomicFAddEXT [[Half]] {{[0-9]+}} [[HalfPointer]] [[Scope_Device]] [[MemSem_SequentiallyConsistent]] [[HalfValue]]
+
+  ret void
+}
+
+attributes #0 = { nounwind "correctly-rounded-divide-sqrt-fp-math"="false" "disable-tail-calls"="false" "frame-pointer"="all" "less-precise-fpmad"="false" "min-legal-vector-width"="0" "no-infs-fp-math"="false" "no-jump-tables"="false" "no-nans-fp-math"="false" "no-signed-zeros-fp-math"="false" "no-trapping-math"="false" "stack-protector-buffer-size"="8" "unsafe-fp-math"="false" "use-soft-float"="false" }
+
+!llvm.module.flags = !{!0}
+
+!0 = !{i32 1, !"wchar_size", i32 4}
--- a/test/extensions/EXT/SPV_EXT_shader_atomic_float_/atomicrmw_fsub_half.ll
+++ b/test/extensions/EXT/SPV_EXT_shader_atomic_float_/atomicrmw_fsub_half.ll
@@ -1,0 +1,38 @@
+; RUN: llvm-as < %s -o %t.bc
+; RUN: llvm-spirv --spirv-ext=+SPV_EXT_shader_atomic_float16_add %t.bc -o %t.spv
+; RUN: spirv-val %t.spv
+; RUN: llvm-spirv -to-text %t.spv -o - | FileCheck --check-prefix=CHECK-SPIRV %s
+
+; RUN: llvm-spirv -r %t.spv -o %t.rev.bc
+; RUN: llvm-dis %t.rev.bc -o - | FileCheck %s --check-prefixes=CHECK-LLVM
+
+; CHECK-SPIRV-DAG: Extension "SPV_EXT_shader_atomic_float16_add"
+; CHECK-SPIRV-DAG: Capability AtomicFloat16AddEXT
+; CHECK-SPIRV: TypeInt [[Int:[0-9]+]] 32 0
+; CHECK-SPIRV-DAG: Constant [[Int]] [[Scope_Device:[0-9]+]] 1 {{$}}
+; CHECK-SPIRV-DAG: Constant [[Int]] [[MemSem_SequentiallyConsistent:[0-9]+]] 16
+; CHECK-SPIRV: TypeFloat [[Half:[0-9]+]] 16
+; CHECK-SPIRV: Variable {{[0-9]+}} [[HalfPointer:[0-9]+]]
+; CHECK-SPIRV: Constant [[Half]] [[HalfValue:[0-9]+]] 15360
+
+target datalayout = "e-i64:64-v16:16-v24:32-v32:32-v48:64-v96:128-v192:256-v256:256-v512:512-v1024:1024"
+target triple = "spir64"
+
+@f = common dso_local local_unnamed_addr addrspace(1) global half 0.000000e+00, align 4
+
+; Function Attrs: nounwind
+define dso_local spir_func void @test_atomicrmw_fadd() local_unnamed_addr #0 {
+entry:
+ %0 = atomicrmw fsub ptr addrspace(1) @f, half 1.0e+00 seq_cst
+; CHECK-SPIRV: FNegate [[Half]] [[NegateValue:[0-9]+]] [[HalfValue]]
+; CHECK-SPIRV: AtomicFAddEXT [[Half]] {{[0-9]+}} [[HalfPointer]] [[Scope_Device]] [[MemSem_SequentiallyConsistent]] [[NegateValue]]
+; CHECK-LLVM: [[FNegateLLVM:%[0-9]+]] = fneg half 0xH3C00
+; CHECK-LLVM: call spir_func half {{.*}}atomic_add{{.*}}(ptr addrspace(1) @f, half [[FNegateLLVM]])
+  ret void
+}
+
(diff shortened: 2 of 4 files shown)


MrSidims:
in-tree builds failures aren't related with the PR

test/extensions/EXT/SPV_EXT_shader_atomic_float_/atomicrmw_fadd_half.ll
```diff
@@ -0,0 +1,33 @@
+; RUN: llvm-as < %s -o %t.bc
+; RUN: llvm-spirv --spirv-ext=+SPV_EXT_shader_atomic_float16_add %t.bc -o %t.spv
+; RUN: spirv-val %t.spv
+; RUN: llvm-spirv -to-text %t.spv -o - | FileCheck %s
+
+; CHECK-DAG: Extension "SPV_EXT_shader_atomic_float16_add"
+; CHECK-DAG: Capability AtomicFloat16AddEXT
+; CHECK: TypeInt [[Int:[0-9]+]] 32 0
```
MrSidims, line 8: Just 1 nit about the tests, I'm not expecting it to be fixed, as they are following style of the tests added before for atomic_float extensions.
```suggestion
; CHECK: TypeInt [[#Int:]] 32 0
```
slightly better way of checking integer variables in IR within LIT infrastructure. Just integer values can be checked by `[[#]]` instead of `{{[0-9]+}}`

MrSidims (approved):
For the most part LGTM, just few nits.

> Including SPV_EXT_shader_atomic_float16_add into --spirv-ext argument of llvm-spirv doesn't result in producing the correspondent capability (AtomicFloat16AddEXT) and extension in a SPIRV output.

Not 100% sure what do you mean, in your test I see `; CHECK-DAG: Capability AtomicFloat16AddEXT` which apparently is not failing. May ask you to clarify this?

Also we add test line for reverse translation to check the result LLVM IR, but in this case this is already tested in test/extensions/EXT/SPV_EXT_shader_atomic_float_/AtomicFAddEXT.ll , so I don't mind leaving it unchecked (yet again it follows existing rules).

MrSidims:
I probably got it

> AtomicFAddEXT.ll.tmp.spv $ llvm-spirv -to-text AtomicFAddEXT.ll.tmp.spv -o /dev/stdout

AtomicFAddEXT.ll doesn't have atomic instructions operating over half type, hence even if the extension is enabled - it remains to be unused hence fp16 capability shouldn't be added.

vmaksimo (commented):
LGTM! Only one question to agree on with @MrSidims:

> 1. SPIRVAtomicFAddEXTInst::getRequiredExtension() should return a list of required extension to support the requirement to list both SPV_EXT_shader_atomic_float16_add and SPV_EXT_shader_atomic_float_add extensions in the module 

will it be OK to implicitly enable `SPV_EXT_shader_atomic_float_add ` if we are enabling `SPV_EXT_shader_atomic_float16_add`?

VyacheslavLevytskyy:
> LGTM! Only one question to agree on with @MrSidims:
> 
> > 1. SPIRVAtomicFAddEXTInst::getRequiredExtension() should return a list of required extension to support the requirement to list both SPV_EXT_shader_atomic_float16_add and SPV_EXT_shader_atomic_float_add extensions in the module
> 
> will it be OK to implicitly enable `SPV_EXT_shader_atomic_float_add ` if we are enabling `SPV_EXT_shader_atomic_float16_add`?

A new commit is to implement this approach as a temporarily workaround, until a bigger refactoring resolves it in a better way.

lib/SPIRV/libSPIRV/SPIRVModule.cpp
```diff
@@ -655,6 +655,15 @@ void SPIRVModuleImpl::addExtension(ExtensionID Ext) {
     return;
   }
   SPIRVExt.insert(ExtName);
+
+  // The following lines is a temporary workaround to satisfy the specification
+  // until prospective refactoring will make getRequiredExtension() return
+  // a vector of capabilities instead of a single capability
```
asudarsa, line 661:
```suggestion
  // a vector of extensionIDs instead of a single extensionID
```

lib/SPIRV/libSPIRV/SPIRVModule.cpp
```diff
@@ -655,6 +655,15 @@ void SPIRVModuleImpl::addExtension(ExtensionID Ext) {
...
+  if (Ext == ExtensionID::SPV_EXT_shader_atomic_float16_add) {
+    // TranslationOpts doesn't know about the workaround, don't ask TranslationOpts is it allowed
+    SPIRVMap<ExtensionID, std::string>::find(ExtensionID::SPV_EXT_shader_atomic_float_add, &ExtName);
+    SPIRVExt.insert(ExtName);
+  }
```
asudarsa, line 666: I think this is a corner case where a particular capability requires two extensions and this workaround is good enough. I do not think we need to modify getExtensionID unless there are other use cases. I am ok to be vetoed out here.

Thanks

lib/SPIRV/libSPIRV/SPIRVModule.cpp
```diff
@@ -655,6 +655,15 @@ void SPIRVModuleImpl::addExtension(ExtensionID Ext) {
     return;
   }
   SPIRVExt.insert(ExtName);
+
+  // The following lines is a temporary workaround to satisfy the specification
+  // until prospective refactoring will make getRequiredExtension() return
+  // a vector of capabilities instead of a single capability
+  if (Ext == ExtensionID::SPV_EXT_shader_atomic_float16_add) {
+    // TranslationOpts doesn't know about the workaround, don't ask TranslationOpts is it allowed
```
asudarsa, line 663: I did not understand this comment. Sorry.

test/extensions/EXT/SPV_EXT_shader_atomic_float_/atomicrmw_fadd_half.ll
```diff
@@ -0,0 +1,34 @@
+; RUN: llvm-as < %s -o %t.bc
+; RUN: llvm-spirv --spirv-ext=+SPV_EXT_shader_atomic_float16_add %t.bc -o %t.spv
+; RUN: spirv-val %t.spv
+; RUN: llvm-spirv -to-text %t.spv -o - | FileCheck %s
+
+; CHECK-DAG: Extension "SPV_EXT_shader_atomic_float16_add"
+; CHECK-DAG: Extension "SPV_EXT_shader_atomic_float_add"
+; CHECK-DAG: Capability AtomicFloat16AddEXT
+; CHECK: TypeInt [[Int:[0-9]+]] 32 0
```
asudarsa, line 9: Nit: Think this is a better name:
```suggestion
; CHECK: TypeInt [[TypeIntID:[0-9]+]] 32 0
```

test/extensions/EXT/SPV_EXT_shader_atomic_float_/atomicrmw_fadd_half.ll
```diff
@@ -0,0 +1,34 @@
...
+; CHECK-DAG: Constant [[Int]] [[Scope_Device:[0-9]+]] 1 {{$}}
```
asudarsa, line 10: I think we can avoid underscores in variable names

test/extensions/EXT/SPV_EXT_shader_atomic_float_/atomicrmw_fadd_half.ll
```diff
@@ -0,0 +1,34 @@
...
+; CHECK-DAG: Constant [[Int]] [[MemSem_SequentiallyConsistent:[0-9]+]] 16
+; CHECK: TypeFloat [[Half:[0-9]+]] 16
```
asudarsa, line 12:
```suggestion
; CHECK: TypeFloat [[TypeFloatHalfID:[0-9]+]] 16
```

test/extensions/EXT/SPV_EXT_shader_atomic_float_/atomicrmw_fadd_half.ll
```diff
@@ -0,0 +1,34 @@
+; RUN: llvm-as < %s -o %t.bc
+; RUN: llvm-spirv --spirv-ext=+SPV_EXT_shader_atomic_float16_add %t.bc -o %t.spv
+; RUN: spirv-val %t.spv
+; RUN: llvm-spirv -to-text %t.spv -o - | FileCheck %s
```
asudarsa, line 4: Why are we not checking reverse-generated LLVM code here?

test/extensions/EXT/SPV_EXT_shader_atomic_float_/atomicrmw_fsub_half.ll
```diff
@@ -0,0 +1,39 @@
...
+@f = common dso_local local_unnamed_addr addrspace(1) global half 0.000000e+00, align 4
+
+; Function Attrs: nounwind
+define dso_local spir_func void @test_atomicrmw_fadd() local_unnamed_addr #0 {
+entry:
+ %0 = atomicrmw fsub ptr addrspace(1) @f, half 1.0e+00 seq_cst
+; CHECK-SPIRV: FNegate [[Half]] [[NegateValue:[0-9]+]] [[HalfValue]]
+; CHECK-SPIRV: AtomicFAddEXT [[Half]] {{[0-9]+}} [[HalfPointer]] [[Scope_Device]] [[MemSem_SequentiallyConsistent]] [[NegateValue]]
+; CHECK-LLVM: [[FNegateLLVM:%[0-9]+]] = fneg half 0xH3C00
+; CHECK-LLVM: call spir_func half {{.*}}atomic_add{{.*}}(ptr addrspace(1) @f, half [[FNegateLLVM]])
```
asudarsa, line 31: I think we need to check for declaration/definition of this function as well?

Thanks

asudarsa (changes requested):
I have some comments that need addressing. I also had a higher level question. For LLVM atomic sub operation, we are replacing it with a combination of non-atomic negate operation followed by atomic add operation. Is this a valid transformation?

VyacheslavLevytskyy:
As for questions about two added test, they are following style and approach of existing tests for float and double. So the question is if it makes sense updating all together to polish those imperfections, or update just new half float tests, or leave as is. Please suggest.

MrSidims:
I would suggest to leave as is - it won't harm. If we want to update the style, it should go only to the existing test to not confuse git blame users if any appear.

test/extensions/EXT/SPV_EXT_shader_atomic_float_/atomicrmw_fadd_half.ll
VyacheslavLevytskyy: As @MrSidims commented earlier, reverse translation to check the result LLVM IR is already tested in test/extensions/EXT/SPV_EXT_shader_atomic_float_/AtomicFAddEXT.ll

MrSidims:
> I have some comments that need addressing. I also had a higher level question. For LLVM atomic sub operation, we are replacing it with a combination of non-atomic negate operation followed by atomic add operation. Is this a valid transformation?

Clarification, we only to if for floats. I don't see why it shouldn't work tbh, and as comparison of floats is a tricky thing to do - at least on X86 all float atomic operations falls to atomic compare exchange. For integer atomic operations using

MrSidims:
Merging as no objects are raised

PR #2214, opened 2023-11-14
Add OpCooperativeMatrixApplyFunctionINTEL instruction

VyacheslavLevytskyy:
This PR aims to introduce entities related to OpCooperativeMatrixApplyFunctionINTEL in llvm-spirv translator, according to https://github.com/intel/llvm/blob/sycl/sycl/doc/design/spirv-extensions/SPV_INTEL_joint_matrix.asciidoc.

TODO update the spec

### DIFF
--- a/lib/SPIRV/libSPIRV/SPIRVInstruction.h
+++ b/lib/SPIRV/libSPIRV/SPIRVInstruction.h
@@ -3420,6 +3420,26 @@ protected:
 _SPIRV_OP(CooperativeMatrixPrefetch, false, 8, true, 5)
 #undef _SPIRV_OP
 
+class SPIRVCooperativeMatrixInvocationInstructionsINTELInstBase
+    : public SPIRVInstTemplateBase {
+protected:
+  std::optional<ExtensionID> getRequiredExtension() const override {
+    return ExtensionID::SPV_INTEL_joint_matrix;
+  }
+  SPIRVCapVec getRequiredCapability() const override {
+    return getVec(
+        internal::CapabilityCooperativeMatrixInvocationInstructionsINTEL);
+  }
+};
+
+#define _SPIRV_OP(x, ...)                                                      \
+  typedef SPIRVInstTemplate<                                                   \
+      SPIRVCooperativeMatrixInvocationInstructionsINTELInstBase,               \
+      internal::Op##x##INTEL, __VA_ARGS__>                                     \
+      SPIRV##x##INTEL;
+_SPIRV_OP(CooperativeMatrixApplyFunction, true, 4, true)
+#undef _SPIRV_OP
+
 class SPIRVCooperativeMatrixKHRInstBase : public SPIRVInstTemplateBase {
 protected:
   std::optional<ExtensionID> getRequiredExtension() const override {
--- a/lib/SPIRV/libSPIRV/SPIRVInstruction.h
+++ b/lib/SPIRV/libSPIRV/SPIRVInstruction.h
@@ -3437,7 +3437,7 @@ protected:
       SPIRVCooperativeMatrixInvocationInstructionsINTELInstBase,               \
       internal::Op##x##INTEL, __VA_ARGS__>                                     \
       SPIRV##x##INTEL;
-_SPIRV_OP(CooperativeMatrixApplyFunction, true, 4, true)
+_SPIRV_OP(CooperativeMatrixApplyFunction, true, 5)
 #undef _SPIRV_OP
 
 class SPIRVCooperativeMatrixKHRInstBase : public SPIRVInstTemplateBase {
--- a/lib/SPIRV/libSPIRV/spirv_internal.hpp
+++ b/lib/SPIRV/libSPIRV/spirv_internal.hpp
@@ -77,8 +77,8 @@ enum InternalOp {
   IOpMaskedGatherINTEL = 6428,
   IOpMaskedScatterINTEL = 6429,
   IOpJointMatrixGetElementCoordINTEL = 6440,
-  IOpCooperativeMatrixPrefetchINTEL = 6449,
   IOpCooperativeMatrixApplyFunctionINTEL = 6448,
+  IOpCooperativeMatrixPrefetchINTEL = 6449,
   IOpPrev = OpMax - 2,
   IOpForward
 };
--- a/lib/SPIRV/libSPIRV/spirv_internal.hpp
+++ b/lib/SPIRV/libSPIRV/spirv_internal.hpp
@@ -78,6 +78,7 @@ enum InternalOp {
   IOpMaskedScatterINTEL = 6429,
   IOpJointMatrixGetElementCoordINTEL = 6440,
   IOpCooperativeMatrixPrefetchINTEL = 6449,
+  IOpCooperativeMatrixApplyFunctionINTEL = 6448,
   IOpPrev = OpMax - 2,
   IOpForward
 };
@@ -104,6 +105,7 @@ enum InternalCapability {
   ICapabilityTensorFloat32RoundingINTEL = 6425,
   ICapabilityMaskedGatherScatterINTEL = 6427,
   ICapabilityJointMatrixWIInstructionsINTEL = 6435,
+  ICapabilityCooperativeMatrixInvocationInstructionsINTEL = 6435,
   ICapabilityJointMatrixTF32ComponentTypeINTEL = 6436,
   ICapabilityJointMatrixBF16ComponentTypeINTEL = 6437,
   ICapabilityJointMatrixPackedInt2ComponentTypeINTEL = 6438,
@@ -177,6 +179,9 @@ _SPIRV_OP(Op, JointMatrixGetElementCoordINTEL)
 
 _SPIRV_OP(Capability, CooperativeMatrixPrefetchINTEL)
 _SPIRV_OP(Op, CooperativeMatrixPrefetchINTEL)
+
+_SPIRV_OP(Capability, CooperativeMatrixInvocationInstructionsINTEL)
+_SPIRV_OP(Op, CooperativeMatrixApplyFunctionINTEL)
 
 _SPIRV_OP(Capability, HWThreadQueryINTEL)
 _SPIRV_OP(BuiltIn, SubDeviceIDINTEL)
--- a/test/extensions/INTEL/SPV_INTEL_joint_matrix/cooperative_matrix_apply.ll
+++ b/test/extensions/INTEL/SPV_INTEL_joint_matrix/cooperative_matrix_apply.ll
@@ -1,0 +1,163 @@
+;; compiled from joint_matrix_apply_bf16.cpp from intel/llvm with some modifications
+
+; RUN: llvm-as < %s -o %t.bc
+; RUN: llvm-spirv %t.bc --spirv-ext=+SPV_KHR_cooperative_matrix,+SPV_INTEL_joint_matrix -o %t.spv
+; RUN: llvm-spirv %t.spv -to-text -o %t.spt
+; RUN: FileCheck < %t.spt %s --check-prefix=CHECK-SPIRV
+
+; RUN: llvm-spirv -r %t.spv -o %t.rev.bc
+; RUN: llvm-dis < %t.rev.bc | FileCheck %s --check-prefix=CHECK-LLVM
+
+; CHECK-SPIRV-DAG: Capability CooperativeMatrixKHR
+; CHECK-SPIRV-DAG: Capability CooperativeMatrixInvocationInstructionsINTEL
+; CHECK-SPIRV-DAG: Extension "SPV_INTEL_joint_matrix"
+; CHECK-SPIRV-DAG: Extension "SPV_KHR_cooperative_matrix"
+; CHECK-SPIRV-DAG: TypeCooperativeMatrixKHR [[#MatTy:]]
+; CHECK-SPIRV: CompositeConstruct [[#MatTy]] [[#Mat:]]
+; CHECK-SPIRV: PtrCastToGeneric [[#]] [[#Ptr:]] [[#]]
+; CHECK-SPIRV: CooperativeMatrixApplyFunctionINTEL [[#MatTy]] [[#Apply:]] [[#Ptr]] [[#Mat]]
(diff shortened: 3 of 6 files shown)


lib/SPIRV/libSPIRV/SPIRVInstruction.h
```diff
@@ -3396,6 +3396,26 @@ class SPIRVJointMatrixINTELWorkItemInst : public SPIRVJointMatrixINTELInstBase {
 _SPIRV_OP(JointMatrixGetElementCoord, true, 5)
 #undef _SPIRV_OP
 
+class SPIRVCooperativeMatrixInvocationInstructionsINTELInstBase
```
MrSidims, line 3399: Agree to have this class separated from 'WI matrix instructions' class

test/extensions/INTEL/SPV_INTEL_joint_matrix/cooperative_matrix_apply.ll
bwlodarcz, line 1: Wouldn't it be good to minimize test case to actually containing only necessary's lines for a test?

MrSidims: In general I agree with this, and that is the way to add 90%+ of the tests. Yet ideally for language features we should have tests be almost end-to-end'ish (or Integration'ish), see test/transcoding/*.cl tests, unfortunately sycl is not yet fully unstreamed to llvm.org. So the closest thing until then is to generate IR from some sycl test and use it as an input to the translator.

Why I'm thinking, that language features should be tested like this? Practice showed up, that for example for matrices having just calls to __spirv_JointMatrixLoadINTEL(A), (B), (C) followed by __spirv_JointMatrixMadINTEL(A, B, C) tests almost nothing as the compiler will generate extra code during optimizations that will affect translation (prime example mem2reg pass inserting phi instructions selecting matrices). Counter argument to this is that such cases should be tested in the appropriate repository, where language APIs are being added, but it won't work for features under active development, as tests in those repositories won't be enabled until all components are ready.

MrSidims: And for this very example - I'm not trusting myself to write LLVM IR that have a function, accepting a function object in a reference wrapper.

MrSidims: And btw, the test is quite minimized, all metadata, attributes and other not needed stuff is removed

bwlodarcz: >In general I agree with this, and that is the way to add 90%+ of the tests. Yet ideally for language features we should have tests be almost end-to-end'ish (or Integration'ish), see test/transcoding/*.cl tests, unfortunately sycl is not yet fully unstreamed to llvm.org. So the closest thing until then is to generate IR from some sycl test and use it as an input to the translator.

And it's ok to use frontend to get IR. But the next steps from that point is to strip everything which is unneeded and blurs the essence of the test. 
Minimization here have a simple goal - to separate things which are necessary to provide a functionality and nothing more. So no `@llvm.assume` or `@llvm.lifetime.*`, `!0`, `func attr` etc. - clear minimal IR needed for feature. Then, when you want to extend a feature or correct a bug one year from that point  - on first sight you know what feature looked like when it was implemented and what's needed to fix.
The problem with using straight forward outputs from frontend is that after some time the implementation in frontend will change but the IR version of the test here assumes that it's unchanged - frozen in time. Recompilation of a test issue from frontend will be impossible (because file doesn't exist or is renamed) or it will yield different IR. Is it test still useful and correct or it's not? Clear to read? This is what happened with FPGA_memory_attributes tests - which were implemented few years ago.

maksimsab: We could run `opt -S -passes=metarenamer < reduced.ll > reduced2.ll` on it in order to get rid off of burdened names like `%arrayidx.i29.i.i.i.i`.

MrSidims: It will change, but it doesn't mean, that the current output should be untested. And at this point of time for this very feature the only place, where it can be tested is llvm-spirv.

bwlodarcz: I just think that the feature test should be focused only on thing which is implemented feature - nothing more and nothing less. And implemented code should be tested thoroughly. IMO. If the whole sequence of features need to be tested and for singular frontend then it's some kind of integration test and should be treated differently than typical feature test.

lib/SPIRV/libSPIRV/SPIRVInstruction.h
```diff
@@ -3402,6 +3402,26 @@ class SPIRVJointMatrixINTELWorkItemInst : public SPIRVJointMatrixINTELInstBase {
...
+    : public SPIRVInstTemplateBase {
+protected:
+  std::optional<ExtensionID> getRequiredExtension() const override {
+    return ExtensionID::SPV_INTEL_joint_matrix;
+  }
+  SPIRVCapVec getRequiredCapability() const override {
+    return getVec(
+        internal::CapabilityCooperativeMatrixInvocationInstructionsINTEL);
```
vmaksimo, line 3413: we also need `CapabilityCooperativeMatrixKHR`, right?

lib/SPIRV/libSPIRV/SPIRVInstruction.h
```diff
@@ -3402,6 +3402,26 @@ class SPIRVJointMatrixINTELWorkItemInst : public SPIRVJointMatrixINTELInstBase {
...
+  }
+};
+
+#define _SPIRV_OP(x, ...)                                                      \
+  typedef SPIRVInstTemplate<                                                   \
+      SPIRVCooperativeMatrixInvocationInstructionsINTELInstBase,               \
+      internal::Op##x##INTEL, __VA_ARGS__>                                     \
+      SPIRV##x##INTEL;
+_SPIRV_OP(CooperativeMatrixApplyFunction, true, 5)
```
vmaksimo, line 3422:
```suggestion
_SPIRV_OP(CooperativeMatrixApplyFunction, true, 4, true)
```
I guess we won't have the case with no Arguments, but this thing is usually defined as per spec, which says the instruction has only 4 necessary operands.
Could you please also clarify, what `Argument` values mean by spec?

test/extensions/INTEL/SPV_INTEL_joint_matrix/cooperative_matrix_apply.ll
```diff
@@ -0,0 +1,149 @@
+;; compiled from joint_matrix_apply_bf16.cpp from intel/llvm with some modifications
+
+; RUN: llvm-as < %s -o %t.bc
+; RUN: llvm-spirv %t.bc --spirv-ext=+SPV_KHR_cooperative_matrix,+SPV_INTEL_joint_matrix -o %t.spv
```
vmaksimo, line 4: also add the case where only joint matrix extension is enabled

vmaksimo:
> TODO update the spec

What changes are planned to be made in spec?

lib/SPIRV/libSPIRV/SPIRVInstruction.h
VyacheslavLevytskyy: The suggestion is applied by the next commit.

About arguments, by spec, "Argument N is the object to copy to parameter N".

lib/SPIRV/libSPIRV/SPIRVInstruction.h
VyacheslavLevytskyy: The instruction has an argument of OpTypeCooperativeMatrixKHR, so it looks like yes.
@MrSidims can you please confirm if CapabilityCooperativeMatrixKHR must be also returned by getRequiredCapability() along with internal::CapabilityCooperativeMatrixInvocationInstructionsINTEL?

lib/SPIRV/libSPIRV/SPIRVInstruction.h
MrSidims: I have changed it. And it should be:
_SPIRV_OP(CooperativeMatrixApplyFunction, true, 5)
the spec changes will be the following:
before
<html>
<body>

4 + variable | 6448 | <id>Result Type | Result <id> | <id>Function | <id>Matrix | <id>, <id>, …​, <id>Argument 1, Argument 2, …​, Argument N
-- | -- | -- | -- | -- | -- | --



</body>
</html>
after:
<html>
<body>

5 | 6448 | <id>Result Type | Result <id> | <id>Function Object | <id>Matrix 
-- | -- | -- | -- | -- | -- | --

where Function Object is of a type pointer and should represent a function object. Please also note, that '4 + variable' that currently specified in the spec is a spec bug, it's actually '5 + variable'


</body>
</html>

MrSidims: Sorry for the confusion

lib/SPIRV/libSPIRV/SPIRVInstruction.h
MrSidims: This capability will be added when we translate Cooperative Matrix Type. Please also note, that while the specification says, that we rely on cooperative matrix extension, it doesn't say, that one capability implicitly declares another :)

test/extensions/INTEL/SPV_INTEL_joint_matrix/cooperative_matrix_apply.ll
MrSidims: > opt -S -passes=metarenamer < reduced.ll > reduced2.ll

this actually renames also useful things, like functions and structures

test/extensions/INTEL/SPV_INTEL_joint_matrix/cooperative_matrix_apply.ll
MrSidims: Do you have in mind, that joint_matrix extension should implicitly enable cooperative matrix? I'm not expecting to see results other then SPV_KHR_cooperative_matrix/cooperative_matrix.ll would generate :)

lib/SPIRV/libSPIRV/SPIRVInstruction.h
MrSidims: Please ignore my replay. I'm completely wrong.

test/extensions/INTEL/SPV_INTEL_joint_matrix/cooperative_matrix_apply.ll
vmaksimo: ah, alright then :)

lib/SPIRV/libSPIRV/SPIRVInstruction.h
vmaksimo: so we still should add the capability? or is it enabled implicitly elsewhere?

MrSidims: Slava will update the PR, SPIRV/libSPIRV/SPIRVEnum.h should be updated

vmaksimo (approved):

lib/SPIRV/libSPIRV/spirv_internal.hpp
```diff
@@ -78,6 +78,7 @@ enum InternalOp {
   IOpMaskedScatterINTEL = 6429,
   IOpJointMatrixGetElementCoordINTEL = 6440,
   IOpCooperativeMatrixPrefetchINTEL = 6449,
+  IOpCooperativeMatrixApplyFunctionINTEL = 6448,
```
asudarsa, line 81: nit: Reorder to make 6448 come first?

lib/SPIRV/libSPIRV/spirv_internal.hpp
```diff
@@ -104,6 +105,7 @@ enum InternalCapability {
   ICapabilityTensorFloat32RoundingINTEL = 6425,
   ICapabilityMaskedGatherScatterINTEL = 6427,
   ICapabilityJointMatrixWIInstructionsINTEL = 6435,
```
asudarsa, line 107: I see 6435 token specified twice here. Is that accurate?

Thanks

lib/SPIRV/libSPIRV/spirv_internal.hpp
```diff
@@ -104,6 +105,7 @@ enum InternalCapability {
...
+  ICapabilityCooperativeMatrixInvocationInstructionsINTEL = 6435,
```
asudarsa, line 108: May be this question belongs elsewhere. But I will ask here. How do we make sure that the tokens ppl use in Khronos registry do not overlap with these?

Thanks

asudarsa (commented):
LGTM. Just One question about token duplication. 

Thanks

lib/SPIRV/libSPIRV/spirv_internal.hpp
VyacheslavLevytskyy: This is by intention. We have two specifications in design, new https://github.com/intel/llvm/blob/sycl/sycl/doc/design/spirv-extensions/SPV_INTEL_joint_matrix.asciidoc and legacy https://github.com/intel/llvm/blob/sycl/sycl/doc/design/spirv-extensions/SPV_INTEL_joint_matrix_legacy.asciidoc. They both introduce 6435 token. The older one is kept in spirv_internal for backward compatibility, and the newer one is introduced to support implementation of new SPV_INTEL_joint_matrix.asciidoc specification.

lib/SPIRV/libSPIRV/spirv_internal.hpp
MrSidims: Vendors reserve the range of tokens for themselves in https://github.com/KhronosGroup/SPIRV-Headers/blob/main/include/spirv/spir-v.xml . It's up for a particular vendor to setup an internal process to distribute these tokens among vendor extensions.

lib/SPIRV/libSPIRV/spirv_internal.hpp
VyacheslavLevytskyy: fixed

asudarsa (approved):
looks good. Thanks

PR #2216, opened 2023-11-15
Fix duplicate type declarations when using addXXXType api

bwlodarcz:
Duplicate declarations are non conformant with spec which rendered current implementation of SPIRVModuleImpl::addXXXType useless for some types. This commit is an easy fix for bool, float and void. Integer is changed from std::map to SmallDenseMap with 4 inlined buckets. Most of the time Integer will only have 8, 16, 32 and 64 bitwidth so it's much more efficient to use this specialized map.

### DIFF
--- a/lib/SPIRV/libSPIRV/SPIRVModule.cpp
+++ b/lib/SPIRV/libSPIRV/SPIRVModule.cpp
@@ -70,7 +70,7 @@ public:
         SPIRVVersion(static_cast<SPIRVWord>(VersionNumber::SPIRV_1_0)),
         GeneratorId(SPIRVGEN_KhronosLLVMSPIRVTranslator), GeneratorVer(0),
         InstSchema(SPIRVISCH_Default), SrcLang(SourceLanguageOpenCL_C),
-        SrcLangVer(102000) {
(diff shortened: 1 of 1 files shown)


svenvh (approved):

LU-JOHN:
> Duplicate declarations are non conformant with spec which rendered current implementation of SPIRVModuleImpl::addXXXType useless for some types.

Can we add a testcase that requires this fix?

MrSidims:
> > Duplicate declarations are non conformant with spec which rendered current implementation of SPIRVModuleImpl::addXXXType useless for some types.
> 
> Can we add a testcase that requires this fix?

Unfortunately - no. For this we need API unit tests

LU-JOHN:
> > > Duplicate declarations are non conformant with spec which rendered current implementation of SPIRVModuleImpl::addXXXType useless for some types.
> > 
> > 
> > Can we add a testcase that requires this fix?
> 
> Unfortunately - no. For this we need API unit tests

I thought this change would ensure that translation will only produce one SPIRV-ID for a int/float type.  Can we write a LIT test that would have duplicates without this change and no duplicates with this change?  The test should ensure there are no duplicates.

PR #2222, opened 2023-11-16
Support syncscope LLVM argument of atomic instructions

### LINKED ISSUE #1728: Translator doesn't support syncscope LLVM argument of atomic instructions
LLVM has a native way of representing synchronisation scope of atomic instructions and it is done through [`syncscope`](https://llvm.org/docs/LangRef.html#syncscope) argument, which can be applied to functions like [`cmpxchg`](https://llvm.org/docs/LangRef.html#i-cmpxchg), [`atomicrmw`](https://llvm.org/docs/LangRef.html#i-atomicrmw), [`fence`](https://llvm.org/docs/LangRef.html#i-fence), [`atomic load`](https://llvm.org/docs/LangRef.html#i-load), and [`atomic store`](https://llvm.org/docs/LangRef.html#i-store).

An example of how LLVM IR with that attribute looks like (generated from OpenCL) can be found in [`clang/test/CodeGenOpenCL/atomic-ops.cl`](https://github.com/llvm/llvm-project/blob/main/clang/test/CodeGenOpenCL/atomic-ops.cl).

From what I see, the translator currently ignores that `syncscope` attribute, at least for some atomic instructions:

https://github.com/KhronosGroup/SPIRV-LLVM-Translator/blob/b378140863da53bd9fdc313514e00d8c8f45a98f/lib/SPIRV/SPIRVWriter.cpp#L1598-L1607

https://github.com/KhronosGroup/SPIRV-LLVM-Translator/blob/b378140863da53bd9fdc313514e00d8c8f45a98f/lib/SPIRV/SPIRVWriter.cpp#L1586-L1596

https://github.com/KhronosGroup/SPIRV-LLVM-Translator/blob/b378140863da53bd9fdc313514e00d8c8f45a98f/lib/SPIRV/SPIRVWriter.cpp#L2167-L2194

https://github.com/KhronosGroup/SPIRV-LLVM-Translator/blob/b378140863da53bd9fdc313514e00d8c8f45a98f/lib/SPIRV/SPIRVRegularizeLLVM.cpp#L452-L469

Note: some background about why this may be actual can be found in intel/llvm#7464

vmaksimo:
This fixes #1728


test/transcoding/OpenCL/atomic_syncscope_test.ll
```diff
@@ -0,0 +1,157 @@
...
+; 1 - ScopeDevice
+; 2 - ScopeWorkGroup
+; 3 - ScopeSubgroup
+; 4 - ScopeInvocation (mapped from work item)
+
+; OCL scopes
+; 0 - work_item
+; 1 - work_group
+; 2 - device
+; 3 - all_svm_devices
```
MrSidims, line 25: I'm quite sure that it will work, but still all_svm_devices seem to be untested

test/transcoding/OpenCL/atomic_syncscope_test.ll
```diff
@@ -0,0 +1,157 @@
...
+; CHECK-LLVM: call spir_func i32 @_Z20atomic_load_explicitPU3AS4VU7_Atomici12memory_order12memory_scope(ptr{{.*}}, i32 5, i32 1)
+; CHECK-LLVM: call spir_func i32 @_Z20atomic_load_explicitPU3AS4VU7_Atomici12memory_order12memory_scope(ptr{{.*}}, i32 5, i32 2)
+; CHECK-LLVM: call spir_func i32 @_Z20atomic_load_explicitPU3AS4VU7_Atomici12memory_order12memory_scope(ptr{{.*}}, i32 5, i32 2)
+; CHECK-LLVM: call spir_func i32 @_Z20atomic_load_explicitPU3AS4VU7_Atomici12memory_order12memory_scope(ptr{{.*}}, i32 5, i32 4)
+
+define dso_local void @fi1(ptr addrspace(4) nocapture noundef readonly %i) local_unnamed_addr #0 {
+entry:
+  %0 = load atomic i32, ptr addrspace(4) %i syncscope("workgroup") seq_cst, align 4
+  %1 = load atomic i32, ptr addrspace(4) %i syncscope("device") seq_cst, align 4
+  %2 = load atomic i32, ptr addrspace(4) %i seq_cst, align 4
```
MrSidims, line 51: Funny question, what would happen if we pass a non-empty string, that doesn't match the table? I'd hope that we will fall to default (device?), right?

MrSidims:
@LU-JOHN @maksimsab @ @VyacheslavLevytskyy @bwlodarcz please take a look

test/transcoding/OpenCL/atomic_syncscope_test.ll
```diff
@@ -0,0 +1,157 @@
...
+; CHECK-SPIRV-DAG: Constant [[#]] [[#Const2Power30:]] 1073741824 
+; CHECK-SPIRV-DAG: Constant [[#]] [[#ConstInt42:]] 42
+
+; AtomicLoad ResTypeId ResId PtrId MemScopeId MemSemanticsId
+; CHECK-SPIRV: AtomicLoad [[#]] [[#]] [[#]] [[#ConstInt2]] [[#SequentiallyConsistent]]
+; CHECK-SPIRV: AtomicLoad [[#]] [[#]] [[#]] [[#ConstInt1]] [[#SequentiallyConsistent]]
+; CHECK-SPIRV: AtomicLoad [[#]] [[#]] [[#]] [[#ConstInt1]] [[#SequentiallyConsistent]]
+; CHECK-SPIRV: AtomicLoad [[#]] [[#]] [[#]] [[#ConstInt3]] [[#SequentiallyConsistent]]
+
+; CHECK-LLVM: call spir_func i32 @_Z20atomic_load_explicitPU3AS4VU7_Atomici12memory_order12memory_scope(ptr{{.*}}, i32 5, i32 1)
```
asudarsa, line 42: Just curious, why do we not generate atomic load native LLVM instruction here?

Thanks

lib/SPIRV/SPIRVWriter.cpp
```diff
@@ -2308,7 +2314,9 @@ LLVMToSPIRVBase::transValueWithoutDecoration(Value *V, SPIRVBasicBlock *BB,
     // an OpenCL environment, we can use the same memory scope as OpenCL atomic
```
asudarsa, line 2314: Please modify comment based on latest impl. Thanks

lib/SPIRV/SPIRVWriter.cpp
```diff
@@ -2308,7 +2314,9 @@ LLVMToSPIRVBase::transValueWithoutDecoration(Value *V, SPIRVBasicBlock *BB,
...
     // functions that don't have memory_scope argument i.e. memory_scope_device.
     // See the OpenCL C specification p6.13.11. "Atomic Functions"
-    Operands[1] = getUInt32(M, spv::ScopeDevice);
+    SmallVector<StringRef> SSIDs;
+    ARMW->getContext().getSyncScopeNames(SSIDs);
+    Operands[1] = getUInt32(M, map<Scope>(SSIDs[ARMW->getSyncScopeID()].str()));
```
asudarsa, line 2319: Do we need some error checks here? Invalid SyncScope ID or empty string etc?

asudarsa:
Please check and add support for syncscope for cmpxchg here: https://github.com/KhronosGroup/SPIRV-LLVM-Translator/blob/251a75c1fae1b903c3323263dd3638e6f0d5ef3d/lib/SPIRV/SPIRVRegularizeLLVM.cpp#L441C8-L441C8
and for fence here: https://github.com/KhronosGroup/SPIRV-LLVM-Translator/blob/251a75c1fae1b903c3323263dd3638e6f0d5ef3d/lib/SPIRV/SPIRVWriter.cpp#L4697

Thanks

lib/SPIRV/SPIRVWriter.cpp
MrSidims: Empty string is a valid sync scope in LLVM, it's documented to be 'system' sync scope. Also there are no 'invalid' sync scopes. IMO they all should be defaulted to something, may be 'device'.

test/transcoding/OpenCL/atomic_syncscope_test.ll
vmaksimo: I'm afraid it'll fail in the current implementation with "not found in map" error. Isn't it a correct behavior, or do we expect some graceful handling like setting the default value instead?

lib/SPIRV/SPIRVWriter.cpp
vmaksimo: Indeed, empty sting is mapped to 'device' scope now.

test/transcoding/OpenCL/atomic_syncscope_test.ll
MrSidims: What is the current (without the patch) behavior, when sync scope is passed? I believe it should remain for the unknown scopes.

test/transcoding/OpenCL/atomic_syncscope_test.ll
vmaksimo: It's by the current design - atomic instructions are translated as OpenCL builtins. Maybe it'll change in future, but not planned yet.

lib/SPIRV/OCLUtil.cpp
```diff
@@ -145,6 +145,16 @@ template <> void SPIRVMap<OCLScopeKind, Scope>::init() {
...
 }
 
+template <> void SPIRVMap<std::string, Scope>::init() {
+  add("work_item", ScopeInvocation);
+  add("workgroup", ScopeWorkgroup);
+  add("device", ScopeDevice);
+  add("all_svm_devices", ScopeCrossDevice);
+  add("subgroup", ScopeSubgroup);
+  // default
+  add("", ScopeDevice);
```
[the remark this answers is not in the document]
vmaksimo: if there are some doubts, this scope is used by default without my patch. but it can be changed if concerns are strong

test/transcoding/OpenCL/atomic_syncscope_test.ll
```diff
@@ -0,0 +1,157 @@
+; RUN: llvm-as %s -o %t.bc
+; RUN: llvm-spirv %t.bc --spirv-ext=+SPV_EXT_shader_atomic_float_add -o %t.spv
+; RUN: spirv-val %t.spv
+; RUN: llvm-spirv %t.spv -to-text -o %t.spt
+; RUN: FileCheck < %t.spt %s -check-prefix=CHECK-SPIRV
+; RUN: llvm-spirv %t.spv -r --spirv-target-env=CL2.0 -o - | llvm-dis -o %t.rev.ll
```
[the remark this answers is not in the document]
vmaksimo: Do you mean we don't need to check reverse translation as we don't do any changes to SPIRVReader?

test/transcoding/OpenCL/atomic_syncscope_test.ll
vmaksimo: Added checks for all the scopes

test/transcoding/OpenCL/atomic_syncscope_test.ll
vmaksimo: Fixed to preserve the current behavior, and now we don't need the default value for empty string scope `""` (but we still may have one though, if it's different from device)

MrSidims (approved):

test/transcoding/OpenCL/atomic_syncscope_test.ll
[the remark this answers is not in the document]
MrSidims: yeap

lib/SPIRV/SPIRVRegularizeLLVM.cpp
```diff
@@ -467,7 +467,16 @@ bool SPIRVRegularizeLLVMBase::regularize() {
           // comparator, which matches with semantics of the flag returned by
```
asudarsa, line 467: Please modify comment here. Thanks

lib/SPIRV/SPIRVWriter.cpp
```diff
@@ -1731,11 +1731,16 @@ SPIRVValue *LLVMToSPIRVBase::transAtomicStore(StoreInst *ST,
...
   SmallVector<StringRef> SSIDs;
   ST->getContext().getSyncScopeNames(SSIDs);
-  std::vector<Value *> Ops{
-      ST->getPointerOperand(),
-      getUInt32(M, map<Scope>(SSIDs[ST->getSyncScopeID()].str())),
-      getUInt32(M, transAtomicOrdering(ST->getOrdering())),
-      ST->getValueOperand()};
+
+  spv::Scope S;
+  // Fill unknown synscope value to default Device scope.
```
asudarsa, line 1736:
```suggestion
  // Fill unknown syncscope value to default Device scope.
```

lib/SPIRV/SPIRVWriter.cpp
```diff
@@ -1746,9 +1751,15 @@ SPIRVValue *LLVMToSPIRVBase::transAtomicLoad(LoadInst *LD,
                                              SPIRVBasicBlock *BB) {
   SmallVector<StringRef> SSIDs;
   LD->getContext().getSyncScopeNames(SSIDs);
+
+  spv::Scope S;
+  // Fill unknown synscope value to default Device scope.
```
asudarsa, line 1756:
```suggestion
  // Fill unknown syncscope value to default Device scope.
```

asudarsa (changes requested):
Hi @vmaksimo 

Thanks for updates. Couple of changes requested: (1) Please update preceding comments where necessary and (2) Please add support for fence inst a well.

Sincerely

vmaksimo:
> (2) Please add support for fence inst a well.

@asudarsa  Do you mean this part of code related to fence?
https://github.com/KhronosGroup/SPIRV-LLVM-Translator/blob/a0c4f2e3d4e6c18a994d3e4c02c2a8395f70c1f6/lib/SPIRV/SPIRVWriter.cpp#L4739-L4740

I'm not sure if we want to change this too, because this PR is not covering `llvm.fence` instructions. @MrSidims, WDYT?

lib/SPIRV/OCLUtil.cpp
```diff
@@ -145,6 +145,16 @@ template <> void SPIRVMap<OCLScopeKind, Scope>::init() {
   add(OCLMS_sub_group, ScopeSubgroup);
 }
 
+template <> void SPIRVMap<std::string, Scope>::init() {
+  add("work_item", ScopeInvocation);
+  add("workgroup", ScopeWorkgroup);
```
LU-JOHN, line 150: Should this be "workgroup" or "work_group"?  In atomic_syncscope_test.ll the comment on line 23 has "work_group".

lib/SPIRV/SPIRVRegularizeLLVM.cpp
```diff
@@ -467,7 +467,16 @@ bool SPIRVRegularizeLLVMBase::regularize() {
...
           // cmpxchg.
           Value *Ptr = Cmpxchg->getPointerOperand();
-          Value *MemoryScope = getInt32(M, spv::ScopeDevice);
+          SmallVector<StringRef> SSIDs;
+          Cmpxchg->getContext().getSyncScopeNames(SSIDs);
+
+          spv::Scope S;
+          // Fill unknown synscope value to default Device scope.
```
LU-JOHN, line 474: Minor nit.  "synscope" should be "syncscope".

lib/SPIRV/OCLUtil.cpp
LU-JOHN: From:  https://www.khronos.org/assets/uploads/developers/presentations/opencl20-quick-reference-card.pdf

> The default scope is work_group for local atomics and all_svm_devices for global atomics.

vmaksimo: I actually don't mind adding both in sake of usability

MrSidims: > "workgroup" or "work_group"

IMHO we should target what clang can generate, and at this point of time it's 'workgroup'. And thanks for noticing default scope in the spec!

LU-JOHN: We should only accept what the spec says.  If "workgroup" is meaningless according to the spec, it should be treated the same as "wrong_scope".  Personally, I would prefer if unrecognized scope strings were treated as errors.

vmaksimo: So, am I right that all the scopes except "device" should go with underscore? I'm not sure if clang generates such things for now though

test/transcoding/OpenCL/atomic_syncscope_test.ll
```diff
@@ -0,0 +1,158 @@
...
+; 3 - ScopeSubgroup
+; 4 - ScopeInvocation (mapped from work item)
+
+; OCL scopes
+; 0 - work_item
+; 1 - work_group
+; 2 - device
+; 3 - all_svm_devices
+; 4 - sub_group
+
```
LU-JOHN, line 27: Consider testing all scope types (work_item, work_group, ...) for every atomic instruction type (store, load, compare exchange).

lib/SPIRV/OCLUtil.cpp
MrSidims: >> Personally, I would prefer if unrecognized scope strings were treated as errors.

While it is indeed better - I would still recommend not to threat them as errors as we and community will face unknown number of regressions because of it, so I'd preserve the current behavior with the defaulting it if the scope is unknown.

asudarsa (approved):
LGTM. Thanks so much

lib/SPIRV/OCLUtil.cpp
vmaksimo: I suggest to align namings (if needed) and fix default scopes in a separate PR - need to investigate why different scopes were chosen by default earlier, find which scopes clang is generating now, and analyze how the changes would affect e2e tests.

lib/SPIRV/OCLUtil.cpp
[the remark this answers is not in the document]
vmaksimo: To sum it up, I removed the default value from the map.

Now it's set "in place" for each atomic instruction translation (`ScopeDevice`) and for fence inst (`ScopeCrossWorkgroup`). The default value will be set both for empty sting and for wrong scope values (e.g., `"wrong_scope"`).

PR #2228, opened 2023-11-21
Support both legacy and Khronos version of global variable extensions

vmaksimo:
This is done to provide smooth drivers transition to official extensions.
Also updated tests for new extensions to use new tokens after https://github.com/KhronosGroup/SPIRV-Headers/commit/cca08c63cefa129d082abca0302adcb81610b465


vmaksimo:
Please also take a look @LU-JOHN @bwlodarcz @VyacheslavLevytskyy

MrSidims (approved):

Issue #2224, opened 2023-11-18
DWARFv5: static data members declarations are DW_TAG_variable


jsji:
Clang/LLVM has implemented new DWARFv5 issue 161118.1  "DW_TAG for C++ static data members". 

https://github.com/llvm/llvm-project/pull/72234

https://github.com/llvm/llvm-project/pull/72235

We should update llvm-spirv to support it.

Currently, llvm-spirv will assert for the new LLVM IR.

!11 = !DIDerivedType(tag: DW_TAG_variable, name: "smem", scope: !12, file: !3, line: 4, baseType: !15, flags: DIFlagStaticMember)

MrSidims:
@jsji thanks, what is the assert message?

jsji:
See https://github.com/intel/llvm/actions/runs/6910146011/job/18803283913

 Unxpected debug info type for variable
# | UNREACHABLE executed at /__w/llvm/llvm/src/llvm-spirv/lib/SPIRV/LLVMToSPIRVDbgTran.cpp:362!

jsji:
A workaround was implemented in https://github.com/intel/llvm/pull/11920/commits/3f8d7ed61ce0b2261f24d208a660580b32bb442c

PR #2229, opened 2023-11-21
Support DWARF5 static member as DW_TAG_variable

vmaksimo:
This resolves https://github.com/KhronosGroup/SPIRV-LLVM-Translator/issues/2224


MrSidims (approved):

MrSidims:
IR output seems legit.

test/DebugInfo/NonSemantic/static_member_array.ll
```diff
@@ -66,10 +66,10 @@ target triple = "spir64-unknown-unknown"
 !9 = !DIBasicType(name: "int", size: 32, encoding: DW_ATE_signed)
 !10 = !{!11}
 !11 = !DISubrange(count: 4)
-!12 = !DIDerivedType(tag: DW_TAG_member, name: "smem", scope: !13, file: !3, line: 4, baseType: !16, flags: DIFlagStaticMember)
+!12 = !DIDerivedType(tag: DW_TAG_variable, name: "smem", scope: !13, file: !3, line: 4, baseType: !16, flags: DIFlagStaticMember)
```
jsji, line 69: This might not be a good fix here -- this testcase was generated and claimed to be DWARF 4 (See below !19 = !{i32 2, !"Dwarf Version", i32 4}). We shouldn't try to modify it to DWARF5.

I think we may need to create another DWARF5 test using new clang, while keep this one, as we will need to support DWARF 4 producer as wel.

MrSidims: Thanks for noticing, you are right. We will fix it later, we just want to have less customizations in intel/llvm

PR #2230, opened 2023-11-21
DRAFT: Can apply MemoryINTEL decoration multiple times on same variable

artemrad:
According to [SPIRV spec](https://registry.khronos.org/SPIR-V/specs/unified1/SPIRV.html#_decoration) only one instance of a specific decoration can be on a SPIRV-Id, unless the decoration explicitly allows it.  

`MemoryINTEL` is such a decoration where it does not conflict with other instances of the same decoration. The example below is makes semantic sense. 

```
5 Decorate XX MemoryINTEL "DEFAULT" 
6 Decorate XX MemoryINTEL "BLOCK_RAM"
```

### DIFF
--- a/lib/SPIRV/SPIRVWriter.cpp
+++ b/lib/SPIRV/SPIRVWriter.cpp
@@ -3262,8 +3262,8 @@ void addAnnotationDecorations(SPIRVEntry *E, DecorationsInfoVec &Decorations) {
   for (const auto &I : Decorations) {
     // Such decoration already exists on a type, try to skip it
     if (E->hasDecorate(I.first, /*Index=*/0, /*Result=*/nullptr))
-      // Allow multiple UserSemantic Decorations
-      if (I.first != DecorationUserSemantic)
+      // Allow multiple UserSemantic and MemoryINTEL Decorations
+      if (I.first != DecorationUserSemantic && I.first != DecorationMemoryINTEL)
         continue;
 
     switch (static_cast<size_t>(I.first)) {
@@ -3421,8 +3421,8 @@ void addAnnotationDecorationsForStructMember(SPIRVEntry *E,
     // Such decoration already exists on a type, skip it
     if (E->hasMemberDecorate(I.first, /*Index=*/0, MemberNumber,
                              /*Result=*/nullptr))
-      // Allow multiple UserSemantic Decorations
-      if (I.first != DecorationUserSemantic)
+      // Allow multiple UserSemantic and MemoryINTEL Decorations
+      if (I.first != DecorationUserSemantic && I.first != DecorationMemoryINTEL)
         continue;
 
     switch (I.first) {
--- a/lib/SPIRV/SPIRVReader.cpp
+++ b/lib/SPIRV/SPIRVReader.cpp
@@ -3480,9 +3480,12 @@ void generateIntelFPGAAnnotation(
     Out << "{register:1}";
 
   SPIRVWord Result = 0;
-  if (E->hasDecorate(DecorationMemoryINTEL))
-    Out << "{memory:"
-        << E->getDecorationStringLiteral(DecorationMemoryINTEL).front() << '}';
+  if (E->hasDecorate(DecorationMemoryINTEL)) {
+    auto Annotations = E->getAllDecorationStringLiterals(DecorationMemoryINTEL);
+    for (size_t I = 0; I != Annotations.size(); ++I) {
+      Out << "{memory:" << Annotations[I].front() << '}';
+    }
+  }
   if (E->hasDecorate(DecorationBankwidthINTEL, 0, &Result))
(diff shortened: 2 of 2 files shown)


vmaksimo (commented):
Change itself looks ok, but we need tests to verify.

The main concern is that [specification](https://github.com/KhronosGroup/SPIRV-Registry/blob/main/extensions/INTEL/SPV_INTEL_fpga_memory_attributes.asciidoc) does not mention that we allow `MemoryINTEL` decoration several times. Is it a planned change to the spec?

test/extensions/INTEL/SPV_INTEL_fpga_memory_attributes/IntelFPGAMemoryINTELMultiDecoration.ll
```diff
@@ -0,0 +1,72 @@
...
+@.str.2 = private unnamed_addr addrspace(1) constant [30 x i8] c"{5826:\22DEFAULT\22}{5826:\22MLAB\22}\00", section "llvm.metadata"
+@.str.3 = private unnamed_addr addrspace(1) constant [35 x i8] c"{5826:\22DEFAULT\22}{5826:\22BLOCK_RAM\22}\00", section "llvm.metadata"
+
+; CHECK-SPV-IR: [[empty_annot:]] = private unnamed_addr constant [17 x i8] c"{memory:DEFAULT}\00", align 1
+; CHECK-SPV-IR: [[mlab_annot:]] = private unnamed_addr constant [14 x i8] c"{memory:MLAB}\00", align 1
+; CHECK-SPV-IR: [[block_ram_annot:]] = private unnamed_addr constant [19 x i8] c"{memory:BLOCK_RAM}\00", align 1
+
+; CHECK-LLVM: [[empty_annot:]] = private unnamed_addr constant [17 x i8] c"{memory:DEFAULT}\00", align 1
+; CHECK-LLVM: [[mlab_annot:]] = private unnamed_addr constant [14 x i8] c"{memory:MLAB}\00", align 1
+; CHECK-LLVM: [[block_ram_annot:]] = private unnamed_addr constant [19 x i8] c"{memory:BLOCK_RAM}\00", align 1
```
vmaksimo, line 39: These checks are basically the same, I guess we can have only one tag (e.g., `CHECK-LLVM`) and check it for both reverse runs. Same for the IR below

test/extensions/INTEL/SPV_INTEL_fpga_memory_attributes/IntelFPGAMemoryINTELMultiDecoration.ll
```diff
@@ -0,0 +1,72 @@
...
+; RUN: FileCheck < %t.rev.ll %s --check-prefix=CHECK-SPV-IR
+; RUN: llvm-spirv -r %t.spv -o %t.rev.bc
+; RUN: llvm-dis %t.rev.bc
+; RUN: FileCheck < %t.rev.ll %s --check-prefix=CHECK-LLVM
+
+; CHECK-SPIRV: Capability FPGAMemoryAttributesINTEL
+; CHECK-SPIRV: Extension "SPV_INTEL_fpga_memory_attributes"
+; CHECK-SPIRV: Decorate [[#empty:]] MemoryINTEL "DEFAULT" 
+; CHECK-SPIRV: Decorate [[#mlab:]] MemoryINTEL "MLAB"
+; CHECK-SPIRV: Decorate [[#block_ram:]] MemoryINTEL "BLOCK_RAM"
```
vmaksimo, line 19: Can I ask you to add checks that prove that we don't generate `"DEFAULT"` more than once?

vmaksimo (commented):
Overall LGTM. Please address comments and change the title and the description - we actually don't apply `MemoryINTEL` decoration twice or multiple times

vmaksimo (approved):

lib/SPIRV/SPIRVWriter.cpp
```diff
@@ -3154,6 +3154,28 @@ AnnotationDecorations tryParseAnnotationString(SPIRVModule *BM,
                        internal::DecorationCacheControlLoadINTEL) {
           Decorates.CacheControlVec.emplace_back(
               static_cast<Decoration>(DecorationKind), std::move(DecValues));
+        } else if (DecorationKind == DecorationMemoryINTEL) {
+          // SPIRV doesn't allow the same Decoration to be applied multiple
+          // times on a single SPIRVEntry, unless explicitly allowed by the
+          // language spec. Filter out the less specific MemoryINTEL
+          // decorations, if applied multiple times
+          auto CanFilterOut = [](auto Val) {
```
MrSidims, line 3162: nit
```suggestion
          auto CanFilterOut = [](auto &Val) {
```

MrSidims (approved):

PR #2232, opened 2023-11-24
Fix reverse translation for a decoration applied to a function parameter

MrSidims:
This fixes reverse translation for SPV_INTEL_fpga_memory_accesses. Theoretically it fixes more, but seems like without the extension we don't generate UserSemantic decoration in this case at all. Will be investigated separately.


vmaksimo (approved):
LGTM.
Don't forget to follow up with the tests/fix for UserSemantic issue

PR #2235, opened 2023-11-24
Fixing Intrinsic::ptr_annotation

artemrad:
There is a fundamental misalignment between when translating llvm.ptr.annotation to [SPIRV's OpMemberDecorate](https://registry.khronos.org/SPIR-V/specs/1.0/SPIRV.html#OpMemberDecorate).

llvm.ptr.annotation is applied onto a pointer.

OpMemberDecorate is applied on a type, which means that all variables of this type will have this Decoration.

This change fixes this misalignment by translating llvm.ptr.annotation to [SPIRV's OpDecorate](https://registry.khronos.org/SPIR-V/specs/1.0/SPIRV.html#OpDecorate) instead, which instead places the decoration onto the pointer.

### DIFF
--- a/lib/SPIRV/SPIRVWriter.cpp
+++ b/lib/SPIRV/SPIRVWriter.cpp
@@ -3131,6 +3131,8 @@ AnnotationDecorations tryParseAnnotationString(SPIRVModule *BM,
 
     std::pair<StringRef, StringRef> Split = AnnotatedDecoration.split(':');
     StringRef Name = Split.first, ValueStr = Split.second;
+    SPIRVDBG(spvdbgs() << "[tryParseAnnotationString]: AnnotationString: "
+                       << Name.str() << "\n");
 
     unsigned DecorationKind = 0;
     if (!Name.getAsInteger(10, DecorationKind)) {
@@ -4281,19 +4283,17 @@ SPIRVValue *LLVMToSPIRVBase::transIntrinsicInst(IntrinsicInst *II,
   // i8* <ptr> is a pointer on a GV, which can carry optinal variadic
   // clang::annotation attribute expression arguments.
   case Intrinsic::ptr_annotation: {
-    // Strip all bitcast and addrspace casts from the pointer argument:
-    //   llvm annotation intrinsic only takes i8*, so the original pointer
-    //   probably had to loose its addrspace and its original type.
-    Value *AnnotSubj = II->getArgOperand(0);
-    while (isa<BitCastInst>(AnnotSubj) || isa<AddrSpaceCastInst>(AnnotSubj)) {
-      AnnotSubj = cast<CastInst>(AnnotSubj)->getOperand(0);
+    Value *AnnotSubj = nullptr;
+    if (auto *BI = dyn_cast<BitCastInst>(II->getArgOperand(0))) {
+      AnnotSubj = BI->getOperand(0);
+    } else {
+      AnnotSubj = II->getOperand(0);
     }
 
     std::string AnnotationString;
     processAnnotationString(II, AnnotationString);
     AnnotationDecorations Decorations =
         tryParseAnnotationString(BM, AnnotationString);
-
     // Translate FPGARegIntel annotations to OpFPGARegINTEL.
     if (AnnotationString == kOCLBuiltinName::FPGARegIntel) {
       auto *Ty = transScavengedType(II);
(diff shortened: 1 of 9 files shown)


lib/SPIRV/SPIRVWriter.cpp
```diff
@@ -4281,19 +4283,17 @@ SPIRVValue *LLVMToSPIRVBase::transIntrinsicInst(IntrinsicInst *II,
...
-    // Strip all bitcast and addrspace casts from the pointer argument:
-    //   llvm annotation intrinsic only takes i8*, so the original pointer
-    //   probably had to loose its addrspace and its original type.
-    Value *AnnotSubj = II->getArgOperand(0);
-    while (isa<BitCastInst>(AnnotSubj) || isa<AddrSpaceCastInst>(AnnotSubj)) {
-      AnnotSubj = cast<CastInst>(AnnotSubj)->getOperand(0);
+    Value *AnnotSubj = nullptr;
+    if (auto *BI = dyn_cast<BitCastInst>(II->getArgOperand(0))) {
+      AnnotSubj = BI->getOperand(0);
+    } else {
```
MrSidims, line 4289: Do we even need this code with this approach?

lib/SPIRV/SPIRVWriter.cpp
```diff
@@ -4303,76 +4303,26 @@ SPIRVValue *LLVMToSPIRVBase::transIntrinsicInst(IntrinsicInst *II,
...
-    // to the first member of the struct; if so, attempt to get a struct type
-    // from an alloca instead.
-    SPIRVType *StructTy = nullptr;
-    [[maybe_unused]] SPIRVValue *ResPtr = nullptr;
-    [[maybe_unused]] SPIRVWord MemberNumber = 0;
-    if (auto *const GI = dyn_cast<GetElementPtrInst>(AnnotSubj)) {
-      if (auto *const STy = dyn_cast<StructType>(GI->getSourceElementType())) {
-        StructTy = transType(STy);
-        ResPtr = transValue(GI, BB);
-        MemberNumber = dyn_cast<ConstantInt>(GI->getOperand(2))->getZExtValue();
```
MrSidims, line 4318: That actually seems to be a bug in previous implementation as GEP in bounds won't work with this code.

test/extensions/INTEL/SPV_INTEL_fpga_memory_attributes/IntelFPGAMemoryAttributesForStruct.ll
```diff
@@ -116,21 +87,22 @@ target triple = "spir64-unknown-unknown"
...
+; CHECK-LLVM: call void @llvm.var.annotation{{.*}}(ptr {{.*}}, ptr [[SINGLEPUMP]]
+; CHECK-LLVM: call void @llvm.var.annotation{{.*}}(ptr {{.*}}, ptr [[DOUBLEPUMP]]
+; CHECK-LLVM: call void @llvm.var.annotation{{.*}}(ptr {{.*}}, ptr [[MAX_REPLICATES]]
+; CHECK-LLVM: call void @llvm.var.annotation{{.*}}(ptr {{.*}}, ptr [[SIMPLE_DUAL_PORT]]
+; CHECK-LLVM: call void @llvm.var.annotation{{.*}}(ptr {{.*}}, ptr [[MERGE]]
+; CHECK-LLVM: call void @llvm.var.annotation{{.*}}(ptr {{.*}}, ptr [[BANK_BITS]]
+; CHECK-LLVM: call void @llvm.var.annotation{{.*}}(ptr {{.*}}, ptr [[FORCE_POW_2_DEPTH]]
+; CHECK-LLVM: call void @llvm.var.annotation{{.*}}(ptr {{.*}}, ptr [[STRIDESIZE]]
+; CHECK-LLVM: call void @llvm.var.annotation{{.*}}(ptr {{.*}}, ptr [[WORDSIZE]]
+; CHECK-LLVM: call void @llvm.var.annotation{{.*}}(ptr {{.*}}, ptr [[TRUE_DUAL_PORT]]
```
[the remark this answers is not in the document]
artemrad: This is because the SPIRV reverse translator relies on a heuristic to decide how to translate SPIRV `OpDecoration`. Ignoring casts, if the decoration is on an `alloca` translate to an `llvm.var.annotation` else translate to `llvm.ptr.annotation`. 
Prior to this change `OpMemberDecoration` will always map to `llvm.ptr.annotation`.

lib/SPIRV/SPIRVWriter.cpp
MrSidims: Shall we then also skip addrspace casts?

artemrad: If we don't skip the bitcasts - then in the reverse translation the `OpDecorate` will get translated to `llvm.ptr.annotation` in even more cases, as the heuristic (discussed) above will not trigger. I think it is important to keep as many `llvm.var.annotations` as possible, as they provide stronger guarantees in the LLVM IR than `llvm.ptr.annotations`, and are more likely to get interpenetrated by the backendm rather than dropped.

MrSidims:
@vmaksimo @bwlodarcz please take a look

MrSidims (approved):

PR #2236, opened 2023-11-27
Fix assert failure of spirv.Image constant null value

wenju-he:
There should be no assert for the value in SPIRVConstantNull::validate.

### DIFF
--- a/lib/SPIRV/libSPIRV/SPIRVValue.h
+++ b/lib/SPIRV/libSPIRV/SPIRVValue.h
@@ -272,6 +272,7 @@ protected:
     assert((Type->isTypeComposite() || Type->isTypeOpaque() ||
             Type->isTypeEvent() || Type->isTypePointer() ||
             Type->isTypeReserveId() || Type->isTypeDeviceEvent() ||
+            Type->isTypeImage() ||
             (Type->isTypeSubgroupAvcINTEL() &&
              !Type->isTypeSubgroupAvcMceINTEL())) &&
            "Invalid type");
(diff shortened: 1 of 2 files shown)


wenju-he:
```
RUN: at line 9: /usr/bin/spirv-val /home/runner/work/SPIRV-LLVM-Translator/SPIRV-LLVM-Translator/build/projects/SPIRV-LLVM-Translator/test/test_output/transcoding/Output/spirv-target-types.ll.tmp.spv
+ /usr/bin/spirv-val /home/runner/work/SPIRV-LLVM-Translator/SPIRV-LLVM-Translator/build/projects/SPIRV-LLVM-Translator/test/test_output/transcoding/Output/spirv-target-types.ll.tmp.spv
error: line 64: OpConstantNull Result Type <id> '13[%13]' cannot have a null value.
  %29 = OpConstantNull %13
```
https://registry.khronos.org/SPIR-V/specs/unified1/SPIRV.html#OpConstantNull can't be a image type. 
However, we do have `target("spirv.Image", void, 0, 0, 0, 0, 0, 0, 1) zeroinitializer` in spv IR.
@MrSidims @jcranmer-intel do you have suggestion how to proceed this PR?

MrSidims (changes requested):
The spec doesn't allow it. I suggest not to emit zero initializer in clang in the first place. Is it generated with -O0?

wenju-he:
> The spec doesn't allow it. I suggest not to emit zero initializer in clang in the first place. Is it generated with -O0?

yes, it is generated by frontend in -O0 mode.

PR #2240, opened 2023-11-28
fix scope=0 in spirv.CooperativeMatrixKHR type

VyacheslavLevytskyy:
Change Scope argument to one of two available options: ScopeWorkgroup/ScopeWorkgroup.


VyacheslavLevytskyy:
@MrSidims This fixes unsupported scope argument.

MrSidims (changes requested):
While it's indeed hard to imagine CrossDevice cooperative matrix, the spec doesn't disallow that. @svenvh do you think, we should do such adjustments in the test(s)?

svenvh:
> While it's indeed hard to imagine CrossDevice cooperative matrix, the spec doesn't disallow that. @svenvh do you think, we should do such adjustments in the test(s)?

The spec indeed allows any scope, so it would be good to keep different scope types in the tests, to maximize test coverage.

PR #2251, opened 2023-12-05
add support for the 'fcmp false' instruction

VyacheslavLevytskyy:
There is no direct equivalent for "fcmp false" in the specification [https://registry.khronos.org/SPIR-V/specs/unified1/SPIRV.html#_relational_and_logical_instructions], but we may return ConstantFalse in this case instead of reporting error.

This patch depends on https://github.com/KhronosGroup/SPIRV-LLVM-Translator/pull/2216

### DIFF
--- a/lib/SPIRV/SPIRVWriter.cpp
+++ b/lib/SPIRV/SPIRVWriter.cpp
@@ -2045,6 +2045,8 @@ LLVMToSPIRVBase::transValueWithoutDecoration(Value *V, SPIRVBasicBlock *BB,
   }
 
   if (CmpInst *Cmp = dyn_cast<CmpInst>(V)) {
+    if(Cmp->getPredicate() == CmpInst::Predicate::FCMP_FALSE)
+      return mapValue(V, BM->addConstant(BM->addBoolType(), 0));
     SPIRVInstruction *BI = transCmpInst(Cmp, BB);
     return mapValue(V, BI);
   }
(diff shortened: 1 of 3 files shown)


VyacheslavLevytskyy:
This patch depends on https://github.com/KhronosGroup/SPIRV-LLVM-Translator/pull/2216. In xmain it would fail without the fixing duplicate type declarations.

bwlodarcz:
Unfortunately the 'fcmp' instruction can go through vector also. I prepared test:
```
define spir_func <4 x i1> @g(<4 x float> %0) {
entry:
 %1 = fcmp false <4 x float> %0, %0
 ret <4 x i1> %1
}
``` 
llvm-as and translator passes but spirv-val
```
error: line 28: OpReturnValue Value <id> '15[%false_0]'s type does not match OpFunction's return type.
  OpReturnValue %false_0
```
Without further investigation I interpret this error that ConstantFalse is constant type declaration. In SPIRV `BoolType != Constant BoolType`. So the return type of the function is not the same as the one returned.

VyacheslavLevytskyy:
@bwlodarcz Thank you. Let's represent a vector of false values as a OpConstantNull ("The null value is type dependent, defined as follows: - Scalar Boolean: false ... Members are set recursively to the null constant according to the null value of their constituent types.")

VyacheslavLevytskyy:
Please take a look @asudarsa @LU-JOHN

bwlodarcz:
LGTM.

MrSidims (approved):

asudarsa (approved):
LGTM. Thanks

Issue #2233, opened 2023-11-24
Multiple issues after running llvm-spirv test-suite with ASAN.


bwlodarcz:
**Description**: ASAN reports multiple issues after running test-suite. 
**Steps to reproduce**:
1. Clone llvm-project and go into it.
2. Clone Translator inside llvm/projects.
3. Run `cmake -B build -G Ninja -DCMAKE_BUILD_TYPE=Debug -DCMAKE_EXPORT_COMPILE_COMMANDS=ON -DLLVM_ENABLE_PROJECTS=clang -DLLVM_BUILD_TESTS=ON -DLLVM_OPTIMIZED_TABLEGEN=ON -DLLVM_USE_SANITIZER='Address' -DLLVM_SPIRV_INCLUDE_TESTS=ON ./llvm`.
4. Run `cmake --build build` and wait until build completes.
5. Go to build directory and run `ninja check-llvm-spirv`.

**Results**
Big chunk of ASAN errors and:
```
Failed Tests (119):
  LLVM_SPIRV :: DebugInfo/DebugInfoChecksum.ll
  LLVM_SPIRV :: DebugInfo/DebugInfoChecksumCompileUnit.ll
  LLVM_SPIRV :: DebugInfo/DebugInfoLLVMArg.ll
  LLVM_SPIRV :: DebugInfo/DebugInfoTypeBasic.ll
  LLVM_SPIRV :: DebugInfo/DebugInfoTypeInheritance.ll
  LLVM_SPIRV :: DebugInfo/Generic/inlined-locations.ll
  LLVM_SPIRV :: DebugInfo/NonSemantic/DIImportedEntity.ll
  ...TRUNCATED...
  LLVM_SPIRV :: transcoding/OpGroupAllAny.ll
  LLVM_SPIRV :: transcoding/image_signedness.ll
  LLVM_SPIRV :: transcoding/sub_group_ballot.ll
  LLVM_SPIRV :: transcoding/sub_group_clustered_reduce.ll
  LLVM_SPIRV :: transcoding/sub_group_non_uniform_arithmetic.ll
  LLVM_SPIRV :: transcoding/sub_group_non_uniform_vote.ll


Testing Time: 21.15s

Total Discovered Tests: 751
  Passed           : 628 (83.62%)
  Expectedly Failed:   4 (0.53%)
  Failed           : 119 (15.85%)
```
**Full text file from run:** 
[...]

PR #2259, opened 2023-12-08
Memory leak fix in SPIRVDecoder::getSourceContinuedInstructions

bwlodarcz:
SPIRVEntry created in function getEntry isn't deleted when it is ExtOp and isn't SourceContinued. Fixes most of ASan errors in issue #2233.


VyacheslavLevytskyy:
LGTM

LU-JOHN (commented):
Is there a way to add a testcase that would fail without this change?

asudarsa (approved):
LGTM. Thanks

asudarsa:
> Is there a way to add a testcase that would fail without this change?

I think, for such changes, we can rely on isolated testing strategy (i.e. this no longer is reported as a memory leak in some memory analysis).

Thanks

bwlodarcz:
> > Is there a way to add a testcase that would fail without this change?
> 
> I think, for such changes, we can rely on isolated testing strategy (i.e. this no longer is reported as a memory leak in some memory analysis).
> 
> Thanks

Yes, after the change ASan is silent during DebugSource tests. To recreate failed test cases you can use reproducing steps from #2233. After this change number of failed test cases drops from ~15% to ~2%.

vmaksimo (approved):

PR #2262, opened 2023-12-12
Access out of scope fix in OCLToSPIRVBase::transBuiltin

bwlodarcz:
OldRetTy pointer was passed to lambda by reference. This means that in lambda it's real C-like type was Value ** pointing to the stack within of the transBuiltin stack frame. Logic of lambda is executed in doConversion function executed in destructor of BuiltinCallMutator. The stack frame is unaccessible during the destructor call and it worked only as undfined behaviour.


asudarsa (approved):
LGTM overall.

Thanks

svenvh (approved):

lib/SPIRV/OCLToSPIRV.cpp
```diff
@@ -912,16 +912,16 @@ void OCLToSPIRVBase::transBuiltin(CallInst *CI, OCLBuiltinTransInfo &Info) {
     Info.UniqName = getSPIRVFuncName(BVKind);
   } else
     return;
-  auto Mutator = mutateCallInst(CI, Info.UniqName + Info.Postfix);
+  BuiltinCallMutator Mutator = mutateCallInst(CI, Info.UniqName + Info.Postfix);
   Info.PostProc(Mutator);
+  Type *OldRetTy = CI->getType();
```
vmaksimo, line 917: Do we need to move this definition from if-condition? My guess is that we can leave it as it was

vmaksimo (approved):
LGTM, just one nit

lib/SPIRV/OCLToSPIRV.cpp
bwlodarcz: True.

svenvh (approved):

PR #2265, opened 2023-12-13
Access out of scope fix in OCLToSPIRVBase::visitCallReadWriteImage

bwlodarcz:
ImageOpMask pointer was passed to closure as reference. Closure is executed outside of original stack frame which renders reference invalid. Fixes some ASan errors from issue #2233.

### DIFF
--- a/lib/SPIRV/OCLToSPIRV.cpp
+++ b/lib/SPIRV/OCLToSPIRV.cpp
@@ -1078,8 +1078,9 @@ void OCLToSPIRVBase::visitCallReadWriteImage(CallInst *CI,
     Info.UniqName = kOCLBuiltinName::ReadImage;
     unsigned ImgOpMask = getImageSignZeroExt(DemangledName);
     if (ImgOpMask) {
-      Info.PostProc = [&](BuiltinCallMutator &Mutator) {
-        Mutator.appendArg(getInt32(M, ImgOpMask));
+      Module *Mod = M;
+      Info.PostProc = [ImgOpMask, Mod](BuiltinCallMutator &Mutator) {
+        Mutator.appendArg(getInt32(Mod, ImgOpMask));
       };
     }
   }


lib/SPIRV/OCLToSPIRV.cpp
```diff
@@ -1078,8 +1078,9 @@ void OCLToSPIRVBase::visitCallReadWriteImage(CallInst *CI,
     Info.UniqName = kOCLBuiltinName::ReadImage;
     unsigned ImgOpMask = getImageSignZeroExt(DemangledName);
     if (ImgOpMask) {
-      Info.PostProc = [&](BuiltinCallMutator &Mutator) {
-        Mutator.appendArg(getInt32(M, ImgOpMask));
+      Module *Mod = M;
```
vmaksimo, line 1081: I get the thing with `ImgOpMask` reference, but why do we need to create a new pointer to the module?

bwlodarcz: Because if anything is captured by [] part of lambda definition - than it's not a pure function with mangled name.
It became a closure which in cpp means that's an instance of anonymous class which have members created from parameter captured.
Something like this:
```
int a;
int b;
auto func = [a, b](){ return a + b; };
```
in intuitive simplified view is something similar to this: 
```
class Closure {
  int a;
  int b;
  Closure(int a, int b): a(a), b(b) {}
  int operator()() {
    return this->a + this->b;
  }
};

int a;
int b;
Closure func(a, b);
```
Reality is of course more complicated as closure is an instance of std::function, there are some conversions etc.

Going to the core of the question. Instance of closure doesn't have pointer to Module as member. 
That's why we need to explicitly pass it.

asudarsa (approved):
LGTM. Thanks

MrSidims (approved):

PR #2266, opened 2023-12-13
Support forward translation of fpmath metadata

sarnex:
The Clang frontend uses the `fpmath` metadata to represent FP accuracy in some cases, right now for OpenCL and possibly SYCL in the near future. The long term plan is to expect the frontend to add the `fpbuiltin-max-error` attribute, but since `fpmath` is generated today and currently dropped, let's support it using the existing `SPV_INTEL_fp_max_error` extension.

Upon reverse translation, we generate the same `fpbuiltin-max-error` attribute/metadata that we would generate if the frontend had added `fpbuiltin-max-error` originally, so anybody using the reverse translation result only needs to handle one case.


lib/SPIRV/SPIRVWriter.cpp
```diff
@@ -302,6 +302,33 @@ static bool recursiveType(const StructType *ST, const Type *Ty) {
   return Run(Ty);
 }
 
+// Add decoration if needed
```
asudarsa, line 305: Good catch. I do not know why I was returning the same pointer that I pass in.
I suppose there are no other changes here?

Thanks

sarnex: Thanks for the review!

It wasn't wrong or anything it just turns out we didn't actually need it.

The only other changes to the function are:
1) Instead of asserting that an `IntrinsicInst` is an `fpbuiltin,` just skip them if they are not so that we can call this function in all cases

2) Handle fpmath metadata

asudarsa (approved):
LGTM. Thanks

lib/SPIRV/SPIRVWriter.cpp
```diff
@@ -302,6 +302,33 @@ static bool recursiveType(const StructType *ST, const Type *Ty) {
...
+void addFPBuiltinDecoration(SPIRVModule *BM, Instruction *Inst,
+                            SPIRVInstruction *I) {
+  const bool AllowFPMaxError =
+      BM->isAllowedToUseExtension(ExtensionID::SPV_INTEL_fp_max_error);
+  if (!AllowFPMaxError)
+    return;
```
vmaksimo, line 311: Does it seem better?
```suggestion
  if (!BM->isAllowedToUseExtension(ExtensionID::SPV_INTEL_fp_max_error))
    return;
```

lib/SPIRV/SPIRVWriter.cpp
```diff
@@ -302,6 +302,33 @@ static bool recursiveType(const StructType *ST, const Type *Ty) {
...
+  if (auto *II = dyn_cast_or_null<IntrinsicInst>(Inst);
+      II && II->getCalledFunction()->getName().startswith("llvm.fpbuiltin")) {
```
vmaksimo, line 313: Can we add a test case for such builtins?

Also such syntax construction is not used in the repo, I find it a bit less obvious than usage of `if (II = dyn_cast())`. Do you mind to change the style here?

lib/SPIRV/SPIRVWriter.cpp
sarnex: Yes sure, thanks will do!

lib/SPIRV/SPIRVWriter.cpp
sarnex: Thanks for the review, We need to use II later in the loop but let me change the style.

sarnex: BTW we already have tests for the builtins here: `test/extensions/INTEL/SPV_INTEL_fp_max_error/IntelFPMaxError.ll`

sarnex:
@vmaksimo Feedback should be addressed in latest commit, thank you!

lib/SPIRV/SPIRVWriter.cpp
```diff
@@ -302,6 +302,31 @@ static bool recursiveType(const StructType *ST, const Type *Ty) {
   return Run(Ty);
 }
 
+// Add decoration if needed
+void addFPBuiltinDecoration(SPIRVModule *BM, Instruction *Inst,
```
vmaksimo, line 306: Guess we need to remove the original function now https://github.com/KhronosGroup/SPIRV-LLVM-Translator/blob/main/lib/SPIRV/SPIRVWriter.cpp#L3630 ?

sarnex: ah dang, nice catch

sarnex: fixed, thx

vmaksimo (approved):
Thanks!

sarnex:
Out of tree failures not related

PR #2267, opened 2023-12-14
Document contract between LLVM IR producer and the translator wrt. global variable declarations

VyacheslavLevytskyy:
Global variable declarations are not permitted to have no address space or ``addrspace(0)``. This PR is to document this contract.

### DIFF
--- a/docs/SPIRVRepresentationInLLVM.rst
+++ b/docs/SPIRVRepresentationInLLVM.rst
@@ -398,9 +398,9 @@ Global variables
 A global variable resides in an address space, and the default address space
 in LLVM is zero. The SPIR-V storage class represented by the zero LLVM IR
 address spaces is Function. However, SPIR-V global variable declarations are
-``OpVariable`` instructions whose Storage Class is not ``Function``. This
-means that global variable declarations are not permitted to have no address
-space or ``addrspace(0)``.
+``OpVariable`` instructions whose Storage Class cannot be ``Function``. This
+means that global variable declarations must always have an address space
+specified and that address space cannot be ``0``.
 
 Function metadata
 -----------------
--- a/docs/SPIRVRepresentationInLLVM.rst
+++ b/docs/SPIRVRepresentationInLLVM.rst
@@ -392,6 +392,16 @@ Calling convention
 A function with ``spir_kernel`` calling convention will be translated as an entry
 point of the SPIR-V module.
 
+Global variables
+----------------
+
+A global variable resides in an address space, and the default address space
+in LLVM is zero. The SPIR-V storage class represented by the zero LLVM IR
+address spaces is Function. However, SPIR-V global variable declarations are
+``OpVariable`` instructions whose Storage Class is not ``Function``. This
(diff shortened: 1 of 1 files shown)


docs/SPIRVRepresentationInLLVM.rst
```diff
@@ -392,6 +392,16 @@ Calling convention
 A function with ``spir_kernel`` calling convention will be translated as an entry
 point of the SPIR-V module.
 
+Global variables
+----------------
+
+A global variable resides in an address space, and the default address space
+in LLVM is zero. The SPIR-V storage class represented by the zero LLVM IR
+address spaces is Function. However, SPIR-V global variable declarations are
+``OpVariable`` instructions whose Storage Class is not ``Function``. This
```
asudarsa, line 401:
```suggestion
``OpVariable`` instructions whose Storage Class cannot be ``Function``. This
```

docs/SPIRVRepresentationInLLVM.rst
```diff
@@ -392,6 +392,16 @@ Calling convention
...
+means that global variable declarations are not permitted to have no address
+space or ``addrspace(0)``.
```
asudarsa, line 403:
```suggestion
means that LLVM global variable declarations must always have an address space specified and that address space cannot be 0``.
```

asudarsa (commented):
Looks good. Minor nits

VyacheslavLevytskyy:
@asudarsa Thank you, updated

asudarsa (approved):
LGTM. Thanks

docs/SPIRVRepresentationInLLVM.rst
```diff
@@ -392,6 +392,16 @@ Calling convention
...
 
+Global variables
+----------------
+
+A global variable resides in an address space, and the default address space
+in LLVM is zero. The SPIR-V storage class represented by the zero LLVM IR
+address spaces is Function. However, SPIR-V global variable declarations are
+``OpVariable`` instructions whose Storage Class cannot be ``Function``. This
+means that global variable declarations must always have an address space
+specified and that address space cannot be ``0``.
```
vmaksimo, line 403: I'm not sure about that, sorry if I missed some part of the conversation and/or problem statement.
I think we should not prohibit LLVM from generating no address space for global variables. Maybe we should just use the appropriate storage class in SPIR-V in such case, making adjustment on our own. What do you think?

bader: > I think we should not prohibit LLVM from generating no address space for global variables.

Could you please clarify why we shouldn't do that?

> Maybe we should just use the appropriate storage class in SPIR-V in such case, making adjustment on our own. What do you think?

What would be the appropriate storage class? I think we can't assign meaningful storage class as global variables are used to represent at least three different cases: function scope variable in local address space, program scope variables in global or constant address space. How would you differentiate between these cases?

MrSidims (approved):

svenvh (approved):

PR #2269, opened 2023-12-14
Memory leak fix in LLVMToSPIRVBase::transExecutionMode

bwlodarcz:
Allocated SPIRVExecutionMode class wasn't added as dependant resource in SPIRVModule. Fix some ASan errors in issue #2233.


asudarsa (approved):
This seems like a miss. I checked other sites and we do BM->add everywhere else.
Very good find. 

Thanks

PR #2270, opened 2023-12-15
README.md: State Kernel capability restriction

svenvh:
Mention explicitly in the README that only SPIR-V modules that declare the `Kernel` capability are supported.  This is a key restriction for anyone looking to translate between (arbitrary) SPIR-V and LLVM IR, so mention it at the top of the README.

This came up in the OpenCL Tooling TSG call of Dec 14th; tagging @asudarsa, @bashbaug, @keryell.


keryell (approved):
Thanks for the clarification.

bashbaug (approved):

AlexeySachkov (approved):

asudarsa (approved):
LGTM. Thanks

PR #2274, opened 2023-12-18
Change ConstantInt.getType to ConstantInt.getIntegerType

bwlodarcz:
Due to change in LLVM IR in commit dea16ebd2613a4a218c53045270fc4fcc9b427ad.


svenvh (approved):
LGTM; out-of-tree build failures are due to packages not having the breaking commit yet.

PR #2276, opened 2023-12-19
Revert "Allow cooperative matrix types for OpMatrixTimesScalar"

vmaksimo:
Per discussion in Khronos and the spec itself, `OpMatrixTimesScalar` should not be used in OpenCL environment, as long as the same can be achieved either via an elementwise multiplication with a cooperative matrix object created from the scalar using `OpConstantComposite` or by iterating over the elements of the cooperative matrix to multiply each element by the scalar.

Spec: https://github.com/KhronosGroup/SPIRV-Registry/blob/main/extensions/KHR/SPV_KHR_cooperative_matrix.asciidoc


MrSidims (approved):
LGTM, please address CI issues

PR #2279, opened 2023-12-20
[SPIRV-V Extension] fpbuiltin-max-error w/o extension

LU-JOHN:
Report an error if fpbuiltin-max-error found without extension enabled

### DIFF
--- a/lib/SPIRV/SPIRVWriter.cpp
+++ b/lib/SPIRV/SPIRVWriter.cpp
@@ -305,12 +305,16 @@ static bool recursiveType(const StructType *ST, const Type *Ty) {
 // Add decoration if needed
 void addFPBuiltinDecoration(SPIRVModule *BM, Instruction *Inst,
                             SPIRVInstruction *I) {
-  if (!BM->isAllowedToUseExtension(ExtensionID::SPV_INTEL_fp_max_error))
-    return;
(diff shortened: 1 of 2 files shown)


lib/SPIRV/SPIRVWriter.cpp
```diff
@@ -320,6 +323,8 @@ void addFPBuiltinDecoration(SPIRVModule *BM, Instruction *Inst,
                      convertFloatToSPIRVWord(F));
     }
   } else if (auto *MD = Inst->getMetadata("fpmath")) {
+    if (!AllowFPMaxError)
+      return;
```
vmaksimo, line 327: Should we report an error here like it's done for `llvm.fpbuiltin`?
Also it looks like this change is not tested in the PR.

LU-JOHN: I discussed this with @sarnex, and he said "fpmath" has been used for a long time to represent FP accuracy, so we cannot start reporting errors with it.  

The change to report an error with "llvm.fpbuiltin" is tested with the CHECK_NO_CAPABILITY_ERROR in IntelFPMaxError.ll

vmaksimo (approved):

MrSidims (approved):

PR #2280, opened 2023-12-21
[SPV->LLVM] Fix global c/dtors type if SPV is from opaque type LLVM

wenju-he:
When SPV is generated from LLVM IR with opaque pointer enabled, ctor
function in global ctors has opaque pointer type rather than function
type. When translating the SPV back to LLVM IR with typed pointer like
in LLVM 14, ctor function type is casted to i8* pointer in the global
ctors initializer. This results in error in LLVM verifier.
This PR fixes the issue by removing the cast.

### DIFF
--- a/lib/SPIRV/SPIRVReader.cpp
+++ b/lib/SPIRV/SPIRVReader.cpp
@@ -3816,7 +3816,7 @@ static GlobalVariable *mutateGlobalCtorDtors(GlobalVariable *GV) {
 }
 
 void SPIRVToLLVM::transGlobalCtorDtors(SPIRVVariable *BV) {
-  GlobalVariable *V = cast<GlobalVariable>(transValue(BV, nullptr, nullptr));
+  auto *V = cast<GlobalVariable>(transValue(BV, nullptr, nullptr));
   V = mutateGlobalCtorDtors(V);
   V->setLinkage(GlobalVariable::AppendingLinkage);
 }
--- a/lib/SPIRV/SPIRVReader.cpp
+++ b/lib/SPIRV/SPIRVReader.cpp
@@ -1331,39 +1331,6 @@ void SPIRVToLLVM::transFunctionPointerCallArgumentAttributes(
   }
 }
 
-/// When spirv is generated from LLVM IR with opapque pointer enabled and then
-/// the spirv is translated to LLVM IR with typed pointer, function pointer is
-/// casted to i8* type in \p Ty and \p Initializer, which causes error in LLVM
-/// IR verifier. E.g.
-///   [1 x %0][%0 { i32 1, i8* bitcast (void ()* @ctor to i8*), i8* null }]
-/// This function removes the cast so that LLVM IR is valid.
-static void postProcGlobalCtorDtorTypeInit(Type *&Ty, Constant *&Initializer) {
-  auto *InitArr = cast<ConstantArray>(Initializer);
-  unsigned NumEltsInArr = InitArr->getType()->getNumElements();
-  assert(NumEltsInArr && "array is empty");
-  auto *CS = cast<ConstantStruct>(InitArr->getAggregateElement(0u));
-  assert(CS->getType()->getNumElements() == 3 && "expect 3 elements in struct");
-  auto *Elt1 = CS->getAggregateElement(1);
-  if (!isa<ConstantExpr>(Elt1))
-    return;
-  auto *StTy = CS->getType();
-  auto *NewStTy = StructType::create(Ty->getContext(),
-                                     {StTy->getElementType(0),
-                                      Elt1->stripPointerCasts()->getType(),
-                                      StTy->getElementType(2)},
-                                     StTy->getName(), StTy->isPacked());
-  Ty = ArrayType::get(NewStTy, NumEltsInArr);
-  SmallVector<Constant *, 4> ArrElts;
-  for (unsigned I = 0; I < NumEltsInArr; ++I) {
-    auto *CS = cast<ConstantStruct>(InitArr->getAggregateElement(I));
-    ArrElts.push_back(ConstantStruct::get(
-        NewStTy, {CS->getAggregateElement(0u),
-                  CS->getAggregateElement(1)->stripPointerCasts(),
-                  CS->getAggregateElement(2)}));
-  }
-  Initializer = ConstantArray::get(cast<ArrayType>(Ty), ArrElts);
-}
-
 /// For instructions, this function assumes they are created in order
 /// and appended to the given basic block. An instruction may use a
 /// instruction from another BB which has not been translated. Such
@@ -1589,28 +1556,20 @@ Value *SPIRVToLLVM::transValueWithoutDecoration(SPIRVValue *BV, Function *F,
     SPIRVBuiltinVariableKind BVKind;
     if (BVar->isBuiltin(&BVKind))
       BV->setName(prefixSPIRVName(SPIRVBuiltInNameMap::map(BVKind)));
-    bool IsCtorOrDtor = (BV->getName() == "llvm.global_ctors" ||
-                         BV->getName() == "llvm.global_dtors");
-    if (Init && IsCtorOrDtor) {
-      Initializer = dyn_cast<Constant>(transValue(Init, F, BB, false));
-      postProcGlobalCtorDtorTypeInit(Ty, Initializer);
-    }
     auto LVar = new GlobalVariable(*M, Ty, IsConst, LinkageTy,
                                    /*Initializer=*/nullptr, BV->getName(), 0,
                                    GlobalVariable::NotThreadLocal, AddrSpace);
     auto Res = mapValue(BV, LVar);
-    if (Init) {
-      if (!IsCtorOrDtor)
-        Initializer = dyn_cast<Constant>(transValue(Init, F, BB, false));
-    } else if (LinkageTy == GlobalValue::CommonLinkage) {
+    if (Init)
+      Initializer = dyn_cast<Constant>(transValue(Init, F, BB, false));
+    else if (LinkageTy == GlobalValue::CommonLinkage)
       // In LLVM, variables with common linkage type must be initialized to 0.
       Initializer = Constant::getNullValue(Ty);
-    } else if (BS == SPIRVStorageClassKind::StorageClassWorkgroup) {
+    else if (BS == SPIRVStorageClassKind::StorageClassWorkgroup)
       Initializer = dyn_cast<Constant>(UndefValue::get(Ty));
-    } else if ((LinkageTy != GlobalValue::ExternalLinkage) &&
-               (BS == SPIRVStorageClassKind::StorageClassCrossWorkgroup)) {
+    else if ((LinkageTy != GlobalValue::ExternalLinkage) &&
+             (BS == SPIRVStorageClassKind::StorageClassCrossWorkgroup))
       Initializer = Constant::getNullValue(Ty);
-    }
 
     LVar->setUnnamedAddr((IsConst && Ty->isArrayTy() &&
                           Ty->getArrayElementType()->isIntegerTy(8))
@@ -3327,10 +3286,11 @@ bool SPIRVToLLVM::translate() {
 
   for (unsigned I = 0, E = BM->getNumVariables(); I != E; ++I) {
     auto *BV = BM->getVariable(I);
-    if (BV->getStorageClass() != StorageClassFunction)
-      transValue(BV, nullptr, nullptr);
-    else
+    if (BV->getName() == "llvm.global_ctors" ||
+        BV->getName() == "llvm.global_dtors")
       transGlobalCtorDtors(BV);
+    else if (BV->getStorageClass() != StorageClassFunction)
+      transValue(BV, nullptr, nullptr);
   }
 
   // Then translate all debug instructions.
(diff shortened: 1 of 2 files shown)


wenju-he:
The PR is needed for both llvm_release_140 and llvm_release_150 branches since opaque pointer isn't enabled by default until LLVM 16.

wenju-he:
```
D:\a\SPIRV-LLVM-Translator\SPIRV-LLVM-Translator\llvm-project\llvm\projects\SPIRV-LLVM-Translator\lib\SPIRV\libSPIRV/SPIRVNameMapEnum.h(597,7): error C2065: 'CapabilityLongConstantCompositeINTEL': undeclared identifier (compiling source file D:\a\SPIRV-LLVM-Translator\SPIRV-LLVM-Translator\llvm-project\llvm\projects\SPIRV-LLVM-Translator\lib\SPIRV\OCLToSPIRV.cpp) [D:\a\SPIRV-LLVM-Translator\SPIRV-LLVM-Translator\build\projects\SPIRV-LLVM-Translator\lib\SPIRV\LLVMSPIRVLib.vcxproj]
```
Error in CI in-tree windows build isn't related to this PR

lib/SPIRV/SPIRVReader.cpp
```diff
@@ -1331,6 +1331,39 @@ void SPIRVToLLVM::transFunctionPointerCallArgumentAttributes(
   }
 }
 
+/// When spirv is generated from LLVM IR with opapque pointer enabled and then
```
VyacheslavLevytskyy, line 1334: typo: opaque

lib/SPIRV/SPIRVReader.cpp
```diff
@@ -1331,6 +1331,39 @@ void SPIRVToLLVM::transFunctionPointerCallArgumentAttributes(
...
+/// the spirv is translated to LLVM IR with typed pointer, function pointer is
+/// casted to i8* type in \p Ty and \p Initializer, which causes error in LLVM
+/// IR verifier. E.g.
+///   [1 x %0][%0 { i32 1, i8* bitcast (void ()* @ctor to i8*), i8* null }]
+/// This function removes the cast so that LLVM IR is valid.
```
VyacheslavLevytskyy, line 1339: I'd like to better understand the scope of the problem when we get an opaque pointer type rather than function
type, and so interpret it further like a i8* pointer when reverse translate it back to LLVM. Does this behavior manifest itself for ctor/dtor only, and that is caused by special treatment of ctor/dtor case in other places of the source code base? Or this is rather a wider problem that is not only about ctor/dtor?

lib/SPIRV/SPIRVReader.cpp
```diff
@@ -1331,6 +1331,39 @@ void SPIRVToLLVM::transFunctionPointerCallArgumentAttributes(
...
+static void postProcGlobalCtorDtorTypeInit(Type *&Ty, Constant *&Initializer) {
+  auto *InitArr = cast<ConstantArray>(Initializer);
+  unsigned NumEltsInArr = InitArr->getType()->getNumElements();
+  assert(NumEltsInArr && "array is empty");
+  auto *CS = cast<ConstantStruct>(InitArr->getAggregateElement(0u));
+  assert(CS->getType()->getNumElements() == 3 && "expect 3 elements in struct");
```
VyacheslavLevytskyy, line 1345: This new function is not added to validate, so I'm not sure you need new asserts here at all. It looks like the following logic can be just protected by if conditions, for example, to avoid adding implicit layer of validity checks into a new ad-hoc function that addresses ctor/dtor case. However, if other reviewers think that asserts are ok here and may remain as it is, I'd strongly suggest to consider better error messages, adding more context to the message to facilitate possible future testing/debugging efforts.

lib/SPIRV/SPIRVReader.cpp
```diff
@@ -1556,20 +1589,28 @@ Value *SPIRVToLLVM::transValueWithoutDecoration(SPIRVValue *BV, Function *F,
     SPIRVBuiltinVariableKind BVKind;
     if (BVar->isBuiltin(&BVKind))
       BV->setName(prefixSPIRVName(SPIRVBuiltInNameMap::map(BVKind)));
+    bool IsCtorOrDtor = (BV->getName() == "llvm.global_ctors" ||
+                         BV->getName() == "llvm.global_dtors");
+    if (Init && IsCtorOrDtor) {
+      Initializer = dyn_cast<Constant>(transValue(Init, F, BB, false));
+      postProcGlobalCtorDtorTypeInit(Ty, Initializer);
+    }
```
VyacheslavLevytskyy, line 1597: It's confusing to have this check separated from if-statement below, and to repeat the check for `IsCtorOrDtor` twice, here and below.

lib/SPIRV/SPIRVReader.cpp
```diff
@@ -1331,6 +1331,39 @@ void SPIRVToLLVM::transFunctionPointerCallArgumentAttributes(
   }
 }
 
+/// When spirv is generated from LLVM IR with opapque pointer enabled and then
+/// the spirv is translated to LLVM IR with typed pointer, function pointer is
+/// casted to i8* type in \p Ty and \p Initializer, which causes error in LLVM
+/// IR verifier. E.g.
+///   [1 x %0][%0 { i32 1, i8* bitcast (void ()* @ctor to i8*), i8* null }]
+/// This function removes the cast so that LLVM IR is valid.
+static void postProcGlobalCtorDtorTypeInit(Type *&Ty, Constant *&Initializer) {
```
VyacheslavLevytskyy, line 1340: The function is defined to be void, as if it would cause a side effect rather than convert one statement into another, as it actually does by reassigning an argument `Initializer` . It would require more time to understand the underlying logic comparing to line 1595 style of
` Initializer = dyn_cast<Constant>(transValue(Init, F, BB, false));`.
In my opinion, it's better to return a pointer than to change the `*&` argument.

lib/SPIRV/SPIRVReader.cpp
```diff
@@ -1556,20 +1589,28 @@ Value *SPIRVToLLVM::transValueWithoutDecoration(SPIRVValue *BV, Function *F,
     SPIRVBuiltinVariableKind BVKind;
     if (BVar->isBuiltin(&BVKind))
       BV->setName(prefixSPIRVName(SPIRVBuiltInNameMap::map(BVKind)));
+    bool IsCtorOrDtor = (BV->getName() == "llvm.global_ctors" ||
+                         BV->getName() == "llvm.global_dtors");
+    if (Init && IsCtorOrDtor) {
+      Initializer = dyn_cast<Constant>(transValue(Init, F, BB, false));
+      postProcGlobalCtorDtorTypeInit(Ty, Initializer);
```
VyacheslavLevytskyy, line 1596: As in the comment to line 1340, it's not obvious that the function changes Initializer value. `Initializer = ` would work better.

LU-JOHN:
We may want to look at lib/SPIRV/SPIRVTypeScavenger.cpp, which was written to:

> // This file implements the necessary logic to recover pointer types from LLVM
// IR for the output SPIR-V file from the opaque pointers in LLVM IR.

It would be better if the SPIRVTypeScavenger was able to get the correct type rather than adding custom code for global c/dtors.

lib/SPIRV/SPIRVReader.cpp
wenju-he: I agree with your concern that this behavior could be a wider problem.
I was also thinking why this issue is exposed so late. My guess is that IGC who uses llvm-14 branch switched to use the open-source SPIRV-LLVM-Translator not long time ago.

wenju-he:
> We may want to look at lib/SPIRV/SPIRVTypeScavenger.cpp, which was written to:
> 
> > // This file implements the necessary logic to recover pointer types from LLVM
> > // IR for the output SPIR-V file from the opaque pointers in LLVM IR.
> 
> It would be better if the SPIRVTypeScavenger was able to get the correct type rather than adding custom code for global c/dtors.
> 
> @jcranmer-intel can you comment if this issue is something that is better handled in the type scavenger?

Yeah we probably need to back-port SPIRVTypeScavenger to llvm_release_140 and llvm_release_150 branches. I'm not familiar with the code in SPIRVTypeScavenger. @jcranmer-intel what's your opinion?

lib/SPIRV/SPIRVReader.cpp
```diff
@@ -3286,10 +3286,11 @@ bool SPIRVToLLVM::translate() {
...
   for (unsigned I = 0, E = BM->getNumVariables(); I != E; ++I) {
     auto *BV = BM->getVariable(I);
-    if (BV->getStorageClass() != StorageClassFunction)
-      transValue(BV, nullptr, nullptr);
-    else
+    if (BV->getName() == "llvm.global_ctors" ||
+        BV->getName() == "llvm.global_dtors")
       transGlobalCtorDtors(BV);
+    else if (BV->getStorageClass() != StorageClassFunction)
+      transValue(BV, nullptr, nullptr);
```
wenju-he, line 3293: the change of order here is to fix https://github.com/KhronosGroup/SPIRV-LLVM-Translator/blob/main/test/extensions/INTEL/SPV_INTEL_function_pointers/global_ctor_dtor.ll and https://github.com/KhronosGroup/SPIRV-LLVM-Translator/blob/main/test/extensions/INTEL/SPV_INTEL_function_pointers/global_ctor_dtor_addrspace.ll in the case SPV from opaque pointer LLVM IR is translated to typed pointer LLVM IR.

wenju-he:
> > We may want to look at lib/SPIRV/SPIRVTypeScavenger.cpp, which was written to:
> > > // This file implements the necessary logic to recover pointer types from LLVM
> > > // IR for the output SPIR-V file from the opaque pointers in LLVM IR.
> > 
> > 
> > It would be better if the SPIRVTypeScavenger was able to get the correct type rather than adding custom code for global c/dtors.
> > @jcranmer-intel can you comment if this issue is something that is better handled in the type scavenger?
> 
> Yeah we probably need to back-port SPIRVTypeScavenger to llvm_release_140 and llvm_release_150 branches. I'm not familiar with the code in SPIRVTypeScavenger. @jcranmer-intel what's your opinion?
> 
> edit: it seems @jcranmer-intel is on holiday now.

@LU-JOHN back-porting SPIRVTypeScavenger requires some efforts that probably can't be done in a short time. Can we proceed with PR at the moment since the fix is urgently needed? I can file an issue for back-porting SPIRVTypeScavenger .

wenju-he:
> > > We may want to look at lib/SPIRV/SPIRVTypeScavenger.cpp, which was written to:
> > > > // This file implements the necessary logic to recover pointer types from LLVM
> > > > // IR for the output SPIR-V file from the opaque pointers in LLVM IR.
> > > 
> > > 
> > > It would be better if the SPIRVTypeScavenger was able to get the correct type rather than adding custom code for global c/dtors.
> > > @jcranmer-intel can you comment if this issue is something that is better handled in the type scavenger?
> > 
> > 
> > Yeah we probably need to back-port SPIRVTypeScavenger to llvm_release_140 and llvm_release_150 branches. I'm not familiar with the code in SPIRVTypeScavenger. @jcranmer-intel what's your opinion?
> > edit: it seems @jcranmer-intel is on holiday now.
> 
> @LU-JOHN back-porting SPIRVTypeScavenger requires some efforts that probably can't be done in a short time. Can we proceed with PR at the moment since the fix is urgently needed? I can file an issue for back-porting SPIRVTypeScavenger .

@VyacheslavLevytskyy @LU-JOHN please also note that in main branch SPIRVTypeScavenger is only used for LLVM->SPIRV translation. The scope here is SPIRV->LLVM. So it needs efforts to evaluate if SPIRVTypeScavenger can be modified to suite this scope.
If there is no objection to the current approach, could you please review and merge? We'd like it to be merged ASAP.

LU-JOHN:
> > > > We may want to look at lib/SPIRV/SPIRVTypeScavenger.cpp, which was written to:
> > > > > // This file implements the necessary logic to recover pointer types from LLVM
> > > > > // IR for the output SPIR-V file from the opaque pointers in LLVM IR.
> > > > 
> > > > 
> > > > It would be better if the SPIRVTypeScavenger was able to get the correct type rather than adding custom code for global c/dtors.
> > > > @jcranmer-intel can you comment if this issue is something that is better handled in the type scavenger?
> > > 
> > > 
> > > Yeah we probably need to back-port SPIRVTypeScavenger to llvm_release_140 and llvm_release_150 branches. I'm not familiar with the code in SPIRVTypeScavenger. @jcranmer-intel what's your opinion?
> > > edit: it seems @jcranmer-intel is on holiday now.
> > 
> > 
> > @LU-JOHN back-porting SPIRVTypeScavenger requires some efforts that probably can't be done in a short time. Can we proceed with PR at the moment since the fix is urgently needed? I can file an issue for back-porting SPIRVTypeScavenger .
> 
> @VyacheslavLevytskyy @LU-JOHN please also note that in main branch SPIRVTypeScavenger is only used for LLVM->SPIRV translation. The scope here is SPIRV->LLVM. So it needs efforts to evaluate if SPIRVTypeScavenger can be modified to suite this scope. If there is no objection to the current approach, could you please review and merge? We'd like it to be merged ASAP.

Hi @wenju-he .  I see you have addressed all of @VyacheslavLevytskyy concerns.  Can you look at while "In-tree build & tests / Windows" is failing?

wenju-he:
> Hi @wenju-he . I see you have addressed all of @VyacheslavLevytskyy concerns. Can you look at while "In-tree build & tests / Windows" is failing?

@LU-JOHN the fail is due to wrong commit is checked out from KhronosGroup/SPIRV-Headers repo in "In-tree build & tests / Windows":
```
Checking out the ref
  "C:\Program Files\Git\bin\git.exe" checkout --progress --force -B main refs/remotes/origin/main
  branch 'main' set up to track 'origin/main'.
  Switched to a new branch 'main'
"C:\Program Files\Git\bin\git.exe" log -1 --format='%H'
'1c6bb2743599e6eb6f37b2969acc0aef812e32e3'
```
Linux in-tree build is checking out correct commit:
```
Checking out the ref
 /usr/bin/git checkout --progress --force 9b527c0fb60124936d0906d44803bec51a0200fb
  Note: switching to '9b527c0fb60124936d0906d44803bec51a0200fb'.
  
  You are in 'detached HEAD' state. You can look around, make experimental
  changes and commit them, and you can discard any commits you make in this
  state without impacting any branches by switching back to a branch.
  
  If you want to create a new branch to retain commits you create, you may
  do so (now or later) by using -c with the switch command. Example:
  
    git switch -c <new-branch-name>
  
  Or undo this operation with:
  
    git switch -
  
  Turn off this advice by setting config variable advice.detachedHead to false
  
  HEAD is now at 9b527c0 Add definitions for SPV_KHR_cooperative_matrix
/usr/bin/git log -1 --format='%H'
'9b527c0fb60124936d0906d44803bec51a0200fb'
```
It isn't obvious to me which part of https://github.com/KhronosGroup/SPIRV-LLVM-Translator/blob/llvm_release_140/.github/workflows/check-in-tree-build.yml is wrong.

wenju-he:
> > Hi @wenju-he . I see you have addressed all of @VyacheslavLevytskyy concerns. Can you look at while "In-tree build & tests / Windows" is failing?
> 
> @LU-JOHN my local windows in-tree build is successful. The fail in "In-tree build & tests / Windows" is due to wrong commit is checked out from KhronosGroup/SPIRV-Headers repo :

submitted an issue: https://github.com/KhronosGroup/SPIRV-LLVM-Translator/issues/2284

LU-JOHN (approved):
LGTM

bader:
@wenju-he wants it to be merged asap. @VyacheslavLevytskyy, @jcranmer-intel, please, review post-commit. @wenju-he, be ready to address post-commit comments in follow-up PRs.

wenju-he:
Thank you @bader

PR #2285, opened 2023-12-29
[DebugInfo] Preserve DIExpression in DIGlobalVariableExpression

LU-JOHN:
Preserve DIExpression in DIGlobalVariableExpression.

Preserving DIExpressions is necessary because of this change in LLVM:

commit 638a8393615e911b729d5662096f60ef49f1c65e
Author: Michael Buch <michaelbuch12@gmail.com>
Date:   Mon Nov 13 06:04:27 2023 +0000

    Reland "[clang][DebugInfo] Emit global variable definitions for static data members with constant initializers" (#71780)

### DIFF
--- a/lib/SPIRV/LLVMToSPIRVDbgTran.cpp
+++ b/lib/SPIRV/LLVMToSPIRVDbgTran.cpp
@@ -1184,9 +1184,8 @@ LLVMToSPIRVDbgTran::transDbgGlobalVariable(const DIGlobalVariable *GV) {
   Ops[FlagsIdx] = transDebugFlags(GV);
 
   // Check if GV is the definition of previously declared static member
-  DIDerivedType *StaticMember = GV->getStaticDataMemberDeclaration();
-  Ops.push_back(StaticMember ? transDbgEntry(StaticMember)->getId()
-                             : getDebugInfoNoneId());
+  if (DIDerivedType *StaticMember = GV->getStaticDataMemberDeclaration())
+    Ops.push_back(transDbgEntry(StaticMember)->getId());
 
   // Check if Ops[VariableIdx] has no information
   if (Ops[VariableIdx] == getDebugInfoNoneId()) {
--- a/lib/SPIRV/LLVMToSPIRVDbgTran.cpp
+++ b/lib/SPIRV/LLVMToSPIRVDbgTran.cpp
@@ -1185,11 +1185,11 @@ LLVMToSPIRVDbgTran::transDbgGlobalVariable(const DIGlobalVariable *GV) {
 
   // Check if GV is the definition of previously declared static member
   DIDerivedType *StaticMember = GV->getStaticDataMemberDeclaration();
-  Ops.push_back(StaticMember ? transDbgEntry(StaticMember)->getId() :
-                getDebugInfoNoneId());
+  Ops.push_back(StaticMember ? transDbgEntry(StaticMember)->getId()
+                             : getDebugInfoNoneId());
 
   for (const DIGlobalVariableExpression *G : DIF.global_variables()) {
-    if (G->getVariable()==GV) {
+    if (G->getVariable() == GV) {
       Ops.push_back(transDbgExpression(G->getExpression())->getId());
       break;
     }
--- a/lib/SPIRV/LLVMToSPIRVDbgTran.cpp
+++ b/lib/SPIRV/LLVMToSPIRVDbgTran.cpp
@@ -1188,12 +1188,14 @@ LLVMToSPIRVDbgTran::transDbgGlobalVariable(const DIGlobalVariable *GV) {
   Ops.push_back(StaticMember ? transDbgEntry(StaticMember)->getId()
                              : getDebugInfoNoneId());
 
-  for (const DIGlobalVariableExpression *G : DIF.global_variables()) {
-    if (G->getVariable() == GV) {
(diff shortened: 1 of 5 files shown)


lib/SPIRV/LLVMToSPIRVDbgTran.cpp
```diff
@@ -1184,8 +1184,18 @@ LLVMToSPIRVDbgTran::transDbgGlobalVariable(const DIGlobalVariable *GV) {
...
+                             : getDebugInfoNoneId());
+
+  if (BM->getDebugInfoEIS() == SPIRVEIS_NonSemantic_Shader_DebugInfo_200)
+    // Check if GV has an associated GVE with a DIExpression
+    for (const DIGlobalVariableExpression *GVE : DIF.global_variables()) {
+      if (GVE->getVariable() == GV) {
+        Ops.push_back(transDbgExpression(GVE->getExpression())->getId());
+        break;
+      }
+    }
```
MrSidims, line 1198: If my understanding is correct, `<id> Variable` from `DebugGlobalVariable` serves (or can serve) the same purpose and the expression can be handled by `getGlobalVariable` function. Could you please check what the translator generates for `<id> Variable` and can it be reused later in lib/SPIRV/SPIRVToLLVMDbgTran.cpp to generate the expression?

test/DebugInfo/DebugInfo-GV-with-DIE.ll
```diff
@@ -0,0 +1,24 @@
+;; Ensure that DIExpressions are preserved in DIGlobalVariableExpressions
+;; This utilizes SPIRV DebugGlobalVariable's optional field DIExpression
+
+; RUN: llvm-as %s -o %t.bc
+; RUN: llvm-spirv -o %t.spv %t.bc --spirv-debug-info-version=nonsemantic-shader-200
```
MrSidims, line 5: like this we are not limited to nonsemantic-shader-200, but also include nonsemantic-shader-100

test/DebugInfo/DebugInfo-GV-with-DIE.ll
```diff
@@ -0,0 +1,24 @@
...
+; RUN: llvm-spirv -r -o %t.rev.bc %t.spv
+; RUN: llvm-dis %t.rev.bc -o %t.rev.ll
+; RUN: FileCheck %s --input-file %t.rev.ll
+
+; CHECK: ![[#]] = !DIGlobalVariableExpression(var: ![[#GV:]], expr: !DIExpression(DW_OP_constu, 1, DW_OP_stack_value))
```
MrSidims, line 10: would be nice to have SPV checks as well

test/DebugInfo/DebugInfo-GV-with-SMD-and-DIE.ll
```diff
@@ -0,0 +1,41 @@
+;; Ensure that DIExpressions are preserved in DIGlobalVariableExpressions
```
MrSidims, line 1: same comments are applicable to this test as well

MrSidims (approved):
LGTM, just 2 nits
would be nice if someone else also takes a look

test/DebugInfo/DebugInfo-GV-with-DIE.ll
LU-JOHN: Added checks for SPV.

test/DebugInfo/DebugInfo-GV-with-DIE.ll
LU-JOHN: Added checking for nonsemantic-shader-100

test/DebugInfo/DebugInfo-GV-with-SMD-and-DIE.ll
LU-JOHN: Added checking for nonsemantic-shader-100.  Added SPV testing.

lib/SPIRV/LLVMToSPIRVDbgTran.cpp
```diff
@@ -1187,6 +1187,23 @@ LLVMToSPIRVDbgTran::transDbgGlobalVariable(const DIGlobalVariable *GV) {
   if (DIDerivedType *StaticMember = GV->getStaticDataMemberDeclaration())
     Ops.push_back(transDbgEntry(StaticMember)->getId());
 
+  // Check if Ops[VariableIdx] has no information
+  if (Ops[VariableIdx] == getDebugInfoNoneId()) {
```
vmaksimo, line 1191: Is it possible that we don't get `DebugInfoNone`, and also have DIExpression?

vmaksimo:
Please also leave a comment in description why we are doing the change (some community change I suppose?)

lib/SPIRV/LLVMToSPIRVDbgTran.cpp
LU-JOHN: In all my experiments I could not get Ops[VariableIdx]!=DebugInfoNone and also have a DIExpression, but I cannot guarantee this will not happen.  The DIExpression holds an initial value.  It seems to only be set if the variable is const, and thus the variable seems to be eliminated and Ops[VariableIdx]==DebugInfoNone.

LU-JOHN:
> Please also leave a comment in description why we are doing the change (some community change I suppose?)

Updated description with reference to LLVM change that drove this update to llvm-spirv.

vmaksimo (approved):

PR #2286, opened 2024-01-04
[SPV-IR -> SPIRV] Fix image builtin unsigned type not preserved

wenju-he:
This PR preserves unsigned return type of image read and unsigned texel
type of image write builtins as ZeroExtend image operand in SPIRV.

### DIFF
--- a/lib/SPIRV/SPIRVInternal.h
+++ b/lib/SPIRV/SPIRVInternal.h
@@ -892,6 +892,13 @@ inline bool getParameterTypes(CallInst *CI, SmallVectorImpl<Type *> &ArgTys) {
   return getParameterTypes(CI->getCalledFunction(), ArgTys);
 }
 
+enum ParamSignedness { Signed = 0, Unsigned, Unknown };
+
+/// Extract signedness of return type and parameter types from a mangled
+/// function name.
+bool getRetParamSignedness(Function *F, ParamSignedness &RetSignedness,
+                           SmallVectorImpl<ParamSignedness> &ArgSignedness);
+
 /// Mangle a function from OpenCL extended instruction set in SPIR-V friendly IR
 /// manner
 std::string getSPIRVFriendlyIRFunctionName(OCLExtOpKind ExtOpId,
--- a/lib/SPIRV/SPIRVInternal.h
+++ b/lib/SPIRV/SPIRVInternal.h
@@ -892,7 +892,7 @@ inline bool getParameterTypes(CallInst *CI, SmallVectorImpl<Type *> &ArgTys) {
   return getParameterTypes(CI->getCalledFunction(), ArgTys);
 }
 
-enum ParamSignedness { Signed = 0, Unsigned, Unknown };
+enum class ParamSignedness { Signed = 0, Unsigned, Unknown };
 
 /// Extract signedness of return type and parameter types from a mangled
 /// function name.
--- a/lib/SPIRV/SPIRVUtil.cpp
+++ b/lib/SPIRV/SPIRVUtil.cpp
@@ -881,6 +881,46 @@ bool getParameterTypes(Function *F, SmallVectorImpl<Type *> &ArgTys,
   return DemangledSuccessfully;
 }
 
+bool getRetParamSignedness(Function *F, ParamSignedness &RetSignedness,
+                           SmallVectorImpl<ParamSignedness> &ArgSignedness) {
+  using namespace llvm::itanium_demangle;
+  StringRef Name = F->getName();
+  if (!Name.starts_with("_Z") || F->arg_empty())
+    return false;
+
+  ManglingParser<DefaultAllocator> Demangler(Name.begin(), Name.end());
+  // If it's not a function name encoding, bail out.
+  auto *RootNode = dyn_cast_or_null<FunctionEncoding>(Demangler.parse());
+  if (!RootNode)
+    return false;
+
+  auto GetSignedness = [](const itanium_demangle::Node *N) {
+    if (!N)
+      return ParamSignedness::Unknown;
+    if (const auto *Vec = dyn_cast<itanium_demangle::VectorType>(N))
+      N = Vec->getBaseType();
+    if (const auto *Name = dyn_cast<NameType>(N)) {
+      StringRef Arg(stringify(Name));
+      if (Arg.starts_with("unsigned"))
+        return ParamSignedness::Unsigned;
+      else if (Arg.equals("char") || Arg.equals("short") || Arg.equals("int") ||
+               Arg.equals("long"))
+        return ParamSignedness::Signed;
+    }
+    return ParamSignedness::Unknown;
+  };
+  RetSignedness = GetSignedness(RootNode->getReturnType());
+  ArgSignedness.resize(F->arg_size());
+  for (const auto &[I, ParamType] : llvm::enumerate(RootNode->getParams())) {
+    if (F->getArg(I)->getType()->isIntOrIntVectorTy())
+      ArgSignedness[I] = GetSignedness(ParamType);
+    else
+      ArgSignedness[I] = ParamSignedness::Unknown;
+  }
+
+  return true;
+}
+
 CallInst *mutateCallInst(
     Module *M, CallInst *CI,
     std::function<std::string(CallInst *, std::vector<Value *> &)> ArgMutate,
--- a/lib/SPIRV/SPIRVUtil.cpp
+++ b/lib/SPIRV/SPIRVUtil.cpp
@@ -903,8 +903,8 @@ bool getRetParamSignedness(Function *F, ParamSignedness &RetSignedness,
       StringRef Arg(stringify(Name));
       if (Arg.starts_with("unsigned"))
         return ParamSignedness::Unsigned;
-      else if (Arg.equals("char") || Arg.equals("short") || Arg.equals("int") ||
-               Arg.equals("long"))
+      if (Arg.equals("char") || Arg.equals("short") || Arg.equals("int") ||
+          Arg.equals("long"))
         return ParamSignedness::Signed;
     }
     return ParamSignedness::Unknown;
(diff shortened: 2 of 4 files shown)


lib/SPIRV/SPIRVWriter.cpp
```diff
@@ -5932,6 +5971,32 @@ LLVMToSPIRVBase::transBuiltinToInstWithoutDecoration(Op OC, CallInst *CI,
...
+        (void)Pointee;
+        assert((Pointee == Args[I] || !isa<Function>(Pointee)) &&
+               "Illegal use of a function pointer type");
+      }
+      SPArgs.push_back(SPI->isOperandLiteral(I)
+                           ? cast<ConstantInt>(Args[I])->getZExtValue()
+                           : transValue(Args[I], BB)->getId());
+    }
+    if (Args.size() <= getImageOperandsIndex(OC))
+      if (int SignZeroExt = getImageSignZeroExt(CI->getCalledFunction()))
```
MrSidims, line 5995: We should add a check if 1.4 is supported (see `BM->isAllowedToUseVersion(VersionNumber::SPIRV_1_4)`) and either ignore such image operand coming from SPIR-V friendly LLVM IR or even report and error. Tagging @VyacheslavLevytskyy and @vmaksimo to express an opinion here, which way is better.

MrSidims (commented):
@wenju-he while the patch is doing a good job in determining whether a particular image instruction call is signed or unsigned, it does look odd to me.
The reason is, that [SPIR-V friendly LLVM IR format](https://github.com/KhronosGroup/SPIRV-LLVM-Translator/blob/main/docs/SPIRVRepresentationInLLVM.rst) defines operands for the instructions to be the part of the 'spv friendly' function call, so while translating an instruction with Sign/ZeroExtend mask - we should be already having the appropriate parameter in the function call in LLVM IR, making this demangling redundant (unless I'm missing something).

lib/SPIRV/SPIRVWriter.cpp
VyacheslavLevytskyy: Probably we should ignore rather than report an error, by analogy with https://github.com/KhronosGroup/SPIRV-LLVM-Translator/blob/93f2783317007343da7e8b00be0a363f2171d77f/lib/SPIRV/libSPIRV/SPIRVInstruction.cpp#L161

wenju-he:
> @wenju-he while the patch is doing a good job in determining whether a particular image instruction call is signed or unsigned, it does look odd to me. The reason is, that [SPIR-V friendly LLVM IR format](https://github.com/KhronosGroup/SPIRV-LLVM-Translator/blob/main/docs/SPIRVRepresentationInLLVM.rst) defines operands for the instructions to be the part of the 'spv friendly' function call, so while translating an instruction with Sign/ZeroExtend mask - we should be already having the appropriate parameter in the function call in LLVM IR, making this demangling redundant (unless I'm missing something).

This PR is for translating an image instruction w/o Sign/ZeroExtend mask. The Sign/ZeroExtend mask could be optional in the sense that signedness is already mangled in the function name. For example, functions names in https://github.com/KhronosGroup/SPIRV-LLVM-Translator/pull/2286/commits/27a812a537d90956cf4719412ba2e58ccbf71bb7#diff-52ffa59d3f317359a71e063620bf6052e847aebd24cac407a69a2f2398a2699cR12 are from SYCL FE and they don't have Sign/ZeroExtend mask operand.
The newly added mask in this PR is just a method for preserving the signedness.

lib/SPIRV/SPIRVUtil.cpp
```diff
@@ -881,6 +881,46 @@ bool getParameterTypes(Function *F, SmallVectorImpl<Type *> &ArgTys,
...
+    if (!N)
+      return ParamSignedness::Unknown;
+    if (const auto *Vec = dyn_cast<itanium_demangle::VectorType>(N))
+      N = Vec->getBaseType();
+    if (const auto *Name = dyn_cast<NameType>(N)) {
+      StringRef Arg(stringify(Name));
+      if (Arg.starts_with("unsigned"))
+        return ParamSignedness::Unsigned;
+      if (Arg.equals("char") || Arg.equals("short") || Arg.equals("int") ||
+          Arg.equals("long"))
```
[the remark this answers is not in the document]
MrSidims: Sure

lib/SPIRV/SPIRVWriter.cpp
```diff
@@ -5932,6 +5971,34 @@ LLVMToSPIRVBase::transBuiltinToInstWithoutDecoration(Op OC, CallInst *CI,
...
+    // Image Op needs handling of SignExtend or ZeroExtend image operands.
+    auto Args = getArguments(CI);
+    SPIRVType *SPRetTy =
+        CI->getType()->isVoidTy() ? nullptr : transScavengedType(CI);
+    auto *SPI = SPIRVInstTemplateBase::create(OC);
+    std::vector<SPIRVWord> SPArgs;
+    for (size_t I = 0, E = Args.size(); I != E; ++I) {
+      if (Args[I]->getType()->isPointerTy()) {
+        Value *Pointee = Args[I]->stripPointerCasts();
+        (void)Pointee;
```
MrSidims, line 5986:
```suggestion
        [[maybe_unused]] Value *Pointee = Args[I]->stripPointerCasts();
```

lib/SPIRV/SPIRVWriter.cpp
```diff
@@ -5932,6 +5971,34 @@ LLVMToSPIRVBase::transBuiltinToInstWithoutDecoration(Op OC, CallInst *CI,
...
+               "Illegal use of a function pointer type");
+      }
+      SPArgs.push_back(SPI->isOperandLiteral(I)
+                           ? cast<ConstantInt>(Args[I])->getZExtValue()
+                           : transValue(Args[I], BB)->getId());
+    }
+    if (BM->isAllowedToUseVersion(VersionNumber::SPIRV_1_4)) {
+      if (Args.size() <= getImageOperandsIndex(OC))
+        if (int SignZeroExt = getImageSignZeroExt(CI->getCalledFunction()))
+          SPArgs.push_back(SignZeroExt);
```
MrSidims, line 5997: Basically this line of code has triggered the original question about necessity of the patch.
We can reach this code in 2 cases:
1. As you have described after we have converted LLVM IR with OpenCL flavor to SPIR-V friendly LLVM IR, like this there should be not issues with this code;
2. Just be translating hand-written SPIR-V friendly LLVM IR, like Intel's SYCL device headers do.

If I read the code correctly in this case there will be an issue if we have encoded other operands in the image instruction call - we will add yet another operand, while we should be modifying existing Image Operands parameter by applying the appropriate mask, like we have in the input IR:
`%image = call <4 x i32> @_Z17__spirv_ImageReadIDv4_j14ocl_image3d_roDv4_iET_T0_T1_(target("spirv.Image", void, 2, 0, 0, 0, 0, 0, 0) undef, <4 x i32> zeroinitializer, /* image operand bias*/ i32 0x1, /*literal float value of bias*/ float 2.0f)`
no we want to add zero extend mask, the result is needed to be `...  /* image operand bias + zero extend mask*/ i32 0x1 | 0x2000, /*literal float value of bias*/ float 2.0f) ` , but instead (if I read the code correctly) we would get `...  /* image operand bias + zero extend mask*/ i32 0x1, /*literal float value of bias*/ float 2.0f, i32 0x2000) ` which is incorrect. I might be missing something here, but even if so, may I ask you to add a test proving, that I'm wrong.

wenju-he: > if we have encoded other operands in the image instruction call - we will add yet another operand

New ImageOperands won't be added if there is already an ImageOperands, see the check at https://github.com/KhronosGroup/SPIRV-LLVM-Translator/pull/2286/files/c7826f11bc4730727d144f65845730fadb8a48c6#diff-ba003da77a80df1db76fb378a873a80e758591d72bf82a46fc90069b2a947990R5995

> while we should be modifying existing Image Operands parameter by applying the appropriate mask

Thank you for pointing out this, which was missing in previous commits. I've just added this part. Test is https://github.com/KhronosGroup/SPIRV-LLVM-Translator/pull/2286/files#diff-52ffa59d3f317359a71e063620bf6052e847aebd24cac407a69a2f2398a2699cR31

MrSidims (approved):

MrSidims:
LGTM now

lib/SPIRV/SPIRVWriter.cpp
```diff
@@ -5932,6 +5971,47 @@ LLVMToSPIRVBase::transBuiltinToInstWithoutDecoration(Op OC, CallInst *CI,
...
+                           ? cast<ConstantInt>(Args[I])->getZExtValue()
+                           : transValue(Args[I], BB)->getId());
+    }
+    if (BM->isAllowedToUseVersion(VersionNumber::SPIRV_1_4)) {
+      size_t ImOpIdx = getImageOperandsIndex(OC);
+      if (Args.size() > ImOpIdx) {
+        // Update existing ImageOperands with SignExtendMask/ZeroExtendMask.
+        if (auto *ImOp = dyn_cast<ConstantInt>(Args[ImOpIdx])) {
+          uint64_t ImOpVal = ImOp->getZExtValue();
+          unsigned SignZeroExtMasks =
```
svenvh, line 5999:
```suggestion
          const unsigned SignZeroExtMasks =
```

lib/SPIRV/SPIRVUtil.cpp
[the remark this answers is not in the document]
svenvh: cl_short, cl_long etc are API types, so I don't think we should handle them here.

> Perhaps we can revisit this function if there is use of it for OpenCL builtin in the future?

That sounds reasonable to me.

lib/SPIRV/SPIRVInternal.h
```diff
@@ -892,6 +892,13 @@ inline bool getParameterTypes(CallInst *CI, SmallVectorImpl<Type *> &ArgTys) {
   return getParameterTypes(CI->getCalledFunction(), ArgTys);
 }
 
+enum ParamSignedness { Signed = 0, Unsigned, Unknown };
```
svenvh, line 895:
```suggestion
enum class ParamSignedness { Signed = 0, Unsigned, Unknown };
```

svenvh (commented):
LGTM, just one nit.

svenvh (approved):

PR #2288, opened 2024-01-05
Fix SPIRVRegularizeLLVMBase::regularize fix for shl i1 and lshr i1

bwlodarcz:
The translator failed assertion with V->user_empty() during regularize function when shl i1 or lshr i1 result is used. E.g.

%2 = shl i1 %0 %1
store %2, ptr addrspace(1) @G.1, align 1

Instruction shl i1 is converted to lshr i32 which arithmetic have the same behavior.

### DIFF
--- a/lib/SPIRV/SPIRVRegularizeLLVM.cpp
+++ b/lib/SPIRV/SPIRVRegularizeLLVM.cpp
@@ -322,31 +322,6 @@ void SPIRVRegularizeLLVMBase::expandSYCLTypeUsing(Module *M) {
     expandVIDWithSYCLTypeByValComp(F);
 }
 
-Value *SPIRVRegularizeLLVMBase::extendBitInstBoolArg(Instruction *II) {
-  IRBuilder<> Builder(II);
-  auto *ArgTy = II->getOperand(0)->getType();
-  Type *NewArgType = nullptr;
-  if (ArgTy->isIntegerTy()) {
-    NewArgType = Builder.getInt32Ty();
-  } else if (ArgTy->isVectorTy() &&
-             cast<VectorType>(ArgTy)->getElementType()->isIntegerTy()) {
-    unsigned NumElements = cast<FixedVectorType>(ArgTy)->getNumElements();
-    NewArgType = VectorType::get(Builder.getInt32Ty(), NumElements, false);
-  } else {
-    llvm_unreachable("Unexpected type");
-  }
-  auto *NewBase = Builder.CreateZExt(II->getOperand(0), NewArgType);
-  auto *NewShift = Builder.CreateZExt(II->getOperand(1), NewArgType);
-  switch (II->getOpcode()) {
-  case Instruction::LShr:
-    return Builder.CreateLShr(NewBase, NewShift);
-  case Instruction::Shl:
-    return Builder.CreateShl(NewBase, NewShift);
-  default:
-    return II;
-  }
-}
-
 bool SPIRVRegularizeLLVMBase::runRegularizeLLVM(Module &Module) {
   M = &Module;
   Ctx = &M->getContext();
@@ -393,19 +368,69 @@ bool SPIRVRegularizeLLVMBase::regularize() {
           }
         }
 
-        // Translator treats i1 as boolean, but bit instructions take
-        // a scalar/vector integers, so we have to extend such arguments
-        if (II.isLogicalShift() &&
-            II.getOperand(0)->getType()->isIntOrIntVectorTy(1)) {
-          auto *NewInst = extendBitInstBoolArg(&II);
-          for (auto *U : II.users()) {
-            if (cast<Instruction>(U)->getOpcode() == Instruction::ZExt) {
-              U->dropAllReferences();
-              U->replaceAllUsesWith(NewInst);
-              ToErase.push_back(cast<Instruction>(U));
+        if (II.isLogicalShift()) {
+          // Translator treats i1 as boolean, but bit instructions take
+          // a scalar/vector integers, so we have to extend such arguments.
+          // shl i1 %a %b and lshr i1 %a %b are now converted on:
+          // %0 = select i1 %a, i32 1, i32 0
+          // %1 = select i1 %b, i32 1, i32 0
+          // %2 = lshr i32 %0, %1
+          // if any other instruction other than zext was dependant:
+          // %3 = icmp ne i32 %2, 0
+          // which converts it back to i1 and replace original result with %3
+          // to dependant instructions.
+          if (II.getOperand(0)->getType()->isIntegerTy(1)) {
+            IRBuilder<> Builder(&II);
+            Value *ExtendedBase = Builder.CreateSelect(
+                II.getOperand(0), Builder.getInt(APInt(32, 1)),
+                Builder.getInt(APInt(32, 0)));
+            Value *ExtendedShift = Builder.CreateSelect(
+                II.getOperand(1), Builder.getInt(APInt(32, 1)),
+                Builder.getInt(APInt(32, 0)));
+            Value *ExtendedShiftedVal =
+                Builder.CreateLShr(ExtendedBase, ExtendedShift);
+            Value *CmpNEInst = nullptr;
+            SmallVector<User *, 8> Users(II.users());
+            for (User *U : Users) {
+              if (auto *UI = dyn_cast<Instruction>(U)) {
+                if (UI->getOpcode() == Instruction::ZExt) {
+                  UI->dropAllReferences();
+                  UI->replaceAllUsesWith(ExtendedShiftedVal);
+                  ToErase.push_back(UI);
+                  continue;
+                }
+              }
+              if (!CmpNEInst) {
+                Value *ConstZero = ConstantInt::get(Builder.getInt32Ty(), 0);
+                CmpNEInst = Builder.CreateICmpNE(ExtendedShiftedVal, ConstZero);
+              }
+              U->replaceUsesOfWith(&II, CmpNEInst);
             }
+            ToErase.push_back(&II);
+          } else if (II.getOperand(0)->getType()->isIntOrIntVectorTy(1)) {
+            IRBuilder<> Builder(&II);
+            unsigned NumElements =
+                cast<FixedVectorType>(II.getOperand(0)->getType())
+                    ->getNumElements();
+            Type *ExtendedTy =
+                VectorType::get(Builder.getInt32Ty(), NumElements, false);
+            Value *ExtendedBase =
+                Builder.CreateZExt(II.getOperand(0), ExtendedTy);
+            Value *ExtendedShift =
+                Builder.CreateZExt(II.getOperand(1), ExtendedTy);
+            Value *ExtendedShiftedVal =
+                Builder.CreateLShr(ExtendedBase, ExtendedShift);
+            SmallVector<User *, 16> Users(II.users());
+            for (auto *U : Users) {
+              if (cast<Instruction>(U)->getOpcode() == Instruction::ZExt) {
+                U->dropAllReferences();
+                U->replaceAllUsesWith(ExtendedShiftedVal);
+                ToErase.push_back(cast<Instruction>(U));
+                continue;
+              }
+            }
+            ToErase.push_back(&II);
           }
-          ToErase.push_back(&II);
         }
 
         // Remove optimization info not supported by SPIRV
--- a/lib/SPIRV/SPIRVRegularizeLLVM.cpp
+++ b/lib/SPIRV/SPIRVRegularizeLLVM.cpp
@@ -412,14 +412,19 @@ bool SPIRVRegularizeLLVMBase::regularize() {
             unsigned NumElements =
                 cast<FixedVectorType>(II.getOperand(0)->getType())
                     ->getNumElements();
-            Type *ExtendedTy =
-                VectorType::get(Builder.getInt32Ty(), NumElements, false);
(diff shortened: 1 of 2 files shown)


asudarsa:
@MrSidims, @vmaksimo, @VyacheslavLevytskyy @maksimsab @LU-JOHN
Please take a look.

Thanks

test/lshr_shl_i1_regularize.ll
```diff
@@ -0,0 +1,40 @@
...
+@G.1 = addrspace(1) global i1 true
+
+define spir_func void @test_lshr_i1(i1 %a, i1 %b) {
+entry:
+  %0 = lshr i1 %a, %b
+; CHECK-LLVM: [[AI32_0:%[0-9]+]] = select i1 %a, i32 1, i32 0
+; CHECK-LLVM: [[BI32_0:%[0-9]+]] = select i1 %b, i32 1, i32 0
+; CHECK-LLVM: [[LSHRI32:%[0-9]+]] = lshr i32 [[AI32_0]], [[BI32_0]]
+; CHECK-LLVM: [[TRUNC_0:%[0-9]+]] = icmp ne i32 [[LSHRI32]], 0
+  %1 = zext i1 %0 to i32
```
asudarsa, line 17: What is the purpose of these 'dead' instructions?

lib/SPIRV/SPIRVRegularizeLLVM.cpp
```diff
@@ -400,13 +375,44 @@ bool SPIRVRegularizeLLVMBase::regularize() {
         // a scalar/vector integers, so we have to extend such arguments
         if (II.isLogicalShift() &&
             II.getOperand(0)->getType()->isIntOrIntVectorTy(1)) {
-          auto *NewInst = extendBitInstBoolArg(&II);
-          for (auto *U : II.users()) {
-            if (cast<Instruction>(U)->getOpcode() == Instruction::ZExt) {
-              U->dropAllReferences();
-              U->replaceAllUsesWith(NewInst);
-              ToErase.push_back(cast<Instruction>(U));
+          IRBuilder<> Builder(&II);
```
asudarsa, line 378: Can you please add example LLVM code as comment? Input IR followed by IR generated by your transformation?
I understand the testcase covers it. But this comment can help in documentation. Thanks

lib/SPIRV/SPIRVRegularizeLLVM.cpp
```diff
@@ -400,13 +375,44 @@ bool SPIRVRegularizeLLVMBase::regularize() {
...
+          Type *Op0Ty = II.getOperand(0)->getType();
```
asudarsa, line 379: Nit: Op0Ty can be moved outside 'if' block and can be used inside the if conditional.

lib/SPIRV/SPIRVRegularizeLLVM.cpp
```diff
@@ -400,13 +375,44 @@ bool SPIRVRegularizeLLVMBase::regularize() {
...
+              Builder.CreateZExt(II.getOperand(1), ExtendedTy);
+          Value *ExtendedShiftedVal = nullptr;
+          if (II.getOpcode() == Instruction::LShr) {
+            ExtendedShiftedVal =
+                Builder.CreateLShr(ExtendedBase, ExtendedShift);
+          } else {
+            ExtendedShiftedVal = Builder.CreateXor(ExtendedBase, ExtendedShift);
+          }
+          Value *CmpNEInst = nullptr;
+          SmallVector<User *, 8> Users(II.users());
```
asudarsa, line 401: Nit: Why do we need a tmp var here? (Users)?

asudarsa (commented):
LGTM overall. A few minor comments to address. Thanks

asudarsa:
mac build issue seems unrelated.

test/lshr_shl_i1_regularize.ll
MrSidims: I believe it's a result of the work of llvm fuzzer

bwlodarcz: @asudarsa This is a check if we are preserving previous behavior. If you look on removed code you will see that `zext` instructions extending result of shift were removed because their presence was redundant.

lib/SPIRV/SPIRVRegularizeLLVM.cpp
bwlodarcz: I consider this whole branch as cold - even freezing so these instructions won't be executed often. 
To the line: I'm not sure how much overhead it introduces (maybe nothing). At first glance you have at least one touch of memory on heap which is unnecessary if the branch isn't taken.

lib/SPIRV/SPIRVRegularizeLLVM.cpp
bwlodarcz: Finding and understanding this was a result of few hours debugging. It's because everything in LLVM use api isn't what it seem.
Shortly use is a bidirectional linked list which nodes points to corresponding `llvm::Value` (because these are not always `llvm::Instruction`). `users()` call give you range of `llvm::User` - directly traverses this linked list to get these pointers (without going through `llvm::Use`). And now there is main surprise. These can be removed during traversal if you do something like e.g `replaceUsesOfWith` - which means that this range can be cut short after for example one removal. This is why you need first traverse the range, store all Value *, and only then replace. Not something that you can expect when you see range of iterators but this is how it works.

test/lshr_shl_i1_regularize.ll
```diff
@@ -0,0 +1,40 @@
...
+  ret void
+}
+
+define spir_func void @test_shl_i1(i1 %a, i1 %b) {
+entry:
+  %0 = shl i1 %a, %b
+; CHECK-LLVM: [[AI32_1:%[0-9]+]] = select i1 %a, i32 1, i32 0
+; CHECK-LLVM: [[BI32_1:%[0-9]+]] = select i1 %b, i32 1, i32 0
+; CHECK-LLVM: [[XORI32:%[0-9]+]] = xor i32 [[AI32_1]], [[BI32_1]]
+; CHECK-LLVM: [[TRUNC_1:%[0-9]+]] = icmp ne i32 [[XORI32]], 0
```
LU-JOHN, line 32: Is shl i1 equivalent to xor?  If %a==0, shifting should produce a 0, but consider when %a==0 and %b==1:

%xor = xor i32 0, 1   ; produces 1
icmp ne i32 %xor, 0  ; produces 1

Thus the new code sequence produces a 1 when it should produce a 0.

bwlodarcz: Left shift of bit:
```
0 << 0 = 0
1 << 0 = 1
0 << 1 = 1
1 << 1 = 0
```
which is the same for xor operation.

Logical right shift of bit:
```
0 >> 0 = 0
1 >> 0 = 1
0 >> 1 = 0
1 >> 1 = 0
```
I don't think that I made a mistake but maybe I'm missing something.

LU-JOHN: > 0 << 1 = 1

I don't think this is correct.  Shifting 0 by anything produces a 0.

LU-JOHN: The correct results are the same for both:

0 >> 0 = 0
1 >> 0 = 1
0 >> 1 = 0
1 >> 1 = 0


0 << 0 = 0
1 << 0 = 1
0 << 1 = 0
1 << 1 = 0

bwlodarcz: Nice catch. Thank you very much.

test/lshr_shl_i1_regularize.ll
```diff
@@ -0,0 +1,40 @@
+; RUN: llvm-as %s -o %t.bc
```
MrSidims, line 1: Vector test is tested elsewhere already, right?

lib/SPIRV/SPIRVRegularizeLLVM.cpp
```diff
@@ -393,19 +368,69 @@ bool SPIRVRegularizeLLVMBase::regularize() {
...
+            unsigned NumElements =
+                cast<FixedVectorType>(II.getOperand(0)->getType())
+                    ->getNumElements();
+            Type *ExtendedTy =
+                VectorType::get(Builder.getInt32Ty(), NumElements, false);
+            Value *ExtendedBase =
+                Builder.CreateZExt(II.getOperand(0), ExtendedTy);
+            Value *ExtendedShift =
+                Builder.CreateZExt(II.getOperand(1), ExtendedTy);
+            Value *ExtendedShiftedVal =
```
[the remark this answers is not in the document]
MrSidims: Discussed offline, it's a know problem to be addressed later. Current patch enhances the behavior for scalars and doesn't change it for vectors.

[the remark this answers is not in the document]
bwlodarcz: Addressed.

bwlodarcz:
CI issues not related to PR.

test/lshr_shl_i1_regularize.ll
```diff
@@ -0,0 +1,57 @@
...
+; RUN: llvm-spirv -s %t.bc -o %t.reg.bc
```
LU-JOHN, line 2: I think this testcase is identical for lshr and shl.  If so, consider using the pre-processor to refactor the test:

; RUN: %clang %s -DOPCODE=shl -E > %t.shl.ll
; RUN: llvm-as %t.shl.ll -o %t.bc
; [rest of run commands]

; RUN: %clang %s -DOPCODE=lshr -E > %t.lshr.ll
; RUN: llvm-as %t.lshr.ll -o %t.bc
; [rest of run commands]

MrSidims:
Restarting CI

lib/SPIRV/SPIRVRegularizeLLVM.cpp
```diff
@@ -393,19 +368,78 @@ bool SPIRVRegularizeLLVMBase::regularize() {
...
+                U->replaceAllUsesWith(ExtendedShiftedVal);
+                ToErase.push_back(cast<Instruction>(U));
+                continue;
+              }
+              if (!CmpNEInst) {
+                CmpNEInst = Builder.CreateICmpNE(ExtendedShiftedVal, VecZeroes);
+              }
+              U->replaceUsesOfWith(&II, CmpNEInst);
             }
+            ToErase.push_back(&II);
```
MrSidims, line 441: ideally it should be a separate function, but I won't insist
```suggestion
          if (II.getOperand(0)->getType()->isIntOrIntVectorTy(1)) {
            IRBuilder<> Builder(&II);
            Value *CmpNEInst = nullptr;
            Constant *ConstZero = ConstantInt::get(Builder.getInt32Ty(), 0);
            Constant *ConstOne = ConstantInt::get(Builder.getInt32Ty(), 1);
            if (auto *VecTy =
                  dyn_cast<FixedVectorType>(II.getOperand(0)->getType())) {
              const unsigned NumElements = VecTy->getNumElements();
              ConstZero =
                  ConstantVector::getSplat(ElementCount::getFixed(NumElements), ConstZero);
              ConstOne =
                  ConstantVector::getSplat(ElementCount::getFixed(NumElements), ConstOne);
            }
            Value *ExtendedBase =
                Builder.CreateSelect(II.getOperand(0), ConstOne, ConstZero);
            Value *ExtendedShift =
                Builder.CreateSelect(II.getOperand(1), ConstOne, ConstZero);
            Value *ExtendedShiftedVal =
                Builder.CreateLShr(ExtendedBase, ExtendedShift);
            SmallVector<User *, 8> Users(II.users());
            for (User *U : Users) {
              if (auto *UI = dyn_cast<Instruction>(U)) {
                if (UI->getOpcode() == Instruction::ZExt) {
                  UI->dropAllReferences();
                  UI->replaceAllUsesWith(ExtendedShiftedVal);
                  ToErase.push_back(UI);
                  continue;
                }
              }
              if (!CmpNEInst) {
                CmpNEInst = Builder.CreateICmpNE(ExtendedShiftedVal, ConstZero);
              }
              U->replaceUsesOfWith(&II, CmpNEInst);
            }
            ToErase.push_back(&II);
          }
```

test/lshr_shl_i1_regularize.ll
MrSidims: IMHO as is it's more readable

MrSidims (approved):
if no objections from @LU-JOHN  I'd merge this

test/lshr_shl_i1_regularize.ll
svenvh: Agree with @MrSidims that it's better to keep the tests written as is.

test/lshr_shl_i1_regularize.ll
```diff
@@ -0,0 +1,57 @@
...
+; CHECK-LLVM: [[LSHR32_2:%[0-9]+]] = lshr <8 x i32> [[AI32_2]], [[BI32_2]]
+; CHECK-LLVM: [[TRUNC_2:%[0-9]+]] = icmp ne <8 x i32> [[LSHR32_2]], zeroinitializer
+  %1 = zext <8 x i1> %0 to <8 x i32>
+  %2 = zext <8 x i1> %0 to <8 x i32>
+; CHECK-LLVM-NOT: zext
+; CHECK-LLVM-NOT: select
+  store <8 x i1> %0, ptr addrspace(1) @G.2, align 1
+; CHECK-LLVM: store <8 x i1> [[TRUNC_2]], ptr addrspace(1) @G.2, align 1
+  ret void
+}
```
svenvh, line 57: Missing an EOL here.

svenvh (approved):

LU-JOHN (approved):
LGTM

PR #2289, opened 2024-01-05
[NFC] [DebugInfo] Test Value field in SPIRV TypeMember

LU-JOHN:
Test Value field in SPIRV TypeMember.

Also remove some unreachable code.

### DIFF
--- a/test/DebugInfo/DebugTypeMember.ll
+++ b/test/DebugInfo/DebugTypeMember.ll
@@ -1,5 +1,6 @@
 ; Exercise Value field in SPIRV TypeMember.
-; It is used to preserve DIDerivedType's extraData argument in LLVM IR.
+; It is used to preserve DIDerivedType's extraData argument in
+; LLVM IR when a DIFlagStaticMember is present.
 
 ; Original *.cpp source
 ; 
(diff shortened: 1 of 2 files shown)


test/DebugInfo/DebugTypeMember.ll
```diff
@@ -0,0 +1,63 @@
+; Exercise Value field in SPIRV TypeMember.
```
MrSidims, line 1: is this test in principle different from https://github.com/KhronosGroup/SPIRV-LLVM-Translator/blob/main/test/DebugInfo/NonSemantic/static_member_array.ll ?

MrSidims (commented):
Removing code LGTM, not sure about the test

test/DebugInfo/DebugTypeMember.ll
LU-JOHN: The unique part of this test is that it has an DIFlagStaticMember and extraData field.  This exercises code in LLVMToSPIRVDbgTran.cpp lines 1016 and 1045.

test/DebugInfo/DebugTypeMember.ll
```diff
@@ -0,0 +1,64 @@
...
+; }
+
+; RUN: llvm-as %s -o %t.bc
+
+; RUN: llvm-spirv -o %t.spv %t.bc
+; RUN: llvm-spirv -r -o %t.rev.bc %t.spv
+; RUN: llvm-dis %t.rev.bc -o %t.rev.ll
+; RUN: FileCheck %s --input-file %t.rev.ll
+
+; RUN: llvm-spirv -o %t.ns.spv %t.bc --spirv-debug-info-version=nonsemantic-shader-200
```
MrSidims, line 23: we should probably extend testing with --spirv-debug-info-version=nonsemantic-shader-100
and by adding few SPIR-V checks, other then that - LGTM

MrSidims (approved):

MrSidims (approved):

PR #2208, opened 2023-11-08
[SPV-IR to OCL] Fix mutated builtin call attribute copying

wenju-he:
When SPIRVToOCL20Pass translates following SPV IR
```
  %call = tail call spir_func noundef ptr @_Z42__spirv_GenericCastToPtrExplicit_ToPrivatePvi(ptr addrspace(4) noundef %Ptr, i32 noundef 0)
declare spir_func ptr @_Z42__spirv_GenericCastToPtrExplicit_ToPrivatePvi(ptr addrspace(4), i32)
```
to OCL IR:
```
  %call = tail call spir_func noundef ptr @__to_private(ptr addrspace(4) noundef %Ptr)
declare spir_func ptr @__to_private(ptr addrspace(4))
```
, there is error `Attribute after last parameter` at copying attribute
to the new call because the old call has an additional AttributeSet for
the last argument. The last argument is eliminated in the new call.

Set tail call for new call if old call is tail call.

Revert cafd7e0. Partially revert c60327f.

PR #2290, opened 2024-01-08
Add pass plugin for in-tree build with shared libs

wenju-he:
The plugin registers llvm-spirv passes to llvm pass manager and enables
LIT test for llvm-spirv passes using opt tool.


wenju-he:
For instance, this PR allows adding LIT test for SPIRVToOCL20Pass in https://github.com/KhronosGroup/SPIRV-LLVM-Translator/pull/2208

PR #2292, opened 2024-01-08
Use triple subarch to emit exact SPIR-V version

### LINKED ISSUE #1509: Use exact SPIR-V version from the target triple

linehill:
Use target triple's subarch component, if present, for setting exact SPIR-V version for the SPIR-V emission.

Resolves #1509.


MrSidims (approved):
LGTM, thanks for the contribution!

svenvh (approved):

PR #2256, opened 2023-12-07
Fix reverse translation of non-constant values of OpCompositeConstruct

vmaksimo:
This patch introduces a way to use runtime values for structure fields.

PR #2296, opened 2024-01-08
Fix reverse translation of non-constant values of OpCompositeConstruct pt.2

vmaksimo:
This patch introduces a way to use runtime values for array and vector types. It continues #2256

### DIFF
--- a/test/composite_construct_vector_non_constant.spvasm
+++ b/test/composite_construct_vector_non_constant.spvasm
@@ -1,0 +1,40 @@
+; REQUIRES: spirv-as
+; RUN: spirv-as --target-env spv1.0 -o %t.spv %s
+; RUN: spirv-val %t.spv
+; RUN: llvm-spirv -r %t.spv -o %t.rev.bc
+; RUN: llvm-dis %t.rev.bc
+; RUN: FileCheck < %t.rev.ll %s
+
+; CHECK: define spir_func <3 x i32> @non_constant_vector_elements(i32 %[[#ARG0:]], i32 %[[#ARG1:]])
+; CHECK: %[[#VEC0:]] = insertelement <3 x i32> poison, i32 %[[#ARG1]], i32 0
+; CHECK: %[[#VEC1:]] = insertelement <3 x i32> %[[#VEC0]], i32 %[[#ARG0]], i32 1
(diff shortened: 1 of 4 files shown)


vmaksimo:

test/composite_construct_vector_non_constant.spvasm
```diff
@@ -0,0 +1,40 @@
...
+; RUN: spirv-as --target-env spv1.0 -o %t.spv %s
+; RUN: spirv-val %t.spv
+; RUN: llvm-spirv -r %t.spv -o %t.rev.bc
+; RUN: llvm-dis %t.rev.bc
+; RUN: FileCheck < %t.rev.ll %s
+
+; CHECK: define spir_func <3 x i32> @non_constant_vector_elements(i32 %[[#ARG0:]], i32 %[[#ARG1:]])
+; CHECK: %[[#VEC0:]] = insertelement <3 x i32> poison, i32 %[[#ARG1]], i32 0
+; CHECK: %[[#VEC1:]] = insertelement <3 x i32> %[[#VEC0]], i32 %[[#ARG0]], i32 1
+; CHECK: %[[#VEC2:]] = insertelement <3 x i32> %[[#VEC1]], i32 1, i32 2
```
MrSidims, line 11: there is a room for a peephole optimization:
```suggestion
`; CHECK: %[[#VEC0:]] = insertelement <3 x i32> < poison, poison, 1> i32 %[[#ARG1]], i32 0
; CHECK: %[[#VEC1:]] = insertelement <3 x i32> %[[#VEC0]], i32 %[[#ARG0]], i32 1`
```
would you mind to take a look if it's implementable?

MrSidims: Or actually, may be not needed: https://godbolt.org/z/v5o8sjanP

MrSidims (approved):
LGTM
I withdraw my comment about peephole optimization as llc generates same assembly for both cases.
There are other ways to implement both vector and array translation:
for vector one can also do shuffle, and while it's usually more efficient for constant vectors - it's not the case for inserting runtime-known values.
For structures one can also implement this functionally via `insertvalue`, but in case of bit structures it becomes less efficient as it would require to load the entire structure first.

VyacheslavLevytskyy:
LGTM

PR #2298, opened 2024-01-09
add API call to display general information about the module

VyacheslavLevytskyy:
Partially load SPIR-V from the stream and decode only selected for the report instructions, needed to retrieve general information about the module: capabilities, extensions, version, memory model and addressing model.

In addition to immediately helpful for back-ends lists of capabilities and extensions declared in SPIR-V module, a general intent also is to extend report details in future by feedbacks about further potentially useful analysis, statistics, etc.

### DIFF
--- a/include/LLVMSPIRVLib.h
+++ b/include/LLVMSPIRVLib.h
@@ -106,6 +106,21 @@ std::unique_ptr<SPIRVModule> readSpirvModule(std::istream &IS,
                                              const SPIRV::TranslatorOpts &Opts,
                                              std::string &ErrMsg);
 
+/// \brief Partially load SPIR-V from the stream and decode only selected
+/// for the report instructions, needed to retrieve general information
+/// about the module.
+/// \returns nullopt on failure.
+struct SPIRVModuleReport {
+  SPIRV::VersionNumber Version;
+  uint32_t MemoryModel;
+  uint32_t AddrModel;
+  std::vector<std::string> Extensions;
+  std::vector<std::string> ExtensionImports;
+  std::vector<uint32_t> Capabilities;
+};
+std::optional<SPIRVModuleReport> getSpirvReport(std::istream &IS);
+std::optional<SPIRVModuleReport> getSpirvReport(std::istream &IS, int &ErrCode);
+
 } // End namespace SPIRV
 
 namespace llvm {
--- a/include/LLVMSPIRVOpts.h
+++ b/include/LLVMSPIRVOpts.h
@@ -68,6 +68,11 @@ enum class VersionNumber : uint32_t {
   MaximumVersion = SPIRV_1_4
 };
 
+inline bool isSPIRVVersionKnown(uint32_t Ver) {
+  return Ver >= static_cast<uint32_t>(VersionNumber::MinimumVersion) &&
+         Ver <= static_cast<uint32_t>(VersionNumber::MaximumVersion);
+}
+
 enum class ExtensionID : uint32_t {
   First,
 #define EXT(X) X,
--- a/lib/SPIRV/SPIRVReader.cpp
+++ b/lib/SPIRV/SPIRVReader.cpp
@@ -4825,6 +4825,80 @@ Instruction *SPIRVToLLVM::transRelational(SPIRVInstruction *I, BasicBlock *BB) {
   return cast<Instruction>(Mutator.getMutated());
 }
 
+std::optional<SPIRVModuleReport> getSpirvReport(std::istream &IS) {
+  int IgnoreErrCode;
+  return getSpirvReport(IS, IgnoreErrCode);
+}
+
+std::optional<SPIRVModuleReport> getSpirvReport(std::istream &IS,
+                                                int &ErrCode) {
+  SPIRVWord Word;
+  std::string Name;
+  std::unique_ptr<SPIRVModule> BM(SPIRVModule::createSPIRVModule());
+  SPIRVDecoder D(IS, *BM);
+  D >> Word;
+  if (Word != MagicNumber) {
+    ErrCode = SPIRVEC_InvalidMagicNumber;
+    return {};
+  }
+  D >> Word;
+  if (!isSPIRVVersionKnown(Word)) {
+    ErrCode = SPIRVEC_InvalidVersionNumber;
+    return {};
+  }
+  SPIRVModuleReport Report = {.Version =
+                                  static_cast<SPIRV::VersionNumber>(Word),
+                              .MemoryModel = MemoryModelMax,
+                              .AddrModel = AddressingModelMax};
+  // Skip: Generator’s magic number, Bound and Reserved word
+  D.ignore(3);
+
+  for (bool IsLastInstr = false;
+       !IS.bad() && !IsLastInstr && D.getWordCountAndOpCode();) {
+    switch (D.OpCode) {
+    case OpCapability:
+      D >> Word;
+      Report.Capabilities.push_back(Word);
+      break;
+    case OpExtension:
+      D >> Name;
+      Report.Extensions.push_back(Name);
+      break;
+    case OpExtInstImport:
+      D >> Word >> Name;
+      Report.ExtensionImports.push_back(Name);
+      break;
+    case OpMemoryModel:
+      SPIRVAddressingModelKind AddrModel;
+      SPIRVMemoryModelKind MemoryModel;
+      D >> AddrModel >> MemoryModel;
+      if (!isValid(AddrModel)) {
+        ErrCode = SPIRVEC_InvalidAddressingModel;
+        return {};
+      } else if (!isValid(MemoryModel)) {
+        ErrCode = SPIRVEC_InvalidMemoryModel;
+        return {};
+      } else {
+        Report.MemoryModel = MemoryModel;
+        Report.AddrModel = AddrModel;
+        IsLastInstr = true;
+      }
+      break;
+    default:
+      IsLastInstr = true;
+    }
+  }
+  if (IS.bad()) {
+    ErrCode = SPIRVEC_InvalidModule;
+    return {};
+  } else if (Report.MemoryModel == MemoryModelMax) {
+    ErrCode = SPIRVEC_UnspecifiedMemoryModel;
+    return {};
+  }
+  ErrCode = SPIRVEC_Success;
+  return std::make_optional(std::move(Report));
+}
+
 std::unique_ptr<SPIRVModule> readSpirvModule(std::istream &IS,
                                              const SPIRV::TranslatorOpts &Opts,
                                              std::string &ErrMsg) {
--- a/lib/SPIRV/libSPIRV/SPIRVErrorEnum.h
+++ b/lib/SPIRV/libSPIRV/SPIRVErrorEnum.h
@@ -22,3 +22,8 @@ _SPIRV_OP(InvalidWordCount,
 _SPIRV_OP(Requires1_1, "Feature requires SPIR-V 1.1 or greater:")
 _SPIRV_OP(RequiresExtension,
           "Feature requires the following SPIR-V extension:\n")
+_SPIRV_OP(InvalidMagicNumber,
+          "Invalid Magic Number.")
+_SPIRV_OP(InvalidVersionNumber,
+          "Invalid Version Number.")
+_SPIRV_OP(UnspecifiedMemoryModel, "Unspecified Memory Model")
--- a/tools/llvm-spirv/llvm-spirv.cpp
+++ b/tools/llvm-spirv/llvm-spirv.cpp
@@ -196,6 +196,12 @@ static cl::opt<bool> SpecConstInfo(
     cl::desc("Display id of constants available for specializaion and their "
              "size in bytes"));
 
+static cl::opt<bool>
+    SPIRVPrintReport("spirv-print-report", cl::init(false),
+                     cl::desc("Display general information about the module "
+                              "(capabilities, extensions, version, memory model"
+                              " and addresing model)"));
+
 static cl::opt<SPIRV::FPContractMode> FPCMode(
     "spirv-fp-contract", cl::desc("Set FP Contraction mode:"),
     cl::init(SPIRV::FPContractMode::On),
@@ -797,7 +803,7 @@ int main(int Ac, char **Av) {
     return convertSPIRV();
 #endif
 
-  if (!IsReverse && !IsRegularization && !SpecConstInfo)
+  if (!IsReverse && !IsRegularization && !SpecConstInfo && !SPIRVPrintReport)
     return convertLLVMToSPIRV(Opts);
 
   if (IsReverse && IsRegularization) {
@@ -823,6 +829,35 @@ int main(int Ac, char **Av) {
       std::cout << "Spec const id = " << SpecConst.ID
                 << ", size in bytes = " << SpecConst.Size
                 << ", type = " << SpecConst.Type << "\n";
+  }
+
+  if (SPIRVPrintReport) {
+    std::ifstream IFS(InputFile, std::ios::binary);
+    int ErrCode = 0;
+    std::optional<SPIRV::SPIRVModuleReport> Report = SPIRV::getSpirvReport(IFS, ErrCode);
+    if (!Report) {
+      std::cout << "Invalid SPIR-V binary, error code is " << ErrCode << "\n";
+      return -1;
+    }
+
+    std::cout << "SPIR-V module report:"
+              << "\n Version: " << static_cast<uint32_t>(Report->Version)
+              << "\n Memory model: " << Report->MemoryModel
+              << "\n Addressing model: " << Report->AddrModel << "\n";
+
+    std::cout << " Number of capabilities: " << Report->Capabilities.size()
+              << "\n";
+    for (auto Capability : Report->Capabilities)
+      std::cout << "  Capability: " << Capability << "\n";
+
+    std::cout << " Number of extensions: " << Report->Extensions.size() << "\n";
+    for (auto Extension : Report->Extensions)
+      std::cout << "  Extension: " << Extension << "\n";
+
+    std::cout << " Number of extension imports: "
+              << Report->ExtensionImports.size() << "\n";
+    for (auto ExtensionImport : Report->ExtensionImports)
+      std::cout << "  Extension Import: " << ExtensionImport << "\n";
   }
   return 0;
 }
--- a/lib/SPIRV/libSPIRV/SPIRVModule.cpp
+++ b/lib/SPIRV/libSPIRV/SPIRVModule.cpp
@@ -2138,9 +2138,7 @@ std::istream &operator>>(std::istream &I, SPIRVModule &M) {
   }
 
   Decoder >> MI.SPIRVVersion;
-  bool SPIRVVersionIsKnown =
-      static_cast<uint32_t>(VersionNumber::MinimumVersion) <= MI.SPIRVVersion &&
-      MI.SPIRVVersion <= static_cast<uint32_t>(VersionNumber::MaximumVersion);
+  bool SPIRVVersionIsKnown = isSPIRVVersionKnown(MI.SPIRVVersion);
   if (!M.getErrorLog().checkError(
           SPIRVVersionIsKnown, SPIRVEC_InvalidModule,
           "unsupported SPIR-V version number '" + to_string(MI.SPIRVVersion) +


include/LLVMSPIRVLib.h
```diff
@@ -106,6 +106,21 @@ std::unique_ptr<SPIRVModule> readSpirvModule(std::istream &IS,
                                              const SPIRV::TranslatorOpts &Opts,
                                              std::string &ErrMsg);
 
+/// \brief Partially load SPIR-V from the stream and decode only selected
+/// for the report instructions, needed to retrieve general information
+/// about the module.
+/// \returns nullopt on failure.
```
AlexeySachkov, line 112: Should this be moved below to be next to a function instead of a struct? Should we note that if this API returned failure, then `readSPIRVModule` is expected to fail as well?

tools/llvm-spirv/llvm-spirv.cpp
```diff
@@ -824,5 +830,34 @@ int main(int Ac, char **Av) {
                 << ", size in bytes = " << SpecConst.Size
                 << ", type = " << SpecConst.Type << "\n";
   }
+
+  if (SPIRVPrintReport) {
```
AlexeySachkov, line 834: For the human-readable form of the report, we should be more user-friendly, I think. I mean that all binary fields should be decoded (version, capabilities, etc.)

include/LLVMSPIRVLib.h
VyacheslavLevytskyy: Agree, thanks

tools/llvm-spirv/llvm-spirv.cpp
VyacheslavLevytskyy: Probably the human-readable form of the report deserves even a separate (additional) API call that takes SPIRV::SPIRVModuleReport as an argument and returns all binary fields decoded.

include/LLVMSPIRVLib.h
```diff
@@ -106,6 +106,35 @@ std::unique_ptr<SPIRVModule> readSpirvModule(std::istream &IS,
...
+std::optional<SPIRVModuleReport> getSpirvReport(std::istream &IS);
+std::optional<SPIRVModuleReport> getSpirvReport(std::istream &IS, int &ErrCode);
+
+struct SPIRVModuleTextReport {
+  std::string Version;
+  std::string MemoryModel;
+  std::string AddrModel;
+  std::vector<std::string> Extensions;
+  std::vector<std::string> ExtensionImports;
+  std::vector<std::string> Capabilities;
```
vmustya, line 131: I actually have some questions here:
1. What is the difference between `Extensions` and `ExtensionImports`?
2. The translator defines [extension enum](https://github.com/KhronosGroup/SPIRV-LLVM-Translator/blob/main/include/LLVMSPIRVExtensions.inc). I believe it would be better to get a vector of enum values instead of strings. The same is related to another fields.

include/LLVMSPIRVLib.h
```diff
@@ -106,6 +106,35 @@ std::unique_ptr<SPIRVModule> readSpirvModule(std::istream &IS,
                                              const SPIRV::TranslatorOpts &Opts,
                                              std::string &ErrMsg);
 
+struct SPIRVModuleReport {
+  SPIRV::VersionNumber Version;
+  uint32_t MemoryModel;
+  uint32_t AddrModel;
+  std::vector<std::string> Extensions;
+  std::vector<std::string> ExtensionImports;
```
vmustya, line 114: I actually have some questions here:
1. What is the difference between `Extensions` and `ExtensionImports`?
2. The translator defines [extension enum](https://github.com/KhronosGroup/SPIRV-LLVM-Translator/blob/main/include/LLVMSPIRVExtensions.inc). I believe it would be better to get a vector of enum values instead of strings.

VyacheslavLevytskyy: 1. These are OpExtension and OpExtInst.
2. Thanks, it may be that you are right, I would need a feedback here actually from future customers of the feature. @AlexeySachkov Can you please suggest, maybe numbers indeed make more sense in this binary version of the report that is SPIRVModuleReport?

vmustya: > 1. These are OpExtension and OpExtInst.

Then it probably should be `ExtendedInstructionSets` instead of `ExtensionImports`. Current names look confusing to me.

> 2. Thanks, it may be that you are right, I would need a feedback here actually from future customers of the feature. @AlexeySachkov Can you please suggest, maybe numbers indeed make more sense in this binary version of the report that is SPIRVModuleReport?

As a future customer of this feature, I would say, that number/enum comparison is better for my code, than string comparison.

VyacheslavLevytskyy: Thank you @vmustya

AlexeySachkov: I think that `ExtensionImports` is not properly named indeed. It corresponds to [`OpExtInstImport`](https://registry.khronos.org/SPIR-V/specs/unified1/SPIRV.html#OpExtInstImport) instruction, which imports not an extension, but _extended instruction set_ which is a different mechanism.

> Thanks, it may be that you are right, I would need a feedback here actually from future customers of the feature. @AlexeySachkov Can you please suggest, maybe numbers indeed make more sense in this binary version of the report that is SPIRVModuleReport?

My vision for this new API is that it would be invoked by a SPIR-V consumer before calling `readSpirvModule` as a lightweight and quick operation to parse SPIR-V module "header" to understand if it makes sense (i.e. if it is possible at all) to read and parse the rest of the SPIR-V module.

From that point of view, returning integers/enumerators is better, because it allows to write more performant code. Human-readable names of those extensions are only needed in a negative case when SPIR-V consumer found an unsupported extension simply to prepare a nice error message for a user. The second case is when the tool itself provides a human-readable output in that new `--spirv-print-report` mode, but it should be easy enough to convert enum to a string for the translator.

VyacheslavLevytskyy: 1. I changed names: ExtensionImports into ExtendedInstructionSets, thanks for the suggestion.

2. @vmustya Talking abouts extensions -- I've checked how extensions enum is created and used, and it is not possible to use integer identifier instead of strings for extensions. We are forced to use string names.

Enum that you are referring to does not guarantee that the same extension has the same identifier in two different builds of the library. It is a performance improvement and a way to refer to extensions internally (e.g., for recording users options choice during one run), but not in any case an attempt to assign a unique identifier to an extension: 

https://github.com/KhronosGroup/SPIRV-LLVM-Translator/blob/9edf71445c624ab4cc6c2213cef917d2e4906e54/include/LLVMSPIRVOpts.h#L71
 
According to spec, extensions are identified by a name, and we are forced to refer to extension using a string name. This is a difference from capabilities, where we should use and will use integers.

VyacheslavLevytskyy:
@MrSidims @svenvh May I ask you for a feedback when you have time please?

include/LLVMSPIRVLib.h
vmustya: @VyacheslavLevytskyy, we still can use the enum values. We just shouldn't serialize/save them into the files which can be re-used with a different translator build. For example, we use the enum values to configure SPIR-V writer: https://github.com/intel/cm-compiler/blob/4ba19c8d93f3aa7099ccff2d77038b8a5bd08658/clang/lib/CodeGen/BackendUtil.cpp#L965

vmustya: @VyacheslavLevytskyy, from the other side, we may have a SPIR-V module with a new extension which is not implemented by the translator yet. Obviously we have no enum entries for such extensions. In this case we still should be able to get the list of the extensions from the module. So, the strings might be the only solution.

lib/SPIRV/SPIRVReader.cpp
```diff
@@ -4825,6 +4825,137 @@ Instruction *SPIRVToLLVM::transRelational(SPIRVInstruction *I, BasicBlock *BB) {
   return cast<Instruction>(Mutator.getMutated());
 }
 
+std::optional<SPIRVModuleReport> getSpirvReport(std::istream &IS) {
+  int IgnoreErrCode;
+  return getSpirvReport(IS, IgnoreErrCode);
+}
+
+std::optional<SPIRVModuleReport> getSpirvReport(std::istream &IS,
```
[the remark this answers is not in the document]
VyacheslavLevytskyy: Thank you for the suggestion. However, code in SPIRVModule.cpp:2125 is too closely interlinked with other logics undesirable in this particular API code, for example, with error reporting resulting in eventual abort() call. I started this implementation from analyzing our re-use option and avoiding any logic duplication. Unfortunately, existing code was not created to be re-used. It's not possible to reuse existing logic without big refactoring work on rewriting it.

include/LLVMSPIRVOpts.h
```diff
@@ -68,6 +68,27 @@ enum class VersionNumber : uint32_t {
...
+    return "1.2";
+  case static_cast<uint32_t>(VersionNumber::SPIRV_1_3):
+    return "1.3";
+  case static_cast<uint32_t>(VersionNumber::SPIRV_1_4):
+    return "1.4";
+  }
+  return "unknown";
+}
+
+inline bool isSPIRVVersionKnown(uint32_t Ver) {
```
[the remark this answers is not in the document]
VyacheslavLevytskyy: Sorry, I don't understand your question. This is exactly what I did. I extract this logics from SPIRVModule.cpp, created a separated function and re-used it in the new code to avoid Ctrl-C Ctrl-V approach.

include/LLVMSPIRVOpts.h
```diff
@@ -68,6 +68,27 @@ enum class VersionNumber : uint32_t {
   MaximumVersion = SPIRV_1_4
 };
 
+inline constexpr std::string_view formatVersionNumber(uint32_t Version) {
```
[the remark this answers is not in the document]
VyacheslavLevytskyy: Please see the answer for the previous question. This is exactly what was done. I was extracting and re-using where it's possible without causing a bigger refactoring.

VyacheslavLevytskyy:
@bwlodarcz Thank you for the code review. Summarizing your concerns, I may respond that this feature addition was carefully checked against existing code to avoid simplistic Copy/Paste of existing higher level scheme of translation that is of course by its own nature is quite similar to the presented header translation into a brief stats report.

include/LLVMSPIRVOpts.h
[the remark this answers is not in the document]
bwlodarcz: Missed that. Thx.

include/LLVMSPIRVOpts.h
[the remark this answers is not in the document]
bwlodarcz: Missed that. Thx

bwlodarcz:
@VyacheslavLevytskyy Thank you. That's exactly the question here - isn't it possible to integrate these two functionalities? I mean parsing header of spirv and checking caps and extensions looks more less like this in code:
```
char * mmaped_file = (char *) mmap(...);
char * mmaped_file_end = mmaped_file + mmaped_size;
struct [[gnu::packed]] Header {
  uint32_t magic_num;
  uint32_t translator;
  uint32_t bound;
  uint32_t reserved;
};
Header H{};
memcpy(&H, mmaped_file, sizeof(Header));
validate_header(&H);
std::vector<Capabilities> caps;
while (mmaped_file < mmaped_file_end && *mmaped_file == 0x11) {
   caps.emplace_back((Capability) (mmaped_file + 32));
   mmaped_file += 64;    
}
std::vector<std::string> exten;
while (mmaped_file < mmaped_file_end && *mmaped_file = 0x0a) {
   std::string &ref = exten.emplace_back(mmaped_file + 32);
   ... here literal alignment logic ...
  mmaped_file += 32 + ref.size() + alignment;
}
```
This is a main logic plus add missing validations etc.

VyacheslavLevytskyy:
> ... isn't it possible to integrate these two functionalities? I mean parsing header of spirv and checking caps and extensions looks more less like this in code: ...

I'm not sure that this is a relevant question vs. the scope of this particular PR. However, I understand your higher-level concerns and probably we should continue in future discussing such matters, though in a wider context of possible road map of this project development, (re-)using spirv-tools functionality, and other relevant topics.

bwlodarcz:
> > ... isn't it possible to integrate these two functionalities? I mean parsing header of spirv and checking caps and extensions looks more less like this in code: ...
> 
> I'm not sure that this is a relevant question vs. the scope of this particular PR. However, I understand your higher-level concerns and probably we should continue in future discussing such matters, though in a wider context of possible road map of this project development, (re-)using spirv-tools functionality, and other relevant topics.

There are at least three ways to add this API so it's in scope of the PR. First we need to establish one fact - the code in snippet above is more or less parser essential complexity (minimum required to deliver). To that we also need to add the code to print which is a few more lines. 

First way is the way above which created separate parser routine which is independent from the main one used in SPIRVModule.cpp:2125 `std::istream &operator>>(std::istream &I, SPIRVModule &M)` and use Decoder, SPIRVModuleImpl etc. - this is what we have here.

Second way is to try to refactor `std::istream &operator>>(std::istream &I, SPIRVModule &M)` to explicitly handling similar logic     like in above snippet and then move iostream by the parsed distance if this is full parsing or end if this report api call.
e.g
```
std::istream &operator>>(std::istream &I, SPIRVModule &M, bool report) {
  Header H{};
  parse_header(I ,H);
  validate_header(H);
  std::vector<Capabilites> caps;
  parse_capabilities(I, caps);
  validate_caps(caps);
  std::vector<std::String> ext;
  parse_extensions(I, ext);
  validate_exts(ext);
  if (report) {
     Report R(caps, ext);
     print_report(R);
     return I;
  }
  SPIRVModuleImpl impl;
  ... fill caps and ext to impl ...
  Decoder D;
  while (...) { 
  ... rest of logic in parser ...
}
```
and because `I` is mutated during parse you are on exact point where next instructions from logical order are. It should have better memory footprint both in binary size (because there is less code) and heap. We are also using the same logic both in parsing and report.   

Third way is to contain report api call in absolute separation in a few lines of essential complexity snippet in possible maximum containment without even touching existing parser logic, iostream etc. This is a most performant option and the one which cost least memory. E.g. 
```
get_report(char *file_name) {
// In dependency of OS
  mmap();
//
  essential_complexity_parse();
  map_caps_to_text();
  print();
}
```
It's having the similar properties like the first one - it's separate parser logic.

Isn't preferable to consider second and third approach?

VyacheslavLevytskyy:
@asudarsa @MrSidims @vmaksimo @maksimsab @LU-JOHN Can you have a look at the PR please

VyacheslavLevytskyy:
@bwlodarcz Thank you for the thorough analysis. In my opinion, we should use existing library style and approach by utilizing its internal interface calls (that is basically just Decoder in this case) to implement this simple and clear feature.

bwlodarcz:
@VyacheslavLevytskyy Is there any particular technical or nontechnical advantage of this approach in respect to second and third variant?

VyacheslavLevytskyy:
> @VyacheslavLevytskyy Is there any particular technical or nontechnical advantage of this approach in respect to second and third variant?

In my opinion, a nontechnical advantage of this approach is to keep existing library style by utilizing its internal interface calls.

lib/SPIRV/SPIRVReader.cpp
```diff
@@ -4825,6 +4825,137 @@ Instruction *SPIRVToLLVM::transRelational(SPIRVInstruction *I, BasicBlock *BB) {
   return cast<Instruction>(Mutator.getMutated());
 }
 
+std::optional<SPIRVModuleReport> getSpirvReport(std::istream &IS) {
+  int IgnoreErrCode;
```
asudarsa, line 4829: Should this be initialized to something? Thanks

include/LLVMSPIRVLib.h
```diff
@@ -106,6 +106,35 @@ std::unique_ptr<SPIRVModule> readSpirvModule(std::istream &IS,
...
+  std::vector<std::string> Extensions;
+  std::vector<std::string> ExtendedInstructionSets;
+  std::vector<uint32_t> Capabilities;
+};
+/// \brief Partially load SPIR-V from the stream and decode only selected
+/// for the report instructions, needed to retrieve general information
+/// about the module. If this call returns failure, readSPIRVModule is
+/// expected to fail as well.
+/// \returns nullopt on failure.
+std::optional<SPIRVModuleReport> getSpirvReport(std::istream &IS);
```
asudarsa, line 122: Why do we need this variant?

test/spirv_report.spt
```diff
@@ -0,0 +1,43 @@
+; RUN: llvm-spirv %s -to-binary -o %t.spv
+; RUN: llvm-spirv --spirv-print-report %t.spv | FileCheck %s --check-prefix=CHECK-DAG
```
asudarsa, line 2: Can we use --spirv-print-report with other options '-r' for example?

Thanks

lib/SPIRV/SPIRVReader.cpp
```diff
@@ -4825,6 +4825,137 @@ Instruction *SPIRVToLLVM::transRelational(SPIRVInstruction *I, BasicBlock *BB) {
   return cast<Instruction>(Mutator.getMutated());
 }
 
+std::optional<SPIRVModuleReport> getSpirvReport(std::istream &IS) {
```
asudarsa, line 4828: This is a self-contained change. Is it possible to move this to a separate .cpp file (something like SPIRVAnalyzer.cpp)? Of course, we need to consider code bloating.

Thanks

tools/llvm-spirv/llvm-spirv.cpp
```diff
@@ -824,5 +830,39 @@ int main(int Ac, char **Av) {
...
+    std::ifstream IFS(InputFile, std::ios::binary);
+    int ErrCode = 0;
+    std::optional<SPIRV::SPIRVModuleReport> BinReport =
+        SPIRV::getSpirvReport(IFS, ErrCode);
+    if (!BinReport) {
+      std::cout << "Invalid SPIR-V binary, error code is " << ErrCode << "\n";
```
asudarsa, line 840: Will it be possible to print out a human-readable ErrCode info?

Thanks

asudarsa, line 840: Can we add a test for this error reporting, if possible?

tools/llvm-spirv/llvm-spirv.cpp
```diff
@@ -824,5 +830,39 @@ int main(int Ac, char **Av) {
...
+              << "\n";
+    for (auto Capability : TextReport.Capabilities)
+      std::cout << "  Capability: " << Capability << "\n";
+
+    std::cout << " Number of extensions: " << TextReport.Extensions.size()
+              << "\n";
+    for (auto Extension : TextReport.Extensions)
+      std::cout << "  Extension: " << Extension << "\n";
+
+    std::cout << " Number of extension imports: "
```
asudarsa, line 862:
```suggestion
    std::cout << " Number of extended instruction sets: "
```

tools/llvm-spirv/llvm-spirv.cpp
```diff
@@ -824,5 +830,39 @@ int main(int Ac, char **Av) {
...
+              << TextReport.ExtendedInstructionSets.size() << "\n";
+    for (auto ExtendedInstructionSet : TextReport.ExtendedInstructionSets)
+      std::cout << "  Extension Import: " << ExtendedInstructionSet << "\n";
```
asudarsa, line 865:
```suggestion
      std::cout << "  Extended Instruction Set: " << ExtendedInstructionSet << "\n";
```

lib/SPIRV/SPIRVReader.cpp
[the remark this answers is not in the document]
asudarsa: Non-blocking comment: We should try to unify such implementations at some point of time.

Thanks

lib/SPIRV/SPIRVReader.cpp
```diff
@@ -4825,6 +4825,137 @@ Instruction *SPIRVToLLVM::transRelational(SPIRVInstruction *I, BasicBlock *BB) {
...
+  if (!isSPIRVVersionKnown(Word)) {
+    ErrCode = SPIRVEC_InvalidVersionNumber;
+    return {};
+  }
+  SPIRVModuleReport Report;
+  Report.Version = static_cast<SPIRV::VersionNumber>(Word);
+  // Skip: Generator’s magic number, Bound and Reserved word
+  D.ignore(3);
+
+  bool IsLastInstr = false, IsMemoryModelDefined = false;
```
asudarsa, line 4854: IsLastInstr can be replaced with IsReportGenCompleted?

include/LLVMSPIRVLib.h
```diff
@@ -106,6 +106,35 @@ std::unique_ptr<SPIRVModule> readSpirvModule(std::istream &IS,
...
+std::optional<SPIRVModuleReport> getSpirvReport(std::istream &IS, int &ErrCode);
```
asudarsa, line 123: Will it be useful to unconditionally call this API before SPIR-V to LLVM translation and exit early in case of error or will that be redundant?

Thanks

asudarsa (commented):
Overall LGTM. Few minor issues raised.

Thanks

include/LLVMSPIRVLib.h
```diff
@@ -106,6 +106,35 @@ std::unique_ptr<SPIRVModule> readSpirvModule(std::istream &IS,
...
+struct SPIRVModuleReport {
+  SPIRV::VersionNumber Version;
+  uint32_t MemoryModel;
+  uint32_t AddrModel;
+  std::vector<std::string> Extensions;
+  std::vector<std::string> ExtendedInstructionSets;
+  std::vector<uint32_t> Capabilities;
+};
+/// \brief Partially load SPIR-V from the stream and decode only selected
+/// for the report instructions, needed to retrieve general information
```
asudarsa, line 118:
```suggestion
/// instructions that are needed to retrieve general information
```

include/LLVMSPIRVLib.h
```diff
@@ -106,6 +106,35 @@ std::unique_ptr<SPIRVModule> readSpirvModule(std::istream &IS,
...
+/// about the module. If this call returns failure, readSPIRVModule is
```
asudarsa, line 119:
```suggestion
/// about the module. If this call fails, readSPIRVModule is
```

include/LLVMSPIRVLib.h
```diff
@@ -106,6 +106,35 @@ std::unique_ptr<SPIRVModule> readSpirvModule(std::istream &IS,
...
+struct SPIRVModuleTextReport {
+  std::string Version;
+  std::string MemoryModel;
+  std::string AddrModel;
+  std::vector<std::string> Extensions;
+  std::vector<std::string> ExtendedInstructionSets;
+  std::vector<std::string> Capabilities;
+};
+/// \brief Create a human-readable form of the report returned by a call to
+/// getSpirvReport by decoding its binary fields
```
asudarsa, line 134:
```suggestion
/// getSpirvReport by decoding its binary fields.
```

lib/SPIRV/SPIRVReader.cpp
```diff
@@ -4825,6 +4825,137 @@ Instruction *SPIRVToLLVM::transRelational(SPIRVInstruction *I, BasicBlock *BB) {
...
+  case AddressingModelPhysical64:
+    return "Physical64";
+  case AddressingModelPhysicalStorageBuffer64:
+    return "PhysicalStorageBuffer64";
+  default:
+    return "Unknown";
+  }
+}
+
+constexpr std::string_view formatMemoryModel(uint32_t MemoryModel) {
```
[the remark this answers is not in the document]
asudarsa: Not sure. But it does seem to be beyond the scope of this PR. A TODO can be added if necessary.

Thanks

lib/SPIRV/SPIRVReader.cpp
```diff
@@ -4825,6 +4825,137 @@ Instruction *SPIRVToLLVM::transRelational(SPIRVInstruction *I, BasicBlock *BB) {
...
+  TextReport.AddrModel = formatAddressingModel(Report.AddrModel);
+  TextReport.MemoryModel = formatMemoryModel(Report.MemoryModel);
+  // format capability codes as strings
+  std::string Name;
+  for (auto Capability : Report.Capabilities) {
+    bool Found = SPIRVCapabilityNameMap::find(
+        static_cast<SPIRVCapabilityKind>(Capability), &Name);
+    TextReport.Capabilities.push_back(Found ? Name : "Unknown");
+  }
+  // copy other string content as is
```
asudarsa, line 4953: Comment is not clear. Please consider rewording.

Thanks

lib/SPIRV/libSPIRV/SPIRVErrorEnum.h
```diff
@@ -22,3 +22,9 @@ _SPIRV_OP(InvalidWordCount,
 _SPIRV_OP(Requires1_1, "Feature requires SPIR-V 1.1 or greater:")
 _SPIRV_OP(RequiresExtension,
           "Feature requires the following SPIR-V extension:\n")
+_SPIRV_OP(InvalidMagicNumber,
+          "Invalid Magic Number.")
+_SPIRV_OP(InvalidVersionNumber,
+          "Invalid Version Number.")
+_SPIRV_OP(UnspecifiedMemoryModel, "Unspecified Memory Model.")
+_SPIRV_OP(RepeatedMemoryModel, "Expects the single OpMemoryModel instruction.")
```
asudarsa, line 30:
```suggestion
_SPIRV_OP(RepeatedMemoryModel, "Expects a single OpMemoryModel instruction.")
```

lib/SPIRV/SPIRVReader.cpp
```diff
@@ -4825,6 +4825,137 @@ Instruction *SPIRVToLLVM::transRelational(SPIRVInstruction *I, BasicBlock *BB) {
...
+
+std::optional<SPIRVModuleReport> getSpirvReport(std::istream &IS,
+                                                int &ErrCode) {
+  SPIRVWord Word;
+  std::string Name;
+  std::unique_ptr<SPIRVModule> BM(SPIRVModule::createSPIRVModule());
+  SPIRVDecoder D(IS, *BM);
+  D >> Word;
+  if (Word != MagicNumber) {
+    ErrCode = SPIRVEC_InvalidMagicNumber;
```
asudarsa, line 4841: Should we try to return the invalid data (in all possible cases) as well? Or will that be redundant?

Thanks

include/LLVMSPIRVLib.h
VyacheslavLevytskyy: Actually it may be the main version of the call, but we include a version with an error code for sake of completeness. From prospective customers we here about a use case when they check if a SDPIRV binary chunk is correct, that's all they basically need.

lib/SPIRV/SPIRVReader.cpp
VyacheslavLevytskyy: That's a good point in general, thank you. However, I'm afraid this is a much wider issue, because lib/SPIRV/SPIRVReader.cpp is bloated indeed. This feature is not exactly self-contained, rather it is integrated with existing internal API, so probably the question of splitting of lib/SPIRV/SPIRVReader.cpp deserves actually a separate PR that should address a question of proper management of features presented in lib/SPIRV/SPIRVReader.cpp. If we started discussing how to proper split lib/SPIRV/SPIRVReader.cpp as a part of this PR it would effectively block the PR instead of implementing this as a small and incremental change that uses existing internal api.

lib/SPIRV/SPIRVReader.cpp
VyacheslavLevytskyy: No, we don't need its value whatsoever. It's a destination of a value that is always ignored.

lib/SPIRV/SPIRVReader.cpp
VyacheslavLevytskyy: A general scheme is to validate the SPIRV header and either return valid data & success code, or return empty std::optional with a properly set error code.

lib/SPIRV/SPIRVReader.cpp
VyacheslavLevytskyy: I think comparing to D and BM it's not so obviously misleading name, but thank you for the suggestion, I will certainly apply it.

lib/SPIRV/SPIRVReader.cpp
VyacheslavLevytskyy: Sure, thank you.

test/spirv_report.spt
VyacheslavLevytskyy: Adding of a new option is non-disruptive in the meaning that it mimics all similar options all around it, and can be as much combined with other options as similar existing options do.

tools/llvm-spirv/llvm-spirv.cpp
VyacheslavLevytskyy: Very good point indeed, thank you. It would certainly make sense to add human-readable error info. However, the library doesn't support conversion from an error code into a human-readable explanation yet. This could be done if we added one more API call for implementing this conversion. So our choice here basically either to keep this PR as is in the part of reporting error code, or to add the third API call, something like:

`/// \brief Returns the message associated with the error code.
/// \returns empty string if no known error code is found.
std::string getErrorMessage(int ErrCode);`

@asudarsa @MrSidims What would you prefer in this case? To add getErrorMessage() as a part of this PR, or not?

VyacheslavLevytskyy: Sure, thank you.

VyacheslavLevytskyy: Added in a new commit a test file to test negative condition.

VyacheslavLevytskyy:
@asudarsa Thank you for the code review! Please see changes applied

VyacheslavLevytskyy:
Failing test DebugInfo/NonSemantic/Shader200/FortranArray.ll is not related to this PR

tools/llvm-spirv/llvm-spirv.cpp
MrSidims, line 840: nit
```suggestion
      std::cerr<< "Invalid SPIR-V binary, error code is " << ErrCode << "\n";
```

tools/llvm-spirv/llvm-spirv.cpp
```diff
@@ -824,5 +830,39 @@ int main(int Ac, char **Av) {
...
+        SPIRV::formatSpirvReport(BinReport.value());
+
+    std::cout << "SPIR-V module report:"
+              << "\n Version: " << TextReport.Version
+              << "\n Memory model: " << TextReport.MemoryModel
+              << "\n Addressing model: " << TextReport.AddrModel << "\n";
+
+    std::cout << " Number of capabilities: " << TextReport.Capabilities.size()
+              << "\n";
+    for (auto Capability : TextReport.Capabilities)
```
MrSidims, line 854: nit here and after
```suggestion
    for (auto &Capability : TextReport.Capabilities)
```

MrSidims (commented):
Few nits, other then that LGTM

tools/llvm-spirv/llvm-spirv.cpp
VyacheslavLevytskyy: ok, thank you

tools/llvm-spirv/llvm-spirv.cpp
VyacheslavLevytskyy: thank you

VyacheslavLevytskyy:
@MrSidims Thank you for the code review! Please see changes applied

VyacheslavLevytskyy:
Failing test DebugInfo/NonSemantic/Shader200/FortranArray.ll is not related to this PR

lib/SPIRV/SPIRVReader.cpp
asudarsa: I agree that this can be done as separate refactoring. A TODO can be handy here.

tools/llvm-spirv/llvm-spirv.cpp
asudarsa: Please add a TODO here. Thanks

lib/SPIRV/SPIRVReader.cpp
asudarsa: To clarify, I was thinking if we need to report the error as follows:
'5' is not a valid magic number.
But, reading your earlier comment about the error reporting not being the primary feature, I suppose it's not that important. Thanks

asudarsa (approved):
Good job. LGTM. Thanks.

tools/llvm-spirv/llvm-spirv.cpp
VyacheslavLevytskyy: Thank you. I created https://github.com/KhronosGroup/SPIRV-LLVM-Translator/pull/2304 to cover another API call of getErrorMessage(int ErrCode), so that we are able to resolve it quickly after this PR get merged.

tools/llvm-spirv/llvm-spirv.cpp
svenvh: This hasn't been done yet as far as I can tell?

test/negative/spirv_report_bad_input.spt
```diff
@@ -0,0 +1,32 @@
+; RUN: llvm-spirv %s -to-binary -o %t.spv
+; RUN: echo "0" > %t_corrupted.spv && cat %t.spv >> %t_corrupted.spv
+; RUN: not llvm-spirv --spirv-print-report %t_corrupted.spv 2>&1 | FileCheck %s --check-prefix=CHECK-ERROR
+;
+; CHECK-ERROR: Invalid SPIR-V binary
+
```
svenvh, line 6: Perhaps add a comment explaining what's bad about this input.

tools/llvm-spirv/llvm-spirv.cpp
VyacheslavLevytskyy: indeed, thank you

test/negative/spirv_report_bad_input.spt
VyacheslavLevytskyy: I'll add a comment to the test file explaining what's bad about this input.

However, if talking about possible output from `llvm-spirv --spirv-print-report`, more user friendly output will be available later, after https://github.com/KhronosGroup/SPIRV-LLVM-Translator/pull/2304 will add a call to get human-readable info by an error code (std::string getErrorMessage(int ErrCode);) This test file will then serve to check both #2298 and #2304

VyacheslavLevytskyy: @asudarsa Please see also the above comment wrt. your comment about testing of https://github.com/KhronosGroup/SPIRV-LLVM-Translator/pull/2304

MrSidims (approved):

PR #2301, opened 2024-01-12
Revert "Don't wrap kernels that are not being called in the module (#2119)"

karolherbst:
kernels can be called from other translation units, so we can't know if a kernel is called or not. Reverting this so we get back `Export` Linkage declarations which are required to support linking of spirv files.

This reverts commit 46285e43abcc606791a0dd18858fd88af2894294.

### DIFF
--- a/lib/SPIRV/SPIRVRegularizeLLVM.cpp
+++ b/lib/SPIRV/SPIRVRegularizeLLVM.cpp
@@ -363,10 +363,9 @@ bool SPIRVRegularizeLLVMBase::runRegularizeLLVM(Module &Module) {
 /// Remove entities not representable by SPIR-V
 bool SPIRVRegularizeLLVMBase::regularize() {
   eraseUselessFunctions(M);
+  addKernelEntryPoint(M);
   expandSYCLTypeUsing(M);
 
-  // Kernels called by other kernels
-  std::vector<Function *> CalledKernels;
   for (auto I = M->begin(), E = M->end(); I != E;) {
     Function *F = &(*I++);
     if (F->isDeclaration() && F->use_empty()) {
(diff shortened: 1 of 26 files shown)


MrSidims (commented):
Just for my understanding, what you see is the following:
lets say test.cl
```
kernel void foo();

kernel void boo() {
      foo();
}
```
test1.cl is:
`kernel void foo() { }`

after linking with spirv-link we get:
`error: 0: Unresolved external reference to "foo".`
because in the module for test.cl we see:
```
Name 4 "foo"
...
Decorate 4 LinkageAttributes "foo" Import
...
Function 2 4 0 3
```
Correct?

I was thinking if we can explore other options other than revert, like hacking the module and changing the linkage type, but it not the correct way and it also makes the validator unhappy.

MrSidims (approved):

karolherbst:
> Correct?

yes

> 
> I was thinking if we can explore other options other than revert, like hacking the module and changing the linkage type, but it not the correct way and it also makes the validator unhappy.

Yeah.. I don't think we can actually do something else here as all functions not being `static` need `Extern` linkage one way or the other. The situation with kernels are just that their interface is special so we need that wrapper.

I wonder however if we could make the kernels itself `Export`? Haven't through about that possibility much, but then we have to make the wrapper `static` so those don't get exported.

svenvh (approved):

karolherbst:
> `transcoding/KernelArgTypeInOpString2.ll` regresses, so I'm gonna push a fix for that shortly

done and while at it, I also kept the additions to the `entry_point_func.ll` and fixed it for the revert.

karolherbst:
fyi, on llvm 414ea3a77181ef01d2cc2ad34950fc1c03ce0d41 I have those failures:

```
Failed Tests (6):
  LLVM_SPIRV :: DebugInfo/Generic/debug-info-eis-option.ll
  LLVM_SPIRV :: DebugInfo/Generic/linear-dbg-value.ll
  LLVM_SPIRV :: SpecConstants/specconstantop-arraylength.spvasm
  LLVM_SPIRV :: SpecConstants/specconstantop-copymemsized.spvasm
  LLVM_SPIRV :: SpecConstants/specconstantop-init.spvasm
  LLVM_SPIRV :: freeze.ll
```

but they seem to be there regardless of my patch

MrSidims:
FortranArray.ll failure is fixed here: https://github.com/KhronosGroup/SPIRV-LLVM-Translator/pull/2303
I actually don't see any other test failures in either CI or locally with fresh-new LLVM.
Anyway, merging.

PR #2304, opened 2024-01-18
add API to query error message by an error code

VyacheslavLevytskyy:
The goal of the PR is to add API to SPIR-V LLVM Translator to query error message by an error code as discussed in https://github.com/KhronosGroup/SPIRV-LLVM-Translator/pull/2298

A need and possible application is a way to generate human-readable error info by error codes returned by other SPIRV Translator API calls, including getSpirvReport().

### DIFF
--- a/include/LLVMSPIRVLib.h
+++ b/include/LLVMSPIRVLib.h
@@ -135,6 +135,10 @@ struct SPIRVModuleTextReport {
 /// \returns String with the human-readable report.
 SPIRVModuleTextReport formatSpirvReport(const SPIRVModuleReport &Report);
 
+/// \brief Returns the message associated with the error code.
+/// \returns empty string if no known error code is found.
+std::string getErrorMessage(int ErrCode);
+
 } // End namespace SPIRV
 
 namespace llvm {
--- a/lib/SPIRV/libSPIRV/SPIRVError.cpp
+++ b/lib/SPIRV/libSPIRV/SPIRVError.cpp
@@ -1,4 +1,4 @@
-//===- SPIRVError.cpp - SPIR-V error code and checking ------------*- C++ -*-===//
+//===- SPIRVError.cpp - SPIR-V error code and checking ----------*- C++ -*-===//
 //
 //                     The LLVM/SPIRV Translator
 //
--- a/lib/SPIRV/libSPIRV/SPIRVError.cpp
+++ b/lib/SPIRV/libSPIRV/SPIRVError.cpp
@@ -1,0 +1,56 @@
+//===- SPIRVError.cpp - SPIR-V error code and checking ------------*- C++ -*-===//
+//
+//                     The LLVM/SPIRV Translator
+//
+// This file is distributed under the University of Illinois Open Source
+// License. See LICENSE.TXT for details.
+//
+// Copyright (c) 2014 Advanced Micro Devices, Inc. All rights reserved.
+//
+// Permission is hereby granted, free of charge, to any person obtaining a
+// copy of this software and associated documentation files (the "Software"),
+// to deal with the Software without restriction, including without limitation
+// the rights to use, copy, modify, merge, publish, distribute, sublicense,
+// and/or sell copies of the Software, and to permit persons to whom the
+// Software is furnished to do so, subject to the following conditions:
+//
+// Redistributions of source code must retain the above copyright notice,
+// this list of conditions and the following disclaimers.
+// Redistributions in binary form must reproduce the above copyright notice,
+// this list of conditions and the following disclaimers in the documentation
+// and/or other materials provided with the distribution.
+// Neither the names of Advanced Micro Devices, Inc., nor the names of its
+// contributors may be used to endorse or promote products derived from this
+// Software without specific prior written permission.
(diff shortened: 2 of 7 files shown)


include/LLVMSPIRVLib.h
```diff
@@ -106,6 +106,10 @@ std::unique_ptr<SPIRVModule> readSpirvModule(std::istream &IS,
                                              const SPIRV::TranslatorOpts &Opts,
                                              std::string &ErrMsg);
 
+/// \brief Returns the message associated with the error code.
+/// \returns empty string if no known error code is found.
+std::string getErrorMessage(int ErrCode);
```
asudarsa, line 111: Do we want to have a separate API for this? or can we just modify getSpirvReport to return a error string instead of an Id?

Thanks

asudarsa (commented):
Looks ok. Just a high level clarification question. Please take a look. Thanks

include/LLVMSPIRVLib.h
VyacheslavLevytskyy: We don't want getSpirvReport to return strings, it should be as lightweight as it's possible. So we need to keep integer error code in getSpirvReport, and I doubt even that it will be much used, I'd expect that in backend it would be ignored. So the main use for error codes and error messages is for human-readable info.

Re: testing -- Sure, after the merge of https://github.com/KhronosGroup/SPIRV-LLVM-Translator/pull/2298, we'll be able to add more tests. Until the merge of #2298 that's a draft PR.

lib/SPIRV/libSPIRV/SPIRVError.h
```diff
@@ -174,6 +174,18 @@ inline bool SPIRVErrorLog::checkError(bool Cond, SPIRVErrorCode ErrCode,
   return Cond;
 }
 
+// Return the message associated with the error code or empty string if the
+// error code is invalid
+inline std::string getErrorMessage(int ErrCode) {
+  std::string ErrorMessage;
+  // It's ok if the error code is not present in the map, so we don't need to
```
LU-JOHN, line 181: Why is it ok if getErrorMessage is called with an invalid error code?  What if the programmer makes a mistake and passes a wrong int variable to getErrorMessage?  getErrorMessage 
return an empty string which is the same string returned by the Success error code:

> _SPIRV_OP(Success, "")

VyacheslavLevytskyy: The goal of a new API call is to return a human-readable info about an error code. It's not aimed to be a check whether the argument is an error or success -- the programmer uses the integer value for such a check. This makes ambiguity of the empty string less important. However, if we really wished to address the issue of returning an empty string for both Success error code and for a missing error code, I'd say it's a problem of a choice of the value for _SPIRV_OP(Success, ""). Personally, I'd vote for changing it into _SPIRV_OP(Success, "Success") that both looks logical and resolves the issue of distinguishing between Success error code and a missing error code.

LU-JOHN: I would prefer changing success to "Success" and also changing an invalid error code to something like "Invalid error code input".   "Invalid error code input" will let the user more quickly understand what is happening than returning an empty string.   I think it also makes the code clearer.  When I see the comment:

>  // It's ok if the error code is not present in the map

it is unsettling to me as a code reader.  I will think: "Why should it be OK if the error code is not in the map?  This is not what I would normally expect, so is there some property of the code that I am missing."

include/LLVMSPIRVLib.h
VyacheslavLevytskyy: @asudarsa There is a test case now in test/negative/spirv_report_bad_input.spt for this API call

VyacheslavLevytskyy:
@bwlodarcz @LU-JOHN @vmaksimo @MrSidims @svenvh Can you please have a look?

lib/SPIRV/libSPIRV/SPIRVError.cpp
```diff
@@ -0,0 +1,56 @@
+//===- SPIRVError.cpp - SPIR-V error code and checking ----------*- C++ -*-===//
+//
+//                     The LLVM/SPIRV Translator
+//
+// This file is distributed under the University of Illinois Open Source
+// License. See LICENSE.TXT for details.
+//
+// Copyright (c) 2014 Advanced Micro Devices, Inc. All rights reserved.
```
asudarsa, line 8:
```suggestion
// Copyright (c) 2024 Intel Corporation. All rights reserved.
```

asudarsa (approved):
LGTM. please modify copyright text before merge Thanks

lib/SPIRV/libSPIRV/SPIRVError.cpp
MrSidims: Neither of them! Please use Khronos copy rights from https://github.com/KhronosGroup/SPIRV-LLVM-Translator/issues/80

VyacheslavLevytskyy:
@asudarsa @MrSidims Thank you, fixed the copyright header

svenvh (approved):

PR #2306, opened 2024-01-19
[NFC] Test behavior of common global variables

LU-JOHN:
Add tests to ensure that:

1. A common global without addrspace is reported as an error
2. A common addrspace global is not transformed to a locally allocated variable

Updated error message from "can not" to "cannot".  Otherwise this is a NFC.


vmaksimo (approved):

test/CommonAddrspaceGlobalVar.ll
```diff
@@ -0,0 +1,20 @@
+; Ensure that a "common global" is not converted to a locally allocated
```
MrSidims, line 1: test/CommonAddrspaceGlobalVar.ll -> test/transcoding/CommonAddrspaceGlobalVar.ll

test/CommonGlobalVarNoAddrspace.ll
```diff
@@ -0,0 +1,24 @@
+; Ensure "common global" with no addrspace is reported as an error
```
MrSidims, line 1: quite unused directory, but should be placed here IMHO

test/CommonGlobalVarNoAddrspace.ll -> test/negative/CommonGlobalVarNoAddrspace.ll

test/CommonAddrspaceGlobalVar.ll
LU-JOHN: done

test/CommonGlobalVarNoAddrspace.ll
LU-JOHN: done

MrSidims (approved):
test/GlobalVarNoAddrspace.ll - ideally should also be moved to negative, but it wasn't added in this patch, so I don't mind

LU-JOHN:
> test/GlobalVarNoAddrspace.ll - ideally should also be moved to negative, but it wasn't added in this patch, so I don't mind

I've moved test/GlobalVarNoAddrspace.ll.     This is more consistent.

PR #2314, opened 2024-01-23
[DebugInfo] Fix instruction enumeration for template parameter

MrSidims:
These two opcodes were mistakenly swapped when they were originally added, at least according to the DebugInfo, OpenCL.DebugInfo.100, and NonSemantic.Shader.DebugInfo.100 extended instruction sets.

This might break existing third-party SPIR-V translators if they are accommodating this bug, mistakenly or otherwise.

Original author: Fraser Cormack


MrSidims:
@frasercrmck sorry for the delay and please take a look. The plan is to get approvals for this PR, afterwards I'll create backport PRs to other release branches, that have OpenCL DebugInfo implemented and after merge with some graceful period - I'll merge the PR on the main branch.

lib/SPIRV/SPIRVToLLVMDbgTran.cpp
```diff
@@ -1267,6 +1267,23 @@ SPIRVToLLVMDbgTran::transTypeTemplateParameter(const SPIRVExtInst *DebugInst) {
                                                              false);
 }
 
+// Workaround for the incorrect enum used for OpTypeTemplateTemplateParameter
+// and OpTypeTemplateParameterPack instructions previously. The W/A added
+// to enforce compatibility with previously generated SPIR-V modules.
+DINode *SPIRVToLLVMDbgTran::transTypeTemplateParameterInst(
```
MrSidims, line 1273: This code is only tested locally. Ideally we should add an infrastructure to test pre-compiled binaries to ensure backward compatibility (that was probably also discussed during tooling wg call). I'll open an issue in github for this.

vmaksimo (approved):

LU-JOHN (approved):
LGTM

svenvh (approved):

lib/SPIRV/SPIRVToLLVMDbgTran.cpp
```diff
@@ -1267,6 +1267,23 @@ SPIRVToLLVMDbgTran::transTypeTemplateParameter(const SPIRVExtInst *DebugInst) {
...
+    const SPIRVExtInst *DebugInst) {
+  constexpr SPIRVWord IndexToCheck =
+      SPIRVDebug::Operand::TypeTemplateTemplateParameter::TemplateNameIdx;
+  const SPIRVWordVec &Ops = DebugInst->getArguments();
+  // We can't distinguish these 2 instructions by number of SPIR-V words which
+  // would be better. But we know, that the second parameter of
+  // OpTypeTemplateTemplateParameter is 'Template Name', which for
+  // OpTypeTemplateParameterPack it is 'Source'.
+  if (BM->get<SPIRVValue>(Ops[IndexToCheck])->getOpCode() == OpString)
```
frasercrmck, line 1282: Ah yeah, we have a workaround like this locally. However, I was worried that it could be `DebugInfoNone` (I don't find the spec very clear about which operands are allowed to be `DebugInfoNone`)

I came up with a multi-stage guesstimation firstly using the number of operands, which may be more than 10 for only `DebugTypeTemplateParameterPack`, or failing that this operand being `OpString` for `DebugTypeTemplateTemplateParameter`, or this operand being `DebugSource` for `DebugTypeTemplateParameterPack` or if both of those are `DebugInfoNone` then some tie breaker I've forgotten about. It was all a bit complicated.

If this can only truly ever be `OpString` and never `DebugInfoNone` then this looks fine to me.

MrSidims: Indeed

PR #2316, opened 2024-01-25
[NFC] Split of SPT and SPIR-V in header parsing

bwlodarcz:
Translator accepting two formats currently - SPIR-V and SPT.
SPT is a textual format used internally for tests without any formal specification.
Parsing a textual format is slower than binary format and SPT also is less strict in comparison to SPIRV. 
Because parsing in Translator is organized around iostream interfaces and implementations are mostly shared, the binary format is parsed much slower than it can be.
The commit starts an effort of splitting implementations to make SPIR-V hot path faster and easier to read.
In addition to that from both implementations the Decoder class is removed - it doesn't bring any important functionality which couldn't be solved by few lines of explicit simple code.

### DIFF
--- a/lib/SPIRV/libSPIRV/SPIRVModule.cpp
+++ b/lib/SPIRV/libSPIRV/SPIRVModule.cpp
@@ -572,6 +572,8 @@ private:
   SPIRVAliasInstMDMap AliasInstMDMap;
 
   void layoutEntry(SPIRVEntry *Entry);
+  std::istream &parseSPT(std::istream &I);
+  std::istream &parseSPIRV(std::istream &I);
 };
 
 SPIRVModuleImpl::~SPIRVModuleImpl() {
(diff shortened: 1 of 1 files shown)


bwlodarcz:
This is a draft now and somewhat consider it a R&D - discussion welcome.

MrSidims:
Restarting CI

bwlodarcz:
Moved as available to review.
What's done:
* Decoder is removed from SPIRV and SPT parser. I cannot remove it completely from codebase cause it's used in reporting feature. Merging these procedures can and should be next goal.
* From that point there is a possibility to investigate and try to fix getEntry and addEntry - two biggest cycle consumers in translation from SPIRV to LLVM. 
* Begin to split SPT and SPIRV code paths. SPIRV is hot and SPT mainly used for test so less relevant.

lib/SPIRV/libSPIRV/SPIRVModule.cpp
```diff
@@ -2103,70 +2105,275 @@ static std::string to_string(VersionNumber Version) {
...
+  MI.resolveUnknownStructFields();
+  return I;
+}
+
+std::istream &SPIRVModuleImpl::parseSPIRV(std::istream &I) {
+  SPIRVModuleImpl &MI = *this;
+  MI.setAutoAddCapability(false);
+  MI.setAutoAddExtensions(false);
+
+  struct [[gnu::packed]] Header {
```
asudarsa, line 2296: I have read both advantages and disadvantages of using packed structs. Given that we do not use packed structs anywhere else, I would recommend to not use it.

lib/SPIRV/libSPIRV/SPIRVModule.cpp
```diff
@@ -2103,70 +2105,275 @@ static std::string to_string(VersionNumber Version) {
   return to_string(static_cast<uint32_t>(Version));
 }
 
-std::istream &operator>>(std::istream &I, SPIRVModule &M) {
-  SPIRVDecoder Decoder(I, M);
-  SPIRVModuleImpl &MI = *static_cast<SPIRVModuleImpl *>(&M);
-  // Disable automatic capability filling.
+namespace {
```
asudarsa, line 2108: I think it might be a good idea to start populating a new file for this functionality SPIRVParser.cpp. You can do that in this PR or the next.

Thanks

lib/SPIRV/libSPIRV/SPIRVModule.cpp
```diff
@@ -2103,70 +2105,275 @@ static std::string to_string(VersionNumber Version) {
...
+SPIRVEntry *parseAndCreateSPIRVEntry(SPIRVWord &WordCount, Op &OpCode,
```
asudarsa, line 2109: Will it be a good idea to split this into 'parseEntry and createEntry?

lib/SPIRV/libSPIRV/SPIRVModule.cpp
```diff
@@ -2103,70 +2105,275 @@ static std::string to_string(VersionNumber Version) {
...
+  }
   MI.resolveUnknownStructFields();
   return I;
 }
 
+std::istream &operator>>(std::istream &I, SPIRVModule &M) {
+  SPIRVModuleImpl &MI = *static_cast<SPIRVModuleImpl *>(&M);
+#ifdef _SPIRV_SUPPORT_TEXT_FMT
+  if (SPIRVUseTextFormat) {
+    return MI.parseSPT(I);
```
asudarsa, line 2371: Maybe we can call these parseText and parseBinary?

lib/SPIRV/libSPIRV/SPIRVModule.cpp
```diff
@@ -2103,70 +2105,275 @@ static std::string to_string(VersionNumber Version) {
...
-    SPIRVEntry *Entry = Decoder.getEntry();
-    if (Entry != nullptr)
-      M.add(Entry);
+  SPIRVWord WordCount = 0;
+  Op OpCode = OpNop;
+  SPIRVEntry *Scope = nullptr;
+  while (true) {
+    WordCount = ReadSPIRVWord(I);
+    if (I.fail()) {
+      SPIRVDBG(spvdbgs() << "[SPIRVDecoder] getWordCountAndOpCode FAIL 0 0\n");
```
asudarsa, line 2258: Remove [SPIRVDecoder]?

lib/SPIRV/libSPIRV/SPIRVModule.cpp
```diff
@@ -2103,70 +2105,275 @@ static std::string to_string(VersionNumber Version) {
...
+      break;
+    }
+    SPIRVDBG(spvdbgs() << "[SPIRVDecoder] getWordCountAndOpCode " << WordCount
+                       << " " << OpCodeNameMap::map(OpCode) << '\n');
+    if (!MI.isModuleValid()) {
+      break;
+    }
+
+    SPIRVEntry *Entry =
+        parseAndCreateSPIRVEntry(WordCount, OpCode, Scope, MI, I);
```
asudarsa, line 2277: We can just call this parseAndCreateEntry?

asudarsa:
> Moved as available to review. What's done:
> 
> * Decoder is removed from SPIRV and SPT parser. I cannot remove it completely from codebase cause it's used in reporting feature. Merging these procedures can and should be next goal.
> * From that point there is a possibility to investigate and try to fix getEntry and addEntry - two biggest cycle consumers in translation from SPIRV to LLVM.
> * Begin to split SPT and SPIRV code paths. SPIRV is hot and SPT mainly used for test so less relevant.

I agree fully with the first two points. But, not much with splitting SPIRV and SPT code paths. Code management becomes a bit cumbersome as there is considerable 'common' code between the two paths. Please provide justification for split.

Thanks again for this refactoring.

asudarsa (commented):
Overall design looks good. Please consider suggestions. Also, an overall comment. Please consider not using the term 'SPIRV parser'. It might be a bit misleading as we do not have any stand-alone SPIRV Parser (as compared to SPIRV Reader or SPIRV Writer).
Foe example, the topic can be renamed: "Removal of Decoder class from logic used to parse spirv"

lib/SPIRV/libSPIRV/SPIRVModule.cpp
bwlodarcz: You must consider that in this case packed struct is a necessity - it's a direct target of memcpy (I.read in this case). If packed wouldn't be used you won't have any guarantee that sizeof(Header) is equal to 20 bytes (struct alignment rules comes into play and filling a struct with hidden members).  
If codebase clarity is a major consideration against `[[gnu::packed]]` please note that in this context `Header` is a helper structure reproducing SPIR-V Header layout which have fixed 20 byte (5 x Word) size. It's scope is local to the function and the symbol isn't visible anywhere else.

lib/SPIRV/libSPIRV/SPIRVModule.cpp
bwlodarcz: Yes and it should be considered one of the next steps. Unfortunately, basing on my today (02/09/24) evaluation this function requires much more work to make full split possible. It's worth noting that the function (it's origin getEntry) is also one of the major contributor to overall slowness of the translator - it's responsible for ~40% of overall runtime (SPIR-V->IR) based on my recent benchmark perf measurement.

bwlodarcz:
> > Moved as available to review. What's done:
> > 
> > * Decoder is removed from SPIRV and SPT parser. I cannot remove it completely from codebase cause it's used in reporting feature. Merging these procedures can and should be next goal.
> > * From that point there is a possibility to investigate and try to fix getEntry and addEntry - two biggest cycle consumers in translation from SPIRV to LLVM.
> > * Begin to split SPT and SPIRV code paths. SPIRV is hot and SPT mainly used for test so less relevant.
> 
> I agree fully with the first two points. But, not much with splitting SPIRV and SPT code paths. Code management becomes a bit cumbersome as there is considerable 'common' code between the two paths. Please provide justification for split.
> 
> Thanks again for this refactoring.

@asudarsa Thank you. I answered to the conversations not related to the splitting SPIR-V and SPT code paths.
This will be combined answer to the whole split topic.

Unfortunately, there are major technical prohibitors which renders the unification almost impossible without serous sacrifice of  performance.

SPT is an internal text format which specification is hidden inside our parser implementation. This stays in direct opposition of SPIR-V which is a binary format with written official specification. SPT is not only one text format which translates directly to SPIR-V e.g. spirv asm from spirv-tools have the same property. That's why calling it just text (in parseText) is somewhat misleading because it's ambiguous without knowledge about SPT itself.

The SPT, as a format, have much looser parse requirements than SPIR-V. One of the examples can be option to include comments in arbitrary places which can have arbitrary lengths. This property alone means that e.g. parsing header in one read is impossible for SPT because there is a possibility that comment can be inserted between relevant fields.
That's also a reason why in SPT version every Header field needs to be checked with `I.fail()` when in SPIR-V check must be done only once after `I.read` of the whole Header. 
Parsing text is also slower than parsing binary formats - 4 bytes in binary format (Word) can contain much more information than in text format where 4 bytes represent 4 alphanumeric symbols.
If we stay with unification of both formats we would have to obey the rules of loosest (slowest) common denominator - SPT format. This will prohibit any potential advantage  from SPIR-V binary format and his written specification. 

Last but not least. It seems that Translator hot path (critical functionality) is SPIR-V translation to IR and SPT is used only as a format for testing which renders it much colder code path from the user perspective. This indicates that potential performance works should be performed in SPIR-V code path which much less consideration of SPT. From that point of view split is necessary to unlock full potential of hot path optimizations.

lib/SPIRV/libSPIRV/SPIRVModule.cpp
bwlodarcz: Basing on my current investigation, when we don't have separated building blocks of parser and `SPIRVModule` (`getEntry` problem) this kind of split will yield more trouble than it's worth. Please note that `SPIRVModule` fields like SPIRVVersion are private and `operator>>` is a kind of factory/constructor. If I would try to separate it now then code changes would be more significant which would yield PR larger.

lib/SPIRV/libSPIRV/SPIRVModule.cpp
bwlodarcz: The name of the function in this case have a precise meaning - `SPIRVEntry` which is translator internal type used for Entity/Object parsed by parser. It's part of current internal parser algorithm and not something which should be used by anyone outside parser scope. That's also why the symbol is not exported externally.

MrSidims (approved):
LGTM

lib/SPIRV/libSPIRV/SPIRVModule.cpp
svenvh: The [Windows CI build log](https://github.com/KhronosGroup/SPIRV-LLVM-Translator/actions/runs/7845239975/job/21409234150?pr=2316) for this PR contains

> warning C5030: attribute 'gnu::packed' is not recognized

So I guess portability might be a concern here?

bwlodarcz: @svenvh Nice catch. It slipped my mind that the application needs to also be buildable by MSVC. 
There are two alternatives here:
* `#pragma pack` in windows build
* because all header contain 4-byte words we can just use array - which is most portable solution.

It seems that the latter is preferable option - rules for size of arrays in standard are more strict so it will have required number of bytes. This solves portability problem.

lib/SPIRV/libSPIRV/SPIRVModule.cpp
```diff
@@ -2103,70 +2105,275 @@ static std::string to_string(VersionNumber Version) {
...
+                                     std::istream &IS) {
+  if (WordCount == 0 || OpCode == OpNop) {
+    return nullptr;
+  }
+  SPIRVEntry *Entry = SPIRVEntry::create(OpCode);
+  assert(Entry);
+  Entry->setModule(&M);
+  if (isModuleScopeAllowedOpCode(OpCode) && !Scope) {
+  } else
+    Entry->setScope(Scope);
```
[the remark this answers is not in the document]
bwlodarcz: Done.

lib/SPIRV/libSPIRV/SPIRVModule.cpp
```diff
@@ -2103,70 +2105,275 @@ static std::string to_string(VersionNumber Version) {
...
+      break;
+    }
+  }
+
+  MI.resolveUnknownStructFields();
+  return I;
+}
+
+std::istream &SPIRVModuleImpl::parseSPIRV(std::istream &I) {
+  SPIRVModuleImpl &MI = *this;
```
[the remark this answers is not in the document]
bwlodarcz: Check for eof is added.

MrSidims:
Restarting CI just in case. The patch remains LGTM, just waiting for another approval from the commentators.

MrSidims:
@asudarsa please take a look

MrSidims:
Apparently Arvind is on a sick leave.

@svenvh are you OK with this split? We plan to enhance decoder in the future speeding it up.

@bwlodarcz please resolve merge conflict

MrSidims:
Sure, I'll forward invitation for Bertrand, though probably it's outside of his working hours.

svenvh (approved):

lib/SPIRV/libSPIRV/SPIRVModule.cpp
asudarsa: Agreed. We can attempt this code refactoring at a later stage when things are more amenable to this change. Thanks

lib/SPIRV/libSPIRV/SPIRVModule.cpp
asudarsa: Please add a TODO above the function definition with this info. Thanks

asudarsa:
> > > Moved as available to review. What's done:
> > > 
> > > * Decoder is removed from SPIRV and SPT parser. I cannot remove it completely from codebase cause it's used in reporting feature. Merging these procedures can and should be next goal.
> > > * From that point there is a possibility to investigate and try to fix getEntry and addEntry - two biggest cycle consumers in translation from SPIRV to LLVM.
> > > * Begin to split SPT and SPIRV code paths. SPIRV is hot and SPT mainly used for test so less relevant.
> > 
> > 
> > I agree fully with the first two points. But, not much with splitting SPIRV and SPT code paths. Code management becomes a bit cumbersome as there is considerable 'common' code between the two paths. Please provide justification for split.
> > Thanks again for this refactoring.
> 
> @asudarsa Thank you. I answered to the conversations not related to the splitting SPIR-V and SPT code paths. This will be combined answer to the whole split topic.
> 
> Unfortunately, there are major technical prohibitors which renders the unification almost impossible without serous sacrifice of performance.
> 
> SPT is an internal text format which specification is hidden inside our parser implementation. This stays in direct opposition of SPIR-V which is a binary format with written official specification. SPT is not only one text format which translates directly to SPIR-V e.g. spirv asm from spirv-tools have the same property. That's why calling it just text (in parseText) is somewhat misleading because it's ambiguous without knowledge about SPT itself.
> 
> The SPT, as a format, have much looser parse requirements than SPIR-V. One of the examples can be option to include comments in arbitrary places which can have arbitrary lengths. This property alone means that e.g. parsing header in one read is impossible for SPT because there is a possibility that comment can be inserted between relevant fields. That's also a reason why in SPT version every Header field needs to be checked with `I.fail()` when in SPIR-V check must be done only once after `I.read` of the whole Header. Parsing text is also slower than parsing binary formats - 4 bytes in binary format (Word) can contain much more information than in text format where 4 bytes represent 4 alphanumeric symbols. If we stay with unification of both formats we would have to obey the rules of loosest (slowest) common denominator - SPT format. This will prohibit any potential advantage from SPIR-V binary format and his written specification.
> 
> Last but not least. It seems that Translator hot path (critical functionality) is SPIR-V translation to IR and SPT is used only as a format for testing which renders it much colder code path from the user perspective. This indicates that potential performance works should be performed in SPIR-V code path which much less consideration of SPT. From that point of view split is necessary to unlock full potential of hot path optimizations.

Thanks Bertrand for detailed answer. It is a good point to have best possible performance of the translator. Thinking more on this, I also suppose we can get rid of SPT flow at some point of time. We do have spirv-as and spirv-dis to go between binary and text formats.

asudarsa (approved):
Overall looks great. Thanks Bertrand. I have no objection.

PR #2317, opened 2024-01-25
Add support for SPV_INTEL_subgroup_requirements

jzc:
Spec: https://github.com/intel/llvm/pull/11301

More accurately, this PR adds support for the named subgroup related features of SPV_INTEL_subgroup_requirements to support implementation of [sycl_ext_named_sub_group_sizes](https://github.com/intel/llvm/blob/sycl/sycl/doc/extensions/proposed/sycl_ext_oneapi_named_sub_group_sizes.asciidoc) (also see https://github.com/intel/llvm/pull/12335). The features related to subgroup lane mapping are not added yet.


lib/SPIRV/PreprocessMetadata.cpp
```diff
@@ -167,11 +167,21 @@ void PreprocessMetadataBase::visit(Module *M) {
 
     // !{void (i32 addrspace(1)*)* @kernel, i32 35, i32 size}
     if (MDNode *ReqdSubgroupSize = Kernel.getMetadata(kSPIR2MD::SubgroupSize)) {
-      EM.addOp()
+      // A primary named subgroup size is encoded as 
+      // the metadata intel_reqd_sub_group_size with value 0.
+      auto val = getMDOperandAsInt(ReqdSubgroupSize, 0);
```
AlexeySachkov, line 172: `val` -> `Val`

lib/SPIRV/PreprocessMetadata.cpp
```diff
@@ -167,11 +167,21 @@ void PreprocessMetadataBase::visit(Module *M) {
...
+      if (val == 0)
+        EM.addOp()
           .add(&Kernel)
-          .add(spv::ExecutionModeSubgroupSize)
-          .add(getMDOperandAsInt(ReqdSubgroupSize, 0))
+          .add(spv::internal::ExecutionModeNamedSubgroupSizeINTEL)
+          .add(0U)
```
AlexeySachkov, line 177:
```suggestion
          .add(/* PrimarySubgroupSizeINTEL = */ 0U)
```

The comment could be spelled differently, but I think it makes sense to explain this magic zero in here

lib/SPIRV/PreprocessMetadata.cpp
```diff
@@ -167,11 +167,21 @@ void PreprocessMetadataBase::visit(Module *M) {
 
     // !{void (i32 addrspace(1)*)* @kernel, i32 35, i32 size}
     if (MDNode *ReqdSubgroupSize = Kernel.getMetadata(kSPIR2MD::SubgroupSize)) {
-      EM.addOp()
+      // A primary named subgroup size is encoded as
```
AlexeySachkov, line 170: Strictly speaking, we are changing default behavior here. Right now the translator propagates sub-group size 0 as-is, but with the change applied it will only happen if the extension is enabled.

I would maybe suggest that we preserve the original behavior to avoid accidentally breaking someone's else workflow, i.e. we should only react to `0` if the extension is enabled even at this step, where we only emit some internal metadata.

MrSidims:
Restarting CI with a new HEAD

lib/SPIRV/SPIRVWriter.cpp
```diff
@@ -5717,6 +5717,13 @@ bool LLVMToSPIRVBase::transExecutionMode() {
           BF->addExecutionMode(BM->add(
               new SPIRVExecutionMode(BF, static_cast<ExecutionMode>(EMode))));
       } break;
+      case spv::internal::ExecutionModeNamedSubgroupSizeINTEL: {
+        if (!BM->isAllowedToUseExtension(ExtensionID::SPV_INTEL_subgroup_requirements))
+          break;
+        BM->addExtension(ExtensionID::SPV_INTEL_subgroup_requirements);
+        BM->addCapability(spv::internal::CapabilitySubgroupRequirementsINTEL);
```
MrSidims, line 5724: It seem like we have inconsistency in our code base in terms of adding capabilities and extensions. Yet I'd recommend to modify lib/SPIRV/libSPIRV/SPIRVEnum.h to add mapping between execution mode and the capability. Optionally, please consider modifying getRequiredExtension from SPIRVCapability class (lib/SPIRV/libSPIRV/SPIRVEntry.h) to add mapping between the extension and capability, but it's some sort of a grey area, is it seems like we have almost never used it.

lib/SPIRV/PreprocessMetadata.cpp
jzc: I changed the behavior and added a test case: https://github.com/KhronosGroup/SPIRV-LLVM-Translator/pull/2317/commits/193c25d6f6b9f7f10c8b10deb7057a0050a3b1b7

Note this code is preprocessing the LLVM module and adding extra execution metadata to that module, so I changed this code to add this metadata in addition to the regular sub group metadata (as opposed to instead). Then in `SPIRVWriter`, sub group execution mode metadata will be present and be processed like before and the named sub group metadata will be ignored when the extension is turned off. https://github.com/KhronosGroup/SPIRV-LLVM-Translator/pull/2317/files#diff-ba003da77a80df1db76fb378a873a80e758591d72bf82a46fc90069b2a947990R5838-R5843

PR #2273, opened 2023-12-18
[SPV -> SPV-IR] Fix image builtin mangling of unsigned type

wenju-he:
Return type of image read and Texel type of image write builtins may be
unsigned. Before this PR, the builtin names in SPIR-V Friendly IR were
always mangled with signed type.

PR #2327, opened 2024-01-30
[Bckport to 14][SPV -> SPV-IR] Fix image builtin mangling of unsigned type (#2273)

wenju-he:
Return type of image read and Texel type of image write builtins may be unsigned. Before this PR, the builtin names in SPIR-V Friendly IR were always mangled with signed type.

(cherry picked from commit e9b95fbc3f58c7d8084620fd58b8bd936c640d37)


MrSidims (approved):
ideally we should also do cherry-picks to newer branches, like llvm_17,16,15 in this case.

wenju-he:
> ideally we should also do cherry-picks to newer branches, like llvm_17,16,15 in this case.

@MrSidims I've just cherry-picked to 17,16,15 branches.
I've encountered a stability issue that sext is causing signed overflow when texel type is unsigned, so the fix is necessary.

PR #2335, opened 2024-01-31
Fix BufferLocationINTEL decoration translation

MrSidims:
There was an assumption, that ptr.annotation encoding buffer_location should be used by load or store instructions. But there is no such restriction in the specification (there were such checks in `checkMemUser` used by `allowDecorateWithBufferLocationOrLatencyControlINTEL`).

This fix also required adjustments in the reader part as traversing dominator tree and replacing old instructions with ptr.annotation won't work for for return basic blocks. So the patch refactors this part causally fixing an issue with multiple structure annotations (instead of analysis in `replaceOperandWithAnnotationIntrinsicCallResult` - we should already update a Value map with pairing ptr.annotation with previously decorated SPIR-V instructions), exposed by a change in test/transcoding/multiple_user_semantic_on_struct.ll test. Ideally it should be separated into 2 different PRs, but I beg your pardon for merging these 2 changes together.

### DIFF
--- a/lib/SPIRV/SPIRVReader.cpp
+++ b/lib/SPIRV/SPIRVReader.cpp
@@ -3639,7 +3639,7 @@ void SPIRVToLLVM::transIntelFPGADecorations(SPIRVValue *BV, Value *V) {
       for (SPIRVWord I = 0; I < STS->getMemberCount(); ++I) {
         std::vector<SmallString<256>> AnnotStrVec;
         generateIntelFPGAAnnotationForStructMember(ST, I, AnnotStrVec);
-        [[maybe_unused]] CallInst *AnnotationCall = nullptr;
+        CallInst *AnnotationCall = nullptr;
         for (const auto &AnnotStr : AnnotStrVec) {
           auto *GS = Builder.CreateGlobalStringPtr(AnnotStr);
 
@@ -3684,7 +3684,7 @@ void SPIRVToLLVM::transIntelFPGADecorations(SPIRVValue *BV, Value *V) {
 
     std::vector<SmallString<256>> AnnotStrVec;
     generateIntelFPGAAnnotation(BV, AnnotStrVec);
-    [[maybe_unused]] CallInst *AnnotationCall = nullptr;
+    CallInst *AnnotationCall = nullptr;
     for (const auto &AnnotStr : AnnotStrVec) {
       Constant *GS = nullptr;
       const auto StringAnnotStr = static_cast<std::string>(AnnotStr);
--- a/lib/SPIRV/SPIRVReader.cpp
+++ b/lib/SPIRV/SPIRVReader.cpp
@@ -1093,66 +1093,6 @@ static void applyFPFastMathModeDecorations(const SPIRVValue *BV,
   }
 }
 
-namespace {
-
-// A pointer annotation may have been generated for the operand. If the operand
-// is used further in IR, it should be replaced with the intrinsic call result.
-// Otherwise, the generated pointer annotation call is left unused.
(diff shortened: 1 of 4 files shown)


MrSidims:
@wangdi4 please take a look

wangdi4:
Thank you for working on this! Just a few concerns:

- It seems in the updated function `transIntelFPGADecorations` (in SPIRVReader.cpp), all the decorations are converted to `llvm.ptr.annotation`s (which replace the uses of the pointer being annotated). I wonder if there is any case that is expected to be converted to metadata?
- Also in the case of multiple decoration on the same pointer, is it expected that each decoration is translated to a separate `llvm.ptr.annotation` call, rather than one `llvm.ptr.annotation` call that contains all the decorations? 
- Please add the test case for GEP in the test `sycl-buffer-location-with-ptr-annotation.ll` so that all the possible paths affected by this changes are covered.

MrSidims:
> * It seems in the updated function `transIntelFPGADecorations` (in SPIRVReader.cpp), all the decorations are converted to `llvm.ptr.annotation`s (which replace the uses of the pointer being annotated). I wonder if there is any case that is expected to be converted to metadata?

spirv.Decorations metadata can be generated if one pass --spirv-target-env=SPV-IR during reverse translation. This mechanism is tested elsewhere (for example in spirv_global_variable_decoration.ll), so I didn't add it in this test. However, it seems that it is a bit bugged for my taste - I've tried sycl-buffer-location-with-ptr-annotation.ll , everything works, but I do see both ptr.annotation and spirv.Decoration metadata when --spirv-target-env=SPV-IR is passed. While IMO it should be just metadata. Not sure if we want to fix it in this patch, its unlikely to be a critical issue.

> * Also in the case of multiple decoration on the same pointer, is it expected that each decoration is translated to a separate `llvm.ptr.annotation` call, rather than one `llvm.ptr.annotation` call that contains all the decorations?

Per https://github.com/KhronosGroup/SPIRV-LLVM-Translator/blob/main/docs/SPIRVRepresentationInLLVM.rst#member-decoration-through-pointer-annotations we can added several MemberDecorations via this mechanism. As for now I don't see this contract covering regular decorations, so encoding Alignment or something like this is not expected to work.

> * Please add the test case for GEP in the test `sycl-buffer-location-with-ptr-annotation.ll` so that all the possible paths affected by this changes are covered.

Not sure, if I understand correctly, what do you mean. There are several other tests doing that. Please also note changes in test/transcoding/multiple_user_semantic_on_struct.ll , where an existing bug was fixed.

wangdi4:
> > * Please add the test case for GEP in the test `sycl-buffer-location-with-ptr-annotation.ll` so that all the possible paths affected by this changes are covered.
> 
> Not sure, if I understand correctly, what do you mean. There are several other tests doing that. Please also note changes in test/transcoding/multiple_user_semantic_on_struct.ll , where an existing bug was fixed.

I mean the case of
```
%x = call %llvm.ptr.annotation(%ptr, ...)
return %x
```
combined with GEP (as the function `transIntelFPGADecorations` contains the parts of handling pointers of struct types and scalar types, both of which are changed in the PR)

MrSidims:
@wangdi4 please take another look. There are many other possible test cases, like usage of annotation of a memcpy or other function parameters etc. But these cases are covered by Brox and myself for example in test/extensions/INTEL/SPV_INTEL_fpga_buffer_location/FPGABufferLocation.ll and I'm not changing these cases. Since the tests are passing and not being modyfied, then change of logic doesn't affect fixes for previously discovered and fixed bugs.

wangdi4:
The PR looks good to me. Thank you!

PR #2339, opened 2024-02-02
Struct as result type in OpGroupNonUniformShuffleDown translation fix

bwlodarcz:
The contract between some frontends and Translator is that calls to functions, 
which are matching to hardcoded name, are converted to matching Ops in SPIR-V. 
The translation from LLVM to SPIR-V of OpGroupNonUniformShuffleDown call was done incorrectly.
The struct member which is correct result type for the op was left unwrapped which lead to
SPIR-V nonconforming with specification.

### DIFF
--- a/lib/SPIRV/SPIRVWriter.cpp
+++ b/lib/SPIRV/SPIRVWriter.cpp
@@ -6345,12 +6345,12 @@ LLVMToSPIRVBase::transBuiltinToInstWithoutDecoration(Op OC, CallInst *CI,
   }
   case OpGroupNonUniformShuffleDown: {
     Function *F = CI->getCalledFunction();
-    Type *LResTy = F->getParamStructRetType(0);
-    SPIRVValue *InValue =
-        transValue(CI->getArgOperand(0)->stripPointerCasts(), BB);
-    SPIRVId ScopeId = transValue(CI->getArgOperand(1), BB)->getId();
-    SPIRVValue *Delta = transValue(CI->getArgOperand(3), BB);
-    if (StructType *St = dyn_cast<StructType>(LResTy)) {
+    if (F->arg_size() && F->getArg(0)->hasStructRetAttr()) {
+      StructType *St = cast<StructType>(F->getParamStructRetType(0));
+      SPIRVValue *InValue =
+          transValue(CI->getArgOperand(0)->stripPointerCasts(), BB);
+      SPIRVId ScopeId = transValue(CI->getArgOperand(1), BB)->getId();
+      SPIRVValue *Delta = transValue(CI->getArgOperand(3), BB);
       SPIRVValue *Composite0 = BM->addLoadInst(InValue, {}, BB);
       Type *MemberTy = St->getElementType(0);
       SPIRVType *ElementTy = transType(MemberTy);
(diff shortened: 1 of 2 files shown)


lib/SPIRV/SPIRVWriter.cpp
```diff
@@ -6287,6 +6287,28 @@ LLVMToSPIRVBase::transBuiltinToInstWithoutDecoration(Op OC, CallInst *CI,
     return BM->addCompositeConstructInst(transType(CI->getType()), Operands,
                                          BB);
   }
+  case OpGroupNonUniformShuffleDown: {
```
MrSidims, line 6290: as they probably share the same issue
```suggestion
  case OpGroupNonUniformShuffle:
  case OpGroupNonUniformShuffleXor:
  case OpGroupNonUniformShuffleUp:
  case OpGroupNonUniformShuffleDown: {
```

lib/SPIRV/SPIRVWriter.cpp
```diff
@@ -6287,6 +6287,28 @@ LLVMToSPIRVBase::transBuiltinToInstWithoutDecoration(Op OC, CallInst *CI,
...
+    Function *F = CI->getCalledFunction();
+    if (F->arg_size() && F->getArg(0)->hasStructRetAttr()) {
+      StructType *St = cast<StructType>(F->getParamStructRetType(0));
```
MrSidims, line 6293: In general it's SYCL details, that `half` (and also `bfloat16`) are hidden within classes. There should be checks, like `isSYCLHalfType` around in the translator, should we add them here to error out, if just a random structure is passed to the instruction?

lib/SPIRV/SPIRVWriter.cpp
```diff
@@ -6287,6 +6287,28 @@ LLVMToSPIRVBase::transBuiltinToInstWithoutDecoration(Op OC, CallInst *CI,
...
+      Type *MemberTy = St->getElementType(0);
+      SPIRVType *ElementTy = transType(MemberTy);
+      SPIRVValue *Element0 =
+          BM->addCompositeExtractInst(ElementTy, Composite0, {0}, BB);
+      SPIRVValue *Src =
+          BM->addGroupInst(OpGroupNonUniformShuffleDown, ElementTy,
+                           static_cast<Scope>(ScopeId), {Element0, Delta}, BB);
+      SPIRVValue *Composite1 =
+          BM->addCompositeInsertInst(Src, Composite0, {0}, BB);
+      return BM->addStoreInst(InValue, Composite1, {}, BB);
```
MrSidims, line 6308: Ideally this block of code terminating with 'add store' should be placed in a separate function. Because we probably have more problems we sret'ed half structures, so this code could be reused instead of copy pasted.

MrSidims (commented):
Generated SPIR-V and LLVM IR LGTM, just few nits

lib/SPIRV/SPIRVWriter.cpp
```diff
@@ -6343,6 +6343,30 @@ LLVMToSPIRVBase::transBuiltinToInstWithoutDecoration(Op OC, CallInst *CI,
     return BM->addCompositeConstructInst(transType(CI->getType()), Operands,
                                          BB);
   }
+  case OpGroupNonUniformShuffleDown: {
+    Function *F = CI->getCalledFunction();
+    if (F->arg_size() && F->getArg(0)->hasStructRetAttr()) {
+      StructType *St = cast<StructType>(F->getParamStructRetType(0));
+      assert(St->getElementType(0)->isHalfTy() ||
+             St->getElementType(0)->isBFloatTy());
```
MrSidims, line 6351:
```suggestion
      assert(isSYCLHalfType(St) || isSYCLBfloat16Type(St));
```

MrSidims (approved):

lib/SPIRV/SPIRVWriter.cpp
MrSidims: Agreed to fix later

PR #2340, opened 2024-02-02
Implement SPV_INTEL_task_sequence extension

vmaksimo:
Spec: https://github.com/KhronosGroup/SPIRV-Registry/pull/192


vmaksimo:

lib/SPIRV/libSPIRV/SPIRVInstruction.h
```diff
@@ -3854,5 +3858,85 @@ template <Op OC> class SPIRVReadClockKHRInstBase : public SPIRVUnaryInst<OC> {
...
+    SPIRVType *ResTy = this->getType();
+    SPVErrLog.checkError(
+        ResTy->isTypeTaskSequenceINTEL(), SPIRVEC_InvalidInstruction,
+        InstName + "\nResult must be TaskSequenceINTEL type\n");
+
+    SPIRVValue *Func =
+        const_cast<SPIRVTaskSequenceCreateINTELInst *>(this)->getOperand(0);
+    SPVErrLog.checkError(
+        Func->getOpCode() == OpFunction, SPIRVEC_InvalidInstruction,
+        InstName + "\nFirst argument is expected to be a function.\n");
```
VyacheslavLevytskyy, line 3891: Could it be OpConstantFunctionPointerINTEL if SPV_INTEL_function_pointers is on?

test/extensions/INTEL/SPV_INTEL_task_sequence/negative/invalid_cluster_mode.ll
```diff
@@ -0,0 +1,59 @@
+; RUN: llvm-as %s -o %t.bc
```
VyacheslavLevytskyy, line 1: Would it make sense to mention in comments source C++ code if any was used to produce ll tests, just for reference purpose, so that it's easier to understand test logic?

lib/SPIRV/libSPIRV/SPIRVInstruction.h
vmaksimo: Enabling of this extension doesn't make `ptr @_Z4multii` to be translated to `OpConstantFunctionPointerINTEL`.
And if it would, we should have discussed with the specification writers if this is an acceptable argument (which now is true only for `OpFunction`)

test/extensions/INTEL/SPV_INTEL_task_sequence/negative/invalid_cluster_mode.ll
vmaksimo: Not sure if it's needed for the negative tests which are the small reduced versions of the `task_sequence.ll` test. I guess I can add the C++ example to the original test though.

lib/SPIRV/libSPIRV/SPIRVInstruction.h
VyacheslavLevytskyy: Thank you

test/extensions/INTEL/SPV_INTEL_task_sequence/negative/invalid_cluster_mode.ll
VyacheslavLevytskyy: Great, thank you

VyacheslavLevytskyy (approved):

lib/SPIRV/SPIRVUtil.cpp
```diff
@@ -1361,7 +1361,8 @@ static SPIR::RefParamType transTypeDesc(Type *Ty,
     auto *ET = TPT->getElementType();
     SPIR::ParamType *EPT = nullptr;
     if (isa<FunctionType>(ET)) {
-      assert(isVoidFuncTy(cast<FunctionType>(ET)) && "Not supported");
+      // TODO: implement support for non void function type
+      // assert(isVoidFuncTy(cast<FunctionType>(ET)) && "Not supported");
```
MrSidims, line 1365: We have discussed it offline, but I still don't understand why we need to comment this assert and what are the consequences. While the PR is approved, I'd suggest to hold merge a bit until every one understands, what is happening here :)

MrSidims (approved):

PR #2341, opened 2024-02-05
[DebugInfo] Fix SPIR-V consumption of DebugInfoNone for debug types

MrSidims:
OpenCL and NonSemantic DebugInfo specifications are flexible in terms of allowing any debug information be replaced with DebugInfoNone, so various of SPIR-V producers follow that and generate it for base types of several debug instructions, leaving SPIR-V consumers to handle this. By default the translator replaces missing debug info with *tag*: null, which is in most cases correct. Yet, there are situations, where it's not allowed by both LLVM and DWARF, for example for DW_TAG_array_type DWARF spec sets, that DW_AT_type attribute is mandatory. For such cases new `transNonNullDebugType` wrapper function was added to the translator, generating "DIBasicType(tag: DW_TAG_unspecified_type, name: "SPIRV unknown type")" where DebugInfoNone was used as the type. This function doesn't replace all calls to `transDebugInst<DIType>` as there are cases, where we can generate null type, for example DWARF doesn't require it for DW_TAG_typedef, hence I'm not changing translation flow in this case. Additionally to this, while DWARF requires type attribute for DW_TAG_pointer_type, LLVM does not, hence I'm not changing translation flow in this case as well.


test/DebugInfo/DebugInfoGlobalVariableNone.spvasm
```diff
@@ -0,0 +1,54 @@
+; Tests translation of GlobalVariable with DebugInfoNone type
+
+; REQUIRES: spirv-as
+
+; RUN: spirv-as --target-env spv1.1 -o %t.spv
```
vmaksimo, line 5:
```suggestion
; RUN: spirv-as %s --target-env spv1.1 -o %t.spv
```

MrSidims:
@vmaksimo  @asudarsa @LU-JOHN @bwlodarcz @VyacheslavLevytskyy please take a look

vmaksimo (commented):
LGTM overall, would be nice to restructure the commit message a bit (optional: to clarify which DI tags are expected to have the type).
Also, is there a plan to cover all the changes with the appropriate tests, or does it look a bit overkill?

lib/SPIRV/SPIRVToLLVMDbgTran.cpp
```diff
@@ -1412,6 +1410,13 @@ MDNode *SPIRVToLLVMDbgTran::transExpression(const SPIRVExtInst *DebugInst) {
   return getDIBuilder(DebugInst).createExpression(Addr);
 }
 
+DIType *
```
asudarsa, line 1413: Just curious. why do we need a separate function here? Can we just not return the unknown type inside the 'case DebugInfoNone' construct inside transDebugInstImpl?

Just a nit.

Thanks

MrSidims: Because `transDebugInstImpl` handles every debug instruction, not only types. And we can't even specialize it for <DIType> for the reasons I mentioned in the description:
`This function doesn't replace all calls to transDebugInst<DIType> as there are cases, where we can generate null type, for example DWARF doesn't require it for DW_TAG_typedef, hence I'm not changing translation flow in this case. Additionally to this, while DWARF requires type attribute for DW_TAG_pointer_type, LLVM does not, hence I'm not changing translation flow in this case as well.`

MrSidims:
Added test for inheritance and local variable written using nonsemantic debug info. It also appears to be, that LLVM can handle Type inheritance with null parent on its own without crashing/erroring out, so remove it's handling.

lib/SPIRV/SPIRVToLLVMDbgTran.cpp
```diff
@@ -479,7 +479,7 @@ SPIRVToLLVMDbgTran::transTypeVector(const SPIRVExtInst *DebugInst) {
   const SPIRVWordVec &Ops = DebugInst->getArguments();
   assert(Ops.size() >= MinOperandCount && "Invalid number of operands");
   DIType *BaseTy =
-      transDebugInst<DIType>(BM->get<SPIRVExtInst>(Ops[BaseTypeIdx]));
+      transNonNullDebugType(BM->get<SPIRVExtInst>(Ops[BaseTypeIdx]));
```
vmaksimo, line 482: I can't find if this case is covered with the test now

lib/SPIRV/SPIRVToLLVMDbgTran.cpp
```diff
@@ -853,9 +851,9 @@ SPIRVToLLVMDbgTran::transTypePtrToMember(const SPIRVExtInst *DebugInst) {
   const SPIRVWordVec &Ops = DebugInst->getArguments();
   assert(Ops.size() >= OperandCount && "Invalid number of operands");
   SPIRVExtInst *Member = BM->get<SPIRVExtInst>(Ops[MemberTypeIdx]);
-  DIType *PointeeTy = transDebugInst<DIType>(Member);
+  DIType *PointeeTy = transNonNullDebugType(Member);
```
vmaksimo, line 854: For these too

test/DebugInfo/NonSemantic/Shader200/FortranArrayNoType.spt
```diff
@@ -0,0 +1,73 @@
+; RUN: llvm-spirv -to-binary %s -o %t.spv
```
LU-JOHN, line 1: Add intro comment describing the purpose of this test, similar to the other testcases.

lib/SPIRV/SPIRVToLLVMDbgTran.cpp
asudarsa: Ah. Got it. Thanks for pointing back to the description.

test/DebugInfo/NonSemantic/Shader200/FortranArrayNoType.spt
MrSidims: Applied

lib/SPIRV/SPIRVToLLVMDbgTran.cpp
MrSidims: Added

lib/SPIRV/SPIRVToLLVMDbgTran.cpp
MrSidims: Added

lib/SPIRV/SPIRVToLLVMDbgTran.cpp
```diff
@@ -1224,11 +1224,11 @@ DINode *SPIRVToLLVMDbgTran::transTypeInheritance(const SPIRVExtInst *DebugInst,
   // No Child operand for NonSemantic debug spec
   SPIRVWord Offset = isNonSemanticDebugInfo(DebugInst->getExtSetKind()) ? 1 : 0;
   DIType *Parent =
-      transNonNullDebugType(BM->get<SPIRVExtInst>(Ops[ParentIdx - Offset]));
+      transDebugInst<DIType>(BM->get<SPIRVExtInst>(Ops[ParentIdx - Offset]));
```
LU-JOHN: Should we add a testcase that requires the reversions on lines 1227 and 1231?  

Also, how is "SPIRV unknown type" generated as expected in DebugInfoTypeInheritance.spvasm with these reversions?

MrSidims: I have removed special handling of DebugInfoTypeInheritance with DebugInfoNonde. Reason is that LLVM actually can handle absence of parent on its own, just by not emitting the appropriate metadata. It's tested here: https://github.com/KhronosGroup/SPIRV-LLVM-Translator/pull/2341/files#diff-60e39e2de73507f9b785b3994221453dc593d76f50be7f3355d2766f5524d2fdR13

LU-JOHN (approved):
LGTM

PR #2344, opened 2024-02-06
Support SPV_INTEL_maximum_registers extension

vmaksimo:
Spec:
https://github.com/KhronosGroup/SPIRV-Registry/pull/235/


vmaksimo:

vmaksimo:
> LGTM conditionally if tokens are placed in https://github.com/KhronosGroup/SPIRV-Headers

Created PR to Headers, will update the translator code once Headers are merged (can be done in a separate PR) https://github.com/KhronosGroup/SPIRV-Headers/pull/416

lib/SPIRV/SPIRVReader.cpp
```diff
@@ -4318,6 +4318,50 @@ bool SPIRVToLLVM::transMetadata() {
       F->setMetadata(kSPIR2MD::IntelFPGAIPInterface,
                      MDNode::get(*Context, InterfaceMDVec));
     }
+    if (auto *EM = BF->getExecutionMode(
```
LU-JOHN, line 4321: Should we check to ensure that only one mode is chosen?

We should use "else if" for the second and third cases.

Consider refactoring common code.

lib/SPIRV/SPIRVReader.cpp
```diff
@@ -4318,6 +4318,50 @@ bool SPIRVToLLVM::transMetadata() {
...
+
+      SmallVector<Metadata *, 4> ValueVec;
+      ValueVec.push_back(ConstantAsMetadata::get(F));
+      ValueVec.push_back(
+          ConstantAsMetadata::get(getUInt32(M, EM->getExecutionMode())));
+      ValueVec.push_back(
+          ConstantAsMetadata::get(getUInt32(M, EM->getLiterals()[0])));
+      ExecModeMD->addOperand(MDNode::get(*Context, ValueVec));
+    }
+    if (auto *EM = BF->getExecutionMode(
```
LU-JOHN, line 4334: This should be "else if"

lib/SPIRV/SPIRVWriter.cpp
```diff
@@ -977,7 +977,10 @@ SPIRVFunction *LLVMToSPIRVBase::transFunctionDecl(Function *F) {
 
   transFPGAFunctionMetadata(BF, F);
 
-  transFunctionMetadataAsUserSemanticDecoration(BF, F);
+  if (BM->isAllowedToUseExtension(ExtensionID::SPV_INTEL_maximum_registers))
+    transFunctionMetadataAsExecutionMode(BF, F);
+  else
+    transFunctionMetadataAsUserSemanticDecoration(BF, F);
```
asudarsa, line 983: Do we want to support the 'old' implementation or throw an error here? Sorry if this is ignorant question.

lib/SPIRV/SPIRVWriter.cpp
```diff
@@ -1118,6 +1121,38 @@ void LLVMToSPIRVBase::transFPGAFunctionMetadata(SPIRVFunction *BF,
     transMetadataDecorations(FDecoMD, BF);
 }
 
+void LLVMToSPIRVBase::transFunctionMetadataAsExecutionMode(SPIRVFunction *BF,
+                                                           Function *F) {
+  if (!BM->isAllowedToUseExtension(ExtensionID::SPV_INTEL_maximum_registers))
```
asudarsa, line 1126: Do we not expect this function to be reused for other extensions like this in the future? This check seems redundant.

Thanks

lib/SPIRV/SPIRVWriter.cpp
```diff
@@ -1118,6 +1121,38 @@ void LLVMToSPIRVBase::transFPGAFunctionMetadata(SPIRVFunction *BF,
...
+      auto *RegisterAllocNodeMDOp =
+          getMDOperandAsMDNode(RegisterAllocModeMD, 0);
+      int Num = getMDOperandAsInt(RegisterAllocNodeMDOp, 0);
+      auto *Const =
+          BM->addConstant(transType(Type::getInt32Ty(F->getContext())), Num);
+      BF->addExecutionMode(BM->add(new SPIRVExecutionMode(
+          OpExecutionModeId, BF, internal::ExecutionModeMaximumRegistersIdINTEL,
+          Const->getId())));
+    } else {
+      int64_t RegisterAllocNodeMDOp =
```
asudarsa, line 1147: Can we please rename this to differentiate with variable used in Line 1138?
May be RegisterAllocModeId?

asudarsa (commented):
LGTM. Some minor comments. Thanks

lib/SPIRV/SPIRVWriter.cpp
vmaksimo: Indeed redundant, will remove

lib/SPIRV/SPIRVReader.cpp
vmaksimo: Actually no. Even so it seems to have no sense to enable several different `maxregisters` execution modes, neither SPIR-V core spec, nor extension prohibits usage of several execution modes at the same time. So, I believe this code should not be changed to `else if`.

LU-JOHN: From https://github.com/KhronosGroup/SPIRV-Registry/pull/235, in extensions/INTEL/SPV_INTEL_maximum_registers.asciidoc

* Each *OpEntryPoint* must contain at most one of the
*MaximumRegistersINTEL*, *MaximumRegistersIdINTEL*, or
*NamedMaximumRegistersINTEL* execution modes.

vmaksimo: Applied in 84a3ead. Unfortunately I can't add meaningful tests for that, because llvm-spirv fails with an assertion in case we add more than one execution mode of the certain type. 
Verified locally that this check both on forward and reverse translation in case of invalid IR/SPV.

PR #2345, opened 2024-02-06
Implement lowering of llvm.bitreverse.*

LU-JOHN:
If SPV_KHR_bit_instructions is not enabled lower llvm.bitreverse.* to a function in LLVM IR.


lib/SPIRV/LLVMBitreverse.h
```diff
@@ -0,0 +1,148 @@
+//===- LLVMBitreverse.h - implementation of llvm.bitreverse -===//
+//
+//                     The LLVM/SPIRV Translator
+//
+// This file is distributed under the University of Illinois Open Source
+// License. See LICENSE.TXT for details.
+//
+// Copyright (c) 2024 Intel Corporation. All rights reserved.
```
MrSidims, line 8: Please use the following license: https://github.com/KhronosGroup/SPIRV-LLVM-Translator/issues/80#issuecomment-476781085

LU-JOHN: License has been udpated.

lib/SPIRV/LLVMBitreverse.h
```diff
@@ -0,0 +1,134 @@
...
+//       ret|=1;
+//     a>>=1;
+//   }
+//   return ret;
+// }
+//
+// Clang options: -emit-llvm -O1 -g0 -fno-discard-value-names
+// -O1 is used instead of -O2 to avoid generating a call to llvm.bitreverse.i16
+
+static const char LLVMBitreverse[]{R"(
```
MrSidims, line 74: That is an interesting approach which is worth exploring, but without tablegen it won't work.
1. There are many more integers, then i16/i32/i64;
2. The intrinsic function works not only with scalars, but with vectors as well.

MrSidims:
It also would be nice to figure out passing --spirv-ext option to the pass, as the intrinsic should be mapped on KHR extension if it's enabled. Otherwise community, who uses KHR extension would see a change of the behavior of the translator in this case.

LU-JOHN:
> It also would be nice to figure out passing --spirv-ext option to the pass, as the intrinsic should be mapped on KHR extension if it's enabled. Otherwise community, who uses KHR extension would see a change of the behavior of the translator in this case.

The pass only maps an intrinsic if a supporting extension is not enabled:

https://github.com/KhronosGroup/SPIRV-LLVM-Translator/blob/27da693af8a1966ce7c6c1e00986a900b3a91597/lib/SPIRV/SPIRVLowerLLVMIntrinsic.cpp#L107C1-L107C9

lib/SPIRV/SPIRVLowerLLVMIntrinsic.cpp
```diff
@@ -0,0 +1,207 @@
...
+    return;
+
+  Function *IntrinsicFunc = I.getCalledFunction();
+  assert(IntrinsicFunc && "Missing function");
+  StringRef IntrinsicName = IntrinsicFunc->getName();
+
+  for (const LLVMIntrinsicMapEntryType &LLVMIntrinsicMapEntry :
+       LLVMIntrinsicMapEntries) {
+    if (II->getIntrinsicID() == LLVMIntrinsicMapEntry.ID &&
+        IntrinsicName == LLVMIntrinsicMapEntry.LLVMFuncName &&
```
svenvh, line 118: There might be a performance issue here.  If I understand correctly, for any intrinsic other than `llvm.bitreverse` and `llvm.sadd.with.overflow`, the pass now has to iterate over all `LLVMIntrinsicMapEntries` entries.

LU-JOHN: Updated to use a std::map

lib/SPIRV/LLVMBitreverse.h
LU-JOHN: Added support for (i8/i16/i32/i64)x(scalar,vec-2,vec-4).

LU-JOHN:
Please take a look @MrSidims @asudarsa @bwlodarcz @VyacheslavLevytskyy @vmaksimo @maksimsab

VyacheslavLevytskyy (approved):

lib/SPIRV/SPIRVLowerLLVMIntrinsic.cpp
```diff
@@ -0,0 +1,195 @@
+//===- SPIRVLowerLLVMIntrinsic.cpp - Lower llvm-intrinsics -----===//
```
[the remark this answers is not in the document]
LU-JOHN: Copyright updated.

lib/SPIRV/SPIRVLowerLLVMIntrinsic.cpp
```diff
@@ -0,0 +1,195 @@
...
+
+  { "llvm.bitreverse.v2i8",        {ExtensionID::SPV_KHR_bit_instructions, "llvm_bitreverse_v2i8",        LLVMBitreverseV2}},
+  { "llvm.bitreverse.v2i16",       {ExtensionID::SPV_KHR_bit_instructions, "llvm_bitreverse_v2i16",       LLVMBitreverseV2}},
+  { "llvm.bitreverse.v2i32",       {ExtensionID::SPV_KHR_bit_instructions, "llvm_bitreverse_v2i32",       LLVMBitreverseV2}},
+  { "llvm.bitreverse.v2i64",       {ExtensionID::SPV_KHR_bit_instructions, "llvm_bitreverse_v2i64",       LLVMBitreverseV2}},
+
+  { "llvm.bitreverse.v4i8",        {ExtensionID::SPV_KHR_bit_instructions, "llvm_bitreverse_v4i8",        LLVMBitreverseV4}},
+  { "llvm.bitreverse.v4i16",       {ExtensionID::SPV_KHR_bit_instructions, "llvm_bitreverse_v4i16",       LLVMBitreverseV4}},
+  { "llvm.bitreverse.v4i32",       {ExtensionID::SPV_KHR_bit_instructions, "llvm_bitreverse_v4i32",       LLVMBitreverseV4}},
+  { "llvm.bitreverse.v4i64",       {ExtensionID::SPV_KHR_bit_instructions, "llvm_bitreverse_v4i64",       LLVMBitreverseV4}},
```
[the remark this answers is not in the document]
LU-JOHN: Support added for 3,8,16-element vectors.

svenvh (approved):

PR #2347, opened 2024-02-12
[NFC] Fix DebugTypeVector test

MrSidims:
It should have tested DebugInfoNone base type


MrSidims:
My mistake, shouldn't have merge the patch without this fix - temporary CI failure has blinded me to not see this test not passing

LU-JOHN (commented):
LGTM

PR #2348, opened 2024-02-12
map to unordered_map for SPIRVIdToEntryMap type

bwlodarcz:
In cpp std::map type is a red-black tree implementation. It's search time is O(log N) and 
pointers are scattered through memory. 
In comparison unordered_map is a hash table which doesn't have the same iteration guarantees 
as red-black tree but it's search time is O(1) plus the constants like calculating hash function and 
load from the table.
The SPIRVModuleImpl::SPIRVIdToEntryMap is used in SPIRVModuleImpl::getEntry function which is widely used
as a key component of the parser.
It's a small fix but yields around 30% speedup for translation SPIR-V to IR.


MrSidims (approved):

MrSidims:
CI failures are unrelated and fixed

PR #2356, opened 2024-02-12
[NFC] Use unordered_map for better performance

LU-JOHN:
Use unordered_map instead of map for better performance.

### DIFF
--- a/lib/SPIRV/Mangler/Mangler.cpp
+++ b/lib/SPIRV/Mangler/Mangler.cpp
@@ -66,7 +66,7 @@ public:
           ThistypeStr << NType;
       }
 #endif
-      std::map<std::string, unsigned>::iterator I =
+      std::unordered_map<std::string, unsigned>::iterator I =
           Substitutions.find(ThistypeStr.str());
       if (I == Substitutions.end())
         return false;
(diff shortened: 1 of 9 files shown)


MrSidims:
I'm expecting, that lots of test would need to be updated, as order of IDs wouldn't remain after it. We would also need to check, if new order of IDs makes sense.
Please also note, that while map to hash table replacement seems legit for {int, int} or {ptr, ptr} pairs, it's less obvious, that it will bring performance benefits for {string, int} pairs.

PS DebugInfo failures are related to https://github.com/KhronosGroup/SPIRV-LLVM-Translator/issues/2357 . You may want to incorporate https://github.com/KhronosGroup/SPIRV-LLVM-Translator/pull/2362 for testing

LU-JOHN:
> I'm expecting, that lots of test would need to be updated, as order of IDs wouldn't remain after it. We would also need to check, if new order of IDs makes sense. Please also note, that while map to hash table replacement seems legit for {int, int} or {ptr, ptr} pairs, it's less obvious, that it will bring performance benefits for {string, int} pairs.
> 
> PS DebugInfo failures are related to #2357 . You may want to incorporate #2362 for testing

I've incorporated #2362.  I think I've avoided changing maps that will affect the output order.

LU-JOHN:
Please take a look @asudarsa @bwlodarcz @MrSidims

MrSidims (approved):
Tentatively approving, but would be nice to have some numbers for performance improvements.
While in most cases it's LGTM, yet it's less obvious, that it improves std::string key cases.

bwlodarcz:
@MrSidims It's valid improvement on algorithmic level but won't be visible in performance metrics due that other things (getEntry chain) are dominating execution time. Still, the gain is here.

svenvh (approved):

Issue #2357, opened 2024-02-13
[DebugInfo] Support non-instrinsic debug-info


jsji:
llvm community has been trying to enable non-instrinsic debug-info.

 * 5c9f7682b090 - (2024-02-12) Reapply "[DebugInfo][RemoveDIs] Turn on non-instrinsic debug-info by default" - Stephen Tozer
 
We should enable translation of the new debug-info and update tests accordingly.

Note all DebugInfo tests and complex-constexpr-vector.ll are failing.

Ref:
[0] https://llvm.org/docs/RemoveDIsDebugInfo.html
[1] https://discourse.llvm.org/t/rfc-instruction-api-changes-needed-to-eliminate-debug-intrinsics-from-ir/68939

PR #2362, opened 2024-02-13
[NFCI] Prepare for non-instrinsic debug info

MrSidims:
For now just convert BB with convertFromNewDbgValues, will figure out something smarter a bit later.

I've updated several tests with dbg.declare intrinsic adding --experimental-debuginfo-iterators=1 to check if it works.

Resolved https://github.com/KhronosGroup/SPIRV-LLVM-Translator/issues/2357 for now

### DIFF
--- a/lib/SPIRV/SPIRVWriter.cpp
+++ b/lib/SPIRV/SPIRVWriter.cpp
@@ -5525,6 +5525,7 @@ void LLVMToSPIRVBase::transFunction(Function *I) {
   // is used to ensure blocks are written in the right order.
   const DominatorTree DT(*I);
   for (BasicBlock &FI : stablePreDominatorTraversal(*I, DT)) {
+    FI.convertFromNewDbgValues();
     transValue(&FI, nullptr);
   }
   for (auto &FI : *I) {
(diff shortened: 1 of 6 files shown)


MrSidims:
FYI @jsji

asudarsa (approved):
Looks good for unblocking our testing. But we will need to handle new debug info. Also, we might need to figure out how to handle this in the reverse translator.
Thanks @MrSidims for a very quick resolution here.

MrSidims:
> But we will need to handle new debug info.

Well, basically we need to remove dbg_decl/var handling by having a case in intrinsic switch and replace it with something. AFAIU there are 2 options:
1. Check associated debug info with alloca's to obtain dbg_decl/var, but it didn't work for me with a single attempt I made there;
2. Smart iteration over basic blocks, as now gdb_decl/var like described here: https://discourse.llvm.org/t/rfc-instruction-api-changes-needed-to-eliminate-debug-intrinsics-from-ir/68939/12 . That is slightly more work, as we would need to rewrite a little bit of code here.

> Also, we might need to figure out how to handle this in the reverse translator.

Nothing to do there.

svenvh (approved):
LGTM, thanks for working on this!

PR #2372, opened 2024-02-19
Report unmet version requirement as error

svenvh:
When restricting the allowed SPIR-V version using `--spirv-max-version`, some constructs may not be translatable due to being unavailable in older SPIR-V versions.  Instead of asserting, report an error.

It would be nice to also report the construct leading to the version conflict, but that information isn't readily available at this point.


svenvh:
Not planning to address the clang-tidy issue in this case, as I think the name `to_string` aligns best with `std::to_string`.

MrSidims (approved):

MrSidims:
@VyacheslavLevytskyy please take a look

VyacheslavLevytskyy (approved):

PR #2378, opened 2024-02-22
[SPIR-V 1.5] Fix OpGroupNonUniformBroadcast version requirement

svenvh:
For SPIR-V 1.4 and earlier, the `Id` operand of `OpGroupNonUniformBroadcast` must come from a constant instruction.

Ensure that any emitted SPIR-V module using a non-constant `Id` operand declares SPIR-V version 1.5.  Bail out if the maximum SPIR-V version was restricted to < 1.5, to avoid producing invalid SPIR-V.

### DIFF
--- a/lib/SPIRV/libSPIRV/SPIRVInstruction.h
+++ b/lib/SPIRV/libSPIRV/SPIRVInstruction.h
@@ -276,6 +276,9 @@ public:
     } else
       SPIRVEntry::setWordCount(WC);
     Ops = TheOps;
+    // The required SPIR-V version depends on the operands for some
+    // instructions.
+    updateModuleVersion();
   }
   void setWordCount(SPIRVWord TheWordCount) override {
     SPIRVEntry::setWordCount(TheWordCount);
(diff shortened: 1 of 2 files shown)


vmaksimo (approved):

lib/SPIRV/libSPIRV/SPIRVInstruction.h
```diff
@@ -295,6 +298,11 @@ class SPIRVInstTemplateBase : public SPIRVInstruction {
 
   /// Get operand as value.
   /// If the operand is a literal, return it as a uint32 constant.
+  const SPIRVValue *getOpValue(int I) const {
+    return isOperandLiteral(I) ? Module->getLiteralAsConstant(Ops[I])
```
MrSidims, line 302: Necessity of adding these lines confuses me, I interpret it as you have attempted to translate:
sub_group_non_uniform_broadcast(a, /*constexpr int*/ 42)
and got
OpGroupNonUniformBroadcast ..... /*id*/ 42 /*literal*/

which is quite surprising, as the instruction is declared as _SPIRV_OP(GroupNonUniformBroadcast, true, 6) (aka has 0 literals, unlike for example _SPIRV_OP(GroupNonUniformBallotBitCount, true, 6, false, 1) which has a literal in a position of one).

MrSidims: Ah, that's it, thanks!

MrSidims (approved):

PR #2392, opened 2024-03-01
Add IAddCarry and ISubBorrow builtins

bwlodarcz:
The fix adds support for IR builtin calls
__spirv_IAddCarry and __spirv_ISubBorrow.
It's also first part of fix which removes noncompliance of uadd/sub_with_overflow intrinsics.
SPIRVUtil changes are needed to support situations where builtin don't have corresponding store instruction.

### DIFF
--- a/lib/SPIRV/SPIRVReader.cpp
+++ b/lib/SPIRV/SPIRVReader.cpp
@@ -2480,20 +2480,12 @@ Value *SPIRVToLLVM::transValueWithoutDecoration(SPIRVValue *BV, Function *F,
     return mapValue(BV,
                     transRelational(static_cast<SPIRVInstruction *>(BV), BB));
   case OpIAddCarry: {
-    IRBuilder Builder(BB);
     auto *BC = static_cast<SPIRVBinary *>(BV);
-    return mapValue(BV, Builder.CreateBinaryIntrinsic(
-                            Intrinsic::uadd_with_overflow,
-                            transValue(BC->getOperand(0), F, BB),
-                            transValue(BC->getOperand(1), F, BB)));
+    return mapValue(BV, transBuiltinFromInst("__spirv_IAddCarry", BC, BB));    
   }
   case OpISubBorrow: {
-    IRBuilder Builder(BB);
     auto *BC = static_cast<SPIRVBinary *>(BV);
-    return mapValue(BV, Builder.CreateBinaryIntrinsic(
-                            Intrinsic::usub_with_overflow,
-                            transValue(BC->getOperand(0), F, BB),
-                            transValue(BC->getOperand(1), F, BB)));
+    return mapValue(BV, transBuiltinFromInst("__spirv_ISubBorrow", BC, BB));    
   }
   case OpGetKernelWorkGroupSize:
   case OpGetKernelPreferredWorkGroupSizeMultiple:
--- a/lib/SPIRV/SPIRVReader.cpp
+++ b/lib/SPIRV/SPIRVReader.cpp
@@ -2481,11 +2481,11 @@ Value *SPIRVToLLVM::transValueWithoutDecoration(SPIRVValue *BV, Function *F,
                     transRelational(static_cast<SPIRVInstruction *>(BV), BB));
   case OpIAddCarry: {
     auto *BC = static_cast<SPIRVBinary *>(BV);
-    return mapValue(BV, transBuiltinFromInst("__spirv_IAddCarry", BC, BB));    
+    return mapValue(BV, transBuiltinFromInst("__spirv_IAddCarry", BC, BB));
   }
   case OpISubBorrow: {
     auto *BC = static_cast<SPIRVBinary *>(BV);
-    return mapValue(BV, transBuiltinFromInst("__spirv_ISubBorrow", BC, BB));    
+    return mapValue(BV, transBuiltinFromInst("__spirv_ISubBorrow", BC, BB));
   }
   case OpGetKernelWorkGroupSize:
   case OpGetKernelPreferredWorkGroupSizeMultiple:
--- a/lib/SPIRV/SPIRVUtil.cpp
+++ b/lib/SPIRV/SPIRVUtil.cpp
@@ -2186,7 +2186,18 @@ bool postProcessBuiltinReturningStruct(Function *F) {
                                      F->getReturnType());
       NewF->addParamAttr(0, SretAttr);
       NewF->setCallingConv(F->getCallingConv());
-      AllocaInst *A = Builder.CreateAlloca(F->getReturnType());
+      SmallVector<User *> Users(CI->users());
+      Value *A = nullptr;
+      for (auto *U : Users) {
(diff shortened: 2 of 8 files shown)


lib/SPIRV/SPIRVWriter.cpp
```diff
@@ -6383,6 +6384,20 @@ LLVMToSPIRVBase::transBuiltinToInstWithoutDecoration(Op OC, CallInst *CI,
     return BM->addCompositeConstructInst(transType(CI->getType()), Operands,
                                          BB);
   }
+  case OpIAddCarry: {
+    Function *F = CI->getCalledFunction();
+    StructType *St = cast<StructType>(F->getParamStructRetType(0));
+    return BM->addBinaryInst(OpIAddCarry, transType(St),
```
MrSidims, line 6390: Like this it will be a dangling OpIAddCarry, we need to store the result to the original sret pointer

test/iaddcarry_builtin.ll
```diff
@@ -0,0 +1,102 @@
...
+  entry:
+  %0 = alloca {i8, i8}
+  call void @_Z17__spirv_IAddCarrycc(ptr sret ({i8, i8}) %0, i8 %a, i8 %b)
+  ret void
+}
+; CHECK-SPIRV:           [[a:%[a-z0-9_]+]] = OpFunctionParameter [[uchar]]
+; CHECK-SPIRV:           [[b:%[a-z0-9_]+]] = OpFunctionParameter [[uchar]]
+; CHECK-SPIRV:       [[entry:%[a-z0-9_]+]] = OpLabel
+; CHECK-SPIRV:      [[var_11:%[a-z0-9_]+]] = OpVariable [[_ptr_Function__struct_9]] Function
+; CHECK-SPIRV:      [[var_12:%[a-z0-9_]+]] = OpIAddCarry [[_struct_9]] [[a]] [[b]]
```
MrSidims, line 37: Like here OpVariable and OpIAddCarry are disconnected

lib/SPIRV/SPIRVReader.cpp
```diff
@@ -2480,20 +2480,12 @@ Value *SPIRVToLLVM::transValueWithoutDecoration(SPIRVValue *BV, Function *F,
     return mapValue(BV,
                     transRelational(static_cast<SPIRVInstruction *>(BV), BB));
   case OpIAddCarry: {
-    IRBuilder Builder(BB);
     auto *BC = static_cast<SPIRVBinary *>(BV);
-    return mapValue(BV, Builder.CreateBinaryIntrinsic(
-                            Intrinsic::uadd_with_overflow,
-                            transValue(BC->getOperand(0), F, BB),
-                            transValue(BC->getOperand(1), F, BB)));
+    return mapValue(BV, transBuiltinFromInst("__spirv_IAddCarry", BC, BB));
```
MrSidims, line 2484: Remove `auto *BC = static_cast<SPIRVBinary *>(BV);` and
```suggestion
    return mapValue(BV, transSPIRVBuiltinFromInst(
                           static_cast<SPIRVInstruction *>(BV), BB));
```

lib/SPIRV/SPIRVUtil.cpp
```diff
@@ -2185,12 +2186,16 @@ bool postProcessBuiltinReturningStruct(Function *F) {
                                      F->getReturnType());
       NewF->addParamAttr(0, SretAttr);
       NewF->setCallingConv(F->getCallingConv());
+      AllocaInst *A = Builder.CreateAlloca(F->getReturnType());
```
MrSidims, line 2189: But we should already have the appropriate alloca, haven't we? Like in the test case:
```
  %0 = alloca {i8, i8} // we should reuse it, otherwise generated code would be on the dead path
  call void @_Z17__spirv_IAddCarrycc(ptr sret ({i8, i8}) %0, i8 %a, i8 %b)
  ret void
```
we would need to create it only in a case, when we have no alloca in the IR before the translator, aka when you would try to translate the intrinsic.

test/isubborrow_builtin.ll
```diff
@@ -0,0 +1,102 @@
+; RUN: llvm-as %s -o %t.bc
+; RUN: llvm-spirv %t.bc -o %t.spv
+; RUN: spirv-dis --raw-id %t.spv | FileCheck --check-prefix CHECK-SPIRV %s
+; RUN: spirv-val %t.spv
+; RUN: llvm-spirv -r %t.spv -o %t.rev.bc
+; llvm-dis %t.rev.bc -o - | FileCheck --check-prefix CHECK-LLVM %s
```
MrSidims, line 6: unused

lib/SPIRV/SPIRVReader.cpp
bwlodarcz: Doesn't `transSPIRVBuiltinFromInst` calls `transBuiltinFromInst` inside after going though switch clause and few ifs which, in that case, are never taken?

MrSidims (approved):

lib/SPIRV/SPIRVUtil.cpp
```diff
@@ -2185,12 +2186,27 @@ bool postProcessBuiltinReturningStruct(Function *F) {
...
+        }
+      }
+      if (!A) {
+        A = Builder.CreateAlloca(F->getReturnType());
+      }
       auto Args = getArguments(CI);
-      Args.insert(Args.begin(), ST->getPointerOperand());
-      auto *NewCI = CallInst::Create(NewF, Args, CI->getName(), CI);
+      Args.insert(Args.begin(), A);
+      CallInst *NewCI = Builder.CreateCall(NewF, Args, CI->getName());
```
MrSidims, line 2203: Wait a sec. NewF here is void with sret, meanwhile Args were set for a previous instance of CI aka without sret parameter. That is incorrect.

MrSidims (changes requested):

lib/SPIRV/SPIRVUtil.cpp
bwlodarcz: The problem here isn't about sret parameter. In llvm IR type system `addrspace` is an integral part of the type so `ptr != ptr addrspace(x)`. Indeed, this code had an error which is now fixed.

lib/SPIRV/SPIRVWriter.cpp
```diff
@@ -65,6 +65,7 @@
 #include "SPIRVUtil.h"
 #include "SPIRVValue.h"
 #include "VectorComputeUtil.h"
+#include "spirv/unified1/spirv.hpp"
```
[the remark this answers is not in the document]
bwlodarcz: My editor inferred that and I overlooked it. Why it's default stance in the editor to do that?
There is a popular 'good practice' to avoid transitive headers inclusion. The principle is that if we directly using type/function/class from header than it should be included. The goal of this practice is to control how much headers are actually in use and in result decrease compile time.
I removed an inclusion.

MrSidims (approved):

svenvh (approved):

PR #2399, opened 2024-03-04
[Backport to 13] Support SPV_INTEL_maximum_registers extension

KorovinVlad:


vmaksimo (commented):
Could you please take a look at clang-tidy errors? 
I worry only about the recursion - I can't see where it is from the review, but maybe the code is too different from LLVM-14 branch. I guess the problem should be somehow resolved before merging

KorovinVlad:
llvm14 translator and above have the following commit: https://github.com/KhronosGroup/SPIRV-LLVM-Translator/commit/6a1cfe9a55253b2913e8d58b5b56df941959a741
I can cherry-pick it to suppress clang-tidy error.

PR #2418, opened 2024-03-07
Fix translation of __spirv_TaskSequenceGetINTEL

vmaksimo:
This change adds support for void return-type of Get instruction and handles the case when Get instruction has 'sret' parameter.

### DIFF
--- a/lib/SPIRV/SPIRVWriter.cpp
+++ b/lib/SPIRV/SPIRVWriter.cpp
@@ -6358,6 +6358,25 @@ LLVMToSPIRVBase::transBuiltinToInstWithoutDecoration(Op OC, CallInst *CI,
     return BM->addStoreInst(transValue(CI->getArgOperand(0), BB), APIntInst, {},
                             BB);
   }
+  case internal::OpTaskSequenceGetINTEL: {
+    Type *ResTy = nullptr;
+    auto OpItr = CI->value_op_begin();
+
+    if (CI->hasStructRetAttr()) {
+      assert(CI->getType()->isVoidTy() && "Return type is not void");
+      ResTy = CI->getParamStructRetType(0);
+      OpItr++;
+    }
+
+    SPIRVType *RetTy = ResTy ? transType(ResTy) : transScavengedType(CI);
+    auto *TaskSeqGet =
+        BM->addTaskSequenceGetINTELInst(RetTy, transValue(*OpItr++, BB), BB);
+
+    if (!CI->hasStructRetAttr())
+      return TaskSeqGet;
+    return BM->addStoreInst(transValue(CI->getArgOperand(0), BB), TaskSeqGet,
+                            {}, BB);
+  }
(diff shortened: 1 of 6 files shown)


vmaksimo:
Please take a look @MrSidims @LU-JOHN @bwlodarcz @VyacheslavLevytskyy

test/extensions/INTEL/SPV_INTEL_task_sequence/sret-get.ll
VyacheslavLevytskyy, line 1: Would it make sense to add source code that was used to generate llvm ir, here and in other test cases, to facilitate reading of test cases logic?

vmaksimo: Unfortunately I don't have source code for these new tests, the "faulty cases" were provided as LLVM IR

VyacheslavLevytskyy (approved):

lib/SPIRV/SPIRVWriter.cpp
```diff
@@ -6053,7 +6053,7 @@ LLVMToSPIRVBase::transBuiltinToInstWithoutDecoration(Op OC, CallInst *CI,
     OC = OpAtomicCompareExchange;
   if (isGroupOpCode(OC))
     BM->addCapability(CapabilityGroups);
-  switch (OC) {
+  switch (static_cast<size_t>(OC)) {
```
LU-JOHN, line 6056: Why is this static_cast necessary?

test/extensions/INTEL/SPV_INTEL_task_sequence/sret-get.ll
```diff
@@ -0,0 +1,88 @@
+; RUN: llvm-as %s -o %t.bc
```
LU-JOHN, line 1: It would be helpful to have a comment at the beginning of the test explaining the purpose.  Something like:

// Test translation of __spirv_TaskSequenceGetIntel with a sret (structure return) parameter.

Otherwise the reader will have to look at the commit history or try to derive the purpose from the filename and CHECK lines.

test/extensions/INTEL/SPV_INTEL_task_sequence/void-get.ll
```diff
@@ -0,0 +1,78 @@
+; RUN: llvm-as %s -o %t.bc
```
LU-JOHN, line 1: A purpose comment would be helpful here too.

lib/SPIRV/SPIRVWriter.cpp
vmaksimo: Because of the added `case internal::` - such OpCodes are not a part of `spv::Op` enum, so this trick helps to fix the build warning. It's used not only here but in all switch cases with internal opcodes in SPIRV writer part.

MrSidims (approved):

PR #2419, opened 2024-03-07
Regularization of uadd/usub_with_overflow intrinsics

bwlodarcz:
The translation of uadd/usub with overflow were problematic. The returning type of these IR intrinsics are tuples with a second i1 type. These intrinsics were translated to OpIAddCarry and OpISubBorrow which, according to standard, should also return a two element tuple but in that case both types should be the same. IR i1 type is translated by default to BoolType which doesn't have the same properties as i1. There isn't also a possibility to exceptionally override the default behavior in current architecture.
Finally the translation was implemented but without compliance with SPIR-V spec - the resulting type of two ops contained BoolType as second element.
This commit fixes non compliant translation. Instrinsics uadd/usub_with_overflow are now translated to sequence of
operations which repacks compliant return type into expected IR type with second element as BoolType.

E.g for i32 uadd_with_overflow:
```
%0 = alloca {i32, i32}, align 8
call spir_func void @_Z17__spirv_IAddCarryii(ptr sret({i32, i32}) %0, i32 %a, i32 %b)
%1 = load {i32, i32}, ptr %0, align 4
%2 = extractvalue {i32, i32} %1, 0
%3 = extractvalue {i32, i32} %1, 1
%4 = icmp ne i32 %3, 0
%5 = insertvalue {i32, i1} undef, i32 %2, 0
%6 = insertvalue {i32, i1} %5, i1 %4, 1
```
All uses of original intrinsic value are replaced with the repacked result (%6 above) which fixes type mismatch.


lib/SPIRV/SPIRVRegularizeLLVM.cpp
```diff
@@ -335,6 +335,62 @@ bool SPIRVRegularizeLLVMBase::runRegularizeLLVM(Module &Module) {
   return true;
 }
 
+namespace {
+void regularizeWithOverflowInstrinsics(StringRef MangledName, CallInst *Call,
```
MrSidims, line 339: I have a feeling, that this can be done easier and cheaper.
Lets imagine we have a case:
```
%res = call {i32, i1} @llvm.uadd.with.overflow.i32(i32 %a, i32 %b)
call void foo({i32, i1} %res)
```
Yes, extracting and extending i1 and storing them to a new structure will work and semantically correct. But LLVM nowadays after the switch to opaque pointers provides us strong and effective mechanisms for bit manipulations for stack allocations. And I wounder if the snippet below is semantically equivalent to the result of the work of this function:
```
%res_alloca = alloca { i32, i32 }
call spir_func void @_Z17__spirv_IAddCarryii(ptr sret(i32, i32) %res_alloca, i32 %a, i32 %b)
%res = load { i32, i1 } %res_alloca
call void foo({i32, i1} %res)
```
?

bwlodarcz: Nice one. Yes it'll work when the second member of result can have 1 or 0 result - which is fulfilled in this case.

lib/SPIRV/SPIRVRegularizeLLVM.cpp
```diff
@@ -335,6 +335,52 @@ bool SPIRVRegularizeLLVMBase::runRegularizeLLVM(Module &Module) {
...
+                                       Module *M,
+                                       std::vector<Instruction *> &ToErase) {
+  IRBuilder Builder(Call);
+  Function *Builtin = Call->getModule()->getFunction(MangledName);
+  AllocaInst *A;
+  if (Builtin) {
+    StructType *StructBuiltinTy =
+        cast<StructType>(Builtin->getParamStructRetType(0));
+    A = Builder.CreateAlloca(Builtin->getParamStructRetType(0));
```
MrSidims, line 348: (not-clang friendly suggesting)
```suggestion
    {
      IRBuilderBase::InsertPointGuard Guard(Builder);
      Builder.SetInsertPointPastAllocas(Call->getParent()->getParent());
      A = Builder.CreateAlloca(Builtin->getParamStructRetType(0));
    }
 ```
 Reason:
If the intrinsic appears in not-the-first basic block of the function the alloca will be created in this basic block, resulting in OpVariable created not-the-first basic block (and not-necessarily in the beginning of it). And this violates:
`All OpVariable instructions in a function must be in the first block in the function. These instructions, together with any intermixed OpLine and OpNoLine instructions, must be the first instructions in that block. (Note the validation rules prevent OpPhi instructions in the first block of a function.)`

I suggest to insert the alloca in the beginning of the function right away as it's cheap to do in the pass.
Same applicable to the create alloca below.

**Review by MrSidims — APPROVED** [review]
LGTM with one nitpick.
Would be nice if @asudarsa @vmaksimo @LU-JOHN @VyacheslavLevytskyy and @svenvh also take a look.

File: lib/SPIRV/SPIRVRegularizeLLVM.cpp (hunk 2 above)
Comment (MrSidims) [reply lib/SPIRV/SPIRVRegularizeLLVM.cpp:348]: Follows a notion from https://github.com/KhronosGroup/SPIRV-LLVM-Translator/pull/2439

File: lib/SPIRV/SPIRVRegularizeLLVM.cpp (hunk 1 above)
Comment (MrSidims) [reply lib/SPIRV/SPIRVRegularizeLLVM.cpp:339]: I unresolve it just for other to double check the logic

File: test/llvm-intrinsics/uadd.with.overflow.ll [hunk 3]
```diff
@@ -1,51 +1,170 @@
...
+; CHECK-LLVM:   %1 = bitcast ptr %0 to ptr
+; CHECK-LLVM:   %2 = load [[structtype_6]], ptr %1, align 16
+; CHECK-LLVM:   %3 = extractvalue [[structtype_6]] %2, 0
+; CHECK-LLVM:   %4 = extractvalue [[structtype_6]] %2, 1
+; CHECK-LLVM:   ret <4 x i1> %4
 declare {i16, i1} @llvm.uadd.with.overflow.i16(i16 %a, i16 %b)
 declare {i32, i1} @llvm.uadd.with.overflow.i32(i32 %a, i32 %b)
 declare {i64, i1} @llvm.uadd.with.overflow.i64(i64 %a, i64 %b)
 declare {<4 x i32>, <4 x i1>} @llvm.uadd.with.overflow.v4i32(<4 x i32> %a, <4 x i32> %b)
+declare void @_Z17__spirv_IAddCarryii(ptr sret({i32, i32}), i32, i32)
```
vmaksimo, line 170: Why this should be declared? I can't see it's used anywhere in the test. Same for usub test.

bwlodarcz: In regularization procedure we have two branches which condition on the existence of the function.
This line is added to cover one of the branches -  situation when somewhere in code direct builtin with __spirv_IAddCarryii is used. 
The line: 
```
  Function *Builtin = Call->getModule()->getFunction(MangledName);
  AllocaInst *A;
  if (Builtin) {
```
to trigger `true` branch.

vmaksimo: Okay, just a bit strange that if we remove the declaration the test will fail (or not?).
Maybe we need to have tests to cover both if-branches.

lib/SPIRV/SPIRVRegularizeLLVM.cpp
bwlodarcz: After discussion, we established that this change is unfortunately, although elegant, is non valid. The cast here it's UB in C and C++ due to type incompatibility. The source of UB here is that because value here is struct, for different types it's alignment is changeable. That means that compiler can insert hidden fields between two members of the struct and in runtime value obtained from such cast can be undefined.

MrSidims (approved):

PR #2431, opened 2024-03-12
Fix globals' type deduction in TypeScavenger

mshelego:
If a global is initialized with a ConstantExpr returning a pointer (getelementptr), the right type has to be deduced instead of defaulting it to i8*.


MrSidims (approved):
LGTM, thanks!

vmaksimo (approved):

svenvh (approved):

PR #2437, opened 2024-03-14
Set WordCount for SPIRVVariable correctly

LU-JOHN:
SPIRVVariable should have a WordCount of 4 + 1 (if optional Initializer is added).
Set WordCount to 4 if Initializer is not added.

### DIFF
--- a/lib/SPIRV/libSPIRV/SPIRVInstruction.h
+++ b/lib/SPIRV/libSPIRV/SPIRVInstruction.h
@@ -455,16 +455,11 @@ public:
                 const std::string &TheName,
                 SPIRVStorageClassKind TheStorageClass, SPIRVBasicBlock *TheBB,
                 SPIRVModule *TheM)
-      : SPIRVInstruction(TheInitializer ? 5 : 4, OpVariable, TheType, TheId,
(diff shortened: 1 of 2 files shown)


lib/SPIRV/libSPIRV/SPIRVInstruction.h
```diff
@@ -459,7 +459,12 @@ class SPIRVVariable : public SPIRVInstruction {
                          TheBB, TheM),
         StorageClass(TheStorageClass) {
     if (TheInitializer && !TheInitializer->isUndef())
+      // SPIRVVariable has optional Initializer field
+      // WordCount is already 5 by default
       Initializer.push_back(TheInitializer->getId());
+    else
+      // No optional Initializer, so set WordCount to 4
+      setWordCount(4);
```
bader, line 467: I think we should update the condition on line 458 instead.

```c++
SPIRVInstruction((TheInitializer && !TheInitializer->isUndef()) ? 5 : 4, OpVariable, TheType, TheId,
```

MrSidims:
Restarting CI

MrSidims (approved):

PR #2438, opened 2024-03-18
[CI] Upgrade to actions/checkout@v4

svenvh:
Node.js 16 actions are [deprecated](https://github.blog/changelog/2023-09-22-github-actions-transitioning-from-node-16-to-node-20/), so update to v4 which uses Node 20.


MrSidims (approved):
Thanks!

PR #2439, opened 2024-03-18
Always insert `OpVariable` instructions at the beginning of a block

victor-eds:
According to the SPIR-V specification:

> All `OpVariable` instructions in a function must be in the first block in the function. These instructions, together with any intermixed `OpLine` and `OpNoLine` instructions, must be the first instructions in that block. (Note the validation rules prevent `OpPhi` instructions in the first block of a function.)

As forcing `OpVariable` insertion at the beginning of the first block might be too invasive by changing original semantics, inserting them at the beginning of the argument block would preserve semantics while always producing valid SPIR-V code.

`llvm.memmove` lowering has been fixed to always generate valid SPIR-V code by inserting needed `alloca` instructions at the beginning of the function's entry block.


test/constexpr_vector.ll
```diff
@@ -73,8 +73,8 @@ entry:
...
...oint (ptr @_Z2f1u2CMvb32_j to i64) to <8 x i8>), i32 4), i8 extractelement (<8 x i8> bitcast (i64 ptrtoint (ptr @_Z2f1u2CMvb32_j to i64) to <8 x i8>), i32 5), i8 extractelement (<8 x i8> bitcast (i64 ptrtoint (ptr @_Z2f1u2CMvb32_j to i64) to <8 x i8>), i32 6), i8 extractelement (<8 x i8> bitcast (i64 ptrtoint (ptr @_Z2f1u2CMvb32_j to i64) to <8 x i8>), i32 7), i8 extractelement (<8 x i8> bitcast (i64 ptrtoint (ptr @_Z2f2u2CMvb32_j to i64) to <8 x i8>), i32 0), i8 extractelement (<8 x i8> bitcast (i64 ptrtoint (ptr @_Z2f2u2CMvb32_j to i64) to <8 x i8>), i32 1), i8 extractelement (<8 x i8> bitcast (i64 ptrtoint (ptr @_Z2f2u2CMvb32_j to i64) to <8 x i8>), i32 2), i8 extractelement (<8 x i8> bitcast (i64 ptrtoint (ptr @_Z2f2u2CMvb32_j to i64) to <8 x i8>), i32 3), i8 extractelement (<8 x i8> bitcast (i64 ptrtoint (ptr @_Z2f2u2CMvb32_j to i64) to <8 x i8>), i32 4), i8 extractelement (<8 x i8> bitcast (i64 ptrtoint (ptr @_Z2f2u2CMvb32_j to i64) to <8 x i8>), i32 5), i8 extractelement (<8 x i8> bitcast (i64 ptrtoint (ptr @_Z2f2u2CMvb32_j to i64) to <8 x i8>), i32 6), i8 extractelement (<8 x i8> bitcast (i64 ptrtoint (ptr @_Z2f2u2CMvb32_j to i64) to <8 x i8>), i32 7)>, ptr %Funcs, align 16
```
victor-eds, line 76: Same line, just moved so that all `alloca`s are at the beginning of the block

lib/SPIRV/libSPIRV/SPIRVBasicBlock.cpp
```diff
@@ -87,6 +87,17 @@ void SPIRVBasicBlock::encodeChildren(spv_ostream &O) const {
 
 _SPIRV_IMP_ENCDEC1(SPIRVBasicBlock, Id)
 
+SPIRVInstruction *SPIRVBasicBlock::getVariableInsertionPoint() const {
```
MrSidims, line 90: If I understand the code in SPIRVBasicBlock.cpp/h correctly, this function will set insert point in the beginning on the current basic block, but not the first basic block of the function. Regardless if I'm right or not, may I ask you to add a test showing, that even if alloca is in the second basic block, then OpVariable will still be added in the first one?

MrSidims (commented):
That is definitely an improvement, thanks!

I have added a comment which is kind of a nitpick, and I probably ready to merge it as is.

lib/SPIRV/libSPIRV/SPIRVBasicBlock.cpp
victor-eds: This function will return the instruction corresponding to the first instruction not being `OpVariable`, `OpLine` or `OpNoLine`, i.e., where the next `OpVariable` should be inserted. In the case you mention, insertion will happen at that given point of the second block. Moving it to the first block may change original LLVM IR semantics, so I'd rather keep generating invalid code in that case as the input LLVM IR would be ill formed from a SPIR-V perspective. See the following case:

```llvm
loop_body:
  %ptr = alloca i32
  store ptr %ptr, ptr %ptrs
  ...
```

Representing the `alloca` instruction there as a single `OpVariable` would change semantics. Generating invalid SPIR-V as the original code is not representable as is in SPIR-V is what we were doing before and what we will keep doing after this patch. I do not intend to fix the general case, just the cases in which the `alloca` instructions are in the first block of a function, but not at the beginning of it.

Naghasan: > Moving it to the first block may change original LLVM IR semantics

how actually ? the memory is uninitialised, the only thing to check is to make sure the alloca is not in a statically unreachable BB.

MrSidims: > Representing the alloca instruction there as a single OpVariable would change semantics

IMHO no, it will only break instruction-level parallelism unless there is an IR analysis taking in account such cases (and I believe there should be something like this to unlock ILP in cases like this one https://godbolt.org/z/7c89dzra1

I'm not insisting to fix this case, the patch is already is a big improvement, just asking if we can/want to fix it now as well (I'm unsure what would be the effort)

svenvh (approved):

lib/SPIRV/libSPIRV/SPIRVBasicBlock.cpp
victor-eds: I was thinking of something like: https://godbolt.org/z/zfe1oq8nx. Moving the variable to the first block would effectively change the semantics to something like: https://godbolt.org/z/heK7To9P7.

Notice the original LLVM module in the first example, with an `alloca` instruction not in the first block, is already not representable in SPIR-V.

MrSidims (approved):

Naghasan (approved):

lib/SPIRV/libSPIRV/SPIRVBasicBlock.cpp
```diff
@@ -87,6 +87,17 @@ void SPIRVBasicBlock::encodeChildren(spv_ostream &O) const {
...
+  auto IP =
+      std::find_if(InstVec.begin(), InstVec.end(), [](SPIRVInstruction *Inst) {
```
MrSidims, line 92: @victor-eds there is an issue with the patch, in situations when we have phi instruction and alloca within the same basic block - OpVariable must not be placed before OpPhi. May I ask you to fix that?

victor-eds: That would be illegal SPIR-V anyway, but I can issue a fix.

MrSidims: Why? The SPIR-V spec says:
```
Within a block, this instruction must appear before all non-OpPhi instructions
(except for OpLine and OpNoLine, which can be mixed with OpPhi).
```

karolherbst: I wonder if this is an overlooked edge case and needs a spec fix. Normally with structured spirv you'll never run into a situation where you'll an `OpVariable` and an `OpPhi` instruction in the same block, but with OpenCL SPIR-V this can happen in the very first one.

PR #2443, opened 2024-03-19
Fix handling of OpenCL convert_ builtins

vmaksimo:
This change fixes cases when the input function name does not match OpenCL spec but starts with `convert_` prefix, e.g. `convert_float_42`.

### DIFF
--- a/lib/SPIRV/OCLToSPIRV.cpp
+++ b/lib/SPIRV/OCLToSPIRV.cpp
@@ -751,9 +751,9 @@ void OCLToSPIRVBase::visitCallConvert(CallInst *CI, StringRef MangledName,
     TargetTyName = TargetTyName.substr(0, FirstUnderscoreLoc);
 
   // Validate target type name
-  std::regex e("([a-z]+)([0-9]*)$");
+  std::regex Expr("([a-z]+)([0-9]*)$");
   std::smatch DestTyMatch;
-  if (!std::regex_match(TargetTyName, DestTyMatch, e))
+  if (!std::regex_match(TargetTyName, DestTyMatch, Expr))
     return;
 
   // The first sub_match is the whole string; the next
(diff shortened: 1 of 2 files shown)


vmaksimo:
Please take a look @MrSidims @LU-JOHN @asudarsa @bwlodarcz @VyacheslavLevytskyy

lib/SPIRV/OCLToSPIRV.cpp
```diff
@@ -734,16 +742,63 @@ void OCLToSPIRVBase::visitCallConvert(CallInst *CI, StringRef MangledName,
...
+  // The first sub_match is the whole string; the next
+  // sub_match is the first parenthesized expression.
+  std::string DestTy = DestTyMatch[1].str();
+
+  // check it's valid type name
+  SmallVector<StringRef, 11> ValidTypes = {"float", "double", "half",   "char",
+                                           "uchar", "short",  "ushort", "int",
+                                           "uint",  "long",   "ulong"};
+  bool IsValidType = false;
+  for (auto I : ValidTypes) {
```
VyacheslavLevytskyy, line 768: Probably this can be done faster if we make `ValidTypes` a static set of strings (e.g., `static std::unordered_set<std::string> ValidTypes = {...}`) that initializes just once and then search in it using smth like `if (auto It = ValidTypes.find(I); It != ValidTypes.end())`.

MrSidims:
Restarting CI with a new head

VyacheslavLevytskyy (approved):

lib/SPIRV/OCLToSPIRV.cpp
```diff
@@ -734,16 +742,56 @@ void OCLToSPIRVBase::visitCallConvert(CallInst *CI, StringRef MangledName,
...
+  // The first sub_match is the whole string; the next
+  // sub_match is the first parenthesized expression.
+  std::string DestTy = DestTyMatch[1].str();
+
+  // check it's valid type name
+  static std::unordered_set<std::string> ValidTypes = {
+      "float",  "double", "half", "char", "uchar", "short",
+      "ushort", "int",    "uint", "long", "ulong"};
+
+  if (auto It = ValidTypes.find(DestTy); It == ValidTypes.end())
```
maarquitos14, line 768: Do we need `auto It` here? If we do, can it be `const auto &it` instead? I think `if (ValidTypes.find(DestTy) == ValidTypes.end())` is good enough.

maarquitos14 (approved):

PR #2456, opened 2024-03-22
Make storage class of builtins to be Input

asudarsa:
In case of builtin variables and their users, the storage class function should be StorageClassInput, irrespective of the incoming addrspace.
Ref: https://registry.khronos.org/OpenCL/specs/3.0-unified/html/OpenCL_Env.html
Sec 2.9. Built-in variables.

This fixes https://github.com/intel/llvm/issues/6703

Thanks
Sincerely

### DIFF
--- a/lib/SPIRV/SPIRVReader.cpp
+++ b/lib/SPIRV/SPIRVReader.cpp
@@ -1553,8 +1553,11 @@ Value *SPIRVToLLVM::transValueWithoutDecoration(SPIRVValue *BV, Function *F,
     // Force SPIRV BuiltIn variable's name to be __spirv_BuiltInXXXX.
     // No matter what BV's linkage name is.
     SPIRVBuiltinVariableKind BVKind;
-    if (BVar->isBuiltin(&BVKind))
+    if (BVar->isBuiltin(&BVKind)) {
       BV->setName(prefixSPIRVName(SPIRVBuiltInNameMap::map(BVKind)));
+      // BuiltinVars are globals that reside in addrspace(1).
+      AddrSpace = SPIRSPIRVAddrSpaceMap::rmap(StorageClassCrossWorkgroup);
+    }
     auto *LVar = new GlobalVariable(*M, Ty, IsConst, LinkageTy,
                                     /*Initializer=*/nullptr, BV->getName(), 0,
                                     GlobalVariable::NotThreadLocal, AddrSpace);
--- a/lib/SPIRV/SPIRVWriter.cpp
+++ b/lib/SPIRV/SPIRVWriter.cpp
@@ -371,7 +371,12 @@ void addFPBuiltinDecoration(SPIRVModule *BM, Instruction *Inst,
   }
 }
 
-SPIRVType *LLVMToSPIRVBase::transType(Type *T) {
+// IsBuiltin is set by caller to be true in two cases:
+// 1. Value associated with the type is a builtin variable.
+// 2. Value associated with the type is a GEPInst with address pointer as a
+//    builtin variable.
+// This variable is just passed onto transPointerType.
+SPIRVType *LLVMToSPIRVBase::transType(Type *T, bool IsBuiltin) {
   LLVMToSPIRVTypeMap::iterator Loc = TypeMap.find(T);
   if (Loc != TypeMap.end())
     return Loc->second;
@@ -415,11 +420,12 @@ SPIRVType *LLVMToSPIRVBase::transType(Type *T) {
   if (T->isPointerTy()) {
     auto *ET = Type::getInt8Ty(T->getContext());
     auto AddrSpc = T->getPointerAddressSpace();
-    return transPointerType(ET, AddrSpc);
+    return transPointerType(ET, AddrSpc, IsBuiltin);
   }
 
   if (auto *TPT = dyn_cast<TypedPointerType>(T)) {
-    return transPointerType(TPT->getElementType(), TPT->getAddressSpace());
+    return transPointerType(TPT->getElementType(), TPT->getAddressSpace(),
+                            IsBuiltin);
   }
 
   if (auto *VecTy = dyn_cast<FixedVectorType>(T)) {
@@ -631,7 +637,14 @@ SPIRVType *LLVMToSPIRVBase::transType(Type *T) {
   return 0;
 }
 
-SPIRVType *LLVMToSPIRVBase::transPointerType(Type *ET, unsigned AddrSpc) {
+// In case of builtin variables and their users, the storage class function
+// should be StorageClassInput, irrespective of the incoming addrspace.
+// Ref:
+// https://registry.khronos.org/OpenCL/specs/3.0-unified/html/OpenCL_Env.html
+// Sec 2.9. Built-in variables.
+// For such cases, we avoid read/writes to PointeeTypeMap.
+SPIRVType *LLVMToSPIRVBase::transPointerType(Type *ET, unsigned AddrSpc,
+                                             bool IsBuiltin) {
   Type *T = PointerType::get(ET, AddrSpc);
   if (ET->isFunctionTy() &&
       !BM->checkExtension(ExtensionID::SPV_INTEL_function_pointers,
@@ -639,9 +652,11 @@ SPIRVType *LLVMToSPIRVBase::transPointerType(Type *ET, unsigned AddrSpc) {
     return nullptr;
 
   std::string TypeKey = (Twine((uintptr_t)ET) + Twine(AddrSpc)).str();
-  auto Loc = PointeeTypeMap.find(TypeKey);
-  if (Loc != PointeeTypeMap.end())
-    return Loc->second;
+  if (!IsBuiltin) {
+    auto Loc = PointeeTypeMap.find(TypeKey);
+    if (Loc != PointeeTypeMap.end())
+      return Loc->second;
+  }
 
   // A pointer to image or pipe type in LLVM is translated to a SPIRV
   // (non-pointer) image or pipe type.
@@ -653,7 +668,7 @@ SPIRVType *LLVMToSPIRVBase::transPointerType(Type *ET, unsigned AddrSpc) {
   if (!BM->isAllowedToUseExtension(
           ExtensionID::SPV_INTEL_usm_storage_classes) &&
       ((AddrSpc == SPIRAS_GlobalDevice) || (AddrSpc == SPIRAS_GlobalHost))) {
-    return transPointerType(ET, SPIRAS_Global);
+    return transPointerType(ET, SPIRAS_Global, IsBuiltin);
   }
   if (ST && !ST->isSized()) {
     Op OpCode;
@@ -670,7 +685,8 @@ SPIRVType *LLVMToSPIRVBase::transPointerType(Type *ET, unsigned AddrSpc) {
 
     auto SaveType = [&](SPIRVType *MappedTy) {
       OpaqueStructMap[Key] = MappedTy;
-      PointeeTypeMap[TypeKey] = MappedTy;
+      if (!IsBuiltin)
+        PointeeTypeMap[TypeKey] = MappedTy;
       return MappedTy;
     };
 
@@ -710,23 +726,29 @@ SPIRVType *LLVMToSPIRVBase::transPointerType(Type *ET, unsigned AddrSpc) {
       }
     }
 
+    auto TheStorageClass = (IsBuiltin)
+                               ? StorageClassInput
+                               : SPIRSPIRVAddrSpaceMap::map(
+                                     static_cast<SPIRAddressSpace>(AddrSpc));
     if (ST->isOpaque()) {
-      return SaveType(BM->addPointerType(
-          SPIRSPIRVAddrSpaceMap::map(static_cast<SPIRAddressSpace>(AddrSpc)),
-          transType(ET)));
+      return SaveType(
+          BM->addPointerType(TheStorageClass, transType(ET, IsBuiltin)));
     }
   } else {
-    SPIRVType *ElementType = transType(ET);
+    SPIRVType *ElementType = transType(ET, IsBuiltin);
     // ET, as a recursive type, may contain exactly the same pointer T, so it
     // may happen that after translation of ET we already have translated T,
     // added the translated pointer to the SPIR-V module and mapped T to this
     // pointer. Now we have to check PointeeTypeMap again.
-    auto Loc = PointeeTypeMap.find(TypeKey);
-    if (Loc != PointeeTypeMap.end()) {
-      return Loc->second;
+    if (!IsBuiltin) {
+      auto Loc = PointeeTypeMap.find(TypeKey);
+      if (Loc != PointeeTypeMap.end()) {
+        return Loc->second;
+      }
     }
-    SPIRVType *TranslatedTy = transPointerType(ElementType, AddrSpc);
-    PointeeTypeMap[TypeKey] = TranslatedTy;
+    SPIRVType *TranslatedTy = transPointerType(ElementType, AddrSpc, IsBuiltin);
+    if (!IsBuiltin)
+      PointeeTypeMap[TypeKey] = TranslatedTy;
     return TranslatedTy;
   }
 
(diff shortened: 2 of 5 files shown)


test/builtin-functions.ll
```diff
@@ -16,7 +16,7 @@
 ; CHECK-FUNCTION-FORMAT-REV: declare spir_func i64 @_Z26__spirv_BuiltInWorkgroupIdi(i32)
 ; CHECK-FUNCTION-FORMAT-OCL-REV: declare spir_func i64 @_Z12get_group_idj(i32)
 
-; CHECK-GLOBAL-FORMAT-REV: @__spirv_BuiltInWorkgroupId = external addrspace(7) constant <3 x i64>
```
asudarsa, line 19: I think this was wrongly being set to addrspace(7) by the reverse translator. Adding @sarnex for comments.

Thanks

test/builtin-storage-class.ll
```diff
@@ -0,0 +1,22 @@
+; RUN: llvm-as %s -o %t.bc
+; RUN: llvm-spirv %t.bc -spirv-text -o - | FileCheck --check-prefix=CHECK-SPIRV %s
+; RUN: llvm-spirv %t.bc -o %t.spv
+; RUN: spirv-val %t.spv
+
+; CHECK-SPIRV: Name [[BuiltinId:[0-9]+]] "__spirv_BuiltInWorkgroupId"
+; CHECK-SPIRV-DAG: Decorate [[BuiltinId]] BuiltIn
+; CHECK-SPIRV-DAG: TypePointer [[PointerTypeId:[0-9]+]] 1
```
asudarsa, line 8: the number '1' corresponds to StorageClassInput here. Thanks

asudarsa:
All the test fails in the CI report this error:
error: line XX: Expected input to have storage class Workgroup, CrossWorkgroup or Function: PtrCastToGeneric
%37 = OpPtrCastToGeneric %_ptr_Generic_uint %__spirv_BuiltInSubgroupMaxSize

This translation is caused by the LLVM IR in the test:
 %2 = addrspacecast ptr addrspace(1) @__spirv_BuiltInSubgroupMaxSize to ptr addrspace(4)

This rule contradicts the rule in OpenCL 3.0 SPIR-V env spec shown in the description.
I see that there is a SPIR-V Registry issue that might be related to this. 
https://github.com/KhronosGroup/SPIRV-Registry/issues/73

Is it possible to allow Input storage class for OpPtrCastToGeneric?

Thanks
Sincerely

test/builtin-functions.ll
sarnex: I'm not sure but the addrspace is not the point of the test so you can change the addrspace as you see fit.

asudarsa:
> > All the test fails in the CI report this error: error: line XX: Expected input to have storage class Workgroup, CrossWorkgroup or Function: PtrCastToGeneric %37 = OpPtrCastToGeneric %_ptr_Generic_uint %__spirv_BuiltInSubgroupMaxSize
> > This translation is caused by the LLVM IR in the test: %2 = addrspacecast ptr addrspace(1) @__spirv_BuiltInSubgroupMaxSize to ptr addrspace(4)
> > This rule contradicts the rule in OpenCL 3.0 SPIR-V env spec shown in the description. I see that there is a SPIR-V Registry issue that might be related to this. [KhronosGroup/SPIRV-Registry#73](https://github.com/KhronosGroup/SPIRV-Registry/issues/73)
> > Is it possible to allow Input storage class for OpPtrCastToGeneric?
> 
> I think it would be better to handle this case on the llvm IR level and copy the value to function memory first. Adding fundamental new address spaces to generic is a problem for some hardware, e.g. Nvidia and AMD where they have native support for generic pointers, but only if it's CL global, local, private memory as they can map their local and private memory into the global address space.
> 
> If generic pointers get fundamental new address spaces it requires different kind of lowering and more runtime overhead.
> 
> Input is also more of a "virtual" address space anyway and it's very hardware specific if the value resides in some accessible memory (e.g. a constant buffer) or if the hardware pulls it via "system value" instructions. In any case, adding Input to generic would make it very painful for runtime compilers to handle in an efficient manner.

Thanks @karolherbst 
I will add a change to copy this over to function memory.

test/builtin-storage-class.ll
```diff
@@ -0,0 +1,22 @@
+; RUN: llvm-as %s -o %t.bc
```
LU-JOHN, line 1: Please add comment describing purpose of this test.

test/builtin-storage-class.ll
LU-JOHN: I think this comment should be put in the source.

lib/SPIRV/SPIRVWriter.cpp
```diff
@@ -371,7 +371,12 @@ void addFPBuiltinDecoration(SPIRVModule *BM, Instruction *Inst,
   }
 }
 
-SPIRVType *LLVMToSPIRVBase::transType(Type *T) {
+// IsBuiltin is set by caller to be true in two cases:
+// 1. Value associated with the type is a builtin variable.
+// 2. Value associated with the type is a GEPInst with address pointer as a
```
LU-JOHN, line 376: Are both these cases hit by testing?

test/builtin-storage-class.ll
asudarsa: Done thanks

lib/SPIRV/SPIRVWriter.cpp
asudarsa: After recent changes, we do not need to worry about the second case. The first one is covered.

Thanks

asudarsa:
I am marking this as draft this I resolve the issue with replacing uses of global variable.

Thanks

lib/SPIRV/SPIRVReader.cpp
```diff
@@ -1167,15 +1167,30 @@ Value *SPIRVToLLVM::mapValue(SPIRVValue *BV, Value *V) {
...
     auto *LD = dyn_cast<LoadInst>(Loc->second);
-    auto *Placeholder = dyn_cast<GlobalVariable>(LD->getPointerOperand());
-    assert(LD && Placeholder &&
-           Placeholder->getName().starts_with(KPlaceholderPrefix) &&
-           "A value is translated twice");
-    // Replaces placeholders for PHI nodes
-    LD->replaceAllUsesWith(V);
-    LD->eraseFromParent();
-    Placeholder->eraseFromParent();
+    if (LD) {
```
asudarsa, line 1184: This change is not related to this PR. It just adds a possibly missing check.

test/builtin-storage-class.ll
asudarsa: Added this info in header comment. Thanks

lib/SPIRV/SPIRVWriter.cpp
```diff
@@ -6610,8 +6646,8 @@ void addPassesForSPIRV(ModulePassManager &PassMgr,
   PassMgr.addPass(PreprocessMetadataPass());
   PassMgr.addPass(SPIRVLowerOCLBlocksPass());
   PassMgr.addPass(OCLToSPIRVPass());
-  PassMgr.addPass(SPIRVRegularizeLLVMPass());
   PassMgr.addPass(SPIRVLowerConstExprPass());
+  PassMgr.addPass(SPIRVRegularizeLLVMPass());
```
asudarsa, line 6650: In SPIRVRegularizeLLVMPass, we have some logic where we try to replace a SPIRV Builtin global constant with a variable in local memory. It helps to have the LowerConstExpr pass run before the regularize pass .

Thanks

asudarsa:
Hi @karolherbst, @MrSidims and @LU-JOHN 

Can you please take a look at this change when convenient?

Thanks

lib/SPIRV/SPIRVReader.cpp
LU-JOHN: The original code has an assertion that LD is not null, so this check should not be necessary unless the original assert was incorrect.

lib/SPIRV/SPIRVReader.cpp
```diff
@@ -1749,6 +1754,36 @@ Value *SPIRVToLLVM::transValueWithoutDecoration(SPIRVValue *BV, Function *F,
...
     transAliasingMemAccess<SPIRVStore>(BS, SI);
+
+    // In SPIRV Writer pass, the SPIRV Builtin variables were processed
+    // in a unique way to ensure we adhered to OpenCL SPIRV Env Spec.
+    // The global variable was copied into a function-level storage.
+    // The following block of code identifies those changes and reverts them.
+    auto *StorePtr = SI->getPointerOperand();
+    if (StorePtr->hasName() &&
+        StorePtr->getName().starts_with("__local_SPIRV_Builtin")) {
+      if (auto *LI = dyn_cast<LoadInst>(SI->getValueOperand())) {
```
LU-JOHN, line 1765: Are these if's necessary?  Is it possible that some __local_SPIRV_Builtin's will not be removed in reverse translation?

lib/SPIRV/SPIRVReader.cpp
asudarsa: Yes. But the assert was placed after the LD was dereferenced in the previous line.

LU-JOHN: Oops.  Yes, I see.

lib/SPIRV/SPIRVRegularizeLLVM.cpp
```diff
@@ -420,6 +420,63 @@ void SPIRVRegularizeLLVMBase::cleanupConversionToNonStdIntegers(Module *M) {
...
+///                               addrspace(1) constant <3 x i64>, align 32
+/// define weak_odr dso_local spir_kernel void @foo() {
+/// ...
+/// %1 = load <3 x i64>, ptr addrspace(1) @__spirv_BuiltInWorkgroupId, align 32
+/// %__local_SPIRV_Builtin* = alloca <3 x i64> addrspace(0), align 32
+/// store <3 x i64> %1, ptr addrspace(0) %__local_SPIRV_Builtin*, align 32
+/// // Replace uses of __spirv_BuiltInWorkgroupId with %__local_SPIRV_Builtin*
+/// ret void
+/// }
+void SPIRVRegularizeLLVMBase::copyBuiltinGVToLocalVar(Module *M) {
```
MrSidims, line 443: There are also cases, when get_..._id() builtin is represented via function call (for OpenCL and SPIR-V friendly). Are these cases being handled?

lib/SPIRV/SPIRVReader.cpp
```diff
@@ -1167,15 +1167,30 @@ Value *SPIRVToLLVM::mapValue(SPIRVValue *BV, Value *V) {
...
     if (Loc->second == V)
       return V;
+
+    // Existing LLVM mapped value could be the local SPIRV Builtin variable
+    // created during forward translation. In this case, we are trying to
+    // replace with mapped value with the global SPIRV Builtin variable
+    // (passed via V here). In this case, we also need a AddrSpaceCast to cast
+    // between the new mapped value and the old.
+    if (Loc->second->hasName() &&
+        Loc->second->getName().starts_with("__local_SPIRV_Builtin")) {
```
MrSidims, line 1177: If reader counterpart is not done, are regressions expected?

lib/SPIRV/SPIRVRegularizeLLVM.cpp
```diff
@@ -420,6 +420,63 @@ void SPIRVRegularizeLLVMBase::cleanupConversionToNonStdIntegers(Module *M) {
...
+/// ...
+/// %0 = load i64, ptr addrspace(1) @__spirv_BuiltInWorkgroupId, align 32
+/// ret void
+/// }
+/// Transformed to --->
+/// @__spirv_BuiltInWorkgroupId = external dso_local local_unnamed_addr
+///                               addrspace(1) constant <3 x i64>, align 32
+/// define weak_odr dso_local spir_kernel void @foo() {
+/// ...
+/// %1 = load <3 x i64>, ptr addrspace(1) @__spirv_BuiltInWorkgroupId, align 32
```
MrSidims, line 437: This is not tested

lib/SPIRV/SPIRVRegularizeLLVM.cpp
```diff
@@ -420,6 +420,63 @@ void SPIRVRegularizeLLVMBase::cleanupConversionToNonStdIntegers(Module *M) {
...
+/// %__local_SPIRV_Builtin* = alloca <3 x i64> addrspace(0), align 32
+/// store <3 x i64> %1, ptr addrspace(0) %__local_SPIRV_Builtin*, align 32
+/// // Replace uses of __spirv_BuiltInWorkgroupId with %__local_SPIRV_Builtin*
+/// ret void
+/// }
+void SPIRVRegularizeLLVMBase::copyBuiltinGVToLocalVar(Module *M) {
+  for (auto I = M->begin(), E = M->end(); I != E;) {
```
MrSidims, line 444: This functionality is not tested

lib/SPIRV/SPIRVRegularizeLLVM.cpp
```diff
@@ -420,6 +420,63 @@ void SPIRVRegularizeLLVMBase::cleanupConversionToNonStdIntegers(Module *M) {
...
+/// %0 = load i64, ptr addrspace(1) @__spirv_BuiltInWorkgroupId, align 32
+/// ret void
+/// }
+/// Transformed to --->
+/// @__spirv_BuiltInWorkgroupId = external dso_local local_unnamed_addr
+///                               addrspace(1) constant <3 x i64>, align 32
+/// define weak_odr dso_local spir_kernel void @foo() {
+/// ...
+/// %1 = load <3 x i64>, ptr addrspace(1) @__spirv_BuiltInWorkgroupId, align 32
+/// %__local_SPIRV_Builtin* = alloca <3 x i64> addrspace(0), align 32
```
MrSidims, line 438: Also, while there were some recent improvements in instruction ordering in the generated module, still, can we ensure, that OpVariable is in the very begging of the basic block? Otherwise this validation rule is being violated:
`All OpVariable instructions in a function must be in the first block in the function.`

lib/SPIRV/SPIRVRegularizeLLVM.cpp
```diff
@@ -420,6 +420,63 @@ void SPIRVRegularizeLLVMBase::cleanupConversionToNonStdIntegers(Module *M) {
...
+            spv::BuiltIn Builtin = spv::BuiltInPosition;
+            if (!(GV->hasName() &&
+                  getSPIRVBuiltin(GV->getName().str(), Builtin)))
+              continue;
+            IRBuilder<> Builder(&Inst);
+            auto *LoadTy = GV->getValueType();
+            auto *LoadedVal = Builder.CreateLoad(LoadTy, GV);
+            if (GV->getAlignment())
+              LoadedVal->setAlignment(llvm::Align(GV->getAlignment()));
+            auto *NewVar = Builder.CreateAlloca(GV->getType(), nullptr,
```
MrSidims, line 460: We might be having several builtins accessed in the kernel (global id, local id, size etc), if I read the code correctly it would result in several __local_SPIRV_Builtin declarations thus SSA violation. Regardless how it's addressed, please reflect it in the test.

lib/SPIRV/SPIRVWriter.cpp
```diff
@@ -670,7 +681,8 @@ SPIRVType *LLVMToSPIRVBase::transPointerType(Type *ET, unsigned AddrSpc) {
 
     auto SaveType = [&](SPIRVType *MappedTy) {
       OpaqueStructMap[Key] = MappedTy;
-      PointeeTypeMap[TypeKey] = MappedTy;
+      if (!IsBuiltin)
+        PointeeTypeMap[TypeKey] = MappedTy;
```
MrSidims, line 685: Does it mean, that for every builtin GV translated we won't visit the map, but instead would translate a type from scratch?

lib/SPIRV/SPIRVReader.cpp
asudarsa: I don't think so...I can try it out in a separate PR.

lib/SPIRV/SPIRVReader.cpp
asudarsa: Good point. They can be turned to 'cast' statements and then asserts to catch any unexpected behavior.

Thanks

lib/SPIRV/SPIRVRegularizeLLVM.cpp
```diff
@@ -420,6 +420,63 @@ void SPIRVRegularizeLLVMBase::cleanupConversionToNonStdIntegers(Module *M) {
...
+/// ret void
+/// }
+/// Transformed to --->
+/// @__spirv_BuiltInWorkgroupId = external dso_local local_unnamed_addr
+///                               addrspace(1) constant <3 x i64>, align 32
+/// define weak_odr dso_local spir_kernel void @foo() {
+/// ...
+/// %1 = load <3 x i64>, ptr addrspace(1) @__spirv_BuiltInWorkgroupId, align 32
+/// %__local_SPIRV_Builtin* = alloca <3 x i64> addrspace(0), align 32
+/// store <3 x i64> %1, ptr addrspace(0) %__local_SPIRV_Builtin*, align 32
```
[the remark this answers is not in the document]
asudarsa: No, the plan here was to copy it to a local variable.

lib/SPIRV/SPIRVRegularizeLLVM.cpp
asudarsa: I do not explicitly control where the OpVariable defined. in all my testing, I do see it occurring at the top.

lib/SPIRV/SPIRVWriter.cpp
asudarsa: unfortunately ... yes.

lib/SPIRV/SPIRVRegularizeLLVM.cpp
```diff
@@ -420,6 +420,63 @@ void SPIRVRegularizeLLVMBase::cleanupConversionToNonStdIntegers(Module *M) {
...
+/// // Replace uses of __spirv_BuiltInWorkgroupId with %__local_SPIRV_Builtin*
```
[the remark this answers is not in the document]
asudarsa: we need to replace one pointer with another.

lib/SPIRV/SPIRVRegularizeLLVM.cpp
asudarsa: yes. I ran into a LIT test fail due to this very same issue and I resolved it.

PR #2457, opened 2024-03-22
Integration of reporting api with translator parser

bwlodarcz:
From it's inception the reporting api had separate parsing pipeline - different from the rest of translator. 
This fix integrate reporting api with the main translator parsing routines.
In addition to that it also fixes a problem of separate validation rules of reporting api.
Now, --spirv-print-report also obey command line validation rules (required extensions) like other commands.


include/LLVMSPIRVOpts.h
```diff
@@ -285,6 +289,8 @@ class TranslatorOpts {
   bool PreserveAuxData = false;
 
   BuiltinFormat SPIRVBuiltinFormat = BuiltinFormat::Function;
+
+  bool IsReport = false;
```
asudarsa, line 293: Please add a small comment about this field. Thanks

lib/SPIRV/libSPIRV/SPIRVModule.cpp
```diff
@@ -2399,7 +2411,10 @@ std::istream &operator>>(std::istream &I, SPIRVModule &M) {
     return MI.parseSPT(I);
   }
 #endif
-  return MI.parseSPIRV(I);
+  if (!MI.TranslationOpts.isReport()) {
+    return MI.parseSPIRV<false>(I);
```
asudarsa, line 2415: Can we not pass the isReport as an argument? Templated code might result in unneeded code bloating here.

Thanks

test/spirv_report.spt
```diff
@@ -1,5 +1,5 @@
 ; RUN: llvm-spirv %s -to-binary -o %t.spv
-; RUN: llvm-spirv --spirv-print-report %t.spv | FileCheck %s --check-prefix=CHECK-DAG
```
asudarsa, line 2: What will be reported if an extension is missing in spirv-ext? Can we add a CHECK for that?

Thanks

test/negative/spirv_report_bad_input.spt
```diff
@@ -1,33 +0,0 @@
-; RUN: llvm-spirv %s -to-binary -o %t.spv
```
asudarsa, line 1: Why is this test being removed?
Just for understanding. Not a blocking question.
Thanks

asudarsa (commented):
Looks good overall. This is a useful refactoring. Thanks.
Few minor comments and clarifications requested.

test/negative/spirv_report_bad_input.spt
bwlodarcz: Report API was added as completely separated parsing pipeline. This fact also means that it needed it's own checks against malformed SPIR-V binary file. The removed file tested these checks. In this PR I'm combining Reporting API with the main parser which also means that it's the subject of the same checks and guards as the hot path pipeline. Main parser is tested already in other tests and IMO doesn't need another one.

lib/SPIRV/libSPIRV/SPIRVModule.cpp
bwlodarcz: Unfortunately, at this stage no. We don't want to introduce unneeded branch in parser hot path for both report and binary cases. User using binary translation shouldn't pay price for Reporting API. This could be avoided if the implementation wouldn't be in iostreams. That makes efficient implementation of early stop very hard.
In this case although you have right that the code size will increase due to two implementations of the same function but prom performance PoV shouldn't make any difference. During first enter icache would be cold anyways and then it's circulating in the same hot routines. There won't be a situation when you have both `<true>` and `<false>` cases loaded to icache in the same time so really problem is negligible. What's more this template can be removed during removal of iostreams as main parsing system.

include/LLVMSPIRVOpts.h
bwlodarcz: This field only purpose is to go around iostream interface to make branch in parser routine and it won't last. Should be removed during removal of iostream implementation.

test/spirv_report.spt
bwlodarcz: Exactly the same like in other tests which covers situation with missing extensions. We already have such tests. For more detailed response please look on the answer below.

lib/SPIRV/libSPIRV/SPIRVModule.cpp
```diff
@@ -2387,6 +2394,11 @@ std::istream &SPIRVModuleImpl::parseSPIRV(std::istream &I) {
       SPIRVDBG(spvdbgs() << "getWordCountAndOpCode EOF 0 0\n");
       break;
     }
+    if constexpr (IS_REPORT) {
+      if (OpCode == OpMemoryModel) {
```
MrSidims, line 2398: That does look fishy, why do we need to skip OpMemoryModel? There should be no more then 2 OpMemoryModel in the model.

test/negative/spirv_report_bad_input.spt
```diff
@@ -1,33 +0,0 @@
...
-; The next line is to corrupt the binary file by changing its Magic Number
-; RUN: echo "0" > %t_corrupted.spv && cat %t.spv >> %t_corrupted.spv
-; RUN: not llvm-spirv --spirv-print-report %t_corrupted.spv 2>&1 | FileCheck %s --check-prefix=CHECK-ERROR
-;
-; CHECK-ERROR: Invalid SPIR-V binary: "InvalidMagicNumber: Invalid Magic Number."
```
MrSidims, line 6: tbh I don't see this very error to be tested elsewhere

include/LLVMSPIRVLib.h
```diff
@@ -107,33 +107,34 @@ std::unique_ptr<SPIRVModule> readSpirvModule(std::istream &IS,
                                              std::string &ErrMsg);
 
 struct SPIRVModuleReport {
-  SPIRV::VersionNumber Version;
+  uint32_t Version;
```
MrSidims, line 110: We probably shouldn't change a type for the Version

asudarsa:
I am good with this PR. Think we should get this merged soon so that Bertrand can continue to work on improving it further. Thanks.

asudarsa (approved):
LGTM. Thanks for working on this Bertrand

asudarsa:
close/open to rerun tests.

asudarsa:
@bwlodarcz Can you please take a look at @MrSidims comments when convenient? Thanks

PR #2461, opened 2024-03-25
switch task sequence enum to new available enum

bowenxue-intel:
Task Sequence Specification was created without explicitly reserving the 6411 token enum. During spec review, that number was reserved by CooperativeMatrixPrefetchINTEL, but Task Sequence was not updated to use a new enum at that point in time, resulting in development with the old enum. 6199 is the newly (confirmed) reserved token enum.


bowenxue-intel:
@MrSidims As discussed, clashing token enum replaced with newly reserved enum for OpTypeTaskSequenceINTEL, please review.

MrSidims (approved):

PR #2466, opened 2024-03-28
Turn on SPIR-V builtin generation for OpenCL CPP sources

asudarsa:
It seems we currently allow OCL to SPIR-V translation only for OpenCL C sources. This PR tries to extend this to OpenCL CPP sources as well.

Thanks


test/extensions/INTEL/SPV_INTEL_arbitrary_precision_fixed_point/capability-arbitrary-precision-fixed-point-numbers.ll
```diff
@@ -541,7 +541,7 @@ attributes #5 = { nounwind }
 
 !0 = !{i32 1, !"wchar_size", i32 4}
 !1 = !{i32 1, i32 2}
-!2 = !{i32 4, i32 100000}
+!2 = !{i32 7, i32 100000}
```
asudarsa, line 544: This test case was actually derived from SYCL source.

lib/SPIRV/OCLToSPIRV.cpp
```diff
@@ -168,9 +168,10 @@ bool OCLToSPIRVBase::runOCLToSPIRV(Module &Module) {
   auto Src = getSPIRVSource(&Module);
   // This is a pre-processing pass, which transform LLVM IR module to a more
   // suitable form for the SPIR-V translation: it is specifically designed to
-  // handle OpenCL C built-in functions and shouldn't be launched for other
+  // handle OpenCL C/CPP built-in functions and shouldn't be launched for other
```
LU-JOHN, line 171: Out of curiosity, what is the reason for only running this transform for these types of sources?

test/ocl_to_spirv_for_opencl_cpp.ll
```diff
@@ -0,0 +1,24 @@
+; RUN: llvm-as %s -o %t.bc
```
LU-JOHN, line 1: Consider adding a comment describing purpose of this test.   I think that looking at the filename or file history is not as clear or convenient.

lib/SPIRV/OCLToSPIRV.cpp
```diff
@@ -168,9 +168,10 @@ bool OCLToSPIRVBase::runOCLToSPIRV(Module &Module) {
...
   // source languages
-  if (std::get<0>(Src) != spv::SourceLanguageOpenCL_C)
+  if (std::get<0>(Src) != spv::SourceLanguageOpenCL_C &&
+      std::get<0>(Src) != spv::SourceLanguageOpenCL_CPP)
```
LU-JOHN, line 174: There is also a SourceLanguageCPP_for_OpenCL.  Should this be allowed?

lib/SPIRV/OCLToSPIRV.cpp
asudarsa: These are the only OpenCL related sources and this particular module deals with OpenCL sources.

lib/SPIRV/OCLToSPIRV.cpp
asudarsa: Yes. Good catch. Thanks. Added.

LU-JOHN (approved):
LGTM

lib/SPIRV/OCLToSPIRV.cpp
```diff
@@ -168,9 +168,11 @@ bool OCLToSPIRVBase::runOCLToSPIRV(Module &Module) {
   auto Src = getSPIRVSource(&Module);
   // This is a pre-processing pass, which transform LLVM IR module to a more
   // suitable form for the SPIR-V translation: it is specifically designed to
-  // handle OpenCL C built-in functions and shouldn't be launched for other
-  // source languages
-  if (std::get<0>(Src) != spv::SourceLanguageOpenCL_C)
+  // handle OpenCL C/CPP and CPP for OpenCL modules and shouldn't be launched
```
svenvh, line 171: In comments we can refer to the official language names:

```suggestion
  // handle OpenCL C/C++ and C++ for OpenCL modules and shouldn't be launched
```

lib/SPIRV/OCLTypeToSPIRV.cpp
```diff
@@ -87,7 +87,13 @@ bool OCLTypeToSPIRVBase::runOCLTypeToSPIRV(Module &Module) {
   AdaptedTy.clear();
   WorkSet.clear();
   auto Src = getSPIRVSource(&Module);
-  if (std::get<0>(Src) != spv::SourceLanguageOpenCL_C)
+  // This is a pre-processing pass, which transform LLVM IR module to a more
+  // suitable form for the SPIR-V translation: it is specifically designed to
+  // handle OpenCL C/CPP and CPP for OpenCL types and shouldn't be launched for
```
svenvh, line 92:
```suggestion
  // handle OpenCL C/C++ and C++ for OpenCL types and shouldn't be launched for
```

svenvh (commented):
Thanks for working on this!

lib/SPIRV/OCLToSPIRV.cpp
asudarsa: Done in commit
Thanks

lib/SPIRV/OCLTypeToSPIRV.cpp
asudarsa: Done in commit. Thanks

svenvh (approved):

Issue #2460, opened 2024-03-25
Missing SPIR-V 1.4 features/changes


MrSidims:
The translator claims SPIR-V 1.4 support, yet there are few features missing

1. `OpCopyLogical` instruction
2. `OpPtrEqual `and `OpPtrNotEqual` instructions
3. `OpPtrDiff` instruction
4. `OpCopyMemory` and `OpCopyMemorySized` instructions since 1.4 have optional Memory Operands
5. `DecorateString` and `OpMemberDecorateString` instructions 
6. CounterBuffer decoration

MrSidims:
Comments:
1. At this point of time I don't see an LLVM counterpart, so guess it's implementable only via SPIR-V friendly translation;

2 and 3. We can consider translating `ptrtoint + cmp / sub` to these instructions during forward translation, but probably we should instead just recognize SPIR-V friendly input and then during reverse translation generate `ptrtoint + cmp / sub`;

4. Pretty much straightforward; 

5. these decorations are odd. I only see a single use-case for UserSemantic decorations for them, meanwhile the translator uses OpDecorate and OpMemberDecorate for the translator. And the translator is right in doing so as the spec puts no restrictions of what kind of literals to used for these 2 instructions;

 6. unlikely to represent right now, but probably should recognize it and skip.

vmaksimo:
The only thing left is to clarify the possible cases where we actually need to use `DecorateString` and `MemberDecorateString` instructions. 
@MrSidims please link registry issue here once it's created. Thanks!

bashbaug:
I created internal SPIR-V issue 810 to clarify OpDecorateString vs. OpDecorate, FYI.

PR #2470, opened 2024-04-02
[SPIR-V 1.4] Allow OpCopyMemorySized to have Memory Operands for both Source and Target 

vmaksimo:
This change addresses p.4 of #2460. 
No changes for `OpCopyMemory` as we currently don't use it in translation.


svenvh (commented):
LGTM, thanks for working on this!  I have just one small naming suggestion.

svenvh (approved):

MrSidims (approved):
Thanks! BTW it's worth cherry-picking in older branches

PR #2395, opened 2024-03-04
[SPIRVToOCL20] Remove tail from atomic_compare_exchange_strong_explicit call

wenju-he:
In translation from  __spirv_AtomicCompareExchange to OpenCL builtin atomic_compare_exchange_strong_explicit, a new alloca `expected` is created and read/written in the OpenCL builtin.
The OpenCL builtin call can't have tail marker since the marker requires that callee doesn't access alloca from the caller. Otherwise llvm alias analysis deduces that the alloca isn't accessed by the call, and instcombine pass replaces the load from the alloca after the call with the value stored to the alloca before the call.

PR #2471, opened 2024-04-03
Fix mangling issue

asudarsa:
In 80dfd864ae556369747074bbde8fc1d4c48547be, we turned on OCLToSPIRV and OCLTypeToSPIRV passes for OpenCL C++ and C++ for OpenCL source languages. In these passes, there are places where some names are mangled/unmangled. Existing code always passed IsCpp=false. With addition of C++ sources, we now need to store the Source language info and use it to figure out IsCpp.

Thanks

### DIFF
--- a/lib/SPIRV/OCLToSPIRV.cpp
+++ b/lib/SPIRV/OCLToSPIRV.cpp
@@ -174,7 +174,7 @@ bool OCLToSPIRVBase::runOCLToSPIRV(Module &Module) {
       std::get<0>(Src) != spv::SourceLanguageOpenCL_CPP &&
       std::get<0>(Src) != spv::SourceLanguageCPP_for_OpenCL)
     return false;
-
+  SrcLang = std::get<0>(Src);
   CLVer = std::get<1>(Src);
 
   LLVM_DEBUG(dbgs() << "Enter OCLToSPIRV:\n");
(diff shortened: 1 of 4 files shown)


asudarsa:
I was not able to come up with a test case that covered these changes. However, it seems like these hookups are required for robustness. 
Moving it out of draft for review.

Thanks

lib/SPIRV/SPIRVReader.cpp
```diff
@@ -3432,7 +3432,9 @@ bool SPIRVToLLVM::translate() {
   if (!transSourceExtension())
     return false;
   transGeneratorMD();
-  if (!lowerBuiltins(BM, M))
+  SPIRVWord Ver = 0;
+  SourceLanguage SrcLang = BM->getSourceLanguage(&Ver);
```
svenvh, line 3436: We should not be using a SPIR-V module's source language to affect the translation flow.  The SPIR-V specification for `OpSource` states:

> This has no semantic impact and can safely be removed from a module.

svenvh (commented):
> I was not able to come up with a test case that covered these changes. However, it seems like these hookups are required for robustness.

I'm a bit hesitant to accept this patch without any tests, because it affects quite a few places in the code.  If an end-to-end test isn't easy to come up with, perhaps try to construct pass-specific tests using the pass plugin (see e.g. #2395).

lib/SPIRV/SPIRVReader.cpp
asudarsa: Hmm. But name mangling happens in a different way depending on what the source language is. If we have to unmangle the names. we need to know the source language. May be we can do something in the regularization pass?

Thanks

PR #2472, opened 2024-04-03
Add convergent attribute for GroupNonUniform ops

Nuullll:


MrSidims (approved):
Nice catch, thanks!

test/group_non_uniform_shuffle_down.ll
```diff
@@ -37,7 +37,7 @@ entry:
 ; CHECK-LLVM: [[ALLOCA_1:%[a-z0-9.]+]] = alloca %"class.sycl::_V1::detail::half_impl::half", align 2
 ; CHECK-LLVM: [[LOAD_0:%[a-z0-9.]+]] = load %"class.sycl::_V1::detail::half_impl::half", ptr [[ALLOCA_1]], align 2
 ; CHECK-LLVM: [[EXTRACT_0:%[a-z0-9.]+]] = extractvalue %"class.sycl::_V1::detail::half_impl::half" [[LOAD_0]], 0
-; CHECK-LLVM: [[CALL_0:%[a-z0-9.]+]] = call spir_func half @_Z22sub_group_shuffle_downDhj(half [[EXTRACT_0]], i32 8) #0
```
MrSidims, line 40: But what happens with the attribute here? I hope - nothing and it's just a bit of a test cleaning. Otherwise it's a bit unexpected and I believe undesired.

Nuullll: GroupNonUniformShuffleDown is also a GroupNonUniform op, so this call is attached with "convergent" attribute too. As a result, the attribute id changed from `#0` to `#1`.

PR #2484, opened 2024-04-08
[SPIR-V 1.4] Add CopyLogical instruction

vmaksimo:
There is no mapping to LLVM instructions, so it can be used only via SPIR-V friendly translation.

This addresses p1. of #2460


vmaksimo:
> Would it be worth adding a SPIR-V-friendly-IR test, to have the newly added code covered?

@svenvh it looks like yes, but unfortunately I wasn't able to come up with the LLVM IR where types will be translated to "not equal but logically matched" types in SPIR-V. Otherwise spirv-val will fail which will stop us from testing reverse translation.

So, my guess is that we can proceed without test till the moment it'll be an issue for users or we'll be able to do a proper translation.

svenvh (approved):
LGTM, thanks!

test/OpCopyLogical.spvasm
```diff
@@ -0,0 +1,23 @@
+; REQUIRES: spirv-as
```
asudarsa, line 1: Please add a comment about the functionality tested here.

asudarsa (approved):
LGTM. Only a minor request for comment. Thanks.

MrSidims (approved):

Issue #2486, opened 2024-04-09
Both GetElementPtrConstantExpr and GetElementPtrInst may represent access to a buildin variable


VyacheslavLevytskyy:
It's possible that access to builtin variable is represented by GetElementPtrConstantExpr rather than GetElementPtrInst. For example, a SPIR-V code at the end of this description presents a a valid pattern that is produced by SPIR-V Backend and results into a ConstantExpr.

SPIRV Translator expects now to see exactly GetElementPtrInst in an access chain for a builtin variable and crashes in lib/SPIRV/SPIRVUtil.cpp in replaceUsesOfBuiltinVar() with llvm_unreachable() call.

```
                                OpCapability Kernel
                                OpCapability Addresses
                                OpCapability Int64
                                OpCapability Int8
                                OpCapability GenericPointer
                                OpCapability Linkage
                                OpMemoryModel Physical64 OpenCL
                                OpEntryPoint Kernel %foo "test"
                                OpDecorate %__spirv_BuiltInWorkgroupSize Constant
                                OpDecorate %__spirv_BuiltInWorkgroupSize Alignment 32
                                OpDecorate %__spirv_BuiltInWorkgroupSize LinkageAttributes "__spirv_BuiltInWorkgroupSize" Import
                                OpDecorate %__spirv_BuiltInWorkgroupSize BuiltIn WorkgroupSize
                        %void = OpTypeVoid
                       %ulong = OpTypeInt 64 0
                        %uint = OpTypeInt 32 0
                       %uchar = OpTypeInt 8 0
                     %v3ulong = OpTypeVector %ulong 3
               %_ptr_CW_uchar = OpTypePointer CrossWorkgroup %uchar
             %_ptr_CW_v3ulong = OpTypePointer CrossWorkgroup %v3ulong
               %_ptr_CW_ulong = OpTypePointer CrossWorkgroup %ulong
                    %foo_type = OpTypeFunction %void
                      %uint_0 = OpConstant %uint 0
                     %ulong_8 = OpConstant %ulong 8
                    %ulong_16 = OpConstant %ulong 16
%__spirv_BuiltInWorkgroupSize = OpVariable %_ptr_CW_v3ulong CrossWorkgroup
                          %c1 = OpSpecConstantOp %_ptr_CW_uchar InBoundsPtrAccessChain %__spirv_BuiltInWorkgroupSize %uint_0 %ulong_8
[...]

PR #2487, opened 2024-04-09
Both GetElementPtrConstantExpr and GetElementPtrInst may represent access to a buildin variable

VyacheslavLevytskyy:
This PR is to fix the issue when SPIRV Translator expects now to see exactly GetElementPtrInst in an access chain for a builtin variable and crashes in lib/SPIRV/SPIRVUtil.cpp in replaceUsesOfBuiltinVar() with llvm_unreachable() call. See https://github.com/KhronosGroup/SPIRV-LLVM-Translator/issues/2486 for the detailed description. Attached is a test case that demonstrates the problem and leads to a crash in current version(s) of SPIRV Translator.

### DIFF
--- a/test/GEPOperator.spvasm
+++ b/test/GEPOperator.spvasm
@@ -1,0 +1,53 @@
+; It's possible that access to builtin variable is represented by
+; GetElementPtrConstantExpr rather than GetElementPtrInst. This
+; test case presents a pattern that results into a ConstantExpr.
+
+; RUN: spirv-as --target-env spv1.0 -o %t.spv %s
+; RUN: spirv-val %t.spv
+; RUN: llvm-spirv -r -o - %t.spv | llvm-dis | FileCheck %s
+
+                                OpCapability Kernel
+                                OpCapability Addresses
+                                OpCapability Int64
+                                OpCapability Int8
+                                OpCapability GenericPointer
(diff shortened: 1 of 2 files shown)


test/GEPOperator.spvasm
```diff
@@ -0,0 +1,53 @@
...
+                         %pv1 = OpBitcast %_ptr_CW_ulong %c1
+                          %v1 = OpLoad %ulong %pv1 Aligned 8
+                        %idx1 = OpIMul %ulong %v0 %v1
+                         %pv2 = OpBitcast %_ptr_CW_ulong %c2
+                          %v2 = OpLoad %ulong %pv2 Aligned 16
+                        %idx2 = OpIMul %ulong %idx1 %v2
+                                OpReturn
+                                OpFunctionEnd
+
+; CHECK: %[[#V0:]] = call spir_func i64 @_Z14get_local_sizej(i32 0)
```
MrSidims, line 49: should we
 `CHECK-NOT: getelementpointer` ?
 and
 `CHECK-NOT: load` ?

VyacheslavLevytskyy: added, thank you

MrSidims (approved):

vmaksimo (approved):

PR #2488, opened 2024-04-09
[Backport to 16] Add support for checked load/store/construct matrix instructions

ficol:
Add support for load/store operations for a cooperative matrix such that original matrix shape is known and implementations are able to reason about how to deal with the out of bounds.

CapabilityCooperativeMatrixCheckedInstructionsINTEL = 6192
CooperativeMatrixLoadCheckedINTEL = 6193
CooperativeMatrixStoreCheckedINTEL = 6194


MrSidims (commented):
To double check, are we sure, that we want target extension matrix type on this branch?

ficol:
It was easier to adapt original code to version 16 using target extension type. Should I avoid it?

MrSidims:
> It was easier to adapt original code to version 16 using target extension type. Should I avoid it?

I assume the patch is needed for https://github.com/intel/intel-graphics-compiler . The question is would it used target extension types or instead typed pointers? I'm not an expert in IGC code base, hence I'm asking.

MrSidims (approved):

PR #2489, opened 2024-04-10
Translate OpConvertF* SPIR-V instructions to llvm.* intrinsics

asudarsa:
Currently OpConvertFtoS and OpConvertFToU SPIR-V instructions get translated to OCL calls. However, this translation is not possible when the conversion happens to non-standard int types (somthing that is not i8, i16,i32, or i64). In such cases, this PR tries to lower it back to the llvm.fptosi.sat and llvm.fptoui.sat instructions respectively.

Thanks

### DIFF
--- a/lib/SPIRV/SPIRVReader.cpp
+++ b/lib/SPIRV/SPIRVReader.cpp
@@ -2666,6 +2666,33 @@ Value *SPIRVToLLVM::transValueWithoutDecoration(SPIRVValue *BV, Function *F,
     return mapValue(BV, Builder.CreateMaskedScatter(InputVector, PtrVector,
                                                     Align(Alignment), Mask));
   }
+  // Conversion instructions are typically converted to OCL builtins.
+  // However conversions involving arbitrary integers are not supported in OCL
+  // builtins. Hence, they are lowered to llvm. builtins.
+  // TODO: can ALL conversions be lowered to llvm. builtins?
+  case OpConvertFToS:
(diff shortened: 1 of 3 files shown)


asudarsa:
@LU-JOHN, Can you please take a look? Thanks

MrSidims (changes requested):
This patch contradicts https://github.com/KhronosGroup/SPIRV-LLVM-Translator/blob/main/docs/SPIRVRepresentationInLLVM.rst#additional-requirements-for-llvm-module . Please consider updating the contract set in the document or provide type regularization mechanism before the translation.

SPV_INTEL_arbitrary_precision_integers was mainly written for C23 BitInt feature. It's quite dangerous to use it for translation of the optimized LLVM IR without knowing, that your target supports the extension.

asudarsa:
Hi @MrSidims 

Thanks so much for the comments and offline chats. It seems like a better solution will be to 'regularize' the incoming LLVM IR. I will attempt that in a separate Jira.
I also think that this PR will be helpful for FPGAs. Should we try to merge this? I am ok with abandoning this as well.

Thanks again

asudarsa:
Hi all

I have uploaded a better PR to address this here: https://github.com/KhronosGroup/SPIRV-LLVM-Translator/pull/2500
I might eventually close this. But we can discuss offline.

Thanks

asudarsa:
Closing this as a better solution is now available. Thanks for review everyone.

PR #2490, opened 2024-04-10
Fix of ParentIdx mismatch for DebugTypeInheritance in SPIRV.debug.h

bwlodarcz:
Fixed ParentIdx was mismatched for DebugTypeInheritance type in context of NonSemantic.Shader.DebugInfo.100. That lead to heap buffer overflow in SPIRVExtInst::getExtOp getter when instruction was incorrectly casted
to SPIRVExtInst as parent of the culprit instruction in SPIRVToLLVMDbgTran::getDIBuilder. The missmatch happen because in previous used standard OpenCL.DebugInfo.100 DebugTypeInheritance had Child field as zero indexed argument. In newer standard that field is removed.

### DIFF
--- a/lib/SPIRV/libSPIRV/SPIRV.debug.h
+++ b/lib/SPIRV/libSPIRV/SPIRV.debug.h
@@ -509,7 +509,6 @@ enum {
   OffsetIdx       = 1,
   SizeIdx         = 2,
   FlagsIdx        = 3,
-  MinOperandCount = 3,
   OperandCount    = 4
 };
 }
@@ -521,7 +520,6 @@ enum {
   OffsetIdx       = 2,
   SizeIdx         = 3,
   FlagsIdx        = 4,
-  MinOperandCount = 4,
   OperandCount    = 5
 };
 }
--- a/lib/SPIRV/libSPIRV/SPIRV.debug.h
+++ b/lib/SPIRV/libSPIRV/SPIRV.debug.h
@@ -991,7 +991,10 @@ inline bool hasDbgInstParentScopeIdx(
     ParentScopeIdx = TypeMember::OpenCL::ParentIdx;
     return true;
   case SPIRVDebug::TypeInheritance:
-    ParentScopeIdx = TypeInheritance::ParentIdx;
+    if (ExtKind == SPIRV::SPIRVEIS_OpenCL_DebugInfo_100)
+      ParentScopeIdx = TypeInheritance::ParentIdx;
+    else
+      ParentScopeIdx = TypeInheritance::ChildIdx;
     return true;
   case SPIRVDebug::TypePtrToMember:
     ParentScopeIdx = TypePtrToMember::ParentIdx;
--- a/lib/SPIRV/libSPIRV/SPIRV.debug.h
+++ b/lib/SPIRV/libSPIRV/SPIRV.debug.h
@@ -503,17 +503,29 @@ enum {
 } // namespace TypeMember
 
 namespace TypeInheritance {
+namespace NonSemantic {
+enum {
+  ParentIdx       = 0,
+  OffsetIdx       = 1,
+  SizeIdx         = 2,
+  FlagsIdx        = 3,
+  MinOperandCount = 3,
+  OperandCount    = 4
+};
+}
+
+namespace OpenCL {
 enum {
   ChildIdx        = 0,
   ParentIdx       = 1,
   OffsetIdx       = 2,
   SizeIdx         = 3,
   FlagsIdx        = 4,
-  // NonSemantic
   MinOperandCount = 4,
   OperandCount    = 5
 };
 }
+} // namespace TypeInheritance
 
 namespace TypePtrToMember {
 enum {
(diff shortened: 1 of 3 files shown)


lib/SPIRV/libSPIRV/SPIRV.debug.h
```diff
@@ -991,7 +991,10 @@ inline bool hasDbgInstParentScopeIdx(
     ParentScopeIdx = TypeMember::OpenCL::ParentIdx;
     return true;
   case SPIRVDebug::TypeInheritance:
-    ParentScopeIdx = TypeInheritance::ParentIdx;
+    if (ExtKind == SPIRV::SPIRVEIS_OpenCL_DebugInfo_100)
```
MrSidims, line 994: Lets separate indexes for `TypeInheritance` in OpenCL and NonSemantic namespaces (like it's done for other instructions).

lib/SPIRV/libSPIRV/SPIRV.debug.h
```diff
@@ -991,7 +991,10 @@ inline bool hasDbgInstParentScopeIdx(
...
+      ParentScopeIdx = TypeInheritance::ParentIdx;
+    else
+      ParentScopeIdx = TypeInheritance::ChildIdx;
```
vmaksimo, line 997: It'd be better to use something like `TypeInheritance::NonSemantic::ParentIdx`. Do you mind to rework this?

vmaksimo (commented):
Do you

lib/SPIRV/libSPIRV/SPIRV.debug.h
bwlodarcz: Done.

lib/SPIRV/libSPIRV/SPIRV.debug.h
bwlodarcz: Done.

asudarsa (changes requested):
The change looks good. Is it possible to add a LIT regression test for this?

Thanks

lib/SPIRV/libSPIRV/SPIRV.debug.h
```diff
@@ -503,17 +503,29 @@ enum {
 } // namespace TypeMember
 
 namespace TypeInheritance {
+namespace NonSemantic {
+enum {
+  ParentIdx       = 0,
+  OffsetIdx       = 1,
+  SizeIdx         = 2,
+  FlagsIdx        = 3,
+  MinOperandCount = 3,
```
vmaksimo, line 512: `MinOperandCount` is not needed here and in OpenCL namespace as long as we have separated them. There are no optional operands in `TypeInheritance`

bwlodarcz:
@asudarsa This is a fix for memory issue which arises from major architectural flaw. The source of bug basically looks like that: I can create DebugInfo by hand in whatever standard and as Scope operation point to something unrelated. That is invalid SPIR-V from spec PoV. DebugInfo doesn't perform any validating checks, nor Parser. In addition to that, LIT testing infrastructure (regex with some anemic macro capabilities) isn't really created to performing such validation checks - it's an backend testing infrastructure (step before asm). It doesn't have enough capabilities.
There are LIT tests in Translator which covers DebugTypeInheritance functionality as is. Without massive reworking DebugInfo parts in Translator covering that kind of bugs is impossible.

MrSidims (approved):
I'm OK with not having a test for this path. Please apply Viktoria's comment

asudarsa:
> @asudarsa This is a fix for memory issue which arises from major architectural flaw. The source of bug basically looks like that: I can create DebugInfo by hand in whatever standard and as Scope operation point to something unrelated. That is invalid SPIR-V from spec PoV. DebugInfo doesn't perform any validating checks, nor Parser. In addition to that, LIT testing infrastructure (regex with some anemic macro capabilities) isn't really created to performing such validation checks - it's an backend testing infrastructure (step before asm). It doesn't have enough capabilities. There are LIT tests in Translator which covers DebugTypeInheritance functionality as is. Without massive reworking DebugInfo parts in Translator covering that kind of bugs is impossible.

I thought this was fixing a bug that was reported by user. May be not. My two cents: If a known issue is being resolved , it is a good idea to have an associated LIT test so that we can catch any regressions. If the test creation is going to be time consuming, I am ok with merging this, but adding the test at a later stage. Thanks

bwlodarcz:
> > @asudarsa This is a fix for memory issue which arises from major architectural flaw. The source of bug basically looks like that: I can create DebugInfo by hand in whatever standard and as Scope operation point to something unrelated. That is invalid SPIR-V from spec PoV. DebugInfo doesn't perform any validating checks, nor Parser. In addition to that, LIT testing infrastructure (regex with some anemic macro capabilities) isn't really created to performing such validation checks - it's an backend testing infrastructure (step before asm). It doesn't have enough capabilities. There are LIT tests in Translator which covers DebugTypeInheritance functionality as is. Without massive reworking DebugInfo parts in Translator covering that kind of bugs is impossible.
> 
> I thought this was fixing a bug that was reported by user. May be not. My two cents: If a known issue is being resolved , it is a good idea to have an associated LIT test so that we can catch any regressions. If the test creation is going to be time consuming, I am ok with merging this, but adding the test at a later stage. Thanks

@asudarsa Yes, it's a bug that was reported by user. The underlying issue is not resolved. The origin of the bug is flawed and insecure architecture of the DebugInfo translation.
The concrete instance of bug reported by user was solved in this PR. Origin of that concrete instance of bug was missed ParentIdx for DebugTypeInheritance but the main source of the bug is in architecture. In DebugInfo most of the fields have Scope/ParentIdx due to equivalence to DWARF format. There aren't any validating routines or passes DebugInfo. That means that any debug instruction can assign as a parent/scope any other (even completely unrelated instruction from execution) as it's parent with every 
consequence of that fact (including memory issues) - and translator isn't aware of that. In addition to that even if I prepare test for that concrete bug how do we catch failing memory issue when sanitizers aren't enabled in CI by default? In that light I don't see any sensile way to test (because what will we be testing and how?).

asudarsa:
I think we can revisit the testing issue when we have sanitizer testing in place. Approving this. Thanks @bwlodarcz for your detailed answers.

Sincerely

asudarsa (approved):
LGTM

lib/SPIRV/libSPIRV/SPIRV.debug.h
bwlodarcz: Addressed

PR #2491, opened 2024-04-10
[SPIR-V 1.5] Deprecate OpLessOrGreater in favor of OpFOrdNotEqual

vmaksimo:
This is a follow-up to 08884df

### DIFF
--- a/lib/SPIRV/SPIRVWriter.cpp
+++ b/lib/SPIRV/SPIRVWriter.cpp
@@ -6097,6 +6097,11 @@ LLVMToSPIRVBase::transBuiltinToInstWithoutDecoration(Op OC, CallInst *CI,
   // Moreover, OpAtomicCompareExchangeWeak has been deprecated.
   if (OC == OpAtomicCompareExchangeWeak)
     OC = OpAtomicCompareExchange;
+
+  if (OC == OpLessOrGreater &&
+      BM->getSPIRVVersion() >= static_cast<SPIRVWord>(VersionNumber::SPIRV_1_5))
+    OC = OpFOrdNotEqual;
+
   if (isGroupOpCode(OC))
     BM->addCapability(CapabilityGroups);
   switch (static_cast<size_t>(OC)) {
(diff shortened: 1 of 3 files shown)


lib/SPIRV/SPIRVWriter.cpp
```diff
@@ -6097,6 +6097,11 @@ LLVMToSPIRVBase::transBuiltinToInstWithoutDecoration(Op OC, CallInst *CI,
   // Moreover, OpAtomicCompareExchangeWeak has been deprecated.
   if (OC == OpAtomicCompareExchangeWeak)
     OC = OpAtomicCompareExchange;
+
+  if (OC == OpLessOrGreater &&
+      BM->getSPIRVVersion() >= static_cast<SPIRVWord>(VersionNumber::SPIRV_1_5))
+    OC = OpFOrdNotEqual;
```
asudarsa, line 6103: Couple of minor questions:
1. Is OpFOrdNotEqual not available before 1.5? Can we not make this substitution unconditionally?
2. Seems like this check (BM->getSPIRVVersion() >= static_cast<SPIRVWord>(VersionNumber::SPIRV_1_5)) will be repeated often. 
Will it help to have a helper function? Like BM->IsSPVVerGE1_5()?

Thanks

vmaksimo: 1. Actually we do make it unconditionally, see: https://github.com/KhronosGroup/SPIRV-LLVM-Translator/blob/main/lib/SPIRV/OCLUtil.cpp#L302
This very change is covering the case where we get OpLessOrGreater as SPIR-V friendly function.
2. It's actually the only change where we do such check (there planned no more changes for spir-v 1.5). But I'll keep it in mind in SPIR-V 1.6 implementation

MrSidims (commented):
Guess this can be tested via SPIR-V friendly LLVM IR test

svenvh (approved):

MrSidims (approved):
CI failure is unrelated and fixed elsewhere

PR #2498, opened 2024-04-10
Start moving away from insert-before Instruction pointers

svenvh:
In preparation for LLVM's [deprecation](https://discourse.llvm.org/t/psa-instruction-constructors-changing-to-iterator-only-insertion/77845) of Instruction constructors and Create methods that accept `Instruction` pointers, upgrade call sites to pass an iterator instead of an `Instruction` pointer.

This patch only upgrades constructor and `Create` calls that were trivial to upgrade with no impact on the test suite.  Some calls remain to be upgraded as future work.

### DIFF
--- a/lib/SPIRV/OCLToSPIRV.cpp
+++ b/lib/SPIRV/OCLToSPIRV.cpp
@@ -499,7 +499,8 @@ CallInst *OCLToSPIRVBase::visitCallAtomicCmpXchg(CallInst *CI) {
 }
 
 void OCLToSPIRVBase::visitCallAtomicInit(CallInst *CI) {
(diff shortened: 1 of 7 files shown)


vmaksimo (approved):

asudarsa:
Thanks for looking into this @svenvh 
One quick comment (might be my ignorance). Do we need a nullptr check before dereferencing these pointers?

Thanks

asudarsa (approved):
LGTM. My earlier comment about null check seems irrelevant, as we already dereference in other places.

Thanks

svenvh:
Yes, in most if not all cases, the variable is already dereferenced for `getName`, `getOperand` or `replaceUsesOfWith` calls.

PR #2499, opened 2024-04-10
Don't call getName() on literal structs.

maarquitos14:


lib/SPIRV/SPIRVUtil.cpp
```diff
@@ -804,8 +804,11 @@ bool getParameterTypes(Function *F, SmallVectorImpl<Type *> &ArgTys,
       HasSret = true;
       unsigned AS = Arg.getType()->getPointerAddressSpace();
       if (auto *STy = dyn_cast<StructType>(Ty))
-        ArgTys.push_back(
-            TypedPointerType::get(GetStructType(STy->getName()), AS));
+        if (!STy->isLiteral())
```
asudarsa, line 807: Can we not just call STy->hasName() here? Suppose that will be more generic? Thanks

asudarsa:
Is it possible to add a LIT test to cover this change?

Thanks

test/ocl_to_spirv_for_opencl_cpp_with_unnamed_struct.ll
```diff
@@ -0,0 +1,17 @@
+;; This test checks if builtins in LLVM IR generated from OpenCL_CPP sources
```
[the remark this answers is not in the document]
maarquitos14: Good catch! Thanks.

test/ocl_to_spirv_for_opencl_cpp_with_unnamed_struct.ll
```diff
@@ -0,0 +1,17 @@
...
+; RUN: llvm-as %s -o %t.bc
+; RUN: llvm-spirv %t.bc -o %t.spv
+; RUN: spirv-val %t.spv
+
+target datalayout = "e-p:32:32-i64:64-v16:16-v24:32-v32:32-v48:64-v96:128-v192:256-v256:256-v512:512-v1024:1024"
+target triple = "spir-unknown-unknown"
+
+%opencl.event_t = type opaque
+
+declare spir_func void @_ZNKSt7complexIfE5__repEv(ptr addrspace(4) dead_on_unwind noalias writable sret({ float, float }) align 4)
```
LU-JOHN, line 13: What would happen with this test case before this PR?

maarquitos14: The following assert is raised: https://github.com/intel/llvm/blob/65bdffb1c9d4c474316d3e330fc3c59338e004f6/llvm/lib/IR/Type.cpp#L591

lib/SPIRV/SPIRVUtil.cpp
```diff
@@ -804,8 +804,11 @@ bool getParameterTypes(Function *F, SmallVectorImpl<Type *> &ArgTys,
       HasSret = true;
       unsigned AS = Arg.getType()->getPointerAddressSpace();
       if (auto *STy = dyn_cast<StructType>(Ty))
-        ArgTys.push_back(
-            TypedPointerType::get(GetStructType(STy->getName()), AS));
+        if (STy->hasName())
+          ArgTys.push_back(
+              TypedPointerType::get(GetStructType(STy->getName()), AS));
+        else
+          ArgTys.push_back(TypedPointerType::get(STy, AS));
```
LU-JOHN, line 811: I added an:

>  assert(STy==GetStructType(STy->getName())); 

at line 808 and all the LIT tests pass.  Can we remove the if-else and always do line 811?

I also tried simplifying lines 799-816 and just always doing:

> ArgTys.push_back(Arg.getType());

and this also passed LIT testing.

At a minimum we should fuse the cases on lines 800 and 815 since they have identical bodies.

LU-JOHN: > I added an:
> 
> > assert(STy==GetStructType(STy->getName()));
> 
> at line 808 and all the LIT tests pass. Can we remove the if-else and always do line 811?
> 
> I also tried simplifying lines 799-816 and just always doing:
> 
> > ArgTys.push_back(Arg.getType());
> 
> and this also passed LIT testing.

I think the simplifications are too risky.  The code was written intentionally to handle these variations.

LU-JOHN (approved):
LGTM

MrSidims (approved):

test/ocl_to_spirv_for_opencl_cpp_with_unnamed_struct.ll
```diff
@@ -0,0 +1,17 @@
+;; This test checks if builtins in LLVM IR generated from OpenCL_CPP sources
+;; including unnamed structs are correctly translated to SPIR-V builtin variables.
```
MrSidims, line 2: Wait, in SPIR-V builtin variables has a concrete meaning, see https://registry.khronos.org/SPIR-V/specs/unified1/SPIRV.html#BuiltIn . I don't see any builtins here, so likely to be a misuse of the word

maarquitos14: Thanks for spotting this!

MrSidims (approved):

Issue #481, opened 2020-03-31
Translator doesn't handle LLVM integer types.


bader:
https://godbolt.org/z/p3D7TX - legit OpenCL code, which translator can't convert to SPIR-V due to `i3` type.
The translator with the assertions gives me:
```
InvalidBitWidth: Invalid bit width in input: 3
```

Proper LLVM back-end should be able to handle such types...

Is there an easy way to avoid such types in LLVM IR by configuring SPIR target in clang?

AlexeySachkov:
Error for this LLVM IR is expected: from [this](https://github.com/KhronosGroup/SPIRV-LLVM-Translator/pull/401#issuecomment-568483832) comment:

> 2.16.1. Universal Validation Rules
>    Data rules
> Scalar integer types can be parameterized only as 32 bit, plus any additional sizes enabled by capabilities.

However, I agree that input source code is pretty valid and shouldn't cause any compilation errors somewhere in the compiler stack

Note: support to arbitrary precision integers is on the way into the translator via support of corresponding extension, see #480

bader:
The validation rule applies to SPIR-V, not LLVM IR. 
Translator is trying to do 1:1 mapping for integer type width, which leads to the rule violation.
Alternative approach is map LLVM integer types to "legal" SPIR-V types. This is was LLVM back-ends are doing when they lower LLVM IR to HW ISA.

keryell:
Interesting code which happens to be naturally FPGA friendly! :-)
I am curious about how it is lowered to FPGA efficiently then?

svenvh:
> The validation rule applies to SPIR-V, not LLVM IR.
> Translator is trying to do 1:1 mapping for integer type width, which leads to the rule violation.
> Alternative approach is map LLVM integer types to "legal" SPIR-V types. This is what LLVM back-ends are doing when they lower LLVM IR to HW ISA.

Agreed, the translator is only doing a best-effort mapping to SPIR-V currently.  We should consider adding a legalization pass to the translator to deal with cases like these.

StuartDBrady:
Note that i1 is currently mapped to bool: any legalization would need to avoid breaking bool, but should still properly legalize any use of i1 as a single-bit integer by frontends.

bader:
I've found "an easy way to avoid such types in LLVM IR by configuring SPIR target in clang".

[...]

PR #2500, opened 2024-04-10
Regularize LLVM code to remove use of non standard integer types in fptoui and fptosi intrinsics

asudarsa:
It is possible that incoming LLVM IR conversion instructions convert floating point to non-standard integer types. 
For e.g. %0 = call i2 @llvm.fptosi.sat.i2.f32(float %input)
Such types are not supported in SPIR-V. This PR cleans up such code and removes occurrence of non-standard integer types.

Thanks

### DIFF
--- a/lib/SPIRV/SPIRVRegularizeLLVM.cpp
+++ b/lib/SPIRV/SPIRVRegularizeLLVM.cpp
@@ -360,7 +360,7 @@ void SPIRVRegularizeLLVMBase::cleanupConversionToNonStdIntegers(Module *M) {
     auto IID = F->getIntrinsicID();
     if (IID != Intrinsic::fptosi_sat && IID != Intrinsic::fptoui_sat)
       continue;
-    for (auto I : F->users()) {
+    for (auto *I : F->users()) {
       if (IntrinsicInst *II = dyn_cast<IntrinsicInst>(I)) {
         // TODO: Vector type not supported yet.
         if (isa<VectorType>(II->getType()))
--- a/lib/SPIRV/SPIRVRegularizeLLVM.cpp
+++ b/lib/SPIRV/SPIRVRegularizeLLVM.cpp
@@ -357,6 +357,8 @@ void SPIRVRegularizeLLVMBase::cleanupConversionToNonStdIntegers(Module *M) {
   for (auto I = M->begin(), E = M->end(); I != E;) {
     Function *F = &(*I++);
     std::vector<Instruction *> ToErase;
+    // TODO: Improve parsing logic by identifying all fptoui.sat and fptosi.sat
+    // intrinsics and then iterating over their users.
     for (BasicBlock &BB : *F) {
       for (Instruction &I : BB) {
         if (IntrinsicInst *II = dyn_cast<IntrinsicInst>(&I)) {
@@ -415,6 +417,7 @@ void SPIRVRegularizeLLVMBase::cleanupConversionToNonStdIntegers(Module *M) {
     }
     for (Instruction *V : ToErase) {
       assert(V->user_empty());
+      V->dropAllReferences();
       V->eraseFromParent();
     }
   }
--- a/lib/SPIRV/SPIRVRegularizeLLVM.cpp
+++ b/lib/SPIRV/SPIRVRegularizeLLVM.cpp
@@ -345,8 +345,8 @@ void SPIRVRegularizeLLVMBase::cleanupConversionToNonStdIntegers(Module *M) {
                     IID, {Inst->getType(), II->getOperand(0)->getType()},
                     II->getOperand(0));
                 Inst->replaceAllUsesWith(NewII);
-                ToErase.push_back(&I);
                 ToErase.push_back(Inst);
+                ToErase.push_back(&I);
               }
             }
           }
--- a/lib/SPIRV/SPIRVRegularizeLLVM.cpp
+++ b/lib/SPIRV/SPIRVRegularizeLLVM.cpp
@@ -354,64 +354,61 @@ void SPIRVRegularizeLLVMBase::expandSYCLTypeUsing(Module *M) {
 // %4 = select i1 %2, i32 -2, i32 %3
 // Replace uses of %1 in Input with %4 in Output
 void SPIRVRegularizeLLVMBase::cleanupConversionToNonStdIntegers(Module *M) {
-  for (auto I = M->begin(), E = M->end(); I != E;) {
-    Function *F = &(*I++);
+  for (auto FI = M->begin(), FE = M->end(); FI != FE;) {
+    Function *F = &(*FI++);
     std::vector<Instruction *> ToErase;
-    // TODO: Improve parsing logic by identifying all fptoui.sat and fptosi.sat
-    // intrinsics and then iterating over their users.
-    for (BasicBlock &BB : *F) {
-      for (Instruction &I : BB) {
-        if (IntrinsicInst *II = dyn_cast<IntrinsicInst>(&I)) {
-          // TODO: Vector type not supported yet.
-          if (isa<VectorType>(II->getType()))
+    auto IID = F->getIntrinsicID();
+    if (IID != Intrinsic::fptosi_sat && IID != Intrinsic::fptoui_sat)
+      continue;
+    for (auto I : F->users()) {
+      if (IntrinsicInst *II = dyn_cast<IntrinsicInst>(I)) {
+        // TODO: Vector type not supported yet.
+        if (isa<VectorType>(II->getType()))
+          continue;
+        auto IID = II->getIntrinsicID();
+        auto IntBitWidth = II->getType()->getScalarSizeInBits();
+        if (IntBitWidth == 8 || IntBitWidth == 16 || IntBitWidth == 32 ||
+            IntBitWidth == 64)
+          continue;
+        if (IID == Intrinsic::fptosi_sat) {
+          // Identify sext (user of II). Make sure that's the only use of II.
+          auto *User = II->getUniqueUndroppableUser();
+          if (!User || !isa<SExtInst>(User))
             continue;
-          auto IID = II->getIntrinsicID();
-          auto IntBitWidth = II->getType()->getScalarSizeInBits();
-          if (IntBitWidth == 8 || IntBitWidth == 16 || IntBitWidth == 32 ||
-              IntBitWidth == 64)
+          auto *SExtI = dyn_cast<SExtInst>(User);
+          auto *NewIType = SExtI->getType();
+          IRBuilder<> IRB(II);
+          auto *NewII = IRB.CreateIntrinsic(
+              IID, {NewIType, II->getOperand(0)->getType()}, II->getOperand(0));
+          Constant *MaxVal = ConstantInt::get(
+              NewIType, APInt::getSignedMaxValue(IntBitWidth).getSExtValue());
+          Constant *MinVal = ConstantInt::get(
+              NewIType, APInt::getSignedMinValue(IntBitWidth).getSExtValue());
+          auto *GTMax = IRB.CreateICmp(CmpInst::ICMP_SGE, NewII, MaxVal);
+          auto *LTMin = IRB.CreateICmp(CmpInst::ICMP_SLE, NewII, MinVal);
+          auto *SatMax = IRB.CreateSelect(GTMax, MaxVal, NewII);
+          auto *SatMin = IRB.CreateSelect(LTMin, MinVal, SatMax);
+          SExtI->replaceAllUsesWith(SatMin);
+          ToErase.push_back(SExtI);
+          ToErase.push_back(II);
+        }
+        if (IID == Intrinsic::fptoui_sat) {
+          // Identify zext (user of II). Make sure that's the only use of II.
+          auto *User = II->getUniqueUndroppableUser();
+          if (!User || !isa<ZExtInst>(User))
             continue;
-          if (IID == Intrinsic::fptosi_sat) {
-            // Identify sext (user of I). Make sure that's the only use of I.
-            auto *User = I.getUniqueUndroppableUser();
-            if (!User || !isa<SExtInst>(User))
-              continue;
-            auto *SExtI = dyn_cast<SExtInst>(User);
-            auto *NewIType = SExtI->getType();
-            IRBuilder<> IRB(&I);
-            auto *NewII = IRB.CreateIntrinsic(
-                IID, {NewIType, II->getOperand(0)->getType()},
-                II->getOperand(0));
-            Constant *MaxVal = ConstantInt::get(
-                NewIType, APInt::getSignedMaxValue(IntBitWidth).getSExtValue());
-            Constant *MinVal = ConstantInt::get(
-                NewIType, APInt::getSignedMinValue(IntBitWidth).getSExtValue());
-            auto *GTMax = IRB.CreateICmp(CmpInst::ICMP_SGE, NewII, MaxVal);
-            auto *LTMin = IRB.CreateICmp(CmpInst::ICMP_SLE, NewII, MinVal);
-            auto *SatMax = IRB.CreateSelect(GTMax, MaxVal, NewII);
-            auto *SatMin = IRB.CreateSelect(LTMin, MinVal, SatMax);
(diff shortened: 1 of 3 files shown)


asudarsa:
Please review once marked ready. Thanks

asudarsa:
This PR tries to address a  very limited set of issues found due to use of non-standard integer types. A more rounded solution may be required in the long run. Tracked here: https://github.com/KhronosGroup/SPIRV-LLVM-Translator/issues/481

Thanks

maksimsab (approved):
LGTM.

maarquitos14 (approved):
LGTM.

VyacheslavLevytskyy (approved):

lib/SPIRV/SPIRVRegularizeLLVM.cpp
```diff
@@ -322,6 +322,108 @@ void SPIRVRegularizeLLVMBase::expandSYCLTypeUsing(Module *M) {
...
+// and Rule 3, we saturate the result of the promoted instruction based on
+// original integer type (iX)
+// Example:
+// Input:
+// %0 = call i2 @llvm.fptoui.sat.i2.f32(float %input)
+// %1 = zext i32 %0
+// Output:
+// %0 = call i32 @_Z17convert_ulong_satf(float %input)
+// %1 = icmp uge i32 %0, 3 <Largest 2-bit unsigned integer>
+// %2 = icmp ule i32 %0, 0 <Smallest 2-bit unsigned integer>
```
LU-JOHN, line 352: For conversion to an unsigned type, the output of the call will already guarantee a minimum value of 0, so there is no need for guard code on that side.

asudarsa: That's a good point. Thanks. Is it ok to leave the code as is for sake of readability?

asudarsa: Thinking again, seems like a useful change. Thanks. Addressed in the next commit.

lib/SPIRV/SPIRVRegularizeLLVM.cpp
```diff
@@ -322,6 +322,104 @@ void SPIRVRegularizeLLVMBase::expandSYCLTypeUsing(Module *M) {
...
+          if (isa<VectorType>(II->getType()))
+            continue;
+          auto IID = II->getIntrinsicID();
+          auto IntBitWidth = II->getType()->getScalarSizeInBits();
+          if (IntBitWidth == 8 || IntBitWidth == 16 || IntBitWidth == 32 ||
+              IntBitWidth == 64)
+            continue;
+          if (IID == Intrinsic::fptosi_sat) {
+            // Identify sext (user of I). Make sure that's the only use of I.
+            auto *User = I.getUniqueUndroppableUser();
```
LU-JOHN, line 373: Could we allow multiple uses by SExtInst/ZExtInst or is this so rare that it is not worth dealing with?

test/llvm-intrinsics/fp_to_arbitrary_size_int_intrinsic.ll
```diff
@@ -0,0 +1,57 @@
+;; Ensure @llvm.fptosi.sat.* and @llvm.fptoui.sat.* intrinsics are translated
+
+; RUN: llvm-as %s -o %t.bc
+; RUN: llvm-spirv %t.bc --spirv-ext=+SPV_INTEL_arbitrary_precision_integers -spirv-text -o - | FileCheck %s --check-prefix=CHECK-SPIRV
+; RUN: llvm-spirv %t.bc --spirv-ext=+SPV_INTEL_arbitrary_precision_integers -o %t.spv
+; RUN: spirv-val %t.spv
+; RUN: llvm-spirv -r %t.spv -o %t.rev.bc
+; RUN: llvm-dis < %t.rev.bc | FileCheck %s --check-prefix=CHECK-LLVM
+
+target datalayout = "e-i64:64-v16:16-v24:32-v32:32-v48:64-v96:128-v192:256-v256:256-v512:512-v1024:1024-n8:16:32:64"
```
MrSidims, line 10: Data layout doesn't match the input LLVM IR ;-)

lib/SPIRV/SPIRVRegularizeLLVM.cpp
```diff
@@ -322,6 +322,104 @@ void SPIRVRegularizeLLVMBase::expandSYCLTypeUsing(Module *M) {
...
+// returned.
+// Rule 4: Otherwise, the result of rounding the argument towards zero is
+// returned.
+// Rules 1 & 4 are preserved when promoting iX to i64. For preserving Rule 2
+// and Rule 3, we saturate the result of the promoted instruction based on
+// original integer type (iX)
+// Example:
+// Input:
+// %0 = call i2 @llvm.fptosi.sat.i2.f32(float %input)
+// %1 = sext i32 %0
```
MrSidims, line 348: Would optimizer always extend generated integer and never use the original type?

lib/SPIRV/SPIRVRegularizeLLVM.cpp
```diff
@@ -322,6 +322,104 @@ void SPIRVRegularizeLLVMBase::expandSYCLTypeUsing(Module *M) {
...
+          auto IntBitWidth = II->getType()->getScalarSizeInBits();
+          if (IntBitWidth == 8 || IntBitWidth == 16 || IntBitWidth == 32 ||
+              IntBitWidth == 64)
+            continue;
+          if (IID == Intrinsic::fptosi_sat) {
+            // Identify sext (user of I). Make sure that's the only use of I.
+            auto *User = I.getUniqueUndroppableUser();
+            if (!User || !isa<SExtInst>(User))
+              continue;
+            auto *SExtI = dyn_cast<SExtInst>(User);
```
MrSidims, line 376: `SExtI` is actually unused in the code. Can we instead check for the integer size, if it's <32 bit, then create i32 version of the emulation and if User is ext instruction - put it to remove container. Same logic for 32 < X < 64 bit range.

lib/SPIRV/SPIRVRegularizeLLVM.cpp
```diff
@@ -322,6 +322,104 @@ void SPIRVRegularizeLLVMBase::expandSYCLTypeUsing(Module *M) {
...
+                II->getOperand(0));
+            Constant *MaxVal = ConstantInt::get(
+                NewIType, APInt::getSignedMaxValue(IntBitWidth).getSExtValue());
+            Constant *MinVal = ConstantInt::get(
+                NewIType, APInt::getSignedMinValue(IntBitWidth).getSExtValue());
+            auto *GTMax = IRB.CreateICmp(CmpInst::ICMP_SGE, NewII, MaxVal);
+            auto *LTMin = IRB.CreateICmp(CmpInst::ICMP_SLE, NewII, MinVal);
+            auto *SatMax = IRB.CreateSelect(GTMax, MaxVal, NewII);
+            auto *SatMin = IRB.CreateSelect(LTMin, MinVal, SatMax);
+            SExtI->replaceAllUsesWith(SatMin);
```
MrSidims, line 390: Moving this under: `for (Instruction *V : ToErase) {` would allow multiple-users case

lib/SPIRV/SPIRVRegularizeLLVM.cpp
```diff
@@ -322,6 +322,104 @@ void SPIRVRegularizeLLVMBase::expandSYCLTypeUsing(Module *M) {
...
+            ZExtI->replaceAllUsesWith(SatMax);
+            ToErase.push_back(ZExtI);
+            ToErase.push_back(II);
+          }
+        }
+      }
+    }
+    for (Instruction *V : ToErase) {
+      assert(V->user_empty());
+      V->eraseFromParent();
```
MrSidims, line 418:
```suggestion
      assert(V->user_empty());
      V->dropAllReferences();
      V->eraseFromParent();
```

test/llvm-intrinsics/fp_to_arbitrary_size_int_intrinsic.ll
```diff
@@ -0,0 +1,57 @@
+;; Ensure @llvm.fptosi.sat.* and @llvm.fptoui.sat.* intrinsics are translated
+
+; RUN: llvm-as %s -o %t.bc
+; RUN: llvm-spirv %t.bc --spirv-ext=+SPV_INTEL_arbitrary_precision_integers -spirv-text -o - | FileCheck %s --check-prefix=CHECK-SPIRV
```
MrSidims, line 4: Do we still need this extension?

lib/SPIRV/SPIRVRegularizeLLVM.cpp
LU-JOHN: > `SExtI` is actually unused in the code. Can we instead check for the integer size, if it's <32 bit, then create i32 version of the emulation and if User is ext instruction - put it to remove container. Same logic for 32 < X < 64 bit range.

Isn't this dangerous?   For example with i3, in the original code users would see values in the range [I3_MIN,I3_MAX].  With the suggested change, users will see values in the range [I32_MIN,I32_MAX].

MrSidims: Why is this dangerous? [I3_MIN,I3_MAX] aren't going anywhere, it just about not checking if a 'single' user is ext instruction and doing transformation regardless.

lib/SPIRV/SPIRVRegularizeLLVM.cpp
```diff
@@ -322,6 +322,104 @@ void SPIRVRegularizeLLVMBase::expandSYCLTypeUsing(Module *M) {
...
+// %2 = icmp sle i32 %0, -2 <Smallest 2-bit signed integer>
+// %3 = select i1 %1, i32 1, i32 %0
+// %4 = select i1 %2, i32 -2, i32 %3
+// Replace uses of %1 in Input with %4 in Output
+void SPIRVRegularizeLLVMBase::cleanupConversionToNonStdIntegers(Module *M) {
+  for (auto I = M->begin(), E = M->end(); I != E;) {
+    Function *F = &(*I++);
+    std::vector<Instruction *> ToErase;
+    for (BasicBlock &BB : *F) {
+      for (Instruction &I : BB) {
```
svenvh, line 361: This can be done more efficiently.  Instead of iterating over all(!) instructions just to find some intrinsic calls, please consider iterating over all users of the intrinsics that need processing.

lib/SPIRV/SPIRVRegularizeLLVM.cpp
LU-JOHN: Thx.  I misunderstood your proposed change.

lib/SPIRV/SPIRVRegularizeLLVM.cpp
asudarsa: Yes.

lib/SPIRV/SPIRVRegularizeLLVM.cpp
asudarsa: We do not expect multiple users here.

test/llvm-intrinsics/fp_to_arbitrary_size_int_intrinsic.ll
asudarsa: Nope. will remove. Thanks

lib/SPIRV/SPIRVRegularizeLLVM.cpp
asudarsa: There can be many intrinsics depending on the signature. But, I suppose we can surely try to parse over all function decls and see if any of them are fptosi.sat/fptoui.sat and process themaccordingly.
Can I please add a TODO for this optimization and address that in a later PR? Thanks for this suggestion.

lib/SPIRV/SPIRVRegularizeLLVM.cpp
asudarsa: Can you please show a IR code snippet example here? Looks like this might be a good point. But I am not fully grasping it. Thanks

asudarsa: FWIW, I initially had a solution with 'not considering' if there is single user. But, in that case I end up running into issues when I try to do replaceAllUsesWith (Type of old result will be iX and new result will be i32/i64 and replaceAllUsesWith wil complain about type mismatch. There are ugly workarounds, but we can avoid by just trying to find that sext/zext user. The optimization we are trying to undo will always create this single user, so we should be ok
Tahnks

MrSidims: If the sole purpose is to handle `AggressiveInstCombine` output for scalar conversions I'm OK with that patch.
If the purpose is different - please see clang's eimplementation of `ScalarExprEmitter` - it doesn't generate any integer extensions for intrinsic's result.

test/llvm-intrinsics/fp_to_arbitrary_size_int_intrinsic.ll
asudarsa: I looked at the original testcase. Data layout is exactly the same as it is here. Can you please let me know how it does not match the input LLVM IR? Thanks

lib/SPIRV/SPIRVRegularizeLLVM.cpp
asudarsa: Yes. handling AggressiveInstCombine is the primary purpose here. Thanks

test/llvm-intrinsics/fp_to_arbitrary_size_int_intrinsic.ll
MrSidims: > n8:16:32:64

There is no "2" and "4" which would stand for i2 and i4

MrSidims: Or if not hacks, may be just aggressive inst combine should respect data layout?

lib/SPIRV/SPIRVRegularizeLLVM.cpp
svenvh: > But, I suppose we can surely try to parse over all function decls

There is no need for parsing. You already have the `Function` object, so you can call `F->getIntrinsicID()`.  If the IntrinsicID matches what you're interested in, you then iterate over `F->users()`.

> Can I please add a TODO for this optimization and address that in a later PR? Thanks for this suggestion.

If you don't intend to rework the loops in this PR, please at least add an early exit based on `F->getIntrinsicID()` in the outer loop to avoid every translation needlessly iterating over all instructions.

asudarsa: I think early exit may not be applicable here as we want to parse functions that contain uses of such intrinsics.

I might have over-thought the resolution here though. You make a very good point. We do already have a handle to F that we can use here. I will try to put up a change here. 

Thanks so much again for feedback.

svenvh: Awesome, thanks for addressing this!

test/llvm-intrinsics/fp_to_arbitrary_size_int_intrinsic.ll
asudarsa: It seems the entries in datalayout are expected to be more of a guide that strict rules. for example, It is not documented anywhere that i2 is illegal if it is not listed in the target datalayout, thought it does sound counter-intuitive. I think we might have to assume that this LLVM IR is legal and proceed with the change here.

Thanks

etiotto: > It seems the entries in datalayout are expected to be more of a guide that strict rules. for example, It is not documented anywhere that i2 is illegal if it is not listed in the target datalayout, thought it does sound counter-intuitive. I think we might have to assume that this LLVM IR is legal and proceed with the change here.
> 
> Thanks

I agree. The fact the `datalayout` field does not explicitly "mention" `i2` doesn't prevent the LLVM optimizer to use i2 as an intermediat type during optimization IMO. The LLVM language spec simply states that the " data layout string that specifies how data is to be laid out in memory". There isn't anything that says using i2 is illegal.

MrSidims (approved):

asudarsa:
Thanks everyone for the reviews.

PR #2508, opened 2024-04-18
Revert "Turn on SPIR-V builtin generation for OpenCL CPP sources (#2466)"

bwlodarcz:
This reverts commit 80dfd864ae556369747074bbde8fc1d4c48547be. The commit breaks Translator in cases when user supplied function is named with the same prefix, after demangling, as OpenCL builtin. The culprit code is in SPIRWriter OCL pass in function OCLToSPIRVBase::visitCallInst(CallInst &).
Failing name would be for example atomic_fetch_and_sub_uint32_explicit.


MrSidims (approved):

svenvh (approved):
Given that #2512 still doesn't solve the problem for user function declarations, I think we should proceed with the revert and think of a more robust solution to #2513.

asudarsa:
I am trying to check if this is a OpenCL C++ specific problem or if this issue can be reproduced in OpenCL C. Will revert once confirmed as a OpenCL C++ specific problem. Thanks

asudarsa:
Hi @svenvh,

We tested my change https://github.com/KhronosGroup/SPIRV-LLVM-Translator/pull/2512 and all our existing test failures are resolved. This particular change (OCLToSPIRV for OpenCL C++) is important for one of our compilation flows. I would prefer to keep this in and continue discussion about covering user-declared functions.
Please let me know if that' acceptable.

Thanks

MrSidims:
> is important for one of our compilation flows

Can it be moved to downstream then?

PR #2512, opened 2024-04-18
[OClToSPIRV] Do not identify user defined functions as OpenCL builtins

asudarsa:
There are scenarios where users define functions with names as 'atomic_fetch_and_sub_uint32'. Such functions are being wrongly identified as OpenCL builtins.
This change adds the following to OCLIsBuiltin(...) function
if (!F->isDeclaration()) return false;

This resolves several existing test failures.

Thanks


asudarsa:
A better solution has been submitted here: https://github.com/KhronosGroup/SPIRV-LLVM-Translator/pull/2525
Once approved, I will close this one.

Thanks

PR #2525, opened 2024-04-22
Add check to early exit when OpenCL builtin is not found in the map

asudarsa:
Several user defined/declared functions were being identified as 'possible' OpenCL builtins as they had prefixes (e.g. atomic_) similar to OpenCL builtin function names. However, they were not present in the OCLSPIRVBuiltinMap.
One of the OCLToSPIRV utility functions calls OCLSPIRVBuiltinMap::map function on such user defined functions and this results in a crash.
The correct way to do this will be to check if the OCL function name key is present in the map and exit early if not.

This PR does that.

Thanks

### DIFF
--- a/lib/SPIRV/OCLToSPIRV.cpp
+++ b/lib/SPIRV/OCLToSPIRV.cpp
@@ -650,8 +650,10 @@ void OCLToSPIRVBase::visitCallAtomicCpp11(CallInst *CI, StringRef MangledName,
 void OCLToSPIRVBase::transAtomicBuiltin(CallInst *CI,
                                         OCLBuiltinTransInfo &Info) {
   llvm::Type *AtomicBuiltinsReturnType = CI->getType();
-  auto SPIRVFunctionName =
-      getSPIRVFuncName(OCLSPIRVBuiltinMap::map(Info.UniqName));
+  Op OC = OpNop;
+  if (!OCLSPIRVBuiltinMap::find(Info.UniqName, &OC))
+    return;
+  auto SPIRVFunctionName = getSPIRVFuncName(OC);
   bool NeedsNegate = false;
   if (AtomicBuiltinsReturnType->isFloatingPointTy()) {
     // Translate FP-typed atomic builtins. Currently we only need to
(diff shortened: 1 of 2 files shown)


asudarsa:
This PR should help us to get https://github.com/KhronosGroup/SPIRV-LLVM-Translator/pull/2500 back by reverting https://github.com/KhronosGroup/SPIRV-LLVM-Translator/pull/2508

Thanks

svenvh (approved):

bwlodarcz:
The fix of the issue is likely a little bigger than "atomic". The changes covers "atomic" prefix but there are other prefixes in `OCLToSPIRVBase::visitCallInst` in OCLToSPIRV.cpp:196 like "convert" or "VLoad" which needs to be handled with care as well. I didn't do deep dive but by quickly glimpsing the code of the function handling that cases it's imaginable to lead Translator again into bogus state by carefully crafted function name. What do you think @MrSidims, @asudarsa?

MrSidims (approved):

asudarsa:
> The fix of the issue is likely a little bigger than "atomic". The changes covers "atomic" prefix but there are other prefixes in `OCLToSPIRVBase::visitCallInst` in OCLToSPIRV.cpp:196 like "convert" or "VLoad" which needs to be handled with care as well. I didn't do deep dive but by quickly glimpsing the code of the function handling that cases it's imaginable to lead Translator again into bogus state by carefully crafted function name. What do you think @MrSidims, @asudarsa?

Hi @bwlodarcz

Thanks much for the feedback.
I am now trying several tests trying to manually create the user defined functions with OCL looking names. I will update the PR with my findings.

Sincerely

VyacheslavLevytskyy (approved):

asudarsa:
> The fix of the issue is likely a little bigger than "atomic". The changes covers "atomic" prefix but there are other prefixes in `OCLToSPIRVBase::visitCallInst` in OCLToSPIRV.cpp:196 like "convert" or "VLoad" which needs to be handled with care as well. I didn't do deep dive but by quickly glimpsing the code of the function handling that cases it's imaginable to lead Translator again into bogus state by carefully crafted function name. What do you think @MrSidims, @asudarsa?

The transBuiltin(...) function already has the check to see if the function we are looking at has a valid OCL builtin (See https://github.com/KhronosGroup/SPIRV-LLVM-Translator/blob/5e53bcc9fb15c7ac33e947bfb8919b5d71dbeff9/lib/SPIRV/OCLToSPIRV.cpp#L915).
This check was missing in transAtomicBuiltin(...) and this PR adds that check. I do not see any other site where this check is required. So, I think this PR can be submitted as is.

Thanks for all the feedback.

Issue #2509, opened 2024-04-18
`ID '19' decorated with NoSignedWrap multiple times is not allowed.`


ben-clayton:
With the latest Vulkan Validation Layers, the internal Google build / test is now failing with:

```
> spirv-val OpPhi_ArgumentsPlaceholders.ll

error: line 49: ID '19' decorated with NoSignedWrap multiple times is not allowed.
  %inc = OpIAdd %uint %j_0 %uint_1

Command failed: exit status 1
```

Note: The new no duplicate decorations validation has triggered for many other SPIR-V projects.

svenvh:
All issues with duplicate decorations should now be fixed on the `main` branch.

PR #2529, opened 2024-04-22
Skip adding decorations for OpForward

svenvh:
When a temporary `OpForward` instruction is needed during translation to SPIR-V, do not add the decorations yet, as that would result in duplicate decorations when the actual instruction is visited and the `OpForward` is replaced by a real SPIR-V instruction.

The SPIR-V Validator has recently started checking for duplicate decorations; this fixes some but not all issues arising from the new checks.

Contributes to https://github.com/KhronosGroup/SPIRV-LLVM-Translator/issues/2509 .

### DIFF
--- a/lib/SPIRV/SPIRVWriter.cpp
+++ b/lib/SPIRV/SPIRVWriter.cpp
@@ -1454,7 +1454,12 @@ SPIRVValue *LLVMToSPIRVBase::transValue(Value *V, SPIRVBasicBlock *BB,
          "Invalid SPIRV BB");
 
   auto *BV = transValueWithoutDecoration(V, BB, CreateForward, FuncTrans);
-  if (!BV || !transDecoration(V, BV))
+  if (!BV)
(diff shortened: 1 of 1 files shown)


vmaksimo (commented):
Should we enable `spirv-val` checks in some of our tests if they have such a problem? Or even it'd be nice to see a separate test for the fix.

svenvh:
> Should we enable `spirv-val` checks in some of our tests if they have such a problem? Or even it'd be nice to see a separate test for the fix.

With the latest version of `spirv-val` the `test/transcoding/OpPhi_ArgumentsPlaceholders.ll` test currently fails validation, see #2509.  This PR fixes that test.

We don't see the failure in GitHub CI yet because it uses an older version of `spirv-val`.

MrSidims (approved):

vmaksimo (approved):

LU-JOHN (approved):
LGTM

PR #2532, opened 2024-04-23
Incorporate SPV_KHR_non_semantic_info into SPIR-V 1.6

vmaksimo:


MrSidims (approved):

lib/SPIRV/LLVMToSPIRVDbgTran.cpp
```diff
@@ -284,8 +284,12 @@ SPIRVEntry *LLVMToSPIRVDbgTran::transDbgEntryImpl(const MDNode *MDN) {
   if (!MDN)
     return BM->addDebugInfo(SPIRVDebug::DebugInfoNone, getVoidTy(),
                             SPIRVWordVec());
-  if (isNonSemanticDebugInfo())
-    BM->addExtension(SPIRV::ExtensionID::SPV_KHR_non_semantic_info);
+  if (isNonSemanticDebugInfo()) {
```
LU-JOHN, line 287: This is indirectly related to this PR, but this seems like a costly place to do this check.   Can this check be moved to the beginning of transDebugMetadata?

vmaksimo: @LU-JOHN I am a bit confused why do we need to do that? I mean, we don't need to add extension/version number earlier than we actually meet not-empty nonsemantic instruction.

vmaksimo: It does seem relevant for almost all the cases, except the one where we have empty `MDN` node as an input in `transDbgEntryImpl()`. In case we are adding `DebugInfoNone`, we should not add any extension earlier. And if the compile unit has only such instructions it'd be wrong to add such extension.
Please let me know if I'm missing something here, maybe I didn't understand your idea fully.

vmaksimo: Sorry, you're right. We will always enter `transDbgEntryImpl()` with non-empty `DICompileUnit`, and the check for non-semantic debug info can be done earlier.
Applied the suggestion.

LU-JOHN (commented):
LGTM

LU-JOHN (approved):

PR #2534, opened 2024-04-23
[Backport to 18] Switch to SPIR-V 1.5

vmaksimo:


lib/SPIRV/libSPIRV/SPIRVModule.cpp
```diff
@@ -58,6 +58,18 @@
 
 namespace SPIRV {
 
+namespace {
+std::string to_string(uint32_t Version) {
```
vmaksimo, line 62: @svenvh what do you thing, should we fix clang-tidy suggestion here (and in main branch to be aligned)? to rename `to_string` to `toString`

svenvh (approved):

lib/SPIRV/libSPIRV/SPIRVModule.cpp
vmaksimo: Agree to ignore here and whenever it's needed :) Thanks!

MrSidims (approved):

PR #2537, opened 2024-04-24
Avoid duplicate Alignment decorations

svenvh:
The SPIR-V Validator has recently started checking for duplicate decorations.  This commit fixes duplicate Alignment decorations that affected the `test/read_image.cl` test.

Alignment decorations have two potential sources during LLVM to SPIR-V translation: the instruction's alignment property and `spirv.Decorations` metadata.  Handle both of these through the `setAlignment` method, so that duplicates can be avoided.

Calling `setAlignment` with different alignments for the same entity is probably an error, so add an assert.

Contributes to https://github.com/KhronosGroup/SPIRV-LLVM-Translator/issues/2509

### DIFF
--- a/lib/SPIRV/libSPIRV/SPIRVValue.cpp
+++ b/lib/SPIRV/libSPIRV/SPIRVValue.cpp
@@ -51,6 +51,14 @@ void SPIRVValue::setAlignment(SPIRVWord A) {
     eraseDecorate(DecorationAlignment);
     return;
   }
+  SPIRVWord PrevAlignment;
+  if (hasAlignment(&PrevAlignment)) {
+    // Do nothing if the Id already has an Alignment decoration, provided
+    // it matches the new alignment.
+    assert(A == PrevAlignment &&
+           "New alignment does not match existing alignment");
+    return;
+  }
   addDecorate(new SPIRVDecorate(DecorationAlignment, this, A));
   SPIRVDBG(spvdbgs() << "Set alignment " << A << " for obj " << Id << "\n")
 }
--- a/lib/SPIRV/SPIRVWriter.cpp
+++ b/lib/SPIRV/SPIRVWriter.cpp
@@ -1060,7 +1060,7 @@ void LLVMToSPIRVBase::transVectorComputeMetadata(Function *F) {
   }
 }
 
-static void transMetadataDecorations(Metadata *MD, SPIRVEntry *Target);
+static void transMetadataDecorations(Metadata *MD, SPIRVValue *Target);
 
 void LLVMToSPIRVBase::transFPGAFunctionMetadata(SPIRVFunction *BF,
                                                 Function *F) {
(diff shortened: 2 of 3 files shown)


svenvh:
The clang-tidy failure is a pre-existing issue.

lib/SPIRV/libSPIRV/SPIRVValue.cpp
```diff
@@ -51,6 +51,14 @@ void SPIRVValue::setAlignment(SPIRVWord A) {
     eraseDecorate(DecorationAlignment);
     return;
   }
+  SPIRVWord PrevAlignment;
+  if (hasAlignment(&PrevAlignment)) {
+    // Do nothing if the Id already has an Alignment decoration, provided
+    // it matches the new alignment.
+    assert(A == PrevAlignment &&
```
MrSidims, line 58: Mmm, some would argue, saying that if 2 sources provides different alignment, then we should use the biggest among them. I personally slightly leaning to not allow such case (aka keep assert here), but can imagine seeing complains about it. WDYT?

svenvh: That's the alternative approach I had considered, but it feels a bit wrong because different alignments kind of means there is a contradiction in the input.  So I first tried with an assert and the test suite passed.  If there is a good reason for allowing different alignments then I'm happy to drop the assert.

lib/SPIRV/libSPIRV/SPIRVValue.cpp
```diff
@@ -51,6 +51,14 @@ void SPIRVValue::setAlignment(SPIRVWord A) {
     eraseDecorate(DecorationAlignment);
     return;
   }
+  SPIRVWord PrevAlignment;
```
asudarsa, line 54: This might not be a relevant issue. But I will go ahead and ask. If we had A==0 and the prevAlignment != 0, do we want to error out or just delete the decoration?

Thanks

asudarsa (commented):
Change looks good to me except for a clarifying question I had. Also, is the existing test sufficient? Will we benefit from a new LIT test here?

Thanks

lib/SPIRV/libSPIRV/SPIRVValue.cpp
svenvh: The current behaviour with `A == 0` is to remove all existing Alignment decorations and my patch doesn't change this.  I think that behaviour is fine for now.

svenvh:
I've added a test case for just this issue, as it was probably by chance that `test/read_image.cl` got affected.

asudarsa (approved):
LGTM. Thanks for adding the test.

lib/SPIRV/libSPIRV/SPIRVValue.cpp
MrSidims: Well, good thing about such assertion - we can always immediately get a notification of such use case and remove the assertion if necessary, so lets keep it as is :)

MrSidims (approved):

PR #2543, opened 2024-04-25
[NFC] Document intrinsic lowering

LU-JOHN:
Document how llvm intrinsics are lowered by SPIRV-LLVM-Translator.


MrSidims (commented):
Thanks for working on this! I'll post extra comments later. There are 2 cases missing:
1. Just mapping on a SPIR-V instruction from either core spec or extension. Q: do we want to mention this case?
2. Some emulation done straight in SPIR-V writer, so not intermediate transformation is being done.

General Q: do we want to unify approaches? All of them have pros and cons, for example if we need to do some types adjustments, it's either to do it just during SPIR-V generation as we create module from scratch, on LLVM IR level replacing types is much harder. Also while inserting definitions of intrinsic function overloads is also possible, do we really want to do it in every case as we would need to insert every possible type combination?

Another Q: do we want (and if yes or no - mention it in the document) emulate some intrinsic that are being mapped on OpenCL lib calls or extension instruction in case, if lets say we know, that target device supports only native builtins or doesn't support a particular extension?

asudarsa:
Thanks so much for adding this @LU-JOHN 
High level comment - I think you should add this under the 'docs' directory. 

I will add my feedback in a bit.

Thanks

lib/SPIRV/Lowering.md
```diff
@@ -0,0 +1,132 @@
...
+          case Intrinsic::ctlz:
+          case Intrinsic::cttz: {
+            SPIRVWord ExtOp = IID == Intrinsic::ctlz ? OpenCLLIB::Clz : OpenCLLIB::Ctz;
+            SPIRVType *Ty = transType(II->getType());
+            std::vector<SPIRVValue *> Ops(1, transValue(II->getArgOperand(0), BB));
+            return BM->addExtInst(Ty, BM->getExtInstSetId(SPIRVEIS_OpenCL), ExtOp, Ops,
+                                  BB);
+          }
+
+   When these ExtInst are reverse translated they are converted to calls:
```
asudarsa, line 24: Can you please document what happens when reverse translation is called with --spirv-target-env=SPV-IR i.e the user requests SPIR-V friendly IR?

lib/SPIRV/Lowering.md
```diff
@@ -0,0 +1,132 @@
+The SPIRV-LLVM-Translator will "lower" some LLVM intrinsic calls to another function or implementation
```
asudarsa, line 1: Is it possible to document what logic/criteria is used to select between these three options?

Thanks

LU-JOHN:
> Thanks for working on this! I'll post extra comments later. There are 2 cases missing:
> 
> 1. Just mapping on a SPIR-V instruction from either core spec or extension. Q: do we want to mention this case?
> 2. Some emulation done straight in SPIR-V writer, so not intermediate transformation is being done.
> 
> General Q: do we want to unify approaches? All of them have pros and cons, for example if we need to do some types adjustments, it's either to do it just during SPIR-V generation as we create module from scratch, on LLVM IR level replacing types is much harder. Also while inserting definitions of intrinsic function overloads is also possible, do we really want to do it in every case as we would need to insert every possible type combination?
> 
> Another Q: do we want (and if yes or no - mention it in the document) emulate some intrinsic that are being mapped on OpenCL lib calls or extension instruction in case, if lets say we know, that target device supports only native builtins or doesn't support a particular extension?



> Thanks for working on this! I'll post extra comments later. There are 2 cases missing:
> 
> 1. Just mapping on a SPIR-V instruction from either core spec or extension. Q: do we want to mention this case?
> 2. Some emulation done straight in SPIR-V writer, so not intermediate transformation is being done.
> 
> General Q: do we want to unify approaches? All of them have pros and cons, for example if we need to do some types adjustments, it's either to do it just during SPIR-V generation as we create module from scratch, on LLVM IR level replacing types is much harder. Also while inserting definitions of intrinsic function overloads is also possible, do we really want to do it in every case as we would need to insert every possible type combination?
> 
> Another Q: do we want (and if yes or no - mention it in the document) emulate some intrinsic that are being mapped on OpenCL lib calls or extension instruction in case, if lets say we know, that target device supports only native builtins or doesn't support a particular extension?

1. I've documented lowering with an instruction in an extension.
2. Emulation done in SPIRVWriter.cpp.

I think we do want to simplify the lowering process, but that is a bigger discussion than what this document addresses.  I'm only trying to give an overview of what is currently done.

lib/SPIRV/Lowering.md
LU-JOHN: I've added documentation to explain the pros/cons of each approach.  The approaches are ordered from best to worst in my opinion.

lib/SPIRV/Lowering.md
LU-JOHN: I've added a little documentation, but this is just duplicating text from the --help option.   I do not understand enough to explain why --spirv-target-env=SPV-IR should be used.  Perhaps someone else can re-write this seciton.

LU-JOHN:
> Thanks so much for adding this @LU-JOHN High level comment - I think you should add this under the 'docs' directory.
> 
> I will add my feedback in a bit.
> 
> Thanks

Moved to docs directory.

docs/Lowering.md
```diff
@@ -0,0 +1,230 @@
+The SPIRV-LLVM-Translator will "lower" some LLVM intrinsic calls to another function or implementation
```
MrSidims, line 1: Lets rename the file from Lowering to LLVMIntrinsicFunctionsLowering or something similar

docs/Lowering.md
```diff
@@ -0,0 +1,230 @@
...
+        5 BitwiseOr 8 46 44 45
+        2 ReturnValue 46
+
+   In reverse translation, the lowering is undone.  Calls are reverted to the original llvm.bswap.i16 intrinsic
+
+        %ret = call i16 @llvm.bswap.i16(i16 %0)
+
+   The emulation functions are deleted.
+
+   This is the next best way of lowering a function.  The functionality of the intrinsic is created by a call to LLVM's CreateIntrinsic,
```
MrSidims, line 176: It's great to spell out pros and cons of each approach. But TBH I'd prefer to avoid saying "the next best way" or "least preferable way" as it might make a reader to lean towards one way of lowering instead of another despite the fact, that another approach might be more suitable in their case.

LU-JOHN: Removed comments rating alternatives.  Let reader decide.

docs/LLVMIntrinsicFunctionLowering.md
```diff
@@ -1,7 +1,9 @@
 The SPIRV-LLVM-Translator will "lower" some LLVM intrinsic calls to another function or implementation
-in one of three ways:
+in one of four ways:
 
-1. In transIntrinsicInst in SPIRVWriter.cpp, calls to LLVM intrinsics are replaced with a SPIRV ExtInst.
+1. A.
+
+   In transIntrinsicInst in SPIRVWriter.cpp, calls to LLVM intrinsics are replaced with a SPIRV ExtInst.
```
asudarsa, line 6:
```suggestion
Method 1

   Variation A: In transIntrinsicInst in SPIRVWriter.cpp, calls to LLVM intrinsics are replaced with a SPIRV ExtInst.
```

docs/LLVMIntrinsicFunctionLowering.md
```diff
@@ -1,7 +1,9 @@
 The SPIRV-LLVM-Translator will "lower" some LLVM intrinsic calls to another function or implementation
-in one of three ways:
+in one of four ways:
```
asudarsa, line 2:
```suggestion
using one of the following four methods:
```

docs/LLVMIntrinsicFunctionLowering.md
```diff
@@ -0,0 +1,230 @@
...
+
+1. A.
+
+   In transIntrinsicInst in SPIRVWriter.cpp, calls to LLVM intrinsics are replaced with a SPIRV ExtInst.
+   For example:
+
+        %0 = tail call i32 @llvm.ctlz.i32(i32 %x, i1 true)
+
+   is translated into SPIRV with an SPIRV ExtInst clz:
```
asudarsa, line 11: Is this not an OpenCL ExtInst?

VyacheslavLevytskyy (approved):

docs/LLVMIntrinsicFunctionLowering.md
LU-JOHN: Done.

docs/LLVMIntrinsicFunctionLowering.md
LU-JOHN: Done.

docs/LLVMIntrinsicFunctionLowering.md
LU-JOHN: Changed to OpenCL.

docs/LLVMIntrinsicFunctionLowering.md
```diff
@@ -67,7 +69,9 @@ in one of four ways:
             return BM->addUnaryInst(OpBitReverse, Ty, Op, BB);
           }
 
-2. Some intrinsics are emulated by basic operations in SPIRVWriter.cpp.  For example:
+Method 2:
+
+   Some intrinsics are emulated by basic operations in SPIRVWriter.cpp.  For example:
```
asudarsa, line 74: Nit: I think it is better to not refer to cpp files. Here you can just refer to SPIRVWriter pass. Also, please try to avoid using cpp filenames in the text below as well.

Thanks

asudarsa (approved):
Thanks @LU-JOHN for addressing all the comments.

Looks good. Just a nit.

docs/LLVMIntrinsicFunctionLowering.md
LU-JOHN: Done.

MrSidims (approved):

PR #2554, opened 2024-05-06
Disable new debug record format

svenvh:
After LLVM commit 91446e2aa687 ("Repply#2 "[RemoveDIs] Load into new debug info format by default in LLVM (#89799)"", 2024-05-03), we need to disable the new debug record format explicitly until we have migrated to the new format.


LU-JOHN (approved):
This is necessary for testing pass with newer LLVM.

PR #2561, opened 2024-05-15
cmake: Add spirv-dis lit feature

svenvh:
Various tests already declare `REQUIRES: spirv-dis`, but this feature was never enabled so those tests were always skipped as "unsupported".


MrSidims (approved):
Thanks!

PR #2564, opened 2024-05-16
[CI] Fix code style checks by installing libllvmlibc

svenvh:
LLVMExports.cmake from already installed packages references files from the libllvmlibc package.


svenvh:
I should have realized in #2563 that both the out-of-tree and the code style checker jobs were affected...

MrSidims (approved):

PR #2565, opened 2024-05-16
Support for record debug format

bwlodarcz:
Recently, LLVM migrated to new debug format replacing debug intrinsics with so-called records. The DIBuilder api supports both formats now with grace deprecation period for intrinsics. Currently Translator during reverse translation have new format disabled by flag and still supports only old one.
Documentation for the migration: https://llvm.org/docs/RemoveDIsDebugInfo.html


asudarsa:
LGTM. Thanks @bwlodarcz . Please take a look at the build fail. Seems like an easy fix. Thanks

svenvh (approved):
LGTM, thanks!

> Please take a look at the build fail. Seems like an easy fix.

Out-of-tree build issues will be solved by #2574

PR #2570, opened 2024-05-17
[Backport to 14] Support both legacy and Khronos version of global va…

mateuszchudyk:
…riable extensions (#2228)

This is done to provide smooth drivers transition to official extensions.
Also updated tests for new extensions to use new tokens after https://github.com/KhronosGroup/SPIRV-Headers/commit/cca08c63cefa129d082abca0302adcb81610b465


MrSidims:
@mateuszchudyk please also update other branches

PR #2576, opened 2024-05-21
[SPV_INTEL_bindless_images] Add ret type to convert-SampledImage builtin name

wenju-he:
Similar as ConvertHandleToImageINTEL, ConvertHandleToSampledImageINTEL builtin may have the same argument type but different return types. So we need to add ret type as name suffix when translating to LLVM IR.


MrSidims (approved):
CI failures are unrelated

Issue #1486, opened 2022-05-27
Twice as many kernel arguments as specified in SPIR


jonasbhjulstad:
[This issue](https://github.com/llvm/llvm-project/issues/55675) shows that there are problems with the integration of llvm-spirv when used to compile `-cl-std=clc++2021` with clang. The number of input arguments recognized for kernel functions (when loading SPIRV in OpenCL) are twice the number specified in the corresponding c++ code.

Compilation works fine for the following revisions:
```
clang revision: 0045d01af96ff56c5b62d133be076020a33867ff
llvm-spirv revision: e8f1964d0e58aacfe39c8ce7b43b64d6cabaed49
```
But not for the newest releases of Clang and LLVM-SPIRV.

AnastasiaStulova:
Btw I was surprised to see the wrapper for such case but my guess is that there is no easy way to do something more optimal, perhaps even with some help from the fronted? My guess is that when the kernel is defined in one translation unit but called in another it is not detectable? I think we had a thread at some point that clang might emit 2 version for each kernel one to use in call expressions and another to use as entry point… but this would lead to code duplication.

Also I was running spirv-opt on the test case from this issue and the function was not inlined into the wrapper but I think such optimisation should still happen when SPIR-V is compiled down to machine instructions?

AnastasiaStulova:
Another surprising thing to me is that such simple case of calling a function from the kernel is not handled by the driver… my guess is that when this comes from llvm-spirv transformation it has slightly different format compared to when similar code generated through the frontend + a 1:1 conversation… not sure if it makes sense for such transformations to match the format of equivalent constructs written in a high-level language…

svenvh:
> there is no easy way to do something more optimal, perhaps even with some help from the fronted? My guess is that when the kernel is defined in one translation unit but called in another it is not detectable?

I don't see how the translator could handle the various multi-translation-unit scenarios indeed, let alone mixed source and IL scenarios.

> Also I was running spirv-opt on the test case from this issue and the function was not inlined into the wrapper but I think such optimisation should still happen when SPIR-V is compiled down to machine instructions?

I'd expect that spirv-opt cannot undo this transformation, because then it would have to generate invalid SPIR-V (which is what #1149 fixed).  How it gets transformed into machine instructions is up to the driver implementation, but `SPIRVReader.cpp` indeed drops the wrapper kernel and promotes the function back to a kernel when translating to LLVM IR.

[...]

PR #2585, opened 2024-05-24
Add a new flag "--no-entry"

lucasz93:
Add a new flag "--no-entry" which disables kernel entry point generation. This is a workaround for some OpenCL drivers that are incompatible with entry points, leading them to have twice the number of kernel arguments. See https://github.com/KhronosGroup/SPIRV-LLVM-Translator/issues/1486 for more details.

Adding this option allows a single toolchain to be used when compiling kernels for all devices. Some real world instances of this problem:
https://community.intel.com/t5/GPU-Compute-Software/OpenCL-CL-INVALID-KERNEL-ARGS-with-driver-gt-30-0-101-1340/td-p/1420021
https://community.intel.com/t5/GPU-Compute-Software/OpenCL-Kernel-Name-appends-quot-1-quot/m-p/1419720#M644

The workaround is to use ocloc but that adds another toolchain to dev environments and it doesn't have CLC++ support in master just yet.

### DIFF
--- a/lib/SPIRV/SPIRVWriter.cpp
+++ b/lib/SPIRV/SPIRVWriter.cpp
@@ -871,7 +871,7 @@ SPIRVFunction *LLVMToSPIRVBase::transFunctionDecl(Function *F) {
       static_cast<SPIRVFunction *>(mapValue(F, BM->addFunction(BFT)));
   BF->setFunctionControlMask(transFunctionControlMask(F));
   if (F->hasName()) {
-    if (isKernel(F)) {
+    if (isKernel(F) && BM->getGenerateKernelEntryPoints()) {
       /* strip the prefix as the runtime will be looking for this name */
       std::string Prefix = kSPIRVName::EntrypointPrefix;
       std::string Name = F->getName().str();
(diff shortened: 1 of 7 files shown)


MrSidims (approved):
Understandable, I'm fine with the option. Please create a test for it.

tools/llvm-spirv/llvm-spirv.cpp
```diff
@@ -278,6 +278,10 @@ static cl::opt<SPIRV::BuiltinFormat> SPIRVBuiltinFormat(
         clEnumValN(SPIRV::BuiltinFormat::Global, "global",
                    "Use globals to represent SPIR-V builtin variables")));
 
+static cl::opt<bool>
+    NoEntry("no-entry", cl::init(false),
```
svenvh, line 282: For consistency with other llvm-spirv options:

```suggestion
    NoEntry("spirv-no-entry", cl::init(false),
```

lib/SPIRV/SPIRVWriter.cpp
```diff
@@ -5635,7 +5635,9 @@ void LLVMToSPIRVBase::transFunction(Function *I) {
 
   if (isKernel(I)) {
     auto Interface = collectEntryPointInterfaces(BF, I);
-    BM->addEntryPoint(ExecutionModelKernel, BF->getId(), BF->getName(),
+    auto Name =
+        BM->getGenerateKernelEntryPoints() ? BF->getName() : I->getName().str();
+    BM->addEntryPoint(ExecutionModelKernel, BF->getId(), Name,
```
LU-JOHN, line 5640: Is this change necessary?  If !getGenerateKernelEntryPoints, then BF->getName()==I->getName().str()  since BF->getName() will be set by line 885.

PR #2587, opened 2024-05-26
Re-implement Intel cache controls decoration

MrSidims:
The patch adds CacheControls(Load/Store)INTEL decorations representation as metadata placed on memory accessing instruction with the following form: !spirv.DecorationCacheControlINTEL !X
!X = !{i32 %decoration_kind%, i32 %level%, i32 %control%,
       i32 %operand of the instruction to decorate%}
It's being done this way as it's less likely for the metadata attached to function calls or memory instructions being optimized out than in a case when it's attached to pointer in LLVM IR.

Also patch creates a dummy GEP accessing pointer operand of the instruction to perform SSA copy of the pointer and attaches !spirv.Decorations metadata to the GEP, so several memory access instructions with different cache locality would use different decorated copies of the same pointer.

Few notes about this implementation and other options. It is sub-optimal as it requires iteration over all instructions in the module. Alternatives are:
1. In SPIRVWriter during transDecorations rewrite already translated instruction (lets say load) with the new one, but with GEP operand. But while the translator provides API to erase instructions and rewriting SPIRVBasicBlocks, yet unfortunately it's not really usable outside of of SPIRVModule context;
2. In SPIRVWriter during transValueWithoutDecoration add special handling of the instructions with spirv.DecorationCacheControlINTEL metadata. And it would work with one exception - it also would require to unprivate some methods of SPIRVScavenger to create GEP with the correct type, which I'm not sure if it's a good idea. Note, in this implementation in the worst case scenario SPIRVScavenger would create a bitcast later during translation.

### DIFF
--- a/test/extensions/INTEL/SPV_INTEL_cache_controls/basic_load_store_arg_dec.ll
+++ b/test/extensions/INTEL/SPV_INTEL_cache_controls/basic_load_store_arg_dec.ll
@@ -1,0 +1,66 @@
+; RUN: llvm-as %s -o %t.bc
+; RUN: llvm-spirv --spirv-ext=+SPV_INTEL_cache_controls -spirv-text %t.bc -o - | FileCheck %s --check-prefix=CHECK-SPIRV
+; RUN: llvm-spirv --spirv-ext=+SPV_INTEL_cache_controls %t.bc -o %t.spv
+; RUN: llvm-spirv -r %t.spv --spirv-target-env=SPV-IR -o - | llvm-dis -o - | FileCheck %s --check-prefix=CHECK-LLVM
+
+; CHECK-SPIRV-DAG: TypeInt [[#Int32:]] 32 0
+; CHECK-SPIRV-DAG: Constant [[#Int32]] [[#Zero:]] 0
+; CHECK-SPIRV-DAG: Decorate [[#Load1Ptr:]] CacheControlLoadINTEL 0 1
+; CHECK-SPIRV-DAG: Decorate [[#Load2Ptr:]] CacheControlLoadINTEL 1 1
+; CHECK-SPIRV-DAG: Decorate [[#Store1Ptr:]] CacheControlStoreINTEL 0 1
+; CHECK-SPIRV-DAG: Decorate [[#Store2Ptr:]] CacheControlStoreINTEL 1 1
+
+; CHECK-SPIRV: FunctionParameter [[#]] [[#Buffer:]]
+; CHECK-SPIRV: PtrAccessChain [[#]] [[#Load1GEPPtr:]] [[#Buffer]] [[#Zero]]
+; CHECK-SPIRV: Load [[#]] [[#]] [[#Load1GEPPtr]]
+; CHECK-SPIRV: PtrAccessChain [[#]] [[#Load2GEPPtr:]] [[#]] [[#Zero]]
+; CHECK-SPIRV: Load [[#]] [[#]] [[#Load2GEPPtr]]
+; CHECK-SPIRV: PtrAccessChain [[#]] [[#Store1GEPPtr:]] [[#]] [[#Zero]]
+; CHECK-SPIRV: Store [[#Store1GEPPtr]]
+; CHECK-SPIRV: PtrAccessChain [[#]] [[#Store2GEPPtr:]] [[#]] [[#Zero]]
+; CHECK-SPIRV: Store [[#Store2GEPPtr]]
+
+; CHECK-LLVM: %[[#GEPLoad1:]] = getelementptr i32, ptr addrspace(1) %{{.*}}, i32 0, !spirv.Decorations ![[#Cache1:]]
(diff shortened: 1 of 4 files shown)


MrSidims:
If somebody prefers another implementation (2nd mentioned for example) here is a draft to express opinion if unprivating a method in the scavenger is OK: https://github.com/MrSidims/SPIRV-LLVM-Translator/commit/156fe23c403ba005efd6b5efea2f2d51169af89e

aratajew (commented):
Overall, the change LGTM. Should we document it in the [SPIR-V Friendly IR doc](https://github.com/KhronosGroup/SPIRV-LLVM-Translator/blob/main/docs/SPIRVRepresentationInLLVM.rst)?

MrSidims:
> Should we document it in the [SPIR-V Friendly IR doc](https://github.com/KhronosGroup/SPIRV-LLVM-Translator/blob/main/docs/SPIRVRepresentationInLLVM.rst)?

In general - yes, but I would like to see if our experiment goes well and if any other changes are required

aratajew (approved):

VyacheslavLevytskyy:
@MrSidims Is there any chance to make this annotation more general than "spirv.DecorationCacheControlINTEL" in the nearest future? The PR itself looks good to me, and it's better than another option, in my opinion, but my concern is that while "spirv.Decorations" and "spirv.ParameterDecorations" looks natural and neatly, as a single entry point/bridge between LLVM IR/SPIR-V necessities, a newly introduced "spirv.DecorationCacheControlINTEL" is so evidently a particular case brought on a higher interface level between LLVM IR and SPIR-V that it doesn't seem elegant and feels not quite right. I don't have any valuable comments wrt. the code of the PR, it's just the approach with an exposure of a corner case as something normal makes me wonder whether it should be generalized before applying.

test/extensions/INTEL/SPV_INTEL_cache_controls/basic_load_store_arg_dec.ll
```diff
@@ -0,0 +1,66 @@
+; RUN: llvm-as %s -o %t.bc
+; RUN: llvm-spirv --spirv-ext=+SPV_INTEL_cache_controls -spirv-text %t.bc -o - | FileCheck %s --check-prefix=CHECK-SPIRV
+; RUN: llvm-spirv --spirv-ext=+SPV_INTEL_cache_controls %t.bc -o %t.spv
+; RUN: llvm-spirv -r %t.spv --spirv-target-env=SPV-IR -o - | llvm-dis -o - | FileCheck %s --check-prefix=CHECK-LLVM
+
+; CHECK-SPIRV-DAG: TypeInt [[#Int32:]] 32 0
+; CHECK-SPIRV-DAG: Constant [[#Int32]] [[#Zero:]] 0
+; CHECK-SPIRV-DAG: Decorate [[#Load1Ptr:]] CacheControlLoadINTEL 0 1
```
VyacheslavLevytskyy, line 8: This capture of Load1Ptr doesn't check anything. Should it be Load1GEPPtr?

MrSidims:
> Is there any chance to make this annotation more general than "spirv.DecorationCacheControlINTEL" in the nearest future

@VyacheslavLevytskyy yes, it's named like this for now for experiments. If we find it's useful in other cases it can be renamed. Right now it follows notation from spirv.Decorations metadata, so the (possible) switch would be painless. @svenvh would you find such new format useful for your extensions?

test/extensions/INTEL/SPV_INTEL_cache_controls/basic_load_store_arg_dec.ll
MrSidims: Good catch! It's a leftover from a previous test case a previous [attempt of](https://github.com/MrSidims/SPIRV-LLVM-Translator/commit/156fe23c403ba005efd6b5efea2f2d51169af89e) implementation

aratajew:
I think that it is reasonable that SPV_INTEL_cache_controls extension has its own metadata `spirv.DecorationCacheControlINTEL` in SPIR-V Friendly IR. This is because SPV_INTEL_cache_controls extension is kind of extraordinary as it allows to decorate a pointer, but it only has any effect if the decorated pointer is directly used by memory-reading/memory-writing instruction. So it actually should be treated as a property of an instruction not a pointer, but due to SPIRV limitations (OpStore cannot be decorated) the decoration is applied to a pointer. I've provided a more extensive explanation about this [here](https://github.com/KhronosGroup/SPIRV-LLVM-Translator/pull/2566#issuecomment-2116734834).
But if we find that this kind of representation can also be useful for any other extension, I'm not against making it more generic.

VyacheslavLevytskyy:
@aratajew Indeed, however, there is one more argument towards making this a general case eventually. After experiments are over and we'll get the final version of this feature, we'd need to add it upstream, to SPIR-V Backend in llvm.org.

VyacheslavLevytskyy (approved):

PR #2589, opened 2024-05-28
[CI] Upgrade out-of-tree job to Ubuntu 22.04

svenvh:


MrSidims (approved):

svenvh:
After a few attempts, it seems like we have a clean out-of-tree release build again; so the DenseMap warnings that we saw in #2588 appear to be false positives in GCC 9.4.0 that no longer happen with GCC 11.4.0.

wenju-he:
@svenvh does this change need to backport to 14~19 branches? Out-of-tree build of 14 branch is broken, see https://github.com/KhronosGroup/SPIRV-LLVM-Translator/actions/runs/10382683754/job/28746179650?pr=2671
please check if the failure is related to this PR.

svenvh:
@wenju-he I think you're right, I'll prepare the backports.

PR #2592, opened 2024-05-30
Update SPIRVRepresentationInLLVM.rst to include loop controls

aejjehint:
Add section about loop controls

### DIFF
--- a/docs/SPIRVRepresentationInLLVM.rst
+++ b/docs/SPIRVRepresentationInLLVM.rst
@@ -495,6 +495,7 @@ branch instruction.
 An example with the ``DependencyAccessesINTEL`` loop control:
 
 .. code-block:: llvm
+  
   br i1 %cond, label %loop, label %exit !spirv.loop.dependency_accesses !4
   ...
   !1 = distinct !{}      ; metadata corresponding to distinct access group
--- a/docs/SPIRVRepresentationInLLVM.rst
+++ b/docs/SPIRVRepresentationInLLVM.rst
@@ -488,8 +488,8 @@ Loop controls and loop metadata
 
 SPIR-V Loop controls that do not have a corresponding community LLVM metadata
 can be translated by creating a new loop metadata named
-``spirv.<LoopControlID>`` where `LoopControlID` is an identifier that represents
-the corresponding SPIR-V loop control. 
+``spirv.<LoopControlID>`` where ``LoopControlID`` is an identifier that
+represents the corresponding SPIR-V loop control. 
 
 Member decoration through pointer annotations
 ---------------------------------------------
--- a/docs/SPIRVRepresentationInLLVM.rst
+++ b/docs/SPIRVRepresentationInLLVM.rst
@@ -487,8 +487,10 @@ Loop controls and loop metadata
 -------------------------------
 
 SPIR-V Loop controls that do not have a corresponding community LLVM metadata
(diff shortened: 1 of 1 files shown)


docs/SPIRVRepresentationInLLVM.rst
```diff
@@ -483,6 +483,13 @@ extra operands ``i32 1`` and ``i32 2``.
 decorates the argument ``b`` of ``k`` with ``Restrict`` in SPIR-V while not
 adding any decoration to argument ``a``.
 
+Loop controls and loop metadata
+-------------------------------
+
+SPIR-V Loop controls that do not have a corresponding community LLVM metadata
```
bader, line 489: What does "community" mean in this context? I think just "LLVM metadata" should be good enough to describe the intention.

bader (approved):

docs/SPIRVRepresentationInLLVM.rst
aejjehint: @bader the intention is to distinguish between vendor-specific metadata and metadata that is present in community-LLVM

MrSidims, line 489: Should we specify where the metadata should be applied eg latch basic block?

aejjehint: Updated

MrSidims (approved):
LGTM
I do believe that now every such change in the translator's contract should also be reviewed by SPIR-V backend's devs, @VyacheslavLevytskyy please take a look

@aejjehint may be you can provide here some examples, how such metadata can be generated form a high level language. Also I have a question, while LLVM passes are aware of llvm.loop metadata and even try to preserve it if possible/makes sense, what is the plan for the new metadata?

docs/SPIRVRepresentationInLLVM.rst
```diff
@@ -483,6 +483,15 @@ extra operands ``i32 1`` and ``i32 2``.
 decorates the argument ``b`` of ``k`` with ``Restrict`` in SPIR-V while not
 adding any decoration to argument ``a``.
 
+Loop controls and loop metadata
```
VyacheslavLevytskyy, line 486: Please add at least one example of llvm code, as it's done in the previous section (that is "Function parameter, instruction and global variable decoration through metadata") to add the context and make the section more clear. Thank you.

VyacheslavLevytskyy (changes requested):
An example is needed

aejjehint:
> while LLVM passes are aware of llvm.loop metadata and even try to preserve it if possible/makes sense, what is the plan for the new metadata?

Like any other metadata, these metadata should be safe to be dropped if not understood.

VyacheslavLevytskyy (approved):

aejjehint:
@svenvh @bashbaug can you please provide your review at your earliest convenience? Thanks!

docs/SPIRVRepresentationInLLVM.rst
svenvh: I think "community" is ambiguous, it could refer to any community.  What about the following?
```suggestion
SPIR-V Loop controls that do not have corresponding `llvm.loop` metadata
```

aejjehint: @svenvh I have updated the text according to your suggestion.

svenvh (approved):

PR #2594, opened 2024-06-05
[NFC] Update complex-constexpr.ll test

MrSidims:
It started to fail again after deab451e7a7f . Only bitcast to inttoptr has left in constexpr context in this test.

TODO: do code cleanup in lib/SPIRV/SPIRVLowerConstExpr.cpp

### DIFF
--- a/test/complex-constexpr.ll
+++ b/test/complex-constexpr.ll
@@ -1,75 +1,32 @@
 ; RUN: llvm-as %s -o %t.bc
-; RUN: llvm-spirv %t.bc --spirv-max-version=1.3 -o %t.spv
-; RUN: spirv-val %t.spv
-; RUN: llvm-spirv %t.spv --spirv-max-version=1.3 -o %t.spt --to-text
-; RUN: llvm-spirv -r %t.spv -o %t.rev.bc
-; RUN: llvm-dis %t.rev.bc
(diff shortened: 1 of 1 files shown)


asudarsa:
As a first step, I think this particular test can be removed. There are other tests that do cover some of the supported const expressions. A cleanup to remove any unsupported const expressions may be warranted. Or I am also ok to react to fails if/when they crop up.

Thanks for addressing this @MrSidims

asudarsa (approved):
LGTM

MrSidims:
> lib/SPIRV/SPIRVLowerConstExpr.cpp

We still have tests like test/constexpr_phi.ll and test/constexpr_vector.ll (and may be elsewhere, which I can not find with ease using gitlog on SPIRVLowerConstExpr.cpp), both of them have bitcasts and ptrtoint used in constexpr context. I have an idea, will update PR in few minutes.

MrSidims:
@svenvh @asudarsa should be slightly better. We ran out of the instructions to be used in constexpr context, but like this should be fine (bitcast from inttoptr) as it keeps original purpose of the test "calling one instruction from another to determine argument of a function call).

svenvh (approved):

asudarsa (approved):
Ah..this looks good. Thanks

PR #2606, opened 2024-06-24
[LLVM->SPIRV] Fix target id in decorates when there is forward decl

wenju-he:
When there is forward declaration of a spirv entry, its decorates are not translated until its definition is seen. Forward id is re-used for its entry. Id in entry decorates should use forward id as well.


MrSidims (approved):

svenvh (approved):

wenju-he:
@MrSidims please help to merge, thank you

Issue #2302, opened 2024-01-15
Missing translation for OpIAddCarry, OpISubCarry, OpSMulExtended, OpUMulExtended


Snektron:
It seems that the overflow family of instructions mentioned in the issue title are currently not being translated, and result in an error when attempting to use them.

bwlodarcz:
OpIAddCarry and OpISubBorrow are currently implemented. OpSMulExtended and OpUMulExtended are not.

Snektron:
Apologies, I did't see that the carry instructions were already implemented.

LLVMs smul.with.overflow intrinsic also returns a struct {T, i1}, so I guess that will be similarly annoying.

vmaksimo:
@Snektron `MulExtended` instructions are supported now. Guess the issue can be closed

Snektron:
Very nice, thank you!!

PR #2608, opened 2024-06-26
Support SMulExtended and UMulExtended SPIR-V instructions

vmaksimo:
They can be used as SPIR-V friendly calls as there is no obvious mapping to any existing LLVM instruction.

The tests are basically copying the ones added in #2392

This change will address the second part of #2302 issue


lib/SPIRV/libSPIRV/SPIRVOpCode.h
```diff
@@ -71,7 +71,9 @@ inline bool isAtomicOpCode(Op OpCode) {
 }
 inline bool isBinaryOpCode(Op OpCode) {
   return ((unsigned)OpCode >= OpIAdd && (unsigned)OpCode <= OpFMod) ||
-         OpCode == OpDot || OpCode == OpIAddCarry || OpCode == OpISubBorrow;
+         OpCode == OpDot ||
+         ((unsigned)OpCode >= OpIAddCarry &&
```
LU-JOHN, line 75: I think it is clearer if every OpCode that is included is listed.  Using ranges (i.e. OpCode>=X && OpCode<=Y) is harder to understand.

vmaksimo: Applied

LU-JOHN (approved):
LGTM

svenvh (approved):

MrSidims (approved):

PR #2612, opened 2024-06-28
[Backport to 14][LLVM->SPIRV] Fix target id in decorates when there is forward decl (#2606)

wenju-he:
When there is forward declaration of a spirv entry, its decorates are not translated until its definition is seen. Forward id is re-used for its entry. Id in entry decorates should use forward id as well.

(cherry picked from commit 305f48884606abf18306db9cdac1487ab6c5d3f1)


wenju-he:
The original issue is in llvm_release_140 branch that is used by intel-graphics-compiler, thus do the backport.

MrSidims (approved):

PR #2617, opened 2024-06-28
[Backport to 15] Fix SPV_INTEL_runtime_aligned implementation part 1

vmaksimo:
It was implemented via new decoration, which is not correct. Instead it should be Function Parameter Attribute decoration.

In this commit starts fixing this in step-by-step manner.

Co-authored-by: Maksimova, Viktoria <viktoria.maksimova@intel.com>

(This is a cherry-pick of 0db0f98)


vmaksimo:
@MrSidims not sure if we should add `*.spt` test because we have full transcoding tests from 974d4db. WDYT?

MrSidims (approved):

PR #2631, opened 2024-07-04
Set spirv.Source according to the opencl.ocl.version and opencl.cxx.version.

haonanya:
Align with https://github.com/llvm/llvm-project/commit/5fd5b8ada70d9cbdaa8cc5bea50a6314e3140364.

### DIFF
--- a/lib/SPIRV/OCLUtil.cpp
+++ b/lib/SPIRV/OCLUtil.cpp
@@ -788,26 +788,39 @@ decodeOCLVer(unsigned Ver) {
 }
 
 unsigned getOCLVersion(Module *M, bool AllowMulti) {
-  NamedMDNode *NamedMD = M->getNamedMetadata(kSPIR2MD::OCLVer);
-  if (!NamedMD)
+  NamedMDNode *NamedMDOCLVer = M->getNamedMetadata(kSPIR2MD::OCLVer);
+  NamedMDNode *NamedMDOCLCXXVer = M->getNamedMetadata(kSPIR2MD::OCLCXXVer);
+  if (!NamedMDOCLVer && !NamedMDOCLCXXVer)
     return 0;
-  assert(NamedMD->getNumOperands() > 0 && "Invalid SPIR");
-  if (!AllowMulti && NamedMD->getNumOperands() != 1)
-    report_fatal_error(
-        llvm::Twine("Multiple OCL version metadata not allowed"));
 
   // If the module was linked with another module, there may be multiple
   // operands.
-  auto GetVer = [=](unsigned I) {
+  auto GetVerPair = [](unsigned I, NamedMDNode *NamedMD) {
     auto *MD = NamedMD->getOperand(I);
     return std::make_pair(getMDOperandAsInt(MD, 0), getMDOperandAsInt(MD, 1));
   };
-  auto Ver = GetVer(0);
-  for (unsigned I = 1, E = NamedMD->getNumOperands(); I != E; ++I)
-    if (Ver != GetVer(I))
-      report_fatal_error(llvm::Twine("OCL version mismatch"));
-
-  return encodeOCLVer(Ver.first, Ver.second, 0);
+  auto GetVer = [=](NamedMDNode *NamedMD) {
+    assert(NamedMD->getNumOperands() && "Invalid SPIR");
+    if (!AllowMulti && NamedMD->getNumOperands() != 1)
+      report_fatal_error(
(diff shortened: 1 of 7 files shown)


haonanya:
@svenvh , @MrSidims, can you please take a look? Thanks very much.

lib/SPIRV/PreprocessMetadata.cpp
```diff
@@ -294,12 +294,16 @@ void PreprocessMetadataBase::preprocessOCLMetadata(Module *M, SPIRVMDBuilder *B,
   // !{x} = !{i32 3, i32 102000}
   B->addNamedMD(kSPIRVMD::Source)
       .addOp()
-      .add(CLVer == kOCLVer::CL21 ? spv::SourceLanguageOpenCL_CPP
-                                  : spv::SourceLanguageOpenCL_C)
+      .add(M->getNamedMetadata(kSPIR2MD::OCLCXXVer) &&
+                   (CLVer == kOCLVer::CLCXX10 || CLVer == kOCLVer::CLCXX2021)
+               ? spv::SourceLanguageOpenCL_CPP
```
svenvh, line 299: Perhaps we should consider using `SourceLanguageCPP_for_OpenCL` instead, but I'm not entirely sure.  `SourceLanguageOpenCL_CPP` is for the "old" OpenCL C++ that was part of OpenCL 2.2.

lib/SPIRV/OCLUtil.cpp
```diff
@@ -788,26 +788,39 @@ decodeOCLVer(unsigned Ver) {
...
+  unsigned OCLCXXVer = NamedMDOCLCXXVer ? GetVer(NamedMDOCLCXXVer) : 0;
+  // Check if OCLCXXVer is compatible with OCLVer
+  if (OCLVer && OCLCXXVer) {
+    if ((OCLVer == kOCLVer::CL20 && OCLCXXVer == kOCLVer::CLCXX10) ||
+        (OCLVer == kOCLVer::CL30 && OCLCXXVer == kOCLVer::CLCXX2021))
+      return OCLCXXVer;
+    report_fatal_error(llvm::Twine(
+        "opencl cxx version is not compatible with opencl c version!"));
+  }
+  return OCLVer ? OCLVer : OCLCXXVer;
```
svenvh, line 823: Will we ever have an OCLCXXVer without an OCLVer?  If not, we can just do

```suggestion
  return OCLVer;
```

lib/SPIRV/PreprocessMetadata.cpp
haonanya: +@MrSidims, @asudarsa, there are several places using SourceLanguageOpenCL_CPP, should we change to SourceLanguageCPP_for_OpenCL? Thanks very much.

MrSidims: I'm also slightly leaning towards SourceLanguageCPP_for_OpenCL, technically it doesn't bring any semantic changes and from debugger perspective CPP_for_OpenCL and OpenCL_CPP are semantically equivalent.

MrSidims (approved):

lib/SPIRV/PreprocessMetadata.cpp
svenvh: We can do this as a separate PR, so not blocking this PR on that.

svenvh (approved):

haonanya:
@MrSidims , @svenvh, can we land the PR? Thanks.

lib/SPIRV/SPIRVReader.cpp
```diff
@@ -4830,7 +4831,7 @@ void SPIRVToLLVM::transSourceLanguage() {
   SPIRVWord Ver = 0;
   SourceLanguage Lang = BM->getSourceLanguage(&Ver);
   if (Lang != SourceLanguageUnknown && // Allow unknown for debug info test
-      Lang != SourceLanguageOpenCL_C && Lang != SourceLanguageOpenCL_CPP)
+      Lang != SourceLanguageOpenCL_C && Lang != SourceLanguageCPP_for_OpenCL)
```
haonanya, line 4834: Hi, @svenvh , @MrSidims , can we preserve old SourceLanguageOpenCL_CPP metadata? Currently spirv reader drop SourceLanguageOpenCL_CPP for sycl source https://github.com/intel/llvm/blob/sycl/clang/lib/CodeGen/CodeGenModule.cpp#L1360,L1362 on spirv->LLVM IR.

PR #2633, opened 2024-07-05
Allow llvm-spirv to be built natively in LLVM cross builds.

hvdijk:
When llvm-spirv is built in-tree as part of LLVM, and the LLVM build also includes libclc, libclc will require a native version of llvm-spirv to be built. To accommodate this, call setup_host_tool. This change by itself is intended to have no effect at the moment, but enables a future LLVM change.

### DIFF
--- a/tools/llvm-spirv/CMakeLists.txt
+++ b/tools/llvm-spirv/CMakeLists.txt
@@ -16,6 +16,8 @@ add_llvm_tool(llvm-spirv
   NO_INSTALL_RPATH
 )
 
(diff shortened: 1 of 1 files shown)


MrSidims (commented):
Fine with me,

>> future LLVM change

If there is a link to share, may I ask you to add it?

svenvh (approved):

MrSidims (approved):
Thanks! @hvdijk Would you prefer the PR be merged now or after LLVM commit?

hvdijk:
Preferably before, that way I can update the LLVM PR to not say that an external change to SPIRV-LLVM-Translator is needed :)

PR #2634, opened 2024-07-07
Do not pass -DLLVM_EXTERNAL_PROJECTS="SPIRV-Headers" for an out of tree build

Zopolis4:
As per the CI logs:
```
CMake Warning:
  Manually-specified variables were not used by the project:

    LLVM_EXTERNAL_PROJECTS


-- Build files have been written to: /home/runner/work/SPIRV-LLVM-Translator/SPIRV-LLVM-Translator/build
```

According to https://reviews.llvm.org/D20838 it looks like this is only relevant for in-tree builds, so no need to pass it here.


svenvh (approved):
LGTM, thanks!  It seems we should also update README.md which mentions this argument too.

MrSidims (approved):

MrSidims:
> It seems we should also update README.md which mentions this argument too.

Guess we can do it on our own

PR #2637, opened 2024-07-08
Don't add initializer to local variables with external linkage

maarquitos14:
If a variable is external, it will be defined during linking, the translator should not do it.


MrSidims:
I agree with the change and will approve after adding a test.

But would like to hear from Sven.

maarquitos14:
> I agree with the change and will approve after adding a test.
> 
> But would like to hear from Sven.

Test added.

svenvh (approved):

PR #2648, opened 2024-07-12
[Backport to 15] [DebugInfo] Round up #elts of TypeVector when calculating memory size

maarquitos14:


maarquitos14:
`llvm::bit_ceil` was introduced in LLVM 16, and `std::bit_ceil` requires C++20. Should I implement our own `bit_ceil` function for backports to 14 and 15?

lib/SPIRV/SPIRVInternal.h
```diff
@@ -1098,6 +1098,97 @@ bool postProcessBuiltinsWithArrayArguments(Module *M, bool IsCpp = false);
...
+    unsigned ZeroBits = 0;
+    for (T Shift = std::numeric_limits<T>::digits >> 1; Shift; Shift >>= 1) {
+      T Tmp = Val >> Shift;
+      if (Tmp)
+        Val = Tmp;
+      else
+        ZeroBits |= Shift;
+    }
+    return ZeroBits;
+  }
```
MrSidims, line 1118: shouldn't it be just the entire algorithm (with few additions, like if <2 return 1 and if already power of 2 return Val?):
```suggestion
    Val--;
    for (T Shift = std::numeric_limits<T>::digits >> 1; Shift; Shift >>= 1) {
      Val | = Val >> Shift;
    }
    return Val++;
  }
```

MrSidims (approved):

PR #2650, opened 2024-07-17
Adjust OpVariable's users and Lifetime ptr operand storage classes

MrSidims:
While clang is generating alloca in private address space for OpenCL following target's datalayout, some of the passes might generate it in generic AS and for this cases we have to create OpVariable in function storage class and create a cast to generic. Additionally if such alloca was an operand of lifetime instruction - then SPIR-V without storage class adjustment validation for lifetime instruction will be violated (pointer operand must be with Function storage class).

This became exposed after https://github.com/llvm/llvm-project/pull/97306 which was fixing handling of byval pointer parameter of a function during inlining, byval pointer for OpenCL will be generated in default (under an option - generic) address space making the alloca to store value of the function's operand also be generic.

### DIFF
--- a/test/llvm-intrinsics/lifetime.ll
+++ b/test/llvm-intrinsics/lifetime.ll
@@ -6,14 +6,45 @@
 ; RUN: llvm-spirv -r %t.spv -o %t.spv.bc
 ; RUN: llvm-dis < %t.spv.bc | FileCheck %s --check-prefix=CHECK-LLVM
 
-; CHECK-SPIRV: 3 LifetimeStart [[tmp:[0-9]+]] 0
-; CHECK-SPIRV: 3 LifetimeStop [[tmp]] 0
+; CHECK-SPIRV-DAG: Name [[#SimpleF:]] "lifetime_simple"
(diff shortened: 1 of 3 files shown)


test/llvm-intrinsics/lifetime.ll
```diff
@@ -6,14 +6,45 @@
...
+; CHECK-SPIRV: LifetimeStop [[#Tmp]] 0
+
+; CHECK-SPIRV: Function [[#]] [[#SizedF:]]
+; CHECK-SPIRV: LifetimeStart [[#Tmp:]] 1
+; CHECK-SPIRV: Bitcast [[#]] [[#Cast:]]
+; CHECK-SPIRV: LifetimeStop [[#Cast]] 1
+
+; CHECK-SPIRV: Function [[#]] [[#GenericF:]]
+; CHECK-SPIRV: Variable [[#]] [[#Var:]] 7
+; CHECK-SPIRV: PtrCastToGeneric [[#]] [[#Cast1:]] [[#Var]]
```
MrSidims, line 24: potentially extra analysis might be done removing useless (in this case) casts, do we want it? probably such analysis wouldn't be generic

svenvh (approved):

PR #2657, opened 2024-08-01
Add translation for Intrinsic::{atan,acos,asin,cosh,sinh,tanh}

linehill:
Add translation for atan, acos, asin, cosh, sinh and tanh LLVM intrinsics which are mapped to corresponding OpenCL extended instructions.


MrSidims (commented):
Thanks for the contribution! Lets also add these intrinsics to the list within getNativeBuiltinIdForIntrinsic function in SPIRVWriter, so when an intrinsic has afn fast-math flag it would be mapped on a native OpenCL builtin.

linehill:
Are there native versions for these OpenCL builtins? It seems to me there aren't.

MrSidims (approved):

MrSidims:
Restarting CI with the new HEAD

MrSidims:
Merging, will get approve from Sven offline if necessary

linehill:
Thanks for the review and merging!

svenvh (commented):
LGTM after the fact. :)

PR #2658, opened 2024-08-02
[CI] Upgrade in-tree job to Ubuntu 22.04

svenvh:
The spirv-tools package used by the job seems no longer available for Ubuntu 20.04.


MrSidims (approved):

MrSidims:
I don't mind the update, though I don't see significant changes in https://github.com/actions/runner-images/pull/10323 so github actions have removed some packages from ubuntu 20.04 CI

PR #2667, opened 2024-08-08
[SPIR-V 1.4] Deprecate OpDecorationGroup, OpGroupDecorate, and OpGroupMemberDecorate

vmaksimo:


lib/SPIRV/libSPIRV/SPIRVDecorate.h
```diff
@@ -433,6 +437,13 @@ class SPIRVGroupDecorateGeneric : public SPIRVEntryNoIdGeneric {
   }
   virtual void decorateTargets() = 0;
   _SPIRV_DCL_ENCDEC
+  void validate() const override {
+    getModule()->getErrorLog().checkError(
+        this->getModule()->getSPIRVVersion() < VersionNumber::SPIRV_1_4,
+        SPIRVEC_InvalidModule,
+        "OpGroup(Member)Decorate is deprecated starting from SPIR-V 1.4");
```
LU-JOHN, line 444: Maybe re-write as:

"OpGroupDecorate and OpGroupMemberDecoreate are deprecated starting from SPIR-V 1.4"

vmaksimo:
> I think it would be fine to identify use of these deprecated instructions with a warning, but failing translation seems a bit too strong.

@bashbaug  I agree, but unfortunately for now we don't have a mechanism to report a warning in llvm-spirv (am I wrong?).
If so, I think that maybe it'd be a solution to just ignore and close this PR, as it does not change much.

PR #2683, opened 2024-08-22
Fix element type when reading OpPtrDiff

svenvh:
The first argument to IRBuilder's `CreatePtrDiff` must be the element type, not the pointer type; otherwise the resulting difference will be scaled incorrectly.

### DIFF
--- a/lib/SPIRV/SPIRVReader.cpp
+++ b/lib/SPIRV/SPIRVReader.cpp
@@ -2185,7 +2185,9 @@ Value *SPIRVToLLVM::transValueWithoutDecoration(SPIRVValue *BV, Function *F,
     auto *BC = static_cast<SPIRVBinary *>(BV);
     auto Ops = transValue(BC->getOperands(), F, BB);
     IRBuilder<> Builder(BB);
-    Value *V = Builder.CreatePtrDiff(transType(BC->getType()), Ops[0], Ops[1]);
+    Type *ElemTy =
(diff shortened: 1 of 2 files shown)


vmaksimo (approved):
Nice catch, thanks!

lib/SPIRV/SPIRVReader.cpp
```diff
@@ -2185,7 +2185,9 @@ Value *SPIRVToLLVM::transValueWithoutDecoration(SPIRVValue *BV, Function *F,
     auto *BC = static_cast<SPIRVBinary *>(BV);
     auto Ops = transValue(BC->getOperands(), F, BB);
     IRBuilder<> Builder(BB);
-    Value *V = Builder.CreatePtrDiff(transType(BC->getType()), Ops[0], Ops[1]);
+    Type *ElemTy =
+        transType(BC->getOperands()[0]->getType()->getPointerElementType());
```
vmaksimo, line 2189: can we simplify a bit, as BC operands are already translated?
```suggestion
        Ops[0]->getType()->getPointerElementType());
```

svenvh: Not really, as the translated type of `Ops[0]` will just be an opaque pointer, while we need the element type.  We only have the element type information in the SPIR-V representation, it is not propagated into LLVM by the earlier `transValue` call.

PR #2693, opened 2024-08-28
[Backport to 16] add API call to display general information about the module (#2298)

ustachow:
Partially load SPIR-V from the stream and decode only selected for the report instructions, needed to retrieve general information about the module: capabilities, extensions, version, memory model and addressing model.

In addition to immediately helpful for back-ends lists of capabilities and extensions declared in SPIR-V module, a general intent also is to extend report details in future by feedbacks about further potentially useful analysis, statistics, etc.


MrSidims (approved):
Build failure is unrelated

PR #2703, opened 2024-09-02
Report error when finding variadic function other than printf

maarquitos14:
Variadic functions are not supported in SPIR-V, the only known exception is printf.


lib/SPIRV/SPIRVWriter.cpp
```diff
@@ -814,6 +814,11 @@ SPIRVType *LLVMToSPIRVBase::transSPIRVOpaqueType(StringRef STName,
 SPIRVType *LLVMToSPIRVBase::transScavengedType(Value *V) {
   if (auto *F = dyn_cast<Function>(V)) {
     FunctionType *FnTy = Scavenger->getFunctionType(F);
+    // VarArg functions other than printf are not supported in SPIR-V.
+    BM->getErrorLog().checkError(
+        !(FnTy->isVarArg() && F->getName() != "printf"),
```
LU-JOHN, line 819: Is this test too strict?  In the test array_of_matrices.ll, I see:

declare dso_local spir_func i32 @_Z18__spirv_ocl_printfPU3AS2Kcz(ptr addrspace(2), ...)

VyacheslavLevytskyy: I agree, there is mangling, and at least 3 names for the printf, so `F->getName() != "printf"` would not work properly.

maarquitos14: I considered `F->getName().contains("printf")`, but thought that was too loose. What do you think, would that be better?

VyacheslavLevytskyy: In SPIR-V Backend we have something along the lines of
```
  if ((DemangledName.starts_with("__spirv_ocl_printf(") ||
       DemangledName.starts_with("printf(")) ...
```
so I guess as a first approach this could work after you've got `DemangledName` from `F->getName()` from `llvm::itaniumDemangle`.

test/negative/unsupported_vararg_func.ll
```diff
@@ -0,0 +1,18 @@
...
+target datalayout = "e-i64:64-v16:16-v24:32-v32:32-v48:64-v96:128-v192:256-v256:256-v512:512-v1024:1024-n8:16:32:64-G1"
+target triple = "spir64_unknown_unknown"
+
+; CHECK: UnsupportedVarArgFunction: Variadic functions other than 'printf' are not supported in SPIR-V.
+; Function Attrs: convergent
+declare spir_func void @for__issue_diagnostic(i32 noundef, i32 noundef, ...) local_unnamed_addr
+
+define i32 @foo() nounwind {
+  call spir_func void (i32, i32, ...) @for__issue_diagnostic(i32 noundef 41, i32 noundef 0)
+    ret i32 0
```
VyacheslavLevytskyy, line 16: two spaces before ret :)

lib/SPIRV/SPIRVWriter.cpp
```diff
@@ -814,6 +814,16 @@ SPIRVType *LLVMToSPIRVBase::transSPIRVOpaqueType(StringRef STName,
...
+    // VarArg functions other than printf are not supported in SPIR-V.
+    if (FnTy->isVarArg()) {
+      StringRef DemangledName;
+      oclIsBuiltin(F->getName(), DemangledName);
+      BM->getErrorLog().checkError(
+          (DemangledName.starts_with("printf(") ||
+           DemangledName.starts_with("__spirv_ocl_printf(")),
+          SPIRVEC_UnsupportedVarArgFunction);
+    }
+
```
VyacheslavLevytskyy, line 826: LGTM!

VyacheslavLevytskyy (approved):

svenvh (approved):

MrSidims (approved):

lib/SPIRV/SPIRVWriter.cpp
```diff
@@ -814,6 +814,16 @@ SPIRVType *LLVMToSPIRVBase::transSPIRVOpaqueType(StringRef STName,
 SPIRVType *LLVMToSPIRVBase::transScavengedType(Value *V) {
   if (auto *F = dyn_cast<Function>(V)) {
     FunctionType *FnTy = Scavenger->getFunctionType(F);
+    // VarArg functions other than printf are not supported in SPIR-V.
+    if (FnTy->isVarArg()) {
+      StringRef DemangledName;
+      oclIsBuiltin(F->getName(), DemangledName);
+      BM->getErrorLog().checkError(
+          (DemangledName.starts_with("printf(") ||
+           DemangledName.starts_with("__spirv_ocl_printf(")),
```
LU-JOHN, line 823: oclIsBuiltin demangles "printf" to "__spirv_ocl_printf".  I think only checking for:  DemangledName.starts_with("__spirv_ocl_printf") should be sufficient.  I think we should possibly be able to check for equality instead of starts_with.

Also does DemangledName include an open parenthesis?

maarquitos14: I have actually realized none of `printf` variants can reach this piece of code because they are skipped in https://github.com/KhronosGroup/SPIRV-LLVM-Translator/blob/7dacb7cdedc9c01e0ce88f1f3766c42ab8f59b7c/lib/SPIRV/SPIRVWriter.cpp#L5789. More specifically, `isBuiltinTransToExtInst(&F)` returns true for `printf`, so it is just skipped and never added to neither `Decls` nor `Defs`. Considering this, I think we should be safe by just checking if the function is variadic to report the error at this point. What do you think? Also tagging the rest of approvers to gather their opinions: @svenvh @MrSidims @VyacheslavLevytskyy

LU-JOHN (approved):
LGTM

lib/SPIRV/SPIRVWriter.cpp
```diff
@@ -814,6 +814,18 @@ SPIRVType *LLVMToSPIRVBase::transSPIRVOpaqueType(StringRef STName,
...
     FunctionType *FnTy = Scavenger->getFunctionType(F);
+    // VarArg functions other than printf are not supported in SPIR-V. None of
+    // printf variants can actually reach this point because they're skipped in
+    // 'LLVMToSPIRVBase::translate()'. Specifically,
+    // 'isBuiltinTransToExtInst(&F)' returns true for printf and its variants,
+    // and so they are never added to 'Decls' and 'Defs'. Therefore, it's safe
+    // to only check here if we're dealing with a variadic function to report an
+    // error. To be on the safe side, an assertion is added to check printf
+    // never reaches this point.
+    assert(F->getName().str() != "printf");
```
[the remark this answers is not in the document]
maarquitos14: Done, thanks :)

svenvh (approved):

PR #2687, opened 2024-08-26
SPV_KHR_untyped_pointers - implement OpTypeUntypedPointerKHR

vmaksimo:
This is the first part of the extension implementation. 
Introducing untyped pointer type.

Spec: https://htmlpreview.github.io/?https://github.com/KhronosGroup/SPIRV-Registry/blob/main/extensions/KHR/SPV_KHR_untyped_pointers.html

PR #2709, opened 2024-09-06
SPV_KHR_untyped_pointers - implement OpUntypedVariableKHR

vmaksimo:
This continues #2687 by introducing untyped variables.

Spec:
https://htmlpreview.github.io/?https://github.com/KhronosGroup/SPIRV-Registry/blob/main/extensions/KHR/SPV_KHR_untyped_pointers.html


lib/SPIRV/SPIRVReader.cpp
```diff
@@ -1527,9 +1527,15 @@ Value *SPIRVToLLVM::transValueWithoutDecoration(SPIRVValue *BV, Function *F,
...
-    auto *BVar = static_cast<SPIRVVariable *>(BV);
-    auto *PreTransTy = BVar->getType()->getPointerElementType();
+  case OpVariable:
+  case OpUntypedVariableKHR: {
+    auto *BVar = static_cast<SPIRVVariableBase *>(BV);
+    SPIRVType *PreTransTy = BVar->getType()->getPointerElementType();
+    if (BVar->getType()->isTypeUntypedPointerKHR()) {
+      auto *UntypedVar = static_cast<SPIRVUntypedVariableKHR *>(BVar);
+      if (SPIRVType *DT = UntypedVar->getDataType())
+        PreTransTy = DT;
```
LU-JOHN, line 1537: Is it possible for 1536 to fail?  
If it fails then PreTransTy will get the old def `PreTransTy = BVar->getType()->getPointerElementType()` 
but `BVar->getType()->isTypeUntypedPointerKHR()` which seems odd.

MrSidims:
Restarting CI

lib/SPIRV/SPIRVReader.cpp
vmaksimo: By specification, we can have untyped variable without data type. Guess I need to add test for global variable without type to see if it's valid

test/extensions/KHR/SPV_KHR_untyped_pointers/infinite-phi.ll
```diff
@@ -0,0 +1,55 @@
...
+target datalayout = "e-p:32:32-i64:64-v16:16-v24:32-v32:32-v48:64-v96:128-v192:256-v256:256-v512:512-v1024:1024"
+target triple = "spir-unknown-unknown"
+
+; CHECK-SPIRV: TypeInt [[#INT:]] 32 0
+; CHECK-SPIRV: TypeUntypedPointerKHR [[#PTRTY:]] 7
+; CHECK-SPIRV: TypeFloat [[#FLOAT:]] 32
+
+; CHECK-SPIRV: UntypedVariableKHR [[#PTRTY]] [[#IPTR:]] 7 [[#INT]]
+; CHECK-SPIRV: UntypedVariableKHR [[#PTRTY]] [[#FPTR:]] 7 [[#FLOAT]]
+; CHECK-SPIRV: Bitcast [[#PTRTY]] [[#IPTRI:]] [[#IPTR]]
```
MrSidims, line 22: This bitcast feels redundant. Guess it's being generated by type scavenger, right? It might be done in a separate PR, but probably we can get rid of it by either modifying the scavenger or by checking the types during bitcast generation.

lib/SPIRV/libSPIRV/SPIRVInstruction.h
```diff
@@ -530,6 +531,78 @@ class SPIRVVariable : public SPIRVInstruction {
...
+    validate();
+  }
+  SPIRVUntypedVariableKHR() : SPIRVVariableBase(OpUntypedVariableKHR) {}
+  SPIRVType *getDataType() const {
+    if (DataType.empty())
+      return nullptr;
+    assert(DataType.size() == 1);
+    return get<SPIRVType>(DataType[0]);
+  }
+  // bool hasDataType() const { return DataType; }
```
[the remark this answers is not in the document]
vmaksimo: Yep, thanks, will remove

test/extensions/KHR/SPV_KHR_untyped_pointers/infinite-phi.ll
vmaksimo: You're right, the bitcast appeared because of the type scavenger. This happens before any translation, so for now I'm not sure how to avoid this issue. I'll add TODO to the test

lib/SPIRV/SPIRVReader.cpp
vmaksimo: I've added test that shows it works as expected. 
That's because `PreTransTy = BVar->getType()->getPointerElementType()` returns the same `OpTypeUntypedPointerKHR` for `OpTypeUntypedPointerKHR` type.

svenvh:
LGTM; closing/reopening once again for CI to retrigger.

MrSidims (approved):

lib/SPIRV/SPIRVReader.cpp
LU-JOHN: From my understanding, when we are handling OpUntypedVariableKHR,
`SPIRVType *PreTransTy = BVar->getType()->getPointerElementType();`
handles the case when the untyped variable has no data type.   I guess this is because getPointerElementType() returns the correct type when there is no data type.
What will PreTransTy be set to if the untyped variable does have a data type?  I am hoping that getPointerElementType() will also return the correct data type, and thus lines 1534 to 1538 can be removed.

vmaksimo: The idea seems good at first glance, but unfortunately it's not doable and we still do need to have the special condition for untyped variable.

Return type of `OpVariable` is a pointer to the variable type, so it's enough to call `BVar->getType()->getPointerElementType()` to get it.
However, the return type of `OpUntypedVariableKHR` is untyped pointer, which does not hold information about data type by its definition. The only place where we can get info about data type (if it's not empty) is from `UntypedVar->getDataType()`, so the condition can't be removed.

LU-JOHN (approved):

PR #2711, opened 2024-09-09
Relax operand constraints on max_work_group_size metadata

frasercrmck:
This allows `max_work_group_size` metadata to behave like `reqd_work_group_size` and `work_group_size_hint` in that it may have between 1 and 3 operands. Missing dimensions are filled in with 1s.

### DIFF
--- a/lib/SPIRV/PreprocessMetadata.cpp
+++ b/lib/SPIRV/PreprocessMetadata.cpp
@@ -129,32 +129,28 @@ void PreprocessMetadataBase::visit(Module *M) {
     // of ExecutionMode instructions.
 
     // !{void (i32 addrspace(1)*)* @kernel, i32 17, i32 X, i32 Y, i32 Z}
-    if (MDNode *WGSize = Kernel.getMetadata(kSPIR2MD::WGSize)) {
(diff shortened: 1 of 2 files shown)


frasercrmck:
Thanks for the reviews! It seems like the CI isn't behaving well, but it doesn't look like it's due to this PR?

AlexeySachkov:
> Thanks for the reviews! It seems like the CI isn't behaving well, but it doesn't look like it's due to this PR?

LLVM binary packages used in out-of-tree builds are built from slightly outdated (for a few days) sources. Therefore there could be a situation where they are incompatible with APIs used in the translator.

However, I'm not following the translator repo that closely to know the actual status, so I will leave it up to @MrSidims and other maintainers to merge this PR

svenvh:
Closing/reopening, as @MrSidims recently merged a patch that should fix the CI.

MrSidims:
Apparently clang-format also started to require libPollyISL, will create PR to fix that.

I believe the PR is good to go, I just usually wait for a working day after approval before merging in case if anybody wants to express an opinion.

MrSidims:
Restarting CI one more time

PR #2713, opened 2024-09-09
Handle constant sized array allocations using no extension

victor-eds:
In LLVM, array allocations might have constant size:

```llvm
%array = alloca i32, i64 4, align 4
```

Represent this kind of allocations using `OpVariable + OpBitcast`.

Before this patch, the `SPV_INTEL_variable_length_array` extension was used.

### DIFF
--- a/test/simple.ll
+++ b/test/simple.ll
@@ -126,13 +126,6 @@ entry:
   ret void
 }
 
-; CHECK: "test_array_alloca"
-; Function Attrs: nounwind
-define dso_local void @test_array_alloca() #0 {
-  %arr = alloca i32, i64 4, align 4
-  ret void
-}
-
 attributes #0 = { nounwind "less-precise-fpmad"="false" "no-frame-pointer-elim"="false" "no-infs-fp-math"="false" "no-nans-fp-math"="false" "no-realign-stack" "stack-protector-buffer-size"="8" "unsafe-fp-math"="false" "use-soft-float"="false" }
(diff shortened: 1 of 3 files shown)


test/simple.ll
```diff
@@ -126,6 +126,13 @@ entry:
   ret void
 }
 
+; CHECK: "test_array_alloca"
+; Function Attrs: nounwind
+define dso_local void @test_array_alloca() #0 {
+  %arr = alloca i32, i64 4, align 4
+  ret void
+}
```
victor-eds, line 134: As no extension is used, this test would fail if an extension was needed.

victor-eds:
@MrSidims can I get a review here?

test/simple.ll
```diff
@@ -126,6 +126,13 @@ entry:
   ret void
 }
 
+; CHECK: "test_array_alloca"
+; Function Attrs: nounwind
+define dso_local void @test_array_alloca() #0 {
```
MrSidims, line 131: Probably it's worth to outline the test in a separate file with checks for generated SPIR-V and LLVM IR after the translation. Reason: LLVM IR changes after roundtrip (which is fine) and this should be exposed in the test.

MrSidims (approved):
Thanks for fixing this! Please consider a test change

test/simple.ll
victor-eds: Created standalone test

victor-eds:
Failing tests not because of lines I've even touched.

MrSidims:
Restarting CI

PR #2716, opened 2024-09-11
Add an option to translate LLVM IR to SPIR-V using the LLVM SPIRV Backend target

VyacheslavLevytskyy:
This PR adds an option to translate LLVM IR to SPIR-V using the LLVM SPIRV Backend target:
* cmake is modified to search for the LLVM SPIRV Backend target while configuring the project,
* if LLVM is built with SPIRV Backend support, we may use the interface exposed by SPIRV BE to translate Module to SPIR-V code if a user explicitly asks for this way of LLVM IR transaformation.

### DIFF
--- a/include/LLVMSPIRVOpts.h
+++ b/include/LLVMSPIRVOpts.h
@@ -149,6 +149,9 @@ public:
       ExtStatusMap[Extension] = Allow;
   }
 
+  std::vector<std::string>
+  getAllowedSPIRVExtensionNames(std::function<bool(ExtensionID)> &Filter) const;
+
   VersionNumber getMaxVersion() const { return MaxVersion; }
 
   bool isGenArgNameMDEnabled() const { return GenKernelArgNameMD; }
--- a/include/LLVMSPIRVOpts.h
+++ b/include/LLVMSPIRVOpts.h
@@ -241,11 +241,11 @@ public:
   }
   BuiltinFormat getBuiltinFormat() const noexcept { return SPIRVBuiltinFormat; }
 
-  void useLLVMSPIRVBackendTarget(bool Flag) noexcept {
-    UseLLVMSPIRVBackendTarget = Flag;
+  void useLLVMTarget(bool Flag) noexcept {
+    UseLLVMTarget = Flag;
   }
   bool getUseLLVMSPIRVBackendTarget() const noexcept {
-    return UseLLVMSPIRVBackendTarget;
+    return UseLLVMTarget;
   }
 
 private:
@@ -296,7 +296,7 @@ private:
   BuiltinFormat SPIRVBuiltinFormat = BuiltinFormat::Function;
 
   // Convert LLVM to SPIR-V using the LLVM SPIR-V Backend target
-  bool UseLLVMSPIRVBackendTarget = false;
+  bool UseLLVMTarget = false;
 };
 
 } // namespace SPIRV
--- a/include/LLVMSPIRVOpts.h
+++ b/include/LLVMSPIRVOpts.h
@@ -241,12 +241,8 @@ public:
   }
   BuiltinFormat getBuiltinFormat() const noexcept { return SPIRVBuiltinFormat; }
 
-  void useLLVMTarget(bool Flag) noexcept {
-    UseLLVMTarget = Flag;
-  }
-  bool getUseLLVMSPIRVBackendTarget() const noexcept {
-    return UseLLVMTarget;
-  }
+  void useLLVMTarget(bool Flag) noexcept { UseLLVMTarget = Flag; }
+  bool getUseLLVMSPIRVBackendTarget() const noexcept { return UseLLVMTarget; }
 
 private:
   // Common translation options
--- a/include/LLVMSPIRVOpts.h
+++ b/include/LLVMSPIRVOpts.h
@@ -242,7 +242,7 @@ public:
   BuiltinFormat getBuiltinFormat() const noexcept { return SPIRVBuiltinFormat; }
 
   void setUseLLVMTarget(bool Flag) noexcept { UseLLVMTarget = Flag; }
-  bool getUseLLVMSPIRVBackendTarget() const noexcept { return UseLLVMTarget; }
+  bool getUseLLVMTarget() const noexcept { return UseLLVMTarget; }
 
 private:
   // Common translation options
--- a/include/LLVMSPIRVOpts.h
+++ b/include/LLVMSPIRVOpts.h
@@ -241,6 +241,13 @@ public:
   }
   BuiltinFormat getBuiltinFormat() const noexcept { return SPIRVBuiltinFormat; }
 
+  void useLLVMSPIRVBackendTarget(bool Flag) noexcept {
+    UseLLVMSPIRVBackendTarget = Flag;
+  }
+  bool getUseLLVMSPIRVBackendTarget() const noexcept {
+    return UseLLVMSPIRVBackendTarget;
+  }
+
 private:
   // Common translation options
   VersionNumber MaxVersion = VersionNumber::MaximumVersion;
@@ -287,6 +294,9 @@ private:
   bool PreserveAuxData = false;
 
   BuiltinFormat SPIRVBuiltinFormat = BuiltinFormat::Function;
+
+  // Convert LLVM to SPIR-V using the LLVM SPIR-V Backend target
+  bool UseLLVMSPIRVBackendTarget = false;
 };
 
 } // namespace SPIRV
--- a/include/LLVMSPIRVOpts.h
+++ b/include/LLVMSPIRVOpts.h
@@ -241,7 +241,7 @@ public:
   }
   BuiltinFormat getBuiltinFormat() const noexcept { return SPIRVBuiltinFormat; }
 
-  void useLLVMTarget(bool Flag) noexcept { UseLLVMTarget = Flag; }
+  void setUseLLVMTarget(bool Flag) noexcept { UseLLVMTarget = Flag; }
   bool getUseLLVMSPIRVBackendTarget() const noexcept { return UseLLVMTarget; }
 
 private:
--- a/lib/SPIRV/SPIRVWriter.cpp
+++ b/lib/SPIRV/SPIRVWriter.cpp
@@ -6753,7 +6753,93 @@ bool isValidLLVMModule(Module *M, SPIRVErrorLog &ErrorLog) {
   return true;
 }
 
+#if defined(LLVM_SPIRV_BACKEND_TARGET_PRESENT)
+namespace llvm {
+extern "C" bool
+SPIRVTranslateModule(Module *M, std::string &SpirvObj, std::string &ErrMsg,
+                     const std::vector<std::string> &AllowExtNames,
+                     const std::vector<std::string> &Opts);
+} // namespace llvm
+#endif
+
 namespace {
+
+#if defined(LLVM_SPIRV_BACKEND_TARGET_PRESENT)
+static inline Triple::SubArchType spirvVersionToSubArch(VersionNumber VN) {
+  switch (VN) {
+  case VersionNumber::SPIRV_1_0:
+    return Triple::SPIRVSubArch_v10;
+  case VersionNumber::SPIRV_1_1:
+    return Triple::SPIRVSubArch_v11;
+  case VersionNumber::SPIRV_1_2:
+    return Triple::SPIRVSubArch_v12;
+  case VersionNumber::SPIRV_1_3:
+    return Triple::SPIRVSubArch_v13;
+  case VersionNumber::SPIRV_1_4:
+    return Triple::SPIRVSubArch_v14;
+  case VersionNumber::SPIRV_1_5:
+    return Triple::SPIRVSubArch_v15;
+  case VersionNumber::SPIRV_1_6:
+    return Triple::SPIRVSubArch_v16;
+  }
+  return Triple::NoSubArch;
+}
+
+bool runSpirvBackend(Module *M, std::ostream *OS, std::string &ErrMsg,
+                     const SPIRV::TranslatorOpts &TranslatorOpts) {
+  static const std::set<ExtensionID> ImplementedBySpirvBE{
+      SPIRV::ExtensionID::SPV_EXT_shader_atomic_float_add,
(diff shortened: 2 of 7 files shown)


VyacheslavLevytskyy:
@MrSidims @asudarsa The PR is in draft because I should do docs/tests in addition, but what's available right now is pretty much all code that I planned for this PR. So if you have any suggestions while I'm working on docs/tests please let me know. Consider this a kind of early bird access)

lib/SPIRV/SPIRVWriter.cpp
```diff
@@ -6753,8 +6753,94 @@ bool isValidLLVMModule(Module *M, SPIRVErrorLog &ErrorLog) {
...
+      SPIRV::ExtensionID::SPV_KHR_shader_clock,
+      SPIRV::ExtensionID::SPV_KHR_cooperative_matrix,
+      SPIRV::ExtensionID::SPV_KHR_non_semantic_info};
+  static const std::string DefaultTriple = "spirv64-unknown-unknown";
+  static const std::vector<std::string> Opts{"--avoid-spirv-capabilities",
+                                             "--translator-compatibility-mode"};
+
+  std::function<bool(SPIRV::ExtensionID)> Filter =
+      [](SPIRV::ExtensionID Ext) -> bool {
+    return ImplementedBySpirvBE.find(Ext) != ImplementedBySpirvBE.end();
```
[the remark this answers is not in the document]
VyacheslavLevytskyy: > this table has to be updated
Yes, this is by intent. The rationale is that Translator is a client of SPIRV BE service in this case, and a set of extensions supported by SPIRV BE is a part of the contract. When the contract is updated on the service side (new extension is added), the client must update its part of the contract as well (allow to pass this extension to SPIRV BE).

From the technical point of view, this list must be filtered anyway, either by SPIRV BE, or by Translator, otherwise, SPIRV BE fails with the complaint about unknown extension. So this PR proposes an explicit way of resolving this, by clearly state the contract (limitations, in a form of supported extensions that Translator aware of). If decision were rather to implement this on SPIRV BE side by ignoring unknown extension, we would be unable to differ whether client passes a valid but yet unsupported extension, or it's a user error and a valid and supported extension is just misspelled.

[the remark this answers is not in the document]
MrSidims: Lets say we have a call to `__spirv_MyBrandNewInstruction` in LLVM IR, which is recognizable by SPIR-V translator as a part of `MyBrandNew` SPIR-V extension. Right now, if we pass it to the translator with the enabled `MyBrandNew` extension, then the translator will generate `OpMyBrandNewInstruction`. If `MyBrandNew` is not enabled - the translator will report an error. Meanwhile, if the extension **is not known** to the translator - it won't generate an error and instead will leave this external function call untouched (may be to consider to report an error in this case actually).

What is the backend's behavior, if it sees `__spirv_MyBrandNewInstruction` function call, but knows nothing about `MyBrandNew` extension? If it reports an error - it's great and I have no concerns with filtering out the extensions. If it doesn't report an error and just generates the external function call - well, it matches the translator's behavior for unknown extensions, but it doesn't match the translator's behavior for this very extension, so instead of truthfully fail with an error, saying that the extension passed along with --spirv-ext option is unsupported by the backend, the backend will silently emit some function call, which later SPIR-V consumer has to resolve somehow, and this might be a problem.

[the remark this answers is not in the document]
VyacheslavLevytskyy: Yes, indeed new users/use cases would get just OpFunctionCall instead of an error, as you've described in Translator's behavior. However, consider existing users.

For example, the Triton compiler enables all extensions. This means that without filtering in Translator's side calls from the Triton compiler will always result in an error.

Another example is SYCL very (really) long list of extensions. Without filtering on Translator side SYCL users need to refactor their calls to be sure that SPIRV BE serves their request.

In other words, if we don't use filtering, we just move the problem deeper, closer to an end-user. If we filter in SPIRV BE we lose information/chance to report misspelling names. Filtering in Translator does not solve all the problems, as your example demonstrated, but it looks like a minimal of all alternative issues.

As a quick and naive resolution of the problem that you've mentioned, it would be easier to introduce an option in SPIRV BE that controls whether __spriv_XYZ has to be resolved to a known instruction, by the analogy with Translator's option to allow unknown intrinsics by prefix.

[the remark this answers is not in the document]
MrSidims: > For example, the Triton compiler enables all extensions (just enableAllExtensions()). This means that without filtering in Translator's side calls from the Triton compiler will always result in an error.

Seems like not the best idea for me. Can't stop myself but wonder if they will report a 100% regression with the translator flow due to recent work related to opaque pointers extension started by Viktoria. Btw, is --spirv-ext=+all is being handled implicitly with this patch?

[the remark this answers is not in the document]
VyacheslavLevytskyy: This patch is just going through the list of extensions, after +all had already been parsed, so from the patch perspective it doesn't matter if the list was explicit or +all, the patch deals with parsed representation for both.

MrSidims (approved):
LGTM with added tests

lib/SPIRV/SPIRVWriter.cpp
```diff
@@ -6753,8 +6752,123 @@ bool isValidLLVMModule(Module *M, SPIRVErrorLog &ErrorLog) {
...
+      SPIRV::ExtensionID::SPV_INTEL_bfloat16_conversion,
+      SPIRV::ExtensionID::SPV_KHR_subgroup_rotate,
+      SPIRV::ExtensionID::SPV_INTEL_variable_length_array,
+      SPIRV::ExtensionID::SPV_INTEL_function_pointers,
+      SPIRV::ExtensionID::SPV_KHR_shader_clock,
+      SPIRV::ExtensionID::SPV_KHR_cooperative_matrix,
+      SPIRV::ExtensionID::SPV_KHR_non_semantic_info};
+  // The fallback for the Triple value.
+  static const std::string DefaultTriple = "spirv64-unknown-unknown";
+  // SPIR-V BE uses the following command line options to conform with
```
svenvh, line 6824:
```suggestion
  // SPIR-V backend uses the following command line options to conform with
```

tools/llvm-spirv/llvm-spirv.cpp
```diff
@@ -704,6 +720,27 @@ int main(int Ac, char **Av) {
...
+  // be used by a user anyway. After that we may safely add the instance of
+  // "spirv-ext" required by LLVM/SPIRV Translator from the corresponding auto
+  // variable (SPVExt).
+  StringMap<llvm::cl::Option *> &RegisteredOptions =
+      llvm::cl::getRegisteredOptions();
+  if (RegisteredOptions.count("spirv-ext") == 1) {
+    llvm::cl::Option *OptToDisable = RegisteredOptions["spirv-ext"];
+    OptToDisable->setArgStr("spirv-ext-coming-from-spirv-backend");
+    OptToDisable->setHiddenFlag(cl::Hidden);
+  }
```
svenvh, line 736: If you `#endif` here instead of below, would that allow you to remove the (now) duplicate static definition of `SPVExt` higher up in this file?

svenvh (commented):
Wondering if we should also add a configuration to the CI to test this integration... Any thoughts?

tools/llvm-spirv/llvm-spirv.cpp
VyacheslavLevytskyy: Good point, thank you!

VyacheslavLevytskyy:
> Wondering if we should also add a configuration to the CI to test this integration... Any thoughts?

I'd say that it's better to add changes into CI as well. At least, in ideal case we would need to test both cases when LLVM is built with/without SPIRV Backend support.

tools/llvm-spirv/llvm-spirv.cpp
```diff
@@ -116,12 +116,6 @@ static cl::opt<VersionNumber> MaxSPIRVVersion(
                clEnumValN(VersionNumber::SPIRV_1_6, "1.6", "SPIR-V 1.6")),
     cl::init(VersionNumber::MaximumVersion));
 
-static cl::list<std::string>
```
asudarsa, line 119: I am not sure if we should be removing this option. Anything that will modify the current behavior of the translator must be avoided. May be I am missing something here?

Thanks

tools/llvm-spirv/llvm-spirv.cpp
```diff
@@ -277,6 +271,11 @@ static cl::opt<SPIRV::BuiltinFormat> SPIRVBuiltinFormat(
         clEnumValN(SPIRV::BuiltinFormat::Global, "global",
                    "Use globals to represent SPIR-V builtin variables")));
 
+static cl::opt<bool> SPIRVUseLLVMSPIRVBackendTarget(
+    "spirv-use-llvm-backend-target",
+    cl::desc("Convert LLVM to SPIR-V using the LLVM SPIR-V Backend target"),
```
asudarsa, line 276: Do we want to add an extra sentence to say what will happen if SPIR-V target is not found? Also, add a quick phrase to specify default setting.

include/LLVMSPIRVOpts.h
```diff
@@ -238,6 +241,9 @@ class TranslatorOpts {
   }
   BuiltinFormat getBuiltinFormat() const noexcept { return SPIRVBuiltinFormat; }
 
+  void useLLVMTarget(bool Flag) noexcept { UseLLVMTarget = Flag; }
```
asudarsa, line 244: May be rename as setUseLLVMTarget?

include/LLVMSPIRVOpts.h
```diff
@@ -238,6 +241,9 @@ class TranslatorOpts {
...
+  bool getUseLLVMSPIRVBackendTarget() const noexcept { return UseLLVMTarget; }
```
asudarsa, line 245: May be rename UseLLVMTarget to be UseLLVMSPIRVBackendTarget everywhere (to be consistent)?

asudarsa (commented):
Changes look good to me (few nits). I am interested to know how this works on E2E tests. We can quickly make some changes in the LLVM driver and build scripts and test this out. I am specifically interested to know what happens if we use the SPIR-V backend for clang++ -fsycl test.cpp compilation. Please let me know if I can help with that.

Thanks

tools/llvm-spirv/llvm-spirv.cpp
VyacheslavLevytskyy: Sure

include/LLVMSPIRVOpts.h
VyacheslavLevytskyy: Sure

include/LLVMSPIRVOpts.h
VyacheslavLevytskyy: Let's please rename the function only, please see https://github.com/KhronosGroup/SPIRV-LLVM-Translator/pull/2716#discussion_r1756820076 for the rationale

VyacheslavLevytskyy:
> Changes look good to me (few nits). I am interested to know how this works on E2E tests. We can quickly make some changes in the LLVM driver and build scripts and test this out. I am specifically interested to know what happens if we use the SPIR-V backend for clang++ -fsycl test.cpp compilation. Please let me know if I can help with that.
> 
> Thanks

Thank you @asudarsa ! Suggestions are applied. As for E2E tests, hopefully after we merge this PR running the SYCL test suite with SPIR-V Backend should become easier, so that we may continue checking our options with that respect. One of easier/more flexible possibilities to test this in SYCL may be to implement a switch from Translator to SPIRV Backend having a new environment variable set ON. This may be even easier if we modify Translator's way of parsing command line in one of future P's either by supporting `const char *EnvVar` parameter cl::ParseCommandLineOptions(), or just directly adding a check of a kind `if sys::Process::GetEnv("SPIRV_USE_LLVM_BACKEND_TARGET") is "ON" then set SPIRVUseLLVMSPIRVBackendTarget true`. This would make it possible to utilize SPIR-V Backend transparently without a need to modify existing code base on the early stage.

asudarsa:
> > Changes look good to me (few nits). I am interested to know how this works on E2E tests. We can quickly make some changes in the LLVM driver and build scripts and test this out. I am specifically interested to know what happens if we use the SPIR-V backend for clang++ -fsycl test.cpp compilation. Please let me know if I can help with that.
> > Thanks
> 
> Thank you @asudarsa ! Suggestions are applied. As for E2E tests, hopefully after we merge this PR running the SYCL test suite with SPIR-V Backend should become easier, so that we may continue checking our options with that respect.
> 
> One of easier/more flexible possibilities to test this in SYCL may be to implement a switch from Translator to SPIRV Backend having a new environment variable set ON (a change in SYCL code base).
> 
> This may be even easier if we modify Translator's code base by changing the way of applying command line arguments in one of future PRs either by supporting `const char *EnvVar` parameter cl::ParseCommandLineOptions(), or just directly adding a check of a kind `if sys::Process::GetEnv("SPIRV_USE_LLVM_BACKEND_TARGET") is "ON" then set SPIRVUseLLVMSPIRVBackendTarget true`. This would make it possible to utilize SPIR-V Backend transparently without a need to modify existing SYCL code base on the early stage.

Hi @VyacheslavLevytskyy 

Thanks for the details. I think there is an alternate option. We already have a user-facing option that can be used to turn on the SPIR-V backend use. Currently, it is used to call 'llc' from the driver. We can refactor it to add the llvm-spirv option. Also, we can start building the SPIR-V backend by default in our build process. It is a quite small backend and also we are quite close to turning it on by default (atleast in intel/llvm compiler). 
I think we can move further discussions out of this PR as it is beyond the scope of this PR. 

Thanks

asudarsa (approved):
Looks good to me. It will be good to announce this in the OpenCL tooling WG at some point of time.

Thanks

include/LLVMSPIRVOpts.h
svenvh:
```suggestion
  bool getUseLLVMTarget() const noexcept { return UseLLVMTarget; }
```

VyacheslavLevytskyy: thank you

VyacheslavLevytskyy:
@asudarsa agree and thank you

PR #2718, opened 2024-09-13
Allow short16 for ushort16 intel_sub_group_block_read/write

MrSidims:
Per cl_intel_subgroups_short V 1.1.0 short16 is allowed for these builtins.

See: https://registry.khronos.org/OpenCL/extensions/intel/cl_intel_subgroups_short.html

Also add missing tests for char16

### DIFF
--- a/lib/SPIRV/OCLUtil.cpp
+++ b/lib/SPIRV/OCLUtil.cpp
@@ -1480,7 +1480,7 @@ std::string getIntelSubgroupBlockDataPostfix(unsigned ElementBitSize,
     break;
   case 16:
     assert((ElementBitSize == 8 || ElementBitSize == 16) &&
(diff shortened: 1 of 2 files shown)


lib/SPIRV/OCLUtil.cpp
```diff
@@ -1479,7 +1479,7 @@ std::string getIntelSubgroupBlockDataPostfix(unsigned ElementBitSize,
     OSS << VectorNumElements;
     break;
   case 16:
-    assert(ElementBitSize == 8 &&
+    assert((ElementBitSize == 8 || ElementBitSize == 16) &&
            "16 elements vector allowed only for char builtins");
```
MrSidims, line 1483: need to update this

vmaksimo (approved):

Issue #2721, opened 2024-09-18
Incorrect translation of calls to a builtin that returns a structure


VyacheslavLevytskyy:
Khronos Translator incorrectly translates calls to builtins that returns a structure. For example,
```
         %16 = OpIAddCarry %_struct_7 %e %uint_1
       %math = OpCompositeExtract %uint %16 0
```
is translated as
```
  %2 = alloca %structtype, align 8
  call spir_func void @_Z17__spirv_IAddCarryii(ptr sret(%structtype) %2, i32 %e, i32 1)
  %math = extractvalue ptr %2, 0
```
that generates incorrect `extractvalue` applied to a pointer.

A full reproducer is in the comment to the reported issue.

VyacheslavLevytskyy:
```
               OpCapability Kernel
               OpCapability Addresses
               OpCapability Int8
               OpCapability GenericPointer
               OpCapability Linkage
          %1 = OpExtInstImport "OpenCL.std"
               OpMemoryModel Physical64 OpenCL
               OpSource OpenCL_CPP 100000
               OpName %a "a"
               OpName %p "p"
               OpName %foo "foo"
               OpName %e "e"
               OpName %math "math"
               OpName %ov "ov"
               OpDecorate %foo LinkageAttributes "foo" Export
       %uint = OpTypeInt 32 0
      %uchar = OpTypeInt 8 0
%_ptr_Generic_uchar = OpTypePointer Generic %uchar
          %5 = OpTypeFunction %uint %uint %_ptr_Generic_uchar
       %bool = OpTypeBool
  %_struct_7 = OpTypeStruct %uint %uint
     %uint_1 = OpConstant %uint 1
   %uchar_42 = OpConstant %uchar 42
         %10 = OpConstantNull %uint
        %foo = OpFunction %uint None %5
          %a = OpFunctionParameter %uint
          %p = OpFunctionParameter %_ptr_Generic_uchar
         %19 = OpLabel
               OpBranch %20
         %20 = OpLabel
          %e = OpPhi %uint %a %19 %math %21
         %16 = OpIAddCarry %_struct_7 %e %uint_1
       %math = OpCompositeExtract %uint %16 0
         %17 = OpCompositeExtract %uint %16 1
         %ov = OpINotEqual %bool %17 %10
               OpBranchConditional %ov %22 %21
         %21 = OpLabel
               OpStore %p %uchar_42 Aligned 1
               OpBranch %20
[...]

PR #2722, opened 2024-09-18
Fix incorrect translation of calls to a builtin that returns a structure

VyacheslavLevytskyy:
This PR fixes https://github.com/KhronosGroup/SPIRV-LLVM-Translator/issues/2721
The root cause is that https://github.com/bwlodarcz/SPIRV-LLVM-Translator/blob/ca48eaf52ae791abc61cba13ef383f022e151e9f/lib/SPIRV/SPIRVUtil.cpp#L2207 doesn't preserve equivalence of values, inserting a pointer where a value of function return's type was before.

A new test case is added from the reproducer of https://github.com/KhronosGroup/SPIRV-LLVM-Translator/issues/2721
Existing test cases are passing as is.

### DIFF
--- a/lib/SPIRV/SPIRVUtil.cpp
+++ b/lib/SPIRV/SPIRVUtil.cpp
@@ -2253,7 +2253,8 @@ bool postProcessBuiltinReturningStruct(Function *F) {
       NewCI->setCallingConv(CI->getCallingConv());
       SmallVector<User *> CIUsers(CI->users());
       for (auto *CIUser : CIUsers) {
-        CIUser->replaceUsesOfWith(CI, A);
+        CIUser->replaceUsesOfWith(CI,
+                                  Builder.CreateLoad(F->getReturnType(), A));
       }
       InstToRemove.push_back(CI);
     }
--- a/lib/SPIRV/SPIRVUtil.cpp
+++ b/lib/SPIRV/SPIRVUtil.cpp
@@ -2227,8 +2227,9 @@ bool postProcessBuiltinReturningStruct(Function *F) {
       Builder.SetInsertPoint(CI);
       SmallVector<User *> Users(CI->users());
       Value *A = nullptr;
+      StoreInst *SI = nullptr;
       for (auto *U : Users) {
-        if (auto *SI = dyn_cast<StoreInst>(U)) {
+        if ((SI = dyn_cast<StoreInst>(U)) != nullptr) {
           A = SI->getPointerOperand();
           InstToRemove.push_back(SI);
           break;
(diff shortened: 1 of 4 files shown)


lib/SPIRV/SPIRVUtil.cpp
```diff
@@ -2253,7 +2253,8 @@ bool postProcessBuiltinReturningStruct(Function *F) {
       NewCI->setCallingConv(CI->getCallingConv());
       SmallVector<User *> CIUsers(CI->users());
       for (auto *CIUser : CIUsers) {
-        CIUser->replaceUsesOfWith(CI, A);
+        CIUser->replaceUsesOfWith(CI,
```
MrSidims, line 2256: This would be a little awkward, but we have to check if the user is Load. If it's Load we should either:
- not create a new Load;
- remove the old load
Because otherwise...

test/llvm-intrinsics/uadd.with.overflow.ll
```diff
@@ -143,17 +145,18 @@ entry:
...
-; CHECK-LLVM:   %3 = extractvalue [[structtype_3]] %1, 1
-; CHECK-LLVM:   %4 = icmp ne i64 %3, 0
-; CHECK-LLVM:   %5 = insertvalue [[structtype_4]] undef, i64 %2, 0
-; CHECK-LLVM:   %6 = insertvalue [[structtype_4]] %5, i1 %4, 1
-; CHECK-LLVM:   %7 = extractvalue [[structtype_4]] %6, 0
-; CHECK-LLVM:   %8 = extractvalue [[structtype_4]] %6, 1
-; CHECK-LLVM:   ret i1 %8
+; CHECK-LLVM:   %0 = alloca %structtype.3, align 8
+; CHECK-LLVM:   call spir_func void @_Z17__spirv_IAddCarryll(ptr sret(%structtype.3) %0, i64 %a, i64 %b)
+; CHECK-LLVM:   %1 = load %structtype.3, ptr %0, align 8
```
MrSidims, line 150: ...we will be left with the dangling load here.

lib/SPIRV/SPIRVUtil.cpp
VyacheslavLevytskyy: good point, thank you

VyacheslavLevytskyy: Updated, please have a look. The point was not in load user, but rather in that (1) we shouldn't update StoreInst that is going to be removed further, (2) it's enough to have one load per all users.

MrSidims (approved):

svenvh (approved):

PR #2723, opened 2024-09-19
Fix OCL builtins translation in case SPV_KHR_untyped_pointers is enabled

vmaksimo:
This change allows to preserve the correct builtin mangling in reverse translation.

### DIFF
--- a/lib/SPIRV/SPIRVReader.cpp
+++ b/lib/SPIRV/SPIRVReader.cpp
@@ -4757,6 +4757,17 @@ Instruction *SPIRVToLLVM::transOCLBuiltinFromExtInst(SPIRVExtInst *BC,
          "Not OpenCL extended instruction");
 
   std::vector<Type *> ArgTypes = transTypeVector(BC->getArgTypes(), true);
+  for (unsigned I = 0; I < ArgTypes.size(); I++) {
+    // Special handling for "truly" untyped pointers to preserve correct OCL
+    // bultin mangling.
+    if (isa<PointerType>(ArgTypes[I]) && BC->getArg(I)->isUntypedVariable()) {
+      auto *BVar = static_cast<SPIRVUntypedVariableKHR *>(BC->getArg(I));
+      ArgTypes[I] = TypedPointerType::get(
+          transType(BVar->getDataType()),
+          SPIRSPIRVAddrSpaceMap::rmap(BVar->getStorageClass()));
+    }
+  }
+
   Type *RetTy = transType(BC->getType());
   std::string MangledName =
       getSPIRVFriendlyIRFunctionName(ExtOp, ArgTypes, RetTy);
--- a/lib/SPIRV/libSPIRV/SPIRVInstruction.h
+++ b/lib/SPIRV/libSPIRV/SPIRVInstruction.h
@@ -2003,6 +2003,12 @@ public:
     }
     return VArgs;
   }
+  SPIRVValue *getArg(SPIRVWord I) {
+    if (isOperandLiteral(I))
+      return Module->getLiteralAsConstant(Args[I]);
+    return getValue(Args[I]);
+  }
+
   std::vector<SPIRVType *> getArgTypes() {
     std::vector<SPIRVType *> ArgTypes;
     auto VArgs = getArgValues();
--- a/test/transcoding/float16.ll
+++ b/test/transcoding/float16.ll
@@ -1,7 +1,13 @@
 ; RUN: llvm-as %s -o %t.bc
 ; RUN: llvm-spirv %t.bc -o %t.spv
 ; RUN: spirv-val %t.spv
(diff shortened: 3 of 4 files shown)


test/transcoding/float16.ll
```diff
@@ -1,23 +1,31 @@
 ; RUN: llvm-as %s -o %t.bc
 ; RUN: llvm-spirv %t.bc -o %t.spv
 ; RUN: spirv-val %t.spv
-; RUN: llvm-spirv %t.spv -to-text -o - | FileCheck %s --check-prefix=CHECK-SPIRV
+; RUN: llvm-spirv %t.spv -to-text -o - | FileCheck %s --check-prefixes=CHECK-SPIRV,CHECK-SPIRV-NOEXT
+; RUN: llvm-spirv -r %t.spv -o %t.rev.bc
+; RUN: llvm-dis %t.rev.bc -o - | FileCheck %s --check-prefix=CHECK-LLVM
+
+; RUN: llvm-spirv %t.bc -o %t.spv --spirv-ext=+SPV_KHR_untyped_pointers
+; RUN: spirv-val %t.spv
```
vmaksimo, line 9:
```suggestion
; TODO: enable back once spirv-tools are updated
; R/UN: spirv-val %t.spv
```

lib/SPIRV/libSPIRV/SPIRVInstruction.h
```diff
@@ -2003,6 +2003,12 @@ class SPIRVExtInst : public SPIRVFunctionCallGeneric<OpExtInst, 5> {
     }
     return VArgs;
   }
+  SPIRVValue *getArg(SPIRVWord I) {
```
LU-JOHN, line 2006: I think this should be renamed getArgValue to be consistent with the pre-existing getArgValues which returns a vector pointers.  Also, consider implementing getArgValues with getArgValue.

test/transcoding/float16.ll
```diff
@@ -1,23 +1,32 @@
 ; RUN: llvm-as %s -o %t.bc
 ; RUN: llvm-spirv %t.bc -o %t.spv
 ; RUN: spirv-val %t.spv
-; RUN: llvm-spirv %t.spv -to-text -o - | FileCheck %s --check-prefix=CHECK-SPIRV
+; RUN: llvm-spirv %t.spv -to-text -o - | FileCheck %s --check-prefixes=CHECK-SPIRV,CHECK-SPIRV-NOEXT
+; RUN: llvm-spirv -r %t.spv -o %t.rev.bc
+; RUN: llvm-dis %t.rev.bc -o - | FileCheck %s --check-prefix=CHECK-LLVM
+
+; RUN: llvm-spirv %t.bc -o %t.spv --spirv-ext=+SPV_KHR_untyped_pointers
```
LU-JOHN, line 8: How did you pick this file to do additional testing with SPV_KHR_untyped_pointers?
Please add a comment explaining why this additional testing is necessary.

lib/SPIRV/SPIRVReader.cpp
```diff
@@ -4757,6 +4757,17 @@ Instruction *SPIRVToLLVM::transOCLBuiltinFromExtInst(SPIRVExtInst *BC,
          "Not OpenCL extended instruction");
 
   std::vector<Type *> ArgTypes = transTypeVector(BC->getArgTypes(), true);
+  for (unsigned I = 0; I < ArgTypes.size(); I++) {
```
LU-JOHN, line 4760: Does function transBuiltinFromInst() also need this special handling?  It also calls transTypeVector() and gets a mangled name.

test/transcoding/float16.ll
vmaksimo: Actually I'm just running all the tests with the extension enabled and trying to understand where the real fixes should be made, or where we just need to update Checks.
This one is testing OCL builtin so I've checked if we are working fine with them when the extension enabled (turned out that not).
Will add the comment about additional testing though.

lib/SPIRV/SPIRVReader.cpp
vmaksimo: My experiments show that e.g. for `OpEnqueKernel` we don't meet untyped variables in arguments, but the mangling is still wrong - should be fixed elsewhere.
It's still possible to unify this fix inside `transTypeVector()`, but not sure if it'd be used by `transBuiltinFromInst()`.

MrSidims (approved):
Thanks!

lib/SPIRV/SPIRVReader.cpp
MrSidims: I'm fine with the merge as is for now. In general we probably should implement the following algorithm:
- translate SPIR-V pointer type as LLVM's Typed pointer
- get the mangling
- 'opaquefy' the pointer to construct the function and call instruction

svenvh (approved):

PR #2724, opened 2024-09-23
Update after Instruction iterator-insertion deprecation

svenvh:
Avoid deprecation warnings after LLVM commit 2f50b280dc8e ("[DebugInfo] Enable deprecation of iterator-insertion methods (#102608)", 2024-09-20).


MrSidims (commented):
Sorry for the late review, the patch is LGTM :)

PR #2727, opened 2024-09-25
Centralise & tweak handling of SyncScope to SPIR-V translation

AlexVlx:
This patch collects the mapping from LLVM SyncScope ID, which was spread out across multiple sites, into a single utility function. Furthermore, it realigns the mapping to match LLVM conventions, namely it defaults to `System` / `CrossDevice` (please see <https://github.com/llvm/llvm-project/pull/106429>) for more context for the proposed changes.

### DIFF
--- a/test/atomic-load-store.ll
+++ b/test/atomic-load-store.ll
@@ -3,11 +3,10 @@
 ; RUN: spirv-val %t.spv
 ; RUN: llvm-spirv -to-text %t.spv -o - | FileCheck %s
 
-; CHECK-DAG: Constant [[#]] [[#Relaxed:]] 0
-; CHECK-DAG: Constant [[#]] [[#DeviceScope:]] 1
-; CHECK-DAG: Constant [[#]] [[#Acquire:]] 2
+; CHECK-DAG: Constant [[#]] [[#CrossDeviceScope:]] 0
 ; CHECK-DAG: Constant [[#]] [[#Release:]] 4
 ; CHECK-DAG: Constant [[#]] [[#SequentiallyConsistent:]] 16
+; CHECK-DAG: Constant [[#]] [[#Acquire:]] 2
 
 target datalayout = "e-i64:64-v16:16-v24:32-v32:32-v48:64-v96:128-v192:256-v256:256-v512:512-v1024:1024"
 target triple = "spir64"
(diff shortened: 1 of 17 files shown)


test/atomic-load-store.ll
```diff
@@ -18,18 +17,18 @@ entry:
 ; CHECK: Variable [[#]] [[#PTR:]]
   %0 = alloca i32
 
-; CHECK: AtomicStore [[#PTR]] [[#DeviceScope]] [[#Relaxed]] [[#]]
+; CHECK: AtomicStore [[#PTR]] [[#CrossDeviceScope]] {{.+}} [[#]]
```
MrSidims, line 20: To double check, the memory order is still Relaxed, you just don't want to reuse [[#CrossDeviceScope:]] constant in the check, right?

test/atomicrmw.ll
```diff
@@ -26,37 +25,37 @@ target triple = "spir64"
 define dso_local spir_func void @test_atomicrmw() local_unnamed_addr #0 {
 entry:
   %0 = atomicrmw xchg ptr addrspace(1) @ui, i32 42 acq_rel
-; CHECK: AtomicExchange [[Int]] {{[0-9]+}} [[Pointer]] [[Scope_Device]] [[MemSem_AcquireRelease]] [[Value]]
+; CHECK: AtomicExchange [[Int]] {{[0-9]+}} [[Pointer]] {{.+}} [[MemSem_AcquireRelease]] [[Value]]
```
MrSidims, line 28: And here is an opposite situation?

test/atomic-load-store.ll
AlexVlx: Correct, re-using the same "named" variable in lit is not quite possible (or I've not figured out how), and instead of just picking something like `Const0` it seemed more desirable to pattern match one and keep the other fully described (it retains most of the information about what is being tested).

test/atomicrmw.ll
AlexVlx: Yes, just to have at least one test where Relaxed explicitly shows up. I can flip it to pattern match like the other test, if you think that is preferable.

MrSidims: I don't have strong objections against the current version of the tests

MrSidims:
LGTM
@AlexVlx please apply clang-format suggestion

MrSidims (approved):

svenvh (approved):

Issue #2702, opened 2024-09-02
Violation of the requirement to have exactly one Parent ID operand for each parent block of the current block in the CFG


VyacheslavLevytskyy:
It's a requirement of the SPIR-V specification (https://registry.khronos.org/SPIR-V/specs/unified1/SPIRV.html#OpPhi) that "There must be exactly one Parent i for each parent block of the current block in the CFG.". However, for the snippet `%retval.0.i = phi i32 [ 0, %sw.default.i ], [ 1, %entry ], [ 1, %entry ]` from the following reproducer

```
define spir_kernel void @foo(i8 %arg) {
entry:
  switch i8 %arg, label %sw.default.i [
    i8 -127, label %exit
    i8 -111, label %exit
  ]

sw.default.i:                                     ; preds = %entry
  br label %exit

exit: ; preds = %sw.default.i, %entry, %entry
  %retval.0.i = phi i32 [ 0, %sw.default.i ], [ 1, %entry ], [ 1, %entry ]
  ret void
}
```

Translator would generate invalid SPIR-V output containing the following fragment:

```
%retval_0_i = OpPhi %uint %uint_0 %sw_default_i %uint_1 %entry %uint_1 %entry
```

When validating Translator's output we get the following error:

```
$ llvm-as phi-multiple-preds.ll -o - | /localdisk2/vlevytsk/sycl/llvm/build/bin/llvm-spirv.bak -o - | spirv-val
error: line 26: OpPhi's number of incoming blocks (3) does not match block's predecessor count (3).
  %retval_0_i = OpPhi %uint %uint_0 %sw_default_i %uint_1 %entry %uint_1 %entry
```

SPIR-V Backend generates correctly `%retval_0_i = OpPhi %uint %uint_0 %13 %uint_1 %12` where there is just one record per 2 original `%entry` predecessors, however, this output of SPIR-V Backend gets rejected by Translator's reverse conversion of SPIR-V to LLVM IR as the following:

```
$ llvm-spirv -r phi-preds.spv
Fails to verify module: PHINode should have one entry for each predecessor of its parent basic block!
  %retval.0.i = phi i32 [ 0, %2 ], [ 1, %0 ]
```

This error is caused by the 1:1 mapping of PHI operands back to LLVM IR so that llvm's machine verifier gets unhappy when it sees just 1 phi entry for two predecessors.

[...]

PR #2736, opened 2024-10-01
Ensure that PHI node has an incoming value per each predecessor instance

VyacheslavLevytskyy:
This PR partially fixes issue https://github.com/KhronosGroup/SPIRV-LLVM-Translator/issues/2702 in the part that is responsible for SPIR-V to LLVM IR translation. Namely, this PR ensures that all PHI nodes of a `Function` has the number of incoming blocks matching block's predecessor count. When a PHI node doesn't conform to this rule, this PR inserts missing number of (Value, Basic Block) pairs to make the PHI node valid.

Another problem from https://github.com/KhronosGroup/SPIRV-LLVM-Translator/issues/2702, that is violation of the requirement to OpPhi's to have exactly one Parent ID operand for each parent block of the current block in the CFG in the output SPIR-V code, is out of scope of this PR.

### DIFF
--- a/lib/SPIRV/SPIRVReader.cpp
+++ b/lib/SPIRV/SPIRVReader.cpp
@@ -3070,6 +3070,47 @@ void SPIRVToLLVM::transFunctionAttrs(SPIRVFunction *BF, Function *F) {
   });
 }
 
+// One basic block can be a predecessor to another basic block more than
+// once (https://github.com/KhronosGroup/SPIRV-LLVM-Translator/issues/2702).
+// This function checks if created PHI's require a fix wrt. this rule.
+static void validatePhiPredecessors(Function *F) {
+  for (BasicBlock &BB : *F) {
+    bool UniquePreds = true;
+    DenseMap<BasicBlock *, unsigned> PredsCnt;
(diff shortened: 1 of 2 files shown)


lib/SPIRV/SPIRVReader.cpp
```diff
@@ -3070,6 +3070,49 @@ void SPIRVToLLVM::transFunctionAttrs(SPIRVFunction *BF, Function *F) {
...
+    }
+    if (UniquePreds)
+      continue;
+    // `phi` requires an incoming value per each predecessor instance, even
+    // it's the same basic block that has been already inserted as an incoming
+    // value of the `phi`.
+    for (Instruction &Instr : BB) {
+      PHINode *Phi = dyn_cast<PHINode>(&Instr);
+      if (!Phi)
+        continue;
```
svenvh, line 3096: We should be able to break out of the loop as soon as we see a non-PHI instruction, right?  Alternatively, could we use `BasicBlock::phis()`?

svenvh (approved):

MrSidims (approved):

Issue #2670, opened 2024-08-12
handling UserSemantic decoration on Input variables


bashbaug:
I've been slowly adding testing for SPIR-V 1.4 features and I found an issue UserSemantic decorations on Input variables.  My test can be found here:

https://github.com/KhronosGroup/OpenCL-CTS/compare/main...bashbaug:OpenCL-CTS:spirv-14-part3

There seems to be two problems:

1. The SPIR-V LLVM Translator currently doesn't handle the OpDecorateString and OpMemberDecorateString instructions (see also #2460).  This one seems pretty straightforward, and we should just parse OpDecorateString similar to the way we are currently parsing OpDecorate, at least for now.  Once we sort out the difference between OpDecorate and OpDecorateString we might only want to parse the UserSematic decoration using OpDecorateString, but we can cross that bridge when we come to it.

2. The SPIR-V LLVM Translator also doesn't seem to be properly handling the UserSemantic decoration on Input variables.  When there are UserSemantic decorations on an input variable, it turns into an "llvm.global.annotation":

```llvm
@__spirv_BuiltInGlobalInvocationId = external addrspace(7) constant <3 x i64>
@0 = private unnamed_addr constant [4 x i8] c"FOO\00", section "llvm.metadata"
@1 = private unnamed_addr constant [4 x i8] c"BAR\00", section "llvm.metadata"
@2 = private unnamed_addr constant [15 x i8] c"FOO? BAR. BAZ!\00", section "llvm.metadata"
@llvm.global.annotations = appending global [3 x { ptr, ptr, ptr, i32, ptr }] [{ ptr, ptr, ptr, i32, ptr } { ptr addrspacecast (ptr addrspace(7) @__spirv_BuiltInGlobalInvocationId to ptr), ptr @0, ptr undef, i32 undef, ptr undef }, { ptr, ptr, ptr, i32, ptr } { ptr addrspacecast (ptr addrspace(7) @__spirv_BuiltInGlobalInvocationId to ptr), ptr @1, ptr undef, i32 undef, ptr undef }, { ptr, ptr, ptr, i32, ptr } { ptr addrspacecast (ptr addrspace(7) @__spirv_BuiltInGlobalInvocationId to ptr), ptr @2, ptr undef, i32 undef, ptr undef }], section "llvm.metadata"
```

This case is not currently handled by `replaceUsesOfBuiltinVar`, so we hit an `llvm_unreachable` and we crash.

[...]

PR #2742, opened 2024-10-07
Ignore UserSemantic decorations on BuiltIn variables

svenvh:
UserSemantic decorations on BuiltIn variables would be added into `llvm.global.annotations`.  During lowering of builtin variables to function calls, `replaceUsesOfBuiltinVar` would not know how to handle the annotation and crash.

Right now there is no clear use case for UserSemantic decorations on BuiltIn variables, so just ignore the decoration and drop it during translation from SPIR-V to LLVM.

Test-case-by: Ben Ashbaugh <ben.ashbaugh@intel.com>

### DIFF
--- a/lib/SPIRV/SPIRVReader.cpp
+++ b/lib/SPIRV/SPIRVReader.cpp
@@ -3911,6 +3911,13 @@ void SPIRVToLLVM::transIntelFPGADecorations(SPIRVValue *BV, Value *V) {
     if (AnnotationCall && !AnnotationCall->getType()->isVoidTy())
       ValueMap[BV] = AnnotationCall;
   } else if (auto *GV = dyn_cast<GlobalVariable>(V)) {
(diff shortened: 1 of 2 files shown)


svenvh:
I've created this as a draft now as we're still not completely sure if we want to just drop these decorations.  This PR right now just drops the decoration (and avoids crashing on the input).

Contributes to/fixes https://github.com/KhronosGroup/SPIRV-LLVM-Translator/issues/2670

MrSidims (approved):
I'm fine with the approach. If we see a value we can add handling via [spirv.Decorations metadata](https://github.com/KhronosGroup/SPIRV-LLVM-Translator/blob/main/docs/SPIRVRepresentationInLLVM.rst#id25).

bashbaug (approved):
Works for me.  Thanks!

Issue #417, opened 2020-01-27
Error : InvalidArraySize: Array size must be at least 1


davidrohr:
I am compiling the attached example with llvm (commit) 99464b7eb1d7bbeb7380a6ccc19b0ffdb2f08d40 and then convert it to spirv (master of converter):

```
clang++ -cl-std=clc++ -x cl -emit-llvm --target=spir64-unknown-unknown -O0 -cl-denorms-are-zero -cl-mad-enable -cl-no-signed-zeros -ferror-limit=1000 -Xclang -finclude-default-header -Dcl_clang_storage_class_specifiers -c test.cl -o test.bc
llvm-spirv test.bc
```

It fails with the following error:
```
InvalidArraySize: Array size must be at least 1: [0 x i16] [Src: /home/qon/clang-build/SPIRV-LLVM-Translator/lib/SPIRV/SPIRVWriter.cpp:352 T->getArrayNumElements() >= 1 ]
Stack dump:
0.      Program arguments: /home/qon/clang-build/SPIRV-LLVM-Translator/build/tools/llvm-spirv/llvm-spirv test.bc 
1.      Running pass 'LLVMToSPIRV' on module 'test.bc'.
 #0 0x00005569e6e63a55 llvm::sys::PrintStackTrace(llvm::raw_ostream&) /home/qon/clang-build/llvm-project/llvm/lib/Support/Unix/Signals.inc:564:22
 #1 0x00005569e6e63ae8 PrintStackTraceSignalHandler(void*) /home/qon/clang-build/llvm-project/llvm/lib/Support/Unix/Signals.inc:625:1
 #2 0x00005569e6e61891 llvm::sys::RunSignalHandlers() /home/qon/clang-build/llvm-project/llvm/lib/Support/Signals.cpp:68:20
 #3 0x00005569e6e633d5 SignalHandler(int) /home/qon/clang-build/llvm-project/llvm/lib/Support/Unix/Signals.inc:406:1
 #4 0x00007f7a6c8ae370 __restore_rt (/lib64/libpthread.so.0+0x15370)
 #5 0x00007f7a6c2b71e1 raise (/lib64/libc.so.6+0x3a1e1)
 #6 0x00007f7a6c29f55b abort (/lib64/libc.so.6+0x2255b)
[...]

PR #2743, opened 2024-10-07
Translate LLVM-IR zero-length arrays to 1-length arrays in SPIR-V

lbushi25:
The LLVM-IR type of zero-length array does not seem to have a mapping into SPIR-V type, rather it outputs an error since the obvious choice of zero-length array is not valid in SPIR-V. This, for example, prohibits us from using unbounded arrays in SYCL device kernels because they are typically represented in LLVM-IR through zero-length arrays. This PR attempts to introduce a workaround to this by lowering a zero-length array in LLVM-IR to a 1-length array in SPIR-V.


lib/SPIRV/SPIRVWriter.cpp
lbushi25, line 1: Only line 452-469 have functional changes in this file. The rest are formatter artifacts.

svenvh (changes requested):
Unconditionally generating non-conformant SPIR-V seems not right.  The suggestion in https://github.com/KhronosGroup/SPIRV-LLVM-Translator/issues/417 was to hide such functionality behind an option (or perhaps we should consider a SPIR-V extension?).

lbushi25:
> Unconditionally generating non-conformant SPIR-V seems not right. The suggestion in #417 was to hide such functionality behind an option (or perhaps we should consider a SPIR-V extension?).

Why would this be non-conformant SPIR-V?
I'm not generating arrays of size zero. I'm instead modifying the lowering of LLVM IR arrays of zero-length to SPIR-V arrays of length 1 which are, at least to the best of my knowledge, valid in SPIR-V

svenvh:
> Why would this be non-conformant SPIR-V? I'm not generating arrays of size zero. I'm instead modifying the lowering of LLVM IR arrays of zero-length to SPIR-V arrays of length 1 which are, at least to the best of my knowledge, valid in SPIR-V.

My apologies, I completely misread your initial description.  Lowering to a SPIR-V array of length 1 is of course completely conformant.

lib/SPIRV/SPIRVWriter.cpp
svenvh: Ideally we should leave out the formatter artifacts as they may complicate potential backports of this commit.  This repo ought to be clang-formatted, so we should fix the formatting in a separate PR; I can do so if you wish.

lbushi25: > Ideally we should leave out the formatter artifacts as they may complicate potential backports of this commit. This repo ought to be clang-formatted, so we should fix the formatting in a separate PR; I can do so if you wish.
> 
> Update: I've created #2744 . When that has landed, you can just rebase on top of `main` and the unrelated formatting changes should disappear from your patch.

Thanks for creating the formatting patch. I have merged it into my branch and also made the changes you suggested.

svenvh (approved):

MrSidims (approved):

PR #2745, opened 2024-10-08
Fix unannotated fall-through warning in SPIRVWriter.cpp

DavidSpickett:
By adding the missing `break`.
```
SPIRV-LLVM-Translator/lib/SPIRV/SPIRVWriter.cpp:3718:5: warning: unannotated fall-through between switch labels [-Wimplicit-fallthrough]
 3718 |     default:
      |     ^
SPIRV-LLVM-Translator/lib/SPIRV/SPIRVWriter.cpp:3718:5: note: insert 'break;' to avoid fall-through
 3718 |     default:
      |     ^
      |     break;
```
As the default case was just `break`, the behaviour is the same but this should prevent future mistakes here.


MrSidims (approved):
Thanks!

svenvh (approved):

PR #2746, opened 2024-10-08
Fix string to bool conversion warnings in some asserts

DavidSpickett:
```
/SPIRV-LLVM-Translator/lib/SPIRV/SPIRVRegularizeLLVM.cpp:535:15: warning: implicit conversion turns string literal into bool: 'const char[39]' to 'bool' [-Wstring-conversion]
  535 |       assert(!"Cache controls must decorate a pointer");
      |              ~^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
/usr/include/assert.h:93:27: note: expanded from macro 'assert'
   93 |      (static_cast <bool> (expr)                                         \
      |                           ^~~~
```
The original code is not wrong but `false &&` is only a few characters more and does the same thing without warning.


MrSidims (approved):

DavidSpickett:
I see clang-tidy issues, will fix shortly.

svenvh (approved):

PR #2748, opened 2024-10-09
UnXFAIL untyped_var test after spirv-tools update

svenvh:
With the latest release of spirv-tools (1.3.296) this test now passes.


vmaksimo (approved):
Thanks! I'll uncomment `spirv-val` runs in other tests in a follow-up PR

PR #2752, opened 2024-10-11
SPV_KHR_untyped_pointers - implement OpUntypedPrefetchKHR

vmaksimo:
When the extension is enabled, we should replace `prefetch` OpenCL ExtInst with the `OpUntypedPrefetchKHR` instruction.

Spec:
https://htmlpreview.github.io/?https://github.com/KhronosGroup/SPIRV-Registry/blob/main/extensions/KHR/SPV_KHR_untyped_pointers.html


lib/SPIRV/libSPIRV/SPIRVInstruction.h
```diff
@@ -4225,5 +4225,82 @@ _SPIRV_OP(ConvertHandleToSamplerINTEL)
...
+  }
+
+  SPIRVUntypedPrefetchKHR() : SPIRVInstruction(OC) {
+    setHasNoId();
+    setHasNoType();
+  }
+
+  void validate() const override {
+    SPIRVInstruction::validate();
+    assert(getValueType(PtrTy)->isTypePointer());
```
MrSidims, line 4257: Lets use
```
 SPIRVErrorLog &SPVErrLog = this->getModule()->getErrorLog();
 SPVErrLog.checkError(....
```

lib/SPIRV/libSPIRV/SPIRVInstruction.h
```diff
@@ -4225,5 +4225,82 @@ _SPIRV_OP(ConvertHandleToSamplerINTEL)
...
+    assert(RW.empty() || (RW.size() == 1 && getValueType(RW[0])->isTypeInt()));
+    assert(Locality.empty() ||
+           (Locality.size() == 1 && getValueType(Locality[0])->isTypeInt()));
+    assert(CacheTy.empty() ||
+           (CacheTy.size() == 1 && getValueType(CacheTy[0])->isTypeInt()));
+  }
+
+  void setWordCount(SPIRVWord TheWordCount) override {
+    SPIRVEntry::setWordCount(TheWordCount);
+    if (TheWordCount > 3)
```
MrSidims, line 4270: I might be missing something, but why do we need these checks?

lib/SPIRV/SPIRVReader.cpp
```diff
@@ -2475,6 +2475,36 @@ Value *SPIRVToLLVM::transValueWithoutDecoration(SPIRVValue *BV, Function *F,
...
+  case OpUntypedPrefetchKHR: {
+    // Do the same as transOCLBuiltinFromExtInst() but for OpUntypedPrefetchKHR.
+    auto *BC = static_cast<SPIRVUntypedPrefetchKHR *>(BV);
+
+    std::vector<Type *> ArgTypes =
+        transTypeVector(BC->getValueTypes(BC->getArguments()), true);
+    Type *RetTy = Type::getVoidTy(*Context);
+
+    std::string MangledName =
+        getSPIRVFriendlyIRFunctionName(OpenCLLIB::Prefetch, ArgTypes, RetTy);
```
MrSidims, line 2487: I started to remember (missed it in our chat). We need to map the instruction to llvm.prefetch intrinsic. So probably the logic should be:
1. if RW, locality and cache type optional parameters are set, then map the instruction to llvm.prefetch;
2. else if some of the optional parameters present, then we should emit SPIR-V friendly LLVM IR call (btw, @alan-baker I believe some routine lines should be added to the spec like: "if Cache Type present, then other optional parameters must be set");
3. if no optional parameters are set - map it to OpenCL prefetch call.

Note, No.2 is discussible, we may want to still emit llvm intrinsic but with some default values set. In the case No.2 RW will always be set, which is good as we won't need to guess a default value, locality can be defaulted to '0'. Not sure though for Cache Type, '1' seems reasonable to me. Would be nice to hear from @alan-baker @bashbaug and @svenvh about that.

lib/SPIRV/SPIRVWriter.cpp
```diff
@@ -5329,6 +5329,12 @@ SPIRVValue *LLVMToSPIRVBase::transDirectCallInst(CallInst *CI,
         BM->addExtension(
             ExtensionID::SPV_EXT_relaxed_printf_string_address_space);
       }
+    } else if (DemangledName.find("__spirv_ocl_prefetch") != StringRef::npos) {
```
MrSidims, line 5332: Please also add translation from prefetch llvm intrinsic.

lib/SPIRV/libSPIRV/SPIRVInstruction.h
vmaksimo: This is the way to handle optional arguments that is used through the translator's codebase. For the current test case without optional parameters these checks are not used, but need to verify if they are  in case we have optional operands.

lib/SPIRV/SPIRVReader.cpp
alan-baker: We could add that restriction in the future if that is the only practical use case. I left it vague because I am not familiar with generating the equivalent LLVM intrinsic.

As for defaults I simply tried to provide the same arguments as the LLVM intrinsics, though I can't imagine how instruction cache fetching could be done via OpenCL C. So for Cache Type I'd expect it's always safe to default to data.

vmaksimo: About p.1, I'm not sure we should use `llvm.prefetch` intrinsic for now because it does not contain the number of bytes to prefetch field.
The similar concern is for forward translation. Although `llvm.prefetch` contains `RW`, `Locality` and `Cache Type` fields, it does not have a number of bytes to prefetch, which is a mandatory field in `OpUntypedPrefetchKHR` instruction.

MrSidims: > it does not contain the number of bytes to prefetch field

Indeed. Probably it can go to another PR, but we can add the following logic:
1. (as before) if no optional parameters are set - map it to OpenCL prefetch call;
2. if one of the optional parameters (RW, locality or cache type) optional parameters is set and number of bytes = 64, then map it on llvm.prefetch (on x86-64 prefetch would prefetch 64 bytes, other magic number is also possible);
3. if optional parameter present and number of bytes != 64 - leave it as SPIR-V friendly call.

MrSidims (approved):
Feel free to address llvm.prefetch later/create a tracker.
I'd also like to hear from @svenvh if he approves the PR

svenvh (approved):

MrSidims:
Clang-tidy error is misleading, merging the PR.
/home/runner/work/SPIRV-LLVM-Translator/SPIRV-LLVM-Translator/lib/SPIRV/SPIRVReader.cpp:2502:15: error: 'CI' is confusable with 'C1' [misc-confusable-identifiers,-warnings-as-errors]
 2502 |     CallInst *CI = CallInst::Create(F, Args, BC->getName(), BB);

      |               ^
/home/runner/work/SPIRV-LLVM-Translator/SPIRV-LLVM-Translator/lib/SPIRV/SPIRVReader.cpp:2036:14: note: other declaration found here
 2036 |     unsigned C1 = M1->getType()->getArrayNumElements();
      |              ^

PR #2759, opened 2024-10-16
Add missing requirement spirv-dis in GroupAndSubgroupInstructions.spv…

mateuszchudyk:
…asm test


mateuszchudyk:
Some actions failed on tests DebugInfo/X86/constant-loclist.ll and transcoding/pipe_builtins.cl. This PR doesn't affect on these tests so the problem seems to be on HEAD.

MrSidims (approved):

PR #2760, opened 2024-10-16
Add missing requirements spirv-val to tests

mateuszchudyk:
There is a feature that checks if spirv-val exists, but no test that uses spirv-val requires that it exists.

### DIFF
--- a/test/CXX/global-ctor.cl
+++ b/test/CXX/global-ctor.cl
@@ -1,3 +1,4 @@
+// REQUIRES: spirv-val
 // RUN: %clang_cc1 -cl-std=clc++ -emit-llvm-bc -triple spir -O0 %s -o %t.bc
 // RUN: llvm-spirv %t.bc -o %t.spv
 // RUN: spirv-val %t.spv
--- a/test/ComparePointers.cl
+++ b/test/ComparePointers.cl
@@ -8,6 +8,7 @@ kernel void test(int global *in, int global *in2) {
   if (in < in2)
     return;
 }
+// REQUIRES: spirv-val
 // RUN: %clang_cc1 -triple spir64 -x cl -cl-std=CL2.0 -O0 -emit-llvm-bc %s -o %t.bc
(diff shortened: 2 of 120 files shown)


mateuszchudyk:
Two actions failed on the test constant-loclist.ll. This PR doesn't touch this file so the problem seems to be on HEAD.

svenvh:
It would be good to hear @MrSidims 's opinion on this too, but I am not keen on making spirv-val a hard requirement for this many tests.  Ideally, if spirv-val is not available, the test would skip validation but should still run the other steps.

MrSidims:
I agree with Sven, we shouldn't skip that many tests.
Btw, I've just make a quick check on my machine with spirv-val non-installed and got the following result:
Total Discovered Tests: 877
  Unsupported      :  **74** (8.44%)
  Passed           : 799 (91.11%)
  Expectedly Failed:   4 (0.46%)
  
I interpret the result like this: there tests that require spirv-as/spirv-dis - and they are skipped. Meanwhile spirv-val tests are still being run as the number of Unsupported tests doesn't match the number of tests modified in this PR. So the question is: why do we need it?

mateuszchudyk:
Makes sense. So basically what we want is that if there's no spirv-val then we want to still run these test and skip just a spirv-val step, right? Now I see that in the [lit.cfg.py:L82](https://github.com/KhronosGroup/SPIRV-LLVM-Translator/blob/main/test/lit.cfg.py#L82) there's a code path which is most likely used to bypass this check. So to sum up I think we can drop this PR.

svenvh:
> the else statement is most likely used to bypass this check (I guess).

Yes exactly, the `else` turns any `spirv-val` into a no-op if spirv-val is not detected.

svenvh:
I think a comment to explain the optionality of spirv-val would be helpful, so I've opened https://github.com/KhronosGroup/SPIRV-LLVM-Translator/pull/2765

PR #2762, opened 2024-10-16
Add fast math flag translation for OpenCL std lib

MrSidims:
Such possibility was added in SPIR-V 1.6.
This patch also introduces limited translation of nofpclass LLVM parameter attribute.

### DIFF
--- a/lib/SPIRV/SPIRVWriter.cpp
+++ b/lib/SPIRV/SPIRVWriter.cpp
@@ -3038,7 +3038,8 @@ bool LLVMToSPIRVBase::transDecoration(Value *V, SPIRVValue *BV) {
     if (Opcode == Instruction::FAdd || Opcode == Instruction::FSub ||
         Opcode == Instruction::FMul || Opcode == Instruction::FDiv ||
         Opcode == Instruction::FRem ||
-        ((Opcode == Instruction::FNeg || Opcode == Instruction::FCmp) &&
+        ((Opcode == Instruction::FNeg || Opcode == Instruction::FCmp ||
+          BV->isExtInst()) &&
          BM->isAllowedToUseVersion(VersionNumber::SPIRV_1_6))) {
       FastMathFlags FMF = BVF->getFastMathFlags();
       SPIRVWord M{0};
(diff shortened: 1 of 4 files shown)


lib/SPIRV/SPIRVWriter.cpp
```diff
@@ -3038,7 +3038,8 @@ bool LLVMToSPIRVBase::transDecoration(Value *V, SPIRVValue *BV) {
     if (Opcode == Instruction::FAdd || Opcode == Instruction::FSub ||
         Opcode == Instruction::FMul || Opcode == Instruction::FDiv ||
         Opcode == Instruction::FRem ||
-        ((Opcode == Instruction::FNeg || Opcode == Instruction::FCmp) &&
+        ((Opcode == Instruction::FNeg || Opcode == Instruction::FCmp ||
```
bader, line 3041: @MrSidims, could you check if SPIRVReader handles fast-math flags for FNeg and FCmp correctly, please?
If not, please, file a tracker to follow-up.

Unfortunately, SPIR-V format is not able to translate LLVM semantics correctly. @bashbaug, SPIR-V doesn't allow applying fast-math flags to OpSelect and OpCall instructions, but LLVM IR can have fast math flags attached to `select` and `call` instructions. Do you know if this is going to be addressed by future SPIR-V versions?

lib/SPIRV/SPIRVWriter.cpp
```diff
@@ -3065,8 +3066,49 @@ bool LLVMToSPIRVBase::transDecoration(Value *V, SPIRVValue *BV) {
...
+          FPClassTest RetTest =
+              RetAttrs.getAttribute(Attribute::NoFPClass).getNoFPClass();
+          AttributeSet RetAttrs = FAttrs.getRetAttrs();
+          // Only Nan and Inf tests are representable in SPIR-V now.
+          bool ToAddNoNan = RetTest & fcNan;
+          bool ToAddNoInf = RetTest & fcInf;
+          if (ToAddNoNan || ToAddNoInf) {
+            const size_t NumParams = F->getFunctionType()->getNumParams();
+            for (size_t I = 0; I != NumParams; ++I) {
+              if (!F->hasParamAttribute(I, Attribute::NoFPClass)) {
```
bader, line 3089: There are extended instructions with non-FP parameters (e.g. ldexp).

lib/SPIRV/SPIRVWriter.cpp
MrSidims: > SPIR-V doesn't allow applying fast-math flags to OpSelect and OpCall instructions, but LLVM IR can have fast math flags attached to select and call instructions

This is a part of [SPV_KHR_float_controls2](https://htmlpreview.github.io/?https://github.com/KhronosGroup/SPIRV-Registry/blob/main/extensions/KHR/SPV_KHR_float_controls2.html):

> Replace the text following "Only valid on …​" with:
> All core instructions which use any floating-point type for either operands or result.
> OpExInst extended instructions, where expressly permitted by the extended instruction set in use.

It may be a bit too much and some instructions can or should be excluded from the list, but it is what it is.

lib/SPIRV/SPIRVWriter.cpp
MrSidims: Yes, there are builtins, where one can not apply the Decoration, above there is a comment from me:

> Ideally we should also check, if fast math flags are applicable to a particular function from OpenCL std lib according to environmental spec, but I beg your pardon for not doing that.

I can't imagine something better, then to check the names of the builtins in this case, and it will be not officiant and TBH might be even unnecessary (see below).

> non-FP parameters

AFAIU clang or any other LLVM IR producer won't put fast math flag and / or attribute in this case as LLVM wouldn't allow it, there are checks if an LLVM value is castable to FPMathOperator resulting in errors/assertions. So I'm not really expecting such IR be ever an input to the translator.

lib/SPIRV/SPIRVWriter.cpp
bashbaug: Given:

> OpExInst extended instructions, where expressly permitted by the extended instruction set in use.

Do we need some updates to the OpenCL Extended Instruction set, or perhaps the OpenCL SPIR-V environment spec?  Feel free to file a SPIR-V or OpenCL issue for tracking, if so, so this doesn't slip through the cracks.

We should also add some targeted CTS tests for this functionality, even if they aren't exhaustive.

lib/SPIRV/SPIRVWriter.cpp
bader: The logic of your code is "use FP Fast Math flag is **all** parameters" use the same `nofpclass` attribute". 
I suggest we change it to "use FP Fast Math flag is **all FP** parameters" use the same `nofpclass` attribute".

MrSidims: Got it, so the suggestion is to:
if a function is `float nofpclass(nan) foo(float nofpclass(nan), i32, i32)` - we still need to apply the `NotNaN` FP math mode.

bader: Right. 👍

lib/SPIRV/SPIRVWriter.cpp
MrSidims: > could you check if SPIRVReader handles fast-math flags for FNeg and FCmp correctly, please

@bader according to passing tests https://github.com/KhronosGroup/SPIRV-LLVM-Translator/blob/main/test/transcoding/fcmp.ll and https://github.com/KhronosGroup/SPIRV-LLVM-Translator/blob/main/test/transcoding/fneg.ll it handles correctly. Manual run also doesn't show anything fishy.

bader: Thanks!

MrSidims: > Do we need some updates to the OpenCL Extended Instruction set, or perhaps the OpenCL SPIR-V environment spec?

@bashbaug no, https://registry.khronos.org/SPIR-V/specs/unified1/OpenCL.ExtendedInstructionSet.100.pdf already contains the required lines like:
```
For environments that allow use of FPFastMathMode decorations on OpExtInst instructions,
FPFastMathMode decorations may be applied to the math instructions.
```
and there are other several references about it.
But I believe Alexey had another question related to `select` and `call` instructions.

> We should also add some targeted CTS tests for this functionality, even if they aren't exhaustive.

We should, https://github.com/KhronosGroup/OpenCL-CTS/pulls?q=is%3Apr+is%3Aopen+FPFastMathMode search shows nothing

MrSidims:
@svenvh please take a look at the PR when you have time

svenvh (approved):

PR #2766, opened 2024-10-18
[Backport to 19] Ignore UserSemantic decorations on BuiltIn variables

svenvh:
UserSemantic decorations on BuiltIn variables would be added into `llvm.global.annotations`.  During lowering of builtin variables to function calls, `replaceUsesOfBuiltinVar` would not know how to handle the annotation and crash.

Right now there is no clear use case for UserSemantic decorations on BuiltIn variables, so just ignore the decoration and drop it during translation from SPIR-V to LLVM.

Test-case-by: Ben Ashbaugh <ben.ashbaugh@intel.com>
(cherry picked from commit 0317fa82d24ad7b2f63e4e630f2372f0e80c5b78)


svenvh:
This might need https://github.com/KhronosGroup/SPIRV-LLVM-Translator/pull/2767 first.

MrSidims (approved):

Issue #2769, opened 2024-10-19
problems with OpCopyMemory


bashbaug:
Reverse translation (from SPIR-V to LLVM) does not appear to be working propery for `OpCopyMemory`.  Note that `OpCopyMemorySized` is working correctly, and there do not appear to be any tests for `OpCopyMemory`.

I am encountering two problems with the following tester (note: this is an except of a test for the SPIR-V 1.4 feature where `OpCopyMemor`y can accept two optional memory operands):

```
; SPIR-V
; Version: 1.4
               OpCapability Addresses
               OpCapability Kernel
               OpMemoryModel Physical64 OpenCL
               OpEntryPoint Kernel %kernel "copymemory_test"
       %uint = OpTypeInt 32 0
       %void = OpTypeVoid
   %gptr_int = OpTypePointer CrossWorkgroup %uint
   %pptr_int = OpTypePointer Function %uint
 %kernel_sig = OpTypeFunction %void %gptr_int
    %uint_42 = OpConstant %uint 42
     %kernel = OpFunction %void None %kernel_sig
        %dst = OpFunctionParameter %gptr_int
      %entry = OpLabel
     %pvalue = OpVariable %pptr_int Function %uint_42
               OpCopyMemory %dst %pvalue
               OpReturn
               OpFunctionEnd
```

1. The asserts are incorrect in `SPIRV::SPIRVCopyMemory::validate`, because they are checking `Id`, but this instruction does not set an `Id`.  This one is pretty easy to fix - something like:

    ```c++
    void validate() const override {
        assert(getValueType(Target)->isTypePointer() && "Invalid Target type");
        assert(getValueType(Source)->isTypePointer() && "Invalid Source type");
        assert(!(getValueType(Target)->getPointerElementType()->isTypeVoid()) &&
               "Invalid Target pointer type");
        assert(!(getValueType(Source)->getPointerElementType()->isTypeVoid()) &&
               "Invalid Target pointer type");
        assert(getValueType(Target)->getPointerElementType() ==
                   getValueType(Source)->getPointerElementType() &&
               "Mismatched Target and Source types");
        SPIRVInstruction::validate();
    }
    ```

[...]

PR #2770, opened 2024-10-21
Fix SPIRVCopyMemory::validate()

svenvh:
The asserts should be checking the `Target` member variable; `Id` is not used for this class.

Only fix the wrong asserts for now; proper handling and testing of `OpCopyMemory` will be done in a followup commit.

Contributes to https://github.com/KhronosGroup/SPIRV-LLVM-Translator/issues/2769


bashbaug (approved):
Thanks!

vmaksimo (approved):

MrSidims (approved):

PR #2793, opened 2024-10-23
[CI] Do not run check-in-tree after pushes to llvm_release_*

svenvh:
The check-in-tree tests take about 50 minutes to run compared to <5 minutes for check-out-of-tree.  This causes considerable congestion on the runners when backports get merged to several llvm_release_* branches around the same time.

Since the value of these post-push checks is relatively low, disable them; they will still run on pull requests.


MrSidims (approved):
makes total sense!

PR #2817, opened 2024-10-28
SPV_KHR_untyped_pointers - fix images translation

vmaksimo:
This patch ensures that we are loading image type, not untyped pointer type when the extension is used. This way we also preserve correct mangling during the reverse translation.

### DIFF
--- a/lib/SPIRV/SPIRVWriter.cpp
+++ b/lib/SPIRV/SPIRVWriter.cpp
@@ -2218,13 +2218,19 @@ LLVMToSPIRVBase::transValueWithoutDecoration(Value *V, SPIRVBasicBlock *BB,
                              MemoryAccessNoAliasINTELMaskMask);
     if (MemoryAccess.front() == 0)
       MemoryAccess.clear();
-    return mapValue(
-        V,
-        BM->addLoadInst(
-            transValue(LD->getPointerOperand(), BB), MemoryAccess, BB,
-            BM->isAllowedToUseExtension(ExtensionID::SPV_KHR_untyped_pointers)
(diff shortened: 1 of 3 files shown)


MrSidims:
Is such behavior unique for images or may happen for any SPIR-V opaque type?

vmaksimo:
> Is such behavior unique for images or may happen for any SPIR-V opaque type?

That's a good question. For now I haven't meet the same problem with other types, but I verified only the core SPIR-V functionality tests.
Which types do you think I should check?

vmaksimo:
> That's a good question. For now I haven't meet the same problem with other types, but I verified only the core SPIR-V functionality tests. Which types do you think I should check?

@MrSidims 
UPD: found that it'd be also used for device event type, and I guess for event type too.

vmaksimo:
> I just wonder if we want to fix them all together in this very PR or not. I can imagine the check you have for images `->isTypeImage()` being replaced with something like `isSPIRVOpaqueType`. 

Hm, it seems reasonable, and most likely will fix most problems with matrices and task sequence extensions. If it doesn't help fully that simple, I suggest this fix be added later together with those test updates. Will keep you posted

vmaksimo:
So, this more general check would fix 2 task sequence tests right away. Maybe the number is more but need to investigate every other test and fix the needed checks to understand. 
Does not seem reasonable to me to do this change in this very PR, but rather follow up with more affected tests (if any more)

MrSidims (approved):
Fine with me, but would like to hear from Sven

svenvh (approved):

PR #2818, opened 2024-10-29
Fix llvm.frexp intrinsic translation with SPV_KHR_untyped_pointers

vmaksimo:


lib/SPIRV/SPIRVWriter.cpp
```diff
@@ -2243,7 +2243,17 @@ LLVMToSPIRVBase::transValueWithoutDecoration(Value *V, SPIRVBasicBlock *BB,
           auto *FrexpResult = transValue(RV, BB);
           SPIRVValue *IntFromFrexpResult =
               static_cast<SPIRVExtInst *>(FrexpResult)->getArgValues()[1];
-          IntFromFrexpResult = BM->addLoadInst(IntFromFrexpResult, {}, BB);
+
+          if (BM->isAllowedToUseExtension(
```
MrSidims, line 2247: May be checks for the extension are redundant in these patches. If IntFromFrexpResult is untyped variable, then the extension is already enabled, right?

MrSidims (approved):
LGTM, just a nit

lib/SPIRV/SPIRVWriter.cpp
vmaksimo: I believe you're right, will update

PR #2819, opened 2024-10-29
Fix OpGroupAsyncCopy used with SPV_KHR_untyped_pointers

vmaksimo:


lib/SPIRV/SPIRVWriter.cpp
```diff
@@ -755,7 +755,8 @@ SPIRVType *LLVMToSPIRVBase::transPointerType(SPIRVType *ET, unsigned AddrSpc) {
     return transPointerType(ET, SPIRAS_Private);
   if (BM->isAllowedToUseExtension(ExtensionID::SPV_KHR_untyped_pointers) &&
       !(ET->isTypeArray() || ET->isTypeVector() || ET->isTypeStruct() ||
-        ET->isTypeImage() || ET->isTypeSampler() || ET->isTypePipe())) {
+        ET->isTypeImage() || ET->isTypeSampler() || ET->isTypePipe() ||
```
MrSidims, line 758: Guess we should merge https://github.com/KhronosGroup/SPIRV-LLVM-Translator/pull/2820 first, right?

MrSidims (changes requested):
After offline discussion: we have to re-evaluate this patch. Reasons:
 - OpGroupAsyncCopy  doesn't work well with opaque pointers in general;
 - regardless how OpGroupAsyncCopy  would work - insertion of a bitcast is incorrect in this case

PR #2825, opened 2024-11-02
Generate memcpy for OpCopyLogical

### LINKED ISSUE #2768: handling OpCopyLogical
Possibly related to #2460 and #2484.

I've been adding SPIR-V 1.4 CTS testing and testing for `OpCopyLogical` specifically and I'd like to confirm how this instruction is supposed to behave.  If helpful, my WIP tests are here:

https://github.com/KhronosGroup/OpenCL-CTS/compare/main...bashbaug:OpenCL-CTS:spirv-14-part5

The SPIR-V file I am using for this test is:

```
; SPIR-V
; Version: 1.4
               OpCapability Addresses
               OpCapability Kernel
               OpEntryPoint Kernel %kernel "copylogical_test"
       %uint = OpTypeInt 32 0
      %float = OpTypeFloat 32
       %void = OpTypeVoid
   %struct_a = OpTypeStruct %uint %float
%ptr_struct_a = OpTypePointer CrossWorkgroup %struct_a
   %struct_b = OpTypeStruct %uint %float
%ptr_struct_b = OpTypePointer CrossWorkgroup %struct_b
 %kernel_sig = OpTypeFunction %void %ptr_struct_b
  %uint_1024 = OpConstant %uint 1024
   %float_pi = OpConstant %float 3.1415
%struct_a_src = OpConstantComposite %struct_a %uint_1024 %float_pi
     %kernel = OpFunction %void None %kernel_sig
        %dst = OpFunctionParameter %ptr_struct_b
      %entry = OpLabel
%struct_b_dst = OpCopyLogical %struct_b %struct_a_src
               OpStore %dst %struct_b_dst
               OpReturn
               OpFunctionEnd
```

It essentially copies a structure with an integer and a float, then stores it to memory.  This was inspired by a similar existing test that uses `OpCopyObject` instead of `OpCopyLogical`.

After SPIR-V to LLVM translation, this case generates:

```llvm
; ModuleID = '/home/bashbaug/git/OpenCL-CTS/test_conformance/spirv_new/spirv_bin/spv1.4/copylogical_struct.bc'
target datalayout = "e-p:64:64:64-i1:8:8-i8:8:8-i16:16:16-i32:32:32-i64:64:64-f32:32:32-f64:64:64-v16:16:16-v24:32:32-v32:32:32-v48:64:64-v64:64:64-v96:128:128-v128:128:128-v192:256:256-v256:256:256-v512:512:512-v1024:1024:1024-G1"

%structtype.0 = type { i32, float }
%structtype = type { i32, float }

; Function Attrs: nounwind
define spir_kernel void @copylogical_test(ptr addrspace(1) %0) #0 !kernel_arg_addr_space !5 !kernel_arg_access_qual !6 !kernel_arg_type !7 !kernel_arg_type_qual !8 !kernel_arg_base_type !7 !spirv.ParameterDecorations !9 {
  call spir_func void @_Z19__spirv_CopyLogical12structtype.0(ptr addrspace(1) sret(%structtype) %0, %structtype.0 { i32 1024, float 0x400921CAC0000000 })
  ret void
}

declare spir_func void @_Z19__spirv_CopyLogical12structtype.0(ptr addrspace(1) sret(%structtype), %structtype.0)
```

Is this the expected behavior?  Should SPIR-V consumers need to provide an implementation of `__spirv_CopyLogical`?  Or, could/should we translate this to something like `llvm.memcpy` instead?

For reference, a similar test using `OpCopyObject` instead of `OpCopyLogical` generates:

```llvm
; ModuleID = '/home/bashbaug/git/OpenCL-CTS/test_conformance/spirv_new/spirv_bin/spv1.4/copyobject_struct.bc'
target datalayout = "e-p:64:64:64-i1:8:8-i8:8:8-i16:16:16-i32:32:32-i64:64:64-f32:32:32-f64:64:64-v16:16:16-v24:32:32-v32:32:32-v48:64:64-v64:64:64-v96:128:128-v128:128:128-v192:256:256-v256:256:256-v512:512:512-v1024:1024:1024"

%structtype = type { i32, float }

; Function Attrs: nounwind
define spir_kernel void @copylogical_test(ptr addrspace(1) %0) #0 !kernel_arg_addr_space !5 !kernel_arg_access_qual !6 !kernel_arg_type !7 !kernel_arg_type_qual !8 !kernel_arg_base_type !7 !spirv.ParameterDecorations !9 {
  %2 = alloca %structtype, align 8
  store %structtype { i32 1024, float 0x400921CAC0000000 }, ptr %2, align 4
  %3 = load %structtype, ptr %2, align 4
  store %structtype %3, ptr addrspace(1) %0, align 4
  ret void
}
```

bashbaug:
fixes #2768 

Generate an LLVM memcpy for OpCopyLogical, rather than a call to an OpCopyLogical function.

### DIFF
--- a/lib/SPIRV/SPIRVReader.cpp
+++ b/lib/SPIRV/SPIRVReader.cpp
@@ -2109,7 +2109,23 @@ Value *SPIRVToLLVM::transValueWithoutDecoration(SPIRVValue *BV, Function *F,
   }
   case OpCopyLogical: {
     SPIRVCopyLogical *CL = static_cast<SPIRVCopyLogical *>(BV);
-    return mapValue(BV, transSPIRVBuiltinFromInst(CL, BB));
+
+    auto *SrcTy = transType(CL->getOperand()->getType());
+    auto *SrcAI = new AllocaInst(SrcTy, 0, "", BB);
+    new StoreInst(transValue(CL->getOperand(), F, BB), SrcAI, BB);
+
+    auto *DstTy = transType(CL->getType());
+    auto *DstAI = new AllocaInst(DstTy, 0, "", BB);
+
+    IRBuilder<> Builder(BB);
+    uint64_t Size = M->getDataLayout().getTypeStoreSize(SrcTy).getFixedValue();
+    assert(Size == M->getDataLayout().getTypeStoreSize(DstTy).getFixedValue() &&
+           "Size mismatch in OpCopyLogical");
+    Builder.CreateMemCpy(DstAI, DstAI->getAlign(), SrcAI, SrcAI->getAlign(),
+                         Size);
+
+    auto *LI = new LoadInst(DstTy, DstAI, "", BB);
+    return mapValue(BV, LI);
   }
 
   case OpAccessChain:
--- a/lib/SPIRV/SPIRVReader.cpp
+++ b/lib/SPIRV/SPIRVReader.cpp
@@ -2991,7 +2991,8 @@ void SPIRVToLLVM::transFunctionAttrs(SPIRVFunction *BF, Function *F) {
     auto *BA = BF->getArgument(I->getArgNo());
     mapValue(BA, &(*I));
     setName(&(*I), BA);
-    AttributeMask IllegalAttrs = AttributeFuncs::typeIncompatible(I->getType());
+    AttributeMask IllegalAttrs =
+        AttributeFuncs::typeIncompatible(I->getType(), I->getAttributes());
     BA->foreachAttr([&](SPIRVFuncParamAttrKind Kind) {
       // Skip this function parameter attribute as it will translated among
       // OpenCL metadata
--- a/test/OpCopyLogical.spvasm
+++ b/test/OpCopyLogical.spvasm
@@ -22,4 +22,6 @@
                OpReturn
                OpFunctionEnd
 
-; CHECK-LLVM: @_Z19__spirv_CopyLogical12structtype.0(ptr sret(%structtype) %[[#]], %structtype.0 zeroinitializer)
+; CHECK-LLVM: [[SRC_ALLOCA:%[a-z0-9.]+]] = alloca [[SRC_TYPE:%[a-z0-9.]+]], align 8
+; CHECK-LLVM: [[DST_ALLOCA:%[a-z0-9.]+]] = alloca [[DST_TYPE:%[a-z0-9.]+]], align 8
+; CHECK-LLVM: call void @llvm.memcpy.p0.p0.i64(ptr align 8 [[DST_ALLOCA]], ptr align 8 [[SRC_ALLOCA]], i64 0, i1 false)


lib/SPIRV/SPIRVReader.cpp
```diff
@@ -2109,7 +2109,23 @@ Value *SPIRVToLLVM::transValueWithoutDecoration(SPIRVValue *BV, Function *F,
   }
   case OpCopyLogical: {
     SPIRVCopyLogical *CL = static_cast<SPIRVCopyLogical *>(BV);
-    return mapValue(BV, transSPIRVBuiltinFromInst(CL, BB));
+
+    auto *SrcTy = transType(CL->getOperand()->getType());
+    auto *SrcAI = new AllocaInst(SrcTy, 0, "", BB);
```
svenvh, line 2114: Since you have an IRBuilder below, you might as well hoist that up and use its `CreateAlloca`, `CreateStore` and `CreateLoad` member functions; that will save you from passing the `"", BB` arguments all the time.

lib/SPIRV/SPIRVReader.cpp
```diff
@@ -2109,7 +2109,23 @@ Value *SPIRVToLLVM::transValueWithoutDecoration(SPIRVValue *BV, Function *F,
...
+    new StoreInst(transValue(CL->getOperand(), F, BB), SrcAI, BB);
```
svenvh, line 2115: Perhaps add a comment explaining why we need a store first, instead of using the result of `transValue(CL->getOperand(), F, BB)` directly?

lib/SPIRV/SPIRVReader.cpp
```diff
@@ -2991,7 +3007,8 @@ void SPIRVToLLVM::transFunctionAttrs(SPIRVFunction *BF, Function *F) {
     auto *BA = BF->getArgument(I->getArgNo());
     mapValue(BA, &(*I));
     setName(&(*I), BA);
-    AttributeMask IllegalAttrs = AttributeFuncs::typeIncompatible(I->getType());
+    AttributeMask IllegalAttrs =
+        AttributeFuncs::typeIncompatible(I->getType(), I->getAttributes());
```
vmaksimo, line 3011: Maybe I'm missing something, but is it really related to the OpCopyLogical translation?

lib/SPIRV/SPIRVReader.cpp
bashbaug: Done!

lib/SPIRV/SPIRVReader.cpp
bashbaug: Do you have a specific suggestion?  I'm open to improving this.  I'm not aware of other viable options, though...

vmaksimo (approved):

lib/SPIRV/SPIRVReader.cpp
svenvh: To me it feels like there is some redundancy when doing a `store` and then a `memcpy`.  Thinking about this again, could we not avoid the `memcpy` altogether, and pretty much follow `OpCopyObject` above?  The only difference would be that the type on the `load` is different from the type on the `store`; we have both types already available (as `SrcTy` and `DstTy` in your current patch).

bashbaug: Interesting - I thought it would have been a problem to store as one type and load as a different type, but it seemed to work just fine.  Have a look at the latest version - much simpler now!

(I'll edit the PR description also, since there is no longer a memcpy involved.)

svenvh (approved):

lib/SPIRV/SPIRVReader.cpp
svenvh: LGTM, thanks!

PR #2826, opened 2024-11-05
SPIRVReader: Support OpConstantComposite for cooperative matrix

svenvh:
This was hitting a "not implemented UNREACHABLE".  Like the other cooperative matrix operations, map this construct to a SPIR-V friendly IR function call.

Let `transDbgInfo` skip over `OpConstantComposite` because we're mapping `OpConstantComposite` to an LLVM `Instruction` without having a corresponding `SPIRVInstruction`.


MrSidims (approved):
Originally this PR brought a lot of confusion to me, as I though: "then how do we handle matrix creation and why I have seen no regressions, when was working on cooperative matrix switch in SYCL?". But apparently we are using OpCompositeConstruct for this, and OpConstantComposite was added to represent OpMatrixTimesScalar functionality.

PR #2853, opened 2024-11-11
[SPIR-V 1.1] SPIRVReader: Add OpSizeOf support

svenvh:
The `OpSizeOf` instruction was added in SPIR-V 1.1, but not supported yet.


test/OpSizeOf.spvasm
```diff
@@ -0,0 +1,33 @@
+; REQUIRES: spirv-as
+; RUN: spirv-as --target-env spv1.1 -o %t.spv %s
+; TODO: reenable spirv-val when OpSizeOf is supported.
+; R/U/N: spirv-val %t.spv
```
vmaksimo, line 4: Is there already an issue in SPIRV-Tools?

vmaksimo (commented):
LGTM
Do we some day expect to have forward translation as well?

test/OpSizeOf.spvasm
svenvh: Yes, spirv-val doesn't know much about `OpSizeOf` yet.  I'm preparing a PR to fix that.

svenvh:
> Do we some day expect to have forward translation as well?

Some day hopefully.  For now I'm focusing on the consumption side only.

test/OpSizeOf.spvasm
vmaksimo: Just an issue reported to SPIR-V Tolls would have been good enough for now, but that's great your fix is already in progress :)

vmaksimo (approved):

MrSidims (approved):

test/OpSizeOf.spvasm
svenvh: For reference, the fix is here: https://github.com/KhronosGroup/SPIRV-Tools/pull/5879
(but even if that lands soon, we'll need to wait for the next SPIRV-Tools release before we can re-enable this check in CI.)

PR #2857, opened 2024-11-13
Improve matrices translation with SPV_KHR_untyped_pointers

vmaksimo:
Ensure that we do correct translation of matrices (`SPV_KHR_cooperative_matrix` and `SPV_INTEL_joint_matrix` extensions) when untyped pointers are enabled.
This mainly fixes mangling issues in reverse translation for the untyped pointer. 
Also added handling for typed and untyped SPIR-V friendly access chain instructions in forward translation (a point to review and discuss).

### DIFF
--- a/lib/SPIRV/SPIRVReader.cpp
+++ b/lib/SPIRV/SPIRVReader.cpp
@@ -2216,8 +2216,10 @@ Value *SPIRVToLLVM::transValueWithoutDecoration(SPIRVValue *BV, Function *F,
     auto *AC = static_cast<SPIRVAccessChainBase *>(BV);
     auto *Base = transValue(AC->getBase(), F, BB);
     SPIRVType *BaseSPVTy = AC->getBaseType();
-    if (BaseSPVTy->isTypePointer() &&
-        BaseSPVTy->getPointerElementType()->isTypeCooperativeMatrixKHR()) {
+    if ((BaseSPVTy->isTypePointer() &&
+         BaseSPVTy->getPointerElementType()->isTypeCooperativeMatrixKHR()) ||
+        (isUntypedAccessChainOpCode(OC) &&
+         BaseSPVTy->isTypeCooperativeMatrixKHR())) {
       return mapValue(BV, transSPIRVBuiltinFromInst(AC, BB));
     }
     Type *BaseTy =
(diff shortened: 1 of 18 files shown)


test/extensions/KHR/SPV_KHR_cooperative_matrix/access_store_untyped.ll
```diff
@@ -0,0 +1,66 @@
...
+; CHECK-SPIRV: TypePointer [[#TypeMatrixPtr:]] 7 [[#TypeMatrix]]
+; CHECK-SPIRV: TypeUntypedPointerKHR [[#TypePtr:]] 7
+
+; CHECK-SPIRV: Variable [[#TypeMatrixPtr]] [[#VarMatrixPtr:]] 7
+; CHECK-SPIRV: CompositeConstruct [[#TypeMatrix]] [[#Composite:]] [[#Const0]]
+; CHECK-SPIRV: Store [[#VarMatrixPtr]] [[#Composite]]
+; CHECK-SPIRV: UntypedAccessChainKHR [[#TypePtr]] [[#Res:]] [[#TypeMatrix]] [[#VarMatrixPtr]] [[#Const1]]
+; CHECK-SPIRV: Store [[#Res]] [[#Const42]]
+
+; CHECK-LLVM: %0 = alloca target("spirv.CooperativeMatrixKHR", i32, 3, 12, 12, 0)
```
MrSidims, line 27:
```suggestion
; CHECK-LLVM: %[[#Matrix:]] = alloca target("spirv.CooperativeMatrixKHR", i32, 3, 12, 12, 0)
```
and so on

lib/SPIRV/SPIRVReader.cpp
```diff
@@ -3426,23 +3428,65 @@ Instruction *SPIRVToLLVM::transBuiltinFromInst(const std::string &FuncName,
...
+        // Pointer could be a function parameter. Assume that the type of the
+        // pointer is the same as the return type.
+        Type *Ty = nullptr;
+        // it return type is array type, assign its element type to Ty
+        if (RetTy->isArrayTy())
+          Ty = RetTy->getArrayElementType();
+        else if (RetTy->isVectorTy())
+          Ty = cast<VectorType>(RetTy)->getElementType();
+        else
+          Ty = RetTy;
```
MrSidims, line 3483: So for store pointer parameter would be considered as `void *`? If so it doesn't feel right, can we instead query type of the value operand of the store?

test/extensions/KHR/SPV_KHR_cooperative_matrix/access_store_untyped.ll
vmaksimo: It's basically just a copy of `access_store.ll` test :) do we need these checks there too?

lib/SPIRV/SPIRVReader.cpp
vmaksimo: Sorry, I don't get this at all. Don't we already query the type from the llvm instructions? Or do you mean specific SPIR-V opcode in mind?

test/extensions/KHR/SPV_KHR_cooperative_matrix/access_store_untyped.ll
MrSidims: May be in another PR, not here :)

MrSidims (approved):

PR #2858, opened 2024-11-13
Fix reverse translation of `OpPtrDiff` with untyped pointers

vmaksimo:
We should not lose operand type when `SPV_KHR_untyped_pointers` is used.

### DIFF
--- a/lib/SPIRV/SPIRVReader.cpp
+++ b/lib/SPIRV/SPIRVReader.cpp
@@ -2284,8 +2284,17 @@ Value *SPIRVToLLVM::transValueWithoutDecoration(SPIRVValue *BV, Function *F,
     auto *BC = static_cast<SPIRVBinary *>(BV);
     auto Ops = transValue(BC->getOperands(), F, BB);
     IRBuilder<> Builder(BB);
-    Type *ElemTy =
-        transType(BC->getOperands()[0]->getType()->getPointerElementType());
+    Value *Operand1 = Ops[0]->stripPointerCasts();
+
+    Type *ElemTy = nullptr;
+    // Should we better get the type info from alloca instruction or from the
+    // OpUntypedVariableKHR DataType field?
+    if (auto *AI = dyn_cast<AllocaInst>(Operand1))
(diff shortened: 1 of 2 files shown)


lib/SPIRV/SPIRVReader.cpp
```diff
@@ -2284,8 +2284,17 @@ Value *SPIRVToLLVM::transValueWithoutDecoration(SPIRVValue *BV, Function *F,
...
     auto Ops = transValue(BC->getOperands(), F, BB);
     IRBuilder<> Builder(BB);
-    Type *ElemTy =
-        transType(BC->getOperands()[0]->getType()->getPointerElementType());
+    Value *Operand1 = Ops[0]->stripPointerCasts();
+
+    Type *ElemTy = nullptr;
+    // Should we better get the type info from alloca instruction or from the
+    // OpUntypedVariableKHR DataType field?
+    if (auto *AI = dyn_cast<AllocaInst>(Operand1))
```
MrSidims, line 2292: What happens if we remove this `if`, leaving only `else`?

MrSidims: May be resolving this can go to a different patch. This patch already improves the code base.

MrSidims (approved):

lib/SPIRV/SPIRVReader.cpp
```diff
@@ -2284,8 +2284,17 @@ Value *SPIRVToLLVM::transValueWithoutDecoration(SPIRVValue *BV, Function *F,
     auto *BC = static_cast<SPIRVBinary *>(BV);
     auto Ops = transValue(BC->getOperands(), F, BB);
     IRBuilder<> Builder(BB);
-    Type *ElemTy =
-        transType(BC->getOperands()[0]->getType()->getPointerElementType());
+    Value *Operand1 = Ops[0]->stripPointerCasts();
+
+    Type *ElemTy = nullptr;
+    // Should we better get the type info from alloca instruction or from the
```
MrSidims, line 2290: Both is fine IMO

svenvh (approved):

PR #2860, opened 2024-11-14
Make LLVMToSPIRVBase comply with Rule of Three.

maarquitos14:
`LLVMToSPIRVBase` had a custom destructor, but no copy constructor and no copy assignment operator, so it was not complying with the Rule of Three. Explicitly add the two latter as deleted to comply.

### DIFF
--- a/lib/SPIRV/SPIRVWriter.h
+++ b/lib/SPIRV/SPIRVWriter.h
@@ -72,6 +72,8 @@ namespace SPIRV {
 class LLVMToSPIRVBase : protected BuiltinCallHelper {
 public:
   LLVMToSPIRVBase(SPIRVModule *SMod);
+  LLVMToSPIRVBase(const LLVMToSPIRVBase &other) = delete;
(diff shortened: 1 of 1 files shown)


svenvh:
Thanks!  Would you mind extending this patch to also comply with the rule of five, by including the move constructor & move assignment operator?

maarquitos14:
> Thanks! Would you mind extending this patch to also comply with the rule of five, by including the move constructor & move assignment operator?

Of course, done now :)

svenvh (approved):

MrSidims (approved):

MrSidims:
@maarquitos14 should we also rename the PR with the description? They are usually used as commit name and message appropriately when PR is merged with 'squash and merge' strategy.

maarquitos14:
> @maarquitos14 should we also rename the PR with the description? They are usually used as commit name and message appropriately when PR is merged with 'squash and merge' strategy.

Good catch, thanks. Done.

PR #2868, opened 2024-11-19
Replace llvm.fmuladd.v1f* with llvm.fmuladd.f* since OpenCL doesn't support length-1 vector.

haonanya1:


haonanya1:
@svenvh , @MrSidims, can you please take a look? Thanks.

test/llvm-intrinsics/fmuladdv1f.ll
```diff
@@ -0,0 +1,109 @@
+; RUN: llvm-as < %s -o %t.bc
+; RUN: llvm-spirv %t.bc --spirv-ext=+SPV_INTEL_vector_compute -o %t.spv
```
MrSidims, line 2: This extension is not standard and we shouldn't rely on it. Meanwhile you have to use it as there are <1 x ...> vectors in the module.

MrSidims (commented):
I won't resist to merge the patch, yet my preference to fix this earlier in the compilation stack by replacing all <1 x ...> vectors in the module. My opinion that we better not to do such replacements in the translator. The reason is that we don't know the environment for which LLVM IR was compiled and I can imagine an unlikely yet possible situation, when LLVM IR compiled for RISC-V architecture comes as an input to the translator and AFAIK <1 x ...> vectors in LLVM IR are used to represent scalable vectors RISC-V extension. So replacing them with scalars in the translator would be incorrect. Opposite argument would be: for this RISC-V folks should add a new extension in SPIR-V, so until it's enabled we could replace vectors with scalars, but IMO it's still better not to do so and continue to emit an error.

haonanya1:
Thanks for quick reply, @MrSidims, the IR is valid, the number of elements for vector type is a constant integer value larger than 0 in llvm ir https://llvm.org/docs/LangRef.html#vector-type. Component Count  of spirv OpTypeVector must be at least 2 https://registry.khronos.org/SPIR-V/specs/unified1/SPIRV.html#OpTypeVector. There are gaps for IR and spirv. @bashbaug , what do you think? Thanks very much.

svenvh:
> I'm a bit afraid to break existing or future flows as <1 x ...> can be also a scalable vector

Aren't scalable vectors represented using `<vscale x ...>`?

I don't have strong objections against doing a vec1 -> scalar replacement in the translator.  However I would suggest to implement that as a separate pass, and not only for the `fmuladd` intrinsic, but for all instructions in the module with `<1 x ...>` vector types.

PR #2869, opened 2024-11-19
[SPIR-V 1.2] SPIRVReader: Add AlignmentId support

svenvh:
If there is no `OpDecorate .. Alignment` in the input, see if there is an `OpDecorateId .. AlignmentId` and take the alignment from the referenced constant instead.

Once `AlignmentId` has been translated to LLVM IR, it is indistinguishable from an (non-ID) `Alignment` decoration.

### DIFF
--- a/lib/SPIRV/SPIRVReader.cpp
+++ b/lib/SPIRV/SPIRVReader.cpp
@@ -288,6 +288,29 @@ Value *SPIRVToLLVM::mapFunction(SPIRVFunction *BF, Function *F) {
   return F;
 }
 
+std::optional<uint64_t> SPIRVToLLVM::transIdAsConstant(SPIRVId Id) {
+  auto *V = BM->get<SPIRVValue>(Id);
+  const auto *ConstValue =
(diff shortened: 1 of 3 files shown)


MrSidims (approved):

lib/SPIRV/libSPIRV/SPIRVValue.cpp
```diff
@@ -64,7 +64,16 @@ void SPIRVValue::setAlignment(SPIRVWord A) {
...
 
 bool SPIRVValue::hasAlignment(SPIRVWord *Result) const {
-  return hasDecorate(DecorationAlignment, 0, Result);
+  if (bool HasDecorate = hasDecorate(DecorationAlignment, 0, Result))
+    return HasDecorate;
+
+  SPIRVId AlignId;
+  bool HasDecorateId = hasDecorateId(DecorationAlignmentId, 0, &AlignId);
+  if (HasDecorateId) {
+    SPIRVConstant *Cst = Module->get<SPIRVConstant>(AlignId);
```
svenvh, line 73: This doesn't work with specialization constants, I'll upload a new version of this patch.

PR #2876, opened 2024-11-25
[DebugInfo] Fix EnumClass support

bwlodarcz:
This commit addresses issues with EnumClass handling in DebugEnumType. 
In reverse translation, a bug caused the EnumClass flag to be incorrectly applied to every DebugEnumType with an UnderlyingType.
Additionally, in forward translation, the support for EnumClassFlag was missing entirely.


MrSidims (commented):
Changes are LGTM, can we have a LIT for this?

bwlodarcz:
@MrSidims We do have such test in `test/DebugInfo/X86/enum-class.ll`.

MrSidims (approved):

PR #2935, opened 2024-12-16
Limit NO_INSTALL_RPATH to out-of-tree builds.

### LINKED ISSUE #2906: NO_INSTALL_RPATH breaks in-tree builds with shared libs
The README states:

> LLVM in-tree build
> 
> The translator can be built as a regular LLVM subproject. To do that you need to clone it into the llvm/projects or llvm/tools directory.
> 
>     git clone https://github.com/llvm/llvm-project.git
>     cd llvm-project/llvm/projects
>     git clone https://github.com/KhronosGroup/SPIRV-LLVM-Translator.git
> 
> Run (or re-run) cmake as usual for LLVM. After that you should have llvm-spirv and check-llvm-spirv targets available.

Running CMake as usual for LLVM, for me, means including `-DLLVM_BUILD_LLVM_DYLIB=ON -DLLVM_LINK_LLVM_DYLIB=ON`.

Following these instructions, and installing LLVM, results in:

```console
harald@noble:~/llvm-project/main/install/aarch64-linux$ bin/llvm-spirv --help
bin/llvm-spirv: error while loading shared libraries: libLLVM.so.20.0git: cannot open shared object file: No such file or directory
```

This is because of https://github.com/KhronosGroup/SPIRV-LLVM-Translator/blob/c1828a93d050afe8a44cc239a2908270c578c345/tools/llvm-spirv/CMakeLists.txt#L19-L20 setting `NO_INSTALL_RPATH`. But in this LLVM configuration, setting the rpath is required to make things work, and if I comment out the `NO_INSTALL_RPATH`, things do work.

Would it be possible to only set `NO_INSTALL_RPATH` in those cases where it is necessary (I do not know which cases), or if that is not possible, to provide an option so that users can choose whether or not to set the rpath depending on what is best for their situation?

hvdijk:
For in-tree builds with shared libraries, setting the rpath not only works, but is required.

Fixes #2906

### DIFF
--- a/tools/llvm-spirv/CMakeLists.txt
+++ b/tools/llvm-spirv/CMakeLists.txt
@@ -14,10 +14,18 @@ if(SPIRV_BACKEND_FOUND)
   list(APPEND LLVM_LINK_COMPONENTS "SPIRVCodeGen")
 endif()
 
+# llvm_setup_rpath messes with the rpath making llvm-spirv not
+# executable from the build directory in out-of-tree builds
+set(add_llvm_tool_options)
+if(LLVM_SPIRV_BUILD_EXTERNAL)
+  set(add_llvm_tool_options
+    NO_INSTALL_RPATH
+  )
+endif()
+
 add_llvm_tool(llvm-spirv
   llvm-spirv.cpp
-  # llvm_setup_rpath messes with the rpath making llvm-spirv not executable from the build directory
-  NO_INSTALL_RPATH
+  ${add_llvm_tool_options}
 )
 
 setup_host_tool(llvm-spirv LLVM_SPIRV llvm-spirv_exe llvm-spirv_target)


hvdijk:
The formatting may raise some eyebrows, but this is intentional. To work around #2906, currently, we comment out `NO_INSTALL_RPATH` by prepending `# `. This PR keeps `NO_INSTALL_RPATH` on a line of its own to allow us to continue to use that workaround until we no longer need to worry about older branches. Alternatively, if this (plus any edits as a result of review comments) could also be applied to the branches, we could just delete our workaround entirely.

svenvh:
> Alternatively, if this (plus any edits as a result of review comments) could also be applied to the branches, we could just delete our workaround entirely.

Backporting the fix should be fine; which of the older branches are you interested in?

hvdijk:
> Backporting the fix should be fine; which of the older branches are you interested in?

We use the `main`, `llvm_release_190`, and `llvm_release_180` branches, thank you.

MrSidims (approved):
Thanks!

svenvh (approved):

PR #2944, opened 2024-12-24
Translate nonsemantic attribute and metadata of GlobalVariable

wenju-he:
Motivations is similar as f729c49. This PR addresses SYCL device global which may have attributes "sycl-device-image-scope", "sycl-host-access" and "sycl-unique-id". Failure to preserve "sycl-unique-id" after llvm-spirv translation triggers assert at
https://github.com/intel/llvm/blob/2824f61dd36790448a224cd596985bd01cbcd0f3/llvm/lib/SYCLLowerIR/DeviceGlobals.cpp#L85

Also preserve GlobalVariable metadata as an improvement, though there is no test to show this is really needed.

### DIFF
--- a/lib/SPIRV/SPIRVReader.cpp
+++ b/lib/SPIRV/SPIRVReader.cpp
@@ -5141,36 +5141,53 @@ void SPIRVToLLVM::transAuxDataInst(SPIRVExtInst *BC) {
     return;
   auto Args = BC->getArguments();
   // Args 0 and 1 are common between attributes and metadata.
-  // 0 is the function, 1 is the name of the attribute/metadata as a string
-  auto *SpvFcn = BC->getModule()->getValue(Args[0]);
-  auto *F = static_cast<Function *>(getTranslatedValue(SpvFcn));
-  assert(F && "Function should already have been translated!");
+  // 0 is the global object, 1 is the name of the attribute/metadata as a string
+  auto *Arg0 = BC->getModule()->getValue(Args[0]);
+  auto *GO = cast<GlobalObject>(getTranslatedValue(Arg0));
+  auto *F = dyn_cast_or_null<Function>(GO);
+  auto *GV = dyn_cast_or_null<GlobalVariable>(GO);
+  assert((F || GV) && "Value should already have been translated!");
   auto AttrOrMDName = BC->getModule()->get<SPIRVString>(Args[1])->getStr();
   switch (BC->getExtOp()) {
-  case NonSemanticAuxData::FunctionAttribute: {
+  case NonSemanticAuxData::FunctionAttribute:
+  case NonSemanticAuxData::GlobalVariableAttribute: {
     assert(Args.size() < 4 && "Unexpected FunctionAttribute Args");
     // If this attr was specially handled and added elsewhere, skip it.
     Attribute::AttrKind AsKind = Attribute::getAttrKindFromName(AttrOrMDName);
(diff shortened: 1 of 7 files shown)


lib/SPIRV/SPIRVReader.cpp
```diff
@@ -5141,36 +5141,53 @@ void SPIRVToLLVM::transAuxDataInst(SPIRVExtInst *BC) {
...
   // Args 0 and 1 are common between attributes and metadata.
-  // 0 is the function, 1 is the name of the attribute/metadata as a string
-  auto *SpvFcn = BC->getModule()->getValue(Args[0]);
-  auto *F = static_cast<Function *>(getTranslatedValue(SpvFcn));
-  assert(F && "Function should already have been translated!");
+  // 0 is the global object, 1 is the name of the attribute/metadata as a string
+  auto *Arg0 = BC->getModule()->getValue(Args[0]);
+  auto *GO = cast<GlobalObject>(getTranslatedValue(Arg0));
+  auto *F = dyn_cast_or_null<Function>(GO);
+  auto *GV = dyn_cast_or_null<GlobalVariable>(GO);
```
MrSidims, line 5148: This makes me a little bit suspicious, why do we need dyn_cast**_or_null** here when `GO` is a result of a `cast`? May be it's leftover from a previous patch iteration, where some test was failing.

lib/SPIRV/SPIRVWriter.cpp
```diff
@@ -1230,23 +1230,27 @@ void LLVMToSPIRVBase::transFunctionMetadataAsUserSemanticDecoration(
   }
 }
 
-void LLVMToSPIRVBase::transAuxDataInst(SPIRVFunction *BF, Function *F) {
-  auto *BM = BF->getModule();
+void LLVMToSPIRVBase::transAuxDataInst(SPIRVValue *BV, Value *V) {
+  auto *GO = cast<GlobalObject>(V);
+  auto *F = dyn_cast_or_null<Function>(GO);
```
MrSidims, line 1235: Same question about **_or_null**

lib/SPIRV/SPIRVReader.cpp
wenju-he: done, yet it is a misuse.

lib/SPIRV/SPIRVWriter.cpp
wenju-he: done

MrSidims (approved):

lib/SPIRV/libSPIRV/NonSemantic.AuxData.h
```diff
@@ -28,6 +28,8 @@ namespace NonSemanticAuxData {
 enum Instruction {
   FunctionMetadata = 0,
   FunctionAttribute = 1,
-  PreserveCount = 2
+  GlobalVariableMetadata = 2,
+  GlobalVariableAttribute = 3,
+  PreserveCount = 4
```
VyacheslavLevytskyy, line 33: If `PreserveCount` is (and has been previously) unused it should be probably deleted entirely. Otherwise, this line introduces a concern about binary incompatibility due to changing `PreserveCount`'s value.

wenju-he: deleted. It is unused.

VyacheslavLevytskyy (approved):

PR #2948, opened 2024-12-29
handle OpBitcast between pointers and non-pointers

bashbaug:
Adds support for SPIR-V [OpBitcast](https://registry.khronos.org/SPIR-V/specs/unified1/SPIRV.html#OpBitcast) instructions where the source is a pointer and the destination is not a pointer, and where the source is not a pointer and the destination is a pointer.  This needs to be handled as a special case because the LLVM [bitcast](https://llvm.org/docs/LangRef.html#bitcast-to-instruction) instruction does not support this.

Handles bitcasts between pointers and scalar integers, which is supported by all SPIR-V versions, and pointers and vectors of integers, which is supported by SPIR-V 1.5 (though only for vectors of 32-bit integers).

### DIFF
--- a/lib/SPIRV/SPIRVReader.cpp
+++ b/lib/SPIRV/SPIRVReader.cpp
@@ -1086,12 +1086,34 @@ Value *SPIRVToLLVM::transConvertInst(SPIRVValue *BV, Function *F,
     CO = IsExt ? Instruction::FPExt : Instruction::FPTrunc;
     break;
   case OpBitcast:
+    CO = Instruction::BitCast;
     if (Src->getType()->isPointerTy() && !Dst->isPointerTy()) {
-      CO = Instruction::PtrToInt;
+      if (auto* DstVecTy = dyn_cast<FixedVectorType>(Dst)) {
+        assert(DstVecTy->getElementType()->isIntegerTy(32) &&
+               DstVecTy->getNumElements() == 2 &&
+               "Expected vector of two 32-bit integer components");
+        auto *Int64Ty = Type::getInt64Ty(BB->getContext());
+        if (BB) {
+          Src = CastInst::CreatePointerCast(Src, Int64Ty, "", BB);
+        } else {
+          Src = ConstantExpr::getPointerCast(dyn_cast<Constant>(Src), Int64Ty);
+        }
+      } else {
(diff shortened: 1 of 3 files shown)


test/OpBitcast_ptr_vector.spvasm
```diff
@@ -0,0 +1,34 @@
...
+; RUN: llvm-spirv -r %t.spv -o %t.rev.bc
+; RUN: llvm-dis %t.rev.bc
+; RUN: FileCheck < %t.rev.ll %s --check-prefix=CHECK-LLVM
+               OpCapability Addresses
+               OpCapability Kernel
+               OpCapability Int64
+               OpMemoryModel Physical64 OpenCL
+               OpEntryPoint Kernel %kernel "test"
+       %uint = OpTypeInt 32 0
+      %ulong = OpTypeInt 64 0
```
svenvh, line 16: Seems unused.

lib/SPIRV/SPIRVReader.cpp
```diff
@@ -1085,6 +1085,37 @@ Value *SPIRVToLLVM::transConvertInst(SPIRVValue *BV, Function *F,
   case OpFConvert:
     CO = IsExt ? Instruction::FPExt : Instruction::FPTrunc;
     break;
+  case OpBitcast:
+    CO = Instruction::BitCast;
```
svenvh, line 1089: Perhaps it would be nice to add a comment here to highlight the difference between `OpBitcast` and llvm's `bitcast` (repeating the PR description would work for me).

test/OpBitcast_ptr_vector.spvasm
bashbaug: Thanks!  I adapted this test from the CTS tests I've been working on and this one slipped through.

Out of curiosity, are you using a tool to find these?

lib/SPIRV/SPIRVReader.cpp
bashbaug: Sure, done.

test/OpBitcast_ptr_vector.spvasm
svenvh: > Out of curiosity, are you using a tool to find these?

Nope, just manual inspection. :-)

PR #2960, opened 2025-01-10
Use `poison` instead of `undef` values wherever possible

### LINKED ISSUE #2953: Generate `poison` values instead of `undef` in reverse translation to align with the LLVM community
> Undef is now deprecated and should be used only to represent loads of uninitialized memory.

https://llvm.org/docs/UndefinedBehavior.html
https://llvm.org/docs/LangRef.html#undefined-values

vmaksimo:
In reverse translation always use `poison` values except cases where we translate `OpUndef`. In other cases common sense was used to define where it is safe to replace `undef` to `poison` (e.g. in creating structures, regularization passes).

This resolves #2953 and aligns us with the LLVM community in terms of generating IR.


lib/SPIRV/SPIRVRegularizeLLVM.cpp
```diff
@@ -231,7 +231,7 @@ void SPIRVRegularizeLLVMBase::buildUMulWithOverflowFunc(Function *UMulFunc) {
   // umul.with.overflow intrinsic return a structure, where the first element
   // is the multiplication result, and the second is an overflow bit.
   auto *StructTy = UMulFunc->getReturnType();
-  auto *Agg = Builder.CreateInsertValue(UndefValue::get(StructTy), Mul, {0});
+  auto *Agg = Builder.CreateInsertValue(PoisonValue::get(StructTy), Mul, {0});
```
MrSidims, line 234: Why changes in SPIRVRegularizeLLVM don't break tests? We don't have poison in SPIR-V, meanwhile you add it to the input LLVM IR.

test/extensions/KHR/SPV_KHR_untyped_pointers/basic.ll
```diff
@@ -0,0 +1,35 @@
+; RUN: llvm-as %s -o %t.bc
```
MrSidims, line 1: Are this and following few tests from a different patch?

vmaksimo: Oopsy, yes, will remove

lib/SPIRV/SPIRVRegularizeLLVM.cpp
vmaksimo: I'm not sure, maybe because I changed check-lines in tests as well?
But whatever the reason, I think that we can safely use `poison` instead of `undef` here because it's not the actual input, but the value we insert ourselves. This way we should not break input semantics. 
What do you think?

MrSidims: > But whatever the reason, I think that we can safely use poison instead of undef here because it's not the actual input, but the value we insert ourselves. This way we should not break input semantics.

Incorrect. If we insert something in LLVM IR that SPIRVWriter doesn't support - it will emit an error.

So I meant to say: if the translator doesn't handle `poison`, then this change would result in a regression when compiling a user code. But I've double checked it, and it seems like the translator is actually handling poison values (at least at some extend). For example in `transConstant` for the following LLVM IR snippet:
```
%structtype = type { i32, float }

define spir_func %structtype @foo() {
entry:
  ret %structtype { i32 1, float poison }
}

```
[this check](https://github.com/KhronosGroup/SPIRV-LLVM-Translator/blob/main/lib/SPIRV/SPIRVWriter.cpp#L1499) results in `true`, so `OpUndef` is added instead of `poison`.

So at least for constants we are safe. Revisiting our offline conversation: guess we can replace more 'undef' code to 'poison' during forward translation.

MrSidims:
Restarting CI

lib/SPIRV/SPIRVUtil.cpp
```diff
@@ -2032,7 +2032,7 @@ static void replaceUsesOfBuiltinVar(Value *V, const APInt &AccumulatedOffset,
           if (!Index.isZero() || DL.getTypeSizeInBits(VecTy) !=
                                      DL.getTypeSizeInBits(Load->getType()))
             llvm_unreachable("Illegal use of a SPIR-V builtin variable");
-          Replacement = UndefValue::get(VecTy);
+          Replacement = PoisonValue::get(VecTy);
```
svenvh, line 2035: With this change you should probably also update the `%2 = insertelement <3 x i64> undef, i64 %1, i32 0` in `docs/SPIRVRepresentationInLLVM.rst`?

svenvh (commented):
LGTM, just a small nit.

svenvh (approved):

PR #2965, opened 2025-01-17
Adapt to change introduced by #122928 in upstream

AlexVlx:
Change #122928 in LLVM modifies the signature of `createObjectPointerType` in a backwards incompatible fashion; this adapts the translator's single use of that interface to match the new signature, whilst retaining existing behaviour.


AlexVlx:
This is predicated on / will become useful once the CI hoovers in the upstream change. @MrSidims for attention since I cannot add reviewers / labels.

svenvh (approved):

PR #2258, opened 2023-12-08
Update LongConstantCompositeINTEL to LongCompositesINTEL capability after Headers change

vmaksimo:
Continue #2257
This fixes https://github.com/KhronosGroup/SPIRV-LLVM-Translator/issues/2261

The change is needed after https://github.com/KhronosGroup/SPIRV-Registry/commit/5d60c7d04589c3a84dee5abc6bc9bc5d1c045e3f and https://github.com/KhronosGroup/SPIRV-Headers/commit/d5acd42cb7cdaa5d0f2c5f739208eeeed395f491

PR #2968, opened 2025-01-24
Replace SPV_INTEL_long_composites ext with the SPV_INTEL_long_composites

vmaksimo:
This continues #2258
All the backports with the rename has reached the backend drivers, so now it's safe to remove the old naming.


MrSidims (approved):
Our E2E tests pass with this change, right?

vmaksimo:
> Our E2E tests pass with this change, right?

@MrSidims looks so!

maksimsab:
Hi @vmaksimo !

This change should be adopted in Clang Driver.

https://github.com/intel/llvm/blob/20d4bec7e6413c1f6350981e420931ea1b8457ea/clang/lib/Driver/ToolChains/Clang.cpp#L10790

Otherwise, it will start failing.

PR #2978, opened 2025-01-29
Update the list of supported by the SPIR-V Backend extensions and remove unneeded options in the call to the SPIR-V Backend API call

VyacheslavLevytskyy:
This PR is to fix usage of the SPIR-V Backend API by updating the list of supported by the SPIR-V Backend extensions and removing no longer needed options in the call to the SPIR-V Backend API call. More details about the context of the change are available in https://github.com/llvm/llvm-project/pull/124745 where The SPIR-V Backend API was refactored wrt. a way it's working with user facing options.


svenvh (approved):
Thanks!

MrSidims (approved):

PR #2981, opened 2025-01-30
Fix DebugLinePriority.spt SPIR-V validation errors

svenvh:
This test had the following issues:

 - The ID bound in the header was set too low.

 - The OpVariable with ID 6 was declared with the Function (7) storage class, but defined outside of a Function.  Change its storage class to CrossWorkgroup (5).

Fix those and run spirv-val on this test.

### DIFF
--- a/test/DebugInfo/NonSemantic/Shader200/DebugLinePriority.spt
+++ b/test/DebugInfo/NonSemantic/Shader200/DebugLinePriority.spt
@@ -1,14 +1,15 @@
 ; RUN: llvm-spirv -to-binary %s -o %t.spv
+; RUN: spirv-val %t.spv
 ; RUN: llvm-spirv -r %t.spv -o %t.rev.bc
 ; RUN: llvm-dis %t.rev.bc -o %t.rev.ll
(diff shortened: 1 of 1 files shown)


test/DebugInfo/NonSemantic/Shader200/DebugLinePriority.spt
```diff
@@ -1,14 +1,15 @@
 ; RUN: llvm-spirv -to-binary %s -o %t.spv
+; RUN: spirv-val %t.spv
```
MrSidims, line 2: Interesting, I haven't yet added NonSemantic.Shader.DebugInfo.200 to spirv-val and spirv-headers, does it mean, that spir-v val skips nonsemantic instruction sets?

svenvh: I suppose so; I haven't looked in detail in the spirv-val sources, but it seems to exclude NonSemantic instructions in various places by checking for a `NonSemantic.` prefix.  The best "proof" I have is that this test passes `spirv-val` after fixing the issues. :-)

MrSidims (approved):

PR #2999, opened 2025-02-06
[CI] Move away from deprecated set-output

svenvh:
See https://github.blog/changelog/2022-10-11-github-actions-deprecating-save-state-and-set-output-commands/


MrSidims (approved):
Thanks!

PR #3010, opened 2025-02-11
[SPIR-V 1.6] Allow UniformDecoration capability for Uniform

svenvh:
Starting from SPIR-V 1.6, the `UniformDecoration` capability also enables the `Uniform` decoration (in addition to the `Shader` capability).  llvm-spirv was not aware of this, and would always add the `Shader` (and implied `Matrix`) capabilities when consuming a SPIR-V module with a `Uniform` decoration.

### DIFF
--- a/lib/SPIRV/libSPIRV/SPIRVDecorate.h
+++ b/lib/SPIRV/libSPIRV/SPIRVDecorate.h
@@ -83,6 +83,10 @@ public:
       BuiltIn BI = static_cast<BuiltIn>(Literals.back());
       return getCapability(BI);
     }
+    case DecorationUniform:
+      if (Module->isAllowedToUseVersion(VersionNumber::SPIRV_1_6))
+        return getVec(CapabilityUniformDecoration);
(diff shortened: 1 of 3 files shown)


vmaksimo (approved):

lib/SPIRV/libSPIRV/SPIRVDecorate.h
```diff
@@ -83,6 +83,10 @@ class SPIRVDecorateGeneric : public SPIRVAnnotationGeneric {
       BuiltIn BI = static_cast<BuiltIn>(Literals.back());
       return getCapability(BI);
     }
+    case DecorationUniform:
```
MrSidims, line 86: We don't have `DecorationUniformId` enlisted in `lib/SPIRV/libSPIRV/SPIRVEnum.h` yet it seem to be known to the translator (at least in `lib/SPIRV/libSPIRV/SPIRVNameMapEnum.h`). Shouldn't we add the appropriate check for it as well?

MrSidims (approved):

Issue #2638, opened 2024-07-08
Missing dimidx checks when translating get_global_offset


bashbaug:
For the following kernel, the `dimidx` value passed to `get_global_offset` will be out-of-range for at least some calls regardless of the value of `j`:

```opencl
kernel void dynamic(global int* out, int j) {
	out[0] = get_global_offset(j-2);
	out[1] = get_global_offset(j-1);
	out[2] = get_global_offset(j);
	out[3] = get_global_offset(j+1);
	out[4] = get_global_offset(j+2);
}
```

The OpenCL C spec has defined out-of-range behavior for `get_global_offset` though, so this program is correct:

> Valid values of dimindx are 0 to get_work_dim() - 1. For other values, get_global_offset() returns 0.

However, when I generate SPIR-V using Clang and the SPIR-V LLVM Translator there are no range checks:

```spir-v
    %dynamic = OpFunction %void None %9
         %11 = OpFunctionParameter %_ptr_CrossWorkgroup_uint
         %12 = OpFunctionParameter %uint
         %13 = OpLabel
         %15 = OpIAdd %uint %12 %uint_4294967294
         %16 = OpLoad %v3ulong %__spirv_BuiltInGlobalOffset Aligned 32
         %17 = OpVectorExtractDynamic %ulong %16 %15
         %18 = OpUConvert %uint %17
               OpStore %11 %18 Aligned 4
         %20 = OpIAdd %uint %12 %uint_4294967295
         %21 = OpLoad %v3ulong %__spirv_BuiltInGlobalOffset Aligned 32
         %22 = OpVectorExtractDynamic %ulong %21 %20
         %23 = OpUConvert %uint %22
         %25 = OpInBoundsPtrAccessChain %_ptr_CrossWorkgroup_uint %11 %ulong_1
               OpStore %25 %23 Aligned 4
         %26 = OpLoad %v3ulong %__spirv_BuiltInGlobalOffset Aligned 32
         %27 = OpVectorExtractDynamic %ulong %26 %12
         %28 = OpUConvert %uint %27
         %30 = OpInBoundsPtrAccessChain %_ptr_CrossWorkgroup_uint %11 %ulong_2
               OpStore %30 %28 Aligned 4
         %32 = OpIAdd %uint %12 %uint_1
         %33 = OpLoad %v3ulong %__spirv_BuiltInGlobalOffset Aligned 32
         %34 = OpVectorExtractDynamic %ulong %33 %32
[...]

PR #3023, opened 2025-02-19
Emit select for out-of-range builtin var indices

svenvh:
The behaviour for out-of-range dimension arguments to work-item functions is well defined in OpenCL C.  For example, `get_global_size` must return 1 if its argument is larger than `get_work_dim() - 1`.

Ensure the generated `extractelement` index never exceeds the vector size and return the correct out-of-range value (which is either 0 or 1 depending on the builtin).

Fixes https://github.com/KhronosGroup/SPIRV-LLVM-Translator/issues/2638 .

### DIFF
--- a/test/DebugInfo/builtin-get-global-id.ll
+++ b/test/DebugInfo/builtin-get-global-id.ll
@@ -33,7 +33,8 @@ entry:
 ; CHECK-NEXT: [[I3:%[0-9]]] = insertelement <3 x i64> [[I1]], i64 [[I2]], i32 1, !dbg [[DBG]]
 ; CHECK-NEXT: [[I4:%[0-9]]] = call spir_func i64 @_Z13get_global_idj(i32 2) #1, !dbg [[DBG]]
 ; CHECK-NEXT: [[I5:%[0-9]]] = insertelement <3 x i64> [[I3]], i64 [[I4]], i32 2, !dbg [[DBG]]
-; CHECK-NEXT: %call = extractelement <3 x i64> [[I5]], i32 0, !dbg [[DBG]]
+; CHECK-NEXT: [[I6:%[0-9]]] = extractelement <3 x i64> [[I5]], i32 0, !dbg [[DBG]]
+; CHECK-NEXT: %call = select i1 true, i64 [[I6]], i64 0, !dbg [[DBG]]
   store i64 %call, ptr %gid, align 8, !dbg !11
   ret void, !dbg !12
 }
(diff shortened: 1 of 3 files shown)


test/DebugInfo/builtin-get-global-id.ll
```diff
@@ -33,7 +33,8 @@ entry:
 ; CHECK-NEXT: [[I3:%[0-9]]] = insertelement <3 x i64> [[I1]], i64 [[I2]], i32 1, !dbg [[DBG]]
 ; CHECK-NEXT: [[I4:%[0-9]]] = call spir_func i64 @_Z13get_global_idj(i32 2) #1, !dbg [[DBG]]
 ; CHECK-NEXT: [[I5:%[0-9]]] = insertelement <3 x i64> [[I3]], i64 [[I4]], i32 2, !dbg [[DBG]]
-; CHECK-NEXT: %call = extractelement <3 x i64> [[I5]], i32 0, !dbg [[DBG]]
+; CHECK-NEXT: [[I6:%[0-9]]] = extractelement <3 x i64> [[I5]], i32 0, !dbg [[DBG]]
+; CHECK-NEXT: %call = select i1 true, i64 [[I6]], i64 0, !dbg [[DBG]]
```
MrSidims, line 37: Hm, a bit odd, that IRBuilder doesn't fold this select, but it should be fine

test/get_global_size.cl
```diff
@@ -0,0 +1,36 @@
...
+  // CHECK-DAG: Constant [[#]] [[#CONST64_1:]] 1 0
+  // CHECK-DAG: Constant [[#]] [[#CONST3:]] 3
+  // CHECK-DAG: Constant [[#]] [[#CONST0:]] 0
+
+  // CHECK: FunctionParameter [[#]] [[#PARAMOUT:]]
+  // CHECK: FunctionParameter [[#]] [[#PARAMX:]]
+
+  // CHECK: Load [[#]] [[#LD0:]]
+  // CHECK: CompositeExtract [[#]] [[#SCAL0:]] [[#LD0]] 0
+  // CHECK: Select [[#]] [[#RES0:]] [[#]] [[#SCAL0]] [[#CONST64_1]]
```
MrSidims, line 19: I assume, that `IRBuilder` folds constant expressions here and we get `OpSelect` with `OpConstantTrue` compare parameter, right? Should we check it explicitly?

MrSidims (approved):
just test nits, other then that the patch is LGTM

test/DebugInfo/builtin-get-global-id.ll
svenvh: I suppose the IRBuilder cannot fold this because the true branch isn't a constant.  It would be up to something like InstCombine to fold the `select` into another instruction.

test/get_global_size.cl
svenvh: That's right, I've updated the `OpSelect` CHECKs.

PR #3027, opened 2025-02-25
[Backport to 19] Fix incorrect translation of calls to a builtin that returns a structure (#2722)

VyacheslavLevytskyy:
This PR fixes the issue with Khronos Translator incorrectly translating calls to builtins that returns a structure and generating incorrect extractvalue applied to a pointer. The PR is to fix the issue by preserving equivalence of newly inserted values with prior values of function return's type.


VyacheslavLevytskyy:
> LGTM, just please fix the description - I believe it's better we don't reference the code in the fork codebase

ok, thank you!

PR #3028, opened 2025-02-25
[Backport to 18] Fix incorrect translation of calls to a builtin that returns a structure (#2722)

VyacheslavLevytskyy:
This PR fixes the issue with Khronos Translator incorrectly translating calls to builtins that returns a structure and generating incorrect extractvalue applied to a pointer. The PR is to fix the issue by preserving equivalence of newly inserted values with prior values of function return's type.

### DIFF
--- a/lib/SPIRV/SPIRVReader.cpp
+++ b/lib/SPIRV/SPIRVReader.cpp
@@ -2758,20 +2758,12 @@ Value *SPIRVToLLVM::transValueWithoutDecoration(SPIRVValue *BV, Function *F,
     return mapValue(BV,
                     transRelational(static_cast<SPIRVInstruction *>(BV), BB));
   case OpIAddCarry: {
-    IRBuilder Builder(BB);
     auto *BC = static_cast<SPIRVBinary *>(BV);
-    return mapValue(BV, Builder.CreateBinaryIntrinsic(
-                            Intrinsic::uadd_with_overflow,
-                            transValue(BC->getOperand(0), F, BB),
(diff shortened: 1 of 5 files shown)


lib/SPIRV/SPIRVReader.cpp
```diff
@@ -2758,20 +2758,12 @@ Value *SPIRVToLLVM::transValueWithoutDecoration(SPIRVValue *BV, Function *F,
     return mapValue(BV,
                     transRelational(static_cast<SPIRVInstruction *>(BV), BB));
   case OpIAddCarry: {
-    IRBuilder Builder(BB);
     auto *BC = static_cast<SPIRVBinary *>(BV);
-    return mapValue(BV, Builder.CreateBinaryIntrinsic(
```
MrSidims, line 2763: Could you please remind me, if this code was chronologically added after initial translation of OpIAddCarry? If so, why to remove it?

VyacheslavLevytskyy: No, this code existed before translation into a spirv-friendly builtin was introduced with `mapValue(BV, transBuiltinFromInst("__spirv_IAddCarry", BC, BB))`. I don't know the reason for this llvm18 branch to differ in this respect from all other branches, but the fact is that this llvm18 was behind all other and did this translation to `Intrinsic::uadd_with_overflow,` that was replaced in all other branches by `__spirv_IAddCarry`.

MrSidims (approved):

PR #3030, opened 2025-02-25
[Backport to 17] Fix incorrect translation of calls to a builtin that returns a structure

VyacheslavLevytskyy:
This PR fixes https://github.com/KhronosGroup/SPIRV-LLVM-Translator/issues/2721
The root cause is that https://github.com/bwlodarcz/SPIRV-LLVM-Translator/blob/ca48eaf52ae791abc61cba13ef383f022e151e9f/lib/SPIRV/SPIRVUtil.cpp#L2207 doesn't preserve equivalence of values, inserting a pointer where a value of function return's type was before.

A new test case is added from the reproducer of https://github.com/KhronosGroup/SPIRV-LLVM-Translator/issues/2721

### DIFF
--- a/lib/SPIRV/SPIRVUtil.cpp
+++ b/lib/SPIRV/SPIRVUtil.cpp
@@ -2259,7 +2259,7 @@ bool postProcessBuiltinReturningStruct(Function *F) {
       if (!A) {
         A = Builder.CreateAlloca(F->getReturnType());
       }
-      std::vector<Type *> ArgTys;
+      SmallVector<Type *> ArgTys;
       getFunctionTypeParameterTypes(F->getFunctionType(), ArgTys);
       ArgTys.insert(ArgTys.begin(), A->getType());
       auto *NewF =
--- a/lib/SPIRV/SPIRVUtil.cpp
+++ b/lib/SPIRV/SPIRVUtil.cpp
@@ -2248,8 +2248,9 @@ bool postProcessBuiltinReturningStruct(Function *F) {
       Builder.SetInsertPoint(CI);
       SmallVector<User *> Users(CI->users());
       Value *A = nullptr;
+      StoreInst *SI = nullptr;
       for (auto *U : Users) {
-        if (auto *SI = dyn_cast<StoreInst>(U)) {
+        if ((SI = dyn_cast<StoreInst>(U)) != nullptr) {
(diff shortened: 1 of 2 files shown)


lib/SPIRV/SPIRVUtil.cpp
```diff
@@ -2269,12 +2270,18 @@ bool postProcessBuiltinReturningStruct(Function *F) {
       NewF->setCallingConv(F->getCallingConv());
       auto Args = getArguments(CI);
       Args.insert(Args.begin(), A);
-      CallInst *NewCI = Builder.CreateCall(NewF, Args, CI->getName());
+      CallInst *NewCI = Builder.CreateCall(
+          NewF, Args, NewF->getReturnType()->isVoidTy() ? "" : CI->getName());
```
vmaksimo, line 2274: Why do we change Name argument to `CreateCall`? Neither the llvm-17 branch, nor your fix to the main branch contains this change

VyacheslavLevytskyy: I've hit an error on the 18th LLVM version with test/builtin_returns_struct.spvasm -- given a newly inserted `call spir_func void @_Z17__spirv_IAddCarryii(ptr sret(%structtype) %[[#Var:]]` the code tried to use a non-empty name for the instruction and LLVM failed with an error. This check prevents this fail.

vmaksimo: Ok, understandable. Although, I don't get why we don't face this issue on the main branch or even on the llvm-19. Do you have any thoughts/observations?

VyacheslavLevytskyy: I've added the same check for a void-typed instruction into the "Backport to 19" patch.

MrSidims (approved):

lib/SPIRV/SPIRVUtil.cpp
VyacheslavLevytskyy: @vmaksimo do you have any followup to the Name argument to `CreateCall` please?

vmaksimo: @MrSidims actually I approved before this check was added.
That's still strange, but let's proceed with it.

PR #3035, opened 2025-03-04
[Backport to 14] Add option to control builtin format for reverse translation.

haonanya1:
Currently, we always convert SPIR-V bultins to globals for forward translation and to functions for reverse translation.
I have a use case where I want to keep them as globals for reverse translation, so I added this mode.
Implementations for both cases already existed, I just consolidated them and added the option.

Backport of https://github.com/KhronosGroup/SPIRV-LLVM-Translator/pull/1986.

### DIFF
--- a/include/LLVMSPIRVOpts.h
+++ b/include/LLVMSPIRVOpts.h
@@ -109,6 +109,8 @@ enum class DebugInfoEIS : uint32_t {
   NonSemantic_Shader_DebugInfo_200
 };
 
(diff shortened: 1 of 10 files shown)


haonanya1:
@MrSidims and  @svenvh , we want to backport your fix https://github.com/KhronosGroup/SPIRV-LLVM-Translator/pull/3023 to 140 branch, the fix is based on several commit, I will backport them one by one. Could you please help trigger CI? Thanks very much.

PR #3041, opened 2025-03-06
[Backport to 18] Emit select for out-of-range builtin var indices

haonanya1:
The behaviour for out-of-range dimension arguments to work-item functions is well defined in OpenCL C. For example, get_global_size must return 1 if its argument is larger than get_work_dim() - 1.

Ensure the generated extractelement index never exceeds the vector size and return the correct out-of-range value (which is either 0 or 1 depending on the builtin).


haonanya1:
@svenvh , @MrSidims, please review, thanks.

MrSidims (approved):

PR #3043, opened 2025-03-06
[Backport to 20] Emit select for out-of-range builtin var indices

haonanya1:


haonanya1:
@svenvh , @MrSidims, please review, thanks.

svenvh (approved):

PR #3050, opened 2025-03-07
properly handle LOD values when nontemporal image operand is present

### LINKED ISSUE #3049: Bad Reverse Translation with SPIR-V 1.6 Nontemporal Image Operand
We seem to be translating SPIR-V 1.6 modules with the optional "Nontemporal" image operand incorrectly.

Without the "Nontemporal" image operand, we turn this SPIR-V instruction:

```
%data0 = OpImageSampleExplicitLod %v4float %si %coord_0_0 Lod %float_0
```

We translate this into a call to `read_imagef` without the `float` LOD parameter:

```
%4 = call spir_func <4 x float> @_Z11read_imagef14ocl_image2d_ro11ocl_samplerDv2_i(ptr addrspace(1) %0, ptr addrspace(2) %3, <2 x i32> zeroinitializer) #0
```

With the "Nontemporal" image operand, though, we include a `float` LOD parameter:

```
%5 = call spir_func <4 x float> @_Z11read_imagef14ocl_image2d_ro11ocl_samplerDv2_if(ptr addrspace(1) %0, ptr addrspace(2) %3, <2 x i32> zeroinitializer, float 0.000000e+00) #0
```

This is fine for implementations that support the [cl_khr_mipmap_image](https://registry.khronos.org/OpenCL/specs/3.0-unified/html/OpenCL_C.html#cl_khr_mipmap_image) extension, but it won't work for implementations that do not support the extension.  We should fix this so we generate the same function call with "Nontemporal" that we do without.

bashbaug:
Instead of checking that the image operands are equal to the LOD mask, check that the LOD bit is in the image operands.  This way an LOD value of zero may be ignored even when image operands contains bits other than the LOD bit, such as for the SPIR-V 1.6 nontemporal image operand.

fixes #3049

### DIFF
--- a/lib/SPIRV/SPIRVToOCL.cpp
+++ b/lib/SPIRV/SPIRVToOCL.cpp
@@ -752,8 +752,10 @@ SPIRVToOCLBase::mutateCallImageOperands(CallInst *CI, StringRef NewFuncName,
       ConstantFP *LodVal = dyn_cast<ConstantFP>(Mutator.getArg(ImOpArgIndex));
       // If the image operand is LOD and its value is zero, drop it too.
       if (LodVal && LodVal->isNullValue() &&
-          ImOpValue == ImageOperandsMask::ImageOperandsLodMask)
+          ImOpValue & ImageOperandsMask::ImageOperandsLodMask) {
         Mutator.removeArgs(ImOpArgIndex, Mutator.arg_size() - ImOpArgIndex);
+        ImOpValue &= ~ImageOperandsMask::ImageOperandsLodMask;
+      }
     }
   }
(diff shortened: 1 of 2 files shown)


svenvh (approved):

MrSidims (approved):

MrSidims:
restarting CI

bashbaug:
If you need me to pick up the latest changes for clean CI just let me know.

MrSidims:
> If you need me to pick up the latest changes for clean CI just let me know.

Close/reopen PR updates HEAD, so no need to update the branch on your side.

PR #3066, opened 2025-03-13
Fix S/UMulExtended instructions translation on llvm_release_15

vsemenov368:


MrSidims:
Seems like that the necessary code is already there https://github.com/KhronosGroup/SPIRV-LLVM-Translator/blob/llvm_release_150/lib/SPIRV/SPIRVWriter.cpp#L5556

PR #3084, opened 2025-03-24
[SPIRV-to-LLVM] Set LocalId parameter of GroupBroadcast to unsigned in SPV-IR

wenju-he:
LocalId should not be negative. Change its type from signed to unsigned in SPV-IR aligns with OpenCL work_group_broadcast built-in and use of GroupBroadcast built-in in SYCL header.

Also change NumElements/Stride parameters of OpGroupAsyncCopy to unsigned.


MrSidims (approved):
LGTM
out-of-tree build CI failures are unrelated, looking into it

PR #3103, opened 2025-04-01
[DebugInfo] Fix translation of some cases of mem2reged local variables

MrSidims:
We should translate null/undef declare records to DebugDeclare with OpDebugInfoNone instruction referenced in OpVariableId operand.

TODO: currently the specification says, that only OpVariable can be referenced by this operand. This seem to be a spec bug, we should allow OpDebugInfoNone there.

### DIFF
--- a/test/DebugInfo/mem2reged_local_var.ll
+++ b/test/DebugInfo/mem2reged_local_var.ll
@@ -1,4 +1,4 @@
-; Check if the translator handles #dbg_declare(ptr null ... correctly
+; Check if the translator handles #dbg_declare(ptr null ...) correctly
 
 ; RUN: llvm-as %s -o %t.bc
 ; RUN: llvm-spirv %t.bc -o %t.spv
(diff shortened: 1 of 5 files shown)


MrSidims:
For out-of-tree build failures - it's a regression in llvm's LLVM_INSTALL_PREFIX resolution.

test/DebugInfo/mem2reged_local_var.ll
```diff
@@ -0,0 +1,59 @@
+; Check if the translator handles #dbg_declare(ptr null ... correctly
```
maarquitos14, line 1:
```suggestion
; Check if the translator handles #dbg_declare(ptr null ...) correctly
```

maarquitos14 (approved):
Just a nit.

maarquitos14:
> We should translate null/undef declare records to DebugDeclare with OpDebugInfoNone instruction referenced in OpVariableId operand.
> 
> TODO: currently the specification says, that only OpVariable can be referenced by this operand. This seem to be a spec bug, we should allow OpDebugInfoNone there.

This makes sense to me. Have we reported this to Spec people so that they can address it?

svenvh (approved):

PR #3118, opened 2025-04-14
[NFC] Update tests after community changes in llorg

vmaksimo:


MrSidims:
transcoding/SampledImage.cl remains not fixes, as there seem to be a bug in the community's patch, see: https://godbolt.org/z/dc3T7Mo3G

vmaksimo:
@MrSidims WDYT, maybe we should adapt `SampledImage.cl` test before it is fixed in the community? In order to fix our in-tree CI

MrSidims (approved):

PR #3123, opened 2025-04-18
[OCLToSPIRV] Takes param size and align from literal block in OpEnqueueKernel

EbinJose2002:
- Implemented the TODO in the visitCallEnqueueKernel function
- Takes param size and align from literal block
- If dynamic allocation is used store instructions are processed to obtain size and align
- Modified test

### DIFF
--- a/lib/SPIRV/OCLToSPIRV.cpp
+++ b/lib/SPIRV/OCLToSPIRV.cpp
@@ -1504,14 +1504,54 @@ void OCLToSPIRVBase::visitCallEnqueueKernel(CallInst *CI,
   // Param: Pointer to block literal
   Value *BlockLiteral = CI->getArgOperand(BlockFIdx + 1);
   Args.push_back(BlockLiteral);
+  BlockLiteral = BlockLiteral->stripPointerCasts();
+  if (auto *GV = dyn_cast<GlobalVariable>(BlockLiteral)) {
+    assert(GV->hasInitializer() && "Block literal should have an initializer");
+    Constant *Init = GV->getInitializer();
+    if (auto *Struct = dyn_cast<ConstantStruct>(Init)) {
+      Constant *SizeConst = Struct->getOperand(0);
+      Constant *AlignConst = Struct->getOperand(1);
+      auto *SizeVal = dyn_cast<ConstantInt>(SizeConst);
+      auto *AlignVal = dyn_cast<ConstantInt>(AlignConst);
+      Args.push_back(SizeVal);
+      Args.push_back(AlignVal);
+    }
+  } else {
 
-  // Param Size: Size of block literal structure
-  // Param Aligment: Aligment of block literal structure
-  // TODO: these numbers should be obtained from block literal structure
-  Type *ParamType = getBlockStructType(BlockLiteral);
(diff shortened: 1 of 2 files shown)


lib/SPIRV/OCLToSPIRV.cpp
```diff
@@ -1504,14 +1504,55 @@ void OCLToSPIRVBase::visitCallEnqueueKernel(CallInst *CI,
   // Param: Pointer to block literal
   Value *BlockLiteral = CI->getArgOperand(BlockFIdx + 1);
   Args.push_back(BlockLiteral);
+  BlockLiteral = BlockLiteral->stripPointerCasts();
+  if (auto *GV = dyn_cast<GlobalVariable>(BlockLiteral)) {
+    if (GV->hasInitializer()) {
```
MrSidims, line 1509: Q: is it a possible scenario when Block GV doesn't have an initializer? It might be for the best to add an assertion. If yes - this case would be unhandled with this change.

lib/SPIRV/OCLToSPIRV.cpp
```diff
@@ -1504,14 +1504,55 @@ void OCLToSPIRVBase::visitCallEnqueueKernel(CallInst *CI,
...
+      if (auto *GEP = dyn_cast<GetElementPtrInst>(U)) {
+        if (GEP->getNumIndices() < 2)
+          continue;
+        auto *CI1 = dyn_cast<ConstantInt>(GEP->getOperand(2));
+        if (!CI1)
+          continue;
+        uint64_t FieldIndex = CI1->getZExtValue();
+        for (User *GEPUser : GEP->users()) {
+          if (auto *Store = dyn_cast<StoreInst>(GEPUser)) {
+            if (FieldIndex == 0 && !ParamSizeVal)
```
MrSidims, line 1536: This looks fishy, why do we need to ensure, that `ParamSizeVal` was not previously assigned? Is it a possible situation, when we store to the same GEP twice for block literals?

lib/SPIRV/OCLToSPIRV.cpp
```diff
@@ -1504,14 +1504,55 @@ void OCLToSPIRVBase::visitCallEnqueueKernel(CallInst *CI,
...
+              ParamSizeVal = Store->getValueOperand();
```
MrSidims, line 1537: Shouldn't we check if the value a `ConstantInt`?

lib/SPIRV/OCLToSPIRV.cpp
EbinJose2002: So far the only 2 cases I have seen are the static block literal like these (having initializer) and using pointers to assign the values using store. So what should I do?

MrSidims: I can neither imagine a case of a block GV not having an initializer. I'd put an assertion for this just in case.

MrSidims (approved):
LGTM, thanks for the contribution!

PS for future contributions, please avoid force pushing (unless you must) to update the PR. We can always merge PRs with squash and merge strategy, so there won't be any unnecessary commits in the git log.

PR #3124, opened 2025-04-18
Refactoring Two functions into One

EbinJose2002:
- Refactored addUntypedPointerKHRType and addPointerType functions into a single function (It was marked as TODO)
- Combined both functions into one with return type SPIRVType *
- Also re inserted an assert and removed unnecessary if (Marked as TODO)


MrSidims (approved):
LGTM, but lets hold the merge for a while, so we could fix the CI first

lib/SPIRV/SPIRVWriter.cpp
```diff
@@ -2350,7 +2352,7 @@ LLVMToSPIRVBase::transValueWithoutDecoration(Value *V, SPIRVBasicBlock *BB,
       // TODO: refactor addPointerType and addUntypedPointerKHRType in one
       // method if possible.
```
vmaksimo, line 2353: can remove TODO now :)

vmaksimo (approved):

PR #3141, opened 2025-05-08
Enabled RuntimeAlignedINTEL functionParameterAttribute.

EbinJose2002:
- Current implemention was non-conformant with spec. So removed it.Uncommented actual implementation.
- `error: 24: Invalid decoration operand: 5940`
This error would show up on trying to run _/test/extensions/INTEL/SPV_INTEL_runtime_aligned/RuntimeAligned.ll_ . So corrected it.


vmaksimo:
We can merge this fix only after verifying that all the drivers were actually updated to consume"new" (spec conformant) format. It's been a year since #1796 was merged but I believe we'd better do a test run in intel/llvm. WDYT, @MrSidims?

Issue #3126, opened 2025-04-23
Disassembled spv file on Windows contains `OpString` with Linux like separator


anmyachev:
```llvm
%41 = OpString "...\\pytest-3\\test_line_info_ir_source_ttir_0/test.ttir"
```

I guess this happened during the translation from `.llir` to `spv`:
```llvm
!1 = !DIFile(filename: "test.ttir", directory: "...\\pytest-3\\test_line_info_ir_source_ttir_0")
```

anmyachev:
@whitneywhtsang maybe you know someone I can ping here about this issue? If one know where the concatenation is happening, it should be easy to fix.

whitneywhtsang:
@MrSidims @asudarsa Would you be able to help?

MrSidims:
Created https://github.com/KhronosGroup/SPIRV-LLVM-Translator/pull/3157

MrSidims:
Fixed with https://github.com/KhronosGroup/SPIRV-LLVM-Translator/pull/3157

PR #3157, opened 2025-05-12
[DebugInfo] Use native for the system separator for source path string

MrSidims:
Previously it used posix, causing strange behaviour like reported in https://github.com/KhronosGroup/SPIRV-LLVM-Translator/issues/3126

It's NFC change for the translator-based tool chains. Yet if SPIR-V consumer is not llvm-spirv - the fix should help.

Note, the patch is not attempting to guess separators depending on the user's IR input, so for the linux <-> windows cross compilation the result will be odd (like it was before).

### DIFF
--- a/test/DebugInfo/LocalAddressSpace.ll
+++ b/test/DebugInfo/LocalAddressSpace.ll
@@ -22,7 +22,7 @@
 ; CHECK: DW_TAG_variable
 ; CHECK-NEXT: DW_AT_name {{.*}} = "a")
 ; CHECK-NEXT: DW_AT_type {{.*}} "int")
-; CHECK-NEXT: DW_AT_decl_file {{.*}} ("/work/tmp{{[/\\]}}tmp.cl")
+; CHECK-NEXT: DW_AT_decl_file {{.*}} ("/work{{[/\\]}}tmp{{[/\\]}}tmp.cl")
 ; CHECK-NEXT: DW_AT_decl_line {{.*}} (2)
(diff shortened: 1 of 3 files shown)


test/DebugInfo/LocalAddressSpace.ll
```diff
@@ -22,7 +22,7 @@
 ; CHECK: DW_TAG_variable
 ; CHECK-NEXT: DW_AT_name {{.*}} = "a")
 ; CHECK-NEXT: DW_AT_type {{.*}} "int")
-; CHECK-NEXT: DW_AT_decl_file {{.*}} ("/work/tmp{{[/\\]}}tmp.cl")
+; CHECK-NEXT: DW_AT_decl_file {{.*}} ("/work{{[/\\]}}tmp{{[/\\]}}tmp.cl")
```
vmaksimo, line 25: Maybe silly question, but should not we also check the first slash/backslash?
```suggestion
; CHECK-NEXT: DW_AT_decl_file {{.*}} ("{{[/\\]}}work{{[/\\]}}tmp{{[/\\]}}tmp.cl")
```

MrSidims: Nope. This one remains to be the same as the dir name starts with /work in the test. See from the PR description:
`Note, the patch is not attempting to guess separators depending on the user's IR input, so for the linux <-> windows cross compilation the result will be odd (like it was before).`

vmaksimo (approved):

PR #3166, opened 2025-05-13
SPIRV-LLVM-Translator: fix handling of built-in SPIRV targer

lumag:
If the SPIRV-LLVM-Translator is being built together with the LLVM, set the defines declaring whether SPIRV target is also included into the LLVM library or not. Otherwise llvm-spirv execution fails because spirv-ext option is registered twice.


lumag:
According to my information, this contribution should be covered by the Corporate OS CLA between Qualcomm and Khronos and as such should not require any extra CLA. Could you please check it?

MrSidims:
> According to my information, this contribution should be covered by the Corporate OS CLA between Qualcomm and Khronos and as such should not require any extra CLA. Could you please check it?

@KhronosWebservices could you please comment on that?

KhronosWebservices:
I do not see @lumag in our Corporate CLA system. Can you please email me webservices@khronosgroup.org. Thanks.

PR #3167, opened 2025-05-14
[CI] Pick SHA of commit to cherry-pick from merge commit

MrSidims:
Previous logic was picking commits with committer's SHA. But instead we should cherry-pick the merge commit itself.

TODO: if a PR is merged with "Rebase and Merge" strategy current logic will pick only the last commit. We usually don't merged like this, yet for rare accousions we should gather all of the merged commits.


MrSidims:
Found this in https://github.com/KhronosGroup/SPIRV-LLVM-Translator/actions/runs/15019661298/job/42205593168

PR #3177, opened 2025-05-14
[CI] Add listener for other workflows to trigger them on a backport

MrSidims:
Otherwise we need to start CI manually.


MrSidims:
Without it I had to start CI in https://github.com/KhronosGroup/SPIRV-LLVM-Translator/pull/3169 manually

PR #3186, opened 2025-05-27
Prevent duplicate array type insertion into TypeMap.

maarquitos14:
Check if an array type exists in TypeMap before inserting it to prevent duplications.

### DIFF
--- a/lib/SPIRV/SPIRVWriter.cpp
+++ b/lib/SPIRV/SPIRVWriter.cpp
@@ -2696,8 +2696,7 @@ LLVMToSPIRVBase::transValueWithoutDecoration(Value *V, SPIRVBasicBlock *BB,
 
 SPIRVType *LLVMToSPIRVBase::mapType(Type *T, SPIRVType *BT) {
   assert(!T->isPointerTy() && "Pointer types cannot be stored in the type map");
-  [[maybe_unused]] auto EmplaceStatus = TypeMap.try_emplace(T, BT);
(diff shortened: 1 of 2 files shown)


vmaksimo (approved):

svenvh:
As far as I understand, the point of the assertion is not to prevent duplication in the map itself, but to warn *if* duplication would have happened.  This was added by @MrSidims in #1047 so it would be good to have his opinion too.

That said, I think this might have been an issue with non-opaque pointers only, which is now past us, so maybe we indeed no longer need this check?

maarquitos14:
> As far as I understand, the point of the assertion is not to prevent duplication in the map itself, but to warn _if_ duplication would have happened. This was added by @MrSidims in #1047 so it would be good to have his opinion too.
>
> That said, I think this might have been an issue with non-opaque pointers only, which is now past us, so maybe we indeed no longer need this check?

@MrSidims is OOO for a few days now. If no strong opposition, I'd like to have this merged and follow up with him later, if he has any concerns.

vmaksimo:
> I think we can solve this by adding a check in this branch to see if the type already exists in the map before calling `mapType`.

@maarquitos14 makes total sense to me, let's try this approach :)
Please also add a simple test case which will crash without the fix you suggested.

vmaksimo (approved):

svenvh (approved):
LGTM, thanks!

PR #3188, opened 2025-05-28
[LLVM->SPV-IR] Add const to pointer arg of prefetch, change num_elements arg type to unsigned

wenju-he:
Per SPIR-V OpenCL.ExtendedInstructionSet spec, the num_elements arg must be size_t. So this PR changes its type to unsigned in SPV-IR.
Adding const to pointer arg aligns with OpenCL spec and allows const pointer input.


svenvh (approved):

wenju-he:
@svenvh @MrSidims please help to merge, thanks.

Issue #2415, opened 2024-03-06
Build failure with llvm 18.1.0


TimB87:
Hey,

trying to update spirv-llvm-translator to v18.1.0 against the llvm 18.1.0 release from earlier today results in a build failure for me on x86_64 (on CRUX 3.7):
[spirv-llvm-translator.log](https://github.com/KhronosGroup/SPIRV-LLVM-Translator/files/14515806/spirv-llvm-translator.log)

I did disable ccache as my setup invokes distcc and would distribute it to other hosts that have llvm 17 installed.
Our current spirv-headers and spirv-tools versions are vulkan-sdk-1.3.275.0.

svenvh:
I believe you will need a more recent version of the SPIRV-Headers.

TimB87:
There is no more recent release available though.

svenvh:
Maybe @vmaksimo or @MrSidims can provide more detail, but I guess the backport #2388 introduced a dependency on SPIRV-Headers code that's not part of any official release yet?  I wonder if that's something we should try to avoid for releases (i.e., make sure that any SPIRV-Headers work that a SPIRV-LLVM-Translator release depends on is part of a SPIRV-Headers release).

TimB87:
It would certainly be helpful for packagers :)

MrSidims:
I vote to revert the backport while waiting for the official release, but would like to hear from @KorovinVlad and @vmaksimo

MrSidims:
It also means that the patch must be reverted from other branches and may be we should consider checking out headers not from github headers repo, but loading a package. Though I must say, this will complicate some backpropagation of the features to intel graphics compiler, but guess it's our problem, not community's :)

svenvh:
> may be we should consider on release branches of the translator to check out headers not from github headers repo, but loading a package

Yeah something like that, or only allow checking out tags (not raw SHA-1s) of the SPIRV-Headers on release branches.

bashbaug:
Hi!  We discussed this issue (actually, https://github.com/KhronosGroup/SPIRV-Headers/issues/422) in our weekly SPIR-V teleconference.  In short, we are open to tagging the SPIR-V Headers more frequently, on a cadence suitable for the SPIR-V LLVM Translator.  The main things we would need to know are the tagging convention this project would like to use, which commits to tag, how frequently to tag, etc.

[...]

PR #3189, opened 2025-05-28
Revert "[Backport to 19] Add SPV_INTEL_int4 extension"

vmaksimo:
This is a follow-up to #3187

To avoid tool issues as we've seen in #2415, set the SPIR-V Headers tag to the latest available release ([vulkan-sdk-1.4.313.0](https://github.com/KhronosGroup/SPIRV-Headers/releases/tag/vulkan-sdk-1.4.313.0)). 
That postpones the introduction of `SPV_INTEL_int4` extension till the next official release, which will include the headers with the new definitions.


bashbaug (approved):
Works for me, but will removing SPV_INTEL_int4 cause any issues?

Should we try to get a new SPIR-V headers tag instead?

PR #3190, opened 2025-05-29
Fix crash on coopmat conversion without decoration

svenvh:
The SPIR-V reader would crash on the added test during reverse translation.  Ensure any cooperative matrix type conversion is handled the same way as a cooperative matrix type conversion with an FPRoundingMode or SaturatedConversion decoration.


vmaksimo (approved):
LGTM.
I don't know why we didn't encounter any problems before this fix, but I also can't find why the intel/llvm frontend couldn't generate conversion instructions without rounding or saturation mode. 
Either we are lucky, or we just do not generate such IR to cope with this very translation issue.

PR #3216, opened 2025-06-06
Update tests after OpenCLKernel attribute rename

svenvh:
Update after llvm-project commit 3b9ebe92011b ("[clang] Simplify device kernel attributes  (#137882)", 2025-06-05).


svenvh:
> Probably also make sense to update https://github.com/KhronosGroup/SPIRV-LLVM-Translator/blob/main/test/transcoding/annotation_dbg_info_drop.ll while we are here

Nice catch; I wondered why that test didn't fail, turns out it was XFAILed:

```
; need to rewrite the entire test since with opaque pointers a lot of bitcasts
; are being eliminated, and those bitcasts were carrying debug information
; XFAIL: *
```

Perhaps it's time to remove this test, given that debug info intrinsics are [being retired](https://llvm.org/docs/RemoveDIsDebugInfo.html); or do you think the original problem covered by this test may still occur with the new debug info record infrastructure?

PR #3221, opened 2025-06-10
[DOC] Specify FP8 representation

MrSidims:


docs/FP8RepresentationInLLVM.rst
```diff
@@ -0,0 +1,121 @@
...
+since clang now supports vectors of sub-byte integers) for the conversion builtin functions as LLVM IR input;
+*packed* format is translated to a *vector*. Meanwhile SPIR-V consumer
+will never pack a *vector* back to *packed* format.
+
+Following list of function calls will be handled by the translator:
+Translated to *FConvert*
+
+.. code-block:: C
+
+  __builtin_spirv_ConvertHF16ToHF8EXT, __builtin_spirv_ConvertBF16ToHF8EXT,
```
MrSidims: HF8 stands for FP8E4M3 and BF8 stands for FP8E5M2. The names can be updated.

docs/FP8RepresentationInLLVM.rst
```diff
@@ -0,0 +1,121 @@
...
+There is limited number of conversions that will be supported now. Most of them will be represented by *FConvert* instruction
+in SPIR-V, which itself doesn't carry information about floating-point value's width and encoding, so there is nothing to expose
+in SPIR-V friendly LLVM IR call. This document adds a new set of external function calls
+each of which has a name that is formed from encoding a specific conversion that it performs.
+This name has a *__builtin_spirv_* prefix and *EXT* postfix (from SPV_EXT_fp8). These calls will
+be translated to SPIR-V conversion instruction operating over the appropriate types. These functions are expected to
+be mangled following Itanium C++ ABI. SPIR-V consumer will apply Itanium mangling during
+translation to LLVM IR as well.
+
+SPIR-V generator will support *scalar*, *vector* and *packed* (*packed* is for *Int4*, it's tentitive and a subject of removal,
```
svenvh:
```suggestion
SPIR-V generator will support *scalar*, *vector* and *packed* (*packed* is for *Int4*, it's tentative and a subject of removal,
```

docs/FP8RepresentationInLLVM.rst
```diff
@@ -0,0 +1,121 @@
...
+SPIR-V representation in LLVM IR for FP8 datatypes
+==========================================================
+.. contents::
+   :local:
+
+Overview
+========
+
+Open Compute and other projects are adding various new data-types and SPIR-V
+(starting from SPV_EXT_fp8) is now adopting them. None of these data-types
```
svenvh: Did you mean
```suggestion
(starting from SPV_EXT_float8) is now adopting them. None of these data-types
```

docs/FP8RepresentationInLLVM.rst
```diff
@@ -0,0 +1,121 @@
...
+The 'type resolution' instruction can be either a conversion instruction or *OpCooperativeMatrixMulAddKHR*.
+
+SPIR-V conversion instructions
+==============================
+
+There is limited number of conversions that will be supported now. Most of them will be represented by *FConvert* instruction
+in SPIR-V, which itself doesn't carry information about floating-point value's width and encoding, so there is nothing to expose
+in SPIR-V friendly LLVM IR call. This document adds a new set of external function calls
+each of which has a name that is formed from encoding a specific conversion that it performs.
+This name has a *__builtin_spirv_* prefix and *EXT* postfix (from SPV_EXT_fp8). These calls will
```
svenvh:
```suggestion
This name has a *__builtin_spirv_* prefix and *EXT* postfix (from SPV_EXT_float8). These calls will
```

docs/FP8RepresentationInLLVM.rst
svenvh: I haven't seen HF8/BF8 before, so I think I would prefer to follow the standard naming, i.e. `FP8E*M*` (although I can see the point of having shorter names as these are quite verbose).

docs/FP8RepresentationInLLVM.rst
```diff
@@ -0,0 +1,121 @@
...
+be mangled following Itanium C++ ABI. SPIR-V consumer will apply Itanium mangling during
+translation to LLVM IR as well.
+
+SPIR-V generator will support *scalar*, *vector* and *packed* (*packed* is for *Int4*, it's tentitive and a subject of removal,
+since clang now supports vectors of sub-byte integers) for the conversion builtin functions as LLVM IR input;
+*packed* format is translated to a *vector*. Meanwhile SPIR-V consumer
+will never pack a *vector* back to *packed* format.
+
+Following list of function calls will be handled by the translator:
+Translated to *FConvert*
```
svenvh:
```suggestion
Translated to *OpFConvert*
```

docs/FP8RepresentationInLLVM.rst
MrSidims: Yeah, it's basically a trimmed version of an internal spec I drafted some time ago and I forgot to update the names. Now they are just E4M3 and E5M2.

docs/FP8RepresentationInLLVM.rst
MrSidims: Thanks! Applied this and other comments

MrSidims:
@Keenuts @michalpaszkowski please also take a look as I plan to introduce the extension to the SPIR-V backend as well, and contact for the frontends should be identical for both SPIR-V generators.

Remaining TODOs:
1. Way to make it working with cooperative matrices. Currently internally I'm using bitmasks on the place of cooperative matrix operands, but it barely acceptable solution as some might claim the same bits for different purposes;
2. Way to represent **SaturatedToLargestFloat8NormalConversionEXT**. There are 2 considirations:
a. (used in internal implementation) a new builtin-like function for clamp conversions - quite straight forward and robust solution;
b. spirv.Decorations metadata attached to a one of the appropriate conversion buitins described in this PR - it will work out of a box for the translator, but in general relying semantics on metadata attached to an instruction is not a great design, though here it should work, as it would be attacked on an external function call and middle end transformations should not touch such call.

svenvh (approved):

PR #3238, opened 2025-06-16
Support casting of `OpConstantNull` to `SPIRVConstant`

vmaksimo:
Current design does not imply inheritance between `SPIRVConstantNull` and `SPIRVConstant` classes, however it is possible that we do the `static_cast` of `SPIRVValue` to `SPIRVConstant` for null constant. Return zero-value in that case.

### DIFF
--- a/lib/SPIRV/libSPIRV/SPIRVValue.h
+++ b/lib/SPIRV/libSPIRV/SPIRVValue.h
@@ -180,6 +180,8 @@ protected:
 
   // Common method for getting values of size less or equal to 64 bits.
   template <typename T> T getValue() const {
+    if (OpCode == OpConstantNull)
+      return 0;
     constexpr auto ValueSize = static_cast<unsigned>(sizeof(T));
     assert((ValueSize <= 8) && "Incorrect result type of requested value");
     T TheValue{};
--- a/lib/SPIRV/libSPIRV/SPIRVValue.h
+++ b/lib/SPIRV/libSPIRV/SPIRVValue.h
@@ -180,8 +180,11 @@ protected:
 
   // Common method for getting values of size less or equal to 64 bits.
   template <typename T> T getValue() const {
-    if (OpCode == OpConstantNull)
+    if (OpCode == OpConstantNull) {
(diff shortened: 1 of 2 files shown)


lib/SPIRV/libSPIRV/SPIRVValue.h
```diff
@@ -180,6 +180,8 @@ template <spv::Op OC> class SPIRVConstantBase : public SPIRVValue {
 
   // Common method for getting values of size less or equal to 64 bits.
   template <typename T> T getValue() const {
+    if (OpCode == OpConstantNull)
+      return 0;
```
Naghasan: nit

```suggestion
    if (OpCode == OpConstantNull) {
      assert(std::is_arithmetic_v<T> && "OpConstantNull has no defined value for this type");
      return 0;
    }
```

MrSidims: I actually have another Q. In which contexts getValue is called? Can it be called to get value of an object of Event type or a Coop Matrix or any composite? If yes note, that ConstantNull is defined as:
```
Result Type must be one of the following types:
- Scalar or vector Boolean type
- Scalar or vector integer type
- Scalar or vector floating-point type
- Pointer type
- Event type
- Device side event type
- Reservation id type
- Queue type
- Composite type
```
hence here we should:
a. not add the proposed assetion;
b. return not null, but a value constructed by the T's default c'tor.

MrSidims: Actually, I got the answer: the method is protected and used only by:
```
  uint64_t getZExtIntValue() const { return getValue<uint64_t>(); }
  float getFloatValue() const { return getValue<float>(); }
  double getDoubleValue() const { return getValue<double>(); }
```
hence we can even make the proposed by Viktor assert to be a static_assert

Naghasan: > Actually, I got the answer: the method is protected and used only by:

yeah, that's why I suggested the nit, I suggested an assert because the function is all constants not just null constant.

PR #3265, opened 2025-07-16
Update LLVM version from 21 to 22

svenvh:


svenvh:
Builds succeed now, failures are due to #3276.  Merging.

PR #3266, opened 2025-07-16
Add Check for WordCount Against SPIRV File Size

YixingZhang007:
A check is inserted in `parseAndCreateSPIRVEntry` function to ensure the `WordCount` for the current SPIR-V instruction is not smaller than the remaining input stream size. If the stream does not contain enough data for the specified `WordCount`, an SPIRV error of `InvalidWordCount` is raised.


lib/SPIRV/libSPIRV/SPIRVModule.cpp
```diff
@@ -2379,6 +2379,42 @@ void SPIRVModuleImpl::addUnknownStructField(SPIRVTypeStruct *Struct, unsigned I,
...
+  std::streamoff RemainingBytes = 0;
+
+#ifdef _SPIRV_SUPPORT_TEXT_FMT
+  if (SPIRVUseTextFormat) {
+    std::string Word;
+    while (IS >> Word) {
+      RemainingBytes += sizeof(SPIRVWord);
+    }
+  } else
+#endif
```
MrSidims: This is probably not needed. Unlike binary mode - here this check is quite expensive. And while text mode is for debugging and not productization - such security checks might be redundant. So I'd recommend to do an early exit from the function for text format.

YixingZhang007: Thanks for the suggestion! I have made the changes :)

svenvh (approved):

MrSidims:
Restarting CI

lib/SPIRV/libSPIRV/SPIRVModule.cpp
```diff
@@ -2385,6 +2385,23 @@ SPIRVEntry *parseAndCreateSPIRVEntry(SPIRVWord &WordCount, Op &OpCode,
   if (WordCount == 0 || OpCode == OpNop) {
     return nullptr;
   }
+  if (!SPIRVUseTextFormat) {
```
MrSidims, line 2388: I though this should be fixing clang-tidy error, isn't it the case?
```
static void validateWordCount(SPIRVModuleImpl &M, std::istream &IS,
                              SPIRVWord WordCount) {
/* */
}
```

YixingZhang007: We are currently enforcing both `llvm-prefer-static-over-anonymous-namespace` and `misc-use-anonymous-namespaces` flags for `clang-tidy`. If the function is declared as `static`, it failed with the flag `misc-use-anonymous-namespaces`. And if the function is placed in a namespace, it failed with the flag `llvm-prefer-static-over-anonymous-namespace`.

svenvh: I've opened https://github.com/KhronosGroup/SPIRV-LLVM-Translator/pull/3286 to address this.

YixingZhang007: Thanks for addressing the issue!  I've opened a PR https://github.com/KhronosGroup/SPIRV-LLVM-Translator/pull/3289 to refactor the WordCount validation introduced in this PR into a separate static function.

PR #3268, opened 2025-07-18
Fix potential constructor ambiguity in BuiltinCallMutator::ValueTypePair

jaladreips:
BuiltinCallMutator::ValueTypePair defines its own constructor with arguments (llvm::Value *V, llvm::Type *T)
BuiltinCallMutator::ValueTypePair also uses std::pair constructor with arguments (llvm::Value *&V, llvm::Type *&T)

This causes an ambiguity that confuses GCC when compiling for C++20.
Issue repro on goldbolt: https://godbolt.org/z/MTnvx6E6T

There are multiple possible fixes. I think the one with the least impact is to change BuiltinCallMutator::ValueTypePair constructor signature to conform with std::pair constructor signature.
Fixed version: https://godbolt.org/z/49WzvjK63

Other possibility is to remove the constructors that are redundant. But maybe they are very important.

### DIFF
--- a/lib/SPIRV/SPIRVBuiltinHelper.h
+++ b/lib/SPIRV/SPIRVBuiltinHelper.h
@@ -139,7 +139,6 @@ public:
              "Must specify a pointer element type if value is a pointer.");
     }
     ValueTypePair(std::pair<llvm::Value *, llvm::Type *> P) : pair(P) {}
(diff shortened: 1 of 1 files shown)


svenvh (approved):
> I deleted the construtors and all tests failed. So, I guess they are used

My suggestion was to only remove the offending constructor, which you have done in the latest patch. LGTM, thanks!

PR #3274, opened 2025-07-22
Fix build when linking against SPIRVLib with SPIR-V backend support enabled

sarnex:
When building with SPIR-V backend support, naturally we need to link against the SPIR-V backend. Currently this is done as part of the tool (`llvm-spirv`)'s CMake dependencies, instead of `SPIRVLib` which is the actual CMake target that requires the SPIR-V backend (`SPIRVTranslateModule` call inside `SPIRVWriter.cpp`).

This causes a linker error if someone depends on `SPIRVLib` directly and doesn't depend on `llvm-spirv`.

Move the dependency from the tool to the lib.


sarnex:
Thanks Sven! CI gets further now but fails on what seem to be unrelated LIT fails again

PR #3277, opened 2025-07-23
Fix LoopMerge Instruction Placement

pvelesko:
The SPIR-V specification requires that OpLoopMerge instructions must be placed as the second-to-last instruction in their basic block, immediately preceding the branch instruction (OpBranch or OpBranchConditional). The original implementation was not ensuring this correct placement, which could lead to SPIR-V validation failures.

1. **Added validation logic** to check if the basic block has a proper terminating instruction (OpBranch or OpBranchConditional)
2. **Conditional placement**:
- If a proper terminator exists → Place the LoopMerge instruction immediately before it
- If no proper terminator exists → Add the instruction at the end of the block
1. **Added a test case** (test/OpLoopMerge_loopMerge.ll) that:
- Verifies SPIR-V validation passes with spirv-val
- Checks that LoopMerge appears before BranchConditional in the text output


test/OpLoopMerge_loopMerge.ll
```diff
@@ -0,0 +1,10 @@
+; Test case for OpLoopMerge instruction placement validation.
+; This test verifies that OpLoopMerge instructions are properly placed as the second-to-last
+; instruction in their basic block, immediately preceding the branch instruction.
+
+; RUN: llvm-spirv %S/loopMerge.bc -o %t.spv
```
MrSidims: Please move .bc file content to the text and call llvm-as in the test before the translation.

test/OpLoopMerge_loopMerge.ll
```diff
@@ -0,0 +1,10 @@
...
+; RUN: spirv-val %t.spv
+
+; RUN: llvm-spirv --to-text %t.spv -o - | FileCheck %s --check-prefix=CHECK-SPIRV-TEXT
+; CHECK-SPIRV-TEXT: LoopMerge
+; CHECK-SPIRV-TEXT: BranchConditional
```
MrSidims:
```suggestion
; CHECK-SPIRV-TEXT: LoopMerge
; CHECK-SPIRV-TEXT-NEXT: BranchConditional 
```

lib/SPIRV/libSPIRV/SPIRVModule.cpp
```diff
@@ -1855,20 +1855,42 @@ SPIRVInstruction *SPIRVModuleImpl::addSelectionMergeInst(
...
+        new SPIRVLoopMerge(MergeBlock, ContinueTarget, LoopControl,
+                           LoopControlParameters, BB),
+        BB, TermInst);
+  } else {
+    // If there's no proper terminator, add at the end
+    return addInstruction(
+        new SPIRVLoopMerge(MergeBlock, ContinueTarget, LoopControl,
+                           LoopControlParameters, BB),
+        BB);
+  }
```
MrSidims: Probably clang-tidy will recommend to remove else as previous if terminates with return.

lib/SPIRV/libSPIRV/SPIRVModule.cpp
```diff
@@ -1855,20 +1855,42 @@ SPIRVInstruction *SPIRVModuleImpl::addSelectionMergeInst(
 SPIRVInstruction *SPIRVModuleImpl::addLoopMergeInst(
     SPIRVId MergeBlock, SPIRVId ContinueTarget, SPIRVWord LoopControl,
     std::vector<SPIRVWord> LoopControlParameters, SPIRVBasicBlock *BB) {
-  return addInstruction(
-      new SPIRVLoopMerge(MergeBlock, ContinueTarget, LoopControl,
-                         LoopControlParameters, BB),
-      BB, const_cast<SPIRVInstruction *>(BB->getTerminateInstr()));
+  SPIRVInstruction *TermInst = const_cast<SPIRVInstruction *>(BB->getTerminateInstr());
+  // OpLoopMerge must be the second-to-last instruction in the block,
+  // immediately preceding the branch instruction (OpBranch or OpBranchConditional)
```
maarquitos14:
```suggestion
  // immediately preceding the branch instruction (OpBranch or OpBranchConditional).
```

lib/SPIRV/libSPIRV/SPIRVModule.cpp
```diff
@@ -1855,20 +1855,42 @@ SPIRVInstruction *SPIRVModuleImpl::addSelectionMergeInst(
...
+  if (TermInst && (TermInst->getOpCode() == OpBranch || 
+                   TermInst->getOpCode() == OpBranchConditional)) {
+    return addInstruction(
+        new SPIRVLoopMerge(MergeBlock, ContinueTarget, LoopControl,
+                           LoopControlParameters, BB),
+        BB, TermInst);
+  } else {
+    // If there's no proper terminator, add at the end
```
maarquitos14:
```suggestion
    // If there's no proper terminator, add at the end.
```

maarquitos14: Also, is this correct? You mentioned that the spec requires this instruction to immediately precede either an OpBranch or an OpBranchConditional. My understanding is that, if none of these exist, this instruction could not be placed. Maybe we should assert here instead?

lib/SPIRV/libSPIRV/SPIRVModule.cpp
```diff
@@ -1855,20 +1855,42 @@ SPIRVInstruction *SPIRVModuleImpl::addSelectionMergeInst(
...
     SPIRVWord LoopControl, std::vector<SPIRVWord> LoopControlParameters,
     SPIRVBasicBlock *BB) {
   addCapability(CapabilityUnstructuredLoopControlsINTEL);
   addExtension(ExtensionID::SPV_INTEL_unstructured_loop_controls);
-  return addInstruction(
-      new SPIRVLoopControlINTEL(LoopControl, LoopControlParameters, BB), BB,
-      const_cast<SPIRVInstruction *>(BB->getTerminateInstr()));
+  SPIRVInstruction *TermInst = const_cast<SPIRVInstruction *>(BB->getTerminateInstr());
+  // OpLoopControlINTEL must be the second-to-last instruction in the block,
+  // immediately preceding the branch instruction (OpBranch or OpBranchConditional)
```
maarquitos14:
```suggestion
  // immediately preceding the branch instruction (OpBranch or OpBranchConditional).
```

lib/SPIRV/libSPIRV/SPIRVModule.cpp
```diff
@@ -1855,20 +1855,42 @@ SPIRVInstruction *SPIRVModuleImpl::addSelectionMergeInst(
...
+  if (TermInst && (TermInst->getOpCode() == OpBranch || 
+                   TermInst->getOpCode() == OpBranchConditional)) {
+    return addInstruction(
+        new SPIRVLoopControlINTEL(LoopControl, LoopControlParameters, BB), BB,
+        TermInst);
+  } else {
+    // If there's no proper terminator, add at the end
```
maarquitos14:
```suggestion
    // If there's no proper terminator, add at the end.
```

maarquitos14: Same than earlier. Is this compliant with the spec? Should we assert/report error instead?

test/OpLoopMerge_loopMerge.ll
maarquitos14: +1.

test/OpLoopMerge_loopMerge.ll
```diff
@@ -0,0 +1,10 @@
+; Test case for OpLoopMerge instruction placement validation.
+; This test verifies that OpLoopMerge instructions are properly placed as the second-to-last
+; instruction in their basic block, immediately preceding the branch instruction.
+
+; RUN: llvm-spirv %S/loopMerge.bc -o %t.spv
+; RUN: spirv-val %t.spv
+
+; RUN: llvm-spirv --to-text %t.spv -o - | FileCheck %s --check-prefix=CHECK-SPIRV-TEXT
+; CHECK-SPIRV-TEXT: LoopMerge
+; CHECK-SPIRV-TEXT-NEXT: BranchConditional
```
YuriPlyakhin: Please, add newline at end of file.

PR #3278, opened 2025-07-23
Adjust Lifetime intrinsic translation after 92c55a315eab

MrSidims:
This patch removes bitcast introduced in TypeScavenger that was previously served as W/A to solve a discrepancy between LLVM's and SPIR-V's lifetime inst definitions.

### DIFF
--- a/lib/SPIRV/SPIRVReader.cpp
+++ b/lib/SPIRV/SPIRVReader.cpp
@@ -253,6 +253,16 @@ IntrinsicInst *SPIRVToLLVM::getLifetimeStartIntrinsic(Instruction *I) {
   auto *II = dyn_cast<IntrinsicInst>(I);
   if (II && II->getIntrinsicID() == Intrinsic::lifetime_start)
     return II;
+  // Bitcast might be inserted during translation of OpLifetimeStart
+  auto *BC = dyn_cast<BitCastInst>(I);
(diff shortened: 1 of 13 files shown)


MrSidims:
@svenvh @vmaksimo please take a look. Note, it's incomplete patch as I also need to fix an odd translation of allocas wrapped in lifetime intrinsics, when allocation happens in generic AS - I'll visit it later this day.

The reason I'm opening PR right away is because it might be a bit not backward compatible because generated SPIR-V changes and the patch requires backports to release branches (the reader part). SPIR-V generation changes due to differences between SPIR-V and LLVM way to deal with Size operand, see SPIR-V definition:
```
Size is an unsigned 32-bit integer. Size must be 0 if Pointer is a
pointer to a non-void type or the Addresses capability is not declared.
If Size is non-zero, it is the number of bytes of memory whose lifetime
is starting.
```

meanwhile LLVM's is:
`The first argument is a constant integer representing the size of the object, or -1 if it is variable sized. The second argument is a pointer to an alloca instruction.`

svenvh:
Thanks for looking into this!

> My idea would be to merge the PR and see if something breaks (with following conditional partial of the PR), but may be you have another thoughts.

Sounds fine to me.

svenvh (approved):

PR #3281, opened 2025-07-30
Translate llvm.minimumnum/maximumnum intrinsics to OpenCL extended inst fmin/fmax

wenju-he:
The intrinsics were added in https://github.com/llvm/llvm-project/commit/89881480 and their behavior matches with OpenCL fmin/fmax built-in. This PR fixes llvm-spirv error when compiling __builtin_elementwise_maximumnum for libclc spirv target.

### DIFF
--- a/lib/SPIRV/SPIRVWriter.cpp
+++ b/lib/SPIRV/SPIRVWriter.cpp
@@ -3933,10 +3933,12 @@ bool LLVMToSPIRVBase::isKnownIntrinsic(Intrinsic::ID Id) {
   case Intrinsic::log:
   case Intrinsic::log10:
   case Intrinsic::log2:
+  case Intrinsic::maximumnum:
   case Intrinsic::maximum:
   case Intrinsic::maxnum:
   case Intrinsic::smax:
   case Intrinsic::umax:
+  case Intrinsic::minimumnum:
   case Intrinsic::minimum:
   case Intrinsic::minnum:
   case Intrinsic::smin:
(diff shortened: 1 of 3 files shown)


lib/SPIRV/SPIRVWriter.cpp
```diff
@@ -4064,12 +4066,12 @@ static SPIRVWord getBuiltinIdForIntrinsic(Intrinsic::ID IID) {
     return OpenCLLIB::Log10;
   case Intrinsic::log2:
     return OpenCLLIB::Log2;
+  case Intrinsic::maximumnum:
   case Intrinsic::maximum:
-    return OpenCLLIB::Fmax;
   case Intrinsic::maxnum:
     return OpenCLLIB::Fmax;
+  case Intrinsic::minimumnum:
   case Intrinsic::minimum:
```
svenvh, line 4074: Revisiting this, it's actually incorrect to map `minimum` to `OpenCLLIB::Fmin` due to differences in NaN behaviour:

`llvm.minimum`:
> If either operand is a NaN, returns NaN.

`fmin`:
> If one operand is a NaN, fmin returns the other argument. If both arguments are NaNs, fmin returns a NaN. 

I've raised #3282 for this.

`llvm.minimumnum` follows `fmin`'s NaN behaviour, so this change looks good to me.

svenvh (approved):

lib/SPIRV/SPIRVWriter.cpp
wenju-he: > Revisiting this, it's actually incorrect to map `minimum` to `OpenCLLIB::Fmin` due to differences in NaN behaviour:

yes, both `minimum` and `maxnum` have the difference in NaN behaviour compared to OpenCL fmax/fmin

PR #3284, opened 2025-07-31
Move instead of copy.

maarquitos14:


maarquitos14:
@MrSidims @svenvh please take a look :)

svenvh (approved):

PR #3290, opened 2025-08-05
Implement SPV_INTEL_bfloat16_arithmetic

MrSidims:
The extension relaxes rules for bf16 type allowing to use it in some arithmetic operations.

Spec is available here:
https://github.com/intel/llvm/pull/18352

Co-authered by: Michael Aziz <michael.aziz@intel.com>


MrSidims:
@maarquitos14 @YixingZhang007 please take a look

lib/SPIRV/SPIRVWriter.cpp
```diff
@@ -5487,6 +5501,11 @@ SPIRVValue *LLVMToSPIRVBase::transDirectCallInst(CallInst *CI,
   SmallVector<std::string, 2> Dec;
   if (isBuiltinTransToExtInst(CI->getCalledFunction(), &ExtSetKind, &ExtOp,
                               &Dec)) {
+    if (const auto *FirstArg = F->getArg(0)) {
+      const auto *Type = FirstArg->getType();
```
maarquitos14, line 5505:
```suggestion
      const Type*Ty = FirstArg->getType();
```

Just a nit, but I prefer the type to be explicit as much as possible. Feel free to ignore.

lib/SPIRV/SPIRVWriter.cpp
```diff
@@ -5487,6 +5501,11 @@ SPIRVValue *LLVMToSPIRVBase::transDirectCallInst(CallInst *CI,
   SmallVector<std::string, 2> Dec;
   if (isBuiltinTransToExtInst(CI->getCalledFunction(), &ExtSetKind, &ExtOp,
                               &Dec)) {
+    if (const auto *FirstArg = F->getArg(0)) {
```
maarquitos14, line 5504:
```suggestion
    if (const Argument *FirstArg = F->getArg(0)) {
```
Just a nit, but I prefer the type to be explicit as much as possible. Feel free to ignore.

lib/SPIRV/SPIRVUtil.cpp
```diff
@@ -525,6 +525,11 @@ ParamType lastFuncParamType(StringRef MangledName) {
   char Mangled = Copy.back();
   std::string Mangled2 = Copy.substr(Copy.size() - 2);
 
+  std::string Mangled6 = Copy.substr(Copy.size() - 6);
+  if (Mangled6 == "__bf16") {
```
YixingZhang007, line 529: We may consider combining this check with the `Mangled` and `Mangled2` check below, since all three return `ParamType::FLOAT`.

lib/SPIRV/SPIRVUtil.cpp
```diff
@@ -525,6 +525,11 @@ ParamType lastFuncParamType(StringRef MangledName) {
...
+    return ParamType::FLOAT;
+  }
+
   if (isMangledTypeFP(Mangled) || isMangledTypeHalf(Mangled2)) {
```
YixingZhang007, line 533: I also noticed that `isMangledTypeFP` and `isMangledTypeHalf` are only used once here. We could consider combining these two functions along with the Mangled6 check into a single helper function for simplicity.

lib/SPIRV/SPIRVUtil.cpp
MrSidims: Let me address this in a separate PR

lib/SPIRV/SPIRVWriter.cpp
MrSidims: Lets ignore it for now.

PR #3292, opened 2025-08-06
Addition of OpSwitchBranches.ll, OpSwitchChar.ll, OpSwitchUnreachable.ll, Two_OpSwitch_same_register.ll and switch-range-check.ll

aadeshps-mcw:
--Test ported from branching subdirectory of llvm-project

### DIFF
--- a/test/branching/OpSwitchChar.ll
+++ b/test/branching/OpSwitchChar.ll
@@ -1,0 +1,57 @@
+;; __kernel void test_switch(__global int* res, uchar val)
+;; {
+;;   switch(val)
+;;   {
+;;   case 0:
+;;     *res = 1;
+;;     break;
+;;   case 1:
+;;     *res = 2;
+;;     break;
+;;   case 2:
+;;     *res = 3;
+;;     break;
+;;   }
+;; }
+
+; RUN: llvm-as < %s -o %t.bc
+; RUN: llvm-spirv %t.bc -o %t.spv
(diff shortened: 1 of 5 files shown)


test/branching/OpSwitchChar.ll
```diff
@@ -0,0 +1,57 @@
+;; __kernel void test_switch(__global int* res, uchar val)
```
MrSidims: Is it different from https://github.com/KhronosGroup/SPIRV-LLVM-Translator/blob/main/test/transcoding/OpSwitchChar.ll ?

MrSidims (commented):
Please double check, if these tests already present in the repository. If they (or similar) tests exist I'd suggest not to duplicate them. I don't mind to scatter them to meaningful directory(ies) like /branching/.

test/branching/OpSwitchChar.ll
aadeshps-mcw: I noticed the same thing when I went through it, and I’ll keep in mind to reduce duplication. Thanks a lot.
Also, should I move the other OpSwitchChar.ll file from the transcoding subdirectory into the branching subdirectory for better organization?

MrSidims: @aadeshps-mcw if we are adding dedicated tests from branching - please go ahead with moving 'char' version to the sub-directory.

MrSidims (approved):

Issue #3287, opened 2025-08-04
SPIRVReader: reverse-translated argument types differ from clang output since opaque pointers


svenvh:
The LLVM argument type for any of the OpenCL "Other Built-in Data Types" (such as image types) has changed since the transition to opaque pointers.  This only affects the OpenCL target env modes of llvm-spirv.

For example, for the following OpenCL C kernel:
```
kernel void foo(image2d_t img, global int *out) {
  *out = get_image_channel_data_type(img);
}
```
`clang -target spir` currently [generates](https://compiler-explorer.com/z/cxzaxnvoc):
```
call spir_func i32 @_Z27get_image_channel_data_type14ocl_image2d_ro(target("spirv.Image", void, 1, 0, 0, 0, 0, 0, 0) %img)
```

Generating SPIR-V from the above kernel using `clang -c --target=spirv64` and taking the SPIR-V through `llvm-spirv -r` gives:
```
call spir_func i32 @_Z27get_image_channel_data_type14ocl_image2d_ro(ptr addrspace(1) %img)
```

That is, the mangled function name remains the same, but the image argument has become a plain pointer instead of a `TargetExtType`.

My proposal would be to align to clang as closely as possible, so for this case we'd ideally generate `TargetExtType`s instead of pointers.

PR #3293, opened 2025-08-06
SPIRVReader: make image and sampler types match clang output

svenvh:
For e.g. a `foo(image2d_t %img)` function type, clang produces

    @foo(target("spirv.Image", void, 1, 0, 0, 0, 0, 0, 0) %img)

but after OpenCL -> SPIR-V translation, the SPIR-V reader would translate it into

    @foo(ptr addrspace(1) %img)

Preserve image argument types as `TargetExtType`s when translating to an OpenCL target environment.  `TargetExtType`s are already emitted for SPIR-V friendly IR.

Make sure not to mangle the `TargetExtType` argument in the usual way, but continue using the special type mangling emitted by clang (e.g. `ocl_image1d_rw`).

Contributes to https://github.com/KhronosGroup/SPIRV-LLVM-Translator/issues/3287


lib/SPIRV/SPIRVInternal.h
```diff
@@ -512,6 +512,7 @@ class BuiltinFuncMangleInfo {
   virtual void init(StringRef UniqUnmangledName) {
     UnmangledName = UniqUnmangledName.str();
   }
+  virtual bool isOpenCL() const { return false; }
```
MrSidims, line 515: Do we need this method? Per the bug description: in SPIR-V friendly LLVM IR we always get target extension type in the output, while for OpenCL we get opaque pointer (which is wrong). Since we are aligning these 2 outputs (and since they are the only possible outputs for the translator) - then having this method and following flags surprises me.

svenvh: Yes, we will emit TargetExtTypes for both SPIR-V friendly LLVM IR and OpenCL-flavoured LLVM IR.  But the mangling for both cases will be different, so this member function distinguishes between how argument TargetExtTypes should be mangled.

MrSidims (approved):

PR #3303, opened 2025-08-08
[SPIRV] Porting tests from llvm-project

EbinJose2002:
Add tests for atomic exchange, type emission, bitcasts, and block reordering


test/bitcast.ll
```diff
@@ -0,0 +1,38 @@
+; RUN: llvm-as %s -o %t.bc
```
MrSidims: I agree, that more formal unit tests is something, what this repository is missing. But I also believe, that we should try to avoid duplication of tests, where we can. With adding this test file, should be remove https://github.com/KhronosGroup/SPIRV-LLVM-Translator/blob/main/test/transcoding/bitcast.ll ?

EbinJose2002: [bitcast](https://github.com/KhronosGroup/SPIRV-LLVM-Translator/blob/main/test/transcoding/bitcast.ll) tests only the bitcast in the reverse translated IR. Maybe I can combine that with this test. Will that be okay?

MrSidims: Ah, I see now. I don't mind to leave it as is now. Probably better to have ported tests to test both forward and reverse translation, but I won't insist. @svenvh WDYT?

svenvh: Since we're looking at this now, I'd say it's nice to take the opportunity to combine them into one.  I agree we should try to avoid (too much) overlap between tests.

EbinJose2002: Thanks for the suggestion. I’ve updated the test to combine both cases into one and removed the overlap.

MrSidims:
Restarting CI with new HEAD

EbinJose2002:
> please add checks for reverse translation

I have added reverse translation checks for other tests . But I didnt add for test/block-ordering.ll as that seemed unnecessary. Is it okay?

PR #3305, opened 2025-08-10
allow casting nullptr to any addrspace

simeonschaub:
This allows translation of code like the following, which previously errored:

```llvm
target datalayout = "e-i64:64-v16:16-v24:32-v32:32-v48:64-v96:128-v192:256-v256:256-v512:512-v1024:1024-G1"

define spir_kernel void @bar(ptr addrspace(1) %arg) {
pass26:
  %expr = icmp eq ptr addrspace(1) addrspacecast (ptr null to ptr addrspace(1)), %arg
  ret void
}
```

It is important this works because the addrspacecast is needed for addrspaces where NULL doesn't correspond to a zero value. We previously ran into correctness errors on some platforms without this cast, ref ROCm/llvm-project#281

### DIFF
--- a/lib/SPIRV/SPIRVWriter.cpp
+++ b/lib/SPIRV/SPIRVWriter.cpp
@@ -1652,6 +1652,9 @@ SPIRVValue *LLVMToSPIRVBase::transUnaryInst(UnaryInstruction *U,
       } else {
         BOC = OpCrossWorkgroupCastToPtrINTEL;
       }
+    } else if (isa<ConstantPointerNull>(Cast->getPointerOperand())) {
+      SPIRVType *TransTy = transScavengedType(U);
+      return BM->addNullConstant(bcast<SPIRVTypePointer>(TransTy));
     } else {
       getErrorLog().checkError(
           SrcAddrSpace == SPIRAS_Generic, SPIRVEC_InvalidModule, U,
--- a/test/addrspacecast_null.ll
+++ b/test/addrspacecast_null.ll
@@ -1,0 +1,19 @@
+; RUN: llvm-as %s -o %t.bc
+; RUN: llvm-spirv %t.bc -o %t.spv
+; RUN: spirv-dis %t.spv | FileCheck %s
+
+; Test that addrspacecast of null pointer generates appropriate OpConstantNull
+; instruction in SPIR-V.
+
+; CHECK: %_ptr_CrossWorkgroup_uchar = OpTypePointer CrossWorkgroup %uchar
+; CHECK: %[[NULL:[0-9]+]] = OpConstantNull %_ptr_CrossWorkgroup_uchar
+; CHECK: OpPtrEqual %bool %[[NULL]]
+
+target datalayout = "e-i64:64-v16:16-v24:32-v32:32-v48:64-v96:128-v192:256-v256:256-v512:512-v1024:1024-G1"
+target triple = "spir64-unknown-unknown"
+
+define spir_kernel void @bar(ptr addrspace(1) %arg) {
+pass26:
+  %expr = icmp eq ptr addrspace(1) addrspacecast (ptr null to ptr addrspace(1)), %arg
+  ret void
+}


lib/SPIRV/SPIRVWriter.cpp
```diff
@@ -1652,6 +1652,9 @@ SPIRVValue *LLVMToSPIRVBase::transUnaryInst(UnaryInstruction *U,
       } else {
         BOC = OpCrossWorkgroupCastToPtrINTEL;
       }
+    } else if (isa<ConstantPointerNull>(Cast->getPointerOperand())) {
+      SPIRVType *TransTy = transScavengedType(U);
+      return BM->addNullConstant(bcast<SPIRVTypePointer>(TransTy));
```
MrSidims, line 1657: Oof, the grey area. I believe, that this change won't harm, as all the 'sane' versions of such cast would be resolved under `DestAddrSpace == SPIRAS_Generic` branch. Also 'UB' means 'UB' and we can do whatever we want if null or whatever pointer delivered from null is dereferenced. But would this solution indeed work for you without bringing hard-to-detect bugs? Per https://llvm.org/docs/LangRef.html#pointer-type :
`The semantics of non-zero address spaces are target-specific. Memory access through a non-dereferenceable pointer is undefined behavior in any address space. Pointers with the bit-value 0 are only assumed to be non-dereferenceable in address space 0`
so technically dereferencing ptr addrspace(5) null might be not-UB per LLVM.
Meanwhile SPIR-V spec is saying in https://registry.khronos.org/SPIR-V/specs/unified1/SPIRV.html#_null_pointers :
`A "null pointer" can be formed from an OpConstantNull instruction with a pointer result type. The resulting pointer value is abstract, and will not equal the pointer value formed from any declared object or access chain into a declared object. Behavior is undefined if a load or store through OpConstantNull is executed.`

so it's always UB and per my understanding just replacing AS cast with a pointer can potentially change the semantics. While having null in the appropriate address space right after the frontend would make us avoiding all possible misinterpretations.

Anyway, if we decide to go with the current solution, please add a test for it.

MrSidims: > While having null in the appropriate address space right after the frontend would make us avoiding all possible misinterpretations.

Mmm, actually, how? if in SPIR-V it's always UB...

simeonschaub: What would be your proposed solution? FWIW, `addrspacecast`s of `null` are actually tested in https://github.com/llvm/llvm-project/blob/main/llvm/test/Transforms/Attributor/AMDGPU/do-not-replace-addrspacecast-with-constantpointernull.ll, so what our frontend emits should be legal, no? What do you propose this should produce when lowering to SPIR-V?

MrSidims: > so what our frontend emits should be legal, no

From LLVM stand-point: yes, but I can't tell if it's legal for AMDGPU target, as I'm not an expert there.

> What do you propose this should produce when lowering to SPIR-V?

Tough question, but the very test you have sent makes me thinking, that replacing address space cast with just a pointer is invalid (or at least the sole purpose of the test is to check, that it is not happening). And commit message of https://github.com/llvm/llvm-project/commit/d2992423e336b06a3c75b6f99aed6dd73abc3112 suggests the same. So if replacing the address space cast with a pointer operand for AMDGPU is illegal earlier in the stack, why it should become legal during translation to SPIR-V?

simeonschaub: I thought the translation is legal exactly because

> The resulting pointer value is abstract, and will not equal the pointer value formed from any declared object or access chain into a declared object.

which is not true for llvm's `ConstantPointerNull` as I gather from the discussion in https://discourse.llvm.org/t/rfc-introduce-sentinel-pointer-value-to-datalayout and https://github.com/ROCm/llvm-project/issues/281

simeonschaub: This probably also means that when translating the other direction, `OpConstantNull` should actually become `addrspacecast (ptr null to ptr addrspace(AS))`

MrSidims: > This probably also means that when translating the other direction, OpConstantNull should actually become addrspacecast (ptr null to ptr addrspace(AS))

Better not. We can't tell, from where **OpConstantNull** was originated after the round trip. If you find it's necessary to make things working correctly on your end - we may do it under an option.

MrSidims (approved):
LGTM as dereferrencing null in SPIR-V is always UB, disregarding the storage class.

Lets wait a bit it to see, what other maintainers will say before merging.

PR #3307, opened 2025-08-12
Addition of alloca-load-store.ll, identity-function.ll, multiple-anonymous-functions.ll, trivial-function-definition.ll, trivial-function-with-call.ll

aadeshps-mcw:
--Test for function definitions and calls, ported from function subdirectory of llvm project.

### DIFF
--- a/test/function/alloca-load-store.ll
+++ b/test/function/alloca-load-store.ll
@@ -1,0 +1,64 @@
+; RUN: llvm-as < %s -o %t.bc
+; RUN: llvm-spirv %t.bc -o %t.spv
+; RUN: spirv-val %t.spv
+; RUN: llvm-spirv -to-text %t.spv -o - | FileCheck %s
+
+target datalayout = "e-i64:64-v16:16-v24:32-v32:32-v48:64-v96:128-v192:256-v256:256-v512:512-v1024:1024-n8:16:32:64"
+target triple = "spir64-unknown-unknown"
+
+; CHECK: Name [[#BAR_ID:]] "bar" 
+; CHECK: Name [[#FOO_ID:]] "foo" 
+; CHECK: Name [[#GOO_ID:]] "goo" 
+
+; CHECK: TypeInt [[#I32:]] 32 0 
+; CHECK: TypeFunction [[#FTY_BARFOO:]] [[#I32]] [[#I32]] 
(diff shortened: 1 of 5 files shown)


MrSidims (approved):
Yeah, these tests are good examples of functionality that is definitely tested implicitly elsewhere in the translator repository, yet IMHO it's better to have some dedicated simple tests.

test/function/alloca-load-store.ll
```diff
@@ -0,0 +1,77 @@
+; RUN: llvm-as %s -o %t.bc
```
MrSidims, line 1: actually, this very test look very familiar to store.ll test. I suggest to remove https://github.com/KhronosGroup/SPIRV-LLVM-Translator/blob/main/test/store.ll in this PR

aadeshps-mcw: Have removed the test.

test/function/alloca-load-store.ll
```diff
@@ -0,0 +1,77 @@
...
+ 
+; CHECK-SPIRV: TypeInt [[#I32:]] 32 0 
+; CHECK-SPIRV: TypeFunction [[#FTY_BARFOO:]] [[#I32]] [[#I32]] 
+; CHECK-SPIRV: TypePointer [[#PTR:]] [[#]] [[#I32]] 
+; CHECK-SPIRV: TypePointer [[#PTR_FUNC:]] [[#]] [[#]] 
+; CHECK-SPIRV: TypeFunction [[#FTY_GOO:]] [[#I32]] [[#I32]] [[#PTR_FUNC]] 
+
+define i32 @bar(i32 %a) {
+  %p = alloca i32
+  store i32 %a, i32* %p
```
MrSidims:
```suggestion
  store i32 %a, ptr %p
```
same for other pointers

MrSidims (approved):
Ultimately LGTM, please adjust pointers type

test/function/alloca-load-store.ll
aadeshps-mcw: Have made the modification

MrSidims: Can we make it for all pointers in the module and not only in this place?

aadeshps-mcw: Have made the changes

PR #3309, opened 2025-08-12
Porting 4 tests from llvm-project

EbinJose2002:
Added tests for linkage attributes, events, variable placement and array in struct.

### DIFF
--- a/test/OpVariable_order.ll
+++ b/test/OpVariable_order.ll
@@ -1,0 +1,26 @@
+; All OpVariable instructions in a function must be the first instructions in the first block
+
+; RUN: llvm-as %s -o %t.bc
+; RUN: llvm-spirv %t.bc -spirv-text -o %t.spt
+; RUN: FileCheck < %t.spt %s --check-prefix=CHECK-SPIRV
+; RUN: llvm-spirv %t.bc -o %t.spv
+; RUN: spirv-val %t.spv
+
+target datalayout = "e-p:32:32-i64:64-v16:16-v24:32-v32:32-v48:64-v96:128-v192:256-v256:256-v512:512-v1024:1024"
+target triple = "spir-unknown-unknown"
+
+; CHECK-SPIRV: Function [[#]] [[#]] [[#]] [[#]]
+; CHECK-SPIRV-EMPTY:
+; CHECK-SPIRV-NEXT: Label
+; CHECK-SPIRV-NEXT: Variable
+; CHECK-SPIRV-NEXT: Variable
+; CHECK-SPIRV: Return
+; CHECK-SPIRV: FunctionEnd
+
+define void @main() {
+entry:
+  %0 = alloca <2 x i32>, align 4
+  %1 = getelementptr <2 x i32>, ptr %0, i32 0, i32 0
+  %2 = alloca float, align 4
+  ret void
+}
--- a/test/event-wait-ptr-type.ll
+++ b/test/event-wait-ptr-type.ll
@@ -1,0 +1,30 @@
+; RUN: llvm-as %s -o %t.bc
+; RUN: llvm-spirv %t.bc -o - -spirv-text | FileCheck %s
+; RUN: llvm-spirv %t.bc -o %t.spv
+; RUN: spirv-val %t.spv
+
+target datalayout = "e-p:32:32-i64:64-v16:16-v24:32-v32:32-v48:64-v96:128-v192:256-v256:256-v512:512-v1024:1024-n8:16:32:64"
+target triple = "spir-unknown-unknown"
+
+; CHECK-DAG: TypeEvent [[#EventTy:]]
+; CHECK-DAG: TypeStruct [[#StructEventTy:]] [[#EventTy]]
+; CHECK-DAG: TypePointer [[#FunPtrStructEventTy:]] 7 [[#StructEventTy]]
+; CHECK-DAG: TypePointer [[#GenPtrEventTy:]] 8 [[#StructEventTy]]
+; CHECK-DAG: TypePointer [[#FunPtrEventTy:]] 8 [[#EventTy]]
+; CHECK: Function
+; CHECK: Variable [[#FunPtrStructEventTy]] [[#Var:]] 7
+; CHECK-NEXT:  PtrCastToGeneric [[#GenPtrEventTy]] [[#GenEvent:]] [[#Var]]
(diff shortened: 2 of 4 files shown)


EbinJose2002:
Thanks for the feedback. I agree with your approach. We’ll focus on porting trivial/basic tests that are missing in the translator, as well as any tests for untested functionality, and will be more careful about avoiding duplication with existing tests. We’ll also make sure to add reverse translation checks for new ported tests where applicable.
For this PR, I will remove test/const-array-in-struct.ll and add reverse translation checks to the relevant tests.

test/event-wait-ptr-type.ll
```diff
@@ -0,0 +1,33 @@
+; RUN: llvm-as %s -o %t.bc
+; RUN: llvm-spirv %t.bc -o - -spirv-text | FileCheck %s --check-prefix=CHECK-SPIRV
+; RUN: llvm-spirv %t.bc -o %t.spv
+; RUN: spirv-val %t.spv
+; RUN: llvm-spirv -r --spirv-target-env=SPV-IR %t.spv -o %t.rev.bc
+; RUN: llvm-dis < %t.rev.bc | FileCheck %s --check-prefix=CHECK-SPV-IR
```
MrSidims, line 6:
```suggestion
; RUN: llvm-spirv -r --spirv-target-env=SPV-IR %t.spv -o %t.rev.bc
; RUN: llvm-dis < %t.rev.bc | FileCheck %s --check-prefix=CHECK-SPV-IR
; RUN: llvm-spirv -r  %t.spv -o %t.rev.bc
; RUN: llvm-dis < %t.rev.bc | FileCheck %s --check-prefix=CHECK-OCL-IR
```

test/extern-weak-linkage.ll
```diff
@@ -0,0 +1,22 @@
...
+; CHECK-DAG: Decorate [[#AbsFun]] LinkageAttributes "abs" Export
+; CHECK-DAG: Decorate [[#ExternalFun]] LinkageAttributes "__devicelib_abs" Import
+
+define weak dso_local spir_func i32 @abs(i32 noundef %x) {
+entry:
+  %call = tail call spir_func i32 @__devicelib_abs(i32 noundef %x)
+  ret i32 %call
+}
+
+declare extern_weak dso_local spir_func i32 @__devicelib_abs(i32 noundef)
```
MrSidims, line 22: What do we have after reverse translation?

test/OpVariable_order.ll
```diff
@@ -0,0 +1,26 @@
+; All OpVariable instructions in a function must be the first instructions in the first block
```
MrSidims, line 1: Yeah, this test can be left without reverse translation part.

test/extern-weak-linkage.ll
EbinJose2002: > What do we have after reverse translation?


I ran the reverse translation on this file and the output was:

```
declare spir_func i32 @__devicelib_abs(i32) #0

; Function Attrs: nounwind
define spir_func i32 @abs(i32 %x) #0 {
entry:
  %call = call spir_func i32 @__devicelib_abs(i32 %x) #0
  ret i32 %call
}
```
The explicit weak linkage from the original is not present in the generated ll.

test/event-wait-ptr-type.ll
EbinJose2002: Is adding the RUN line for reverse translation sufficient here? When I ran the reverse translation without specifying -spirv-target-env, it didn’t reproduce the original call.

MrSidims: My suggestion to have both runs and have 2 dedicated checks for reverse translation.

By default translator emits LLVM IR that would clang produce for OpenCL aka we would see appropriate OpenCL builtin - that is what you see I assume. With SPV-IR it produces so called 'SPIR-V friendly LLVM IR', documented [here](https://github.com/KhronosGroup/SPIRV-LLVM-Translator/blob/main/docs/SPIRVRepresentationInLLVM.rst).

There are also ways to control OpenCL version for the output with -spirv-target-env, but it's too much for this test.

EbinJose2002: Thanks for clarifying. Yes, I was seeing the OpenCL builtin in the default reverse translation output, which matches what you described. I’ll update the test to include both runs and add two dedicated checks for reverse translation as suggested.

test/event-wait-ptr-type.ll
```diff
@@ -0,0 +1,37 @@
...
+; CHECK-SPIRV-DAG: TypePointer [[#FunPtrEventTy:]] 8 [[#EventTy]]
+; CHECK-SPIRV: Function
+; CHECK-SPIRV: Variable [[#FunPtrStructEventTy]] [[#Var:]] 7
+; CHECK-SPIRV-NEXT:  PtrCastToGeneric [[#GenPtrEventTy]] [[#GenEvent:]] [[#Var]]
+; CHECK-SPIRV-NEXT:  Bitcast [[#FunPtrEventTy]] [[#FunEvent:]] [[#GenEvent]]
+; CHECK-SPIRV-NEXT:  GroupWaitEvents [[#]] [[#]] [[#FunEvent]]
+
+; CHECK-SPV-IR: __spirv_GroupWaitEvents
+; CHECK-OCL-IR: spir_func void @_Z17wait_group_events
+;; The call @_Z17wait_group_events is not correctly mapped to SPIRV. It remains as functioncall in SPIRV generation.
```
svenvh: Is this an issue specific to the SPIRV-LLVM-Translator?  It might be worth marking this with a `TODO`.

svenvh (commented):
@MrSidims thanks for coming up with some criteria!  It's important to understand that adding tests comes with a (future) maintenance cost, so we should make sure that any new tests add value and don't overlap with existing tests.

test/event-wait-ptr-type.ll
EbinJose2002: > Is this an issue specific to the SPIRV-LLVM-Translator? It might be worth marking this with a `TODO`.

Yes, the same call @_Z17wait_group_events was correctly translated to OpGroupWaitEvents in llvm-project but it failed to do so in SPIRV-LLVM-Translator. So shall I mark it as TODO ?

svenvh: Good to know; yes it would be great if you could add a TODO there.

svenvh (approved):

PR #3315, opened 2025-08-13
[Backport to 15] Add support for SPV_INTEL_function_variants (#3246)

jdnk:
Original PR adding support for SPV_INTEL_function_variants: https://github.com/KhronosGroup/SPIRV-LLVM-Translator/pull/3246

Spec: https://github.com/intel/llvm/blob/sycl/sycl/doc/design/spirv-extensions/SPV_INTEL_function_variants.asciidoc


jdnk:
This required using `llvm::Optional` instead of `std::optional` due to C++14 used in the older version. Also fixed other misc issues caused by an erroneous merge. I kept it as separate commits to trace it more easily, but can squash if preferred.

MrSidims (approved):

PR #3317, opened 2025-08-14
Porting 3 tests from llvm-project

EbinJose2002:
Tests for pointer comparison and scalar integer and bitwise instructions

### DIFF
--- a/test/instructions/ptrcmp.ll
+++ b/test/instructions/ptrcmp.ll
@@ -1,0 +1,196 @@
+; RUN: llvm-as %s -o %t.bc
+; RUN: llvm-spirv %t.bc -o - -spirv-text | FileCheck %s --check-prefix=CHECK-SPIRV
+; RUN: llvm-spirv %t.bc -o %t.spv
+; RUN: spirv-val %t.spv
+; RUN: llvm-spirv -r %t.spv -o %t.rev.bc
+; RUN: llvm-dis < %t.rev.bc | FileCheck %s --check-prefix=CHECK-LLVM
+
+target datalayout = "e-p:32:32-i64:64-v16:16-v24:32-v32:32-v48:64-v96:128-v192:256-v256:256-v512:512-v1024:1024"
+target triple = "spir"
+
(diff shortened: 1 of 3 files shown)


test/instructions/scalar-bitwise-operations.ll
```diff
@@ -0,0 +1,125 @@
+; RUN: llvm-as %s -o %t.bc
```
MrSidims, line 1: I believe with this test added, we can remove ExtendBitBoolArg.ll

test/instructions/ptrcmp.ll
```diff
@@ -0,0 +1,196 @@
...
+; CHECK-SPIRV: FunctionEnd
+
+; CHECK-LLVM: define spir_func i1 @test_eq
+; CHECK-LLVM: %[[R:.*]] = icmp eq
+; CHECK-LLVM: ret i1 %[[R]]
+
+define i1 @test_eq(i16* %a, i16* %b) {
+  %r = icmp eq i16* %a, %b
+  ret i1 %r
+}
```
MrSidims: same suggestion for other tests
```suggestion
define i1 @test_eq(ptr %a, ptr %b) {
  %r = icmp eq ptr %a, %b
  ret i1 %r
}
```

test/instructions/scalar-bitwise-operations.ll
EbinJose2002: > I believe with this test added, we can remove ExtendBitBoolArg.ll

Yes. The new test is checking for even more bitwise operations and has checks for forward and reverse translation. So I think we can remove ExtendBitBoolArg.ll .

MrSidims: Please accept my apologies, I have misunderstood the purpose of test/ExtendBitBoolArg.ll - it actually checks, if 1-bit lfsr translation works correctly. So lets restore that test back. With the test restore the patch is LGTM.

MrSidims (approved):
LGTM with the test restored.

PR #3318, opened 2025-08-14
Added test files opencl_version_metadata.ll and progvar_prog_scope_uninit.ll

aadeshps-mcw:


test/opencl_version_metadata.ll
```diff
@@ -0,0 +1,48 @@
+; RUN: split-file %s %t
```
MrSidims, line 1: this test both supersets and subset of existing tests. For example we also have here already transcoding/spirvSourceCL2.0.ll and multi_md.ll (and may be more). If we are about to merge this PR - we should:
a. remove transcoding/spirvSourceCL2.0.ll 
b. check, what multi_md.ll does and either incorporate in this test or decide to leave multi_md.ll as a stand alone test

Again, there might be more tests

test/progvar_prog_scope_uninit.ll
```diff
@@ -0,0 +1,62 @@
+; RUN: llvm-as %s -o %t.bc
```
MrSidims, line 1: Not sure if I understand purpose of this test, is it for zeroinitialized GVs?

test/opencl_version_metadata.ll
aadeshps-mcw: I think i can maybe delete this test as the spirvSourceCLCXX.ll serves the purpose of this test.

PR #3319, opened 2025-08-14
Added test files for global-var-name-linkage.ll and keep-tracked-const.ll

aadeshps-mcw:
-Added test cases for global variable name linkage and const tracking from llvm-project.


test/keep-tracked-const.ll
```diff
@@ -0,0 +1,38 @@
+; This test case ensures that cleaning of temporary constants doesn't purge tracked ones.
```
MrSidims: Does it bring anything new to the tests from test/SpecConstants/ ?

aadeshps-mcw: I didn't see any tests having the check for reverse translation.
in test/SpecConstants, i could add check-llvm directives in bool-spirv-specconstant.ll and i can remove this extra file.

MrSidims: I see, lets keep your test then, will figure out what to do with bool-spirv-specconstant.ll later. Yet please move this test to test/SpecConstants/ dir

aadeshps-mcw: Sure, have done that in the latest commit.

PR #3342, opened 2025-09-01
Two tests Ported from the LLVM Project to Translator

SubashBoopathi:
Ported two tests from the LLVM project to the SPIRV-LLVM-Translator: icmp.ll validates integer comparison instructions  for scalar and vector types, and fp-vector-intrinsics.ll verifies floating-point vector intrinsics, ensuring accurate SPIR-V translation.

### DIFF
--- a/test/llvm-intrinsics/fp-vector-intrinsics.ll
+++ b/test/llvm-intrinsics/fp-vector-intrinsics.ll
@@ -1,0 +1,165 @@
+; RUN: llvm-as %s -o %t.bc
+; RUN: llvm-spirv %t.bc -spirv-text -o - | FileCheck %s
+; RUN: llvm-spirv %t.bc -o %t.spv
+; RUN: spirv-val %t.spv
+
+target datalayout = "e-i64:64-v16:16-v24:32-v32:32-v48:64-v96:128-v192:256-v256:256-v512:512-v1024:1024-n8:16:32:64"
+target triple = "spir64"
+
+declare <2 x half> @llvm.rint.v2f16(<2 x half>)
+declare <2 x half> @llvm.nearbyint.v2f16(<2 x half>)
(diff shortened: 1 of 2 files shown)


test/llvm-intrinsics/fp-vector-intrinsics.ll
```diff
@@ -0,0 +1,165 @@
+; RUN: llvm-as %s -o %t.bc
```
MrSidims, line 1: I do remember this test was failing, when I first saw the PR. What has changed since then? Force-pushes are making it impossible for me to find out.

SubashBoopathi: I updated the test by removing some extra CHECK directives that are irrelevant and now passes all RUN commands. I’ll avoid force-pushes or add clearer notes next time to make changes easier to follow.

MrSidims: Do I understand correctly, that reverse translation was failing? I'm not saying, that you should fix this, but I'm trying to understand, whether we have a bug anywhere and we just had an opportunity to catch and file it.

SubashBoopathi: @MrSidims  The issue wasn’t related to reverse translation. Initially, I planned to include checks for some additional intrinsics, but since those intrinsics already had tests elsewhere, I removed their checks from the test file. Unfortunately, my initial removal wasn’t complete, which caused the failures you saw earlier. The fix was simply cleaning up those leftover, unnecessary CHECK directives of the removed intrinsics tests.

MrSidims (approved):
Fine with me, @SubashBoopathi if it's possible, could you please check git reflog to get the removed CHECKs and submitting them as an issue to the translator, so we could verify, that them were wrong or there is something to fix in the translator?

svenvh (approved):

PR #2535, opened 2024-04-23
Add SPV_INTEL_bindless_images preview extension

MrSidims:
Specification: https://github.com/intel/llvm/blob/sycl/sycl/doc/design/spirv-extensions/SPV_INTEL_bindless_images.asciidoc

Co-author: He, Wenju <wenju.he@intel.com>

PR #3370, opened 2025-09-30
Implement SPV_INTEL_predicated_io extension

YixingZhang007:
This extension adds predicated load and store instructions. Predicated load performs load from memory if predicate is true; otherwise, it uses default_value as a result. Predicated store performs store of value to memory if predicate is true; otherwise, it does nothing.

Spec: https://github.com/intel/llvm/pull/20158

Signed-off-by: Zhang, Yixing<yixing.zhang@intel.com>


test/extensions/INTEL/SPV_INTEL_predicated_io/predicated_io_generic.ll
```diff
@@ -0,0 +1,76 @@
+; Generated with:
+; source.cl:
+; int __spirv_PredicatedLoadINTEL(const __global void* pointer, bool predicate, int default_value);
+; int __spirv_PredicatedLoadINTEL(const __global void* pointer, bool predicate, int default_value, int memory_operands);
+; void __spirv_PredicatedStoreINTEL(const __global void* pointer, int object, bool predicate);
+; void __spirv_PredicatedStoreINTEL(const __global void* pointer, int object, bool predicate, int memory_operands);
+;
+; void foo(const __global void* load_pointer, __global void* store_pointer, int default_value, int store_object, bool predicate) {
```
MrSidims: @YuriPlyakhin @bashbaug should we tell, that if _pointer_ is typed pointer, then it's element type should match the default value's type? If no - the example in the test is fine, but we also need to add an overload for the reversly-translated builtin. If yes, better to modify the example in the test a bit changing `void*` to `int*`. Currently it's allowed to have mismatched types:
```
Result Type is the type of the loaded object. It must be a scalar or vector of numerical-type.

Pointer is the pointer to load through. Its type must be a pointer.
```

**UPD**: actually, we should account for return type overloading in this case anyway, as we also have `SPV_KHR_untyped_pointers` extension and when it's enabled there might be declaration collisions. @YixingZhang007 if I'm not mistaken, here is how return value can be accounted in mangling: https://github.com/KhronosGroup/SPIRV-LLVM-Translator/pull/2535/files#diff-842ecc57cba566615d42bfbbb2571dd46eea03f273c6e07298fc9cb00a66ac86R2497

test/extensions/INTEL/SPV_INTEL_predicated_io/predicated_io_generic.ll
```diff
@@ -0,0 +1,76 @@
...
+; CHECK-SPIRV: FunctionParameter [[#BoolTy]] [[#Predicate:]]
+; CHECK-SPIRV: PredicatedLoadINTEL [[#Int32Ty]] [[#Result1:]] [[#LoadPtr]] [[#Predicate]] [[#DefaultVal]]
+; CHECK-SPIRV: PredicatedLoadINTEL [[#Int32Ty]] [[#Result2:]] [[#LoadPtr]] [[#Predicate]] [[#DefaultVal]] [[#Const0]]
+; CHECK-SPIRV: PredicatedStoreINTEL [[#StorePtr]] [[#StoreObj]] [[#Predicate]]
+; CHECK-SPIRV: PredicatedStoreINTEL [[#StorePtr]] [[#StoreObj]] [[#Predicate]] [[#Const0]]
+
+; CHECK-LLVM: call spir_func i32 @_Z27__spirv_PredicatedLoadINTELPU3AS1cbi(ptr addrspace(1) %{{.*}}, i1 %{{.*}}, i32 %{{.*}})
+; CHECK-LLVM: call spir_func i32 @_Z27__spirv_PredicatedLoadINTELPU3AS1cbii(ptr addrspace(1) %{{.*}}, i1 %{{.*}}, i32 %{{.*}}, i32 0)
+; CHECK-LLVM: call spir_func void @_Z28__spirv_PredicatedStoreINTELPU3AS1cib(ptr addrspace(1) %{{.*}}, i32 %{{.*}}, i1 %{{.*}})
+; CHECK-LLVM: call spir_func void @_Z28__spirv_PredicatedStoreINTELPU3AS1cibi(ptr addrspace(1) %{{.*}}, i32 %{{.*}}, i1 %{{.*}}, i32 0)
```
MrSidims: Note, as it is now mangling is changed from` void*` to `char*`. I do expect that if void* is replaced with `integer*` the result would be the same - we will have `char*` in the mangling. It will happen as currently there is no information in the module about the real type for load/store, hence llvm-spirv will default to `char*`. If Ben/Yuri confirm, that pointer should have matching element type to the default value, then we can do something better. Here is TypeScavenger routine in llvm-spirv, scavenging type from whatever placed possible, one of them are load and store memory instructions. We could add extra check along load/store checks for this very builtin to get the type of the pointer.

test/extensions/INTEL/SPV_INTEL_predicated_io/predicated_io_generic.ll
YuriPlyakhin: @MrSidims , @YixingZhang007 ,
From spec: `The type of Default Value must be the same as Result Type.`
Based on this, do we still need to take into account return type mangling, would not mangling of `Default Value` be enough? Where would collisions come from?

Regarding adding `if pointer is typed pointer, then it's element type should match the default value's type` - I think this would add more determinism to spec - less ambiguity. @bashbaug, what do you think?

bashbaug: > Regarding adding `if pointer is typed pointer, then it's element type should match the default value's type` - I think this would add more determinism to spec - less ambiguity.

Yes, I agree, if the pointer is a typed pointer then the pointer type should match the _Result Type_, and hence also the _Default Value_.

Strictly speaking, we might not even need to differentiate between typed pointers and untyped pointers in this spec, due to this section in the [untyped pointers spec](https://github.khronos.org/SPIRV-Registry/extensions/KHR/SPV_KHR_untyped_pointers.html):

> When memory accessed via instructions have a pointer operand with type [OpTypeUntypedPointerKHR](https://github.khronos.org/SPIRV-Registry/extensions/KHR/SPV_KHR_untyped_pointers.html#OpTypeUntypedPointerKHR) (e.g. OpLoad or atomic instructions), the interpreted data type is specified by the Result Type if it exists. The intepreted data type for instructions without a Result Type (e.g. OpStore) comes from the type of the operand of the object being stored.

YuriPlyakhin: good to know, thanks. Based on this, test should be updated to use `int*` instead of `void*`, if result type is `int`.

YixingZhang007: Thank you for the suggestions! I agree that it would be better to require the pointer's element type to match the default value's type. I have updated the tests to use `int*` instead of `void*`.

test/extensions/INTEL/SPV_INTEL_predicated_io/predicated_io_generic.ll
YixingZhang007: Thank you for pointing this out! After updating the pointer type in the test from void* to int*, the compiler can now correctly mangle the pointer type as `int` instead of `char` as shown below:
https://github.com/KhronosGroup/SPIRV-LLVM-Translator/blob/de004391da5a4db6cc4a4b191633a150ad1560f0/test/extensions/INTEL/SPV_INTEL_predicated_io/predicated_io_generic.ll#L47
Please feel free to let me know if there is any changes needed.

test/extensions/INTEL/SPV_INTEL_predicated_io/predicated_io_generic.ll
```diff
@@ -0,0 +1,75 @@
...
+; CHECK-SPIRV: FunctionParameter [[#IntPtrTy]] [[#StorePtr:]]
+; CHECK-SPIRV: FunctionParameter [[#Int32Ty]] [[#DefaultVal:]]
+; CHECK-SPIRV: FunctionParameter [[#Int32Ty]] [[#StoreObj:]]
+; CHECK-SPIRV: FunctionParameter [[#BoolTy]] [[#Predicate:]]
+; CHECK-SPIRV: PredicatedLoadINTEL [[#Int32Ty]] [[#Result1:]] [[#LoadPtr]] [[#Predicate]] [[#DefaultVal]]
+; CHECK-SPIRV: PredicatedLoadINTEL [[#Int32Ty]] [[#Result2:]] [[#LoadPtr]] [[#Predicate]] [[#DefaultVal]] [[#Const0]]
+; CHECK-SPIRV: PredicatedStoreINTEL [[#StorePtr]] [[#StoreObj]] [[#Predicate]]
+; CHECK-SPIRV: PredicatedStoreINTEL [[#StorePtr]] [[#StoreObj]] [[#Predicate]] [[#Const0]]
+
+; CHECK-LLVM: call spir_func i32 @_Z27__spirv_PredicatedLoadINTELPU3AS1ibi(ptr addrspace(1) %{{.*}}, i1 %{{.*}}, i32 %{{.*}})
```
MrSidims, line 47: Ah, so mangling switch was happening only for char (void) pointers. In this case I have no objections.

Issue #3371, opened 2025-09-30
questionable Memory Semantics for OpenCL C atomic functions


bashbaug:
For Khronos folks, this is partially discussed in internal issue 154.

The "Memory Semantics" that are generated for the OpenCL C 2.0 atomic functions are debatably incorrect.  Consider this OpenCL C kernel, which performs an [atomic store](https://registry.khronos.org/OpenCL/specs/3.0-unified/html/OpenCL_C.html#atomic_store) to an atomic variable in the global address space, performing a release operation:

```c
kernel void test(global atomic_int* gptr, local atomic_int* lptr) {
    atomic_store_explicit(gptr, 2, memory_order_release);
}
```

The [SPIR-V atomic instruction](https://registry.khronos.org/SPIR-V/specs/unified1/SPIRV.html#OpAtomicStore) that is generated for this kernel is something like:

```
OpAtomicStore %13 %uint_1 %uint_4 %uint_2
```

Here are the operands for this instruction:
1. `%13` is the pointer operand, from previous instructions.
2. `%uint_1` is the Scope for Memory, and represents Device scope.  This makes sense, since the default memory scope is `memory_scope_device`: "The functions that do not have memory_scope argument have the same semantics as the corresponding functions with the memory_scope argument set to memory_scope_device."
3. `%uint_4` is the Memory Semantics.  This represents the Release memory order, which is good, but strangely there are no affected memory spaces.  
4. `%uint_2` is the value that is stored, and is fine.

Should the SPIR-V LLVM Translator include affected memory spaces in the generated atomic instruction?  If so, how should it determine which memory spaces are affected?  If not, what should it mean to have an atomic memory order, but no affected memory spaces?

MrSidims:
I agree with the proposal you shared internally - use pointer's address space to determine, which memory semantic to use. In case if address space is generic and it's not possible to infer it at the point of the translation - we should use conservative mask CrossWorkgroupMemory | WorkgroupMemory | SubgroupMemory (whatever subgroup memory means).

MrSidims:
Created https://github.com/KhronosGroup/SPIRV-LLVM-Translator/pull/3375

PR #3375, opened 2025-10-04
Add memory semantics to atomic builtins

MrSidims:
It's deduced from a pointer's address space.
Resolves:
https://github.com/KhronosGroup/SPIRV-LLVM-Translator/issues/3371

### DIFF
--- a/lib/SPIRV/OCLToSPIRV.cpp
+++ b/lib/SPIRV/OCLToSPIRV.cpp
@@ -42,15 +42,21 @@
 #include "SPIRVInternal.h"
 #include "libSPIRV/SPIRVDebug.h"
 
+#include "llvm/ADT/SmallPtrSet.h"
+#include "llvm/ADT/SmallVector.h"
 #include "llvm/ADT/StringSwitch.h"
 #include "llvm/Analysis/ValueTracking.h"
+#include "llvm/IR/Constants.h"
 #include "llvm/IR/IRBuilder.h"
 #include "llvm/IR/Instruction.h"
 #include "llvm/IR/Instructions.h"
+#include "llvm/IR/Operator.h"
+#include "llvm/IR/TypedPointerType.h"
 #include "llvm/IR/PatternMatch.h"
 #include "llvm/Support/Debug.h"
 
 #include <algorithm>
+#include <optional>
 #include <regex>
 #include <set>
 
@@ -62,6 +68,91 @@ using namespace SPIRV;
 using namespace OCLUtil;
 
 namespace SPIRV {
+namespace {
+
+static std::optional<unsigned> getAddressSpaceFromType(Type *Ty) {
+  if (!Ty)
+    return std::nullopt;
+  if (auto *TypedPtr = dyn_cast<TypedPointerType>(Ty))
+    return TypedPtr->getAddressSpace();
+  if (auto *Ptr = dyn_cast<PointerType>(Ty))
+    return Ptr->getAddressSpace();
+  return std::nullopt;
+}
+
+static std::optional<unsigned> getAddressSpaceFromValue(Value *Ptr) {
+  if (!Ptr)
+    return std::nullopt;
+
+  SmallPtrSet<Value *, 8> Visited;
+  SmallVector<Value *, 8> Worklist;
+  Worklist.push_back(Ptr);
+  std::optional<unsigned> GenericAS;
+
+  while (!Worklist.empty()) {
+    Value *Current = Worklist.pop_back_val();
+    if (!Visited.insert(Current).second)
+      continue;
+
+    if (auto MaybeAS = getAddressSpaceFromType(Current->getType())) {
+      if (*MaybeAS != SPIRAS_Generic)
+        return MaybeAS;
+      GenericAS = MaybeAS;
+    }
+
+    if (auto *Op = dyn_cast<Operator>(Current)) {
+      switch (Op->getOpcode()) {
+      case Instruction::AddrSpaceCast:
+      case Instruction::BitCast:
+      case Instruction::GetElementPtr:
+        Worklist.push_back(Op->getOperand(0));
+        break;
+      case Instruction::Select:
+        Worklist.push_back(Op->getOperand(1));
+        Worklist.push_back(Op->getOperand(2));
+        break;
+      case Instruction::PHI: {
+        auto *Phi = cast<PHINode>(Op);
+        for (Value *Incoming : Phi->incoming_values())
+          Worklist.push_back(Incoming);
+        break;
+      }
+      default:
+        break;
+      }
+    }
+  }
+
+  return GenericAS;
+}
+
+static unsigned getAtomicPointerMemorySemantics(Value *Ptr, Type *RecordedType) {
+  std::optional<unsigned> AddrSpace = getAddressSpaceFromType(RecordedType);
+  if ((!AddrSpace || *AddrSpace == SPIRAS_Generic) && Ptr)
+    if (auto MaybeAS = getAddressSpaceFromValue(Ptr))
+      AddrSpace = MaybeAS;
+
+  if (!AddrSpace)
+    return MemorySemanticsMaskNone;
+
+  switch (*AddrSpace) {
+  case SPIRAS_Global:
+  case SPIRAS_GlobalDevice:
+  case SPIRAS_GlobalHost:
+    return MemorySemanticsCrossWorkgroupMemoryMask;
+  case SPIRAS_Local:
+    return MemorySemanticsWorkgroupMemoryMask;
+  case SPIRAS_Generic:
+    return MemorySemanticsCrossWorkgroupMemoryMask |
+           MemorySemanticsWorkgroupMemoryMask |
+           MemorySemanticsSubgroupMemoryMask;
+  default:
+    return MemorySemanticsMaskNone;
+  }
+}
+
+} // namespace
+
 static size_t getOCLCpp11AtomicMaxNumOps(StringRef Name) {
   return StringSwitch<size_t>(Name)
       .Cases("load", "flag_test_and_set", "flag_clear", 3)
@@ -700,6 +791,11 @@ void OCLToSPIRVBase::transAtomicBuiltin(CallInst *CI,
   const size_t ScopeIdx = ArgsCount - 1;
   const size_t OrderIdx = ScopeIdx - NumOrder;
 
+  unsigned PtrMemSemantics = MemorySemanticsMaskNone;
+  if (Mutator.arg_size() > 0)
+    PtrMemSemantics =
+        getAtomicPointerMemorySemantics(Mutator.getArg(0), Mutator.getType(0));
+
   if (NeedsNegate) {
     Mutator.mapArg(1, [=](Value *V) {
       IRBuilder<> IRB(CI);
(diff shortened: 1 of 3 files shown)


svenvh:
> @svenvh @bashbaug Tests are currently failing, but if we are good with the approach - I'll investigate and fix the failures.

The approach looks reasonable to me; it would be nice to have some comments on the various functions though (explaining for which cases we need `getAddressSpaceFromValue` for example).

lib/SPIRV/OCLToSPIRV.cpp
```diff
@@ -62,6 +68,91 @@ using namespace SPIRV;
...
+    if (!Visited.insert(Current).second)
+      continue;
+
+    if (auto MaybeAS = getAddressSpaceFromType(Current->getType())) {
+      if (*MaybeAS != SPIRAS_Generic)
+        return MaybeAS;
+      GenericAS = MaybeAS;
+    }
+
+    if (auto *Op = dyn_cast<Operator>(Current)) {
```
bashbaug, line 101: For my own understanding, is this kind of a "light" form of generic addresss space inference?

Is this primarily needed because explicit address space overloads don't exist, or because explicit address space overloads aren't being used, or something else?

To be clear, I think this is fine and a good idea to include, just trying to understand why it is needed.

lib/SPIRV/OCLToSPIRV.cpp
```diff
@@ -62,6 +68,91 @@ using namespace SPIRV;
...
+  if (auto *TypedPtr = dyn_cast<TypedPointerType>(Ty))
+    return TypedPtr->getAddressSpace();
+  if (auto *Ptr = dyn_cast<PointerType>(Ty))
+    return Ptr->getAddressSpace();
+  return std::nullopt;
+}
+
+static std::optional<unsigned> getAddressSpaceFromValue(Value *Ptr) {
+  if (!Ptr)
+    return std::nullopt;
```
bashbaug: Under what circumstances would we call this function and pass a NULL value?

Regardless, should we be conservative and return `SPIRAS_Generic` here?

lib/SPIRV/OCLToSPIRV.cpp
```diff
@@ -62,6 +68,91 @@ using namespace SPIRV;
...
+      default:
+        break;
+      }
+    }
+  }
+
+  return GenericAS;
+}
+
+static unsigned getAtomicPointerMemorySemantics(Value *Ptr, Type *RecordedType) {
```
bashbaug: Minor/nitpick: This isn't getting the full memory semantics from the pointer, because the memory semantics also includes information about the atomic memory order that cannot be derived from the pointer.  Could the name of the function be updated to reflect this, or would it make the name too long?  `getAtomicPointerMemorySemanticsMemoryMask`?

lib/SPIRV/OCLToSPIRV.cpp
```diff
@@ -710,8 +806,19 @@ void OCLToSPIRVBase::transAtomicBuiltin(CallInst *CI,
     return transOCLMemScopeIntoSPIRVScope(V, OCLMS_device, CI);
   });
   for (size_t I = 0; I < NumOrder; ++I) {
-    Mutator.mapArg(OrderIdx + I, [=](Value *V) {
-      return transOCLMemOrderIntoSPIRVMemorySemantics(V, OCLMO_seq_cst, CI);
+    Mutator.mapArg(OrderIdx + I,
+                   [=](IRBuilder<> &Builder, Value *V) -> Value * {
+      Value *MemSem =
+          transOCLMemOrderIntoSPIRVMemorySemantics(V, OCLMO_seq_cst, CI);
+      if (!PtrMemSemantics)
```
bashbaug: Consider:

```suggestion
      if (PtrMemSemantics == MemorySemanticsMaskNone)
```

bashbaug (commented):
The overall approach LGTM, thanks!

lib/SPIRV/OCLToSPIRV.cpp
MrSidims: I'll remove it, there are no such circumstances (now) as number of args is explicitly checked earlier in the call stack. It's better to just put here a debug assertion to silence some static analysis tools.

lib/SPIRV/OCLToSPIRV.cpp
MrSidims: getAtomicPointerMemorySemanticsMemoryMask makes sense to me

lib/SPIRV/OCLToSPIRV.cpp
MrSidims: Yes, it's light version of Infer Address Space pass. It's created because different front ends can pick different default address spaces for pointers. AFAIK even for upstream OpenCL there will be different LLVM IR produced by clang for SPIR and AMDGPU (through SPIR(V) tool chain(s) targets).

I'm actually missing one assumption. If worklist ends up on generic alloca - it should be assumed as private alloca.

lib/SPIRV/OCLToSPIRV.cpp
```diff
@@ -62,6 +68,96 @@ using namespace SPIRV;
...
+    if (auto MaybeAS = getAddressSpaceFromType(Current->getType())) {
+      if (*MaybeAS != SPIRAS_Generic)
+        return MaybeAS;
+      GenericAS = MaybeAS;
+    }
+
+    if (isa<AllocaInst>(Current))
+      // It's safe to assume, that generic alloca is actually a function
+      // storage allocation.
+      return SPIRAS_Private;
```
bashbaug: Is there any way to check for this assumption?

To be clear, I do think that the only possible alloca currently is a function storage class allocation (AKA a private address space allocation), but if this ever changed in the future, this could easily become a nasty bug to track down.

What is generating a generic address space alloca today vs. a private address space alloca?

lib/SPIRV/OCLToSPIRV.cpp
```diff
@@ -62,6 +68,96 @@ using namespace SPIRV;
...
+// address space.
+static unsigned getAtomicPointerMemorySemanticsMemoryMask(Value *Ptr,
+                                                          Type *RecordedType) {
+  std::optional<unsigned> AddrSpace = getAddressSpaceFromType(RecordedType);
+  if ((!AddrSpace || *AddrSpace == SPIRAS_Generic) && Ptr)
+    if (auto MaybeAS = getAddressSpaceFromValue(Ptr))
+      AddrSpace = MaybeAS;
+
+  if (!AddrSpace)
+    return MemorySemanticsMaskNone;
```
bashbaug: Hmm, are there situations where this could legitimately occur and no address space can be determined?

Maybe enforcing this in the SPIR-V LLVM Translator is unnecessary, but I'd like to eventually have a SPIR-V validation rule that says "if the memory semantics includes a memory order (e.g. Acquire, Release, etc.), and it does NOT include a storage class that the memory order applies to (e.g. WorkgroupMemory, CrossWorkgroupMemory, etc.), then the SPIR-V module is invalid".

lib/SPIRV/OCLToSPIRV.cpp
```diff
@@ -62,6 +68,96 @@ using namespace SPIRV;
...
+  case SPIRAS_Global:
+  case SPIRAS_GlobalDevice:
+  case SPIRAS_GlobalHost:
+    return MemorySemanticsCrossWorkgroupMemoryMask;
+  case SPIRAS_Local:
+    return MemorySemanticsWorkgroupMemoryMask;
+  case SPIRAS_Generic:
+    return MemorySemanticsCrossWorkgroupMemoryMask |
+           MemorySemanticsWorkgroupMemoryMask |
+           MemorySemanticsSubgroupMemoryMask;
```
bashbaug: This is probably the safe thing to do since it's "everything", but noting for completeness: there is no "SubgroupMemory" for a generic address space pointer to point to.

lib/SPIRV/OCLToSPIRV.cpp
MrSidims: Technically, default alloca AS is controlled by target triple in clang. For OpenCL it's always '0' aka private. But for example in SYCL, default address space is 4, aka generic (though we expect it be resolved as private).

bashbaug: We're getting a bit out in the weeds beyond the scope of this PR, but I think I convinced myself that this is reasonable, at least for now.  Most allocas should be in a named address space, in which case the resulting pointer will be a non-generic address space pointer.  IMHO we really shouldn't see a generic address space alloca, but if we do, I suppose it's OK to treat it as-if it were a private address space alloca, since it just as much sense as any other address space.

I'll just resolve this conversation for now.

MrSidims: Removed this check

lib/SPIRV/OCLToSPIRV.cpp
MrSidims: Not really. The meaning here is that in case if somehow there is an atomic builtin with not-first-in-order pointer parameter ending up being evaluated by `getAtomicPointerMemorySemanticsMemoryMask` we won't get any issues (note, getAddressSpaceFromType returns std::optional). But actually, let me put an assertion here.

lib/SPIRV/OCLToSPIRV.cpp
```diff
@@ -62,6 +68,90 @@ using namespace SPIRV;
 using namespace OCLUtil;
 
 namespace SPIRV {
+
+static std::optional<unsigned> getAddressSpaceFromType(Type *Ty) {
```
maarquitos14:
```suggestion
static std::optional<unsigned> getAddressSpaceFromType(const Type *Ty) {
```

lib/SPIRV/OCLToSPIRV.cpp
```diff
@@ -62,6 +68,90 @@ using namespace SPIRV;
...
+  assert(Ty && "Can't deduce pointer AS");
+  if (auto *TypedPtr = dyn_cast<TypedPointerType>(Ty))
+    return TypedPtr->getAddressSpace();
+  if (auto *Ptr = dyn_cast<PointerType>(Ty))
+    return Ptr->getAddressSpace();
+  return std::nullopt;
+}
+
+// Performs an address space inference analysis.
+static std::optional<unsigned> getAddressSpaceFromValue(Value *Ptr) {
```
maarquitos14:
```suggestion
static std::optional<unsigned> getAddressSpaceFromValue(const Value *Ptr) {
```

lib/SPIRV/OCLToSPIRV.cpp
```diff
@@ -62,6 +68,90 @@ using namespace SPIRV;
...
+  assert(Ptr && "Can't deduce pointer AS");
+
+  SmallPtrSet<Value *, 8> Visited;
```
maarquitos14:
```suggestion
  SmallPtrSet<const Value *, 8> Visited;
```

lib/SPIRV/OCLToSPIRV.cpp
```diff
@@ -62,6 +68,90 @@ using namespace SPIRV;
...
+  SmallVector<Value *, 8> Worklist;
```
maarquitos14:
```suggestion
  SmallVector<const Value *, 8> Worklist;
```

lib/SPIRV/OCLToSPIRV.cpp
```diff
@@ -62,6 +68,90 @@ using namespace SPIRV;
...
+  Worklist.push_back(Ptr);
+  std::optional<unsigned> GenericAS;
+
+  while (!Worklist.empty()) {
+    Value *Current = Worklist.pop_back_val();
```
maarquitos14:
```suggestion
    const Value *Current = Worklist.pop_back_val();
```

lib/SPIRV/OCLToSPIRV.cpp
```diff
@@ -62,6 +68,90 @@ using namespace SPIRV;
...
+    if (!Visited.insert(Current).second)
+      continue;
+
+    if (auto MaybeAS = getAddressSpaceFromType(Current->getType())) {
+      if (*MaybeAS != SPIRAS_Generic)
```
maarquitos14: Is this safe? Isn't it undefined behaviour when `MaybeAS` is `std::nullopt`?

lib/SPIRV/OCLToSPIRV.cpp
```diff
@@ -62,6 +68,90 @@ using namespace SPIRV;
...
+      }
+    }
+  }
+
+  return GenericAS;
+}
+
+// Sets memory semantic mask of an atomic depending on a pointer argument
+// address space.
+static unsigned getAtomicPointerMemorySemanticsMemoryMask(Value *Ptr,
```
maarquitos14:
```suggestion
static unsigned getAtomicPointerMemorySemanticsMemoryMask(const Value *Ptr,
```

lib/SPIRV/OCLToSPIRV.cpp
```diff
@@ -62,6 +68,90 @@ using namespace SPIRV;
...
+                                                          Type *RecordedType) {
```
maarquitos14:
```suggestion
                                                          const Type *RecordedType) {
```

lib/SPIRV/OCLToSPIRV.cpp
MrSidims: AFAIK std::nullopt != nullptr

maarquitos14: It seems like assigning it to `std::nullopt` returns `has_value()` as false (https://godbolt.org/z/3M6PMffnY). According to [cppreference](https://en.cppreference.com/w/cpp/utility/optional/operator%252A.html):

> If has_value() is false, the behavior is undefined.

PR #3397, opened 2025-10-16
[Backport to 16] Fix incorrect translation for AtomicCmpExchangeWeak builtin ptr arg type. (#3390)

maarquitos14:
Add missing case so that pointer argument types for `OpAtomicCompareExchangeWeak` are resolved correctly. The newly added test, without this patch would introduce an invalid bitcast just before the atomic builtin.


maarquitos14:
@MrSidims  @svenvh Asking for review here because I had to slightly adjust the test here, so just double checking if that is fine for a backport.

svenvh (approved):

PR #3401, opened 2025-10-20
[Backport] Two backports from 17 to 16

bokrzesi:


lib/SPIRV/OCLUtil.cpp
```diff
@@ -898,6 +898,7 @@ SPIRAddressSpace getOCLOpaqueTypeAddrSpace(Op OpCode) {
   case OpTypeSampler:
     return SPIRV_SAMPLER_T_ADDR_SPACE;
   case internal::OpTypeJointMatrixINTEL:
+  case internal::OpTypeJointMatrixINTELv2:
```
MrSidims, line 901: @bokrzesi could you please elaborate, why do we need these backports? They were not done, because our base extension for sycl::joint matrices is SPV_KHR_cooperative_matrix. Hence Intel's version for this functionality no longer contain joint matrix type and other common functionality.

bokrzesi: Hi, I'm trying to figure out what needs to be backported

I need this (`lib/SPIRV/SPIRVReader.cpp`)

image

PR #3403, opened 2025-10-20
Fix memory leaks in addCapability and addConditionalCapability

jmmartinez:
Compiling with `-fsanitize=address` uncovered 2 memory leaks related to the use of raw pointers in `addCapability` and `addConditionalCapability`.

Instead of manually handling memory, use `std::unique_ptr` to track the ownership of the pointed object and free the memory for us.

### DIFF
--- a/lib/SPIRV/libSPIRV/SPIRVModule.cpp
+++ b/lib/SPIRV/libSPIRV/SPIRVModule.cpp
@@ -784,7 +784,7 @@ void SPIRVModuleImpl::addCapability(SPIRVCapabilityKind Cap) {
   SPIRVDBG(spvdbgs() << "addCapability: " << SPIRVCapabilityNameMap::map(Cap)
                      << '\n');
 
-  auto [It, Inserted] = CapMap.insert(std::make_pair(Cap, nullptr));
+  auto [It, Inserted] = CapMap.try_emplace(Cap);
   if (!Inserted)
     return;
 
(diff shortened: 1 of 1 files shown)


lib/SPIRV/libSPIRV/SPIRVModule.cpp
```diff
@@ -787,10 +781,12 @@ void SPIRVModuleImpl::addCapability(SPIRVCapabilityKind Cap) {
   addCapabilities(SPIRV::getCapability(Cap));
   SPIRVDBG(spvdbgs() << "addCapability: " << SPIRVCapabilityNameMap::map(Cap)
                      << '\n');
-  if (hasCapability(Cap))
+
+  auto [It, Inserted] = CapMap.insert(std::make_pair(Cap, nullptr));
```
MrSidims: nit here and in the code below:
```suggestion
  auto [It, Inserted] = CapMap.try_emplace(std::make_pair(Cap, nullptr));
```

jmmartinez: I didn't know `try_emplace` made it into standard containers in C++17. Thanks! I thought it was an LLVM specific thing.

I've updated the PR to use try_emplace.

lib/SPIRV/libSPIRV/SPIRVModule.h
```diff
@@ -109,9 +109,10 @@ struct SPIRVTypeImageDescriptor;
 
 class SPIRVModule {
 public:
-  typedef std::map<SPIRVCapabilityKind, SPIRVCapability *> SPIRVCapMap;
+  typedef std::map<SPIRVCapabilityKind, std::unique_ptr<SPIRVCapability>>
```
maarquitos14, line 112: While we're here, do we really need to keep the order here? Could this be made `std::unordered_map`? That would probably give us better performance. Anyway, I know this is offtopic, so feel free to ignore or address in a different patch.

jmmartinez: I'll add it to my TODO list and do it in a separate PR. It's a low hanging fruit so it's worth checking.

I'll also see if I can remove the rest of the manual memory management from the class, such that the destructor becomes trivial.

jmmartinez: I've gave it a try and I know now why `std::map` is used: If we want to use the unordered containers on these two maps we have to define `std::hash(const SPIRVCapabilityKind&)`.

PR #3411, opened 2025-10-23
[LLVM->SPIRV] Translate `AddrSpaceCast from generic null` to OpGenericCastToPtr

wenju-he:
Some GPU backends materialize non-generic nulls via addrspacecast from generic null, see https://github.com/llvm/llvm-project/blob/14ed67411ce1/clang/lib/CodeGen/Targets/SPIR.cpp#L257

llvm-spirv should preserve the addrspacecast.

### DIFF
--- a/test/addrspacecast_generic_null.ll
+++ b/test/addrspacecast_generic_null.ll
@@ -13,7 +13,6 @@ target triple = "spir64-unknown-unknown"
 ; CHECK: OpPtrEqual %bool %[[CAST]]
 
 define spir_kernel void @bar_generic_null(ptr addrspace(3) %arg) {
-pass26:
   %expr = icmp eq ptr addrspace(3) addrspacecast (ptr addrspace(4) null to ptr addrspace(3)), %arg
   ret void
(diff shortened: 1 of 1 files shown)


lib/SPIRV/SPIRVWriter.cpp
```diff
@@ -1672,7 +1672,8 @@ SPIRVValue *LLVMToSPIRVBase::transUnaryInst(UnaryInstruction *U,
       } else {
         BOC = OpCrossWorkgroupCastToPtrINTEL;
       }
-    } else if (isa<ConstantPointerNull>(Cast->getPointerOperand())) {
+    } else if (isa<ConstantPointerNull>(Cast->getPointerOperand()) &&
+               SrcAddrSpace != SPIRAS_Generic) {
```
MrSidims, line 1676: Ah, seems like I've missed Src == Generic case while writing https://github.com/KhronosGroup/SPIRV-LLVM-Translator/pull/3305#discussion_r2266229718

PR #3417, opened 2025-10-28
Preserve byval/sret typed pointer semantics with SPV_KHR_untyped_pointers

vmaksimo:
Avoid translation of pointer function arguments with `byval`/`sret` attribute as untyped pointers to preserve the information about the pointer element type.
Insert `OpBitcast` to further use such pointers in the untyped pointer semantics (vise-versa bitcast instructions are explicitly allowed by the [SPV_KHR_untyped_pointers ](https://github.khronos.org/SPIRV-Registry/extensions/KHR/SPV_KHR_untyped_pointers.html) extension).

This approach ensures valid reverse translation and correct OpenCL runtime behavior, especially for kernels translation. 
Without the fix `clSetKernelArg()` fails with `CL_INVALID_MEM_OBJECT` error.

### DIFF
--- a/test/transcoding/unused-sret-opaque-ptr.ll
+++ b/test/transcoding/unused-sret-opaque-ptr.ll
@@ -1,14 +1,16 @@
-; RUN: llvm-as %s -o %t.bc
-; RUN: llvm-spirv %t.bc -spirv-text -o %t.txt
+; Ensure that sret pointer semantics is preserved when the parameter is unused
+; (even with the SPV_KHR_untyped_pointers extension enabled).
+
+; RUN: llvm-spirv %s -spirv-text -o %t.txt
 ; RUN: FileCheck < %t.txt %s --check-prefix=CHECK-SPIRV
-; RUN: llvm-spirv %t.bc -o %t.spv
+; RUN: llvm-spirv %s -o %t.spv
 ; RUN: spirv-val %t.spv
 ; RUN: llvm-spirv -r %t.spv -o %t.rev.bc
 ; RUN: llvm-dis < %t.rev.bc | FileCheck %s --check-prefix=CHECK-LLVM
 
-; RUN: llvm-spirv %t.bc -spirv-text -o %t.txt --spirv-ext=+SPV_KHR_untyped_pointers
+; RUN: llvm-spirv %s -spirv-text -o %t.txt --spirv-ext=+SPV_KHR_untyped_pointers
 ; RUN: FileCheck < %t.txt %s --check-prefix=CHECK-SPIRV
-; RUN: llvm-spirv %t.bc -o %t.spv --spirv-ext=+SPV_KHR_untyped_pointers
+; RUN: llvm-spirv %s -o %t.spv --spirv-ext=+SPV_KHR_untyped_pointers
 ; RUNx: spirv-val %t.spv
 ; RUN: llvm-spirv -r %t.spv -o %t.rev.bc
 ; RUN: llvm-dis < %t.rev.bc | FileCheck %s --check-prefix=CHECK-LLVM
--- a/test/extensions/KHR/SPV_KHR_untyped_pointers/byval.ll
+++ b/test/extensions/KHR/SPV_KHR_untyped_pointers/byval.ll
@@ -1,3 +1,5 @@
+; Ensure that a typed pointer passed by value is converted to an untyped pointer prior usage.
+
 ; RUN: llvm-spirv %s -spirv-text -o %t.spt --spirv-ext=+SPV_KHR_untyped_pointers
 ; RUN: FileCheck < %t.spt %s --check-prefix=CHECK-SPIRV
 ; RUN: llvm-spirv %s -o %t.spv --spirv-ext=+SPV_KHR_untyped_pointers
--- a/test/transcoding/unused-byval-opaque-ptr.ll
+++ b/test/transcoding/unused-byval-opaque-ptr.ll
@@ -1,14 +1,16 @@
-; RUN: llvm-as %s -o %t.bc
-; RUN: llvm-spirv %t.bc -spirv-text -o %t.txt
+; Ensure that a pointer passed by value is translated as a typed pointer even
+; with the SPV_KHR_untyped_pointers extension enabled to preserve byval semantics.
+
+; RUN: llvm-spirv %s -spirv-text -o %t.txt
 ; RUN: FileCheck < %t.txt %s --check-prefix=CHECK-SPIRV
-; RUN: llvm-spirv %t.bc -o %t.spv
+; RUN: llvm-spirv %s -o %t.spv
 ; RUN: spirv-val %t.spv
 ; RUN: llvm-spirv -r %t.spv -o %t.rev.bc
 ; RUN: llvm-dis < %t.rev.bc | FileCheck %s --check-prefix=CHECK-LLVM
 
-; RUN: llvm-spirv %t.bc -spirv-text -o %t.txt --spirv-ext=+SPV_KHR_untyped_pointers
+; RUN: llvm-spirv %s -spirv-text -o %t.txt --spirv-ext=+SPV_KHR_untyped_pointers
 ; RUN: FileCheck < %t.txt %s --check-prefix=CHECK-SPIRV
-; RUN: llvm-spirv %t.bc -o %t.spv --spirv-ext=+SPV_KHR_untyped_pointers
+; RUN: llvm-spirv %s -o %t.spv --spirv-ext=+SPV_KHR_untyped_pointers
 ; RUN: spirv-val %t.spv
 ; RUN: llvm-spirv -r %t.spv -o %t.rev.bc
 ; RUN: llvm-dis < %t.rev.bc | FileCheck %s --check-prefix=CHECK-LLVM


lib/SPIRV/SPIRVWriter.cpp
```diff
@@ -2214,7 +2229,33 @@ LLVMToSPIRVBase::transValueWithoutDecoration(Value *V, SPIRVBasicBlock *BB,
...
+    if (BM->isAllowedToUseExtension(ExtensionID::SPV_KHR_untyped_pointers) &&
+        SPVArg->getType()->isTypePointer() &&
+        !SPVArg->getType()->isTypeUntypedPointerKHR()) {
+      // When SPV_KHR_untyped_pointers extension is enabled, bitcast typed
+      // pointer function arguments to untyped pointers for further usage in the
+      // untyped pointers paradigm.
+      // Don't do this if the argument is further used in function calls as it
+      // breaks SPIR-V validity.
+      for (auto *U : V->users()) {
+        if (isa<CallInst>(U) || isa<InvokeInst>(U) || isa<PHINode>(U))
```
MrSidims: Why phi instruction is also guarded?

lib/SPIRV/SPIRVWriter.cpp
```diff
@@ -2214,7 +2229,33 @@ LLVMToSPIRVBase::transValueWithoutDecoration(Value *V, SPIRVBasicBlock *BB,
...
+          return mapValue(V, SPVArg);
```
MrSidims: Shouldn't we also skip bit/addrspace casts?

lib/SPIRV/SPIRVWriter.cpp
vmaksimo: Most likely by mistake - I met a similar failure in e2e tests with PHI instruction but it is related to [other explicit bitcast insertion](https://github.com/KhronosGroup/SPIRV-LLVM-Translator/blob/630f8ef0759944e7d0416e6bb70bd2c1a9c8dc6e/lib/SPIRV/SPIRVWriter.cpp#L2517): `PHI nodes not grouped at top of basic block!`
Will double-check that this check is not needed here and remove.

lib/SPIRV/SPIRVWriter.cpp
vmaksimo: I don't think so. Call instructions are the only places where we should preserve the same arguments types as in the "top" function to avoid signature mismatches.

MrSidims: So we can't have anything like this:
```
void spir_func @foo(ptr sret(%struct.Example) %ret_ptr) {
  %cast = addrspacecast ptr %ret_ptr to ptr addrspace(4)
  call void spir_func @boo(ptr addrspace(4) sret(%struct.Example) %cast)
}
```
?

lib/SPIRV/SPIRVWriter.cpp
vmaksimo: Yep, this can be safely removed, thanks for noticing!

lib/SPIRV/SPIRVWriter.cpp
vmaksimo: Verified locally - it's a valid point...
Per offline discussion - rewriting this block to the opposite logic: do the bitcast where it's needed for sure (load, store, memcpy instructions), and avoid bitcasting in other cases.

test/llvm-intrinsics/memset-opaque.ll
```diff
@@ -8,16 +8,16 @@
 ; RUN: llvm-spirv %t.bc -spirv-text -o %t.spt --spirv-ext=+SPV_KHR_untyped_pointers
 ; RUN: FileCheck < %t.spt %s --check-prefixes=CHECK-SPIRV,CHECK-SPIRV-UNTYPED-PTR
 ; RUN: llvm-spirv %t.bc -o %t.spv --spirv-ext=+SPV_KHR_untyped_pointers
-; RUN: spirv-val %t.spv
+; RUNx: spirv-val %t.spv
```
YuriPlyakhin, line 16: is it possible to add `issue` to not forget to enable it back? Also, comment here about why it was disabled and when it is going to be enabled back?

test/transcoding/unused-sret-opaque-ptr.ll
```diff
@@ -23,8 +46,16 @@ target triple = "spir-unknown-unknown"
...
+  call void @boo(ptr sret(%struct.Example) align 8 %1, i32 noundef 42)
   ret i32 0
 }
 
-declare void @_Z3booi(ptr sret(%struct.Example) align 8, i32 noundef)
+define spir_func void @bar(ptr sret(%struct.Example) %ret_ptr) {
+  %cast = addrspacecast ptr %ret_ptr to ptr addrspace(4)
+  call void @baz(ptr addrspace(4) sret(%struct.Example) %cast)
+  ret void
+}
```
YuriPlyakhin, line 59: the test case is named `unused-sret-opaque-ptr`. I think now it is checking more than unused sret ptr. Should we consider renaming?

lib/SPIRV/SPIRVWriter.cpp
```diff
@@ -2214,7 +2228,43 @@ LLVMToSPIRVBase::transValueWithoutDecoration(Value *V, SPIRVBasicBlock *BB,
...
+      auto PerformBitcastForArg =
+          [&](SPIRVFunctionParameter *BA) -> SPIRVValue * {
+        // Position to insert bitcast should be right after variable insertion
+        // point in the entry basic block.
+        auto *InsertBB = BF->getBasicBlock(0);
+        auto *InsertPoint = InsertBB->getVariableInsertionPoint();
+        auto *UntypedPtrType = BM->addPointerType(
+            BA->getType()->getPointerStorageClass(), nullptr);
+
+        auto *Bitcast = BM->addUnaryInst(OpBitcast, UntypedPtrType, BA,
```
YuriPlyakhin, line 2254: In test changes and new tests in this PR I did not notice new bitcasts. What test/check is covering this code in the PR?

test/transcoding/unused-sret-opaque-ptr.ll
```diff
@@ -6,14 +6,37 @@
...
+
+; CHECK-SPIRV: Function [[#]] [[#FunBoo]] [[#]] [[#BooTy]]
+; CHECK-SPIRV: FunctionParameter [[#PtrTy7]] [[#ParamBoo]]
+
+; CHECK-SPIRV: Function [[#]] [[#FunBaz]] [[#]] [[#BazTy]]
+; CHECK-SPIRV: FunctionParameter [[#PtrTy8]] [[#ParamBaz]]
+
+; CHECK-SPIRV: FunctionParameter [[#PtrTy7]] [[#ParamBar:]]
+; With untyped extension enabled addrspacecast is done to untyped pointer type in addrspace 8.
+; CHECK-SPIRV: PtrCastToGeneric [[#]] [[#Cast:]] [[#ParamBar]]
```
YuriPlyakhin, line 37: If this test now checks more than just unused sret ptr cases, maybe be it worth adding more `CHECK-LLVM` checks corresponding to `CHECK-SPIRV` lines?

vmaksimo: It actually checks the same, just in more extensive style and with the additional test case - the main checks for SPIR-V still include 1) verification of function attribute and 2) validity of SPV when LLVM IR contains opaque pointers.
As for the CHECK-LLVM part, there is nothing more to check as I see.

lib/SPIRV/SPIRVWriter.cpp
vmaksimo: Hm, it was covered with `memset-opaque.ll` before we fixed the core logic.. I'll add a new case for that.

test/llvm-intrinsics/memset-opaque.ll
vmaksimo: I'll add the `TODO` comment (the same way it was already done for other tests for untyped pointers), e.g. see: https://github.com/KhronosGroup/SPIRV-LLVM-Translator/blob/a3a5de742179b342a537df96ad6e9687723279c7/test/long-type-struct.ll#L19

YuriPlyakhin: ok

test/transcoding/unused-sret-opaque-ptr.ll
YuriPlyakhin: ok

test/transcoding/unused-sret-opaque-ptr.ll
vmaksimo: Same as above - don't think the renaming is needed

lib/SPIRV/SPIRVWriter.cpp
vmaksimo: Added `byval.ll` test

YuriPlyakhin (approved):

lib/SPIRV/SPIRVWriter.cpp
```diff
@@ -866,6 +866,20 @@ SPIRVType *LLVMToSPIRVBase::transScavengedType(Value *V) {
...
         Ty = FnTy->getParamType(Arg.getArgNo());
       }
+      // Preserve element type for byval/sret arguments even when
+      // SPV_KHR_untyped_pointers is enabled. Losing pointee type would make it
+      // impossible to reconstruct the original parameter and will lead to
+      // OpenCL runtime failure due to mismatched memory object semantics.
+      if (BM->isAllowedToUseExtension(ExtensionID::SPV_KHR_untyped_pointers) &&
+          (Arg.hasByValAttr() || Arg.hasStructRetAttr())) {
+        TypedPointerType *TPT = cast<TypedPointerType>(Ty);
+        auto *NewType = BM->addPointerType(
```
YixingZhang007, line 876: Should we check for 'Ty' to be 'TypedPointerType' before doing the cast (for example, by adding an 'assert')? Also, I think it might be better to use explicit types here, such as 'SPIRVType *NewType', same as those below.

vmaksimo: Cast without a check is intentional in this case - it is done to avoid masking the core errors and make bug investigation easier.
About types, I'd rather make the first one `auto` as well :) but maybe it's not needed to be changed after all

PR #3418, opened 2025-10-28
Remove entry point wrapper

vmaksimo:
After LLVM community changes we don't need to do it ourselves.

TODO:

- [x] Remove all the unused functions
- [x] Fix test checks (probably revert of https://github.com/intel/llvm/pull/14450/ would do the job)

### DIFF
--- a/test/transcoding/enqueue_kernel.cl
+++ b/test/transcoding/enqueue_kernel.cl
@@ -21,8 +21,6 @@
 // RUN: llvm-dis %t.rev.bc
 // RUN: FileCheck < %t.rev.ll %s --check-prefix=CHECK-SPV-IR
 
-// TODO: review the test
-
 // CHECK-SPIRV: EntryPoint {{[0-9]+}} [[BlockKer1:[0-9]+]] "__device_side_enqueue_block_invoke_kernel"
 // CHECK-SPIRV: EntryPoint {{[0-9]+}} [[BlockKer2:[0-9]+]] "__device_side_enqueue_block_invoke_2_kernel"
 // CHECK-SPIRV: EntryPoint {{[0-9]+}} [[BlockKer3:[0-9]+]] "__device_side_enqueue_block_invoke_3_kernel"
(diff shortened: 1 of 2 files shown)


test/transcoding/enqueue_kernel.cl
```diff
@@ -21,18 +21,15 @@
 // RUN: llvm-dis %t.rev.bc
 // RUN: FileCheck < %t.rev.ll %s --check-prefix=CHECK-SPV-IR
 
+// TODO: review the test
```
vmaksimo: Guess it looks fine, just worried because the test is huge and some code gets duplicated (or just extended?) by clang

lib/SPIRV/LLVMToSPIRVDbgTran.cpp
```diff
@@ -1276,17 +1276,8 @@ SPIRVEntry *LLVMToSPIRVDbgTran::transDbgFunction(const DISubprogram *Func) {
...
-        // Function definition of spir_kernel can have no "spir_kernel" calling
-        // convention because SPIRVRegularizeLLVMBase::addKernelEntryPoint pass
-        // could have turned it to spir_func. The "true" entry point is a
-        // wrapper kernel function, which can be found further in the module.
-        if (FuncDef) {
-          if (F.getCallingConv() == CallingConv::SPIR_KERNEL) {
-            IsEntryPointKernel = true;
-            break;
-          }
+        if (FuncDef)
```
vmaksimo, line 1279: FYI, `FuncDef` was introduced in 55ce223, and `IsEntryPointKernel` - in 8565381

MrSidims:
LLVM_SPIRV :: DebugInfo/DebugInfoTypeVector.spvasm
  
this seem to be a new failure, though not sure if it's related

vmaksimo:
@MrSidims thanks for noticing! It is related, I fixed similar error in the c7ec392 (the error appears because kernel wrapper and kernel itself have the same name - making invalid `call @kernel` in reverse translation

MrSidims (approved):

PR #3343, opened 2025-09-10
Support for SPV_INTEL_shader_atomic_bfloat16 extension

MrSidims:
Spec is available here:
https://github.com/intel/llvm/pull/20009

Author:
"Ratajewski, Andrzej" <andrzej.ratajewski@intel.com>

PR #3424, opened 2025-11-04
Add SPV_INTEL_16bit_atomics extension

vmaksimo:
This continues #3343 and reflects specification update, including extension renaming.
Specification: https://github.com/intel/llvm/pull/20009

TODO:

- [x] Update capabilities' magic numbers once they are finalized.
- [x] add/find tests for atomic load/store/exchange/compareexchange/weak + add support
- [x] verify atomic instructions with 16-bit integers

### DIFF
--- a/test/extensions/KHR/SPV_KHR_untyped_pointers/byval.ll
+++ b/test/extensions/KHR/SPV_KHR_untyped_pointers/byval.ll
@@ -7,7 +7,7 @@
 ; RUN: llvm-spirv -r %t.spv -o %t.rev.bc
 ; RUN: llvm-dis < %t.rev.bc | FileCheck %s --check-prefix=CHECK-LLVM
 
-; CHECK-SPIRV-DAG: Name [[#Fun:]] "kernel"
+; CHECK-SPIRV-DAG: EntryPoint [[#]] [[#Fun:]] "kernel"
 ; CHECK-SPIRV-DAG: Decorate [[#Param:]] FuncParamAttr 2
 ; CHECK-SPIRV-DAG: TypeUntypedPointerKHR [[#UntypedPtrTy:]] 7
 ; CHECK-SPIRV-DAG: TypeStruct [[#StructTy:]]
--- a/test/transcoding/unused-byval-opaque-ptr.ll
+++ b/test/transcoding/unused-byval-opaque-ptr.ll
@@ -15,7 +15,7 @@
 ; RUN: llvm-spirv -r %t.spv -o %t.rev.bc
 ; RUN: llvm-dis < %t.rev.bc | FileCheck %s --check-prefix=CHECK-LLVM
 
-; CHECK-SPIRV-DAG: Name [[#Fun:]] "kernel"
+; CHECK-SPIRV-DAG: EntryPoint [[#]] [[#Fun:]] "kernel"
 ; CHECK-SPIRV-DAG: Decorate [[#Param:]] FuncParamAttr 2
 ; CHECK-SPIRV-DAG: TypeStruct [[#StructTy:]]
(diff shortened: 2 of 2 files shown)


vmaksimo:
~~The refactoring made in `SPIRVInstruction.h` triggered the issue that we add `CapabilityInt64Atomics` only for `atomic store` instruction (see CI failures).~~
Nope, my mistake.

lib/SPIRV/libSPIRV/SPIRVInstruction.h
```diff
@@ -2965,8 +2965,18 @@ class SPIRVAtomicInstBase : public SPIRVInstTemplateBase {
     // Besides, OpAtomicCompareExchangeWeak, OpAtomicFlagTestAndSet and
     // OpAtomicFlagClear instructions require the "kernel" capability. But this
     // capability should be added by setting the OpenCL memory model.
-    if (hasType() && getType()->isTypeInt(64))
-      return {CapabilityInt64Atomics};
+    if (hasType()) {
+      if (getType()->isTypeInt(64))
+        return {CapabilityInt64Atomics};
+      if (getType()->isTypeInt(16))
+        return {internal::CapabilityInt16AtomicsINTEL};
```
MrSidims: This does look like a breaking change. I understand, that OpenCL env disallows any atomics by 32-bits (unless extended by capabilities), but do we want to tighten us to environmental spec during SPIR-V generation? May be safer to check, if SPV_INTEL_16bit_atomics is allowed, before adding the capability (and do not add `getRequiredExtension` for this class).

Though not sure, if there will be any real-life case, where it could break anything. @svenvh WDYT about this?

lib/SPIRV/libSPIRV/SPIRVInstruction.h
```diff
@@ -3046,7 +3069,7 @@ class SPIRVAtomicFMinMaxEXTBase : public SPIRVAtomicInstBase {
 public:
   std::optional<ExtensionID> getRequiredExtension() const override {
     if (getType()->isTypeFloat(16, FPEncodingBFloat16KHR))
-      return ExtensionID::SPV_INTEL_shader_atomic_bfloat16;
+      return ExtensionID::SPV_INTEL_16bit_atomics;
```
MrSidims: This actually seem to be a bug from a previous implementation. It should be
```suggestion
      return {ExtensionID::SPV_EXT_shader_atomic_float_min_max, ExtensionID::SPV_INTEL_16bit_atomics};
```
same applicable to fadd.

lib/SPIRV/libSPIRV/SPIRVInstruction.h
svenvh: Not sure how much we'll see this in the wild indeed, but perhaps it's better to follow @MrSidims's suggestion and check if this new extension is allowed before adding the capability.  Requiring a vendor extension for a standard type may be quite surprising if anyone hits this in the future.

lib/SPIRV/libSPIRV/SPIRVInstruction.h
vmaksimo: Maybe because we even don't have the method to return more than one required extension.
UPD: this actually was the intention of 1aae8db379349 some time ago and we have a W/A there.
UPD2: not sure yet how to handle the same for min/max, because we don't have "subset" extension in this case.

vmaksimo: Ok, that may look fun, multiple extension requirement was intentionally removed about 5 years ago in 584329f:
> In all cases what we have encountered so far each instruction has either
zero or one extension that it requires and therefore there is no need to
use `std::set` to store that information.

Issue #3402, opened 2025-10-20
SPIRV emitted for `memset` fails validation


simeonschaub:
Reproducer:

```llvm
target datalayout = "e-i64:64-v16:16-v24:32-v32:32-v48:64-v96:128-v192:256-v256:256-v512:512-v1024:1024-G1"
target triple = "spir64-unknown-unknown"

; Function Attrs: nocallback nofree nounwind willreturn memory(argmem: write)
declare void @llvm.memset.p0.i64(ptr nocapture writeonly, i8, i64, i1 immarg) #0

define spir_kernel void @f() {
conversion:
  call void @llvm.memset.p0.i64(ptr null, i8 0, i64 0, i1 false)
  %0 = add nsw i64 0, 0
  ret void
}

attributes #0 = { nocallback nofree nounwind willreturn memory(argmem: write) }
```

translates to:

```spv
; SPIR-V
; Version: 1.4
; Generator: Khronos LLVM/SPIR-V Translator; 14
; Bound: 22
; Schema: 0
               OpCapability Addresses
               OpCapability Linkage
               OpCapability Kernel
               OpCapability Int64
               OpCapability Int8
          %1 = OpExtInstImport "OpenCL.std"
               OpMemoryModel Physical64 OpenCL
               OpEntryPoint Kernel %19 "f"
               OpSource Unknown 0
               OpName %f "f"
               OpName %conversion "conversion"
               OpDecorate %f LinkageAttributes "f" Export
               OpDecorate %12 Constant
               OpDecorate %18 NoSignedWrap
      %ulong = OpTypeInt 64 0
      %uchar = OpTypeInt 8 0
    %ulong_1 = OpConstant %ulong 1
    %ulong_0 = OpConstant %ulong 0
       %void = OpTypeVoid
          %3 = OpTypeFunction %void
%_arr_uchar_ulong_1 = OpTypeArray %uchar %ulong_1
%_ptr_UniformConstant__arr_uchar_ulong_1 = OpTypePointer UniformConstant %_arr_uchar_ulong_1
%_ptr_UniformConstant_uchar = OpTypePointer UniformConstant %uchar
%_ptr_Function_uchar = OpTypePointer Function %uchar
         %10 = OpConstantNull %_arr_uchar_ulong_1
         %12 = OpVariable %_ptr_UniformConstant__arr_uchar_ulong_1 UniformConstant %10
         %16 = OpConstantNull %_ptr_Function_uchar
[...]

PR #3436, opened 2025-11-06
Fix OpVariable creation during memset translation

MrSidims:
It should be a function-level OpVariable, not Global var.

Fixes https://github.com/KhronosGroup/SPIRV-LLVM-Translator/issues/3402


MrSidims:
should solve the problem, @simeonschaub

PR #3460, opened 2025-11-20
Fix OCL builtin mangling for the indirect pointer usage

vmaksimo:
This continues #2723 to support correct OCL builtins translation with the `SPV_KHR_untyped_pointers` extension enabled.

It reuses the approach from #2924 to get the "true" pointer element type and restore proper SPIR-V builtin mangling.
The approach has been refactored into a separate function, which is now also used for translating OCL builtins.

### DIFF
--- a/lib/SPIRV/SPIRVReader.cpp
+++ b/lib/SPIRV/SPIRVReader.cpp
@@ -3675,17 +3675,17 @@ Instruction *SPIRVToLLVM::transBuiltinFromInst(const std::string &FuncName,
   std::vector<Type *> ArgTys =
       transTypeVector(SPIRVInstruction::getOperandTypes(Ops), true);
 
-  auto Ptr = findFirstPtrType(ArgTys);
-  if (Ptr < ArgTys.size() &&
-      BI->getValueType(Ops[Ptr]->getId())->isTypeUntypedPointerKHR()) {
+  unsigned PtrIdx = findFirstPtrType(ArgTys);
+  if (PtrIdx < ArgTys.size() &&
+      BI->getValueType(Ops[PtrIdx]->getId())->isTypeUntypedPointerKHR()) {
     // Special handling for "truly" untyped pointers to preserve correct
     // builtin mangling of atomic and matrix operations.
     if (isAtomicOpCodeUntypedPtrSupported(OC)) {
       auto *AI = static_cast<SPIRVAtomicInstBase *>(BI);
-      ArgTys[Ptr] = TypedPointerType::get(
+      ArgTys[PtrIdx] = TypedPointerType::get(
           transType(AI->getSemanticType()),
-          SPIRSPIRVAddrSpaceMap::rmap(
-              BI->getValueType(Ops[Ptr]->getId())->getPointerStorageClass()));
+          SPIRSPIRVAddrSpaceMap::rmap(BI->getValueType(Ops[PtrIdx]->getId())
+                                          ->getPointerStorageClass()));
     }
   }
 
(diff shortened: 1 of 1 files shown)


YuriPlyakhin (approved):

lib/SPIRV/SPIRVReader.cpp
```diff
@@ -5296,16 +5298,31 @@ Instruction *SPIRVToLLVM::transOCLBuiltinFromExtInst(SPIRVExtInst *BC,
...
-    if (!isa<PointerType>(ArgTypes[I]) ||
-        !BC->getArgValue(I)->getType()->isTypeUntypedPointerKHR())
-      continue;
-
-    Type *NewPtrTy = getTypedPtrFromUntypedOperand(BC->getArgValue(I), RetTy);
-    if (NewPtrTy)
-      ArgTypes[I] = NewPtrTy;
+  // Special handling for "truly" untyped pointers to preserve correct
+  // OCL builtin mangling.
+  auto Ptr = findFirstPtrType(ArgTypes);
```
YuriPlyakhin:
```suggestion
  auto PtrIndex = findFirstPtrType(ArgTypes);
```

lib/SPIRV/SPIRVReader.cpp
```diff
@@ -5296,16 +5298,31 @@ Instruction *SPIRVToLLVM::transOCLBuiltinFromExtInst(SPIRVExtInst *BC,
...
+  if (Ptr < ArgTypes.size() && isa<PointerType>(ArgTypes[Ptr])) {
```
YuriPlyakhin: Should we check if it is TypeUntypedPointerKHR? Otherwise, how do we know that it is an untyped pointer?

vmaksimo: It could be counterintuitive, but LLVM `PointerType` means untyped pointer :) 
And this `isa` check would fail for `TypedPointerType` because it's not inherited from "default pointer type".

So here we are sure that the type will be translated from `TypeUntypedPointerKHR`.

lib/SPIRV/SPIRVReader.cpp
```diff
@@ -3816,6 +3773,55 @@ SPIRVToLLVM::SPIRVToLLVM(Module *LLVMModule, SPIRVModule *TheSPIRVModule)
   DbgTran.reset(new SPIRVToLLVMDbgTran(TheSPIRVModule, LLVMModule, this));
 }
 
+Type *SPIRVToLLVM::getTypedPtrFromUntypedOperand(SPIRVValue *Val, Type *RetTy) {
```
maarquitos14, line 3776:
```suggestion
Type *SPIRVToLLVM::getTypedPtrFromUntypedOperand(const SPIRVValue *Val, const Type *RetTy) {
```

My gut tells me both params should be const for a function named like this. Am I right?

test/transcoding/OpenCL/math-builtins.ll
```diff
@@ -0,0 +1,118 @@
...
+
+; CHECK-LLVM-LABEL: define spir_func void @fract
+; CHECK-LLVM-COUNT-3: @_Z5fractfPf
+; CHECK-LLVM-LABEL: ret void
+define spir_func void @fract(ptr %arg) {
+entry:
+  %ptr = alloca float
+  %p = bitcast ptr %ptr to ptr
+  %p2 = bitcast ptr %arg to ptr
+  %res = call spir_func float @_Z17__spirv_ocl_fractfPf(float 1.250000e+00, ptr %ptr)
```
maarquitos14, line 32: Is there any reason to make three calls to each OCL extended instruction? Is it to check that different types are correctly mangled?

lib/SPIRV/SPIRVReader.cpp
```diff
@@ -5290,20 +5296,35 @@ Instruction *SPIRVToLLVM::transOCLBuiltinFromExtInst(SPIRVExtInst *BC,
...
-        BC->getArgValue(I)->isUntypedVariable()) {
-      auto *BVar = static_cast<SPIRVUntypedVariableKHR *>(BC->getArgValue(I));
-      ArgTypes[I] = TypedPointerType::get(
-          transType(BVar->getDataType()),
-          SPIRSPIRVAddrSpaceMap::rmap(BVar->getStorageClass()));
+  // Special handling for "truly" untyped pointers to preserve correct
+  // OCL builtin mangling.
+  auto Ptr = findFirstPtrType(ArgTypes);
+  if (Ptr < ArgTypes.size() && isa<PointerType>(ArgTypes[Ptr])) {
+    if (ExtOp == OpenCLLIB::Frexp || ExtOp == OpenCLLIB::Remquo ||
```
YixingZhang007: Are more instructions type expected to be check against `ExtOp` here? If so, I think we might consider using switch statement for better readability, similar to the following.
```
switch (ExtOp) {
    case OpenCLLIB::Frexp:
    case OpenCLLIB::Remquo:
    case OpenCLLIB::Lgamma_r: { ... }
    case OpenCLLIB::Printf: {...}
   default: {}
}
```

lib/SPIRV/SPIRVReader.cpp
YixingZhang007: LLVM 15+ enables opaque pointers by default, but only until LLVM 17,  the support for typed pointers is removed completely (https://llvm.org/docs/OpaquePointers.html). I guess this approach would work as long as we would not try applying this patch on a earlier LLVM version.

test/transcoding/OpenCL/math-builtins.ll
vmaksimo: Exactly for this! These three ways to pass the argument to the builtin can be handled differently, e.g. the last call for all the builtins was failing with the wrong mangling (pointer to char instead of pointer to float) - fixed with the last PR iterations.

lib/SPIRV/SPIRVReader.cpp
vmaksimo: These are the only special cases listed in OCL extended instructions set where the type should be explicitly set to int, but I like the look of the switch case!

lib/SPIRV/SPIRVReader.cpp
vmaksimo: Hmm, actually I think we would backport this extension to older release branches when it's ready.

After all it seems it'd be better to add a check for `TypeUntypedPointerKHR` in this very PR to avoid problems in future.

lib/SPIRV/SPIRVReader.cpp
vmaksimo: They may as an idea. But inside we call `Value *getTranslatedValue(SPIRVValue *)` and `TypedPointerType *get(Type *, unsigned)` that don't expect the pointers to be constant.

PR #3461, opened 2025-11-20
initial support for SPV_INTEL_device_barrier

bashbaug:
Adds initial (preview) support for device scope barriers via SPV_INTEL_device_barrier.

See: https://github.com/intel/llvm/pull/12092

### DIFF
--- a/test/extensions/INTEL/SPV_INTEL_device_barrier/device_barrier_runtime_scope_spirv.ll
+++ b/test/extensions/INTEL/SPV_INTEL_device_barrier/device_barrier_runtime_scope_spirv.ll
@@ -1,0 +1,116 @@
+;; kernel void test(global uint* dst)
+;; {
+;;    int scope = magic_get_scope();
+;;    __spirv_ControlBarrier(scope, 1, 264);  // local
+;;    __spirv_ControlBarrier(scope, 1, 520);  // global
+;;    __spirv_ControlBarrier(scope, 1, 2056); // image
+;;
+;;    __spirv_ControlBarrier(scope, 0, 520);  // global, all_svm_devices
+;;    __spirv_ControlBarrier(scope, 1, 520);  // global, device
+;;    __spirv_ControlBarrier(scope, 2, 520);  // global, work_group
+;;    __spirv_ControlBarrier(scope, 3, 520);  // global, subgroup
+;;    __spirv_ControlBarrier(scope, 4, 520);  // global, work_item
+;;}
+
+; Test for SPV_INTEL_device_barrier (SPIR-V friendly LLVM IR)
+; RUN: llvm-as %s -o %t.bc
+; RUN: llvm-spirv %t.bc -o %t.spv --spirv-ext=+SPV_INTEL_device_barrier
+; RUN: llvm-spirv %t.spv -o %t.spt --to-text
+; RUN: FileCheck < %t.spt %s --check-prefix=CHECK-SPIRV
+
+; RUN: llvm-spirv %t.spv -o %t.rev.bc -r --spirv-target-env=SPV-IR
+; RUN: llvm-dis %t.rev.bc -o %t.rev.ll
+; RUN: FileCheck < %t.rev.ll %s --check-prefix=CHECK-LLVM
+
+; RUN: llvm-spirv %t.bc -o %t.disabled.spv
+; RUN: llvm-spirv %t.disabled.spv -o %t.disabled.spt --to-text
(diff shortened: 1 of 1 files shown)


lib/SPIRV/libSPIRV/SPIRVInstruction.h
```diff
@@ -2488,13 +2488,45 @@ class SPIRVControlBarrier : public SPIRVInstruction {
...
+    if (ESV && ESV->getOpCode() == OpConstant) {
+      if (static_cast<SPIRVConstant *>(ESV)->getZExtIntValue() == ScopeDevice) {
+        getModule()->getErrorLog().checkError(CanUseDeviceBarrier,
+                                              SPIRVEC_RequiresExtension,
+                                              "SPV_INTEL_device_barrier\n");
+        return true;
+      }
+      return false;
+    }
+    // If we cannot determine the execution scope, assume that it can be a
```
MrSidims: It feels like a breaking change. Searching through specifications I don't see a restriction of a Scope not being SpecConstant or runtime value.

lib/SPIRV/libSPIRV/SPIRVInstruction.h
```diff
@@ -2488,13 +2488,45 @@ class SPIRVControlBarrier : public SPIRVInstruction {
...
   _SPIRV_DEF_ENCDEC3(ExecScope, MemScope, MemSema)
   void validate() const override {
     assert(OpCode == OC);
     assert(WordCount == 4);
     SPIRVInstruction::validate();
   }
+
+  bool isDeviceBarrier() const {
+    bool CanUseDeviceBarrier = getModule()->isAllowedToUseExtension(
+        ExtensionID::SPV_INTEL_device_barrier);
```
MrSidims: May be to avoid breaking changes rewrite the function like this?
```suggestion
    if (!getModule()->isAllowedToUseExtension(ExtensionID::SPV_INTEL_device_barrier))
      return false;
```

lib/SPIRV/libSPIRV/SPIRVInstruction.h
bashbaug: Hmm, I don't think this is a breaking change.  There are two characteristics to consider:

1. Whether the device barrier extension is enabled or not.
2. Whether the execution scope is a constant or not.

If the device barrier extension is disabled (`CanUseDeviceBarrier` is `false`), then `isDeviceBarrier` returns `false`, all good.

Otherwise, when the device barrier extension is enabled (`CanUseDeviceBarrier` is `true`), we need to look at the execution scope.  There are three possibilities:

1. If the execution scope is a constant and it is the device scope, then this is a device barrier, return `true`.
    * Note, this is also an error, if the device barrier extension is NOT enabled.
2. If the execution scope is a constant and it is not the device scope, then this is NOT a device barrier, return `false`.
3. If the execution scope is not a constant, we need to be conservative, since it might be a device barrier.  In this case, we return `CanUseDeviceBarrier`, which is `true`.

Does this logic make sense?

MrSidims: > Does this logic make sense?

Well, kind of (in a sense, that my suggestion to add an early exit actually doesn't help).

Basically, all falls to a check if `SPV_INTEL_device_barrier` is enabled by compiler's frontend. If it's enabled, then we must be sure, that all (targeted) device drivers support the extension, as all runtime or specialization constant scopes would result in a new capability to appear in the module even when device scope is not actually used. Probably, at this point of time we are okay with such behaviour, as we are cautions with enabling new extensions.

bashbaug: Right, I guess we could consider being conservative in the other direction in (3), and NOT declare the extension or the capability for non-constant scopes, but this doesn't feel correct.

If we're told that we're allowed to use the extension, I think we shouldn't shy away from using it.

bashbaug:
> are we ready to merge the change?

Yes, I believe so.  I was holding off for a bit just in case I could switch to the upstream headers vs. relying on spirv_internal.hpp, but I think it'd be best to merge the changes in their current state, then I will switch to the upstream headers as soon as support is available.

vmaksimo:
@bashbaug when you have time - please backport to 22 as well, CI there is stable now

bashbaug:
> please backport to 22 as well, CI there is stable now

Looks like the `llvm_release_220` branch already has this support - can you please confirm?  If so, no backport is needed.  Thanks!

https://github.com/KhronosGroup/SPIRV-LLVM-Translator/blob/llvm_release_220/include/LLVMSPIRVExtensions.inc#L60

PR #3467, opened 2025-12-03
Implement SPV_KHR_fma extension

vmaksimo:
The extension adds support for the `OpFmaKHR` instruction, which provides a native SPIR-V instruction for fused multiply-add operations as an alternative to using OpenCL.std::Fma extended instruction.

Translate both LLVM fma intrinsics as well as OCL builtins to `OpFmaKHR` if the extension is available.

Specification:
https://github.khronos.org/SPIRV-Registry/extensions/KHR/SPV_KHR_fma.html

### DIFF
--- a/test/extensions/KHR/SPV_KHR_fma/fma.ll
+++ b/test/extensions/KHR/SPV_KHR_fma/fma.ll
@@ -1,4 +1,6 @@
 ; RUN: llvm-spirv %s --spirv-ext=+SPV_KHR_fma -o %t.spv
+; TODO: enable once spirv-val supports the extension.
+; RUNx: spirv-val %t.spv
 ; RUN: llvm-spirv %t.spv -to-text -o %t.spt
 ; RUN: FileCheck < %t.spt %s --check-prefix=CHECK-SPIRV
 ; RUN: llvm-spirv -r %t.spv -o - | llvm-dis -o %t.rev.ll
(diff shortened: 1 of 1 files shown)


test/extensions/KHR/SPV_KHR_fma/fma.ll
```diff
@@ -0,0 +1,57 @@
...
+  ret float %result
+}
+
+define spir_func <4 x float> @test_fma_vector(<4 x float> %a, <4 x float> %b, <4 x float> %c) {
+entry:
+  %result = call <4 x float> @llvm.fma.v4f32(<4 x float> %a, <4 x float> %b, <4 x float> %c)
+  ret <4 x float> %result
+}
+
+; case to test fma called via OCL builtins
```
maarquitos14:
```suggestion
; Case to test fma called via OCL builtins.
```

test/extensions/KHR/SPV_KHR_fma/fma.ll
```diff
@@ -0,0 +1,57 @@
+; RUN: llvm-spirv %s --spirv-ext=+SPV_KHR_fma -o %t.spv
```
maarquitos14, line 1: No spirv-val for this one? Is it because it still doesn't recognise FMA? If so, should we add the line commented out with a TODO/FIXME?

vmaksimo: Yeah, the validation was added only ~1 month ago, will take some time to reach our CI. 
Will add todo

Issue #3125, opened 2025-04-19
Fix reassoc flag translation


MrSidims:
See https://github.com/KhronosGroup/SPIRV-Registry/issues/326 for details

The translation should be:
reassoc -> AllowTransform -> reassoc

PR #3475, opened 2025-12-17
Implement extension SPV_KHR_float_controls2

jmmartinez:
First attempt at implementing [SPV_KHR_float_controls2](https://github.khronos.org/SPIRV-Registry/extensions/KHR/SPV_KHR_float_controls2.html).

- When doing SPIRV->LLVM-IR, we first read the `ExecutionModeFPFastMathDefault` for every kernel, and if instructions in that kernel do not specify a particular `FPFastMathMode`, we use the kernel one.
- `ExecutionModeFPFastMathDefault` is **not** propagated down to the callees.
- According to [SPV_KHR_float_controls2#issues](https://github.khronos.org/SPIRV-Registry/extensions/KHR/SPV_KHR_float_controls2.html#_issues); we do not have an equivalent of LLVM's `afn` flag. If we map `fadd fast float %a, %b` to SPIRV and back, it becomes `fadd reassoc nnan ninf nsz arcp contract float %a, %b` losing the `afn` flag.
- `ContractionOff` and `SignedZeroInfNanPreserve` are translated to a `ExecutionModeFPFastMathDefault` with all flags set to 0.

### DIFF
--- a/lib/SPIRV/SPIRVWriter.cpp
+++ b/lib/SPIRV/SPIRVWriter.cpp
@@ -6670,10 +6670,10 @@ bool LLVMToSPIRVBase::transExecutionMode() {
         assert(F);
         BM->addCapability(CapabilityFloatControls2);
         BM->addExtension(ExtensionID::SPV_KHR_float_controls2);
-        PoisonValue *V;
+        PoisonValue *V = nullptr;
         unsigned FlagsLiteral;
         N.get(V).get(FlagsLiteral);
-        assert(V->getType()->isFloatingPointTy());
+        assert(V && V->getType()->isFloatingPointTy());
         SPIRVTypeFloat *FloatSPIRVType =
             static_cast<SPIRVTypeFloat *>(transType(V->getType()));
         setExecutionModeFPFastMathDefault(BF, FloatSPIRVType, FlagsLiteral);


test/transcoding/fadd.ll
```diff
@@ -1,8 +1,12 @@
 ; RUN: llvm-as %s -o %t.bc
-; RUN: llvm-spirv -spirv-text %t.bc -o - | FileCheck %s --check-prefix=CHECK-SPIRV
+; RUN: llvm-spirv -spirv-text %t.bc -o - | FileCheck %s --check-prefixes=CHECK-SPIRV,CHECK-SPIRV-DEFAULT
```
jmmartinez, line 2: Don't forget to add `fcmp` tests!

jmmartinez:
> > I've added a commit related to this, but I'll file a separate patch since this issue is not related to the float_controls2 extension.
> 
> Fine with me.
> 
> Most (if not all) of the folks working on the translator are currently on holidays (including myself), so guess review will be done a bit later :)
> 
> (unless there is a super urgency - in this case I can take a look before New Year)

No problem! It's not urgent.

lib/SPIRV/SPIRVWriter.cpp
```diff
@@ -3118,6 +3118,12 @@ static void transMetadataDecorations(Metadata *MD, SPIRVValue *Target) {
           static_cast<StoreCacheControl>(CacheControl->getZExtValue())));
       break;
     }
+    case DecorationFPFastMathMode: {
+      // Ignore this decoration. FPFastMathMode is set through the fast-math flags associated with the instruction.
+      // It should not be set through metadata.
+      // See https://github.com/KhronosGroup/SPIRV-LLVM-Translator/issues/3410
```
MrSidims: I think the link is a bit misleading. The main reason of not expecting this decoration metadata in LLVM IR is that LLVM opt can do whatever it wants to unknown metadata attached to an instruction, hence semantically important info shouldn't be passed like this.

lib/SPIRV/SPIRVWriter.cpp
```diff
@@ -6545,11 +6589,50 @@ bool LLVMToSPIRVBase::transExecutionMode() {
...
+  SmallVector<SPIRVType *> AllFPTypes;
+  for (auto [_, SPVTy] : TypeMap) {
+    if (!SPVTy->isTypeFloat())
+      continue;
+    AllFPTypes.push_back(SPVTy);
+  }
+
+  // We encode an fp-operaiton with no FPFastMathMode flags set as an
+  // fp-operation with all the flags set to 0. Instead of setting the flag for
+  // every individual operation, we set it once, for the entry-point.
```
MrSidims: I believe spec doesn't mandate setting FPFastMathMode if FP Fast Math Mode is None. Though it might be a good sanity check. So may be worth leaving it as is, but note, that currently the patch is not checking that a particular type is actually used in a kernel.

lib/SPIRV/SPIRVWriter.cpp
```diff
@@ -6545,11 +6589,50 @@ bool LLVMToSPIRVBase::transExecutionMode() {
...
 
   transFPContract();
+  transFPFastMathDefault();
 
   return true;
 }
 
+void LLVMToSPIRVBase::transFPFastMathDefault() {
+  if (!BM->hasCapability(CapabilityFloatControls2))
+    return;
```
MrSidims: Current logic seem to be the following:
**If** during fast math flag translation we ended up generating some decorations (and hence capabilities) **then** enter this function and add None FP Mode to the entry point.

I believe the logic should be instead:
After translation of fast math flags we enter this function anyway if `SPV_KHR_float_controls2` extension is enabled. Then we should check, if [spirv.ExecutionMode](https://github.com/KhronosGroup/SPIRV-LLVM-Translator/blob/main/docs/SPIRVRepresentationInLLVM.rst#spir-v-instructions-mapped-to-llvm-metadata) metadata (that encodes FP Fast Math Mode) is attached to a kernel and then add the appropriate `OpExecutionMode`.

MrSidims: May be it can go to a separate patch.

lib/SPIRV/SPIRVWriter.cpp
```diff
@@ -6545,11 +6589,50 @@ bool LLVMToSPIRVBase::transExecutionMode() {
...
+
+  SmallVector<SPIRVType *> AllFPTypes;
+  for (auto [_, SPVTy] : TypeMap) {
```
MrSidims: nit:
```suggestion
  for (auto [_, SPVTy] : FloatTypeMap) {
```

test/extensions/KHR/SPV_KHR_float_controls2/execution_mode_default.ll
```diff
@@ -0,0 +1,37 @@
+; RUN: llvm-spirv --spirv-ext=+SPV_KHR_float_controls2 -spirv-text %s -o - | FileCheck %s
+; RUN: llvm-spirv --spirv-ext=+SPV_KHR_float_controls2 %s -o %t.spv
+; RUN: spirv-val %t.spv
```
MrSidims, line 3: Lets add reverse translation checks as well.

test/extensions/KHR/SPV_KHR_float_controls2/execution_mode_default.ll
```diff
@@ -0,0 +1,37 @@
...
+
+; CHECK: 4 EntryPoint {{[0-9]+}} [[#foo:]] "foo"
+; CHECK-DAG: 3 Name [[#bar:]] "bar"
```
MrSidims: for new tests lets not add checks for number of words in the instructions.
same comment for other lines

```suggestion
; CHECK: EntryPoint {{[0-9]+}} [[#foo:]] "foo"
; CHECK-DAG: Name [[#bar:]] "bar"
```

test/transcoding/fadd.ll
```diff
@@ -11,13 +15,20 @@
 ; CHECK-SPIRV: 3 Name [[#r5:]] "r5"
 ; CHECK-SPIRV: 3 Name [[#r6:]] "r6"
 ; CHECK-SPIRV: 3 Name [[#r7:]] "r7"
+; CHECK-SPIRV: 3 Name [[#r8:]] "r8"
```
MrSidims, line 18: (here it's fine to have word count check to not destruct the existing style of the test)

test/fp-decorate-twice.ll
```diff
@@ -0,0 +1,19 @@
+; RUN: llvm-spirv -spirv-text %s -o - | FileCheck %s
+; RUN: llvm-spirv %s -o %t.spv
+; RUN: spirv-val %t.spv
+;
+; CHECK: 4 Name [[#CMP:]] "oltRes"
+; CHECK-DAG: 5 FOrdLessThan [[#BOOL:]] [[#CMP]]
+; CHECK-DAG: 4 Decorate [[#CMP]] FPFastMathMode 1
+; CHECK-NOT: 4 Decorate [[#CMP]] FPFastMathMode 16
```
MrSidims:
```suggestion
; CHECK: Name [[#CMP:]] "oltRes"
; CHECK-DAG: FOrdLessThan [[#BOOL:]] [[#CMP]]
; CHECK-DAG: Decorate [[#CMP]] FPFastMathMode 1
; CHECK-NOT: Decorate [[#CMP]] FPFastMathMode 16
```

lib/SPIRV/SPIRVWriter.cpp
jmmartinez: I don't think that this is what we want. This is a member variable of the class.

```cpp
namespace SPIRV {

class LLVMToSPIRVBase : protected BuiltinCallHelper {
  /*...*/
private:
  // This maps LLVM types (except for pointers) to SPIRVType.
  LLVMToSPIRVTypeMap TypeMap;
```

lib/SPIRV/SPIRVWriter.cpp
jmmartinez: I've updated the message in https://github.com/KhronosGroup/SPIRV-LLVM-Translator/pull/3475/changes/8fd4724d3aa4f0b94f15a7a5c040b56d5cd30943

lib/SPIRV/SPIRVWriter.cpp
jmmartinez: Since we're early enough, I added tests in https://github.com/KhronosGroup/SPIRV-LLVM-Translator/pull/3475/changes/924debb8e9a781384cd9110556c1305eaf3270f6 and modified the function in https://github.com/KhronosGroup/SPIRV-LLVM-Translator/pull/3475/changes/8fa049e511a76c2b9b22e06eeba8e20c84917b3e .

Now, if the extension is available and we're using floating-point types, we add the `DefaultFPFastMathFlags`.

jmmartinez:
Just in case, I'd like to bring the attention to one of my previous messages about the issue https://github.com/KhronosGroup/SPIRV-LLVM-Translator/issues/3125 :

Currently, this PR maps LLVM's `reassoc -> AllowReassoc` (this is the behavior that was implemented before `float_controls2`).

In the issue it is suggested that we'd better translate `reassoc -> AllowTransform`. The problem with this is that `AllowTransform` implies both `AllowContract` and `AllowReassoc`.

Then, if we map LLVM's to SPIRV and back to LLVM we end up with different semantics:

```
reassoc -> AllowTransform AllowContract AllowReassoc -> contract reassoc
```

To avoid this, we could translate 

```
reassoc -> AllowReassoc -> no-flags
contract reassoc -> AllowTransform AllowContract AllowReassoc -> contract reassoc
```

lib/SPIRV/SPIRVReader.cpp
```diff
@@ -3443,6 +3439,58 @@ static void validatePhiPredecessors(Function *F) {
...
+    FMF.setAllowReciprocal();
+  static_assert(FPFastMathModeAllowContractFastINTELMask ==
+                FPFastMathModeAllowContractMask);
+  if (V & FPFastMathModeAllowContractFastINTELMask)
+    FMF.setAllowContract();
+  static_assert(FPFastMathModeAllowReassocINTELMask ==
+                FPFastMathModeAllowReassocMask);
+  if (V & FPFastMathModeAllowReassocINTELMask)
+    FMF.setAllowReassoc();
+  // There is no FPFastMathMode flag that represents LLVM approxiamte functions
```
maarquitos14:
```suggestion
  // There is no FPFastMathMode flag that represents LLVM approximate functions
```

lib/SPIRV/SPIRVReader.h
```diff
@@ -274,6 +274,14 @@ class SPIRVToLLVM : private BuiltinCallHelper {
   void
   transFunctionPointerCallArgumentAttributes(SPIRVValue *BV, CallInst *CI,
                                              SPIRVTypeFunction *CalledFnTy);
+
+  using FunctionAndTypeIdPair = std::pair<Function *, SPIRVWord>;
+  using FunctionToFastMathFlagsMap =
+      DenseMap<FunctionAndTypeIdPair, FastMathFlags>;
+  FunctionToFastMathFlagsMap Func2FastMathFlags;
```
maarquitos14:
```suggestion
  FunctionToFastMathFlagsMap FuncToFastMathFlags;
```

Feel free to ignore, it's just a nit.

lib/SPIRV/SPIRVReader.cpp
```diff
@@ -1218,26 +1218,22 @@ static void applyNoIntegerWrapDecorations(const SPIRVValue *BV,
...
-      FMF.setFast();
+    FastMathFlags FMF = translateFastMathFlags(V);
     Inst->setFastMathFlags(FMF);
+    return;
+  }
+
+  SPIRVWord TypeId = BV->getType()->getId();
+  auto Func2FMF = Func2FastMathFlags.find({Inst->getFunction(), TypeId});
+  if (Func2FMF != Func2FastMathFlags.end()) {
+    Inst->setFastMathFlags(Func2FMF->second);
```
maarquitos14: This is intended to add the flags specified by FPFastMathDefault to every instruction in the kernel, right?

lib/SPIRV/SPIRVWriter.cpp
```diff
@@ -3118,6 +3118,12 @@ static void transMetadataDecorations(Metadata *MD, SPIRVValue *Target) {
           static_cast<StoreCacheControl>(CacheControl->getZExtValue())));
       break;
     }
+    case DecorationFPFastMathMode: {
+      // Ignore this decoration. FPFastMathMode is set through the fast-math
+      // flags associated with the instruction. It should not be set through
```
maarquitos14:
```suggestion
      // Ignore this decoration. FPFastMathMode is set through the LLVM IR fast-math
      // flags (e.g. reassoc, contract) associated with the instruction. It should not be set through
```
This just to make this easier to understand for people not so familiar with FP flags. Feel free to ignore or rephrase.

lib/SPIRV/SPIRVWriter.cpp
```diff
@@ -6421,8 +6463,10 @@ bool LLVMToSPIRVBase::transExecutionMode() {
 
       switch (EMode) {
       case spv::ExecutionModeContractionOff:
-        BF->addExecutionMode(BM->add(new SPIRVExecutionMode(
-            OpExecutionMode, BF, static_cast<ExecutionMode>(EMode))));
+        // With SPV_KHR_float_controls2 this is deprecated
```
maarquitos14:
```suggestion
        // With SPV_KHR_float_controls2 this is deprecated.
```

lib/SPIRV/SPIRVWriter.cpp
```diff
@@ -6421,8 +6463,10 @@ bool LLVMToSPIRVBase::transExecutionMode() {
...
+        if (!BM->hasCapability(CapabilityFloatControls2))
```
maarquitos14: Shouldn't we add `FPFastMathDefault` execution mode when `CapabilityFloatControls2` is enabled? It is required to set the equivalent of `ContractionOff`, isn't it?

lib/SPIRV/SPIRVWriter.cpp
```diff
@@ -6509,9 +6553,13 @@ bool LLVMToSPIRVBase::transExecutionMode() {
         BM->addCapability(CapabilityVectorComputeINTEL);
       } break;
 
+      case spv::ExecutionModeSignedZeroInfNanPreserve:
+        // With SPV_KHR_float_controls2 this is deprecated
```
maarquitos14:
```suggestion
        // With SPV_KHR_float_controls2 this is deprecated.
```

lib/SPIRV/SPIRVWriter.cpp
```diff
@@ -6509,9 +6553,13 @@ bool LLVMToSPIRVBase::transExecutionMode() {
...
+        if (BM->hasCapability(CapabilityFloatControls2))
```
maarquitos14: Don't we need to add `FPFastMathDefault` execution mode too? It is required to set the equivalent of `SignedZeroInfNanPreserve`, isn't it?

lib/SPIRV/SPIRVWriter.cpp
```diff
@@ -6549,7 +6597,51 @@ bool LLVMToSPIRVBase::transExecutionMode() {
   return true;
 }
 
+void LLVMToSPIRVBase::transFPFastMathDefault() {
+  if (!BM->isAllowedToUseExtension(ExtensionID::SPV_KHR_float_controls2))
+    return;
+
+  SmallVector<SPIRVType *> AllFPTypes;
+  for (auto [_, SPVTy] : TypeMap) {
+    if (!SPVTy->isTypeFloat())
```
maarquitos14: Does this work for vectors? Should we use `isTypeVectorOrScalarFloat()` instead?

lib/SPIRV/libSPIRV/SPIRVEntry.h
```diff
@@ -668,6 +668,15 @@ class SPIRVExecutionMode : public SPIRVAnnotation {
     WordLiterals.push_back(Z);
     updateModuleVersion();
   }
+  // Complete constructor for FPFastMathDefault
```
maarquitos14:
```suggestion
  // Complete constructor for FPFastMathDefault.
```

lib/SPIRV/libSPIRV/SPIRVEntry.h
```diff
@@ -668,6 +668,15 @@ class SPIRVExecutionMode : public SPIRVAnnotation {
...
+  SPIRVExecutionMode(Op OC, SPIRVEntry *TheTarget,
+                     SPIRVExecutionModeKind TheExecMode, SPIRVWord X,
+                     SPIRVWord Y)
```
maarquitos14:
```suggestion
                     SPIRVExecutionModeKind TheExecMode, SPIRVWord TargetType,
                     SPIRVWord FastMathMode)
```

Nit.

maarquitos14: Actually, I think this isn't allowed. Extra operands for `FPFastMathDefault` are ids, according to the spec, so as far as I know, it should only be allowed with `OpExecutionModeId`. Am I right?

lib/SPIRV/libSPIRV/SPIRVEntry.h
```diff
@@ -728,6 +737,13 @@ class SPIRVExecutionModeId : public SPIRVExecutionMode {
       : SPIRVExecutionMode(OpExecutionModeId, TheTarget, TheExecMode, X, Y, Z) {
     updateModuleVersion();
   }
+  // Complete constructor for FPFastMathDefault
```
maarquitos14: The nit from before applies here as well.

maarquitos14 (commented):
First pass. I'll do a second pass to check tests.

lib/SPIRV/libSPIRV/SPIRVEntry.h
jmmartinez: You're right. This code is odd because `SPIRVExecutionModeId` derives from `SPIRVExecutionMode`. I've added some assertions in [1bf546](https://github.com/KhronosGroup/SPIRV-LLVM-Translator/pull/3475/commits/1bf5462b33e1e4a7ce2ca73b7ad417629f20ae65) to highlight this.

jmmartinez: In fact, I'm tempted to use something like tag-dispatching to enforce at compile-time the signatures of the constructors being called.

Something like:

```cpp
template<SPIRVExecutionModeKind K>
using SPIRVExecutionModeKindK = std::integral_constant<SPIRVExecutionModeKind, K>;
template<Op OC>
using SPIRVOpK = std::integral_constant<Op, OC>;

/*...*/

// Complete constructor for FPFastMathDefault.
SPIRVExecutionMode(SPIRVOpK<OpExecutionModeId>, SPIRVEntry *TheTarget,
                   SPIRVExecutionModeKindK<spv::ExecutionModeFPFastMathDefault>, SPIRVWord TargetType,
                   SPIRVWord FastMathMode)
    : SPIRVAnnotation(OpExecutionModeId, TheTarget, 5), ExecMode(ExecutionModeFPFastMathDefault) {
  WordLiterals.push_back(TargetType);
  WordLiterals.push_back(FastMathMode);
  updateModuleVersion();
}

/*...*/

// Complete constructor for FPFastMathDefault.
SPIRVExecutionModeId(SPIRVEntry *TheTarget,
                     SPIRVExecutionModeKindK<ExecutionModeFPFastMathDefault>,SPIRVWord TargetType,
                     SPIRVWord FastMathMode)
    : SPIRVExecutionMode(SPIRVOpK<OpExecutionModeId>{}, TheTarget, SPIRVExecutionModeKindK<spv::ExecutionModeFPFastMathDefault>{},
                         TargetType, FastMathMode) {
  updateModuleVersion();
}
```

This also applies for the other constructors probably.

lib/SPIRV/SPIRVReader.cpp
jmmartinez: Yes. I'm a bit blocked about how to handle function calls to non-kernels though.

For the moment the no flags are set for those functions; but we could intersect all the reachable `FPFastMathDefault` flags and use those.

lib/SPIRV/SPIRVWriter.cpp
jmmartinez: At the moment, since the default fast-math flags are all disabled, both are preserved (`ContractionOff/SignedZeroInfNanPreserve` disable the `contract/nsz ninf nnan` flags).

I should add a comment explaining how these are preserved.

lib/SPIRV/SPIRVWriter.cpp
jmmartinez: No, it should be a scalar floating point type. However, when reading SPIRV code this is not adding the flags to vector instructions in `SPIRVToLLVM::applyFPFastMathModeDecorations`.

I should fix that.

lib/SPIRV/SPIRVReader.cpp
maarquitos14: I thought we were going with not propagating flags down to the callees. Anyway, I'm happy to address that in a follow up patch to not block this one anymore.

lib/SPIRV/SPIRVWriter.cpp
maarquitos14: Okay, I see what you mean. However, I vaguely recall that having no flags isn't the same as having all flags set to zero from my implementation of this extension in the SPIRV BE. Let me try and find that again. 

Also, a comment would help anyway :)

lib/SPIRV/libSPIRV/SPIRVEntry.h
maarquitos14: I'm happy with your current fix, but feel free to go with your suggestion for extra safety.

lib/SPIRV/SPIRVWriter.cpp
maarquitos14: You're right: https://github.com/llvm/llvm-project/pull/146941#discussion_r2257391634. Ironically, it was myself making the same point you made now in the SPIRV BE implementation. I still haven't reviewed tests, but in case you didn't yet, please, make sure to add tests for vectors.

lib/SPIRV/SPIRVWriter.cpp
maarquitos14: > If an operation is decorated with FPFastMathMode then the flags from that decoration apply. Otherwise, if the current entry point sets any FPFastMathDefault execution mode then all flags specified for any operand type or for the result type of the operation apply. If the operation is not decorated with FPFastMathMode and the entry point sets no FPFastMathDefault execution modes then the flags to be applied are determined by the client API and not by SPIR-V.

My understanding of this quote from the spec is that no decoration is not the same than decoration with all flags set to zero: all flags set to zero clearly specify the fast math mode, while no decoration means the client API can decide. Do you agree?

test/extensions/KHR/SPV_KHR_float_controls2/execution_mode_default.ll
```diff
@@ -0,0 +1,49 @@
...
+; SPIRV-DAG: ExecutionModeId [[#foo]] 6028 [[#double]] [[#zero]]
+; SPIRV-NOT: ExecutionModeId [[#bar]] 6028
+
+target triple = "spirv-unknown-unknown"
+
+define spir_kernel void @foo(half %ah, half %bh, float %af, float %bf, double %ad, double %bd) {
+entry:
+  ; IR-LABEL: define {{.*}} @foo
+  ; IR-NEXT: entry:
+  ; IR-NEXT:   %rh = fadd contract half %ah, %bh
```
maarquitos14: My understanding is that you don't check decorations in SPIRV because you assume that they have to be present in SPIRV if they are present in the reverse translation. Am I right?

test/extensions/KHR/SPV_KHR_float_controls2/execution_mode_id.spvasm
```diff
@@ -0,0 +1,57 @@
...
+; RUN: spirv-as -o %t.spv %s
+; RUN: spirv-val %t.spv
+; RUN: llvm-spirv -r -o %t.bc %t.spv
+; RUN: llvm-dis %t.bc -o - | FileCheck %s --check-prefix=LLVM
+; RUN: llvm-spirv --spirv-ext=+SPV_KHR_float_controls2 %t.bc -o - --spirv-text | FileCheck %s --check-prefix=SPIRV
+; RUN: llvm-spirv --spirv-ext=+SPV_KHR_float_controls2 %t.bc -o %t.back.spv
+; RUN: spirv-val %t.back.spv
+;
+; Check that the FPFastMathDefault set to "contract" on a function is honored
+; when translating from SPIRV to LLVM-IR
```
maarquitos14:
```suggestion
; when translating from SPIRV to LLVM-IR.
```

lib/SPIRV/SPIRVWriter.cpp
```diff
@@ -6549,7 +6598,51 @@ bool LLVMToSPIRVBase::transExecutionMode() {
...
+    AllFPTypes.push_back(SPVTy);
+  }
+
+  if (AllFPTypes.empty())
+    return;
+
+  BM->addCapability(CapabilityFloatControls2);
+  BM->addExtension(ExtensionID::SPV_KHR_float_controls2);
+
+  // We encode an fp-operaiton with no FPFastMathMode flags set as an
```
maarquitos14:
```suggestion
  // We encode an fp-operation with no FPFastMathMode flags set as an
```

lib/SPIRV/SPIRVWriter.cpp
```diff
@@ -6549,7 +6598,51 @@ bool LLVMToSPIRVBase::transExecutionMode() {
...
+  // fp-operation with all the flags set to 0. Instead of setting the flag for
+  // every individual operation, we set it once, for the entry-point.
```
maarquitos14: I'm not really sure if this is correct. Like I said in a different thread, I don't think it's the same having all flags set to 0 than not having the decoration.

test/extensions/KHR/SPV_KHR_float_controls2/execution_mode_default.ll
```diff
@@ -0,0 +1,49 @@
...
+; SPIRV-DAG: Name [[#bar:]] "bar"
+;
+; SPIRV-DAG: Constant {{[0-9]+}} [[#zero:]] 0
+;
+; SPIRV-DAG: TypeFloat [[#half:]] 16
+; SPIRV-DAG: TypeFloat [[#float:]] 32
+; SPIRV-DAG: TypeFloat [[#double:]] 64
+;
+; 6028 is FPFastMathDefault 
+; SPIRV-DAG: ExecutionModeId [[#foo]] 6028 [[#half]] [[#zero]]
```
maarquitos14: It's not clear to me where is the `OpExecutionModeId` coming from. I don't see any `!spirv.ExecutionMode` metadata.

test/extensions/KHR/SPV_KHR_float_controls2/extension_not_needed.ll
```diff
@@ -0,0 +1,16 @@
+; RUN: llvm-spirv --spirv-ext=+SPV_KHR_float_controls2 -spirv-text %s -o - | FileCheck %s --check-prefix=SPIRV
+; RUN: llvm-spirv --spirv-ext=+SPV_KHR_float_controls2 %s -o %t.spv
+; RUN: spirv-val %t.spv
+
+; Do not add extension if no floating-point type is used in this module.
```
maarquitos14: What if there is an fp type but it doesn't have any flags, or there is none of the FP operations? Does it enable the extension in that case? Should it enable the extension in that case?

test/extensions/KHR/SPV_KHR_float_controls2/extension_not_needed_but_used.ll
```diff
@@ -0,0 +1,40 @@
...
+;
+; SPIRV: EntryPoint {{[0-9]+}} [[#foo:]] "foo"
+; SPIRV-DAG: Name [[#bar:]] "bar"
+;
+; SPIRV-DAG: Constant {{[0-9]+}} [[#zero:]] 0
+;
+; SPIRV-DAG: TypeFloat [[#float:]] 32
+;
+; 6028 is FPFastMathDefault 
+; SPIRV-DAG: ExecutionModeId [[#foo]] 6028 [[#float]] [[#zero]]
```
maarquitos14: I'm a bit lost here: the test is named `extension_not_needed_but_used`, however, you expect `FPFastMathDefault` execution mode, which does need the extension, I think?

At the same time, similarly to a previous test, I don't see any `!spirv.ExecutionMode` metadata, so I'm not sure why we expect this execution mode.

lib/SPIRV/SPIRVWriter.cpp
jmmartinez: I've added some test for vectors https://github.com/KhronosGroup/SPIRV-LLVM-Translator/pull/3475/changes/73622cdaed0211a4cb09fa4947bdb4fab30d6e5a

jmmartinez: I still have to check how the matrix type are handled

test/extensions/KHR/SPV_KHR_float_controls2/execution_mode_id.spvasm
jmmartinez: I'm so sorry. I think you corrected stuff like this on my PRs a hundred times already.

lib/SPIRV/SPIRVWriter.cpp
MrSidims: Yeah, I wanted to make the same comment (again) here after the first review, but this time I realized, that only scalar floats should be annotated :D

> I still have to check how the matrix type are handled

TypeMatrix is shader only, it's not handled by the translator.

test/extensions/KHR/SPV_KHR_float_controls2/execution_mode_id.spvasm
maarquitos14: Haha no worries. Tbh, this is the stuff I find most frequently in reviews.

lib/SPIRV/SPIRVWriter.cpp
jmmartinez: I agree.

Currently, if `float_controls2` is available, we enable it **always** with all the flags set to zero. Then an LLVM floating-point operation with no flags has the same semantics in SPIRV.
However, I think there is a problem in my implementation: functions getting called by kernels.

Currently the `FastMathModeDefault` are not preserved when doing `spirv->llvm-ir->spirv`. When doing `spirv->llvm-ir` we set the `FastMathModeDefault` into the kernel operations, but we cannot do that on the called functions. Then, when doing `llvm-ir->spirv` we end up with the right flags on the kernel, but stricter flags (all set to 0 propagated through the new `FastMathModeDefault`) on the called function.

test/extensions/KHR/SPV_KHR_float_controls2/execution_mode_default.ll
jmmartinez: Sort of. I wanted to check only that the `ExecutionModeId` was set correctly (the flags set on the instructions are verified in other tests). And reverse translated it to ensure the `contract` flag doesn't get overridden by it.

I can add the checks for the individual instructions if it make more sense.

lib/SPIRV/SPIRVWriter.cpp
jmmartinez: Just a second contradicting thought. In fact, depending on how you see it, setting no flags in SPIRV can also be seen as enabling all rewrite flags in LLVM: `contract / reassociate / ...` are all permitted and is up to the client to decide if it optimizes it or not.

test/extensions/KHR/SPV_KHR_float_controls2/execution_mode_default.ll
jmmartinez: At the moment this patch is not setting `!spirv.ExecutionMode` and it always sets the `FastMathDefaultFlags` to `0` (which is probably too strict).

lib/SPIRV/libSPIRV/SPIRVEntry.h
jmmartinez: I'll do this as a follow up probably, to avoid adding noise to this PR

lib/SPIRV/SPIRVWriter.cpp
maarquitos14: Exactly, setting everything to zero might prevent possible client optimizations. I think we shouldn't do that.

test/extensions/KHR/SPV_KHR_float_controls2/execution_mode_default.ll
maarquitos14: As long as the intent is clearly specified in the test, I'm happy with that. Can you add a comment explaining this?

test/extensions/KHR/SPV_KHR_float_controls2/execution_mode_default.ll
maarquitos14: So where does the `OpExecutionMode` come from, then?

jmmartinez: At the moment, it's always setting it to 0 by default in SPIRV. So even if the decoration doesn't appear in the LLVM-IR, if we allow `SPV_KHR_float_controls2`, it is added to every kernel.

maarquitos14: Okay, I understand. I have doubts about such an approach for the reasons mentioned in another thread: we would be setting constraints to kernels that actually don't come from the original source. Without those constraints, client APIs have more freedom to optimize. @MrSidims what's your opinion on this?

MrSidims: My understanding, is that **None** `FP Fast Math Mode` set in `FPFastMathDefault` is not overriding default controls of client API. @bashbaug @alan-baker is my understanding correct?

alan-baker: Any default specified overrides the client. None is saying the default is no fast math (which may match what the client requires).

MrSidims: Thanks for clarification, in such case current behaviour is undesired,

jmmartinez: Sounds good. Then, I'll modify the code such that when translating LLVM-IR to SPIRV we should avoid setting the `FPFastMathDefault` (unless there is metadata specifying it).

In that case a fp-operation without flags in LLVM, `fadd %a %b` will map to a fp-operation without flags in SPIRV (instead of having the flags set to `0` through `FPFastMathDefault`).

---

One case that I still have to reflect more on how to translate `ContractionOff/SignedZeroInfNanPreserve`.

We cannot set an `FPFastMathDefault` or flags on individual instructions such that only some flags are set and the rest are left to the client API.

From what I understand, `float_controls2` deprecates those execution modes, but doesn't forbid them. We could still set them and print a warning as a first step.

What do you think?

maarquitos14: > Then, I'll modify the code such that when translating LLVM-IR to SPIRV we should avoid setting the FPFastMathDefault (unless there is metadata specifying it).

I think that is what we want, yes. You can reuse, or at least cross-check with, the tests in the SPIRV BE: https://github.com/llvm/llvm-project/tree/main/llvm/test/CodeGen/SPIRV/extensions/SPV_KHR_float_controls2 

> From what I understand, float_controls2 deprecates those execution modes, but doesn't forbid them. We could still set them and print a warning as a first step.

What we did in the SPIRV BE was to replace them with the appropriate flags through `FPFastMathDefault`. In that case, I think it's okay to do it, because it comes explicitly from the source. They set `ContractionOff/SignedZeroInfNanPreserve`, and we need to translate that.

lib/SPIRV/SPIRVWriter.cpp
jmmartinez: Updated the PR. With https://github.com/KhronosGroup/SPIRV-LLVM-Translator/pull/3475/changes/f8ace9291ab69c3b646ccbf1f38db3066b8ec03c we preserve the flags.

From SPIRV->LLVM we preserve the `FastMathModeDefault` in the `!spirv.ExecutionMode` metadata. Then when doing LLVM->SPIRV the metadata is lowered into the original `FastMathModeDefault`.

We set the default execution to `0` only when:
* `FastMathModeDefault` in `!spirv.ExecutionMode` was 0 to begin with
* We're adding the `ContractionOff` ExecutionMode
* We're adding the `SignedZeroInfNanPreserve`

For these last 2, since these execution-modes are deprecated with FloatControls2, we have to translate them to something equivalent. We cannot unset some flags and leave others for the client API. For simplicity, I've chosen to set all the flags to 0.
This can be argued though.

In both cases, I've chosen to preserve the fast-math flags that are attached to the instructions. (`Fadd` with `contract` flags will still be translated to `fadd contract` even if `ContracitonOff` was set).

test/extensions/KHR/SPV_KHR_float_controls2/execution_mode_default.ll
jmmartinez: Added:

```
; By default, do not set the execution-mode when the extension is used.
; This test doesn't verify directly that the instructions have the SPIRV
; 'contract' flag (this is done in another test).
; As a sanity check, we still reverse-translate and check the IR.
```

lib/SPIRV/SPIRVWriter.cpp
jmmartinez: All this code is gone.

lib/SPIRV/SPIRVReader.cpp
```diff
@@ -1218,26 +1218,24 @@ static void applyNoIntegerWrapDecorations(const SPIRVValue *BV,
...
-      FMF.setFast();
+    FastMathFlags FMF = translateFastMathFlags(V);
     Inst->setFastMathFlags(FMF);
+    return;
+  }
+
+  // Get the scalar type to handle vector operands. And get the first operand
+  // type (instead of the result) due to fcmp instructions.
+  Type *FloatType = Inst->getOperand(0)->getType()->getScalarType();
+  auto Func2FMF = FuncToFastMathFlags.find({Inst->getFunction(), FloatType});
```
jmmartinez, line 1236: I'm tempted to remove this `FuncToFastMathFlags` stuff.

It's used to set the `FPFastMathFlags` that are attached to the execution mode to the individual instructions of a kernel.

But since we're preserving the `FPFastMathFlags` in the metadata; I'm thinking that this is not needed anymore.

@maarquitos14 should I remove this ?

MrSidims: I believe that this logic should be still placed _somewhere_ as the middleend and backend are unlikely to know about this metadata out of the box and honestly it feels like for optimization passes it's easier to work with individual instruction flags. IMHO resolving `ExecutionMode` to FP flag right away in the SPIR-V consumer won't harm and actually make implementation lower-level drivers friendly.

@svenvh @vmaksimo WDYT?

test/extensions/KHR/SPV_KHR_float_controls2/execution_mode_contract_off.ll
```diff
@@ -0,0 +1,38 @@
...
+; RUN: spirv-val %t.spv
+; RUN: llvm-spirv -r %t.spv -o - | llvm-dis | FileCheck %s --check-prefix=IR
+
+; Check that we do not lower to ContractOff when the FloatControls2 extension is available.
+; Instead, we set FPFastMathDefault to 0.
+;
+; SPIRV-ON: Capability FloatControls2
+; SPIRV: EntryPoint {{[0-9]+}} [[FOO:[0-9]+]] "foo"
+; SPIRV-OFF: ExecutionMode [[FOO]] 31
+; SPIRV-ON-DAG: ExecutionModeId [[FOO]] 6028 [[HALF:[0-9]+]] [[ZERO:[0-9]+]]
```
MrSidims: Lets use `#`-style in new tests
```suggestion
; SPIRV-ON-DAG: ExecutionModeId [[#FOO]] 6028 [[#HALF:]] [[#ZERO:]]
```

test/extensions/KHR/SPV_KHR_float_controls2/execution_mode_contract_off.ll
```diff
@@ -0,0 +1,38 @@
...
+; SPIRV-ON-DAG: TypeFloat [[DOUBLE]] 64
+; SPIRV-ON-DAG: Constant {{[0-9]+}} [[ZERO]] 0
+
+target triple = "spirv-unknown-unknown"
+
+define spir_kernel void @foo(half %ah, half %bh, float %af, float %bf, double %ad, double %bd) {
+entry:
+  ; IR-LABEL: define {{.*}} @foo
+  ; IR-NEXT: entry:
+  ; IR-NEXT: %rh = fadd contract half %ah, %bh
```
MrSidims: Lets don't check for concrete names, either define string vars or check for `%{{.*}}`

test/extensions/KHR/SPV_KHR_float_controls2/execution_mode_contract_off.ll
```diff
@@ -0,0 +1,38 @@
...
+; Check that we do not lower to ContractOff when the FloatControls2 extension is available.
+; Instead, we set FPFastMathDefault to 0.
+;
+; SPIRV-ON: Capability FloatControls2
+; SPIRV: EntryPoint {{[0-9]+}} [[FOO:[0-9]+]] "foo"
+; SPIRV-OFF: ExecutionMode [[FOO]] 31
+; SPIRV-ON-DAG: ExecutionModeId [[FOO]] 6028 [[HALF:[0-9]+]] [[ZERO:[0-9]+]]
+; SPIRV-ON-DAG: ExecutionModeId [[FOO]] 6028 [[FLOAT:[0-9]+]] [[ZERO]]
+; SPIRV-ON-DAG: ExecutionModeId [[FOO]] 6028 [[DOUBLE:[0-9]+]] [[ZERO]]
+; SPIRV-ON-DAG: TypeFloat [[HALF]] 16
```
MrSidims: Test is currently failing, not 100% why. May be we should move this line before checks of ExecutionModeId to ensure correct REGEX variables definition:
```
; SPIRV-ON-DAG: TypeFloat [[#HALF:]] 16
...
; SPIRV-ON-DAG: ExecutionModeId [[#FOO]] 6028 [[#HALF]] [[#ZERO]]
```

MrSidims (approved):
Functionally LGTM, but lets make tests passing :)

test/extensions/KHR/SPV_KHR_float_controls2/execution_mode_contract_off.ll
jmmartinez: It's weird. I haven't managed to reproduce the issue. But I think I know what the problem is: I'm scanning the `TypeMap` and there is no guarantee of the order of its elements. I'll fix this to always generate the same code.

jmmartinez: Fixed in the last commit by sorting the floating-point types by bit-width and encoding.

lib/SPIRV/SPIRVWriter.cpp
maarquitos14: I think this makes perfect sense, and align with the spec:

> If an operation is decorated with FPFastMathMode then the flags from that decoration apply.

lib/SPIRV/SPIRVReader.cpp
maarquitos14: I see @MrSidims's point, but at the same time, I think if we do this, the translation wouldn't be 100% accurate. Maybe I'm too picky. Honestly, I don't have a strong opinion here, let's see what other folks think.

lib/SPIRV/SPIRVMDWalker.h
```diff
@@ -91,7 +91,9 @@ class SPIRVMDWalker {
 
     bool atEnd() const { return !(M && I < E); }
 
-    template <typename T> MDWrapper &get(T &V) {
+    template <typename IntegerType,
```
maarquitos14, line 94: This is a nice and convenient change, but I think it's not too related with the PR. I won't block approval for this, but I think it would be better to have this in a separate PR.

lib/SPIRV/SPIRVMDWalker.h
```diff
@@ -115,12 +117,14 @@ class SPIRVMDWalker {
       return *this;
     }
 
-    MDWrapper &get(Function *&F) {
+    template <typename ValueType,
```
maarquitos14, line 120: Same than above.

lib/SPIRV/SPIRVReader.cpp
```diff
@@ -3443,6 +3441,73 @@ static void validatePhiPredecessors(Function *F) {
...
+  if (V & FPFastMathModeAllowContractFastINTELMask)
+    FMF.setAllowContract();
+  static_assert(FPFastMathModeAllowReassocINTELMask ==
+                FPFastMathModeAllowReassocMask);
+  if (V & FPFastMathModeAllowReassocINTELMask)
+    FMF.setAllowReassoc();
+  // There is no FPFastMathMode flag that represents LLVM approximate functions
+  // flag `afn`. Even the FPFastMathMode Fast flag should not imply it, but to
+  // avoid changing the previous behaviour we make it equivalent to LLVM's.
+  if (V & FPFastMathModeFastMask)
```
maarquitos14: > The FPFastMathMode mode bit Fast. Set all the other FPFastMathMode bits instead.

SPV_KHR_float_controls2 deprecates this bit. If we're using SPV_KHR_float_controls2, should we accept this?

lib/SPIRV/SPIRVWriter.cpp
```diff
@@ -6425,8 +6482,35 @@ bool LLVMToSPIRVBase::transExecutionMode() {
...
+        // With SPV_KHR_float_controls2 this execution mode is deprecated.
+        // We cannot set only the contract flag to 0, so we set all flags to 0.
+        if (BM->isAllowedToUseExtension(ExtensionID::SPV_KHR_float_controls2)) {
+          SmallVector<SPIRVTypeFloat *> FloatSPIRVTypes;
+          for (auto [_, FloatSPIRVType] : TypeMap)
+            if (FloatSPIRVType->isTypeFloat())
+              FloatSPIRVTypes.push_back(
+                  static_cast<SPIRVTypeFloat *>(FloatSPIRVType));
+
+          // Sort the types to ensure they are always emited in a consistent
```
maarquitos14, line 6520: If I understand correctly, we added this to make our testing more consistent. I assume this is not too bad performance-wise because we will never have more than 4 float types?

test/extensions/KHR/SPV_KHR_float_controls2/execution_mode_contract_off.ll
```diff
@@ -0,0 +1,38 @@
+; RUN: llvm-spirv %s -o - --spirv-ext=-SPV_KHR_float_controls2 --spirv-text | FileCheck %s --check-prefixes=SPIRV,SPIRV-OFF
```
maarquitos14: It's just a nit, but I think `FLOAT_CONTROLS2-OFF\ON` would convey the intention better.

test/extensions/KHR/SPV_KHR_float_controls2/execution_mode_contract_off.ll
```diff
@@ -0,0 +1,38 @@
...
+; RUN: llvm-spirv %s -o %t.spv --spirv-ext=+SPV_KHR_float_controls2
+; RUN: spirv-val %t.spv
+; RUN: llvm-spirv -r %t.spv -o - | llvm-dis | FileCheck %s --check-prefix=IR
+
+; Check that we do not lower to ContractOff when the FloatControls2 extension is available.
+; Instead, we set FPFastMathDefault to 0.
+;
+; SPIRV-ON: Capability FloatControls2
+; SPIRV: EntryPoint {{[0-9]+}} [[#FOO:]] "foo"
+; SPIRV-OFF: ExecutionMode [[#FOO]] 31
```
maarquitos14: Should we also check that it is NOT present in the other case? Just to be on the safe side.

test/extensions/KHR/SPV_KHR_float_controls2/execution_mode_id.spvasm
```diff
@@ -0,0 +1,63 @@
...
+; RUN: llvm-spirv -r -o %t.bc %t.spv
+; RUN: llvm-dis %t.bc -o - | FileCheck %s --check-prefix=LLVM
+; RUN: llvm-spirv --spirv-ext=+SPV_KHR_float_controls2 %t.bc -o - --spirv-text | FileCheck %s --check-prefix=SPIRV
+; RUN: llvm-spirv --spirv-ext=+SPV_KHR_float_controls2 %t.bc -o %t.back.spv
+; RUN: spirv-val %t.back.spv
+;
+; Check that the FPFastMathDefault set to "contract" on a function is honored
+; when translating from SPIRV to LLVM-IR.
+;
+; However, when generating SPIRV back again from LLVM-IR set FPFastMathMode
```
maarquitos14, line 13: But we also set the ExecMode through the metadata, don't we? We do both things, don't we? That's what I see in the checks as well.

test/extensions/KHR/SPV_KHR_float_controls2/execution_mode_id_vec.spvasm
```diff
@@ -0,0 +1,53 @@
+; REQUIRES: spirv-as, spirv-dis
+; RUN: spirv-as -o %t.spv %s
+; RUN: spirv-val %t.spv
+; RUN: llvm-spirv -r -o %t.bc %t.spv
+; RUN: llvm-dis %t.bc -o - | FileCheck %s --check-prefix=LLVM
+; RUN: llvm-spirv --spirv-ext=+SPV_KHR_float_controls2 %t.bc -o - --spirv-text | FileCheck %s --check-prefix=SPIRV
+; RUN: llvm-spirv --spirv-ext=+SPV_KHR_float_controls2 %t.bc -o %t.back.spv
+; RUN: spirv-val %t.back.spv
+;
```
maarquitos14, line 9: Please, add a brief description of the test.

test/extensions/KHR/SPV_KHR_float_controls2/execution_mode_signed_zero_inf_nan_preserve.ll
```diff
@@ -0,0 +1,35 @@
+; RUN: llvm-spirv %s -o - -spirv-max-version=1.4 --spirv-ext=-SPV_KHR_float_controls2 --spirv-text | FileCheck %s --check-prefixes=SPIRV,SPIRV-OFF
```
maarquitos14: Same than before with `ON/OFF`.

test/extensions/KHR/SPV_KHR_float_controls2/execution_mode_signed_zero_inf_nan_preserve.ll
```diff
@@ -0,0 +1,35 @@
...
+; RUN: spirv-val %t.spv
+; RUN: llvm-spirv -r %t.spv -o - | llvm-dis | FileCheck %s --check-prefix=IR
+
+; Check that we do not lower to SignedZeroInfNanPreserve when the FloatControls2 extension is available.
+; Instead, we set FPFastMathDefault to 0.
+;
+; SPIRV-OFF: Capability SignedZeroInfNanPreserve
+; SPIRV-ON: Capability FloatControls2
+; SPIRV: EntryPoint {{[0-9]+}} [[#FOO:]] "foo"
+; SPIRV-OFF: ExecutionMode [[#FOO]] 4461 32
```
maarquitos14: Should we also check that it is NOT present in the other case? Just to be on the safe side.

test/extensions/KHR/SPV_KHR_float_controls2/vector_type.ll
```diff
@@ -0,0 +1,32 @@
+; RUN: llvm-spirv --spirv-ext=+SPV_KHR_float_controls2 -spirv-text %s -o - | FileCheck %s --check-prefix=SPIRV
+; RUN: llvm-spirv --spirv-ext=+SPV_KHR_float_controls2 %s -o %t.spv
+; RUN: spirv-val %t.spv
+; RUN: llvm-spirv -r %t.spv -o - | llvm-dis | FileCheck %s --check-prefix=IR
+
```
maarquitos14, line 5: Add a brief description of the test, please.

test/fp-decorate-twice.ll
```diff
@@ -0,0 +1,19 @@
+; RUN: llvm-spirv -spirv-text %s -o - | FileCheck %s
+; RUN: llvm-spirv %s -o %t.spv
+; RUN: spirv-val %t.spv
+;
```
maarquitos14, line 4: Please, add a brief description of the test.

lib/SPIRV/SPIRVMDWalker.h
jmmartinez: I've moved this to https://github.com/KhronosGroup/SPIRV-LLVM-Translator/pull/3557 and I'll rebase this PR over that one.

lib/SPIRV/SPIRVReader.cpp
jmmartinez: I've added a check as an assertion (I like that it enforces the comment with actual code).

In practice, I'd expect this to be enforced by tools ensuring they emit valid SPIRV.

lib/SPIRV/SPIRVWriter.cpp
jmmartinez: It's to make code generation deterministic. Non-deterministic code generation is a pain when there is a bug in the user code, and each time you compile you end up with different code.

Performance-wise, yes, I'd expect this to be very small.

test/extensions/KHR/SPV_KHR_float_controls2/execution_mode_id_vec.spvasm
jmmartinez: I've added a description, and also added checks to see that the ExecutionMode is preserved during the round-trip from SPIRV->LLVM_IR->SPIRV

test/extensions/KHR/SPV_KHR_float_controls2/execution_mode_signed_zero_inf_nan_preserve.ll
jmmartinez: I've added them since it's a simple useful addition. I think in any case it would (or at least it should) have been caught by spirv-val if both were enabled.

jmmartinez:
> Also, in terms of deprecations, the spec says:
> 
> ```
> Deprecation
> This extension deprecates the following features:
> 
> 1. The execution modes ContractionOff and SignedZeroInfNanPreserve. Use FPFastMathDefault with the appropriate flags instead.
> 2. The decoration NoContraction. Use the FPFastMathMode decoration instead.
> 3. The FPFastMathMode mode bit Fast. Set all the other FPFastMathMode bits instead.
> 4. Enabling the FPFastMathMode decoration using the Kernel capability. All uses should declare the FloatControls2 capability.
> 5. The OpenCL.std instructions fmin_common, fmax_common. Use fmin, fmax with NInf and NNaN instead.
> ```
> 
> I think we still need to cover 2. and 5., the rest are covered/not necessary.

I've addressed .5 in the last commit, but I probably have to revisit it.

lib/SPIRV/SPIRVReader.cpp
vmaksimo: I tend to agree with @MrSidims take.

vmaksimo (approved):
Just misclicked earlier :)

Issue #3481, opened 2026-01-05
CI - Build with more than 2 threads


spencer-lunarg:
I noticed in https://github.com/KhronosGroup/SPIRV-LLVM-Translator/blob/main/.github/workflows/check-in-tree-build.yml#L122

it has `make llvm-spirv -j2`

why is this only `2`?

I was looking into how the various `KhronosGroup` repos are using our limited, shared Github CI minutes, the `SPIRV-LLVM-Translator` repo is [among the most](https://github.com/KhronosGroup/SPIRV-LLVM-Translator/actions/metrics/usage?dateRangeType=DATE_RANGE_TYPE_LAST_YEAR) but understand it is a constantly and large contributed repo

I see the CI jobs take about 55 minutes, 50 of which are just building it, looking at the Linux build, I noticed it was only using 2 threads and should just use `ninja` or `make llvm-spirv -j$(nproc)`

Thanks

svenvh:
Reopening as we still have the Windows builds using a hardcoded `-j2`.

vmaksimo:
Sorry, some automation accidentally closed this - we still have the Windows builds using a hardcoded -j2

PR #3483, opened 2026-01-07
[CI] Move away from hardcoded -j2

svenvh:
Instead of the hardcoded `-j2`, use the number of available processing units.

Fixes https://github.com/KhronosGroup/SPIRV-LLVM-Translator/issues/3481

### DIFF
--- a/.github/workflows/check-in-tree-build.yml
+++ b/.github/workflows/check-in-tree-build.yml
@@ -119,11 +119,11 @@ jobs:
       - name: Build
         run: |
           cd build
-          make llvm-spirv -j2
+          make llvm-spirv -j$(nproc)
       - name: Build tests & test
         run: |
           cd build
-          make check-llvm-spirv -j2
+          make check-llvm-spirv -j$(nproc)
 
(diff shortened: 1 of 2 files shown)


spencer-lunarg:
The windows build is still using `-j2` https://github.com/KhronosGroup/SPIRV-LLVM-Translator/actions/runs/20784908272/job/59691612041

image

and probably will get that 50 minute build time down more as well

.github/workflows/check-in-tree-build.yml
spencer-lunarg, line 171: @svenvh 

> I have not figured out yet how to query the number of available processing units on Windows (suggestions appreciated).

```suggestion
          cmake --build . --config ${{ matrix.build_type }} --target llvm-spirv -j
```

I use linux, but according to all the resources I found (and an LLM) it says that `-j` will translate to `/m` for MSVC

svenvh: Thanks, I'll try that with a followup PR.

PR #3498, opened 2026-01-14
Fix umul issue

MrSidims:


MrSidims:
Not sure, why CLA has appeared and why re-signing hasn't done any effect. So let me explicitly allow to use this code in any follow up PR. FYI @vmaksimo @YuriPlyakhin

PR #3532, opened 2026-01-26
[CI] Temporarily work around missing llvm 22 packages

svenvh:
apt.llvm.org does not provide llvm-22 packages yet.  While we wait for them to become available, attempt some workarounds to get a bit more testing coverage on the `llvm_release_220` branch.

This commit should be reverted once llvm-22 packages are available.

### DIFF
--- a/.github/workflows/check-in-tree-build.yml
+++ b/.github/workflows/check-in-tree-build.yml
@@ -69,15 +69,15 @@ jobs:
         run: |
           curl -L "https://apt.llvm.org/llvm-snapshot.gpg.key" | sudo apt-key add -
           curl -L "https://packages.lunarg.com/lunarg-signing-key-pub.asc" | sudo apt-key add -
-          #echo "deb https://apt.llvm.org/jammy/ llvm-toolchain-jammy-22 main" | sudo tee -a /etc/apt/sources.list
+          echo "deb https://apt.llvm.org/jammy/ llvm-toolchain-jammy main" | sudo tee -a /etc/apt/sources.list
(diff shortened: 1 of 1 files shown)


.github/workflows/check-in-tree-build.yml
```diff
@@ -69,15 +69,15 @@ jobs:
         run: |
           curl -L "https://apt.llvm.org/llvm-snapshot.gpg.key" | sudo apt-key add -
           curl -L "https://packages.lunarg.com/lunarg-signing-key-pub.asc" | sudo apt-key add -
-          echo "deb https://apt.llvm.org/jammy/ llvm-toolchain-jammy-22 main" | sudo tee -a /etc/apt/sources.list
+          #echo "deb https://apt.llvm.org/jammy/ llvm-toolchain-jammy-22 main" | sudo tee -a /etc/apt/sources.list
```
vmaksimo:
```suggestion
          echo "deb https://apt.llvm.org/jammy/ llvm-toolchain-jammy main" | sudo tee -a /etc/apt/sources.list
```
Just wondering if this can be an option - why can't we get the tools (clang) the same way you're suggesting for code-style workflow? 
I mean, both would fail *if* main package would be updated - which I'm not sure will happen earlier than they'd create 22-release package version.

svenvh: It would require updating LLVM_VERSION in this job, but it turns out the only place that uses `LLVM_VERSION` is here. :)  So yeah, that might work too and should give us a more recent clang; let me try a new patch.

svenvh: > It would require updating LLVM_VERSION in this job

Actually we don't even need to do that.

PR #3538, opened 2026-01-27
Widen integers if arbitrary precision integers are not supported

frengels:
This is a conversion that's also happening in the LLVM SPIRV backend so it would make sense to add support for this.


frengels:
Thanks for the feedback, I'll have a look once I get some time to check the SPIR-V backend. I'll close this PR for now pending investigation.

PR #3558, opened 2026-01-30
Update OCL nan builtin translation to work with bfloat type

vmaksimo:
To avoid name clash between builtins with different return types, (`half` and `bfloat` specifically), introduce return type postfix for `nan` builtin in reverse translation to SPIR-V Friendly IR.

### DIFF
--- a/test/transcoding/OpenCL/nan_bfloat.ll
+++ b/test/transcoding/OpenCL/nan_bfloat.ll
@@ -17,7 +17,7 @@ target triple = "spir64"
 ; CHECK-SPIRV: ExtInst [[#HALF]] [[#]] [[#]] nan
 
 ; CHECK-SPV-IR: call spir_func bfloat @_Z22__spirv_ocl_nan_RDF16bt(
-; CHECK-SPV-IR: call spir_func half @_Z21__spirv_ocl_nan_Rhalft(
+; CHECK-SPV-IR: call spir_func half @_Z15__spirv_ocl_nant(
 
 define dso_local spir_kernel void @test_bfloat(ptr addrspace(1) align 2 %a, i16 %b) {
 entry:
--- a/lib/SPIRV/SPIRVUtil.cpp
+++ b/lib/SPIRV/SPIRVUtil.cpp
@@ -2660,7 +2660,7 @@ class OpenCLStdToSPIRVFriendlyIRMangleInfo : public BuiltinFuncMangleInfo {
 public:
   OpenCLStdToSPIRVFriendlyIRMangleInfo(OCLExtOpKind ExtOpId,
                                        ArrayRef<Type *> ArgTys, Type *RetTy)
-      : ExtOpId(ExtOpId), ArgTys(ArgTys) {
+      : ExtOpId(ExtOpId), ArgTys(ArgTys), RetTy(RetTy) {
 
     std::string Postfix = "";
     if (needRetTypePostfix())
@@ -2675,8 +2675,12 @@ public:
     case OpenCLLIB::Vload_halfn:
     case OpenCLLIB::Vloada_halfn:
     case OpenCLLIB::Vloadn:
-    case OpenCLLIB::Nan:
       return true;
+    case OpenCLLIB::Nan:
+      // Only add return type mangling for bfloat16 to disambiguate from half
+      // (both are represented as i16 in LLVM). Float and half use traditional
+      // naming for backward compatibility.
+      return RetTy->getScalarType()->isBFloatTy();
     default:
       return false;
     }
@@ -2742,6 +2746,7 @@ public:
 private:
   OCLExtOpKind ExtOpId;
   ArrayRef<Type *> ArgTys;
+  Type *RetTy;
 };
 } // namespace
 
--- a/test/transcoding/OpenCL/nan.ll
+++ b/test/transcoding/OpenCL/nan.ll
@@ -16,7 +16,7 @@ target triple = "spir64"
 
 ; CHECK-LLVM: call spir_func float @_Z3nanj(
 
-; CHECK-SPV-IR: call spir_func float @_Z22__spirv_ocl_nan_Rfloatj(
+; CHECK-SPV-IR: call spir_func float @_Z15__spirv_ocl_nanj(
 
 define dso_local spir_kernel void @test(ptr addrspace(1) align 4 %a, i32 %b) {
 entry:


test/transcoding/OpenCL/nan_bfloat.ll
```diff
@@ -0,0 +1,35 @@
...
+; CHECK-SPIRV: ExtInst [[#BFLOAT]] [[#]] [[#]] nan
+; CHECK-SPIRV: ExtInst [[#HALF]] [[#]] [[#]] nan
+
+; CHECK-SPV-IR: call spir_func bfloat @_Z22__spirv_ocl_nan_RDF16bt(
+; CHECK-SPV-IR: call spir_func half @_Z21__spirv_ocl_nan_Rhalft(
+
+define dso_local spir_kernel void @test_bfloat(ptr addrspace(1) align 2 %a, i16 %b) {
+entry:
+  %call = tail call spir_func bfloat @_Z22__spirv_ocl_nan_RDF16bt(i16 %b)
+  %call2 = tail call spir_func half @_Z21__spirv_ocl_nan_Rhalft(i16 %b)
```
MrSidims: Do we already expect such output from clang?

vmaksimo: Nope, it was actually misread of the https://github.com/KhronosGroup/SPIRV-LLVM-Translator/blob/main/docs/SPIRVRepresentationInLLVM.rst#spir-v-extended-instructions-mapped-to-llvm-function-calls
Double underscore is the expected format, I'll adjust the PR and tests accordingly.

MrSidims: Let me clarify myself. Do we expect overload by return value with  be accepted by clang already or in the future (either in llvm.org or intel/llvm)? At least I can't find any tests for this: https://github.com/search?q=repo%3Aintel%2Fllvm+__spirv_ocl_nan&type=code

vmaksimo: Hm, it should be used by Triton, but now I don't see that any FE is producing such friendly call as well.

If we are talking about acceptance by backends - I'm sure it's not the first IR friendly instruction to have explicit return type in its name.

test/transcoding/OpenCL/nan_bfloat.ll
```diff
@@ -0,0 +1,35 @@
...
+target datalayout = "e-i64:64-v16:16-v24:32-v32:32-v48:64-v96:128-v192:256-v256:256-v512:512-v1024:1024-n8:16:32:64-G1"
+target triple = "spir64"
+
+; CHECK-SPIRV: TypeFloat [[#BFLOAT:]] 16 0 {{$}}
+; CHECK-SPIRV: TypeFloat [[#HALF:]] 16 {{$}}
+; CHECK-SPIRV: ExtInst [[#BFLOAT]] [[#]] [[#]] nan
+; CHECK-SPIRV: ExtInst [[#HALF]] [[#]] [[#]] nan
+
+; CHECK-SPV-IR: call spir_func bfloat @_Z22__spirv_ocl_nan_RDF16bt(
+; CHECK-SPV-IR: call spir_func half @_Z21__spirv_ocl_nan_Rhalft(
```
YuriPlyakhin: is it correct that reverse translation would produce 2 underscores: `nan__RDF16bt`
after https://github.com/KhronosGroup/SPIRV-LLVM-Translator/pull/3560 is merged?

test/transcoding/OpenCL/nan_bfloat.ll
```diff
@@ -0,0 +1,35 @@
...
+
+define dso_local spir_kernel void @test_bfloat(ptr addrspace(1) align 2 %a, i16 %b) {
+entry:
+  %call = tail call spir_func bfloat @_Z23__spirv_ocl_nan__RDF16bt(i16 %b)
+  %call2 = tail call spir_func half @_Z22__spirv_ocl_nan__Rhalft(i16 %b)
```
YuriPlyakhin, line 25: for forward translation, will translator accept both variants: `nan_RDF16bt` and `nan__RDF16bt`?

MrSidims: Yes. The translator doesn't care about mangling during forward translation (apart of initial decoding to get how many symbols to read after _Z%X%%X% prefix). You could even name it `_Z23__spirv_ocl_nan__GoGoTeamBlue` and call `_Z23__spirv_ocl_nan__GoGoTeamRed` right after - the translator doesn't care, it will produce `ExtInst # # # nan #` for both whatsoever.

test/transcoding/OpenCL/nan_bfloat.ll
MrSidims: That worries me. On the one hand - the translator doesn't care about mangling for the input LLVM IR, what is important is that the actual name matches a known builtin name and it cares about a type/types in the input IR to correctly generate an instruction (or extend instruction).

But. Consider this example: https://godbolt.org/z/vr936rq1d . You see, the problem is that the mangling is different by the argument type, there is no mangling per return type. And that is expected, unless we move to template specialization - C++ has no overloads by return type. Yet current clang's mangling is fine as per https://registry.khronos.org/OpenCL/specs/3.0-unified/html/OpenCL_C.html#math-functions `float` and `half` overloads of `nan` builin are different by the argument type. Addition of `bfloat` breaks it and that is what this PR is trying to address. Current revision is LGTM, but the original change in the [Writer](https://github.com/KhronosGroup/SPIRV-LLVM-Translator/pull/3558/changes/0ea3ee1cb7659a1badffb10a7203eb0871edb9f2#diff-ba003da77a80df1db76fb378a873a80e758591d72bf82a46fc90069b2a947990R289) has made me thinking, that we'd better to align on such builtins mangling ahead of time and clang seem to be the place for this. So I would suggest you to discuss the matter with @AaronBallman - what should be ideal mangling for builtins returning `bfloat` and how to distinguish them with builtins returning `half` in cases, where argument types match.

test/transcoding/OpenCL/nan_bfloat.ll
vmaksimo: if merged - yes, but i decided to reject the mentioned PR in favor of changing the specification.
so no, this check would not ne changed.

YuriPlyakhin (approved):
LGTM.
Q: do we need to add spirv-val to the test?

vmaksimo:
> LGTM. Q: do we need to add spirv-val to the test?

@YuriPlyakhin my suggestion is no. Here `spirv-val` fails with:
```
error: line 29: ExtInst doesn't support BFloat16 type.
  %call = OpExtInst %bfloat16 %1 nan %b
```

PR #3562, opened 2026-02-04
Add support for debug macro information (DW_MACINFO_define/undef)

mgcarrasco:
Implement bidirectional translation of LLVM [DIMacro](https://llvm.org/doxygen/classllvm_1_1DIMacro.html) to SPIR-V [DebugMacroDef ](https://registry.khronos.org/SPIR-V/specs/unified1/DebugInfo.html#DebugMacroDef)and [DebugMacroUndef](https://registry.khronos.org/SPIR-V/specs/unified1/DebugInfo.html#DebugMacroUndef) instructions.

LLVM to SPIR-V:
- Add transDbgMacros() to process all macro definitions and undefs
- Add transDbgMacroDefine() to translate DW_MACINFO_define to DebugMacroDef
- Add transDbgMacroUndef() to translate DW_MACINFO_undef to DebugMacroUndef
- DW_MACINFO_define are processed first, so they are available for DW_MACINFO_undef cases.

SPIR-V to LLVM:
- Add transMacroDef() to translate DebugMacroDef back to DIMacro
- Add transMacroUndef() to translate DebugMacroUndef back to DIMacro
- Add getOrCreateMacroFile() to manage DIMacroFile creation
- Handle edge case where LLVM IR allows undefining macros that were never defined (DebugMacroUndef with DebugInfoNone reference)

### DIFF
--- a/lib/SPIRV/LLVMToSPIRVDbgTran.cpp
+++ b/lib/SPIRV/LLVMToSPIRVDbgTran.cpp
@@ -66,6 +66,8 @@ void LLVMToSPIRVDbgTran::transDebugMetadata() {
       transDbgEntry(IE);
   }
 
+  transDbgMacros();
+
   for (const DIType *T : DIF.types())
     transDbgEntry(T);
 
@@ -1680,4 +1682,91 @@ SPIRVEntry *LLVMToSPIRVDbgTran::transDbgModule(const DIModule *Module) {
   BM->addExtension(ExtensionID::SPV_INTEL_debug_module);
   BM->addCapability(spv::CapabilityDebugInfoModuleINTEL);
   return BM->addDebugInfo(SPIRVDebug::ModuleINTEL, getVoidTy(), Ops);
+}
+
+void LLVMToSPIRVDbgTran::transDbgMacros() {
+  auto ProcessMacros = [&](unsigned MacroType, auto &&Handler) {
+    for (const auto &[Macro, MacroFile] : DIF.macros()) {
+      if (Macro->getMacinfoType() != MacroType)
+        continue;
+
+      SPIRVString *FileName = BM->getString("");
+      if (MacroFile && MacroFile->getFile())
+        FileName = BM->getString(getFullPath(MacroFile->getFile()));
+
+      Handler(Macro, FileName);
+    }
+  };
+
+  // Process DW_MACINFO_define macros and populate MacroDefMap
+  ProcessMacros(dwarf::DW_MACINFO_define,
+                [&](const DIMacro *Macro, SPIRVString *FileName) {
+                  transDbgMacroDefine(Macro, FileName);
+                });
+
+  // Process DW_MACINFO_undef macros (after DW_MACINFO_define populates
+  // MacroDefMap). DebugMacroUndef in SPIR-V requires a reference to the
+  // DebugMacroDef (not just the macro name as in LLVM IR)
+  ProcessMacros(dwarf::DW_MACINFO_undef,
+                [&](const DIMacro *Macro, SPIRVString *FileName) {
+                  transDbgMacroUndef(Macro, FileName);
+                });
+}
+
+SPIRVEntry *LLVMToSPIRVDbgTran::transDbgMacroDefine(const DIMacro *Macro,
+                                                    SPIRVString *FileName) {
+  SPIRVWordVec Ops;
+  using namespace SPIRVDebug::Operand::MacroDef;
+  Ops.resize(MinOperandCount);
+  Ops[SourceIdx] = FileName->getId();
+  Ops[LineIdx] = Macro->getLine();
+
+  if (isNonSemanticDebugInfo()) {
+    transformToConstant(Ops, {LineIdx});
+  }
+
+  Ops[NameIdx] = BM->getString(Macro->getName().str())->getId();
+
+  // ValueIdx is optional, add it if the value operand exists.
+  // Once the MacroDef is parsed, it is not possible to distinguish between a
+  // macro with an empty value and a macro without a value. The implementation
+  // encodes the empty value case as one without macro value.
+  if (!Macro->getValue().empty()) {
+    StringRef Value = Macro->getValue();
+    Ops.push_back(BM->getString(Value.str())->getId());
+  }
+
+  SPIRVEntry *MacroEntry =
+      BM->addDebugInfo(SPIRVDebug::MacroDef, getVoidTy(), Ops);
+  MDMap[Macro] = MacroEntry;
+  MacroDefMap[Macro->getName().str()] = static_cast<SPIRVExtInst *>(MacroEntry);
+  return MacroEntry;
+}
+
+SPIRVEntry *LLVMToSPIRVDbgTran::transDbgMacroUndef(const DIMacro *Macro,
+                                                   SPIRVString *FileName) {
+  SPIRVWordVec Ops;
+  using namespace SPIRVDebug::Operand::MacroUndef;
+  Ops.resize(OperandCount);
+  Ops[SourceIdx] = FileName->getId();
+  Ops[LineIdx] = Macro->getLine();
+
+  if (isNonSemanticDebugInfo()) {
+    transformToConstant(Ops, {LineIdx});
+  }
+
+  SPIRVId MacroDefId = getDebugInfoNoneId();
+
+  // transDbgMacroDefine is processed first so MacroDefMap is already populated
+  // at this point.
+  auto It = MacroDefMap.find(Macro->getName().str());
+  if (It != MacroDefMap.end()) {
+    MacroDefId = It->second->getId();
+  }
+
+  Ops[MacroIdx] = MacroDefId;
+  SPIRVEntry *MacroEntry =
+      BM->addDebugInfo(SPIRVDebug::MacroUndef, getVoidTy(), Ops);
+  MDMap[Macro] = MacroEntry;
+  return MacroEntry;
 }
--- a/lib/SPIRV/SPIRVToLLVMDbgTran.cpp
+++ b/lib/SPIRV/SPIRVToLLVMDbgTran.cpp
@@ -1407,6 +1407,76 @@ DINode *SPIRVToLLVMDbgTran::transModule(const SPIRVExtInst *DebugInst) {
       Scope, Name, ConfigMacros, IncludePath, ApiNotes, File, Line, IsDecl);
 }
 
+DIMacroFile *
+SPIRVToLLVMDbgTran::getOrCreateMacroFile(DIFile *File,
+                                         const SPIRVExtInst *DebugInst) {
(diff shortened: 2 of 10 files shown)


test/DebugInfo/DebugMacroDef.ll
```diff
@@ -0,0 +1,55 @@
+; Test round-trip translation of debug macro information:
+; LLVM IR -> SPIR-V -> LLVM IR
+
+; RUN: llvm-spirv --spirv-debug-info-version=nonsemantic-shader-100 %s -o %t.spv
```
MrSidims: > Implement bidirectional translation of LLVM [DIMacro](https://llvm.org/doxygen/classllvm_1_1DIMacro.html) to SPIR-V [DebugMacroDef ](https://registry.khronos.org/SPIR-V/specs/unified1/DebugInfo.html#DebugMacroDef)and [DebugMacroUndef](https://registry.khronos.org/SPIR-V/specs/unified1/DebugInfo.html#DebugMacroUndef) instructions.

Note, that the spec links in the PR description are leading to OpenCL.DebugInfo.100 instruction set, while:
a. tests are covering nonsemantic debug info; 
b. code is also build for nonsemantic, see few Literal parameters of the instructions in the referenced OpenCL spec, while in the implementation (and [nonsemantic variant](https://github.com/KhronosGroup/SPIRV-Registry/blob/main/nonsemantic/NonSemantic.Shader.DebugInfo.100.asciidoc#DebugMacroDef)) there operands are Ids.

I'd suggest to test and implement all 3 debug information types: OpenCL, nonsemantic.100 and nonsemantic.200. The later 2 are identical for Macro handling. OpenCL is a bit different, but not that drastically: see the following lines here and there which transforms literals to constants:
```
    if (isNonSemanticDebugInfo())
      transformToConstant(Ops, {ScopeLineIdx});
```

mgcarrasco: thanks, will do.

lib/SPIRV/LLVMToSPIRVDbgTran.cpp
```diff
@@ -1681,3 +1683,75 @@ SPIRVEntry *LLVMToSPIRVDbgTran::transDbgModule(const DIModule *Module) {
...
+}
+
+SPIRVEntry *LLVMToSPIRVDbgTran::transDbgMacroDefine(const DIMacro *Macro,
+                                                    SPIRVString *FileName) {
+  SPIRVWordVec Ops;
+  using namespace SPIRVDebug::Operand::MacroDef;
+  Ops.resize(OperandCount);
+  Ops[SourceIdx] = FileName->getId();
+  Ops[LineIdx] =
+      SPIRVWriter->transValue(getUInt(M, Macro->getLine()), nullptr)->getId();
```
maarquitos14: If I'm looking correctly at the spec, `Line` is a literal number, not an `<id>`.

lib/SPIRV/LLVMToSPIRVDbgTran.cpp
```diff
@@ -1681,3 +1683,75 @@ SPIRVEntry *LLVMToSPIRVDbgTran::transDbgModule(const DIModule *Module) {
...
+  return MacroEntry;
+}
+
+SPIRVEntry *LLVMToSPIRVDbgTran::transDbgMacroUndef(const DIMacro *Macro,
+                                                   SPIRVString *FileName) {
+  SPIRVWordVec Ops;
+  using namespace SPIRVDebug::Operand::MacroUndef;
+  Ops.resize(OperandCount);
+  Ops[SourceIdx] = FileName->getId();
+  Ops[LineIdx] =
```
maarquitos14: Same than above.

lib/SPIRV/LLVMToSPIRVDbgTran.cpp
```diff
@@ -1681,3 +1683,75 @@ SPIRVEntry *LLVMToSPIRVDbgTran::transDbgModule(const DIModule *Module) {
...
+      SPIRVWriter->transValue(getUInt(M, Macro->getLine()), nullptr)->getId();
+
+  SPIRVId MacroDefId = getDebugInfoNoneId();
+
+  // transDbgMacroDefine is processed first so MacroDefMap is already populated
+  // at this point
```
maarquitos14:
```suggestion
  // at this point.
```

lib/SPIRV/LLVMToSPIRVDbgTran.cpp
mgcarrasco: Thanks! This should be fixed now.

For `isNonSemanticDebugInfo()` we must follow https://github.com/KhronosGroup/SPIRV-Registry/blob/main/nonsemantic/NonSemantic.Shader.DebugInfo.100.asciidoc#DebugMacroDef where line is an `<id>` otherwise the literal number (https://registry.khronos.org/SPIR-V/specs/unified1/OpenCL.DebugInfo.100.html 4.9 Section).

lib/SPIRV/LLVMToSPIRVDbgTran.cpp
mgcarrasco: Thanks, fixed too.

test/DebugInfo/DebugMacroDef.ll
mgcarrasco: Thanks! This was very helpful. This should be fixed now.

mgcarrasco: We are testing all debug information types.

lib/SPIRV/LLVMToSPIRVDbgTran.cpp
maarquitos14: Okay, then I think the follow-up is adding SPIRV checks in the tests to check that each debug info type uses the expected format for `Line`.

mgcarrasco: Thanks, great idea. We are now testing the SPIRV instructions.

maarquitos14 (approved):

lib/SPIRV/libSPIRV/SPIRV.debug.h
```diff
@@ -971,6 +971,25 @@ enum {
 } // namespace NonSemantic
 } // namespace ImportedEntity
 
+namespace MacroDef {
+enum {
+  SourceIdx      = 0,
+  LineIdx        = 1,
+  NameIdx        = 2,
+  ValueIdx       = 3,
```
vmaksimo: `Value` id is optional argument for `DebugMacroDef`. Can we test the case where define is made without the value?

lib/SPIRV/LLVMToSPIRVDbgTran.cpp
```diff
@@ -1681,3 +1683,82 @@ SPIRVEntry *LLVMToSPIRVDbgTran::transDbgModule(const DIModule *Module) {
...
+  Ops.resize(OperandCount);
+  Ops[SourceIdx] = FileName->getId();
+  Ops[LineIdx] = Macro->getLine();
+
+  if (isNonSemanticDebugInfo()){
+    transformToConstant(Ops, {LineIdx});
+  }
+
+  Ops[NameIdx] = BM->getString(Macro->getName().str())->getId();
+  Ops[ValueIdx] = BM->getString(Macro->getValue().str())->getId();
```
vmaksimo: `Value` id is optional argument for `DebugMacroDef`. Can we test the case where define is made without the value?
+also need to handle this in reverse translation.

lib/SPIRV/SPIRVToLLVMDbgTran.cpp
```diff
@@ -1407,6 +1407,76 @@ DINode *SPIRVToLLVMDbgTran::transModule(const SPIRVExtInst *DebugInst) {
...
+DIMacro *SPIRVToLLVMDbgTran::transMacroDef(const SPIRVExtInst *DebugInst) {
+  using namespace SPIRVDebug::Operand::MacroDef;
+  const SPIRVWordVec &Ops = DebugInst->getArguments();
+  assert(Ops.size() >= OperandCount && "Invalid number of operands");
+
+  const std::string &SourceFileName = getString(Ops[SourceIdx]);
+  DIFile *File = getDIFile(SourceFileName);
+
+  DIMacroFile *MacroFile =
+      !File ? nullptr : getOrCreateMacroFile(File, DebugInst);
```
vmaksimo, line 1434: Just wondering if it makes sense to create "useless" debug info if there's no file to reference?

lib/SPIRV/SPIRVToLLVMDbgTran.cpp
```diff
@@ -1407,6 +1407,76 @@ DINode *SPIRVToLLVMDbgTran::transModule(const SPIRVExtInst *DebugInst) {
...
+  // DW_MACINFO_def (i.e., undefining a macro that was never defined).
+  // Example: !DIMacro(type: DW_MACINFO_undef, line: 1, name: "SIZE")
+  //          with no corresponding DW_MACINFO_def for "SIZE"
+  //
+  // When translated to SPIR-V, this becomes:
+  //   DebugMacroUndef <source> <line> DebugInfoNone
+  //
+  // This results in ReferencedMacro being nullptr. We must use a non-empty
+  // fallback name because DIBuilder::createMacro() in LLVM rejects empty
+  // strings.
```
vmaksimo: Is not this the case where we should drop the debug info? I know the translator can be not the perfect place to deal with the validity of debug info, and such instruction will likely be removed further, but for me it just doesn't make sense to preserve unusable info.. I might missing something here

lib/SPIRV/SPIRVToLLVMDbgTran.cpp
mgcarrasco: I think in this particular case it would be incomplete but not useless, just the file would be missing but the other fields will be present.

lib/SPIRV/SPIRVToLLVMDbgTran.cpp
mgcarrasco: Thanks. I see your point. The current implementation is not producing meaningful information so it may be better not to emit debug info in this case altogether.

If needed, we could try to handle this more gracefully but in a future PR. I have simplified the code so we don't emit info for this case.

lib/SPIRV/LLVMToSPIRVDbgTran.cpp
mgcarrasco: Thanks! Now the code is taking into account that it could be optional.

mgcarrasco:
> The code LGTM, thanks! But it would be nice to have SPIR-V test checks in all the tests, not only in `DebugMacroDef.ll`, sorry for not mentioning this earlier.

@vmaksimo @maarquitos14 thanks for the reviews. All issues have been addressed already, including the SPIRV checks for all tests.

MrSidims (approved):

PR #3564, opened 2026-02-05
[NFC] Add testing for cross-compilation with SPIR-V backend

MrSidims:
No extensions/ nor DebugInfo/ tests are touched.
For tests that are failing appropriate TODO and FIXME are added. Variants of errors:
1. llc crash run on the test (5 tests), can be found by grep LLC_FAIL;
2. llvm-spirv -r crash when reverse translating llc-generated SPIR-V (5 tests),
can be found by grep LLVM_SPIRV_R_FAIL;
3. FileCheck missmatches that are not yet investigated as a part of this patch (a number of tests),can be found by grep FILECHECK_FAIL.

Note, few IR samples in llvm-spirv LITs are producing empty SPIR-V modules when compiled with SPIR-V backend due to DCE. This has to be fixed.


test/transcoding/fadd.ll
```diff
@@ -4,9 +4,9 @@
 ; RUN: llvm-spirv %t.bc -o %t.spv
 ; RUN: llvm-spirv %t.bc --spirv-ext=+SPV_KHR_float_controls2 -o %t.fc2.spv
 ; RUN: spirv-val %t.spv
-; RUN: spirv-val %t.fc2.spv
```
YuriPlyakhin, line 7: this test doesn't use llc, right? So why it was modified? I guess what I don't understand is when exactly SPIR-V backend is involved in this test scenario.
Similar question is to all other cases like that.

MrSidims: Wrong conflict resolution. Fixing

maarquitos14 (approved):

PR #3571, opened 2026-02-10
Add spirv-use-user-semantic-for-linkage option

MrSidims:
when enabled, LLVM linkage types that have no SPIR-V counterparts are translated as UserSemantic "linkage:%type%" decoration, where %type% can be weak, weak_odr etc.

Ideal solution is to have new extension(s) for this, but we don't have them. Using UserSemantic is not correct, but it doesn't worsen the current situation, which is:
if in 2 modules there is a weak function, and these modules are linked together after llvm-spirv roundtrip, then llvm-link errors out because functions from both modules become extern.

So with this patch, if linkage type is preserved via UserSemantic we won't see this error, and if UserSemantic is ignored (which SPIR-V consumer has right to do) - the linkage error remains.

### DIFF
--- a/lib/SPIRV/SPIRVWriter.cpp
+++ b/lib/SPIRV/SPIRVWriter.cpp
@@ -7269,8 +7269,8 @@ void LLVMToSPIRVBase::transLinkageAsUserSemantic(SPIRVEntry *BE,
     return;
   // Emit a UserSemantic decoration to preserve LLVM linkage types that have
   // no native SPIR-V representation. The string format is "linkage:<type>".
-  // Linkage types that map natively (External/Internal/Private/LinkOnceODR)
-  // are not emitted.
+  // Linkage types that map natively (Export/Import/LinkOnceODR) are not
+  // emitted.
   const char *LinkageName = nullptr;
   switch (GV->getLinkage()) {
   case GlobalValue::WeakAnyLinkage:
@@ -7281,9 +7281,6 @@ void LLVMToSPIRVBase::transLinkageAsUserSemantic(SPIRVEntry *BE,
     break;
   case GlobalValue::LinkOnceAnyLinkage:
     LinkageName = "linkonce";
-    break;
-  case GlobalValue::AvailableExternallyLinkage:
-    LinkageName = "available_externally";
     break;
   case GlobalValue::CommonLinkage:
     LinkageName = "common";


lib/SPIRV/SPIRVReader.cpp
```diff
@@ -1793,6 +1811,17 @@ Value *SPIRVToLLVM::transValueWithoutDecoration(SPIRVValue *BV, Function *F,
                              : GlobalValue::UnnamedAddr::None);
     LVar->setInitializer(Initializer);
 
+    // Override linkage from UserSemantic decoration if the flag is enabled.
+    if (BM->shouldConsumeLinkageUserSemantic()) {
+      for (const auto &UsSem :
```
maarquitos14, line 1817:
```suggestion
      for (const auto &UserSemantic :
```

Just a nit, but `UsSem` is not very descriptive. Feel free to ignore.

lib/SPIRV/SPIRVReader.cpp
```diff
@@ -4358,6 +4387,14 @@ void SPIRVToLLVM::transIntelFPGADecorations(SPIRVValue *BV, Value *V) {
     std::vector<SmallString<256>> AnnotStrVec;
     generateIntelFPGAAnnotation(BV, AnnotStrVec);
 
+    // Filter out "linkage:..." annotations that were consumed for linkage
+    // override, so they don't become spurious global annotations.
+    if (BM->shouldConsumeLinkageUserSemantic()) {
```
maarquitos14: Could they be erased just after being consumed so we don't need to search them again?

test/weak-linkage-user-semantic.ll
```diff
@@ -0,0 +1,71 @@
+; Test that LLVM linkage types not natively supported in SPIR-V are preserved
+; through UserSemantic decorations when the experimental flags are enabled.
+
+; RUN: llvm-spirv --spirv-emit-linkage-user-semantic %s -o %t.spv
+; RUN: spirv-val %t.spv
+; RUN: llvm-spirv %t.spv -to-text -o %t.spt
+; RUN: FileCheck < %t.spt %s --check-prefix=CHECK-SPIRV
+
+; RUN: llvm-spirv %s -o %t.noflg.spv
```
maarquitos14, line 9: No need for `spirv-val` here because, from `spirv-val` POV, it's just the same thing, right?

lib/SPIRV/SPIRVReader.cpp
```diff
@@ -97,6 +98,23 @@ using namespace llvm;
 using namespace SPIRV;
 using namespace OCLUtil;
 
+// Parse a "linkage:<type>" UserSemantic string into an LLVM linkage type.
+static std::optional<GlobalValue::LinkageTypes>
+parseLinkageUserSemantic(StringRef UsSem) {
```
YuriPlyakhin, line 103: nit
```suggestion
parseLinkageUserSemantic(StringRef UserSemantic) {
```

tools/llvm-spirv/llvm-spirv.cpp
```diff
@@ -169,6 +169,18 @@ static cl::opt<bool> SPIRVEmitFunctionPtrAddrSpace(
...
+    cl::desc("Emit UserSemantic decoration with \"linkage:<type>\" string on "
+             "functions and globals whose linkage has no native SPIR-V "
+             "representation."));
+
+static cl::opt<bool> SPIRVConsumeLinkageUserSemantic(
+    "spirv-consume-linkage-user-semantic", cl::init(false),
+    cl::desc("Interpret UserSemantic decoration with \"linkage:<type>\" "
+             "string as the corresponding LLVM linkage on functions and "
+             "globals."));
+
```
YuriPlyakhin: Can we use the same option for forward and reverse translation to avoid errors, when forward-specific option is used for reverse case and vice versa?
```suggestion
static cl::opt<bool> SPIRVUseUserSemanticForLinkage(
    "spirv-use-user-semantic-for-linkage", cl::init(false),
    cl::desc("Emit/consume UserSemantic decoration with \"linkage:<type>\" string on "
             "functions and globals whose linkage has no native SPIR-V "
             "representation."));
```

test/weak-linkage-user-semantic.ll
```diff
@@ -0,0 +1,71 @@
+; Test that LLVM linkage types not natively supported in SPIR-V are preserved
+; through UserSemantic decorations when the experimental flags are enabled.
```
YuriPlyakhin, line 2: flag description in llvm-spirv.cpp doesn't mention that this flag is experimental. If this flag is indeed experimental, we should be more explicit about it. Also, if it is indeed experimental, what are the next steps/conditions to decide, if to change it/remove or otherwise make it non-experimental? It would be good to have it documented in the comments in the code somewhere, maybe in llvm-spirv.cpp near the flag description.

test/weak-linkage-user-semantic.ll
```diff
@@ -0,0 +1,71 @@
...
+; RUN: FileCheck < %t.noflg.spt %s --check-prefix=CHECK-SPIRV-NOFLG
+
+; RUN: llvm-spirv -r --spirv-consume-linkage-user-semantic %t.spv -o %t.rev.bc
+; RUN: llvm-dis %t.rev.bc -o %t.rev.ll
+; RUN: FileCheck < %t.rev.ll %s --check-prefix=CHECK-LLVM
+
+; RUN: llvm-spirv -r %t.spv -o %t.rev.noflg.bc
+; RUN: llvm-dis %t.rev.noflg.bc -o %t.rev.noflg.ll
+; RUN: FileCheck < %t.rev.noflg.ll %s --check-prefix=CHECK-LLVM-NOFLG
+
```
YuriPlyakhin, line 21: should we add a test, where we generate spirv without flag, and consume with?

lib/SPIRV/SPIRVReader.cpp
MrSidims: Well, there are other places in code, where the style is set to call `UserSemantic` variable as `UsSem` and not UserSemantic (probably because this code was written by me in the past XD (Oh, actually, no only by me, I see:
`9c78e6e04 (Alexey Sachkov        2022-11-23 11:52:25 -0500 4469)   for (const auto &UsSem :`)).

I don't have strong opinion and can rename it all over the places, but one think to consider: UserSemantic is decoration aka instance of OpDecorate. UseSem - is a string that this decoration holds.

test/weak-linkage-user-semantic.ll
MrSidims: Good point, I'll remove this confusing comment from the test or adjust llvm-spirv help message.

Some internal knowledge for better context first:
1. Per SPIR-V spec: UserSemantic can not carry semantics;
2. Per SPIR-V spec: UserSemantic can not be applied to a function (yet spir-val is not checking it);
3. Despite 1 and 2, sometime ago, initially in llvm-spirv we decided to allow passing of function attributes and additional info via UserSemantic as a part of prototyping a feature (aka without introducing a specification).

Next steps for productizing translation of weak and other linkage types is to add a new SPIR-V extension, like it was done before with linkonce_odr. The problem is, that I know, that Khronos WG previously had troubles approving C/C++/LLVM linkage types addition to SPIR-V (see gitlab issue 39), and I can't guarantee that there will be support within SPIR-V community for this extension, but I'll give it a chance. But there is a probability, that this "experimental" option would live longer that we would want to.

test/weak-linkage-user-semantic.ll
MrSidims: ~Will add~
added

test/weak-linkage-user-semantic.ll
MrSidims: ~Will add~
added

tools/llvm-spirv/llvm-spirv.cpp
MrSidims: Sure, will change this

lib/SPIRV/SPIRVReader.cpp
MrSidims: https://github.com/KhronosGroup/SPIRV-LLVM-Translator/pull/3571/changes#r2793465820

lib/SPIRV/SPIRVReader.cpp
YuriPlyakhin: both Marcos and I provided this as a nit, so there is no strong opinion from our side either.
Since the rest of the code is using `UsSem`, I guess it is fine to keep it and maybe renaming should be a separate PR.
Both `us` and `use` can be considered as normal words, that's the reason, why I personally don't like shortening the word `user` to `us` or `use`.

test/weak-linkage-user-semantic.ll
YuriPlyakhin: ok, that sounds like flag is indeed experimental, so llvm-spirv help message should be adjusted and it might be good to add the context you provided to the source code comment in some form, so it is not lost in PRs.

lib/SPIRV/SPIRVReader.cpp
maarquitos14: Like I said, feel free to ignore.

lib/SPIRV/SPIRVReader.cpp
MrSidims: Applied

lib/SPIRV/SPIRVReader.cpp
MrSidims: Well, I just wanted to object referencing the "existing" code style, and then realized, that it was created with high influence of myself and kind of found it funny and decided to share :)

test/weak-linkage-user-semantic.ll
MrSidims: Added a note about experimental-ness of the option.

lib/SPIRV/SPIRVReader.cpp
```diff
@@ -97,6 +98,23 @@ using namespace llvm;
...
+  if (!UsSem.starts_with("linkage:"))
+    return std::nullopt;
+  StringRef LinkageName = UsSem.drop_front(strlen("linkage:"));
+  return StringSwitch<std::optional<GlobalValue::LinkageTypes>>(LinkageName)
+      .Case("weak", GlobalValue::WeakAnyLinkage)
+      .Case("weak_odr", GlobalValue::WeakODRLinkage)
+      .Case("linkonce", GlobalValue::LinkOnceAnyLinkage)
+      .Case("available_externally", GlobalValue::AvailableExternallyLinkage)
```
MrSidims, line 111: actually, this should be just Import

lib/SPIRV/SPIRVReader.cpp
```diff
@@ -1793,6 +1811,19 @@ Value *SPIRVToLLVM::transValueWithoutDecoration(SPIRVValue *BV, Function *F,
...
+    // Override linkage from UserSemantic decoration if the flag is enabled.
+    // Erase consumed decorations so they don't appear as spurious annotations.
+    if (BM->shouldUseUserSemanticForLinkage()) {
+      for (const auto &UsSem :
+           BVar->getDecorationStringLiteral(DecorationUserSemantic)) {
+        if (auto LT = parseLinkageUserSemantic(UsSem)) {
+          LVar->setLinkage(*LT);
+          break;
+        }
+      }
```
vmaksimo, line 1823:
```suggestion
      StringRef UsSem = getString(BVar->getDecorationLiterals(DecorationUserSemantic));
      if (auto LT = parseLinkageUserSemantic(UsSem))
        LVar->setLinkage(*LT);
```

YuriPlyakhin (approved):

MrSidims:
Converted to draft to avoid incidental merge and back to the whiteboard.

Observations and ideas:
1. Currently if llvm-spirv is built on top of llvm containing https://github.com/llvm/llvm-project/pull/181661 - compilation of OpenCL samples error out saying, that spirv-use-user-semantic-for-linkage is defined twice;
2. I'm considering to move away from NonSemantic decoration to NonSemantic.AuxData instruction set, which is implemented in the translator, though not implemented in the backend.

PR #3581, opened 2026-02-16
[CMake] Move headers to include/LLVMSPIRVLib to match install layout

wenju-he:
Currently, headers are only available in their source locations during
build time, but are installed to a flattened 'LLVMSPIRVLib' directory.
This creates a discrepancy between in-tree builds (#include
"LLVMSPIRVOpts.h") and out-of-tree builds (#include
"LLVMSPIRVLib/LLVMSPIRVOpts.h") for downstream projects.

This PR restructures the headers by moving them directly to
include/LLVMSPIRVLib/ in the source tree, allowing downstream
projects (e.g., https://github.com/intel/opencl-clang) to use
#include <LLVMSPIRVLib/header.h> consistently in all scenarios.

Organizing headers in subdirectories matching the installed layout is
a common pattern across LLVM projects (e.g., Clang lib/Headers, LLDB, libcxx).


svenvh:
Retriggering CI.

PR #3610, opened 2026-02-26
Handle struct return in FP conversion builtins

vmaksimo:
__builtin conversion functions can use `sret` instead of direct returns (e.g. wrapper type `sycl::half`). When sret is present, operand 0 is the return pointer, not the source value.


vmaksimo:
Makes total sense - we found the way to avoid usage of user-defined types (here - sycl struct wrappers over native types), so this change is actually not needed.
Closing this one.

@MrSidims if you still want to proceed with your idea about ABI change in clang - guess it worth raising in LLVM forum.

PR #3661, opened 2026-03-17
Drop undef/poison initializers for global variables in SPIR-V translation

aobolensk:
Align the translator with the LLVM SPIR-V backend behavior : undef/poison initializers are dropped unless the variable is constant with an aggregate type. This avoids emitting unnecessary OpUndef initializers in the generated SPIR-V.

### DIFF
--- a/lib/SPIRV/SPIRVWriter.cpp
+++ b/lib/SPIRV/SPIRVWriter.cpp
@@ -2073,12 +2073,20 @@ LLVMToSPIRVBase::transValueWithoutDecoration(Value *V, SPIRVBasicBlock *BB,
 
   if (auto *GV = dyn_cast<GlobalVariable>(V)) {
     llvm::Type *Ty = Scavenger->getScavengedType(GV);
-    // Though variables with common linkage type are initialized by 0,
-    // they can be represented in SPIR-V as uninitialized variables with
-    // 'Export' linkage type, just as tentative definitions look in C
-    llvm::Constant *Init = GV->hasInitializer() && !GV->hasCommonLinkage()
-                               ? GV->getInitializer()
-                               : nullptr;
+    // Check whether the global has an initializer for SPIR-V.
+    // Variables with common linkage are represented as uninitialized with
+    // 'Export' linkage, just as tentative definitions look in C.
+    // Undef/poison initializers are dropped unless the variable is constant
+    // with an aggregate type
+    auto HasInit = [](const GlobalVariable *GV) {
+      if (!GV->hasInitializer() || GV->hasCommonLinkage())
+        return false;
+      if (isa<UndefValue>(GV->getInitializer()))
+        return GV->isConstant() &&
+               GV->getInitializer()->getType()->isAggregateType();
+      return true;
+    };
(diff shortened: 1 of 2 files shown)


vmaksimo (commented):
I seriously doubt we should make this change before LLVM totally removes `undef`.
We already aligned to the community wherever possible (cec12d6) in terms of using `poison`, but translation of `OpUndef` to `poison` does not seem reasonable to me right now.

aobolensk:
> I seriously doubt we should make this change before LLVM totally removes `undef`. We already aligned to the community wherever possible ([cec12d6](https://github.com/KhronosGroup/SPIRV-LLVM-Translator/commit/cec12d6cf46306d0a015e883d5adb5a8200df1c0)) in terms of using `poison`, but translation of `OpUndef` to `poison` does not seem reasonable to me right now.

Actually, that was done now for the purpose of alignment with SPIR-V backend in LLVM

lib/SPIRV/SPIRVReader.cpp
```diff
@@ -1704,7 +1704,7 @@ Value *SPIRVToLLVM::transValueWithoutDecoration(SPIRVValue *BV, Function *F,
   }
 
   case OpUndef:
-    return mapValue(BV, UndefValue::get(transType(BV->getType())));
+    return mapValue(BV, PoisonValue::get(transType(BV->getType())));
```
MrSidims: We can't do the replacement unconditionally. If we really want to invest into this, then we should do this following https://registry.khronos.org/SPIR-V/specs/unified1/SPIRV.html#_validity_and_defined_behavior aka check which instruction generates/consume OpUndef and if consuming this undef/poison value is forbidden for an instruction - undef should be generated in LLVM IR, and poison otherwise.

aobolensk: Fair enough. Shall we follow the logic described in the spec directly? As I can see, there could be several "forbidden consumers" such as conditional branch instructions. Or are there any other assumptions? The check seems to be very non-elegant, because it tries to evaluate the context of the users (other mappers here do not perform such checks)

MrSidims: Please correct me if I'm wrong. The patch addresses [this fixme](https://github.com/KhronosGroup/SPIRV-LLVM-Translator/pull/3661/changes#diff-5d0744e30bd7dabf14b2295d366788037895eb7db8ba2396515fe850fe8ff260L10), right? If so, I would like to understand, why roundtrip through llvm-spirv alone and through backend + llvm-spirv are different before suggesting an approach.

aobolensk: The point is that backend omits custom "undef" initializers, because they don't make sense. Updated the patch to focus on this specific situation along with PR title & description

aobolensk: @MrSidims please, take a look at the updated approach

MrSidims (approved):
LGTM

MrSidims:
@vmaksimo @svenvh friendly ping

svenvh (approved):

vmaksimo (approved):

PR #3667, opened 2026-03-24
Support ClampStochasticRound FP16/BF16 to E2M1

vmaksimo:
Add missing support of Float4/E2M1 as a valid destination type for `OpClampStochasticRoundFToFINTEL` builtin - the only thing left to fully support `SPV_INTEL_fp_conversions` extension.

Updated the design doc accordingly.

Spec:
https://github.com/intel/llvm/blob/sycl/sycl/doc/design/spirv-extensions/SPV_INTEL_fp_conversions.asciidoc

### DIFF
--- a/docs/OCPTypesRepresentationInLLVM.rst
+++ b/docs/OCPTypesRepresentationInLLVM.rst
@@ -134,7 +134,8 @@ for storing the last seed value.
 .. code-block:: C
 
   __builtin_spirv_ClampStochasticRoundFP16ToE5M2INTEL, __builtin_spirv_ClampStochasticRoundFP16ToE4M3INTEL,
-  __builtin_spirv_ClampStochasticRoundBF16ToE5M2INTEL, __builtin_spirv_ClampStochasticRoundBF16ToE4M3INTEL
+  __builtin_spirv_ClampStochasticRoundBF16ToE5M2INTEL, __builtin_spirv_ClampStochasticRoundBF16ToE4M3INTEL,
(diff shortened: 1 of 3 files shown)


YuriPlyakhin (approved):
LGTM

vmaksimo:
@MrSidims friendly-ping, have i answered your concern/question?

MrSidims:
Ah, my review is not blocking here. It makes sense to add missing instructions.

I'm just trying to clarify this. Because, lets say, `ConvertFP16ToE2M1INTEL` and `ClampConvertFP16ToE2M1INTEL` are (per my understanding) effectively the same instruction. And if anybody claims other way - they would need to answer uncomfortable question about "to what value Inf is being converted?".

vmaksimo:
Thanks, Dmitry!
I also assume that these two names act as aliases of the same instruction - just for the naming consistency with the conversion instructions to other types.
I'll rase the question internally to confirm, and if needed we'll follow-up on spec/implementation.

For now I see no blocks for merging, thanks for clarifying that.

MrSidims (approved):

vmaksimo:
No longer relevant with the new spec revision [SPV_INTEL_fp_conversions](https://github.com/intel/llvm/blob/sycl/sycl/doc/design/spirv-extensions/SPV_INTEL_fp_conversions.asciidoc):
Clarified fp4 (E2M1) saturation is unconditional, so a separate ClampStochasticRound\*ToE2M1INTEL form is not needed: it collapses into the existing StochasticRound*ToE2M1INTEL builtin.

Closing in favor of #3782

PR #3672, opened 2026-03-31
Make DebugLinePriority test valid SPIR-V

svenvh:
A recent version of SPIRV-Tools found several issues with the test, such as `DebugTypeFunction` having the wrong return type operand and `DebugTypeBasic` missing the flags operand.


vmaksimo (approved):
Oh, thanks! I actually already made a fix for forward translation (how do you think we got this spt file?😅) to align with the spec - will double-check and publish to review shortly https://github.com/KhronosGroup/SPIRV-LLVM-Translator/commit/4f726a37b00d853bb9e05b785aaa28a904fed116

svenvh:
> will double-check and publish to review shortly

@vmaksimo are you still planning to do this?  No pressure, just checking this is still on your list. :-)

vmaksimo:
@svenvh yeah, my "shortly" took guilty amount of time... 😅
actually #3692 is the first one of the several PR fixes coming - stay tuned! 
i just reordered them semantically for simpler and more clear looking PRs

PR #3674, opened 2026-04-07
Remove intrinsic declarations that became unused after lowering

intwanghao:


lib/SPIRV/SPIRVRegularizeLLVM.cpp
```diff
@@ -798,6 +798,16 @@ bool SPIRVRegularizeLLVMBase::regularize() {
...
 
+  // Remove intrinsic declarations that became unused after lowering.
+  // During the loop above, intrinsic calls are replaced with calls to lowered
+  // functions. The original intrinsic declarations may still remain in the
+  // module with no uses.
+  for (auto I = M->begin(), E = M->end(); I != E;) {
+    Function *F = &(*I++);
+    if (F->isDeclaration() && F->use_empty())
+      F->eraseFromParent();
+  }
```
MrSidims, line 809: If my understanding is correct, as it's spelled right now _every_ unused function declaration will be removed from the module and that is wrong. Please check if removed function is an intrinsic.

PR #3675, opened 2026-04-07
Extend INT4/FP4 packed conversions for i16, i64, and vector packed inputs

vmaksimo:
Packed Int4/FP4 conversion builtins only worked correctly when the packed container was `i8` or `i32`.
Using `i16` or `i64` caused an `"Invalid floating point encoding"` assertion during forward translation.
Using a vector of integers (e.g. `<2 x i8>`) as the packed container produced a size-mismatched bitcast that crashed the round-trip translation.

Extend support to all integer widths (8, 16, 32, 64 bits) and vector-of-integer packed containers for both Int4 and FP4 conversions in both directions.

Extended `SPV_INTEL_float4/conversions_packed.ll` and `SPV_INTEL_int4/conversions_packed.ll` with `i16`, `i64`, and `<2 x i8>` packed container cases in both directions.


YuriPlyakhin (approved):

MrSidims:
My review is not gating

PR #3679, opened 2026-04-08
[Backport to LLVM 17] SPIRVWriter - JointMatrix missing logic

bokrzesi:


vmaksimo:
The code this PR restores was intentionally removed in #2088 as part of the opaque-pointer migration, and that removal predates the llvm_release_170 branch cut. Merging this would bring back code that was already cleaned up.
We should not merge this.

PR #3683, opened 2026-04-14
[LLVM 17] Add workaround to translate casts from target ext types to int (Originally: #3476)

bokrzesi:
This patch should be reverted in the future as it supports translation of invalid SPIR-V modules.

This is forward-porting from: https://github.com/KhronosGroup/SPIRV-LLVM-Translator/pull/3476

This is workaround related to: https://github.com/intel/llvm/commit/f34e5458aa63


bokrzesi:
@michalpaszkowski @vmaksimo 

Could you please review it?

lib/SPIRV/SPIRVReader.cpp
```diff
@@ -1226,6 +1226,12 @@ Value *SPIRVToLLVM::transConvertInst(SPIRVValue *BV, Function *F,
       CO = Instruction::IntToPtr;
     }
     break;
+  // TODO: remove this as it's invalid SPIR-V.
```
```suggestion
  // Workaround: OpConvertPtrToU on a target extension type (e.g. spirv.Image) is
  // invalid SPIR-V, but is generated by the DPC++ frontend for ESIMD
  // get_surface_index() on image accessors.
  // Translate as a builtin call (valid LLVM IR) instead of PtrToInt (invalid IR).
  // Removing this requires fixing the root cause: either a new SPIR-V extension
  // for image-handle-to-index conversion, or redesigning ESIMD accessor storage.
```

michalpaszkowski (approved):

vmaksimo (approved):

PR #3691, opened 2026-04-20
Implement SPV_KHR_abort extension

vmustya:
The SPV_KHR_abort extension introduces the `OpAbortKHR` instruction,
allowing shaders to terminate execution early.

Assisted-by: Claude Opus 4.7 <noreply@anthropic.com>


lib/SPIRV/SPIRVWriter.cpp
vmaksimo, line 5150: Now that OpAbortKHR exists, that TODO should be updated to name OpAbortKHR as the intended target. 
I guess all three intrinsics would translate to OpAbortKHR with a null/undefined message since none of them carry anything message-like. 

That work is a follow-up - doesn't need to be in this PR, but please update the comment and mark it explicitly as a TODO so it doesn't get lost.

test/extensions/KHR/SPV_KHR_abort/abort.ll
vmaksimo, line 1: can we have `spirv-val` RUN lines here?

also, please add a test case where unreachable path is also tested:

```
  define spir_func void @test_abort_unreachable(i32 %msg) {
  entry:
    call spir_func void @_Z16__spirv_AbortKHRIiEvT_(i32 %msg)
    unreachable
  }
```

michalpaszkowski (approved):
LGTM from my side!

lib/SPIRV/SPIRVWriter.cpp
```diff
@@ -2333,10 +2333,21 @@ LLVMToSPIRVBase::transValueWithoutDecoration(Value *V, SPIRVBasicBlock *BB,
...
+      if (Last->getOpCode() == OpAbortKHR)
+        return mapValue(V, const_cast<SPIRVInstruction *>(Last));
     return mapValue(V, BM->addUnreachableInst(BB));
+  }
 
   if (auto *RI = dyn_cast<ReturnInst>(V)) {
+    // SPV_KHR_abort: same suppression as for UnreachableInst above.
+    if (auto *Last = BB->getTerminateInstr())
+      if (Last->getOpCode() == OpAbortKHR)
+        return mapValue(V, const_cast<SPIRVInstruction *>(Last));
```
vmaksimo, line 2350: if a non-void function has `call @__spirv_AbortKHR` the return value is silently dropped.
maybe this needs to be moved after  `if (auto *RV = RI->getReturnValue())` block? not sure though what LLVM semantics is expected here

lib/SPIRV/SPIRVWriter.cpp
```diff
@@ -6900,6 +6913,13 @@ LLVMToSPIRVBase::transBuiltinToInstWithoutDecoration(Op OC, CallInst *CI,
     auto BArgs = transValue(getArguments(CI), BB);
     return BM->addControlBarrierInst(BArgs[0], BArgs[1], BArgs[2], BB);
   } break;
+  case OpAbortKHR: {
+    BM->getErrorLog().checkError(
+        BM->isAllowedToUseExtension(ExtensionID::SPV_KHR_abort),
+        SPIRVEC_RequiresExtension, "SPV_KHR_abort\n");
+    auto *Msg = transValue(CI->getArgOperand(0), BB);
```
vmaksimo, line 6920: we should explicitly check that CI does contain exactly one argument - error if not

lib/SPIRV/libSPIRV/SPIRVInstruction.h
```diff
@@ -858,6 +858,54 @@ template <Op TheOpCode> class SPIRVInstNoOperand : public SPIRVInstruction {
...
+  }
+
+  SPIRVValue *getMessage() const { return getValue(MessageId); }
+  SPIRVType *getMessageType() const {
+    return reinterpret_cast<SPIRVType *>(getEntry(MessageTypeId));
+  }
+
+  std::vector<SPIRVValue *> getOperands() override {
+    return std::vector<SPIRVValue *>(1, getMessage());
+  }
```
vmaksimo, line 896: i don't see these are actually used anywhere. redundant?

vmaksimo (commented):
`SPV_KHR_constant_data` is listed as a dependency, but I don't see it used anywhere in the code - maybe `spirv-val` will catch this as well.
Please also address `clang-tidy` failure

vmustya:
> Please also address `clang-tidy` failure

The clang-tidy fails as follows:
```
/home/runner/work/SPIRV-LLVM-Translator/SPIRV-LLVM-Translator/lib/SPIRV/libSPIRV/SPIRVInstruction.h:903:3: error: visibility of function 'decode' is changed from public in class 'SPIRVEntry' to protected [misc-override-with-different-visibility,-warnings-as-errors]
  903 |   _SPIRV_DEF_ENCDEC2(MessageTypeId, MessageId)
      |   ^
/home/runner/work/SPIRV-LLVM-Translator/SPIRV-LLVM-Translator/lib/SPIRV/libSPIRV/SPIRVEntry.h:151:8: note: expanded from macro '_SPIRV_DEF_ENCDEC2'
  151 |   void decode(std::istream &I) override { getDecoder(I) >> (x) >> (y); }
      |        ^
/home/runner/work/SPIRV-LLVM-Translator/SPIRV-LLVM-Translator/lib/SPIRV/libSPIRV/SPIRVEntry.h:402:16: note: function declared here as public
  402 |   virtual void decode(std::istream &I);
      |                ^
```

The reported issue is that the `encode` and `decode` methods generated by the `_SPIRV_DEF_ENCDEC2` macro are `protected`. However, the base `SPIRVEntry` class defines these methods as `public`. I've checked other instruction implementations, like `OpReturn` and `OpUnreachable`, and they also define the `encode` and `decode` methods as protected. So, this `clang-tidy` issue is applicable to the whole translator code.

I don't really want to deviate the style other instructions are implemented. So, I believe, that either **all** the instructions should be changed by a separate commit, or the clang-tidy error should be suppressed. @vmaksimo, what would you suggest?

test/extensions/KHR/SPV_KHR_abort/abort.ll
vmustya: Done: https://github.com/KhronosGroup/SPIRV-LLVM-Translator/pull/3691/commits/0b02b539d9a8c30b489de02a2ebc5d6302ec5348

lib/SPIRV/SPIRVWriter.cpp
vmustya: Done: https://github.com/KhronosGroup/SPIRV-LLVM-Translator/pull/3691/commits/0b02b539d9a8c30b489de02a2ebc5d6302ec5348

lib/SPIRV/SPIRVWriter.cpp
vmustya: Done: https://github.com/KhronosGroup/SPIRV-LLVM-Translator/pull/3691/commits/0b02b539d9a8c30b489de02a2ebc5d6302ec5348

lib/SPIRV/libSPIRV/SPIRVInstruction.h
vmustya: Removed: https://github.com/KhronosGroup/SPIRV-LLVM-Translator/pull/3691/commits/0b02b539d9a8c30b489de02a2ebc5d6302ec5348

lib/SPIRV/SPIRVWriter.cpp
vmustya: Done: https://github.com/KhronosGroup/SPIRV-LLVM-Translator/pull/3691/commits/0b02b539d9a8c30b489de02a2ebc5d6302ec5348

lib/SPIRV/libSPIRV/SPIRVInstruction.h
vmustya: The `getOperand` method override is required for the translation.

test/extensions/KHR/SPV_KHR_abort/abort.ll
vmustya: > can we have `spirv-val` RUN lines here?

@vmaksimo , the `spirv-val` tool used for the CI is too old to recognize the `SPV_KHR_abort` extension. However, locally-built top of trunk version works correctly. Should we disable the `spirv-val` checks until the CI is updated?

lib/SPIRV/SPIRVWriter.cpp
```diff
@@ -5131,9 +5148,10 @@ SPIRVValue *LLVMToSPIRVBase::transIntrinsicInst(IntrinsicInst *II,
   case Intrinsic::invariant_start:
   case Intrinsic::invariant_end:
   case Intrinsic::dbg_label:
-  // llvm.trap and llvm.debugtrap intrinsics are not implemented. But for now
-  // don't crash. This change is pending the trap/abort intrinsic
-  // implementation.
+  // TODO: lower llvm.trap / llvm.ubsantrap / llvm.debugtrap to OpAbortKHR
```
maarquitos14, line 5144: Could we do this in this patch too? Seems very related.

lib/SPIRV/libSPIRV/SPIRVInstruction.h
```diff
@@ -858,6 +858,49 @@ template <Op TheOpCode> class SPIRVInstNoOperand : public SPIRVInstruction {
 typedef SPIRVInstNoOperand<OpReturn> SPIRVReturn;
 typedef SPIRVInstNoOperand<OpUnreachable> SPIRVUnreachable;
 
+class SPIRVAbortKHR : public SPIRVInstruction {
+public:
+  static const Op OC = OpAbortKHR;
+  static const SPIRVWord FixedWordCount = 3;
+  // Complete constructor
```
maarquitos14:
```suggestion
  // Complete constructor.
```

lib/SPIRV/libSPIRV/SPIRVInstruction.h
```diff
@@ -858,6 +858,49 @@ template <Op TheOpCode> class SPIRVInstNoOperand : public SPIRVInstruction {
...
+  SPIRVAbortKHR(SPIRVValue *TheMessage, SPIRVBasicBlock *TheBB)
+      : SPIRVInstruction(FixedWordCount, OC, TheBB),
+        MessageTypeId(TheMessage->getType()->getId()),
+        MessageId(TheMessage->getId()) {
+    setAttr();
+    validate();
+    assert(TheBB && "Invalid BB");
+  }
+  // Incomplete constructor
```
maarquitos14:
```suggestion
  // Incomplete constructor.
```

lib/SPIRV/SPIRVWriter.cpp
```diff
@@ -2333,10 +2333,27 @@ LLVMToSPIRVBase::transValueWithoutDecoration(Value *V, SPIRVBasicBlock *BB,
...
+    // SPV_KHR_abort: OpAbortKHR is itself a SPIR-V block terminator. If the
+    // LLVM IR appends a trailing `ret void` after the abort call, do not emit
+    // a second SPIR-V terminator. We deliberately limit this suppression to
+    // `ret void`: if a non-void function ends with `ret <value>` after
+    // OpAbortKHR, fall through so that the value is still translated and the
+    // user sees a normal SPIR-V validation error rather than silently
+    // dropping the return value.
+    if (auto *Last = BB->getTerminateInstr();
+        !RI->getReturnValue() && Last && Last->getOpCode() == OpAbortKHR)
+      return mapValue(V, const_cast<SPIRVInstruction *>(Last));
```
maarquitos14: I am not sure this is enough to handle all the cases. For example:
```
call void @llvm.trap()/__spirv_AbortKHR
cal void @llvm.lifetime.end()
ret void
```

This would be invalid, because we would have an instruction after a terminator. As far as I can tell, nothing would prevent translating the call to @llvm.lifetime.end() after the OpAbortKHR, which is a terminator. Am I right?

lib/SPIRV/SPIRVReader.cpp
```diff
@@ -1951,6 +1939,21 @@ Value *SPIRVToLLVM::transValueWithoutDecoration(SPIRVValue *BV, Function *F,
                                transValue(RV->getReturnValue(), F, BB), BB));
   }
 
+  case OpAbortKHR: {
+    // OpAbortKHR is a SPIR-V block terminator. In LLVM IR, model it as a call
```
maarquitos14, line 1955: Should we split in two paths? This one for spirv-friendly IR, and another one for non-friendly which would translate into... llvm.trap maybe?

maarquitos14 (commented):
I was looking at SPV_KHR_abort separately, but @MrSidims brought this PR to my attention. I have a few extra tests that I'm more than happy to share if you're interested @vmustya.

vmustya:
> I have a few extra tests that I'm more than happy to share if you're interested @vmustya.

@maarquitos14, sure! I'd really appreciate that.

maarquitos14:
> > I have a few extra tests that I'm more than happy to share if you're interested @vmustya.
> 
> @maarquitos14, sure! I'd really appreciate that.

You can find them in https://github.com/maarquitos14/SPIRV-LLVM-Translator/commit/fe51642c3a75d5cc02f5515668c772fa45f2e509. I think all of them assume llvm.trap gets translated to OpAbortKHR and back, but still I think they can be useful.

lib/SPIRV/SPIRVWriter.cpp
vmaksimo: @maarquitos14 Maybe not to overload this PR, we can follow-up in a separate one (together with Reader part update for default/non-friendly IR path). What do you think?

test/extensions/KHR/SPV_KHR_abort/abort.ll
vmaksimo: Ouch, I was afraid of this.
Yes please, you may put smth like `RUN: not spirv-val ; TODO: remove "not" once SPIR-V tools are updated` so the test will start to fail in CI once tools get updated.

vmaksimo:
> I don't really want to deviate the style other instructions are implemented. So, I believe, that either all the instructions should be changed by a separate commit, or the clang-tidy error should be suppressed. 

@vmustya If can - we should not introduce new errors, and I believe this is the case. So please just put encode/decode to public - we'll take a look at other instructions separately.

lib/SPIRV/SPIRVWriter.cpp
maarquitos14: I won't block this PR for this. If you want to follow-up in a separate PR, go for it.

lib/SPIRV/SPIRVWriter.cpp
vmustya: @maarquitos14, I've moved the tests you've shared from the `llvm.trap` intrinsic to the SPIR-V friendly IR, and adjusted the writer code. Now all the instructions in the basic block, that follow the `__spirv_AbortKHR` function call, are ignored by the writer. Could you please provide your feedback?

maarquitos14 (approved):
LGTM.

lib/SPIRV/SPIRVWriter.cpp
maarquitos14: LGTM! Thanks :)

test/extensions/KHR/SPV_KHR_abort/abort.ll
```diff
@@ -0,0 +1,52 @@
+; RUN: llvm-as %s -o %t.bc
+; RUN: llvm-spirv %t.bc -o %t.spv --spirv-ext=+SPV_KHR_abort
+; RUN: llvm-spirv %t.spv -o %t.spt --to-text
+; RUN: FileCheck < %t.spt %s --check-prefix=CHECK-SPIRV
+; RUN: llvm-spirv %t.spv -o %t.rev.bc -r --spirv-target-env=SPV-IR
+; RUN: llvm-dis %t.rev.bc -o %t.rev.ll
+; RUN: FileCheck < %t.rev.ll %s --check-prefix=CHECK-LLVM
+
+; FIXME: enable the following run when the translator CI is updated to a new
+; verion of the SPIR-V Tools that includes the support for the SPV_KHR_abort
```
vmaksimo: This typo is present in other test with FIXME as well
```suggestion
; version of the SPIR-V Tools that includes the support for the SPV_KHR_abort
```

lib/SPIRV/SPIRVReader.cpp
```diff
@@ -1951,6 +1939,21 @@ Value *SPIRVToLLVM::transValueWithoutDecoration(SPIRVValue *BV, Function *F,
...
+    // to the SPIR-V friendly builtin __spirv_AbortKHR followed by an
+    // 'unreachable' terminator.
+    auto *AbortInst =
+        transSPIRVBuiltinFromInst(static_cast<SPIRVInstruction *>(BV), BB);
```
vmaksimo: nit: just for styling matter
```suggestion
        transSPIRVBuiltinFromInst(static_cast<SPIRVAbortKHR *>(BV), BB);
```

test/extensions/KHR/SPV_KHR_abort/no-abort-unaffected.ll
```diff
@@ -0,0 +1,88 @@
+; Sanity check: enabling SPV_KHR_abort does not affect functions that don't
+; call __spirv_AbortKHR. Normal terminators (OpReturn, OpReturnValue,
+; OpBranch, OpUnreachable) must be preserved.
+
+; RUN: llvm-as %s -o %t.bc
+; RUN: llvm-spirv %t.bc --spirv-ext=+SPV_KHR_abort -o %t.spv
+; RUN: llvm-spirv %t.spv -to-text -o %t.spt
+; RUN: FileCheck < %t.spt %s --check-prefix=CHECK-SPIRV
```
vmaksimo: Better use implicit negative checks for the specified pattern between positive checks:
```suggestion
; RUN: FileCheck < %t.spt %s --check-prefix=CHECK-SPIRV --implicit-check-not AbortKHR
```
This way explicit `CHECK-SPIRV-NOT` can be removed.

test/extensions/KHR/SPV_KHR_abort/abort-multiple-blocks.ll
```diff
@@ -0,0 +1,85 @@
...
+;
+; Return block: normal return
+; CHECK-SPIRV: ReturnValue
+;
+; First abort block
+; CHECK-SPIRV: AbortKHR
+;
+; Second abort block
+; CHECK-SPIRV: AbortKHR
+; CHECK-SPIRV: FunctionEnd
```
vmaksimo, line 53: Please add checks for Label/Branch name so we know that `AbortKHR` are inside the expected branches, e.g.:
```
Name [[#LabelWork:]] "work"
Name [[#LabelErr1:]] "err1"
BranchConditional [[#]] [[#LabelWork]] [[#LabelErr1]]
Label [[#LabelWork]]
...
Label [[#LabelErr1]]
```

test/extensions/KHR/SPV_KHR_abort/abort-in-kernel.ll
```diff
@@ -0,0 +1,93 @@
...
+; RUN: not spirv-val %t.spv
+
+; Round-trip
+; RUN: llvm-spirv -r %t.spv -o %t.rev.bc
+; RUN: llvm-dis < %t.rev.bc | FileCheck %s --check-prefix=CHECK-LLVM
+
+; ---- SPIR-V ----
+; CHECK-SPIRV-DAG: Capability AbortKHR
+; CHECK-SPIRV-DAG: Extension "SPV_KHR_abort"
+; CHECK-SPIRV-DAG: EntryPoint 6 {{[0-9]+}} "test_kernel"
```
vmaksimo:
```suggestion
; CHECK-SPIRV-DAG: EntryPoint 6 [[#KernelId:]] "test_kernel"
```
To reuse it later like: `CHECK-SPIRV: Function {{.*}} [[#KernelId]]`

vmaksimo (commented):
The change with the test coverege looks great, thanks for the work done!

I have the nits about the tests that'd better to be addressed, after that I guess we're perfectly good to go 🙂

vmaksimo (approved):
LGTM

vmaksimo:
@svenvh would you like to review this one prior merging?

svenvh (approved):

PR #3704, opened 2026-04-27
[NFC] Use tool substitution for spirv-val in lit config

aobolensk:
Bare-string substitutions match as substrings and the replacement path contains the tool name, causing corrupted RUN lines

Port of the original patch in LLVM SPIR-V backend: https://github.com/llvm/llvm-project/pull/192462

### DIFF
--- a/test/lit.cfg.py
+++ b/test/lit.cfg.py
@@ -85,7 +85,7 @@ if config.spirv_tools_have_spirv_val:
     llvm_config.add_tool_substitutions(['spirv-val'], [config.spirv_tools_bin_dir])
     using_spirv_tools = True
 else:
-    llvm_config.add_tool_substitutions(['spirv-val'])
+    llvm_config.add_tool_substitutions([ToolSubst('spirv-val', ':')])
 
 if not config.llvm_spirv_build_external and config.llvm_build_shared_libs:
     config.available_features.add('pass-plugin')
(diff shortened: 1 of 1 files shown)


MrSidims (approved):

vmaksimo (approved):

test/lit.cfg.py
```diff
@@ -85,7 +85,7 @@
     llvm_config.add_tool_substitutions(['spirv-val'], [config.spirv_tools_bin_dir])
     using_spirv_tools = True
 else:
-    config.substitutions.append(('spirv-val', ':'))
+    llvm_config.add_tool_substitutions(['spirv-val'])
```
svenvh: Not sure this is equivalent?  Before your patch, we'd skip spirv-val executions without failing.  With your patch, wouldn't this still run any spirv-val steps unconditionally if there happens to be a spirv-val in a default path?

aobolensk: The key point here is boundary check around `spirv-val`. So if it is a part of the path, approach with `config.substitutions.append` is going to replace it, in this case it will loop up for "spirv-val" as a standalone word, so in this case if your path contains substring "spirv-val" it is not going to be replaced

svenvh: To be clear: I'm not questioning that the previous approach can be problematic.  I'm questioning whether your solution also completely disables spirv-val executions as the previous approach does.

aobolensk: I see what you mean, you're right, thank you. Applied the fix, please take a look

svenvh: Thanks, that looks better indeed.

svenvh (approved):

PR #3708, opened 2026-04-28
[NFC] Suppress coverity issue

KornevNikita:


KornevNikita:
Failures are known.

KornevNikita:
@vmaksimo could you please take a look?:))

vmaksimo (commented):
Should not you just mark Coverity hit as false positive instead? 
I'd prefer not to add knowingly dead code

KornevNikita:
> Should not you just mark Coverity hit as false positive instead?

Yep that's what I did actually, just wanted to make sure other people won't hit this. But I do not insist:)

vmaksimo:
My thinking is that we should not be provoked by Coverity that is not able to analyze LLVM framework..
@svenvh do you agree to close this one?

svenvh:
> My thinking is that we should not be provoked by Coverity that is not able to analyze LLVM framework..
> @svenvh do you agree to close this one?

Yes that sounds reasonable to me.

PR #3716, opened 2026-04-29
Configurable address space numbering

mbelicki:
This change adds a new configuration knob that allows the user of SPIR-V reader to provide custom address space numbering to the translator. The user can provide an array with values for each logical SPIR-V address space (private, local, global, generic, constant etc.) that will be used by the translator when generating LLVM IR. The behavior of the reader remains unchanged if user does not explicitly change the setting.

This change was motivated by a need to support compilation target that uses different convention for address space numbering than SPIR-V. Generally the GPU targets in LLVM (AMDGPU, NVPTX) seem to use 0 (LLVM's default address space) as rough equivalent of generic address space. Without this option an additional address space renumbering pass is needed, while such pass is possible, it is somewhat fragile (need to visit all places in module where ptr type can be used, they change as LLVM changes) and consumes compilation time (this can't be done in single pass due to potential type/value dependencies). Configuring this in SPIR-V translator seems to be a simpler and cleaner solution.

The change replaces all hard-coded address space values in SPIR-V reader with a call that resolves the logical SPIR-V address space to value provided by the user. If the mapping was not configured the original values are used. Additionally analogous change was made to builtin name mangling.

### DIFF
--- a/include/LLVMSPIRVOpts.h
+++ b/include/LLVMSPIRVOpts.h
@@ -42,6 +42,7 @@
 #include <llvm/ADT/SmallVector.h>
 #include <llvm/ADT/StringRef.h>
 
+#include <array>
 #include <cassert>
 #include <cstdint>
 #include <map>
@@ -117,6 +118,22 @@ enum class DebugInfoEIS : uint32_t {
 };
 
 enum class BuiltinFormat : uint32_t { Function, Global };
+
+enum SPIRAddressSpace : uint32_t {
+  SPIRAS_Private,
+  SPIRAS_Global,
+  SPIRAS_Constant,
+  SPIRAS_Local,
+  SPIRAS_Generic,
+  SPIRAS_GlobalDevice,
+  SPIRAS_GlobalHost,
+  SPIRAS_Input,
+  SPIRAS_Output,
+  SPIRAS_CodeSectionINTEL,
+  SPIRAS_Count,
+};
+
+using AddrSpaceMap = std::array<uint32_t, SPIRAS_Count>;
 
 /// \brief Helper class to manage SPIR-V translation
 class TranslatorOpts {
@@ -245,6 +262,22 @@ public:
     EmitFunctionPtrAddrSpace = Value;
   }
 
+  void setAddrSpaceMap(AddrSpaceMap Map) noexcept { ASMap = std::move(Map); }
+
+  uint32_t mapAddrSpace(uint32_t SPIRAS) const noexcept {
+    if (ASMap && SPIRAS < SPIRAS_Count)
+      return (*ASMap)[SPIRAS];
+    return SPIRAS;
+  }
+
+  void setFunctionProgramAddrSpace(uint32_t AS) noexcept {
+    FunctionProgramAS = AS;
+  }
+
+  uint32_t getFunctionProgramAddrSpace() const noexcept {
+    return FunctionProgramAS;
+  }
+
   void setBuiltinFormat(BuiltinFormat Value) noexcept {
     SPIRVBuiltinFormat = Value;
   }
@@ -345,6 +378,12 @@ private:
   // Controls if CodeSectionINTEL can be emitted and consumed with a dedicated
   // address space
   bool EmitFunctionPtrAddrSpace = false;
+
+  // Optional per-target mapping from SPIRAddressSpace values to LLVM AS numbers
+  std::optional<AddrSpaceMap> ASMap;
+
+  // Address space for function definitions; independent of the data-pointer map
+  uint32_t FunctionProgramAS = SPIRAS_Private;
 
   bool PreserveAuxData = false;
 
--- a/include/LLVMSPIRVOpts.h
+++ b/include/LLVMSPIRVOpts.h
@@ -264,6 +264,10 @@ public:
 
   void setAddrSpaceMap(AddrSpaceMap Map) noexcept { ASMap = std::move(Map); }
 
+  const AddrSpaceMap *getAddrSpaceMap() const noexcept {
+    return ASMap ? &*ASMap : nullptr;
+  }
+
   uint32_t mapAddrSpace(uint32_t SPIRAS) const noexcept {
     if (ASMap && SPIRAS < SPIRAS_Count)
       return (*ASMap)[SPIRAS];
--- a/include/LLVMSPIRVOpts.h
+++ b/include/LLVMSPIRVOpts.h
@@ -265,7 +265,7 @@ public:
   void setAddrSpaceMap(AddrSpaceMap Map) noexcept { ASMap = std::move(Map); }
 
   const AddrSpaceMap *getAddrSpaceMap() const noexcept {
-    return ASMap ? &*ASMap : nullptr;
+    return ASMap ? &ASMap.value() : nullptr;
   }
 
   uint32_t mapAddrSpace(uint32_t SPIRAS) const noexcept {
@@ -279,7 +279,9 @@ public:
   }
 
   uint32_t getFunctionProgramAddrSpace() const noexcept {
-    return FunctionProgramAS;
+    if (FunctionProgramAS)
+      return *FunctionProgramAS;
+    return mapAddrSpace(SPIRAS_Private);
   }
 
   void setBuiltinFormat(BuiltinFormat Value) noexcept {
@@ -380,14 +382,19 @@ private:
   bool PreserveOCLKernelArgTypeMetadataThroughString = false;
 
   // Controls if CodeSectionINTEL can be emitted and consumed with a dedicated
-  // address space
+  // address space. This option overrides address space set in
+  // FunctionProgramAS for functions that have their address taken with
+  // SPIRAS_CodeSectionINTEL.
   bool EmitFunctionPtrAddrSpace = false;
 
-  // Optional per-target mapping from SPIRAddressSpace values to LLVM AS numbers
+  // Optional per-target mapping from SPIRAddressSpace values to LLVM AS
+  // numbers.
   std::optional<AddrSpaceMap> ASMap;
 
-  // Address space for function definitions; independent of the data-pointer map
-  uint32_t FunctionProgramAS = SPIRAS_Private;
+  // Address space for function definitions. When set, overrides the address
+  // space map. Otherwise getFunctionProgramAddrSpace() falls back to
+  // SPIRAS_Private, which may be resolved unsign ASMap if set.
+  std::optional<uint32_t> FunctionProgramAS;
 
   bool PreserveAuxData = false;
 
--- a/include/LLVMSPIRVOpts.h
+++ b/include/LLVMSPIRVOpts.h
@@ -262,7 +262,7 @@ public:
     EmitFunctionPtrAddrSpace = Value;
   }
 
-  void setAddrSpaceMap(AddrSpaceMap Map) noexcept { ASMap = std::move(Map); }
+  void setAddrSpaceMap(AddrSpaceMap Map) noexcept { ASMap = Map; }
 
   const AddrSpaceMap *getAddrSpaceMap() const noexcept {
     return ASMap ? &ASMap.value() : nullptr;
@@ -393,7 +393,7 @@ private:
 
   // Address space for function definitions. When set, overrides the address
   // space map. Otherwise getFunctionProgramAddrSpace() falls back to
-  // SPIRAS_Private, which may be resolved unsign ASMap if set.
+  // SPIRAS_Private, which may be resolved using ASMap if set.
   std::optional<uint32_t> FunctionProgramAS;
 
   bool PreserveAuxData = false;
--- a/lib/SPIRV/SPIRVReader.cpp
+++ b/lib/SPIRV/SPIRVReader.cpp
@@ -369,16 +369,18 @@ Type *SPIRVToLLVM::transType(SPIRVType *T, bool UseTPT) {
     if (BM->shouldEmitFunctionPtrAddrSpace() &&
         T->getPointerElementType()->getOpCode() == OpTypeFunction)
       AS = SPIRAS_CodeSectionINTEL;
+    unsigned MappedAS = BM->mapAddrSpace(AS);
     Type *ElementTy = transType(T->getPointerElementType(), UseTPT);
     if (UseTPT)
-      return TypedPointerType::get(ElementTy, AS);
-    return mapType(T, PointerType::get(*Context, AS));
+      return TypedPointerType::get(ElementTy, MappedAS);
+    return mapType(T, PointerType::get(*Context, MappedAS));
   }
   case OpTypeUntypedPointerKHR: {
     unsigned AS = SPIRSPIRVAddrSpaceMap::rmap(T->getPointerStorageClass());
     if (AS == SPIRAS_CodeSectionINTEL && !BM->shouldEmitFunctionPtrAddrSpace())
       AS = SPIRAS_Private;
-    return mapType(T, PointerType::get(*Context, AS));
+    unsigned MappedAS = BM->mapAddrSpace(AS);
+    return mapType(T, PointerType::get(*Context, MappedAS));
   }
   case OpTypeVector:
     return mapType(T,
@@ -465,7 +467,7 @@ Type *SPIRVToLLVM::transType(SPIRVType *T, bool UseTPT) {
     Type *Ty = nullptr;
     if (UseTPT) {
       Type *StructTy = getOrCreateOpaqueStructType(M, transVCTypeName(PST));
-      Ty = TypedPointerType::get(StructTy, SPIRAS_Global);
+      Ty = TypedPointerType::get(StructTy, BM->mapAddrSpace(SPIRAS_Global));
     } else {
       std::vector<unsigned> Params;
       if (PST->hasAccessQualifier()) {
@@ -1406,7 +1408,7 @@ Value *SPIRVToLLVM::oclTransConstantPipeStorage(
   return new GlobalVariable(*M, CPSTy, false, GlobalValue::LinkOnceODRLinkage,
                             ConstantStruct::get(CPSTy, CPSElems),
                             BCPS->getName(), nullptr,
-                            GlobalValue::NotThreadLocal, SPIRAS_Global);
+                            GlobalValue::NotThreadLocal, BM->mapAddrSpace(SPIRAS_Global));
 }
 
 // Translate aliasing memory access masks for SPIRVLoad and SPIRVStore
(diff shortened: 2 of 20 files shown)


MrSidims (commented):
I haven't taken a close look at the code, but I love the idea. It should have been done a long time ago, great to see it's being addressed.

test/spirv-addrspace-map.ll
```diff
@@ -0,0 +1,85 @@
...
+}
+
+declare ptr @llvm.stacksave.p0() #0
+declare void @llvm.stackrestore.p0(ptr) #0
+declare spir_func void @_Z24__spirv_FixedSinCosINTELILi34ELi66EEU7_ExtIntIXmlLi2ET0_EEiU7_ExtIntIXT_EEibiiii(ptr addrspace(4) sret(i66) align 8, i34, i1 zeroext, i32, i32, i32, i32) #0
+
+attributes #0 = { nounwind }
+
+; CHECK-ERR-IDX: not a valid SPIRAddressSpace index
+; CHECK-ERR-FMT: Invalid format for --spirv-addrspace-map
```
vmaksimo: Let's move these up right after RUN lines for readability

test/spirv-addrspace-map-untyped.ll
```diff
@@ -0,0 +1,27 @@
...
+; RUN:   -o - | llvm-dis | FileCheck %s --check-prefix=CHECK-MAPPED
+; RUN: llvm-spirv -r %t.untyped.spv \
+; RUN:   -o - | llvm-dis | FileCheck %s --check-prefix=CHECK-IDENTITY
+
+target datalayout = "e-i64:64-v16:16-v24:32-v32:32-v48:64-v96:128-v192:256-v256:256-v512:512-v1024:1024-G1"
+target triple = "spir64-unknown-unknown"
+
+; With mapping 4->0, the generic AS4 pointer remaps to AS0 (printed as bare ptr).
+; CHECK-MAPPED: define{{.*}} @test_generic_atomic(
+; CHECK-MAPPED: call spir_func i32 @{{.*}}atomic{{.*}}(ptr{{( addrspace\(0\))?}}
```
vmaksimo: We'd better have precise CHECK here - with either `ptr addrspace(0)` or `ptr`. Also full function call name would be better. (same for other test cases)
P.S. oh, are we really not able to translate atomic instructions back to atomic instructions? or is it OCL builtin style after reverse translation?

test/spirv-addrspace-map.ll
```diff
@@ -0,0 +1,85 @@
+; Test --spirv-addrspace-map address space remapping on SPIR-V -> LLVM translation.
+
+; RUN: llvm-as < %s -o %t.bc
+; RUN: llvm-spirv %t.bc --spirv-ext=+SPV_INTEL_variable_length_array,+SPV_INTEL_arbitrary_precision_integers,+SPV_INTEL_arbitrary_precision_fixed_point -o %t.spv
```
vmaksimo: I doubt the need to have the main test requiring  any SPIR-V extensions for translation.
Was there a specific use case you wanted to test? If so, please move these test functions in a separate test file.

tools/llvm-spirv/llvm-spirv.cpp
```diff
@@ -758,6 +774,40 @@ bool parseSpecConstOpt(llvm::StringRef SpecConstStr,
...
+      return true;
+    }
+    unsigned From, To;
+    if (KV[0].getAsInteger(10, From) || From >= SPIRV::SPIRAS_Count) {
+      errs() << "Error: --spirv-addrspace-map: \"" << KV[0]
+             << "\" is not a valid SPIRAddressSpace index (0-"
+             << (SPIRV::SPIRAS_Count - 1) << ").\n";
+      return true;
+    }
+    if (KV[1].getAsInteger(10, To)) {
```
vmaksimo, line 801: Just a guess for now, should not we somehow limit the target (`To`) address space value as well as the `From`?
```suggestion
    if (KV[1].getAsInteger(10, To) || To >= SPIRV::SPIRAS_Count) {
```

tools/llvm-spirv/llvm-spirv.cpp
```diff
@@ -169,6 +169,22 @@ static cl::opt<bool> SPIRVEmitFunctionPtrAddrSpace(
...
+    "spirv-addrspace-map",
+    cl::desc("Remap SPIR-V address spaces when translating SPIR-V to LLVM IR.\n"
+             "Comma-separated list of from:to pairs, e.g. \"0:4,4:0\".\n"
+             "from is a SPIRAddressSpace index (0-9); to is the LLVM IR "
+             "address space number to emit.\n"
+             "Unmapped entries keep their original SPIRAS value."),
+    cl::value_desc("from1:to1,from2:to2,..."));
+
+static cl::opt<unsigned> SPIRVFunctionProgramAddrSpace(
+    "spirv-function-program-addrspace",
```
vmaksimo, line 182: Any test for this one?

include/LLVMSPIRVOpts.h
```diff
@@ -245,6 +262,26 @@ class TranslatorOpts {
...
+  }
+
+  uint32_t mapAddrSpace(uint32_t SPIRAS) const noexcept {
+    if (ASMap && SPIRAS < SPIRAS_Count)
+      return (*ASMap)[SPIRAS];
+    return SPIRAS;
+  }
+
+  void setFunctionProgramAddrSpace(uint32_t AS) noexcept {
+    FunctionProgramAS = AS;
```
vmaksimo, line 278: Should we add upper bound validation here?

vmaksimo (commented):
Initial review (only tests and new options for now)

include/LLVMSPIRVOpts.h
```diff
@@ -245,6 +262,26 @@ class TranslatorOpts {
     EmitFunctionPtrAddrSpace = Value;
   }
 
+  void setAddrSpaceMap(AddrSpaceMap Map) noexcept { ASMap = std::move(Map); }
+
+  const AddrSpaceMap *getAddrSpaceMap() const noexcept {
+    return ASMap ? &*ASMap : nullptr;
```
maarquitos14:
```suggestion
    return ASMap ? &ASMap.value() : nullptr;
```

include/LLVMSPIRVOpts.h
```diff
@@ -346,6 +383,12 @@ class TranslatorOpts {
   // address space
   bool EmitFunctionPtrAddrSpace = false;
 
+  // Optional per-target mapping from SPIRAddressSpace values to LLVM AS numbers
```
maarquitos14:
```suggestion
  // Optional per-target mapping from SPIRAddressSpace values to LLVM AS numbers.
```

include/LLVMSPIRVOpts.h
```diff
@@ -346,6 +383,12 @@ class TranslatorOpts {
...
+  std::optional<AddrSpaceMap> ASMap;
+
+  // Address space for function definitions; independent of the data-pointer map
```
maarquitos14:
```suggestion
  // Address space for function definitions; independent of the data-pointer map.
```

maarquitos14 (approved):
I love the idea, just a few nits.

test/spirv-addrspace-map.ll
mbelicki: Fixed in f516762.

test/spirv-addrspace-map-untyped.ll
mbelicki: Fixed in f516762, sorry this was left over from me starting this test from a copy of a different test, it doesn't make sense here as the mangling also is affected by address space mapping and should be checked in full.

test/spirv-addrspace-map.ll
mbelicki: SPV_INTEL_variable_length_array is needed for alloca/stacksave/restore validation, I moved it to a separate test case. The arbitrary int/fp came from a test case that I used to validate mangling, it doesn't make sense to conflate both here, so I removed the extensions.

tools/llvm-spirv/llvm-spirv.cpp
mbelicki: The `To` parameter refers here to the address space that will be emitted by SPIR-V reader. LLVM allows any unsigned number here, targets may choose to use more address spaces than SPIR-V defines (or a different numbering schema e.g. 0, 10, 20, etc.), I think leaving this flexibility here makes sense.

tools/llvm-spirv/llvm-spirv.cpp
mbelicki: Sorry, forgot to include in the PR. I've added the test, while adding the test I expanded it to cover interactions with address space mapping and CodeSectionINTEL. This caused me to modify the behavior of this option a to be a little bit more intuitive.

include/LLVMSPIRVOpts.h
mbelicki: The `AS` parameter currently represents the target address space number in generated code, it can be any numeric value (same as values used in mapping). I can change this behavior to accept SPIR-V logical address spaces here (private, local, etc.) instead. My impression however was that it's more ergonomic to have the same value here as in case of mapping. The SPIR-V consumers are expected to use this in following way right now:

    Opts.setFunctionProgramAddrSpace(CustomLLVMTarget::AddressSpace::GENERIC);

include/LLVMSPIRVOpts.h
mbelicki: Fixed in f516762.

include/LLVMSPIRVOpts.h
mbelicki: Fixed in f516762.

include/LLVMSPIRVOpts.h
mbelicki: Fixed in f516762, although the comment also changed a bit to better reflect the option usage.

mbelicki:
@vmaksimo @maarquitos14 thank you for your reviews! I applied your comments and introduced some changes to how function address space is handled. Previously it was only applied to functions which had their address taken, which caused two problems: (1) it was a little bit unintuitive, (2) the address space used for functions which didn't have their address taken was always SPIRAS_Private without mapping applied, which created a gap in how the mapping is applied.

include/LLVMSPIRVOpts.h
```diff
@@ -343,9 +382,20 @@ class TranslatorOpts {
...
+  // SPIRAS_CodeSectionINTEL.
   bool EmitFunctionPtrAddrSpace = false;
 
+  // Optional per-target mapping from SPIRAddressSpace values to LLVM AS
+  // numbers.
+  std::optional<AddrSpaceMap> ASMap;
+
+  // Address space for function definitions. When set, overrides the address
+  // space map. Otherwise getFunctionProgramAddrSpace() falls back to
+  // SPIRAS_Private, which may be resolved unsign ASMap if set.
```
maarquitos14:
```suggestion
  // SPIRAS_Private, which may be resolved using ASMap if set.
```

I think?

maarquitos14 (approved):
LGTM, just a nit.

lib/SPIRV/SPIRVReader.cpp
```diff
@@ -1887,22 +1886,26 @@ Value *SPIRVToLLVM::transValueWithoutDecoration(SPIRVValue *BV, Function *F,
     auto *VLA = static_cast<SPIRVVariableLengthArrayINTEL *>(BV);
     llvm::Type *Ty = transType(BV->getType()->getPointerElementType());
     llvm::Value *ArrSize = transValue(VLA->getOperand(0), F, BB);
-    return mapValue(BV,
-                    new AllocaInst(Ty, M->getDataLayout().getAllocaAddrSpace(),
-                                   ArrSize, BV->getName(), BB));
+    return mapValue(BV, new AllocaInst(Ty, BM->mapAddrSpace(SPIRAS_Private),
+                                       ArrSize, BV->getName(), BB));
```
MrSidims: I believe this code must be untouched. Instead the patch should change datalayout of the module, re-defining default GV and alloca address spaces.

test/spirv-addrspace-map-builtin-mangle.ll
```diff
@@ -0,0 +1,43 @@
...
+
+; RUN: llvm-as %s -o %t.bc
+; RUN: llvm-spirv %t.bc -o %t.spv
+; RUN: spirv-val %t.spv
+; RUN: llvm-spirv -r --spirv-target-env=SPV-IR %t.spv -o - \
+; RUN:   | llvm-dis | FileCheck %s --check-prefix=CHECK-DEFAULT
+; RUN: llvm-spirv -r --spirv-target-env=SPV-IR --spirv-addrspace-map=1:3 %t.spv -o - \
+; RUN:   | llvm-dis | FileCheck %s --check-prefix=CHECK-MAPPED
+
+target datalayout = "e-i64:64-v16:16-v24:32-v32:32-v48:64-v96:128-v192:256-v256:256-v512:512-v1024:1024"
```
MrSidims, line 12: We should re-define datalayout during reverse translation depending on a map and check it in some test.

tools/llvm-spirv/llvm-spirv.cpp
```diff
@@ -169,6 +169,22 @@ static cl::opt<bool> SPIRVEmitFunctionPtrAddrSpace(
...
+    cl::desc("Address space to assign to function definitions when translating "
+             "SPIR-V to LLVM IR (used for OpConstantFunctionPointerINTEL). "
+             "Defaults to 0 (SPIRAS_Private)."),
+    cl::value_desc("address-space"));
```
MrSidims, line 187: I feel like it should be an option for both forward and reverse translation. Not a hard request for this PR, if you don't plan to do this - just flag it.

tools/llvm-spirv/llvm-spirv.cpp
```diff
@@ -758,6 +774,40 @@ bool parseSpecConstOpt(llvm::StringRef SpecConstStr,
   return false;
 }
 
+// Returns true on error.
+static bool parseAddrSpaceMapOpt(SPIRV::TranslatorOpts &Opts) {
+  SPIRV::AddrSpaceMap ASMap;
```
MrSidims: Local name shadows global name, lets rename either of them to avoid confusion.

include/LLVMSPIRVOpts.h
```diff
@@ -245,6 +262,28 @@ class TranslatorOpts {
     EmitFunctionPtrAddrSpace = Value;
   }
 
+  void setAddrSpaceMap(AddrSpaceMap Map) noexcept { ASMap = std::move(Map); }
```
MrSidims: nit: Not sure if std::move does anything useful here, but idiomatically fine.
`void setAddrSpaceMap(AddrSpaceMap &Map) noexcept { ASMap = Map; }` is also fine

MrSidims (commented):
Overall LGTM, re-writing datalayout is a strong request from me, other comments are nits.

include/LLVMSPIRVOpts.h
mbelicki: Yes, it's indeed a typo, fixed in a295973.

lib/SPIRV/SPIRVReader.cpp
mbelicki: I've fixed this in a295973. Additionally I've appended -P to DL string for consistency, I've considered -G too but unfortunately the reader right now depends on -G being zero.

test/spirv-addrspace-map-builtin-mangle.ll
mbelicki: Added a new test specifically for that in a295973

tools/llvm-spirv/llvm-spirv.cpp
mbelicki: Done in a295973

include/LLVMSPIRVOpts.h
mbelicki: I doesn't matter for array, I removed it in a295973 to not confuse readers.

tools/llvm-spirv/llvm-spirv.cpp
mbelicki: I've added an additional sentence to the description to clarify that it's not supported by the writer (8fcbe3b). I'm not sure if this what you meant by "flagging it"? I agree that it would be great to have the whole mapping mechanism work both ways but the scope of this PR is already bigger than I expected when I planned it out initially ;)

MrSidims: I mean: if you are not planning to add such thing in this PR or near future - say it and potentially I'll handle this :)

mbelicki: I don't plan to add the changes in the writer :) Thanks.

mbelicki:
@MrSidims @vmaksimo I would be grateful for another review or approval if everything is ok right now :)

lib/SPIRV/SPIRVReader.cpp
```diff
@@ -1896,14 +1895,12 @@ Value *SPIRVToLLVM::transValueWithoutDecoration(SPIRVValue *BV, Function *F,
     IRBuilder<> Builder(BB);
     auto *Restore = static_cast<SPIRVRestoreMemoryINTEL *>(BV);
     llvm::Value *Ptr = transValue(Restore->getOperand(0), F, BB);
-    auto *StackRestore = Builder.CreateStackRestore(Ptr);
-    return mapValue(BV, StackRestore);
+    return mapValue(BV, Builder.CreateStackRestore(Ptr));
```
MrSidims: nit: lets move away such changes out of a scope of this PR.

lib/SPIRV/SPIRVReader.cpp
```diff
@@ -369,16 +369,18 @@ Type *SPIRVToLLVM::transType(SPIRVType *T, bool UseTPT) {
     if (BM->shouldEmitFunctionPtrAddrSpace() &&
         T->getPointerElementType()->getOpCode() == OpTypeFunction)
       AS = SPIRAS_CodeSectionINTEL;
+    unsigned MappedAS = BM->mapAddrSpace(AS);
```
MrSidims, line 372: This feels a bit awkward: we first get AS from looking at a `SPIRSPIRVAddrSpaceMap` and then do re-mapping. It should be merged into a single utility. But it's probably a jab on me as I've already introduced such discrepancy with special handling of `CodeSectionINTEL`.

MrSidims (approved):
Thanks!

lib/SPIRV/SPIRVReader.cpp
mbelicki: Fixed. Sorry, after removing mapping from this sites I didn't check if the code remain exactly the same as before my edits.

vmaksimo (approved):

lib/SPIRV/SPIRVReader.cpp
```diff
@@ -3150,10 +3151,13 @@ Value *SPIRVToLLVM::transFixedPointInst(SPIRVInstruction *BI, BasicBlock *BB) {
...
     llvm::PointerType *RetPtrTy =
-        llvm::PointerType::get(*Context, SPIRAS_Generic);
+        llvm::PointerType::get(*Context, BM->mapAddrSpace(SPIRAS_Generic));
     Value *Alloca =
         new AllocaInst(RetTy, M->getDataLayout().getAllocaAddrSpace(), "", BB);
-    Value *RetValPtr = new AddrSpaceCastInst(Alloca, RetPtrTy, "", BB);
+    Value *RetValPtr = (Alloca->getType() == RetPtrTy)
+                           ? Alloca
+                           : static_cast<Value *>(new AddrSpaceCastInst(
+                                 Alloca, RetPtrTy, "", BB));
```
```suggestion
                           : new AddrSpaceCastInst(Alloca, RetPtrTy, "", BB);
```

lib/SPIRV/SPIRVReader.cpp
```diff
@@ -3275,11 +3279,14 @@ Value *SPIRVToLLVM::transArbFloatInst(SPIRVInstruction *BI, BasicBlock *BB,
...
-        llvm::PointerType::get(*Context, SPIRAS_Generic);
+        llvm::PointerType::get(*Context, BM->mapAddrSpace(SPIRAS_Generic));
     ArgTys.push_back(RetPtrTy);
     Value *Alloca =
         new AllocaInst(RetTy, M->getDataLayout().getAllocaAddrSpace(), "", BB);
-    Value *RetValPtr = new AddrSpaceCastInst(Alloca, RetPtrTy, "", BB);
+    Value *RetValPtr = (Alloca->getType() == RetPtrTy)
+                           ? Alloca
+                           : static_cast<Value *>(new AddrSpaceCastInst(
+                                 Alloca, RetPtrTy, "", BB));
```
```suggestion
                           : new AddrSpaceCastInst(Alloca, RetPtrTy, "", BB);
```

vmaksimo (commented):
Last small points for improvement (non-blocking)

svenvh (approved):

PR #3718, opened 2026-04-29
Fix DebugTypeFunction void return type

vmaksimo:
All DebugInfo specs (core SPIR-V, OpenCL.DebugInfo.100, NonSemantic.Shader.DebugInfo.*) require the ReturnType operand of DebugTypeFunction to reference a valid type.
For void-returning functions the translator was emitting DebugInfoNone instead of VoidTy.

Also fix the same issue in existing reverse-translation test inputs that were generated by the Writer with the bug in place.

`spirv-val` on the NonSemantic path cannot be enabled yet due to other spec-compliance issues to be fixed in subsequent changes.

### DIFF
--- a/test/DebugInfo/DebugInfoTypeFunctionVoidReturn.ll
+++ b/test/DebugInfo/DebugInfoTypeFunctionVoidReturn.ll
@@ -1,0 +1,32 @@
+; Test that DebugTypeFunction uses VoidTy (not DebugInfoNone) as the return
+; type for void-returning functions, per all DebugInfo specs.
+
+; RUN: llvm-as %s -o %t.bc
+; RUN: llvm-spirv %t.bc -o %t.spv
+; RUN: spirv-val %t.spv
+; RUN: llvm-spirv %t.bc -spirv-text -o - | FileCheck %s --check-prefix=CHECK-SPIRV
+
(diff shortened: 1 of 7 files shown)


test/DebugInfo/DebugInfoTypeFunctionVoidReturn.ll
svenvh, line 1: This is already covered by other tests, so not sure if we really need a new test?

svenvh (commented):
Thanks!

test/DebugInfo/DebugInfoTypeFunctionVoidReturn.ll
vmaksimo: For now this very bug is covered only by `DebugInfo/NonSemantic/Shader200/InlineNamespace.ll` so I thought it'd be more clear to have a separate one. But I don't mind removing it 🙂

MrSidims (approved):

test/DebugInfo/DebugInfoTypeFunctionVoidReturn.ll
svenvh: Ah of course, the other tests are all inputs, not outputs; makes sense to keep it then!

svenvh (approved):

PR #3719, opened 2026-04-29
Translate llvm.trap and llvm.ubsantrap to OpAbortKHR

vmustya:
When the SPV_KHR_abort is enabled, the `llvm.trap` and `llvm.ubsantrap`
intrinsics are translated to `OpAbortKHR` instructions. The `llvm.trap`
intrinsic is translated with a reserved all-ones message value, while
the `llvm.ubsantrap` intrinsic propagates its i8 failure kind argument
zero-extended to 32 bits as the message value.

The reverse translation doesn't introduce the `llvm.trap` or
`llvm.ubsantrap` intrinsics back into the LLVM IR, as there is no way to
distinguish them from other uses of `OpAbortKHR`. So, the reader
preserves them as calls to the `__spirv_AbortKHR` functions.

The `llvm.debugtrap` intrinsic is still ignored, because it's not
an abort semantically.

Assisted-by: Claude Opus 4.7 <noreply@anthropic.com>

### DIFF
--- a/lib/SPIRV/SPIRVWriter.cpp
+++ b/lib/SPIRV/SPIRVWriter.cpp
@@ -5136,17 +5136,34 @@ SPIRVValue *LLVMToSPIRVBase::transIntrinsicInst(IntrinsicInst *II,
         "-spirv-allow-unknown-intrinsics option.");
     break;
   }
+  case Intrinsic::trap:
+  case Intrinsic::ubsantrap: {
+    // When the SPV_KHR_abort extension is disabled, ignore/drop the llvm.trap
+    // and llvm.ubsantrap intrinsics to not break the translator.
+    if (!BM->isAllowedToUseExtension(ExtensionID::SPV_KHR_abort))
+      return nullptr;
+
+    // For llvm.trap use the constant 32-bit "all ones" value, because there is
+    // no message present in the intrinsic.
+    uint64_t MsgVal = ~0u;
+    if (IID == Intrinsic::ubsantrap) {
+      // For llvm.ubsan intrinsic propagate the operand value.
+      auto *Arg = cast<ConstantInt>(II->getArgOperand(0));
+      MsgVal = Arg->getZExtValue();
+    }
+
+    auto *I32Ty = transType(Type::getInt32Ty(II->getContext()));
+    auto *Msg = BM->addConstant(I32Ty, MsgVal);
+    return BM->addAbortKHRInst(Msg, BB);
+  }
   // We can just ignore/drop some intrinsics, like optimizations hint.
   case Intrinsic::experimental_noalias_scope_decl:
   case Intrinsic::invariant_start:
   case Intrinsic::invariant_end:
   case Intrinsic::dbg_label:
-  // TODO: lower llvm.trap / llvm.ubsantrap / llvm.debugtrap to OpAbortKHR
-  // (with a null/undefined Message) when the SPV_KHR_abort extension is
-  // enabled. For now we silently drop these intrinsics so that the translator
-  // doesn't crash on input that contains them.
-  case Intrinsic::trap:
-  case Intrinsic::ubsantrap:
+  // llvm.debugtrap is silently dropped: it has no direct SPIR-V counterpart
+  // and cannot be lowered into the OpAbortKHR, because it doesn't terminate
+  // the execution.
   case Intrinsic::debugtrap:
   // Just ignore llvm.fake.use intrinsic, as it has no translation in SPIRV.
   case Intrinsic::fake_use:
--- a/test/extensions/KHR/SPV_KHR_abort/debugtrap-not-translated.ll
+++ b/test/extensions/KHR/SPV_KHR_abort/debugtrap-not-translated.ll
@@ -1,0 +1,27 @@
+; Negative test: llvm.debugtrap should NOT produce OpAbortKHR.
+; Only llvm.trap and llvm.ubsantrap are translated to OpAbortKHR. The
+; debugtrap variant is currently dropped (return nullptr). This test ensures
+; it doesn't accidentally trigger the abort path.
+
+; RUN: llvm-as %s -o %t.bc
+; RUN: llvm-spirv %t.bc --spirv-ext=+SPV_KHR_abort -spirv-text -o %t.spt
+; RUN: FileCheck < %t.spt %s --check-prefix=CHECK-SPIRV --implicit-check-not AbortKHR
+
+; ---- debugtrap: no AbortKHR, unreachable still present ----
+; CHECK-SPIRV: Function
+; CHECK-SPIRV: Label
+; CHECK-SPIRV: Unreachable
+; CHECK-SPIRV: FunctionEnd
+
+target datalayout = "e-i64:64-v16:16-v24:32-v32:32-v48:64-v96:128-v192:256-v256:256-v512:512-v1024:1024"
+target triple = "spir64-unknown-unknown"
+
+define spir_func void @uses_debugtrap() {
+entry:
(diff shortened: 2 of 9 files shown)


vmustya:
@vmaksimo, @michalpaszkowski, @maarquitos14, @svenvh, this is a follow-up PR for the #3691. Could you please review? Thanks in advance!

maarquitos14 (approved):
The changes LGTM. However, I'm a bit concerned about the reverse translation always going back to spirv-friendly IR. I understand the reasoning, and it makes sense. I just want to bring up a discussion of what happens with consumers that don't accept spirv-friendly IR. Could we add a flag to translate back to llvm.trap/llvm.ubsantrap? Maybe the user can pass in the flag what intrinsic to use for OpAbortKHR in the reverse translation. Open to other ideas/opinions, like I said, my goal is to open the discussion.

vmaksimo (commented):
Can't we translate `AbortKHR` back to llvm intrinsic when SPIR-V friendly IR is not enabled? How bad is this idea?

lib/SPIRV/SPIRVWriter.cpp
```diff
@@ -5136,17 +5136,34 @@ SPIRVValue *LLVMToSPIRVBase::transIntrinsicInst(IntrinsicInst *II,
         "-spirv-allow-unknown-intrinsics option.");
     break;
   }
+  case Intrinsic::trap:
+  case Intrinsic::ubsantrap: {
+    // When the SPV_KHR_abort extension is disabled, ignore/drop the llvm.trap
+    // and llvm.ubsantrap intrinsics to not break the translator.
+    if (!BM->isAllowedToUseExtension(ExtensionID::SPV_KHR_abort))
+      return nullptr;
```
MrSidims, line 5144: There are plenty of places where translation is aborted like this if some extension is not enabled, but I can't stop myself but wonder, if we should be actually generating `OpNop` instead. Just thinking out loud, not a call for action.

test/extensions/KHR/SPV_KHR_abort/debugtrap-not-translated.ll
```diff
@@ -0,0 +1,27 @@
+; Negative test: llvm.debugtrap should NOT produce OpAbortKHR.
+; Only llvm.trap and llvm.ubsantrap are translated to OpAbortKHR. The
+; debugtrap variant is currently dropped (return nullptr). This test ensures
+; it doesn't accidentally trigger the abort path.
+
+; RUN: llvm-as %s -o %t.bc
+; RUN: llvm-spirv %t.bc --spirv-ext=+SPV_KHR_abort -spirv-text -o %t.spt
```
MrSidims, line 7:
```suggestion

; RUN: llvm-spirv %s --spirv-ext=+SPV_KHR_abort -spirv-text -o %t.spt
```
same comment is applicable to other tests

**UPD**: %t.ll ->%s

test/extensions/KHR/SPV_KHR_abort/debugtrap-not-translated.ll
```diff
@@ -0,0 +1,27 @@
+; Negative test: llvm.debugtrap should NOT produce OpAbortKHR.
```
MrSidims, line 1: Not sure if we need this test. We already have the negative case for the intrinsic covered by https://github.com/KhronosGroup/SPIRV-LLVM-Translator/blob/8f759502ce0d6faf533e31cf313aa085415d934c/test/llvm-intrinsics/debugtrap.ll#L14

Same comment for some other negative tests below.

test/extensions/KHR/SPV_KHR_abort/trap-basic.ll
```diff
@@ -0,0 +1,62 @@
...
+; RUN: llvm-dis < %t.rev.bc | FileCheck %s --check-prefix=CHECK-LLVM
+
+; Round-trip with explicit SPV-IR target environment
+; RUN: llvm-spirv -r %t.spv -o %t.rev.spv-ir.bc --spirv-target-env=SPV-IR
+; RUN: llvm-dis < %t.rev.spv-ir.bc | FileCheck %s --check-prefix=CHECK-LLVM
+
+; FIXME: enable the following run when the translator CI is updated to a new
+; version of the SPIR-V Tools that includes the support for the SPV_KHR_abort
+; extension.
+; RUN: not spirv-val %t.spv
```
MrSidims, line 23:
```suggestion
; RUNx:  spirv-val %t.spv
```

test/extensions/KHR/SPV_KHR_abort/trap-ext-disabled.ll
```diff
@@ -0,0 +1,35 @@
...
+; RUN: llvm-spirv %t.bc --spirv-ext=+all,-SPV_KHR_abort -spirv-text -o %t.spt
+; RUN: FileCheck < %t.spt %s --check-prefix=CHECK-DISABLED
+
+; Explicit disable
+; RUN: llvm-spirv %t.bc --spirv-ext=-SPV_KHR_abort -spirv-text -o %t.spt2
+; RUN: FileCheck < %t.spt2 %s --check-prefix=CHECK-DISABLED
+
+; Verify both produce valid SPIR-V
+; RUN: llvm-spirv %t.bc --spirv-ext=+all,-SPV_KHR_abort -o %t.spv
+; RUN: spirv-val %t.spv
```
MrSidims, line 17: Is it useful/not covered by other existing tests?

test/extensions/KHR/SPV_KHR_abort/trap-in-kernel.ll
```diff
@@ -0,0 +1,90 @@
+; Kernel function with assert-like trap pattern.
+; Models the real-world HIP assert() use case:
+;   kernel calls __assert_fail which calls llvm.trap.
+;
+; Verifies that OpAbortKHR works correctly inside callee functions
+; invoked from kernel entry points.
+;
+; Note: The kernel's assert.fail block has FunctionCall + Unreachable
+; (the unreachable is after the call, not after a trap — this is correct
```
MrSidims, line 9: lets not use non-ascii symbols.
```suggestion
; (the unreachable is after the call, not after a trap - this is correct
```

test/extensions/KHR/SPV_KHR_abort/trap-in-kernel.ll
```diff
@@ -0,0 +1,90 @@
...
+; __assert_fail_internal: contains the trap → OpAbortKHR
+; CHECK-SPIRV: Function
+; CHECK-SPIRV: AbortKHR
+; CHECK-SPIRV: FunctionEnd
+
+; test_kernel: calls __assert_fail_internal, then unreachable (in caller)
+; CHECK-SPIRV: Function {{.*}} [[#KernelId]]
+; CHECK-SPIRV: BranchConditional
+; CHECK-SPIRV: Return
+; CHECK-SPIRV: FunctionCall
```
MrSidims, line 40: The comment above convinced me, that this FunctionCall is important. If so - lets check, what function is being called.

test/extensions/KHR/SPV_KHR_abort/trap-basic.ll
MrSidims: Same for other tests.

maarquitos14:
> Can't we translate `AbortKHR` back to llvm intrinsic when SPIR-V friendly IR is not enabled? How bad is this idea?

The one problem I see with that is that I'm not sure we can easily discern if we need to go back to llvm.trap or llvm.ubsantrap. Maybe we can check the message value for that? As of now, they are hardcoded, IIRC.

maarquitos14:
@vmustya can we reignite this?

PR #3720, opened 2026-04-30
SPIRVReader: Add spirv.nontemporal for Nontemporal image op

svenvh:
Add `spirv.nontemporal` metadata for image operations that have a `Nontemporal` image operand, such that the nontemporal information is preserved in the LLVM IR.

### DIFF
--- a/lib/SPIRV/SPIRVReader.cpp
+++ b/lib/SPIRV/SPIRVReader.cpp
@@ -3847,6 +3847,18 @@ Instruction *SPIRVToLLVM::transBuiltinFromInst(const std::string &FuncName,
   } else {
     Call = CallInst::Create(Func, transValue(Ops, BB->getParent(), BB), "", BB);
   }
+  auto HasImageOperandNonTemporal = [](SPIRVInstruction *Inst) -> bool {
+    size_t Idx = getImageOperandsIndex(Inst->getOpCode());
+    if (Idx == ~0U)
+      return false;
+    auto Ops = Inst->getOperands();
+    if (Ops.size() <= Idx)
+      return false;
+    auto ImOp = static_cast<SPIRVConstant *>(Ops[Idx])->getZExtIntValue();
+    return ImOp & ImageOperandsMask::ImageOperandsNontemporalMask;
+  };
+  if (HasImageOperandNonTemporal(BI))
+    transNonTemporalMetadata(Call);
   setName(Call, BI);
   setAttrByCalledFunc(Call);
   SPIRVDBG(spvdbgs() << "[transInstToBuiltinCall] " << *BI << " -> ";
@@ -4765,7 +4777,10 @@ SPIRVToLLVM::transOCLImageTypeAccessQualifier(SPIRV::SPIRVTypeImage *ST) {
 bool SPIRVToLLVM::transNonTemporalMetadata(Instruction *I) {
   Constant *One = ConstantInt::get(Type::getInt32Ty(*Context), 1);
   MDNode *Node = MDNode::get(*Context, ConstantAsMetadata::get(One));
-  I->setMetadata(M->getMDKindID("nontemporal"), Node);
+  if (isa<LoadInst>(I) || isa<StoreInst>(I))
+    I->setMetadata(LLVMContext::MD_nontemporal, Node);
+  else
+    I->setMetadata("spirv.nontemporal", Node);
   return true;
 }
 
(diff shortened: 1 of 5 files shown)


test/image_operand_nontemporal.spvasm
vmaksimo, line 1: Hm, we're not checking for this anywhere in the translator's code.. is it okay to leave this validation work to the spirv-tools?

lib/SPIRV/SPIRVReader.cpp
```diff
@@ -3847,6 +3847,18 @@ Instruction *SPIRVToLLVM::transBuiltinFromInst(const std::string &FuncName,
...
+    size_t Idx = getImageOperandsIndex(Inst->getOpCode());
+    if (Idx == ~0U)
+      return false;
+    auto Ops = Inst->getOperands();
+    if (Ops.size() <= Idx)
+      return false;
+    auto ImOp = static_cast<SPIRVConstant *>(Ops[Idx])->getZExtIntValue();
+    return ImOp & ImageOperandsMask::ImageOperandsNontemporalMask;
+  };
+  if (HasImageOperandNonTemporal(BI))
```
vmaksimo: Won't it be "easier" to inherit `SPIRVImageInstBase` from `SPIRVMemoryAccess` as it's done for load/store?
If it makes any sense, this block would fit to simple `if (BI->SPIRVMemoryAccess::isNonTemporal())`

vmaksimo (commented):
Interesting! Any plans for contributing the matching Writer part (if such IR will be generated by any compiler frontend)?

test/image_operand_nontemporal.spvasm
svenvh: I think so; if a user wants full validation, it's better to use SPIR-V tools before invoking llvm-spirv.

lib/SPIRV/SPIRVReader.cpp
svenvh: `SPIRVMemoryAccess::isNonTemporal` handles Memory Operands, and this is for Image Operands, so they're not the same.  But actually I'll see if it makes more sense to move this into a proper member function of `SPIRVImageInstBase`.

svenvh:
> Any plans for contributing the matching Writer part (if such IR will be generated by any compiler frontend)?

Not aware of any frontend emitting this yet, but I can do the Writer part in a followup PR if there's interest.

vmaksimo (approved):

lib/SPIRV/SPIRVReader.cpp
```diff
@@ -4765,7 +4769,10 @@ SPIRVToLLVM::transOCLImageTypeAccessQualifier(SPIRV::SPIRVTypeImage *ST) {
 bool SPIRVToLLVM::transNonTemporalMetadata(Instruction *I) {
   Constant *One = ConstantInt::get(Type::getInt32Ty(*Context), 1);
   MDNode *Node = MDNode::get(*Context, ConstantAsMetadata::get(One));
-  I->setMetadata(M->getMDKindID("nontemporal"), Node);
+  if (isa<LoadInst>(I) || isa<StoreInst>(I))
+    I->setMetadata(LLVMContext::MD_nontemporal, Node);
+  else
+    I->setMetadata("spirv.nontemporal", Node);
```
MrSidims, line 4775: What if we remove this line, does LLVM verifier flags an error or metadata is not emitted? Techinicaly function call can be a memory accessing instruction. Looking at LangRef I don't see restrictions to where apply this metadata. Looking at the tests I do see that this metadata can be applied to intrinsic calls, for example:
`CodeGen/RISCV/rvv/nontemporal-vp-scalable.ll:14809:  %x = call <vscale x 4 x i32> @llvm.experimental.vp.strided.load.nxv4i32.i64(ptr %p, i64 %stride, <vscale x 4 x i1> splat(i1 true), i32 %vl), !nontemporal !0`

MrSidims: https://godbolt.org/z/q7ohr6K3d doesn't error out at least

svenvh: The IR Verifier currently doesn't complain about `!nontemporal` on non-Load/Store instructions indeed, but I'm hesitant to rely on that.  The LangRef only mentions `!nontemporal` metadata for Load/Store instructions so using it on builtin calls would give it non-standard meaning.  The use of `!nontemporal` on intrinsic calls seems sporadic, and I couldn't find any precedents of `!nontemporal` on ordinary non-intrinsic calls.

So I went for custom metadata, but I'm open to changing to `!nontemporal` if there's a strong enough reason.

MrSidims: My only reason is Occam's razor :) But I understand your hesitation, so I don't insist. The patch is LGTM.

PR #3722, opened 2026-05-02
Lower llvm.ldexp via OpenCL ldexp ext-inst

pvelesko:
## Summary

clang emits `Intrinsic::ldexp` for `ldexp()`, `scalbn()`, `scalbln()`, and integer-exponent `exp2()` in device code, but `transIntrinsicInst()` had no case for it, so translation aborted at `SPIRVWriter.cpp` with:

```
InvalidFunctionCall: Unexpected llvm intrinsic:
 llvm.ldexp.f64.i32
```

This blocks compilation of any TU using math functions that lower to `ldexp`, including Kokkos tests for `ldexp`, `scalbn`, `scalbln`, `frexp`, and `exp2`.

## Change

Three small inserts in `lib/SPIRV/SPIRVWriter.cpp`:

1. Add `Intrinsic::ldexp` to `isKnownIntrinsic()` so its declaration isn't emitted as a SPIR-V function declaration.
2. Map `Intrinsic::ldexp` to `OpenCLLIB::Ldexp` (opcode 34) in `getBuiltinIdForIntrinsic()`.
3. Add `Intrinsic::ldexp` to the binary-FP intrinsics dispatch case -- the same path used by `Intrinsic::powi`, which has the same `(gentype, int)` signature.

`f16`, `f32`, `f64` scalar and vector forms are all handled because the type is derived from `II->getType()`.

New lit test `test/llvm-intrinsics/ldexp.ll` exercises all three scalar widths plus a `<4 x float>` vector form, validates with `spirv-val`. The existing `test/transcoding/ldexp.cl` only covered the OpenCL-mangled builtin path, which doesn't go through `Intrinsic::ldexp`.

Originally observed compiling Kokkos device code through chipStar.

### DIFF
--- a/test/llvm-intrinsics/ldexp.ll
+++ b/test/llvm-intrinsics/ldexp.ll
@@ -1,0 +1,40 @@
+; RUN: llvm-as < %s -o %t.bc
+; RUN: llvm-spirv %t.bc -o %t.spv
+; RUN: llvm-spirv %t.spv -to-text -o - | FileCheck %s
+; RUN: spirv-val %t.spv
+
+; CHECK: ExtInstImport [[ExtInstSetId:[0-9]+]] "OpenCL.std"
+; CHECK-DAG: TypeFloat [[Half:[0-9]+]] 16
+; CHECK-DAG: TypeFloat [[Float:[0-9]+]] 32
+; CHECK-DAG: TypeFloat [[Double:[0-9]+]] 64
+; CHECK-DAG: TypeInt [[Int32:[0-9]+]] 32 0
+; CHECK-DAG: TypeVector [[Float4:[0-9]+]] [[Float]] 4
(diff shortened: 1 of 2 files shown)


test/llvm-intrinsics/ldexp.ll
```diff
@@ -0,0 +1,40 @@
+; RUN: llvm-as < %s -o %t.bc
+; RUN: llvm-spirv %t.bc -o %t.spv
```
MrSidims:
```suggestion
; RUN: llvm-spirv %s -o %t.spv
```

MrSidims (approved):

pvelesko:
Closing this as it was independently fixed upstream in #3608 (merged 2026-03-02). Our additions produce duplicate case labels against the current main, which is what the CI is reporting.

PR #3735, opened 2026-05-05
Disable spirv-val for SPV_KHR_abort tests

svenvh:
The current approach with `not` causes the spirv-val step to fail on recent enough SPIRV-Tools versions.  Avoid running spirv-val at all until a new SPIRV-Tools release is more widely available.

### DIFF
--- a/test/extensions/KHR/SPV_KHR_abort/abort-conditional.ll
+++ b/test/extensions/KHR/SPV_KHR_abort/abort-conditional.ll
@@ -16,7 +16,7 @@
 ; FIXME: enable the following run when the translator CI is updated to a new
 ; version of the SPIR-V Tools that includes the support for the SPV_KHR_abort
 ; extension.
-; RUN: not spirv-val %t.spv
+; TODO: RUNx: spirv-val %t.spv
 
 ; ---- SPIR-V with extension ----
 ; CHECK-SPIRV-DAG: Extension "SPV_KHR_abort"
--- a/test/extensions/KHR/SPV_KHR_abort/abort-in-kernel.ll
+++ b/test/extensions/KHR/SPV_KHR_abort/abort-in-kernel.ll
@@ -16,7 +16,7 @@
 ; FIXME: enable the following run when the translator CI is updated to a new
 ; version of the SPIR-V Tools that includes the support for the SPV_KHR_abort
 ; extension.
-; RUN: not spirv-val %t.spv
+; TODO: RUNx: spirv-val %t.spv
 
 ; Round-trip
 ; RUN: llvm-spirv -r %t.spv -o %t.rev.bc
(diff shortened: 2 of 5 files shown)


svenvh:
@vmustya I overlooked this little detail on your patch.  Please also keep this in mind for any backports.

vmaksimo:
@svenvh is the failure observed only locally on the latest SPIRV-Tools? 
I didn't thought `not spirv-val` can cause any problems in CI 🤔

svenvh:
> @svenvh is the failure observed only locally on the latest SPIRV-Tools?

It fails in any environment that uses a recent enough SPIRV-Tools (but not in our GitHub CI yet).

> I didn't thought `not spirv-val` can cause any problems in CI 🤔

It's actually a bit of a ticking time bomb too: when packages.lunarg.com/vulkan releases a new spirv-tools version with SPV_KHR_abort support, all branches with `not spirv-val` will simultaneously start failing these tests in the GitHub CI.  I think it's better to avoid such future surprises. :)

vmaksimo:
@svenvh Oh, I asked to do this on purpose 😅 My motivation is that a ticking time bomb is better because we don't forget to enable the "TODO" checks - it's not easier to track these in mind only. 
And the failing CI just explicitly notifies us 1) that the tools are updated, 2) the checks back should be enabled back (and if the validation still fails - this is a point to disable the RUN line with TODO and a filed issue).
Does this make sense to you?

svenvh:
> Oh, I asked to do this on purpose 😅

Ah I completely missed that on the review! 😅

> My motivation is that a ticking time bomb is better because we don't forget to enable the "TODO" checks - it's not easier to track these in mind only.

I see the benefit, it's indeed very easy to forget about enabling spirv-val at a later point.  Unfortunately I think the downsides outweigh that benefit: "sudden" CI failures aren't ideal, it puts the burden on maintainers to find out what's going on and fix it.  But most of all this makes it hard to work with recent spirv-tools (which isn't unreasonable as the latest spirv-tools release is already 1 year old).  Hence I'd prefer sticking to the current convention of disabling spirv-val.

Longer term, if we want to change the convention, perhaps we should consider moving the GitHub CI away from spirv-tools releases and build spirv-tools directly from the `main` branch?

vmaksimo:
>  perhaps we should consider moving the GitHub CI away from spirv-tools releases and build spirv-tools directly from the main branch?

@svenvh I like the idea! Let's keep the current convention and may be switch to building SPIRV-Tools in future

vmaksimo (approved):

svenvh:
> may be switch to building SPIRV-Tools in future

I've opened https://github.com/KhronosGroup/SPIRV-LLVM-Translator/issues/3736 so we don't forget.

PR #3737, opened 2026-05-06
[LLVM 16] ESIMD W/A Feature Check Guard

bokrzesi:
The original workaround (#3476) applied unconditionally, accepting invalid SPIR-V (`OpConvertPtrToU` applied to target extension types) in all contexts.

This PR gates that W/A on `SPV_INTEL_vector_compute` - to limit acceptance of spec-violating SPIR-V to the ESIMD use case that actually needs it.


vmaksimo (approved):
Thanks for minimizing the cases where we accept the invalid SPIR-V! 
Let's do the formatting, and maybe also propose this to the main branch

PR #3734, opened 2026-05-04
[DOC] Add NonSemantic AuxData set specification

MrSidims:
This document describes already existing mechanism to preserve some of LLVM constructs in
SPIR-V during the translation without a formal extension. It is referencing some public LLVM
APIs (which is unusual for formal SPIR-V specification), but quite useful to capture what llvm-spirv
does.

On top of existing functionality added a way to preserve available_externally after the round trip.

Capturing current implementation notes is done with a help from Claude Code Opus 4.7.

PR #3743, opened 2026-05-08
Add NonSemanticAuxData::Linkage translation

MrSidims:
This follows doc added in
https://github.com/KhronosGroup/SPIRV-LLVM-Translator/pull/3734

Now only AvailableExternally is translated

### DIFF
--- a/lib/SPIRV/SPIRVReader.cpp
+++ b/lib/SPIRV/SPIRVReader.cpp
@@ -5477,17 +5477,17 @@ void SPIRVToLLVM::transAuxDataInst(SPIRVExtInst *BC) {
   if (!BC->getModule()->preserveAuxData())
     return;
   auto Args = BC->getArguments();
-  // Args 0 and 1 are common between attributes and metadata.
-  // 0 is the global object, 1 is the name of the attribute/metadata as a string
+  // Arg 0 is common to all instructions in this set: it identifies the
+  // global object the auxiliary data is attached to.
   auto *Arg0 = BC->getModule()->getValue(Args[0]);
   auto *GO = cast<GlobalObject>(getTranslatedValue(Arg0));
   auto *F = dyn_cast<Function>(GO);
   auto *GV = dyn_cast<GlobalVariable>(GO);
   assert((F || GV) && "Value should already have been translated!");
-  auto AttrOrMDName = BC->getModule()->get<SPIRVString>(Args[1])->getStr();
   switch (BC->getExtOp()) {
   case NonSemanticAuxData::FunctionAttribute:
   case NonSemanticAuxData::GlobalVariableAttribute: {
+    auto AttrOrMDName = BC->getModule()->get<SPIRVString>(Args[1])->getStr();
     assert(Args.size() < 4 && "Unexpected FunctionAttribute Args");
     // If this attr was specially handled and added elsewhere, skip it.
     Attribute::AttrKind AsKind = Attribute::getAttrKindFromName(AttrOrMDName);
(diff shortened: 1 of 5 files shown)


lib/SPIRV/SPIRVReader.cpp
```diff
@@ -5566,6 +5545,18 @@ void SPIRVToLLVM::transAuxDataInst(SPIRVExtInst *BC) {
...
     break;
   }
+  case NonSemanticAuxData::Linkage: {
+    auto *LinkageConst =
+        static_cast<SPIRVConstant *>(BC->getModule()->get<SPIRVValue>(Args[1]));
+    switch (LinkageConst->getZExtIntValue()) {
+    case NonSemanticAuxData::AvailableExternally:
+      GO->setLinkage(GlobalValue::AvailableExternallyLinkage);
+      break;
+    default:
```
maarquitos14, line 5555: Should we assert here? Not error out, but simply assert. There is no other linkage type available so far, so likely there is a bug somewhere if we end up here. The assert would help us caught it.

test/extensions/KHR/SPV_KHR_non_semantic_info/preserve-available-externally-linkage.ll
```diff
@@ -0,0 +1,35 @@
...
+
+; CHECK-SPIRV: ExtInstImport [[#Import:]] "NonSemantic.AuxData"
+; CHECK-SPIRV: Name [[#GV:]] "extern_gv"
+; CHECK-SPIRV: Name [[#Fn:]] "inlinable"
+; CHECK-SPIRV: TypeInt [[#I32T:]] 32 0
+; CHECK-SPIRV-DAG: Constant [[#I32T]] [[#LinkageVal:]] 0
+; CHECK-SPIRV: TypeVoid [[#VoidT:]]
+
+; CHECK-SPIRV-DAG: ExtInst [[#VoidT]] [[#]] [[#Import]] NonSemanticAuxDataLinkage [[#GV]] [[#LinkageVal]] {{$}}
+; CHECK-SPIRV-DAG: ExtInst [[#VoidT]] [[#]] [[#Import]] NonSemanticAuxDataLinkage [[#Fn]] [[#LinkageVal]] {{$}}
```
maarquitos14, line 24: Should we do a CHECK-SPIRV-NOEXT-NOT of these two?

maarquitos14 (commented):
LGTM, just a couple of questions.

lib/SPIRV/SPIRVReader.cpp
MrSidims: No. There will be a follow-up patch that actually makes NonSemantic instructions be really NonSemantic - aka if unknow translator is ensured to just skip processing without crashing or erroring out.

MrSidims: I can think of DEBUG output actually, let me try it. But definitely not an assertion IMHO.

test/extensions/KHR/SPV_KHR_non_semantic_info/preserve-available-externally-linkage.ll
MrSidims: Both the OpExtInstImport "NonSemantic.AuxData" and the ExtInst ... NonSemanticAuxDataLinkage lines are emitted in both runs. The only thing that differs between EXT (max-version 1.5) and NOEXT (default 1.6) is the OpExtension "SPV_KHR_non_semantic_info" declaration  SPV_KHR_non_semantic_info was folded into core SPIR-V 1.6, so the explicit extension declaration goes away,  but the non-semantic ExtInst mechanism itself (and our use of it) is still emitted as before.

maarquitos14: Ah, thanks for explaining!

maarquitos14 (approved):

PR #3744, opened 2026-05-09
Skip unknown nonsemantic instructions and sets

MrSidims:
Do nothing instead of crashing/erroring out.

### DIFF
--- a/lib/SPIRV/SPIRVReader.cpp
+++ b/lib/SPIRV/SPIRVReader.cpp
@@ -5507,6 +5507,7 @@ void SPIRVToLLVM::transAuxDataInst(SPIRVExtInst *BC) {
   case NonSemanticAuxData::GlobalVariableAttribute:
   case NonSemanticAuxData::FunctionMetadata:
   case NonSemanticAuxData::GlobalVariableMetadata:
+  case NonSemanticAuxData::Linkage:
     break;
   default:
     return;
--- a/lib/SPIRV/SPIRVReader.cpp
+++ b/lib/SPIRV/SPIRVReader.cpp
@@ -2777,6 +2777,10 @@ Value *SPIRVToLLVM::transValueWithoutDecoration(SPIRVValue *BV, Function *F,
     case SPIRVEIS_NonSemantic_Shader_DebugInfo_200:
       DbgTran->transDebugIntrinsic(ExtInst, BB);
       return mapValue(BV, nullptr);
+    case SPIRVEIS_NonSemantic_Unknown:
+      // Non-semantic instruction sets unknown to the translator are
+      // silently skipped.
+      return mapValue(BV, nullptr);
     default:
       llvm_unreachable("Unknown extended instruction set!");
     }
@@ -5498,6 +5502,15 @@ void SPIRVToLLVM::transAuxDataInst(SPIRVExtInst *BC) {
   assert(BC->getExtSetKind() == SPIRV::SPIRVEIS_NonSemantic_AuxData);
   if (!BC->getModule()->preserveAuxData())
     return;
+  switch (BC->getExtOp()) {
+  case NonSemanticAuxData::FunctionAttribute:
+  case NonSemanticAuxData::GlobalVariableAttribute:
+  case NonSemanticAuxData::FunctionMetadata:
+  case NonSemanticAuxData::GlobalVariableMetadata:
+    break;
+  default:
+    return;
+  }
   auto Args = BC->getArguments();
   // Arg 0 is common to all instructions in this set: it identifies the
   // global object the auxiliary data is attached to.
@@ -5583,7 +5596,7 @@ void SPIRVToLLVM::transAuxDataInst(SPIRVExtInst *BC) {
     break;
   }
   default:
-    llvm_unreachable("Invalid op");
+    break;
   }
 }
 
--- a/lib/SPIRV/SPIRVToLLVMDbgTran.cpp
+++ b/lib/SPIRV/SPIRVToLLVMDbgTran.cpp
@@ -1635,9 +1635,10 @@ MDNode *SPIRVToLLVMDbgTran::transDebugInstImpl(const SPIRVExtInst *DebugInst) {
     // translator are silently ignored to avoid crashing on modules produced
     // by newer producers. The semantic OpenCL/SPIRV.debug paths still
     // require every opcode to be implemented.
-    if (DebugInst->getExtSetKind() ==
-            SPIRVEIS_NonSemantic_Shader_DebugInfo_100 ||
-        DebugInst->getExtSetKind() == SPIRVEIS_NonSemantic_Shader_DebugInfo_200)
+    // TODO: since introduction of rolling version of NonSemanticDebugInfo, we
+    // must no longer just ignore unknown debug info set, but instead process
+    // instructions, that the translator know.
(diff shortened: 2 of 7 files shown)


lib/SPIRV/SPIRVReader.cpp
```diff
@@ -5498,6 +5502,15 @@ void SPIRVToLLVM::transAuxDataInst(SPIRVExtInst *BC) {
   assert(BC->getExtSetKind() == SPIRV::SPIRVEIS_NonSemantic_AuxData);
   if (!BC->getModule()->preserveAuxData())
     return;
+  switch (BC->getExtOp()) {
+  case NonSemanticAuxData::FunctionAttribute:
+  case NonSemanticAuxData::GlobalVariableAttribute:
+  case NonSemanticAuxData::FunctionMetadata:
+  case NonSemanticAuxData::GlobalVariableMetadata:
+    break;
```
MrSidims, line 5511: to update once https://github.com/KhronosGroup/SPIRV-LLVM-Translator/pull/3743 is merged

maarquitos14 (approved):
LGTM. Just thinking out loud: would it be worth it to have a flag to enable a mode where unknown nonsemantic instructions and sets do crash? It might be useful for debugging.

svenvh (approved):

MrSidims:
> LGTM. Just thinking out loud: would it be worth it to have a flag to enable a mode where unknown nonsemantic instructions and sets do crash? It might be useful for debugging.

It's worth to add such diagnostics, but not in a form of crash - instead it should be a non-interrupting warning (which can be set to be treated as errors). I do have such a long standing idea, but never had enough passion to set it up. Regardless, I'd prefer to do such guarding in a separate PR.

lib/SPIRV/SPIRVReader.cpp
vmaksimo: the time has come?

test/extensions/KHR/SPV_KHR_non_semantic_info/skip-unknown-shader-debuginfo.spvasm
```diff
@@ -0,0 +1,30 @@
...
+       %fnTy = OpTypeFunction %void
+        %nop = OpExtInst %void %dbg 100 %u7
+         %fn = OpFunction %void None %fnTy
+        %ent = OpLabel
+                OpReturn
+                OpFunctionEnd
+
+; CHECK-LABEL: define spir_func void @f()
+; CHECK-NEXT:    ret void
+; CHECK-NOT: NonSemantic.Shader.DebugInfo.300
```
vmaksimo: Do we really produce the LLVM metadata with the used SPIR-V DI if it's a known EIS?

MrSidims: Oops, it's a remnant of a sanity check, that a chain of:
```
spirv-as test.spvasm
spirv-dis test.spv
```
do work without errors when unknown nonsemantic set/instructions are used. Will fix this.

lib/SPIRV/SPIRVToLLVMDbgTran.cpp
```diff
@@ -1631,6 +1631,14 @@ MDNode *SPIRVToLLVMDbgTran::transDebugInstImpl(const SPIRVExtInst *DebugInst) {
     return transTypeArrayDynamic(DebugInst);
 
   default:
+    // Non-semantic shader debug info opcodes that are unknown to this
+    // translator are silently ignored to avoid crashing on modules produced
+    // by newer producers. The semantic OpenCL/SPIRV.debug paths still
+    // require every opcode to be implemented.
+    if (DebugInst->getExtSetKind() ==
+            SPIRVEIS_NonSemantic_Shader_DebugInfo_100 ||
+        DebugInst->getExtSetKind() == SPIRVEIS_NonSemantic_Shader_DebugInfo_200)
```
MrSidims: This check is actually not ideal, I'll add a TODO with description, what should be done instead.

Some context, since https://github.com/KhronosGroup/SPIRV-Registry/pull/384 NonSemanticDebugInfo spec has moved to a rolling version, see https://github.com/KhronosGroup/SPIRV-Registry/blob/main/nonsemantic/NonSemantic.Shader.DebugInfo.asciidoc#version-compatibility . It means, that with new version some producer might generate some new instructions, that the translator is not capable to consumer, but under the same new rolling version - old instructions (that the translator knows, how to consume) are still available. So if [NonSemantic][ShaderDebugInfo][%some version%] set is unknown - translator shouldn't ignore instructions under this set, but instead only filter out the new instructions. Problem is, that I don't know, how to do it in "the most clean" way, as there might be a situation, where new rolling version changes semantic of an instruction, which the translator might not know.

lib/SPIRV/SPIRVReader.cpp
MrSidims: yeap

lib/SPIRV/SPIRVToLLVMDbgTran.cpp
```diff
@@ -1631,6 +1631,15 @@ MDNode *SPIRVToLLVMDbgTran::transDebugInstImpl(const SPIRVExtInst *DebugInst) {
     return transTypeArrayDynamic(DebugInst);
 
   default:
+    // Non-semantic shader debug info opcodes that are unknown to this
+    // translator are silently ignored to avoid crashing on modules produced
+    // by newer producers. The semantic OpenCL/SPIRV.debug paths still
+    // require every opcode to be implemented.
+    // TODO: since introduction of rolling version of NonSemanticDebugInfo, we
+    // must no longer just ignore unknown debug info set, but instead process
+    // instructions, that the translator know.
```
vmaksimo, line 1640: Worth filing an issue? Or do you already have the idea/fix handy as a follow-up change?

vmaksimo (approved):

lib/SPIRV/SPIRVToLLVMDbgTran.cpp
MrSidims: Created https://github.com/KhronosGroup/SPIRV-LLVM-Translator/issues/3749

vmaksimo:
@svenvh @MrSidims we prepared backport PRs, but wanted to double-check that no one disagrees on backporting this change to older release branches. 
If not, may I ask you to explicitly approve #3960? Or share your concerns here.
Thanks in advance!

vmaksimo:
@svenvh @MrSidims friendly ping 😊

svenvh:
Sorry for the delay, no concerns from our side!

MrSidims:
> @svenvh @MrSidims we prepared backport PRs, but wanted to double-check that no one disagrees on backporting this change to older release branches. If not, may I ask you to explicitly approve #3960? Or share your concerns here. Thanks in advance!

Apologies, missed this one (as well as discussion of the appropriate agenda in tooling wg).

I believe once PR is approved and merged on `main` branch - no approval to merge a backport PR is required by default. In cases when backport requires a significant re-implementation or might carry a breaking change (significance of the change is decided `by the engineer that is backporting a patch`) - in this case an extra round of review should be called (but such examples are very rare, I recall only few and most of them were related to porting new types to branches with not target extension types in LLVM).

New stand-alone contributions that submitted to release branches still require review.

PR #3746, opened 2026-05-11
[SPIRV->LLVM] Fix missing spir_func CC on builtin FP4/i4 conversion calls

karolzwolak:
When translating SPIRV FP4/FP8/i4 conversions (`OpFConvert` etc.) back to LLVM IR, the code path that creates `__builtin_spirv_*` calls via `getOrInsertFunction` used the default calling convention.

Fix by setting `CallingConv::SPIR_FUNC` on both the function declaration and the call instruction, consistent with all other builtin call sites in the translator.

### DIFF
--- a/lib/SPIRV/SPIRVReader.cpp
+++ b/lib/SPIRV/SPIRVReader.cpp
@@ -1138,10 +1138,7 @@ Value *SPIRVToLLVM::transConvertInst(SPIRVValue *BV, Function *F,
         if (MangledName.empty())
           MangledName = mangleBuiltin(BuiltinName, OpsTys, &Info);
 
-        FunctionType *FTy = FunctionType::get(Dst, OpsTys, false);
-        FunctionCallee Func = M->getOrInsertFunction(MangledName, FTy);
-        if (auto *F = dyn_cast<Function>(Func.getCallee()))
-          F->setCallingConv(CallingConv::SPIR_FUNC);
+        Function *Func = getOrCreateFunction(M, Dst, OpsTys, MangledName);
         auto *CI = CallInst::Create(Func, Ops, "", BB);
         CI->setCallingConv(CallingConv::SPIR_FUNC);
         return CI;
--- a/lib/SPIRV/SPIRVReader.cpp
+++ b/lib/SPIRV/SPIRVReader.cpp
@@ -1140,7 +1140,11 @@ Value *SPIRVToLLVM::transConvertInst(SPIRVValue *BV, Function *F,
 
         FunctionType *FTy = FunctionType::get(Dst, OpsTys, false);
         FunctionCallee Func = M->getOrInsertFunction(MangledName, FTy);
-        return CallInst::Create(Func, Ops, "", BB);
+        if (auto *F = dyn_cast<Function>(Func.getCallee()))
+          F->setCallingConv(CallingConv::SPIR_FUNC);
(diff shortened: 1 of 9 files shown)


lib/SPIRV/SPIRVReader.cpp
```diff
@@ -1140,7 +1140,11 @@ Value *SPIRVToLLVM::transConvertInst(SPIRVValue *BV, Function *F,
 
         FunctionType *FTy = FunctionType::get(Dst, OpsTys, false);
         FunctionCallee Func = M->getOrInsertFunction(MangledName, FTy);
-        return CallInst::Create(Func, Ops, "", BB);
+        if (auto *F = dyn_cast<Function>(Func.getCallee()))
+          F->setCallingConv(CallingConv::SPIR_FUNC);
+        auto *CI = CallInst::Create(Func, Ops, "", BB);
+        CI->setCallingConv(CallingConv::SPIR_FUNC);
+        return CI;
```
MrSidims: Probably we should use https://github.com/KhronosGroup/SPIRV-LLVM-Translator/blob/main/lib/SPIRV/SPIRVUtil.cpp#L277 instead

karolzwolak: good suggestion, fixed

lib/SPIRV/SPIRVReader.cpp
```diff
@@ -1138,9 +1138,10 @@ Value *SPIRVToLLVM::transConvertInst(SPIRVValue *BV, Function *F,
         if (MangledName.empty())
           MangledName = mangleBuiltin(BuiltinName, OpsTys, &Info);
 
-        FunctionType *FTy = FunctionType::get(Dst, OpsTys, false);
-        FunctionCallee Func = M->getOrInsertFunction(MangledName, FTy);
-        return CallInst::Create(Func, Ops, "", BB);
+        Function *Func = getOrCreateFunction(M, Dst, OpsTys, MangledName);
+        auto *CI = CallInst::Create(Func, Ops, "", BB);
+        CI->setCallingConv(CallingConv::SPIR_FUNC);
```
vmaksimo, line 1143: Isn't `getOrCreateFunction` sets calling convention itself? Or is it somehow not set for the call instances?

MrSidims: Yeap, that was the point of my recommendation :)

karolzwolak: It does set the calling convention the function.
The call instruction has no explicit calling convention, so I'm setting it to `spir_func`.
Honestly I'm not sure it makes any actual difference, but the LLVM IR is different.

karolzwolak: Perhaps I should use this function instead of setting the calling conv manually?
https://github.com/KhronosGroup/SPIRV-LLVM-Translator/blob/ba34cbb62a9cbec90fc446e76faadf702e09b6f6/lib/SPIRV/SPIRVReader.cpp#L255-L259

MrSidims: > It does set the calling convention the function.
> The call instruction has no explicit calling convention

Ah, the infamous LLVM calling convention API discrepancy (they've done it for a reason I'll never understand). No concerns from me then.

MrSidims (approved):

vmaksimo (approved):

PR #3748, opened 2026-05-12
[CI] Build SPIRV-Tools from main instead of lunarg packages

### LINKED ISSUE #3736: Build latest spirv-tools in CI
We're currently using packages.lunarg.com/vulkan releases to get SPIRV-Tools in our CI environment.  Those releases aren't very frequent, which is problematic when developing new features, see e.g. https://github.com/KhronosGroup/SPIRV-LLVM-Translator/pull/3735 .

See if we can instead build the SPIRV-Tools `main` branch in CI.  We'd probably want some kind of build cache to avoid burning many credits on repeatedly building the same thing, or running out of cycles for the longer-running in-tree jobs.

vmaksimo:
Resolves #3736 - replace the lunarg apt package for spirv-tools with a source build of SPIRV-Tools main.
The lunarg releases are infrequent and block CI validation of new extensions before they land in a published package.

A dedicated `prebuild-spirv-tools` workflow owns building and caching the tools.
It runs once a day at 23:00 UTC and saves the install tree under key `spirv-tools-main-YYYY-MM-DD`.
A manual `workflow_dispatch` trigger is provided for urgent refreshes and for bootstrapping the cache on first deployment.

The in-tree and out-of-tree build workflows only restore the cache via a restore-keys prefix match on `spirv-tools-main-`, which always pulls the most recent prebuild entry.
The Actions log for each build job records the exact key restored, so reviewers can see which tools build was used.

### DIFF
--- a/.github/workflows/prebuild-spirv-tools.yml
+++ b/.github/workflows/prebuild-spirv-tools.yml
@@ -1,0 +1,50 @@
+# This workflow builds SPIRV-Tools from main once per day and caches the
+# install tree. The build- and test-workflows only restore from this cache
+# (via restore-keys prefix match), so this workflow owns the freshness of
+# the tools.
+#
+# workflow_dispatch is provided for manual refresh when a SPIRV-Tools fix
+# needs to be picked up before the next scheduled run, or to bootstrap the
+# cache on first deployment.
+name: Prebuild SPIRV-Tools
+
+on:
+  schedule:
+    - cron: 0 23 * * *
+  workflow_dispatch:
+
+env:
+  SPIRV_TOOLS_INSTALL: ${{ github.workspace }}/spirv-tools-install
+
+jobs:
+  prebuild:
+    name: Build and cache SPIRV-Tools main
+    runs-on: ubuntu-24.04
+    steps:
+      - name: Get current date for SPIRV-Tools cache key
+        id: date
+        run: echo "date=$(date -u +%Y-%m-%d)" >> $GITHUB_OUTPUT
+      - name: Check for existing cache entry
+        id: spirv-tools-cache
+        uses: actions/cache/restore@v5
+        with:
+          path: ${{ env.SPIRV_TOOLS_INSTALL }}
+          key: spirv-tools-main-${{ steps.date.outputs.date }}
+          lookup-only: true
+      - name: Build and install SPIRV-Tools
+        if: steps.spirv-tools-cache.outputs.cache-hit != 'true'
+        run: |
+          git clone --depth=1 https://github.com/KhronosGroup/SPIRV-Tools SPIRV-Tools
+          python3 SPIRV-Tools/utils/git-sync-deps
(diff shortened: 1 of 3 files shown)


.github/workflows/prebuild-spirv-tools.yml
```diff
@@ -0,0 +1,50 @@
...
+  SPIRV_TOOLS_INSTALL: ${{ github.workspace }}/spirv-tools-install
+
+jobs:
+  prebuild:
+    name: Build and cache SPIRV-Tools main
+    runs-on: ubuntu-24.04
+    steps:
+      - name: Get current date for SPIRV-Tools cache key
+        id: date
+        run: echo "date=$(date -u +%Y-%m-%d)" >> $GITHUB_OUTPUT
```
svenvh, line 26: Instead of date, I wonder if we should take the SHA-1 of the SPIRV-Tools `main` branch as key instead?

svenvh (commented):
Nice, thanks for looking into this!

.github/workflows/prebuild-spirv-tools.yml
vmaksimo: If you're okay with once-per-day tools build - I believe any key will be good🙂 
`date` is more readable (e.g. easy to follow in logs without searching the hash), `SHA-1` is on other hand exact and precise - we can choose either

svenvh: I guess it doesn't make a difference for the CI, as the regular build-and-test jobs will always pick the latest version anyway.  Getting the SPIRV-Tools SHA-1 might be a bit of a faff, so I think I'm okay going with the date (unless @MrSidims would prefer the SHA-1?).

MrSidims: I don't have strong opinion. Both options are fine. Also tools are not that heavy to build.

svenvh (approved):

MrSidims:
Lets merge it and test on https://github.com/KhronosGroup/SPIRV-LLVM-Translator/pull/3751

MrSidims:
@vmaksimo probably the CI job is not working correctly. I've restarted CI in https://github.com/KhronosGroup/SPIRV-LLVM-Translator/pull/3751 and tests started to pass (previously they were failing with unknown capability). It should be a good sign, but apparently not. spirv-val should have failed with:
```
spirv-val freeze.spv
error: line 23: PoisonKHR must appear in a block
  %9 = OpPoisonKHR %uint
```
(see details in https://github.com/KhronosGroup/SPIRV-Registry/issues/416) but it didn't. Therefore spirv-val RUN stopped working.

vmaksimo:
@MrSidims as I see in the latest merged commit 2569028, it "successfully fails" with the freshest `spirv-val`.

I believe your job was not failing because we should have triggered the first tools build manually (right after the merge) so they would have cached and would be used in your PR. Let's retrigger CI to prove the point

PR #3750, opened 2026-05-14
Fix is_fpclass translation for inverted case

### LINKED ISSUE #3696: is_fpclass might be incorrectly expanded in the translator
See https://github.com/KhronosGroup/SPIRV-LLVM-Translator/pull/1922/changes#r3131239187

MrSidims:
Fixes: https://github.com/KhronosGroup/SPIRV-LLVM-Translator/issues/3696


MrSidims:
@karolzwolak please take a look

karolzwolak (approved):
Awesome, it fixes the bug & simplifies the logic.
I *think* the bug isn't actually present right now. The inversion function returns classes that produce only a single check in the `ResultVec`, so there's no OR combining, and no bug.

maarquitos14 (approved):

PR #3751, opened 2026-05-14
Implement SPV_KHR_poison_freeze extension

MrSidims:
Spec is available here:
https://github.com/KhronosGroup/SPIRV-Registry/blob/main/extensions/KHR/SPV_KHR_poison_freeze.asciidoc


vmaksimo:
retrigger CI with the freshest spirv tools

vmaksimo:
out-of-tree to be fixed by #3753

test/extensions/KHR/SPV_KHR_poison_freeze/poison.ll
```diff
@@ -0,0 +1,55 @@
+; RUN: llvm-spirv %s -o %t.spv --spirv-ext=+SPV_KHR_poison_freeze
+; RUN: llvm-spirv %t.spv -o %t.spt --to-text
+; RUN: FileCheck < %t.spt %s --check-prefix=CHECK-SPIRV
+; RUN: llvm-spirv %t.spv -o %t.rev.bc -r --spirv-target-env=SPV-IR
+; RUN: llvm-dis %t.rev.bc -o %t.rev.ll
+; RUN: FileCheck < %t.rev.ll %s --check-prefix=CHECK-LLVM
+
+; RUN: spirv-val %t.spv
```
vmaksimo: Strange ordering. Move to the second line?

test/extensions/KHR/SPV_KHR_poison_freeze/poison.ll
```diff
@@ -0,0 +1,55 @@
...
+
+; RUN: llvm-spirv %s -o %t.noext.spv
+; RUN: llvm-spirv %t.noext.spv -o %t.noext.spt --to-text
```
vmaksimo:
```suggestion
; RUN: llvm-spirv %s -spirv-text -o %t.noext.spt
```

test/extensions/KHR/SPV_KHR_poison_freeze/poison.ll
```diff
@@ -0,0 +1,55 @@
+; RUN: llvm-spirv %s -o %t.spv --spirv-ext=+SPV_KHR_poison_freeze
+; RUN: llvm-spirv %t.spv -o %t.spt --to-text
+; RUN: FileCheck < %t.spt %s --check-prefix=CHECK-SPIRV
+; RUN: llvm-spirv %t.spv -o %t.rev.bc -r --spirv-target-env=SPV-IR
```
vmaksimo: guess we don't need target-env to be set - the path is the same, right?

test/extensions/KHR/SPV_KHR_poison_freeze/auto-emit-arithmetic-poison.ll
```diff
@@ -0,0 +1,54 @@
...
+
+; Per SPV_KHR_poison_freeze, declaring the PoisonFreezeKHR capability requires
+; every entry point to use the ArithmeticPoisonKHR execution mode.
+
+target datalayout = "e-i64:64-v16:16-v24:32-v32:32-v48:64-v96:128-v192:256-v256:256-v512:512-v1024:1024-n8:16:32:64"
+target triple = "spir64-unknown-unknown"
+
+; CHECK-SPIRV: Capability PoisonFreezeKHR
+; CHECK-SPIRV: Extension "SPV_KHR_poison_freeze"
+; CHECK-SPIRV: EntryPoint {{[0-9]+}} [[#KFreeze:]] "kernel_uses_freeze"
```
vmaksimo:
```suggestion
; CHECK-SPIRV: EntryPoint [[#]] [[#KFreeze:]] "kernel_uses_freeze"
```

vmaksimo (commented):
Patch LGTM, thanks!
I have some styling/test questions ideally to be resolved (comments for `SPV_KHR_poison_freeze/poison.ll` are applicable to other tests as well)

lib/SPIRV/libSPIRV/SPIRVValue.h
```diff
@@ -300,6 +300,28 @@ class SPIRVUndef : public SPIRVConstantEmpty<OpUndef> {
   void validate() const override { SPIRVConstantEmpty::validate(); }
 };
 
+class SPIRVPoisonKHR : public SPIRVConstantEmpty<OpPoisonKHR> {
+public:
+  // Complete constructor
```
maarquitos14:
```suggestion
  // Complete constructor.
```

lib/SPIRV/libSPIRV/SPIRVValue.h
```diff
@@ -300,6 +300,28 @@ class SPIRVUndef : public SPIRVConstantEmpty<OpUndef> {
...
+  SPIRVPoisonKHR(SPIRVModule *M, SPIRVType *TheType, SPIRVId TheId)
+      : SPIRVConstantEmpty(M, TheType, TheId) {
+    validate();
+  }
+  // Incomplete constructor
```
maarquitos14:
```suggestion
  // Incomplete constructor.
```

lib/SPIRV/libSPIRV/SPIRVValue.h
maarquitos14: We could also remove the comment, it doesn't add a lot of value. Up to you.

lib/SPIRV/libSPIRV/SPIRVValue.h
maarquitos14: We could also remove the comment, it doesn't add a lot of value. Up to you.

test/extensions/KHR/SPV_KHR_poison_freeze/auto-emit-arithmetic-poison.ll
```diff
@@ -0,0 +1,54 @@
...
+; CHECK-SPIRV: EntryPoint {{[0-9]+}} [[#KPoison:]] "kernel_uses_poison"
+; CHECK-SPIRV: EntryPoint {{[0-9]+}} [[#KPlain:]] "kernel_plain"
+; ArithmeticPoisonKHR (5157) must be present on every entry point, including
+; the one that does not itself emit any poison/freeze op.
+; CHECK-SPIRV-DAG: ExecutionMode [[#KFreeze]] 5157
+; CHECK-SPIRV-DAG: ExecutionMode [[#KPoison]] 5157
+; CHECK-SPIRV-DAG: ExecutionMode [[#KPlain]] 5157
```
maarquitos14, line 26: Does it make sense to add an extra function that is not entry point and check that it doesn't get the execution mode?

test/extensions/KHR/SPV_KHR_poison_freeze/arithmetic-poison-mode.ll
```diff
@@ -0,0 +1,40 @@
...
+; RUN: llvm-spirv %t.noext.spv -o %t.noext.spt --to-text
+; RUN: FileCheck < %t.noext.spt %s --check-prefix=CHECK-NOEXT
+
+target datalayout = "e-i64:64-v16:16-v24:32-v32:32-v48:64-v96:128-v192:256-v256:256-v512:512-v1024:1024-n8:16:32:64"
+target triple = "spir64-unknown-unknown"
+
+; CHECK-SPIRV: Capability PoisonFreezeKHR
+; CHECK-SPIRV: Extension "SPV_KHR_poison_freeze"
+; CHECK-SPIRV: EntryPoint {{[0-9]+}} [[#Kernel:]] "test_kernel"
+; CHECK-SPIRV: ExecutionMode [[#Kernel]] 5157
```
maarquitos14: Should we also check that it doesn't get it twice? The explicit one + the implicit one because of the capability.

test/extensions/KHR/SPV_KHR_poison_freeze/freeze.ll
```diff
@@ -0,0 +1,41 @@
...
+; RUN: FileCheck < %t.spt %s --check-prefix=CHECK-SPIRV
+; RUN: llvm-spirv %t.spv -o %t.rev.bc -r --spirv-target-env=SPV-IR
+; RUN: llvm-dis %t.rev.bc -o %t.rev.ll
+; RUN: FileCheck < %t.rev.ll %s --check-prefix=CHECK-LLVM
+
+; RUN: spirv-val %t.spv
+
+; RUN: llvm-spirv %s -o %t.noext.spv
+; RUN: llvm-spirv %t.noext.spv -o %t.noext.spt --to-text
+; RUN: FileCheck < %t.noext.spt %s --check-prefix=CHECK-NOEXT
```
maarquitos14, line 11: Should we check reverse translation here too? Just to make sure the freeze is not present.

maarquitos14 (commented):
What about the changes to "Conversion Instructions"? For example:

> Change the following sentence in the descriptions of OpConvertFToU and OpConvertFToS from:
> 
> Behavior is undefined if Result Type is not wide enough to hold the converted value.
> 
> to:
> 
> If Result Type is not wide enough to hold the converted value Result is poison if ArithmeticPoisonKHR is declared; otherwise behavior is undefined.

MrSidims:
> What about the changes to "Conversion Instructions"? For example:

I'm not sure how it should affect lowering. Note: UB != error and result of the conversion we will know only in runtime.

maarquitos14:
> > What about the changes to "Conversion Instructions"? For example:
> 
> I'm not sure how it should affect lowering. Note: UB != error and result of the conversion we will know only in runtime.

If the extension is enabled, and the execution mode is `ArithmeticPoisonKHR`, we should emit poison, if I'm reading it correctly.

MrSidims:
> If the extension is enabled, and the execution mode is `ArithmeticPoisonKHR`, we should emit poison, if I'm reading it correctly.

You read this section correct. But for this example:
```
i32 @foo(float %f) {
  i32 %val = call i32 OpConvertFToU(float %f)
  ret %val
}
```
we will know if overflow happens only at runtime. And it will be compiler's runtime/execution environment duty to handle UB or materialize poisoness value depending on the execution mode. It's never SPIR-V emitter's duty to guess a potentially poisoness value.

lib/SPIRV/libSPIRV/SPIRVValue.h
MrSidims: It's basically a copy-paste from other classes. Removed

MrSidims:
@vmaksimo @maarquitos14 thanks for the suggestions, applied

test/extensions/KHR/SPV_KHR_poison_freeze/auto-emit-arithmetic-poison.ll
maarquitos14: @MrSidims Did you miss this one or simply consider it's not needed?

svenvh:
LGTM

vmaksimo:
What about `OpConvertUToPtr`, `OpBitcast`:

> if the storage class of Result Type does not match, <...> Result is poison if ArithmeticPoisonKHR is declared

and ` OpUDiv, OpSDiv, OpUMod, OpSRem, OpSMod`:

> If Operand 2 is 0, Result is poison if ArithmeticPoisonKHR is declared


At least the first one looks like the thing known before runtime - should we handle this somehow in translator?

test/extensions/KHR/SPV_KHR_poison_freeze/auto-emit-arithmetic-poison.ll
MrSidims: Please note `helper` function in the module.

MrSidims: Ah, okay, I should do a better checking on SPIR-V level aka {{[0-9]+}} -> function id.

MrSidims:
> What about `OpConvertUToPtr`, `OpBitcast`:
> 
> > if the storage class of Result Type does not match, <...> Result is poison if ArithmeticPoisonKHR is declared
> 
> and ` OpUDiv, OpSDiv, OpUMod, OpSRem, OpSMod`:
> 
> > If Operand 2 is 0, Result is poison if ArithmeticPoisonKHR is declared
> 
> At least the first one looks like the thing known before runtime - should we handle this somehow in translator?

I believe we will never have LLVM IR input, where `bitcast` results in a pointer in a different address space than a pointer operand has, do you disagree?

`OpBitcast` and `OpConvertUToPtr` are mentioned in the core spec (see: storage mismatch that is not checked statically is UB) and extension spec (aka UB -> poison) not because of the things, that can be statically checked, but because of the cases when we can't do this: imagine we had a function in one translation unit that was doing something with global memory and that got AS casted to generic pointer. And then we pass these generic pointers to a function call, with this function defined in another translation unit. In this function we could AS cast these pointer operands to, lets say, private. And that is not valid, it's UB disregarding, how we did the cast, via CastGenericToPtr or via Bitcast or via cast_ptr_to_int -> cast_int_to_ptr. That is the reason the rule exists.

Regarding Div and other arithemtic operations. These are all valid questions, including suggestion from Marcos can be turn way around saying that we can fold conversion instructions to poison if constexpr input do not fit in the target type. But I'm strongly against adding any constant folding in llvm-spirv - it goes against our design and also it worsen debuggability. So I would left it to execution environment to handle such special cases.

vmaksimo:
Thanks for another one detailed explanation! I agree with all your takes

vmaksimo (approved):

Issue #3742, opened 2026-05-08
Core SPIR-V feature `SubgroupSize` is translated to non-core attribute `intel_reqd_sub_group_size`


AdamBrouwersHarries:
It appears that the[ `SubgroupSize` execution mode](https://registry.khronos.org/SPIR-V/specs/unified1/SPIRV.html#Execution_Mode) is being translated to the attribute `intel_reqd_sub_group_size`. This is surprising, as `SubgroupSize` is a core part of SPIR-V (since version 1.1), while the `intel_reqd_sub_group_size` annotation is part of a (non-core) [OpenCL vendor extension](https://registry.khronos.org/OpenCL/extensions/intel/cl_intel_required_subgroup_size.html).

Attached are an example SPIR-V assembly kernel, along with the llvm IR that the translator emits by default. The pertinent sections are: 

SPIR-V:
```
               OpEntryPoint Kernel %11 "kernel"
               OpExecutionMode %11 ContractionOff
; CHECK: !intel_reqd_sub_group_size [[METADATA:\![0-9]+]]
; CHECK: [[METADATA]] = !{i32 42} 
               OpExecutionMode %11 SubgroupSize 42
```

LLVM: 
```llvm
; Function Attrs: nounwind
define spir_kernel void @kernel(...) ... !intel_reqd_sub_group_size !15
...
!15 = !{i32 42}
``` 

Is this intended behaviour? I'm aware that a consumer of SPIR-V may not always go through the translator, but for consumers who *do*, it seems odd to translate to a vendor-specific OpenCL extension for a core language feature.

My opinion is that this should be changed so that it translates to a vendor-agnostic attribute, which vendors (or consumers) can translate into internal representations (such as intel's existing attribute), or implement directly.

I'd be keen to hear any background or rationale for the current implementation.

[subgroup_size.spvasm.txt](https://github.com/user-attachments/files/27520175/subgroup_size.spvasm.txt)
[subgroup_size_.spvasm.tmp.ll.txt](https://github.com/user-attachments/files/27520174/subgroup_size_.spvasm.tmp.ll.txt)

MrSidims:
That is not a translation to "vendor-specific attribute" rather than a translation to a named LLVM IR metadata that is happened to have "intel" in its name as it's also produced by clang for the mentioned `intel_reqd_sub_group_size`.

[...]

PR #3764, opened 2026-05-22
Add reqd_sub_group_size metadata for SubgroupSize

svenvh:
Add function metadata without a vendor prefix for the `SubgroupSize` Execution Mode, alongside the existing vendor-prefixed metadata. `SubgroupSize` is a core SPIR-V feature, so representing it using vendor-prefixed metadata was misleading.

Contributes to https://github.com/KhronosGroup/SPIRV-LLVM-Translator/issues/3742


MrSidims (approved):
Please link the corresponding PR in clang.

svenvh:
> Please link the corresponding PR in clang.

There won't be any yet, as we'll need an official OpenCL C extension first.  The situation is a bit odd, but it reflects the current state of the specifications: the OpenCL SPIR-V Environment specification requires supporting `SubgroupSize` because of `SubgroupDispatch` (for SPIR-V >= 1.1), but there is no equivalent in OpenCL C yet.

So for now I just want to decouple `SubgroupSize` translation from any vendor-tagged metadata, to avoid confusion or concerns for consumers of the generated LLVM IR.

lib/SPIRV/SPIRVReader.cpp
```diff
@@ -4937,11 +4937,14 @@ bool SPIRVToLLVM::transMetadata() {
...
       CtorKernels.push_back(F);
     }
-    // Generate metadata for intel_reqd_sub_group_size
+    // Generate metadata for reqd_sub_group_size.
     if (auto *EM = BF->getExecutionMode(ExecutionModeSubgroupSize)) {
       auto *SizeMD =
           ConstantAsMetadata::get(getUInt32(M, EM->getLiterals()[0]));
       F->setMetadata(kSPIR2MD::SubgroupSize, MDNode::get(*Context, SizeMD));
+      // Add vendor specific alias for backward compatibility.
+      F->setMetadata("intel_reqd_sub_group_size",
```
vmaksimo, line 5016: So, we will generate two "duplicating" metadata for now, right?
Should we check for both in the test?

svenvh: Sure; I've extended the test.

PR #3766, opened 2026-05-28
Translate llvm.exp10.* intrinsics

maarquitos14:


test/llvm-intrinsics/fp-intrinsics.ll
```diff
@@ -496,3 +496,15 @@ entry:
 }
 
 declare float @llvm.ldexp.f32.i32(float, i32)
+
```
svenvh: Minor nit: perhaps keep this together with the exp/exp2 tests higher up.

svenvh (approved):

test/llvm-intrinsics/fp-intrinsics.ll
maarquitos14: Yeah, I wasn't sure if we preferred to keep them together, or just append because it's a new test. I'd rather have them together too, let's see what other reviewers think and if they agree, I'll move this.

MrSidims (approved):

PR #3767, opened 2026-06-01
Support `SaturatedToLargestFloat8NormalConversionEXT` decoration

vmaksimo:
Add round-trip handling for `SPV_EXT_float8`'s saturating FP-to-FP8 conversion.

At the LLVM IR level it is represented via `_sat` suffix on the existing `__builtin_spirv_Convert<Src>To<E4M3|E5M2>EXT` interface. When `_sat` is present, the resulting OpFConvert is decorated with `SaturatedToLargestFloat8NormalConversionEXT`.


MrSidims:
> At the LLVM IR level it is represented via _sat suffix on the existing __builtin_spirv_Convert<Src>To<E4M3|E5M2>EXT

Shouldn't it be  `__builtin_spirv_ClampConvert<Src>To<E4M3|E5M2>` instead?

vmaksimo:
> > At the LLVM IR level it is represented via _sat suffix on the existing __builtin_spirv_ConvertTo<E4M3|E5M2>EXT
> 
> Shouldn't it be `__builtin_spirv_ClampConvert<Src>To<E4M3|E5M2>` instead?

I believe it can be both, wanted to publish this for discussion to choose the preferrable one.

MrSidims:
I'd not create instances without necessity :)

vmaksimo:
CI failures are caused by `llc` change https://github.com/llvm/llvm-project/pull/200791

MrSidims (approved):
Fine with me, but if I were you - I'd get someone from Intel to approve :)

lib/SPIRV/SPIRVReader.cpp
```diff
@@ -1100,10 +1100,29 @@ Value *SPIRVToLLVM::transConvertInst(SPIRVValue *BV, Function *F,
 
       FPEncodingWrap SrcEnc = GetEncodingAndUpdateType(SPVSrcTy);
       FPEncodingWrap DstEnc = GetEncodingAndUpdateType(SPVDstTy);
+      bool IsSaturatedFP8 = BC->hasDecorate(
+          DecorationSaturatedToLargestFloat8NormalConversionEXT);
+      if (IsSaturatedFP8) {
+        // Per SPV_EXT_float8, this decoration may only target
+        // OpFConvert/OpConvertSToF/OpConvertUToF whose Result Type uses an
+        // Float8E4M3EXT or Float8E5M2EXT FP encoding.
```
YuriPlyakhin: nit: this comment repeats error message. I would remove it.

lib/SPIRV/SPIRVReader.cpp
```diff
@@ -1100,10 +1100,29 @@ Value *SPIRVToLLVM::transConvertInst(SPIRVValue *BV, Function *F,
...
       if (IsFP4OrFP8Encoding(SrcEnc) || IsFP4OrFP8Encoding(DstEnc) ||
           SPVSrcTy->isTypeInt(4) || SPVDstTy->isTypeInt(4)) {
-        FPConversionDesc FPDesc = {
-            SrcEnc, DstEnc, static_cast<SPIRV::SPIRVWord>(BC->getOpCode())};
+        // SPV_EXT_float8: an OpFConvert to E4M3/E5M2 decorated with
+        // SaturatedToLargestFloat8NormalConversionEXT round-trips through the
+        // ClampConvert<Src>To<E4M3|E5M2>INTEL builtin name.
+        SPIRV::SPIRVWord LookupOp =
+            IsSaturatedFP8 ? internal::OpClampConvertFToFINTEL : OC;
+        FPConversionDesc FPDesc = {SrcEnc, DstEnc, LookupOp};
```
YuriPlyakhin: bug:
When `IsSaturatedFP8` is true, the lookup is rewritten to {Integer, E4M3, OpClampConvertFToFINTEL}, which has no entry in FPConvertToEncodingMap -> rmap asserts.

example:
```
               OpCapability Addresses
               OpCapability Linkage
               OpCapability Kernel
               OpCapability Int8
               OpCapability Float8EXT
               OpExtension "SPV_EXT_float8"
               OpMemoryModel Physical32 OpenCL
               OpDecorate %conv SaturatedToLargestFloat8NormalConversionEXT
               OpDecorate %f LinkageAttributes "f" Export
       %void = OpTypeVoid
         %i8 = OpTypeInt 8 0
        %f8a = OpTypeFloat 8 Float8E4M3EXT
       %func = OpTypeFunction %void %i8
          %f = OpFunction %void None %func
          %x = OpFunctionParameter %i8
          %b = OpLabel
       %conv = OpConvertSToF %f8a %x
               OpReturn
               OpFunctionEnd
```

llvm-spirv -r sat-on-stof.spv -o sat-on-stof.bc :

```
llvm-spirv: SPIRV/libSPIRV/SPIRVUtil.h:102: static Ty1 SPIRV::SPIRVMap<Ty1, Ty2, Identifier>::rmap(Ty2) [with Ty1 = llvm::StringRef; Ty2 = SPIRV::FPConversionDesc; Identifier = void]: Assertion `Found && "Invalid key"' failed.
...
Stack dump:
#11 0x00005d1f29bf28a2 SPIRV::SPIRVToLLVM::transConvertInst(SPIRV::SPIRVValue*, llvm::Function*, llvm::BasicBlock*)
```

lib/SPIRV/SPIRVWriter.cpp
```diff
@@ -5736,6 +5730,17 @@ SPIRVValue *LLVMToSPIRVBase::transDirectCallInst(CallInst *CI,
...
+      // ClampConvert<Src>To<E4M3|E5M2>INTEL builtin maps to OpFConvert
+      // decorated with SaturatedToLargestFloat8NormalConversionEXT instead of
+      // the SPV_INTEL_fp_conversions OpClampConvertFToFINTEL opcode.
+      bool IsSaturatedFP8 =
+          FPDesc.ConvOpCode == internal::OpClampConvertFToFINTEL &&
+          (FPDesc.DstEncoding == FPEncodingWrap::E4M3 ||
+           FPDesc.DstEncoding == FPEncodingWrap::E5M2) &&
+          BM->isAllowedToUseExtension(ExtensionID::SPV_EXT_float8);
+      if (IsSaturatedFP8)
+        FPDesc.ConvOpCode = OpFConvert;
```
YuriPlyakhin: https://github.com/intel/llvm/blob/sycl/sycl/doc/design/spirv-extensions/SPV_INTEL_fp_conversions.asciidoc  `OpClampConvertFToFINTEL` doesn't specify what happens when input is (+/-) inf.

This PR replaces it with `OpFConvert + SaturatedToLargestFloat8NormalConversionEXT` from https://github.khronos.org/SPIRV-Registry/extensions/EXT/SPV_EXT_float8.html
which makes a value which is infinite result in the largest representable normal (i.e. non-infinite) value with a matching sign.

Is this substitution expected by users of `OpClampConvertFToFINTEL`?

lib/SPIRV/SPIRVWriter.cpp
```diff
@@ -5736,6 +5730,17 @@ SPIRVValue *LLVMToSPIRVBase::transDirectCallInst(CallInst *CI,
...
           SPIRV::FPConvertToEncodingMap::map(DemangledName);
+      // SPV_EXT_float8: when SPV_EXT_float8 is enabled, the
+      // ClampConvert<Src>To<E4M3|E5M2>INTEL builtin maps to OpFConvert
+      // decorated with SaturatedToLargestFloat8NormalConversionEXT instead of
+      // the SPV_INTEL_fp_conversions OpClampConvertFToFINTEL opcode.
+      bool IsSaturatedFP8 =
+          FPDesc.ConvOpCode == internal::OpClampConvertFToFINTEL &&
+          (FPDesc.DstEncoding == FPEncodingWrap::E4M3 ||
+           FPDesc.DstEncoding == FPEncodingWrap::E5M2) &&
+          BM->isAllowedToUseExtension(ExtensionID::SPV_EXT_float8);
```
YuriPlyakhin: When `SPV_INTEL_fp_conversions` is enabled, the writer used to emit `OpClampConvertFToFINTEL` for `__builtin_spirv_ClampConvert<...>To<E4M3|E5M2>INTEL`. With this PR the writer always rewrites to `OpFConvert` + `SaturatedToLargestFloat8NormalConversionEXT`. There's no longer a way to ask the translator to emit `OpClampConvertFToFINTEL` with an FP8 result type.

Is that intended?

If `SPV_EXT_float8` is meant to supersede `SPV_INTEL_fp_conversions` for FP8 destinations, please call that out in PR description.

test/extensions/INTEL/SPV_INTEL_fp_conversions/spv_intel_fp_conversions.ll
```diff
@@ -61,7 +61,7 @@ target triple = "spir-unknown-unknown"
 ; Followings tests are for clamp rounding
 
 ; CHECK-SPIRV: Function [[#]] [[#hf16_hf8_clamp]] [[#]]
-; CHECK-SPIRV: ClampConvertFToFINTEL [[#HFloat8Ty]] [[#Conv:]] [[#HalfConst]]
+; CHECK-SPIRV: FConvert [[#HFloat8Ty]] [[#Conv:]] [[#HalfConst]]
```
YuriPlyakhin: could you please also add check for `SaturatedToLargestFloat8NormalConversionEXT` in SPIRV part of the checks.

lib/SPIRV/SPIRVReader.cpp
vmaksimo: Fixed the assert: rmap rewrite to `OpClampConvertFToFINTEL` is now scoped to `OC == OpFConvert`, so `OpConvertSToF/OpConvertUToF` with the decoration no longer hit a missing key.

Though spec allows the decoration on `OpConvertSToF/OpConvertUToF`, saturation is semantically meaningful only for i16+ sources, so the test you added exercises a case no producer emits today. 
For now I think the assert fix should be sufficient.

lib/SPIRV/SPIRVWriter.cpp
vmaksimo: Nice catch, I was a bit rush to change the behavior. Restored the behavior for `SPV_INTEL_fp_conversions` in  a577c55 so `SPV_EXT_float8` does not supersede

lib/SPIRV/SPIRVWriter.cpp
vmaksimo: changed the behavior in a577c55, guess this is no longer relevant

lib/SPIRV/SPIRVWriter.cpp
MrSidims: > With this PR the writer always rewrites to OpFConvert + SaturatedToLargestFloat8NormalConversionEXT. There's no longer a way to ask the translator to emit OpClampConvertFToFINTEL with an FP8 result type.

Is it bad?

YuriPlyakhin: I don't know, maybe it is good. I was just asking, if it was the intention. :)
@vmaksimo , you might want to verify with @gmlueck and @bashbaug on how we want these 2 extensions to coexist...

vmaksimo: My intent is to simplify this very change to have the decoration support from the multi-vendor extension.
I'd prefer to resolve any collisions with the INTEL extension in a separate PR when we clarify our spec changes

gmlueck: > If `SPV_EXT_float8` is meant to supersede `SPV_INTEL_fp_conversions` for FP8 destinations, please call that out in PR description.

I think we *do* want `SPV_EXT_float8` to supersede `SPV_INTEL_fp_conversions`.  Before @bashbaug went on vacation, he updated the draft `SPV_INTEL_fp_conversions` to remove the overlapping parts like **OpClampConvertFToFINTEL**.  Therefore, we shouldn't be emitting those instructions.

It would be nice to get this change into the SPRIV Translator before merging intel/llvm#21568, so that we don't have compiled SYCL applications using the outdated **OpClampConvertFToFINTEL**, etc. instructions.

gmlueck: I talked to @YuriPlyakhin offline.  If you want to make the change that causes `SPV_EXT_float8` to supersede `SPV_INTEL_fp_conversions` in a separate PR, that's OK with me.  I only wanted to say that we should have this as a goal.  I don't mind if it happens in a separate PR.

test/extensions/EXT/SPV_EXT_float8/clamp_convert_to_fp8.ll
```diff
@@ -0,0 +1,56 @@
+; Clamping FP-to-FP8 conversions: when SPV_EXT_float8 is enabled, the
```
YuriPlyakhin, line 1: nit
```suggestion
; Clamping FP-to-FP8 conversions: when SPV_EXT_float8 is enabled and SPV_INTEL_fp_conversions is not, the
```

YuriPlyakhin (approved):
LGTM. one nit.

PR #3770, opened 2026-06-04
Support SPV_INTEL_rounded_div_sqrt extension

Qi-Ye-079:
Support the new SPV_INTEL_rounded_div_sqrt extension in the translator.

Currently the `FPRoundingMode` decoration is translated already, but the `Extension` and `Capability` aren't generated. This PR registers the new extension and capability in translator, and ensures their generation in the resulting SPIR-V module if `FPRoundingMode` is decorated on applicable instructions (`fdiv` or `sqrt`).

Spec: https://github.khronos.org/SPIRV-Registry/extensions/INTEL/SPV_INTEL_rounded_divide_sqrt.html


lib/SPIRV/libSPIRV/spirv_internal.hpp
vmaksimo, line 1: We should not modify `spirv_internal.hpp` for the extensions that are published to Khronos - their opcodes should be already defined in [spirv.hpp](https://github.com/KhronosGroup/SPIRV-Headers/blob/main/include/spirv/unified1/spirv.hpp) from SPIR-V Headers.

So, the changes here can be safely removed (check if we need to update `spirv-headers-tag.conf` to contain the newer header's commit hash).
All the `internal::` namespaces in other files to be removed as well.

test/extensions/INTEL/SPV_INTEL_rounded_divide_sqrt/rounding_supported.ll
```diff
@@ -0,0 +1,199 @@
+; Verifies that FPRoundingMode decorations on OpFDiv and sqrt (OpenCL.std
+; extended instruction) trigger the RoundedDivideSqrtINTEL capability and the
+; SPV_INTEL_rounded_divide_sqrt extension.
+
+; RUN: llvm-as %s -o %t.bc
+; RUN: llvm-spirv %t.bc -o %t.spv --spirv-ext=+SPV_INTEL_rounded_divide_sqrt
```
vmaksimo:
```suggestion
; RUN: llvm-spirv %s -o %t.spv --spirv-ext=+SPV_INTEL_rounded_divide_sqrt
```
same for other tests

test/extensions/INTEL/SPV_INTEL_rounded_divide_sqrt/rounding_unsupported.ll
```diff
@@ -0,0 +1,48 @@
+; Verifies that FPRoundingMode is NOT emitted for instructions other than
+; fdiv and sqrt, and that neither RoundedDivideSqrtINTEL capability nor the
+; SPV_INTEL_rounded_divide_sqrt extension is generated as a result.
+
+; RUN: llvm-as %s -o %t.bc
+; RUN: llvm-spirv %t.bc -o %t.spv --spirv-ext=+SPV_INTEL_rounded_divide_sqrt,+SPV_KHR_bfloat16,+SPV_INTEL_bfloat16_arithmetic
```
vmaksimo: Let's not test interaction with other extensions. Moreover, `SPV_INTEL_rounded_divide_sqrt` spec says:
> This extension does not currently support rounding modes on floating-point types with non-IEEE 754 encodings, such as bfloat16.

lib/SPIRV/SPIRVWriter.cpp
```diff
@@ -4691,7 +4754,13 @@ SPIRVValue *LLVMToSPIRVBase::transIntrinsicInst(IntrinsicInst *II,
     auto *BI = BM->addBinaryInst(OpFDiv, transType(II->getType()),
                                  transValue(II->getArgOperand(0), BB),
                                  transValue(II->getArgOperand(1), BB), BB);
-    return applyRoundingModeConstraint(II->getOperand(2), BI);
+    // TODO: we should prevent other constrained instructions having
+    // FPRoundingMode decoration (except for those conversion insts) as they
+    // violate the spec
+    if (isApplicableForRoundingModeINTEL(BI)) {
```
vmaksimo: Hm, would not this be enough?
```suggestion
    if (BM->isAllowedToUseExtension(ExtensionID::SPV_INTEL_rounded_divide_sqrt))
```

vmaksimo (commented):
I'd suggest we'd support `llvm.experimental.constrained.sqrt` intrinsic to demonstrate the SPIR-V extension instead of supporting FP rounding mode via `!spirv.Decorations`. My thinking is that it'll simplify the changes needed in Writer.

lib/SPIRV/SPIRVWriter.cpp
Qi-Ye-079: I think we should also check if the instruction has IEEE-754 FP types, shouldn't we? Which is what the helper function `isApplicableForRoundingModeINTEL` trying to check.

test/extensions/INTEL/SPV_INTEL_rounded_divide_sqrt/rounding_unsupported.ll
Qi-Ye-079: Yeah this is exactly what I want to test, i.e., a non-IEEE-754 type (such as `bfloat16`), which will NOT generate `FPRoundingMode` even though `SPV_INTEL_rounded_divide_sqrt` is enabled. Apparently, `bfloat16` needs these 2 extra extensions.

lib/SPIRV/libSPIRV/spirv_internal.hpp
Qi-Ye-079: Ok got it. I find the capability is currently defined in `include/spirv/unified1` directory only (even in the latest `main` branch) in SPIRV-Header.

Qi-Ye-079:
> I'd suggest we'd support `llvm.experimental.constrained.sqrt` intrinsic to demonstrate the SPIR-V extension instead of supporting FP rounding mode via `!spirv.Decorations`. My thinking is that it'll simplify the changes needed in Writer.

Ok. Since these `!spirv-Decorations` can also generate `FPRoundingMode` decorations, so I thought I needed to check them as well...

lib/SPIRV/libSPIRV/spirv_internal.hpp
vmaksimo: Yep, this is expected :) we use only this header here in our builds

test/extensions/INTEL/SPV_INTEL_rounded_divide_sqrt/rounding_unsupported.ll
vmaksimo: Ah, okay, my bad! My eye didn't catch LLVM supports bfloat for these instructions as well as IEEE-754 types

lib/SPIRV/SPIRVWriter.cpp
vmaksimo: Hm, then maybe:
```suggestion
    if (BM->isAllowedToUseExtension(ExtensionID::SPV_INTEL_rounded_divide_sqrt) && 
          isIEEE754FPType(BI->getType())
```
as we already know what the instruction is being translated, no need to call `isDivOrOCLSqrt()`

vmaksimo:
> Ok. Since these `!spirv-Decorations` can also generate `FPRoundingMode` decorations, so I thought I needed to check them as well...

You're right, but I think for now we can stay with only constrained intrinsics for simplicity - seems to be good enough. 
And we'd always be able to support `!spirv-Decorations` in a separate PR if actually needed.

lib/SPIRV/libSPIRV/spirv_internal.hpp
Qi-Ye-079: Done.

test/extensions/INTEL/SPV_INTEL_rounded_divide_sqrt/rounding_supported.ll
Qi-Ye-079: Done.

lib/SPIRV/SPIRVWriter.cpp
Qi-Ye-079: Done.

lib/SPIRV/SPIRVWriter.cpp
```diff
@@ -3371,8 +3393,13 @@ bool LLVMToSPIRVBase::transDecoration(Value *V, SPIRVValue *BV) {
...
       transMetadataDecorations(IDecoMD, BV);
-    if (BV->isInst())
-      addFPBuiltinDecoration(BM, Inst, static_cast<SPIRVInstruction *>(BV));
+    if (BV->isInst()) {
+      auto *BI = static_cast<SPIRVInstruction *>(BV);
+      addFPBuiltinDecoration(BM, Inst, BI);
+      if (BI->hasFPRoundingMode() && isDivOrOCLSqrt(BI)) {
+        BM->addCapability(CapabilityRoundedDivideSqrtINTEL);
+      }
+    }
```
vmaksimo: Maybe we can add Capability right before `applyRoundingModeConstraint()`? This way we can remove this code block as well as unused `isDivOrOCLSqrt()`

lib/SPIRV/SPIRVWriter.cpp
```diff
@@ -4691,7 +4719,27 @@ SPIRVValue *LLVMToSPIRVBase::transIntrinsicInst(IntrinsicInst *II,
     auto *BI = BM->addBinaryInst(OpFDiv, transType(II->getType()),
                                  transValue(II->getArgOperand(0), BB),
                                  transValue(II->getArgOperand(1), BB), BB);
-    return applyRoundingModeConstraint(II->getOperand(2), BI);
+    // TODO: we should prevent other constrained instructions having
+    // FPRoundingMode decoration (except for those conversion insts) as they
+    // violate the spec
```
vmaksimo: Maybe to remove TODO as you already create a PR to resolve this :)

vmaksimo (commented):
Have you accidentally removed `rounding_unsupported.ll` test? 
Also, clang-format should be fixed

lib/SPIRV/SPIRVWriter.cpp
Qi-Ye-079: This would work too, but I'm adding it here simply because this is the single convergence point where all intrinsic handler will reach. For now we only have 2 intrinsics that supports this extension, but if we were to add supports for more intrinsics in the future, we need to add the capability for each intrinsic case which seems a bit cumbersome. I don't know if this may happen or not, I'm just ok with either.

lib/SPIRV/SPIRVWriter.cpp
Qi-Ye-079: I'm planning to remove it in my next PR, but... I can remove here also

Qi-Ye-079:
> Have you accidentally removed `rounding_unsupported.ll` test? Also, clang-format should be fixed

I removed it because it's no longer applicable - currently all instructions marked with `!spirv-Decoration`  or constrained intrinsics will be translated to `FPRoundingMode` attached. This test can be added back for my next PR.

vmaksimo (approved):

lib/SPIRV/SPIRVWriter.cpp
```diff
@@ -65,6 +65,8 @@
 #include "SPIRVUtil.h"
 #include "SPIRVValue.h"
 #include "VectorComputeUtil.h"
+#include "libSPIRV/SPIRVDecorate.h"
```
YuriPlyakhin: I see only SPIRVWriter implementation. What is the plan for reverse translation implementation?

test/llvm-intrinsics/constrained-arithmetic.ll
```diff
@@ -1,5 +1,5 @@
 ; RUN: llvm-as %s -o %t.bc
-; RUN: llvm-spirv %t.bc -spirv-text -o - | FileCheck %s
+; RUN: llvm-spirv %t.bc -spirv-text -o - --spirv-ext=+SPV_INTEL_rounded_divide_sqrt | FileCheck %s
```
YuriPlyakhin: Could you please also add a run line without passing extension, so we verify no-extension case as well (it should drop FPRoundingMode on FDiv in that case)?
Something like that: `llvm-spirv %t.bc -spirv-text -o - | FileCheck %s --check-prefix=CHECK-NOEXT)`

lib/SPIRV/SPIRVWriter.cpp
YuriPlyakhin: I think adding capability right before `applyRoundingModeConstraint` is less error-prone long term in my opinion.
If anyone in the future adds a path that puts FPRoundingMode on non-IEEE-754 type, this single point will emit RoundedDivideSqrtINTEL.
I think in addition to `isRoundingModeCapableINTEL`, we also need to check `isIEEE754FPType` to avoid errors like that, but it seems it is easier to just drop this code as well as `isRoundingModeCapableINTEL` and move `addCapability` next to the intrinsic-side `applyRoundingModeConstraint`.
I lean toward the latter: it's two extra lines today and removes a redundant helper that has to stay in lock-step with whatever opcodes the spec allows in the future.

YuriPlyakhin: Another similar point is `addCapability` is calling `addExtension` and it checks `isAllowedToUseExtension`, and if this extension was not enabled, it will call `setInvalid`. So if `FPRoundingMode` ever added on Div/sqrt without `SPV_INTEL_rounded_divide_sqrt`, this single point code will mark the module as invalid instead of no-opping. In order to avoid it, we would need to check `BM->isAllowedToUseExtension` here as well.

While this is hypothetical scenario, it seems using addCapability together with applyRoundingModeConstraint helps us avoid this with very small cost.

lib/SPIRV/SPIRVWriter.cpp
```diff
@@ -65,6 +65,8 @@
...
+#include "libSPIRV/SPIRVError.h"
```
YuriPlyakhin: do we have to add those new include lines? I think both are already reachable transitively.
If we need to include them, please, match the style of other includes: no `libSPIRV/`

lib/SPIRV/SPIRVWriter.cpp
vmaksimo: I'm mostly sure that fdiv/sqrt are translated back to OCL builtins/SPIR-V friendly IR calls that will include rounding information in their names - nothing to do from our side here. 
But would be nice to see actual RUN and CHECK lines for the reverse translation, at least for the newly added tests.
@Qi-Ye-079 we'd like to see both paths tested: `llvm-spirv -r --spirv-target-env=SPV-IR` and `llvm-spirv -r` (OCL/default)

lib/SPIRV/SPIRVWriter.cpp
Qi-Ye-079: Done. Thanks for the insight.

test/llvm-intrinsics/constrained-arithmetic.ll
Qi-Ye-079: Done.

lib/SPIRV/SPIRVWriter.cpp
Qi-Ye-079: Removed.

lib/SPIRV/SPIRVWriter.cpp
Qi-Ye-079: I added a new test for the Reader, but I confirmed that the `FPRoundingMode` decoration will be translated to `!spirv-Decorations` metadata in LLVM IR _only_ under SPV-IR env, but not for OCL in `SPIRVReader.cpp`:
```
// Decoration metadata is only enabled in SPIR-V friendly mode
  if (BM->getDesiredBIsRepresentation() == BIsRepresentation::SPIRVFriendlyIR)
    transDecorationsToMetadata(BV, V);
```
Thus the rounding mode is gone after translating back to LLVM IR for OCL. Should we add the metadata also for OCL? I feel like this MD should be a SPIR-V only thing and shouldn't be applicable for OCL, but I may be wrong.

YuriPlyakhin: @Qi-Ye-079 , could you please merge together `test/extensions/INTEL/SPV_INTEL_rounded_divide_sqrt/rounding_supported.ll` and `test/extensions/INTEL/SPV_INTEL_rounded_divide_sqrt/rounding_supported_reader.ll`?

and maybe adding checks for `!spirv.Decorations` for reverse translation on vector types makes sense also.

YuriPlyakhin: As to not producing this MD for OCL, it seems it is expected, but I'll let @vmaksimo to confirm.

Qi-Ye-079: > @Qi-Ye-079 , could you please merge together `test/extensions/INTEL/SPV_INTEL_rounded_divide_sqrt/rounding_supported.ll` and `test/extensions/INTEL/SPV_INTEL_rounded_divide_sqrt/rounding_supported_reader.ll`?
> 
> and maybe adding checks for `!spirv.Decorations` for reverse translation on vector types makes sense also.

Done.

test/extensions/INTEL/SPV_INTEL_rounded_divide_sqrt/rounding_supported.ll
```diff
@@ -0,0 +1,187 @@
...
+; CHECK-SPIRV-DAG: TypeFloat [[#DOUBLE:]] 64
+
+; CHECK-SPIRV-DAG: TypeVector [[#HALFV:]] [[#HALF]] 2
+; CHECK-SPIRV-DAG: TypeVector [[#FLOATV:]] [[#FLOAT]] 4
+; CHECK-SPIRV-DAG: TypeVector [[#DOUBLEV:]] [[#DOUBLE]] 3
+
+; In SPV-IR mode the FPRoundingMode decoration (Decoration 39) is preserved as
+; an !spirv.Decorations metadata node, captured per rounding mode at its first
+; use below and verified at the end. The same node is shared across every
+; fdiv/sqrt of that mode (scalar and vector alike).
```
YuriPlyakhin: nit:
could you please move the comment down to just before the first CHECK-LLVM-SPV-LABEL block (line 58), where the capture ([[RTE:![0-9]+]]) actually first occurs.
That puts the explanation right where the "captured at its first use" pattern begins.

YuriPlyakhin (approved):
LGTM. I'd like @vmaksimo to confirm that reverse-translation works as expected.

test/extensions/INTEL/SPV_INTEL_rounded_divide_sqrt/rounding_supported.ll
Qi-Ye-079: Done.

test/extensions/INTEL/SPV_INTEL_rounded_divide_sqrt/rounding_supported.ll
```diff
@@ -0,0 +1,186 @@
...
+
+; All four rounding modes on vector-typed constrained sqrts.
+; CHECK-SPIRV-DAG: Decorate [[#SV_RTE:]] FPRoundingMode 0
+; CHECK-SPIRV-DAG: Decorate [[#SV_RTZ:]] FPRoundingMode 1
+; CHECK-SPIRV-DAG: Decorate [[#SV_RTP:]] FPRoundingMode 2
+; CHECK-SPIRV-DAG: Decorate [[#SV_RTN:]] FPRoundingMode 3
+; CHECK-SPIRV-DAG: ExtInst [[#HALFV]]   [[#SV_RTE]] {{[0-9]+}} sqrt
+; CHECK-SPIRV-DAG: ExtInst [[#FLOATV]]  [[#SV_RTZ]] {{[0-9]+}} sqrt
+; CHECK-SPIRV-DAG: ExtInst [[#DOUBLEV]] [[#SV_RTP]] {{[0-9]+}} sqrt
+; CHECK-SPIRV-DAG: ExtInst [[#DOUBLEV]] [[#SV_RTN]] {{[0-9]+}} sqrt
```
vmaksimo: let's use modern style :)
```suggestion
; CHECK-SPIRV-DAG: ExtInst [[#HALFV]]   [[#SV_RTE]] [[#]] sqrt
; CHECK-SPIRV-DAG: ExtInst [[#FLOATV]]  [[#SV_RTZ]] [[#]] sqrt
; CHECK-SPIRV-DAG: ExtInst [[#DOUBLEV]] [[#SV_RTP]] [[#]] sqrt
; CHECK-SPIRV-DAG: ExtInst [[#DOUBLEV]] [[#SV_RTN]] [[#]] sqrt
```

test/llvm-intrinsics/constrained-arithmetic.ll
```diff
@@ -19,6 +20,14 @@
 
 ; CHECK-NOT: Decorate {{[0-9]+}} FPRoundingMode
 
+; Without SPV_INTEL_rounded_divide_sqrt enabled, neither the capability nor the
+; extension is emitted, and the FPRoundingMode decoration is dropped from the
+; FDiv ("div", the only towardzero/mode-1 user here).
+; CHECK-NOEXT-NOT: Capability RoundedDivideSqrtINTEL
+; CHECK-NOEXT-NOT: Extension "SPV_INTEL_rounded_divide_sqrt"
+; CHECK-NOEXT: Name [[di:[0-9]+]] "div"
+; CHECK-NOEXT-NOT: Decorate [[di]] FPRoundingMode
```
vmaksimo:
```suggestion
; CHECK-NOEXT: Name [[#di:]] "div"
; CHECK-NOEXT-NOT: Decorate [[#di]] FPRoundingMode
```

test/extensions/INTEL/SPV_INTEL_rounded_divide_sqrt/fconvert_unaffected.ll
```diff
@@ -0,0 +1,27 @@
...
+; RUN: FileCheck %s --input-file %t.spt --check-prefix=CHECK-SPIRV
+
+target datalayout = "e-p:32:32-i64:64-v16:16-v24:32-v32:32-v48:64-v96:128-v192:256-v256:256-v512:512-v1024:1024"
+target triple = "spir-unknown-unknown"
+
+; CHECK-SPIRV-NOT: Capability RoundedDivideSqrtINTEL
+; CHECK-SPIRV-NOT: Extension "SPV_INTEL_rounded_divide_sqrt"
+; The conversion still carries its rounding-mode decoration (core SPIR-V).
+; CHECK-SPIRV: Decorate [[#CVT:]] FPRoundingMode 1
+; CHECK-SPIRV: FConvert {{[0-9]+}} [[#CVT]]
```
vmaksimo:
```suggestion
; CHECK-SPIRV: FConvert [[#]] [[#CVT]]
```
please use it in the metadata checks as well (applicable to  both new tests)

lib/SPIRV/SPIRVWriter.cpp
vmaksimo: Confirming, now we expect `!spirv.Decorations` only in SPIR-V friendly IR mode.

vmaksimo (commented):
Some styling tips for tests.
You can now rebase on top of main (with #3771 merged) - guess some tests need to be aligned as well.

test/extensions/INTEL/SPV_INTEL_rounded_divide_sqrt/rounding_supported.ll
Qi-Ye-079: Done

test/llvm-intrinsics/constrained-arithmetic.ll
Qi-Ye-079: Done

test/extensions/INTEL/SPV_INTEL_rounded_divide_sqrt/fconvert_unaffected.ll
Qi-Ye-079: Done

lib/SPIRV/SPIRVWriter.cpp
```diff
@@ -4704,11 +4711,29 @@ SPIRVValue *LLVMToSPIRVBase::transIntrinsicInst(IntrinsicInst *II,
                              transValue(II->getArgOperand(1), BB), BB);
   }
   case Intrinsic::experimental_constrained_fdiv: {
-    SPIRVDBG(dbgs() << "WARNING: Rounding Mode of constrained intrinsic "
-                       "\"fdiv\" not supported by SPIR-V, dropping it.\n");
```
vmaksimo, line 4708: Hm, should we still emit warning if the extension is not enabled?
If so, we may want to also add a warning to new sqrt intrinsic

test/extensions/INTEL/SPV_INTEL_rounded_divide_sqrt/rounding_supported.ll
```diff
@@ -0,0 +1,186 @@
...
+declare double @llvm.experimental.constrained.sqrt.f64(double, metadata, metadata)
+
+declare <2 x half>   @llvm.experimental.constrained.sqrt.v2f16(<2 x half>, metadata, metadata)
+declare <4 x float>  @llvm.experimental.constrained.sqrt.v4f32(<4 x float>, metadata, metadata)
+declare <3 x double> @llvm.experimental.constrained.sqrt.v3f64(<3 x double>, metadata, metadata)
+
+; Verify the FPRoundingMode (Decoration 39) metadata nodes captured above:
+; mode 0 = RTE, 1 = RTZ, 2 = RTP, 3 = RTN.
+; CHECK-LLVM-SPV-DAG: [[RTE]] = !{[[RTEX:![0-9]+]]}
+; CHECK-LLVM-SPV-DAG: [[RTEX]] = !{i32 39, i32 0}
```
vmaksimo: Please apply this code style for newly added test...
```suggestion
; CHECK-LLVM-SPV-DAG: ![[#RTE]] = !{![[#RTEX:]]}
; CHECK-LLVM-SPV-DAG: ![[#RTEX]] = !{i32 39, i32 0}
```

test/llvm-intrinsics/constrained-arithmetic.ll
```diff
@@ -1,5 +1,6 @@
 ; RUN: llvm-as %s -o %t.bc
-; RUN: llvm-spirv %t.bc -spirv-text -o - | FileCheck %s --implicit-check-not FPRoundingMode
```
vmaksimo, line 2: We should not have removed this line. Why you don't checking with `--implicit-check-not FPRoundingMode` anymore?
Let's keep the diff as small as possible.

test/llvm-intrinsics/constrained-arithmetic.ll
```diff
@@ -1,5 +1,6 @@
...
+; RUN: llvm-spirv %t.bc -spirv-text -o - --spirv-ext=+SPV_INTEL_rounded_divide_sqrt | FileCheck %s
+; RUN: llvm-spirv %t.bc -spirv-text -o - | FileCheck %s --check-prefix=CHECK-NOEXT
```
vmaksimo: if running two modes, please run `spirv-val` on both SPV files translated with/without extension

lib/SPIRV/SPIRVWriter.cpp
```diff
@@ -2971,6 +2971,12 @@ void checkIsGlobalVar(SPIRVEntry *E, Decoration Dec) {
                               ErrStr);
 }
 
+static bool isIEEE754FPType(SPIRVType *Ty) {
+  if (Ty->isTypeVector())
+    Ty = Ty->getVectorComponentType();
+  return Ty->isTypeFloat(16) || Ty->isTypeFloat(32) || Ty->isTypeFloat(64);
+}
```
vmaksimo: Tip: maybe move this to be a method of SPIRVType?

test/extensions/INTEL/SPV_INTEL_rounded_divide_sqrt/rounding_supported.ll
```diff
@@ -0,0 +1,186 @@
...
+
+; In SPV-IR mode the FPRoundingMode decoration (Decoration 39) is preserved as
+; an !spirv.Decorations metadata node, captured per rounding mode at its first
+; use below and verified at the end. The same node is shared across every
+; fdiv/sqrt of that mode (scalar and vector alike).
+; CHECK-LLVM-SPV-LABEL: @test_fdiv_scalar
+; CHECK-LLVM-SPV: fdiv half %{{.*}}, !spirv.Decorations [[RTE:![0-9]+]]
+; CHECK-LLVM-SPV: fdiv float %{{.*}}, !spirv.Decorations [[RTZ:![0-9]+]]
+; CHECK-LLVM-SPV: fdiv double %{{.*}}, !spirv.Decorations [[RTP:![0-9]+]]
+; CHECK-LLVM-SPV: fdiv double %{{.*}}, !spirv.Decorations [[RTN:![0-9]+]]
```
vmaksimo: Hmm, will smth like this work? We can avoid duplication of the same check lines wherever possible (you'll need to use `--check-prefixes=CHECK-LLVM,CHECK-LLVM-SPV` this way)

```suggestion
; CHECK-LLVM-LABEL: @test_fdiv_scalar
; CHECK-LLVM: fdiv half
; CHECK-LLVM-SPV-SAME: !spirv.Decorations ![[#RTE:]]
```

lib/SPIRV/SPIRVWriter.cpp
Qi-Ye-079: Done

lib/SPIRV/SPIRVWriter.cpp
Qi-Ye-079: Done

test/llvm-intrinsics/constrained-arithmetic.ll
Qi-Ye-079: Done

test/llvm-intrinsics/constrained-arithmetic.ll
Qi-Ye-079: Done

test/extensions/INTEL/SPV_INTEL_rounded_divide_sqrt/rounding_supported.ll
Qi-Ye-079: Done

test/extensions/INTEL/SPV_INTEL_rounded_divide_sqrt/rounding_supported.ll
Qi-Ye-079: Done

test/llvm-intrinsics/constrained-arithmetic.ll
```diff
@@ -8,14 +11,25 @@
...
 
 ; CHECK: FAdd {{[0-9]+}} [[ad]]
 ; CHECK: FDiv {{[0-9]+}} [[di]]
 ; CHECK: FSub {{[0-9]+}} [[su]]
 ; CHECK: FMul {{[0-9]+}} [[mu]]
+; CHECK: ExtInst {{[0-9]+}} [[sq]] [[#]] sqrt
 ; CHECK: FMul
 ; CHECK: FAdd
-; CHECK: ExtInst {{[0-9]+}} {{[0-9]+}} {{[0-9]+}} fma
+; CHECK: ExtInst {{[0-9]+}} {{[0-9]+}} [[#]] fma
```
vmaksimo: AI artifact? No need to mix styles - follow the nearby style when editing existing lines; use modern [[#]] when adding new tests or check blocks.

test/llvm-intrinsics/constrained-arithmetic.ll
```diff
@@ -8,14 +11,25 @@
...
+; CHECK: Name [[sq:[0-9]+]] "sqrt"
+
+; CHECK-DAG: Decorate [[di]] FPRoundingMode 1
+; CHECK-DAG: Decorate [[sq]] FPRoundingMode 3
+
+; Without SPV_INTEL_rounded_divide_sqrt enabled, neither the capability nor the
+; extension is emitted, and the FPRoundingMode decoration is dropped from the
+; FDiv and sqrt here
+; CHECK-NOEXT-NOT: Capability RoundedDivideSqrtINTEL
+; CHECK-NOEXT-NOT: Extension "SPV_INTEL_rounded_divide_sqrt"
```
vmaksimo: We have extension/capability at the top of the SPV logical layout, move this as first check lines after RUN lines.

test/llvm-intrinsics/constrained-arithmetic.ll
vmaksimo: Not exactly - we'd better not edit this line at all (we should not change the "core" test). The new extension-related run and check lines should go after. Please review the diff against main prior pushing the next revision.

test/extensions/INTEL/SPV_INTEL_rounded_divide_sqrt/fconvert_unaffected.ll
```diff
@@ -0,0 +1,27 @@
+; Guards target-sensitivity: FPRoundingMode is valid on conversion
+; instructions (OpFConvert) in core SPIR-V, so even with
+; SPV_INTEL_rounded_divide_sqrt enabled on the command line, an FPRoundingMode
+; that lands on an OpFConvert (here from llvm.experimental.constrained.fptrunc)
+; must NOT pull in the RoundedDivideSqrtINTEL capability or the extension.
+
+; RUN: llvm-spirv %s -o %t.spv --spirv-ext=+SPV_INTEL_rounded_divide_sqrt
+; RUN: spirv-val %t.spv
+; RUN: llvm-spirv %t.spv -o %t.spt --to-text
+; RUN: FileCheck %s --input-file %t.spt --check-prefix=CHECK-SPIRV
```
vmaksimo: No need for setting a separate check-prefix - no other check scenarios to distinguish between, default `CHECK` will work.

test/extensions/INTEL/SPV_INTEL_rounded_divide_sqrt/rounding_supported.ll
```diff
@@ -0,0 +1,180 @@
...
+; CHECK-SPIRV-DAG: Decorate [[#S_RTZ:]] FPRoundingMode 1
+; CHECK-SPIRV-DAG: Decorate [[#S_RTP:]] FPRoundingMode 2
+; CHECK-SPIRV-DAG: Decorate [[#S_RTN:]] FPRoundingMode 3
+; CHECK-SPIRV-DAG: ExtInst [[#HALF]]   [[#S_RTE]] [[#]] sqrt
+; CHECK-SPIRV-DAG: ExtInst [[#FLOAT]]  [[#S_RTZ]] [[#]] sqrt
+; CHECK-SPIRV-DAG: ExtInst [[#DOUBLE]] [[#S_RTP]] [[#]] sqrt
+; CHECK-SPIRV-DAG: ExtInst [[#DOUBLE]] [[#S_RTN]] [[#]] sqrt
+
+; The sqrt builtin is mangled differently per mode (plain "sqrt" for OCL,
+; "__spirv_ocl_sqrt" for SPV-IR), so the shared line matches the common prefix.
```
vmaksimo: This very case is not as `fdiv` - the instructions are actually different (SPV OCL builtin as function call vs plain sqrt). Please restore separate CHECK lines for different reverse translation modes

vmaksimo (commented):
Getting closer to the merge, hopefully last review round

test/extensions/INTEL/SPV_INTEL_rounded_divide_sqrt/fconvert_unaffected.ll
```diff
@@ -0,0 +1,27 @@
+; Guards target-sensitivity: FPRoundingMode is valid on conversion
+; instructions (OpFConvert) in core SPIR-V, so even with
+; SPV_INTEL_rounded_divide_sqrt enabled on the command line, an FPRoundingMode
+; that lands on an OpFConvert (here from llvm.experimental.constrained.fptrunc)
+; must NOT pull in the RoundedDivideSqrtINTEL capability or the extension.
```
MrSidims: May I ask you to re-phrase this clarification in your own words? (same applicable to other comments in the tests in the PR)

Opinionated justification: I really had hard times reading and trying to understand the comment.

Not-opinionated justification: just imagine a mis-configured LLM will read the test file in the future. Instead of just dropping unnecessary text from the context it would store it as is, increasing tokens usage and probability of hallucination. Meanwhile such verbosity of this comment brings close to zero value compared to a shorter, humanized comment.

Some examples of shorten comments below.

test/extensions/INTEL/SPV_INTEL_rounded_divide_sqrt/rounding_supported.ll
```diff
@@ -0,0 +1,180 @@
...
+; SPV_INTEL_rounded_divide_sqrt extension, and round-trip back through both
+; reverse-translation paths.
+
+; RUN: llvm-spirv %s -o %t.spv --spirv-ext=+SPV_INTEL_rounded_divide_sqrt
+; RUN: spirv-val %t.spv
+; RUN: llvm-spirv %t.spv -o %t.spt --to-text
+; RUN: FileCheck %s --input-file %t.spt --check-prefix=CHECK-SPIRV
+
+; Reverse translation, OCL/default path: there is no OCL representation for the
+; rounding mode on fdiv/sqrt, so it is dropped (plain fdiv, plain sqrt builtin).
```
MrSidims:
```suggestion
; OCL path: there is no OCL representation for the
; rounding mode on fdiv/sqrt, so it is dropped.
```

test/extensions/INTEL/SPV_INTEL_rounded_divide_sqrt/rounding_supported.ll
```diff
@@ -0,0 +1,180 @@
...
+; The --implicit-check-not flags assert no rounding info leaks anywhere.
```
MrSidims: remove clarification, it's obvious what "--implicit-check-not" does

test/extensions/INTEL/SPV_INTEL_rounded_divide_sqrt/rounding_supported.ll
MrSidims: > there is no OCL representation for the rounding mode on fdiv/sqrt

Strictly speaking, nothing forbids us from reusing `spirv.Decoration` metadata in this case.

test/extensions/INTEL/SPV_INTEL_rounded_divide_sqrt/rounding_supported.ll
```diff
@@ -0,0 +1,180 @@
...
+; RUN: llvm-spirv -r %t.spv -o %t.rev.bc
+; RUN: llvm-dis %t.rev.bc -o - | FileCheck %s --check-prefix=CHECK-LLVM \
+; RUN:   --implicit-check-not=FPRoundingMode --implicit-check-not=spirv.Decorations
+
+; Reverse translation, SPV-IR path: the rounding mode is preserved as an
+; !spirv.Decorations FPRoundingMode (Decoration 39) attached to the fdiv/sqrt.
```
MrSidims:
```suggestion
; SPV-IR path: the rounding mode is preserved as an
; !spirv.Decorations FPRoundingMode metadata attached to the fdiv/sqrt.
```

test/llvm-intrinsics/constrained-arithmetic.ll
Qi-Ye-079: Done

test/llvm-intrinsics/constrained-arithmetic.ll
Qi-Ye-079: Sure. I reverted the style for `fma`, but use modern style for `sqrt`.

test/llvm-intrinsics/constrained-arithmetic.ll
Qi-Ye-079: Done. Maintained the same RUN lines for original tests, and added new RUNs after them.

test/extensions/INTEL/SPV_INTEL_rounded_divide_sqrt/fconvert_unaffected.ll
Qi-Ye-079: Done

test/extensions/INTEL/SPV_INTEL_rounded_divide_sqrt/rounding_supported.ll
vmaksimo: True, but currently not used, and was not intended initially:
https://github.com/KhronosGroup/SPIRV-LLVM-Translator/blob/a34ca53ef882ba11b6f7244e2a156078bcbdb03f/lib/SPIRV/SPIRVReader.cpp#L4810-L4812
For now the comment just describes current expected behavior

test/extensions/INTEL/SPV_INTEL_rounded_divide_sqrt/rounding_supported.ll
Qi-Ye-079: Done

test/extensions/INTEL/SPV_INTEL_rounded_divide_sqrt/fconvert_unaffected.ll
Qi-Ye-079: Done. Made it concise.

test/extensions/INTEL/SPV_INTEL_rounded_divide_sqrt/rounding_supported.ll
Qi-Ye-079: Applied your suggestion.

test/extensions/INTEL/SPV_INTEL_rounded_divide_sqrt/rounding_supported.ll
Qi-Ye-079: Done.

test/extensions/INTEL/SPV_INTEL_rounded_divide_sqrt/rounding_supported.ll
Qi-Ye-079: Done. Checking `@_Z4sqrt*` for OCL and `@_Z16__spirv_ocl_sqrt*` for SPV.

test/extensions/INTEL/SPV_INTEL_rounded_divide_sqrt/fconvert_unaffected.ll
MrSidims: Thank you!

MrSidims:
lgtm (but my review is not gating)

test/extensions/INTEL/SPV_INTEL_rounded_divide_sqrt/rounding_supported.ll
```diff
@@ -0,0 +1,180 @@
...
+; CHECK-SPIRV-DAG: Decorate [[#D_RTN:]] FPRoundingMode 3
+; CHECK-SPIRV-DAG: FDiv [[#HALF]] [[#H_RTE]]
+; CHECK-SPIRV-DAG: FDiv [[#FLOAT]] [[#F_RTZ]]
+; CHECK-SPIRV-DAG: FDiv [[#DOUBLE]] [[#D_RTP]]
+; CHECK-SPIRV-DAG: FDiv [[#DOUBLE]] [[#D_RTN]]
+
+; In SPV-IR mode the FPRoundingMode decoration (Decoration 39) is preserved as
+; an !spirv.Decorations metadata node, captured per rounding mode at its first
+; use below and verified at the end. The same node is shared across every
+; fdiv/sqrt of that mode (scalar and vector alike).
```
vmaksimo:
```suggestion

```
Tip: over-explanation, can be removed (you anyway need to fix clang-format failure - can do this fixup as well)

vmaksimo: My bad, that was infra issue. You can skip this one

vmaksimo (approved):

PR #3771, opened 2026-06-04
Preventing generating FPRoundingMode for arithmetic constrained intrinsics

### LINKED ISSUE #3569: llvm.experimental.constrained intrinsics generate invalid SPIR-V
Currently, at least some of the [llvm.experimental.constrained](https://llvm.org/docs/LangRef.html#constrained-floating-point-intrinsics) intrinsics are generating invalid SPIR-V.  Specifically, they are adding the FPRoundingMode decoration to instructions that do not support this decoration, for the OpenCL SPIR-V environment at least, see:

https://registry.khronos.org/OpenCL/specs/3.0-unified/html/OpenCL_Env.html#_rounding_modes_for_conversions

These validation failures are not currently flagged by spirv-val, but they may be shortly, see:

https://github.com/KhronosGroup/SPIRV-Tools/pull/6537

What should we do in this case?  Some options:

1. Fail translation - there is nothing in SPIR-V we can translate to.
2. Drop the unsupported rounding mode decorations, potentially with a warning diagnostic.
3. Do nothing and keep generating invalid SPIR-V, in which case a few tests will need to be updated to stop running spirv-val.

Qi-Ye-079:
Fixes misuse of `FPRoundingMode`  on some unsupported intrinsics which violates the OpenCL SPIR-V spec:
https://registry.khronos.org/OpenCL/specs/unified/html/OpenCL_Env.html#_rounding_modes_for_conversions

Such misuse can now be detected by `spirv-val`.

Mirrors the [same fix](https://github.com/llvm/llvm-project/pull/182742/) in the SPIR-V backend in LLVM.

Resolves #3569

### DIFF
--- a/test/llvm-intrinsics/constrained-arithmetic.ll
+++ b/test/llvm-intrinsics/constrained-arithmetic.ll
@@ -1,5 +1,5 @@
 ; RUN: llvm-as %s -o %t.bc
-; RUN: llvm-spirv %t.bc -spirv-text -o - | FileCheck %s
+; RUN: llvm-spirv %t.bc -spirv-text -o - | FileCheck %s --implicit-check-not FPRoundingMode
 ; RUN: llvm-spirv %t.bc -o %t.spv
 ; RUN: spirv-val %t.spv
 
@@ -8,15 +8,6 @@
 ; CHECK: Name [[di:[0-9]+]] "div"
 ; CHECK: Name [[su:[0-9]+]] "sub"
 ; CHECK: Name [[mu:[0-9]+]] "mul"
-
-; CHECK-NOT: Decorate {{[0-9]+}} FPRoundingMode
-
-; CHECK-NOT: Decorate [[ad]] FPRoundingMode 0
-; CHECK-NOT: Decorate [[di]] FPRoundingMode 1
-; CHECK-NOT: Decorate [[su]] FPRoundingMode 2
-; CHECK-NOT: Decorate [[mu]] FPRoundingMode 3
-
-; CHECK-NOT: Decorate {{[0-9]+}} FPRoundingMode
 
 ; CHECK: FAdd {{[0-9]+}} [[ad]]
 ; CHECK: FDiv {{[0-9]+}} [[di]]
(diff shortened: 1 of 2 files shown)


test/llvm-intrinsics/constrained-arithmetic.ll
```diff
@@ -1,8 +1,7 @@
 ; RUN: llvm-as %s -o %t.bc
 ; RUN: llvm-spirv %t.bc -spirv-text -o - | FileCheck %s
```
vmaksimo: As we don't expected any `FPRoundingMode` decoration added, I suggest using this and remove all the checks for decorations
```suggestion
; RUN: llvm-spirv %t.bc -spirv-text -o - | FileCheck %s --implicit-check-not FPRoundingMode
```

vmaksimo (commented):
LGTM, thanks!
One minor thing to address and we're good to go

test/llvm-intrinsics/constrained-arithmetic.ll
Qi-Ye-079: Done.

vmaksimo (approved):

MrSidims:
Phew, since you have volunteered to fix this, may I ask you to answer some uncomfortable questions from @jmmartinez in https://github.com/llvm/llvm-project/pull/182742 ? :D

Basically, it seems like these intrinsics are not representable in SPIR-V. If so, may be we should just emit an error when see them in the module? (unless allow-unknown-intrinsic is passed). (I don't have a strong opinion myself tho, so any decision and justification is fine with me).

vmaksimo:
@MrSidims not perfect thing to solve for first-time contributor, huh? 😅

Note, we'd support the rounding mode for two intrinsics via `SPV_INTEL_rounded_div_sqrt` extension (see #3770)
So I'd prefer not to emit an error, but just drop the constrained logic as it's done today (emitting a warning about this might be a good idea though)

jmmartinez:
> @MrSidims not perfect thing to solve for first-time contributor, huh? 😅
> 
> Note, we'd support the rounding mode for two intrinsics via `SPV_INTEL_rounded_div_sqrt` extension (see #3770) So I'd prefer not to emit an error, but just drop the constrained logic as it's done today (emitting a warning about this might be a good idea though)

I'm ok with dropping the constraint + warning the user. It seems like a fair middle ground.

MrSidims (approved):
fine with me

PR #3772, opened 2026-06-06
Add SPV_NV_shader_atomic_fp16_vector extension

MrSidims:
It allowed fp16 atomics translation.

Spec:
https://github.com/KhronosGroup/SPIRV-Registry/blob/main/extensions/NV/SPV_NV_shader_atomic_fp16_vector.asciidoc


MrSidims:
@aobolensk @jmmartinez @AlexVlx please take a look

lib/SPIRV/libSPIRV/SPIRVInstruction.h
```diff
@@ -3135,6 +3135,9 @@ class SPIRVAtomicFAddEXTInst : public SPIRVAtomicInstBase {
 public:
   std::optional<ExtensionID> getRequiredExtension() const override {
     assert(hasType());
+    if (getType()->isTypeVector() &&
+        getType()->getVectorComponentType()->isTypeFloat(16))
```
aobolensk, line 3139: Q: Should that be applicable for both bf16 and fp16?

MrSidims: BF16 is not covered by the extension as it is
`This extension is written against the SPIR-V Specification, Version 1.6 Revision 3` - the time when bf16 (aka fp encoding) wasn't yet introduced in SPIR-V.
https://github.com/llvm/llvm-project/pull/202083 is a hack that is possible in SPIR-V backend as there is a clean guard around AMD target.

aobolensk (approved):

PR #3776, opened 2026-06-10
Use OpImageQuerySize for multisampled image size queries

aobolensk:
OpImageQuerySizeLod requires the image to have MS = 0, so size queries on multisampled images must use OpImageQuerySize without a LOD operand

Inspired by the change in LLVM SPIR-V backend: https://github.com/llvm/llvm-project/pull/190767


MrSidims:
Restarting CI with new HEAD

MrSidims (approved):

PR #3777, opened 2026-06-10
Use SPV_EXT_long_vector extension for non-standard vector size

Qi-Ye-079:
For a vector type, add `SPV_EXT_long_vector` extension and prioritize it over `SPV_INTEL_vector_compute` whenever the vector type doesn't have a standard component size (2, 3, 4, 8, or 16).


lib/SPIRV/SPIRVWriter.cpp
```diff
@@ -428,6 +428,15 @@ SPIRVType *LLVMToSPIRVBase::transType(Type *T) {
...
   if (auto *VecTy = dyn_cast<FixedVectorType>(T)) {
+    if (VecTy->getNumElements() == 1 &&
+        !BM->isAllowedToUseExtension(ExtensionID::SPV_EXT_long_vector)) {
+      BM->getErrorLog().checkError(false, SPIRVEC_RequiresExtension,
+                                   "SPV_EXT_long_vector\n"
+                                   "NOTE: LLVM module contains a 1-element "
+                                   "vector, translation of which requires "
+                                   "this extension");
+      return nullptr;
+    }
```
vmaksimo: Won't the translation just fail in `validate()` if we remove this block?

Qi-Ye-079: Yes, but (1) I'd like to make the failure more descriptive and (2) the assertion failure in `validate()` only happens if debug build (`NDEBUG` is undefined) as in the original code

MrSidims:
> keep using SPV_INTEL_vector_compute for other non-standard vector size at the moment.

Multi-vendor extension must always have bigger priority, than a vendor extension if they do the same.

Qi-Ye-079:
> > keep using SPV_INTEL_vector_compute for other non-standard vector size at the moment.
> 
> Multi-vendor extension must always have bigger priority, than a vendor extension if they do the same.

Yes, I'm prioritizing `SPV_EXT_long_vector` over intel one now.

lib/SPIRV/SPIRVWriter.cpp
Qi-Ye-079: If you still think the check is redundant or can cause regression, I'm ok to remove it

vmaksimo (approved):
LGTM
Nit: I'd add test case for a vector of arbitrary size (now we have only for 1-element vector.)

lib/SPIRV/SPIRVWriter.cpp
```diff
@@ -428,6 +428,20 @@ SPIRVType *LLVMToSPIRVBase::transType(Type *T) {
...
+    if (IsNonStandardCount &&
+        !BM->isAllowedToUseExtension(ExtensionID::SPV_EXT_long_vector) &&
+        !BM->isAllowedToUseExtension(ExtensionID::SPV_INTEL_vector_compute)) {
+      BM->getErrorLog().checkError(
+          false, SPIRVEC_RequiresExtension,
+          "SPV_EXT_long_vector or SPV_INTEL_vector_compute\n"
+          "NOTE: LLVM module contains a vector with an unsupported number of "
+          "components, translation of which requires one of these extensions");
+      return nullptr;
+    }
```
MrSidims, line 444: Why have we started to need this?

Qi-Ye-079: I've answered same the question for @vmaksimo : the same check in `SRIRVType::validate()` is done by `assert` enclosed by `#ifndef NDEBUG` block only; I thought adding this check should be more user friendly and make it easier for lit test.
Actually I think this piece of code should replace the assertion check in `SPIRVType::validate()` instead, but... I don't know if it can break anything or have any side effect by with such change.
If you think this is purely redundant or can cause regression, I'm ok to remove it.

svenvh:
> Nit: I'd add test case for a vector of arbitrary size (now we have only for 1-element vector.)

Yes, I was going to suggest the same.  I'd suggest to simply add another case to each test file.

Qi-Ye-079:
> > Nit: I'd add test case for a vector of arbitrary size (now we have only for 1-element vector.)
> 
> Yes, I was going to suggest the same. I'd suggest to simply add another case to each test file.

I agree adding another arbitrary size for each test, but since the emission of OpExtension is done per module, and there's no usage of `split-file` in this repo, I think it's better to add new files for a new arbitrary size, to ensure each different size actually emits the correct extension and capability (or report error).

vmaksimo:
>> keep using SPV_INTEL_vector_compute for other non-standard vector size at the moment.

> Multi-vendor extension must always have bigger priority, than a vendor extension if they do the same.

Note, we want to rely on SPV_INTEL_vector_compute for vector compute modules (ESIMD) even when SPV_EXT_long_vector is enabled. 
I suggest this PR stays community-valuable, and I will publish INTEL-only related change in a separate PR.

vmaksimo (approved):
I'd merge `vector_size_5,ll` and `negative_size_5.ll` into one test, same for size 1.
Other than that, LGTM

lib/SPIRV/SPIRVWriter.cpp
YuriPlyakhin: sounds reasonable to me. But I'm fine either way, I don't think it should block this PR.

test/extensions/EXT/SPV_EXT_long_vector/priority_over_vector_compute.ll
```diff
@@ -0,0 +1,14 @@
+; RUN: llvm-spirv %s --spirv-ext=+SPV_EXT_long_vector,+SPV_INTEL_vector_compute -o %t.spv
+; RUN: llvm-spirv %t.spv -to-text -o - | FileCheck %s
+
+; CHECK-DAG: Capability LongVectorEXT
+; CHECK-DAG: Extension "SPV_EXT_long_vector"
+; CHECK-NOT: Capability VectorAnyINTEL
```
YuriPlyakhin: nit: we can use `--implicit-check-not` to make this check stronger. Same feedback for other similar tests in this PR.

test/extensions/EXT/SPV_EXT_long_vector/priority_over_vector_compute.ll
```diff
@@ -0,0 +1,14 @@
+; RUN: llvm-spirv %s --spirv-ext=+SPV_EXT_long_vector,+SPV_INTEL_vector_compute -o %t.spv
+; RUN: llvm-spirv %t.spv -to-text -o - | FileCheck %s
```
YuriPlyakhin: nit: consider adding `spirv-val` if makes sense (here and other tests in this PR).

YuriPlyakhin (approved):
LGTM. Few nits. And another nit: I would merge some of those tests, e.g. 2 negative into single test, 2 priority tests into single test, two size 5 and size 1 tests into single test.

Qi-Ye-079:
> LGTM. Few nits. And another nit: I would merge some of those tests, e.g. 2 negative into single test, 2 priority tests into single test, two size 5 and size 1 tests into single test.

@YuriPlyakhin Thanks for the review. I actually split them into different files on purpose, see this comment I replied to Sven: https://github.com/KhronosGroup/SPIRV-LLVM-Translator/pull/3777#issuecomment-4800744477

test/extensions/EXT/SPV_EXT_long_vector/priority_over_vector_compute.ll
Qi-Ye-079: Done.

test/extensions/EXT/SPV_EXT_long_vector/priority_over_vector_compute.ll
Qi-Ye-079: Done.

Qi-Ye-079 (commented):
@vmaksimo @YuriPlyakhin Thanks, I've fixed all nits.

YuriPlyakhin (approved):

vmaksimo:
@svenvh @MrSidims would be nice if one of you'd be able to take a look so we don't merge it with Intel approvals only🙂

YuriPlyakhin:
Planning to merge tomorrow unless anyone objects. Friendly ping @svenvh / @MrSidims  / @maarquitos14

test/extensions/EXT/SPV_EXT_long_vector/priority_over_vector_compute_size_5.ll
maarquitos14, line 1: Can't this be merged into priority_over_vector_compute.ll, so both sizes are in the same file?

maarquitos14 (approved):
Just a nit, otherwise LGTM.

test/extensions/EXT/SPV_EXT_long_vector/priority_over_vector_compute_size_5.ll
YuriPlyakhin: thank you for review, @maarquitos14 , this question was addressed here: https://github.com/KhronosGroup/SPIRV-LLVM-Translator/pull/3777#issuecomment-4800744477

maarquitos14: Ah, thanks, didn't see that one.

PR #3778, opened 2026-06-12
Fix SPIR-V module assert when translating freeze without SPV_KHR_poison_freeze

lslusarczyk:
## Summary

When `SPV_KHR_poison_freeze` is disabled, `llvm-spirv` must not leave `llvm.freeze` in the IR while still treating it as a distinct SSA value during translation (for example via `OpName`). That mismatch led to an assert in `SPIRVModuleImpl::getEntry`. This change strips `freeze` in `SPIRVRegularizeLLVM` when the extension is not allowed (with `TranslatorOpts` passed into that pass) and removes the duplicate operand-reuse path from `SPIRVWriter`.

## Test plan

- [x] `llvm-lit test/extensions/KHR/SPV_KHR_poison_freeze/freeze-phi-loop.ll`
- [x] `llvm-lit test/extensions/KHR/SPV_KHR_poison_freeze/freeze.ll`


lslusarczyk:
@MrSidims , @vmaksimo , this PR fixes regression from some past PR you were working on. 
Could you please review? Could you please approve CI run?

MrSidims:
CI approved.

PR description is way too verbose, is the issue about debug info instruction referencing freeze instruction, which actually is deleted when the extension is not enabled leaving debug info referencing something that doesn't exist? If yes - please spell this out and remove overwhelming information about SYCL, Pytorch, test plan etc etc. A single paragraph about the bug is enough.

test/extensions/KHR/SPV_KHR_poison_freeze/freeze-many-nonsemantic-debug.ll
```diff
@@ -0,0 +1,119 @@
+; Regression for llvm-spirv assert in SPIRVModule::getEntry when translating
+; many freeze instructions with NonSemantic.Shader.DebugInfo.200 and
```
MrSidims: Per my understanding a single pair of freeze + debug info instructions is enough to test the bug.

test/extensions/KHR/SPV_KHR_poison_freeze/freeze-many-nonsemantic-debug.ll
```diff
@@ -0,0 +1,119 @@
...
+;
+; RUN: llvm-spirv %s -o %t.spv --spirv-debug-info-version=nonsemantic-shader-200
+; RUN: llvm-spirv %s -o %t.spv --spirv-ext=+SPV_KHR_poison_freeze --spirv-debug-info-version=nonsemantic-shader-200
+; RUN: llvm-spirv %t.spv -to-text -o %t.spt
+; RUN: FileCheck < %t.spt %s --check-prefix=CHECK-SPIRV
+
+target datalayout = "e-i64:64-v16:16-v24:32-v32:32-v48:64-v96:128-v192:256-v256:256-v512:512-v1024:1024-n8:16:32:64"
+target triple = "spir64-unknown-unknown"
+
+; CHECK-SPIRV: FreezeKHR
```
MrSidims: What about the case, when the extension is disabled?

lib/SPIRV/SPIRVRegularizeLLVM.cpp
```diff
@@ -721,6 +721,17 @@ bool SPIRVRegularizeLLVMBase::regularize() {
             BO->setIsExact(false);
         }
 
+        if (!Opts.isAllowedToUseExtension(ExtensionID::SPV_KHR_poison_freeze)) {
```
MrSidims, line 724: I don't have a strong opinion whether to fix this in SPIRVRegularizeLLVM or SPIRVWriter. But if we fix it here - it would be nice to remove:
```
    // Without the extension, move the freeze away.
    Value *Operand = U->getOperand(0);
    if (isa<UndefValue>(Operand))
      return BM->addNullConstant(transScavengedType(U));
    return transValue(Operand, BB);
```
from the Writer.

test/extensions/KHR/SPV_KHR_poison_freeze/freeze-phi-loop.ll
```diff
@@ -0,0 +1,38 @@
+; Regression for llvm-spirv assert in SPIRVModule::getEntry when translating a
+; freeze in a phi loop without SPV_KHR_poison_freeze (llvm-reduce from device
+; bitcode; debug info is not required).
+;
+; RUN: llvm-as < %s -o %t.bc
+; RUN: llvm-spirv %t.bc -o %t.noext.spv
```
vmaksimo:
```suggestion
; RUN: llvm-spirv %s -o %t.noext.spv
```

test/extensions/KHR/SPV_KHR_poison_freeze/freeze-phi-loop.ll
```diff
@@ -0,0 +1,38 @@
...
+; RUN: llvm-spirv %t.noext.spv -to-text -o %t.noext.spt
+; RUN: FileCheck < %t.noext.spt %s --check-prefix=CHECK-NOEXT
+;
+; RUN: llvm-spirv %t.bc -o %t.spv --spirv-ext=+SPV_KHR_poison_freeze
```
vmaksimo:
```suggestion
; RUN: llvm-spirv %s -o %t.spv --spirv-ext=+SPV_KHR_poison_freeze
```

lib/SPIRV/SPIRVRegularizeLLVM.cpp
vmaksimo: Please apply this suggestion from Dmitry

test/extensions/KHR/SPV_KHR_poison_freeze/freeze-many-nonsemantic-debug.ll
lslusarczyk: test was changed to simpler one, that new test already runs the extension is disabled and when it is not
confirmed it passes with the fix and fails without

lib/SPIRV/SPIRVRegularizeLLVM.cpp
lslusarczyk: done

lslusarczyk:
@vmaksimo, @MrSidims,
Thank you for the review comments. I've addressed them. Could you please re-approve CI, take another look, and merge if everything looks good?

vmaksimo (approved):

test/extensions/KHR/SPV_KHR_poison_freeze/freeze-phi-loop.ll
```diff
@@ -0,0 +1,37 @@
+; Regression for llvm-spirv assert in SPIRVModule::getEntry when translating a
+; freeze in a phi loop without SPV_KHR_poison_freeze (llvm-reduce from device
+; bitcode; debug info is not required).
+;
```
MrSidims: To remove?

MrSidims (approved):
lgtm with one nit

test/extensions/KHR/SPV_KHR_poison_freeze/freeze-phi-loop.ll
lslusarczyk: removed

PR #3779, opened 2026-06-15
[CI] Switch lit to internal shell

vmaksimo:
llvm/llvm-project@57109befac92 `"[lit] Deprecate execute_external=True in ShTest"` makes `execute_external=True` a fatal error. 
Our previous setting (`ShTest(not llvm_config.use_lit_shell)`) evaluated to `True` on Linux/macOS (positional argument is `execute_external`), so test discovery now fails with: `ValueError: execute_external=True is deprected as of LLVM-23 ...`

### DIFF
--- a/test/lit.cfg.py
+++ b/test/lit.cfg.py
@@ -13,9 +13,6 @@ from lit.llvm.subst import FindTool
 config.name = 'LLVM_SPIRV'
 
 # testFormat: The test format to use to interpret tests.
-# Use lit's internal shell. ShTest(execute_external=True) is rejected
-# starting with LLVM-23 (commit 57109befac92, "[lit] Deprecate
-# execute_external=True in ShTest").
 config.test_format = lit.formats.ShTest()
 
 # suffixes: A list of file extensions to treat as test files.
--- a/test/lit.cfg.py
+++ b/test/lit.cfg.py
@@ -13,7 +13,10 @@ from lit.llvm.subst import FindTool
 config.name = 'LLVM_SPIRV'
 
 # testFormat: The test format to use to interpret tests.
-config.test_format = lit.formats.ShTest(not llvm_config.use_lit_shell)
+# Use lit's internal shell. ShTest(execute_external=True) is rejected
+# starting with LLVM-23 (commit 57109befac92, "[lit] Deprecate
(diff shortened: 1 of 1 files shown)


test/lit.cfg.py
```diff
@@ -13,7 +13,10 @@
 config.name = 'LLVM_SPIRV'
 
 # testFormat: The test format to use to interpret tests.
-config.test_format = lit.formats.ShTest(not llvm_config.use_lit_shell)
+# Use lit's internal shell. ShTest(execute_external=True) is rejected
+# starting with LLVM-23 (commit 57109befac92, "[lit] Deprecate
+# execute_external=True in ShTest").
+config.test_format = lit.formats.ShTest()
```
vmaksimo: I wonder if the comment should be removed?
And should we now even set `test_format` explicitly, as the defailt is internal shell now?

svenvh: If tests still work after removing this `test_format` assignment, then I see no reason to keep it. :)

vmaksimo: Ah, quick experiment showed it won't work. 
Without the line, lit silently finds zero tests and exits with "did not discover any tests for provided path(s)". ShTest() invocation is needed to register the test format, even though execute_external=False is now the default.

svenvh: Thanks for trying, I'd suggest to just tidy up the comment then (there's no need to keep these historical options as comments; for that the git log will suffice).

MrSidims: Yeap, lets remove the comment

MrSidims (approved):

vmaksimo:
Seems this fixes only one problem, only Out-of-tree passes CI checks; 
If reviewers don't mind, I'd like to get this one merged, and address In-tree failures in separate follow-up PR(s):
```
Failed Tests (2):
  LLVM_SPIRV :: llvm-intrinsics/frexp.ll
  LLVM_SPIRV :: reqd_work_group_size_md.ll
```

svenvh (approved):

svenvh:
> If reviewers don't mind, I'd like to get this one merged, and address In-tree failures in separate follow-up PR(s):

Yes that sounds reasonable to me.

PR #3781, opened 2026-06-16
[DebugInfo] Distinguish absent and empty DIFile source on round-trip

vmaksimo:
`DebugSource` with no source text was translated into a DIFile with `source: ""` instead of no `source:` field at all. 
LLVM preserves the two distinctly: an empty source becomes `DW_AT_LLVM_source = "\n"` in DWARF, which GDB and other consumers prefer over the on-disk file, displaying a blank line at kernel breakpoints in place of the real source.

Round-trip an absent SPIR-V Source operand to an absent DIFile `source:` field.

### DIFF
--- a/lib/SPIRV/SPIRVToLLVMDbgTran.cpp
+++ b/lib/SPIRV/SPIRVToLLVMDbgTran.cpp
@@ -79,11 +79,14 @@ void SPIRVToLLVMDbgTran::addDbgInfoVersion() {
 DIFile *
 SPIRVToLLVMDbgTran::getDIFile(const std::string &FileName,
                               std::optional<DIFile::ChecksumInfo<StringRef>> CS,
-                              std::optional<StringRef> Source) {
+                              std::optional<std::string> Source) {
   return getOrInsert(FileMap, FileName, [this, FileName, CS, Source]() {
     SplitFileName Split(FileName);
-    // Use the first builder from the map to crete DIFile since it's
+    // Use the first builder from the map to create DIFile since it's
     // relations with other debug metadata is not going through DICompileUnit
+    // Pass nullopt (not an empty string) when the SPIR-V has no Source
+    // operand: LLVM/DWARF preserve "absent" and "present-but-empty"
+    // distinctly.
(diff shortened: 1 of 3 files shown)


lib/SPIRV/SPIRVToLLVMDbgTran.cpp
```diff
@@ -79,11 +79,14 @@ void SPIRVToLLVMDbgTran::addDbgInfoVersion() {
...
-                              std::optional<StringRef> Source) {
+                              std::optional<std::string> Source) {
   return getOrInsert(FileMap, FileName, [this, FileName, CS, Source]() {
     SplitFileName Split(FileName);
-    // Use the first builder from the map to crete DIFile since it's
+    // Use the first builder from the map to create DIFile since it's
     // relations with other debug metadata is not going through DICompileUnit
+    // Pass nullopt (not an empty string) when the SPIR-V has no Source
+    // operand: LLVM/DWARF preserve "absent" and "present-but-empty"
+    // distinctly.
```
MrSidims: IMHO few years from now this comment would be looking confusing. I suggest to not push meta-information from a conversation with an LLM.

MrSidims (approved):

lib/SPIRV/SPIRVToLLVMDbgTran.cpp
MrSidims: I mean, that now I understand what `(not an empty string)` means and that is related to StringRef to std::string change. But when I forget the context - this "clarification" will confuse me without bringing any benefits in understanding the code while reading it.

svenvh: Agreed; too often I see LLMs putting patch delta explanations as code comments, when that really belongs in the commit message only.

vmaksimo: Thanks folks, I focused on shortening the comments instead of reasoning if they are really needed 🫣

svenvh (approved):

PR #3785, opened 2026-06-18
Correctly manage NamedId to avoid dangling IDs

Qi-Ye-079:
Fix assertion failure "Id is not in map" when a `FreezeInst` result is a forward reference when `SPV_KHR_poison_freeze` isn't enabled. Added a minimal lit testcase as regression to cover this situation.

### DIFF
--- a/lib/SPIRV/libSPIRV/SPIRVModule.cpp
+++ b/lib/SPIRV/libSPIRV/SPIRVModule.cpp
@@ -1375,6 +1375,10 @@ SPIRVEntry *SPIRVModuleImpl::replaceForward(SPIRVForward *Forward,
     IdEntryMap.erase(Loc);
     Entry->setId(ForwardId);
     IdEntryMap[ForwardId] = Entry;
+    // Need explicit bookkeeping of named IDs here to make it consistent with
+    // IdEntryMap, otherwise in some cases the original Id may be dangling.
+    if (NamedId.erase(Id))
+      NamedId.insert(ForwardId);
     // Replace current Id with ForwardId in decorates.
     Entry->replaceTargetIdInDecorates(ForwardId);
   }
--- a/test/extensions/KHR/SPV_KHR_poison_freeze/freeze-phi-back-edge.ll
+++ b/test/extensions/KHR/SPV_KHR_poison_freeze/freeze-phi-back-edge.ll
@@ -1,0 +1,46 @@
+; Regression for assertion `Id is not in map' in SPIRVModuleImpl::getEntry
+; (https://github.com/intel/llvm/issues/22308): when SPV_KHR_poison_freeze
+; is disabled, the writer ignores `freeze` by reusing its operand without
+; rewriting IR users of the FreezeInst. If a phi back-edge references the
+; freeze result (the freeze is defined after the phi in program order),
+; the SPIR-V phi forward-references an unmapped id and the writer hits
+; the assert at module serialization. The freeze must be stripped (and
+; users rewritten) in regularize before SPIR-V emission.
+;
+; RUN: llvm-spirv %s -o %t.spv
+; RUN: spirv-val %t.spv
+; RUN: llvm-spirv %t.spv -to-text -o %t.spt
+; RUN: FileCheck < %t.spt %s --check-prefix=CHECK-NOEXT
+;
+; RUN: llvm-spirv %s -o %t.ext.spv --spirv-ext=+SPV_KHR_poison_freeze
+; RUN: spirv-val %t.ext.spv
+; RUN: llvm-spirv %t.ext.spv -to-text -o %t.ext.spt
+; RUN: FileCheck < %t.ext.spt %s --check-prefix=CHECK-EXT
+
+target datalayout = "e-i64:64-v16:16-v24:32-v32:32-v48:64-v96:128-v192:256-v256:256-v512:512-v1024:1024-n8:16:32:64"
+target triple = "spir64-unknown-unknown"
+
+; Without the extension, regularize strips the freeze from the IR and
+; rewrites its users (the phi below), so no FreezeKHR is emitted.
+; CHECK-NOEXT-NOT: Capability PoisonFreezeKHR
+; CHECK-NOEXT-NOT: Extension "SPV_KHR_poison_freeze"
+; CHECK-NOEXT-NOT: FreezeKHR
+
+; With the extension, the freeze survives as OpFreezeKHR and the phi's
+; back-edge operand references it, matching the source IR.
+; CHECK-EXT-DAG: Capability PoisonFreezeKHR
+; CHECK-EXT-DAG: Extension "SPV_KHR_poison_freeze"
+; CHECK-EXT: Phi [[#]] [[#PHI:]] [[#]] [[#]] [[#FR:]] [[#]]
+; CHECK-EXT: FreezeKHR [[#]] [[#FR]] [[#PHI]]
+
+define spir_func void @f(i64 %n) {
+entry:
+  br label %loop
+loop:
+  %prev = phi i64 [ 0, %entry ], [ %fr, %loop ]
+  %fr = freeze i64 %prev
+  %c = icmp slt i64 %fr, %n
+  br i1 %c, label %loop, label %exit
+exit:
+  ret void
+}
--- a/test/extensions/KHR/SPV_KHR_poison_freeze/freeze-phi-back-edge.ll
+++ b/test/extensions/KHR/SPV_KHR_poison_freeze/freeze-phi-back-edge.ll
@@ -1,11 +1,13 @@
-; Regression for assertion `Id is not in map' in SPIRVModuleImpl::getEntry
-; (https://github.com/intel/llvm/issues/22308): when SPV_KHR_poison_freeze
-; is disabled, the writer ignores `freeze` by reusing its operand without
-; rewriting IR users of the FreezeInst. If a phi back-edge references the
-; freeze result (the freeze is defined after the phi in program order),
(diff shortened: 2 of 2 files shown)


lib/SPIRV/libSPIRV/SPIRVModule.cpp
```diff
@@ -1375,6 +1375,10 @@ SPIRVEntry *SPIRVModuleImpl::replaceForward(SPIRVForward *Forward,
     IdEntryMap.erase(Loc);
     Entry->setId(ForwardId);
     IdEntryMap[ForwardId] = Entry;
+    // Need explicit bookkeeping of named IDs here to make it consistent with
+    // IdEntryMap, otherwise in some cases the original Id may be dangling.
+    if (NamedId.erase(Id))
+      NamedId.insert(ForwardId);
```
MrSidims, line 1381: Frankly, I don't understand why an `explicit bookkeeping of named IDs` is needed...

test/extensions/KHR/SPV_KHR_poison_freeze/freeze-phi-back-edge.ll
```diff
@@ -0,0 +1,46 @@
+; Regression for assertion `Id is not in map' in SPIRVModuleImpl::getEntry
+; (https://github.com/intel/llvm/issues/22308): when SPV_KHR_poison_freeze
+; is disabled, the writer ignores `freeze` by reusing its operand without
+; rewriting IR users of the FreezeInst. If a phi back-edge references the
+; freeze result (the freeze is defined after the phi in program order),
+; the SPIR-V phi forward-references an unmapped id and the writer hits
+; the assert at module serialization. The freeze must be stripped (and
+; users rewritten) in regularize before SPIR-V emission.
```
MrSidims: ... given that this comment says, that `The freeze must be stripped (and users rewritten) in regularize before SPIR-V emission`.

MrSidims (commented):
I feel like [this patch](https://github.com/KhronosGroup/SPIRV-LLVM-Translator/pull/3778) is fixing the same bug.

test/extensions/KHR/SPV_KHR_poison_freeze/freeze-phi-back-edge.ll
vmaksimo: Comment is an artifact from previous approach to solve this (with the Lukasz PR you mentioned)
Needs to be removed/rewritten.

lib/SPIRV/libSPIRV/SPIRVModule.cpp
vmaksimo: Yeah so it's we update `IdEntryMap` and decoration targets for the new id but don't touch `NamedId` - that will hit for the following example: 
```
loop:
   %prev = phi i64 [ 0, %entry ], [ %fr, %loop ]   ; uses %fr before it's defined
   %fr   = freeze i64 %prev                        ; resolved here
```

If the entry had a name, the old `Id` stays in `NamedId` while no longer mapping to any entry; when the module is being written, we iterate `NamedId` to write out the OpNames, `getEntry(old_Id)` asserts.

vmaksimo (commented):
> I feel like this patch is fixing the same bug.

You're right, it fixes the same but on the different level. 
3778 strips freeze in SPIRVRegularizeLLVM so the writer never sees forward references for it, while this one fixes the broader problem with the invalid ids.

Per my local experiments, we can meet the same crash with addrspacecast used as a phi operand, but not sure if it's real-world example.

UPD: The crash happens only when the forward-resolved value is further ignored by the writer (like freeze without the freeze extension, or addrspacecast GlobalDevice->Global without USM extension)

test/extensions/KHR/SPV_KHR_poison_freeze/freeze-phi-back-edge.ll
Qi-Ye-079: I have rewritten the comments.

test/extensions/KHR/SPV_KHR_poison_freeze/freeze-phi-back-edge.ll
```diff
@@ -0,0 +1,48 @@
+; When SPV_KHR_poison_freeze is disabled, the Writer won't translate `FreezeInst`
+; itself but its operand.
+; As a consequence, if a phi back-edge references the freeze result (a forward
+; reference at this point), the id of its operand will be replaced by
```
MrSidims, line 4: phi always uses forward references.
```suggestion
; As a consequence, if a phi back-edge references the freeze result, the id of its operand will be replaced by
```

test/extensions/KHR/SPV_KHR_poison_freeze/freeze-phi-back-edge.ll
```diff
@@ -0,0 +1,48 @@
...
+; the forward-reference-Id of `freeze`, but the original id of this operand
+; remains in NamedId set while the same id is gone in IdEntryMap, resulting in
+; assertion failure "Id is not in map" during getEntry() when writing the Module.
+; With explicit bookkeeping of NamedId in replaceForward(),
+; the assertion failure should never occur; this test serves as regression
+; to ensure exactly this.
```
MrSidims, line 10: Can we shorten the comment?

test/extensions/KHR/SPV_KHR_poison_freeze/freeze-phi-back-edge.ll
```diff
@@ -0,0 +1,48 @@
+; When SPV_KHR_poison_freeze is disabled, the Writer won't translate `FreezeInst`
```
MrSidims, line 1: For `freeze` instruction, I strongly believe that [this PR](https://github.com/KhronosGroup/SPIRV-LLVM-Translator/pull/3778) is _the fix_.

MrSidims (commented):
Why don't we see the same issue with `llvm.expect`? Its result may also be referenced by a phi node. And `llvm.expect` will be ignored, if `SPV_KHR_assume_expect` extension is not enabled.

I'm not a fan of this approach. I do believe, that instead of fixing a bug (note 'a' not 'the' because while intention is to "fix" `freeze` translation, it might "fix" some other yet unknown issues) on previous translation stage it just hides it under a carpet.

MrSidims:
> I'm not a fan of this approach. I do believe, that instead of fixing a bug (note 'a' not 'the' because while intention is to "fix" freeze translation, it might "fix" some other yet unknown issues) on previous translation stage it just hides it under a carpet.

And since we are all AI enjoyers, here is Opus 4.8 verdict. Disclaimer: I didn't not assess its output myself, so it's a plain copy-paste without my guidance and it might be a halucination. Please take it with a grain of salt:
**Prompt:**

> Please assess this PR: https://github.com/KhronosGroup/SPIRV-LLVM-Translator/pull/3785 . I strongly believe, that it hides a real and more dangerous bug, but don't have proves. You may either investigate it trying to find, when the code added breaks, or calm me down, saying, that everything is fine.

**Reply:**

> Your instinct is right. I found the deeper bug and proved it.
> 
> Verdict: the PR fixes a symptom, not the root cause
> 
> **Root cause**. In replaceForward (SPIRVModule.cpp:1364), the ForwardId != Id branch renames an already-existing entry: the phi was added with its own id P, then freeze (no extension) collapses to the phi and mapValue calls replaceForward, which does Entry->setId(ForwardId) — changing the phi's id from P to F and re-keying IdEntryMap. The problem: operands everywhere are stored as raw ids (Ops, a vector<SPIRVWord>, resolved lazily via getEntry/getValue). Anything that captured the phi's original id P before the collapse is now dangling. NamedId is just one such structure — the PR re-keys it and stops there.
> 
>  **Proof**. I built the translator at the PR's base, reproduced the assert, applied the PR patch, and rebuilt. The PR's own test passes. But this IR — where the phi result is also consumed by another instruction:
> 
> ```
>   loop:
>     %prev = phi i64 [ 0, %entry ], [ %fr, %loop ]
>     %x = add i64 %prev, 1      ; <-- captures the phi's original id P
>     %fr = freeze i64 %prev
>     ...
> ```
> 
>   still crashes with the PR applied, and from a different code path:
> 
> ```
>   ┌─────────────────────┬─────────────────────────────────────────────────┐
>   │                     │       Frame #12 (the caller of getEntry)        │
>   ├─────────────────────┼─────────────────────────────────────────────────┤
>   │ PR's test (NamedId) │ operator<<(ostream&, SPIRVModule&) → encodeName │
>   ├─────────────────────┼─────────────────────────────────────────────────┤
>   │ My test (operand)   │ SPIRVBinary::validate() → getValue(Ops[0])      │
>   └─────────────────────┴─────────────────────────────────────────────────┘
> ```
> The add's operand still holds P; replaceForward never rewrote it. Same IR with --spirv-ext=+SPV_KHR_poison_freeze translates fine (no collapse), confirming  the bug is the id-hijack itself.
> 
> Why it's more dangerous than it looks. getEntry only asserts on the missing id (SPIRVModule.cpp:998). Instruction encode() emits raw Ops directly  (SPIRVInstruction.h:361). So in a release/no-assert build there's no crash — the translator silently emits SPIR-V whose add references a now-undefined id  P: malformed output instead of a loud failure.
> 
> The proper fix is at the source: replaceForward shouldn't leave dangling ids at all. Either rewrite all operand references P→F when it renames an entry, or  (cleaner) avoid renaming the existing entry — keep the phi at P and point id F at it / merge the other direction. Patching NamedId only patches the one  consumer that happened to be in the failing test; decorates were already handled (replaceTargetIdInDecorates), but instruction operands, and any other id-keyed map, are not.
>

MrSidims (changes requested):

PR #3798, opened 2026-06-25
Replace decode-path asserts with runtime error checks

idubinov:
Replace debug-only asserts in the SPIR-V decode path with runtime error checks, so that malformed input modules are rejected gracefully instead of causing crashes (null-deref / type-confusion) in release builds.

PR #3799, opened 2026-06-25
Bounds-check decoded operand/element counts before indexing

idubinov:
### Summary
Counts derived from the stream (operand-vector size, pair count, MemoryAccess size,
member index, OpWords size) were `assert`ed and then immediately used to index. Under
`-DNDEBUG` the guard vanishes, allowing an out-of-bounds read.

### Changes
- 19 sites guarded with runtime checks before the indexing operation.
- **Off-by-one bug fixed:** `SPIRVDecorate.cpp` `getLiteral` used `assert(I <= size())`
  which permitted `Literals[size()]` (one-past-end). Corrected to a strict `< size()`
  bounds check.
- `memoryAccessUpdate` (SPIRVInstruction.h 411/415/420/431), the sole guards for
  real attacker-reachable OOB reads, now enforce bounds at runtime.
- Files: SPIRVInstruction.h, SPIRVEntry.cpp, SPIRVFnVar.cpp, SPIRVType.h, SPIRVDecorate.cpp.

### Notes
- Where `getErrorLog()` is not in scope (SPIRVMemoryAccess, free functions) the existing
  local early-return / `ErrMsg` idiom is used so the bound is genuinely enforced in release.

### DIFF
--- a/lib/SPIRV/libSPIRV/SPIRVDecorate.cpp
+++ b/lib/SPIRV/libSPIRV/SPIRVDecorate.cpp
@@ -98,7 +98,13 @@ SPIRVDecorateGeneric::SPIRVDecorateGeneric(Op OC)
 Decoration SPIRVDecorateGeneric::getDecorateKind() const { return Dec; }
 
 SPIRVWord SPIRVDecorateGeneric::getLiteral(size_t I) const {
-  assert(I <= Literals.size() && "Out of bounds");
+  // SEC-00717 G2: Literals is sized from the decoded decoration words. The
+  // original assert used "<=" (off-by-one) which permitted Literals[size()].
+  // Use a strict "<" runtime bound and avoid the out-of-bounds read.
+  SPIRVCK(I < Literals.size(), InvalidWordCount,
+          "Decoration literal index out of bounds");
+  if (I >= Literals.size())
+    return 0;
   return Literals[I];
 }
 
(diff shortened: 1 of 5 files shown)


idubinov:
@MrSidims @aobolensk @maarquitos14 @vmaksimo

lib/SPIRV/libSPIRV/SPIRVDecorate.cpp
```diff
@@ -98,7 +98,13 @@ SPIRVDecorateGeneric::SPIRVDecorateGeneric(Op OC)
 Decoration SPIRVDecorateGeneric::getDecorateKind() const { return Dec; }
 
 SPIRVWord SPIRVDecorateGeneric::getLiteral(size_t I) const {
-  assert(I <= Literals.size() && "Out of bounds");
+  // SEC-00717 G2: Literals is sized from the decoded decoration words. The
+  // original assert used "<=" (off-by-one) which permitted Literals[size()].
+  // Use a strict "<" runtime bound and avoid the out-of-bounds read.
```
maarquitos14, line 103: I don't think the comment adds a lot of value.

lib/SPIRV/libSPIRV/SPIRVDecorate.cpp
```diff
@@ -98,7 +98,13 @@ SPIRVDecorateGeneric::SPIRVDecorateGeneric(Op OC)
...
+  SPIRVCK(I < Literals.size(), InvalidWordCount,
+          "Decoration literal index out of bounds");
+  if (I >= Literals.size())
+    return 0;
```
maarquitos14, line 107: Isn't this misleading? We're returning a proper value for something that is illegal.

maarquitos14 (commented):
Like I said in #3798, before going into more detail, I'm not sure all of these need to be converted from assert to proper errors. The asserts are often added to check preconditions or assumptions that will help debugging, but don't necessarily mean that an error should be reported at the same exact place. It could very well be that down the line there will be a proper error already. Could we check that we can indeed exercise crashes/undefined behaviours on those asserts before jumping straight into replacing them with errors?

Additionally, for specific bugfixes we should have tests that should fail without the fix, and pass with the fix.

PR #3800, opened 2026-06-25
Validate decoded operand/result types before unsafe static_cast

idubinov:
### Summary
`validate()` runs on the read path and checks the types of operands resolved from
decoded IDs (`isTypePointer()`, `isTypeBool()`, `isTypeVector()`, element-type/bitwidth
equality, ...). Many sit directly before an unchecked `static_cast<SPIRVType*>` or a
`getPointerElementType()` deref. Compiled out under `-DNDEBUG`, this allows a wrong-type
cast and deref (type confusion).

### Changes
- 54 type-predicate sites converted to `SPIRVCK(..., InvalidInstruction/InvalidFunctionCall)`
  before the cast/use (largest group; dominated by arithmetic / matrix / composite /
  copy `validate()` bodies in SPIRVInstruction.h).
- Where the assert directly guards a `static_cast` (e.g. SPIRVDecorate group members),
  the conversion also skips the unsafe cast on failure (`if (!SPIRVCK(...)) continue;`).
- Files: SPIRVInstruction.h, SPIRVType.h/.cpp, SPIRVValue.h, SPIRVDecorate.cpp.

### Notes
- `validate()` returns void here, so `SPIRVCK` (log + continue) is used, not `SPIRVCKRT`.
- Pure type/pointer identity equality asserts (false compare gives a wrong-but-safe
  result) left as asserts per analysis.

### DIFF
--- a/lib/SPIRV/libSPIRV/SPIRVDecorate.cpp
+++ b/lib/SPIRV/libSPIRV/SPIRVDecorate.cpp
@@ -257,7 +257,9 @@ void SPIRVGroupDecorate::decorateTargets() {
   for (auto &I : Targets) {
     auto *Target = getOrCreate(I);
     for (auto &Dec : DecorationGroup->getDecorations()) {
-      assert(Dec->isDecorate());
+      if (!SPIRVCK(Dec->isDecorate(), InvalidInstruction,
+                   "Expected a Decorate in the decoration group"))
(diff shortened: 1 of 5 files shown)


idubinov:
@MrSidims @aobolensk @maarquitos14 @vmaksimo

PR #3802, opened 2026-06-25
Validate decoded WordCount consistency at runtime

idubinov:
### Summary
`assert(WordCount == FixedWC + ...)` / `WordCount > FixedWC` / `NumWords >= 1` and sizes
of vectors resized from the decoded word count. `WordCount` comes straight from the
stream header; compiled out, wrong-sized buffers are used downstream.

### Changes
- 35 sites converted to `SPIRVCK(..., InvalidWordCount)`, the exact pattern already
  used by the `setWordCount` overrides and SPIRVType.h:676/799/880.
- Includes SPIRVInstruction.h:251 `WordCount > 0` guard that feeds `WordCount - 1`
  (prevents unsigned underflow).
- Files: SPIRVInstruction.h, SPIRVType.h, SPIRVDecorate.h, SPIRVValue.h, SPIRVEntry.cpp,
  SPIRVAsm.h, SPIRVFnVar.h, SPIRVStream.cpp.

### DIFF
--- a/lib/SPIRV/libSPIRV/SPIRVAsm.h
+++ b/lib/SPIRV/libSPIRV/SPIRVAsm.h
@@ -46,7 +46,7 @@ public:
 protected:
   void validate() const override {
     SPIRVEntry::validate();
-    assert(WordCount > FixedWC);
+    SPIRVCK(WordCount > FixedWC, InvalidWordCount, "");
     assert(OpCode == OC);
   }
   _SPIRV_DEF_ENCDEC2(Id, Target)
@@ -86,7 +86,7 @@ protected:
   _SPIRV_DEF_ENCDEC6(Type, Id, FunctionType, Target, Instructions, Constraints)
   void validate() const override {
     SPIRVValue::validate();
-    assert(WordCount > FixedWC);
+    SPIRVCK(WordCount > FixedWC, InvalidWordCount, "");
     assert(OpCode == OC);
   }
   SPIRVAsmTargetINTEL *Target = nullptr;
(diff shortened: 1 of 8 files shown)


MrSidims (commented):
@idubinov @maarquitos14 @aobolensk 

I'm not sure, if any of the recently created PRs are needed. (with exception of truncation checks - this might be real issue).

Do we have data on how much it slows down the translation? If it's not bad - per instruction check won't harm. Otherwise I'd like us to consider other approaches (some of them are actually already implemented). 

Per my understanding, the intention is to prevent OOB. OOB on a single frame (aka instruction) is not critical on its own, a vector of a possible attack is typically looking like: get OOB outside the processed malformed SPIR-V module - and for this we have the whole module check added in https://github.com/KhronosGroup/SPIRV-LLVM-Translator/pull/3266 . Was the patch missing something or am I missing something?

From architecture stand-point, if we want real validation of each instruction we should instead introduce a dependency on SPIR-V tools enforcing spirv-val before performing translation. But it requires few fixed in the translator first.

Also, if we want full security - then we (each vendor) should check how exactly the translator is used in our products and add harness by always spawning de-priveleged process for the translation even for KMD. It's slow, but my gut feeling it's not much slower, than checking each instruction.

svenvh:
> I'm not sure, if any of the recently created PRs are needed.

Seconding that.  The asserts and `validate` functions are nothing more than a debug aid; they are likely to be compiled out of releases anyway.  We do not want to implement full validation here, as that would just become a (likely inferior) duplication of the SPIRV-Tools validator.

As @MrSidims already suggested, for proper input validation we could run the SPIRV-Tools validator upfront, perhaps as an optional feature initially.

PR #561, opened 2020-06-01
[SPIR-V] Exchange abort() to exit() in llvm-spirv tool by default.

fadeeval:
The llvm-spirv tool might be used (and actually used in intel/llvm) by some custom clang toolchains
and handling aborts from executed tools might be tricky and requires additional work.
Regular non-zero exit codes are just better for error reporting in production.

Signed-off-by: Aleksander Fadeev <aleksander.fadeev@intel.com>

PR #3817, opened 2026-06-29
Validate undefined Ids and builtin set ids in release builds

idubinov:
Malformed input runs an unchecked static_cast on objects leading to type confusion/null-pointer deref and out-of-bound reads that can leak heap memory into the translated output.

This change adds checks and rejection of module, closing the leak and the type-confusion.

### DIFF
--- a/lib/SPIRV/SPIRVReader.cpp
+++ b/lib/SPIRV/SPIRVReader.cpp
@@ -356,6 +356,11 @@ Type *SPIRVToLLVM::transType(SPIRVType *T, bool UseTPT) {
     // The length might be an OpSpecConstantOp, that needs to be specialized
     // and evaluated before the LLVM ArrayType can be constructed.
     auto *LenExpr = static_cast<const SPIRVTypeArray *>(T)->getLength();
+    if (!BM->getErrorLog().checkError(
+            LenExpr && isConstantOpCode(LenExpr->getOpCode()),
+            SPIRVEC_InvalidModule,
+            "OpTypeArray length must reference an integer constant"))
+      return nullptr;
     auto *LenValue = cast<ConstantInt>(transValue(LenExpr, nullptr, nullptr));
     return mapType(T, ArrayType::get(transType(T->getArrayElementType()),
                                      LenValue->getZExtValue()));
--- a/lib/SPIRV/libSPIRV/SPIRVModule.cpp
+++ b/lib/SPIRV/libSPIRV/SPIRVModule.cpp
@@ -1001,7 +1001,8 @@ SPIRVEntry *SPIRVModuleImpl::getEntry(SPIRVId Id) const {
 
 SPIRVExtInstSetKind SPIRVModuleImpl::getBuiltinSet(SPIRVId SetId) const {
   auto Loc = IdToInstSetMap.find(SetId);
-  assert(Loc != IdToInstSetMap.end() && "Invalid builtin set id");
+  if (Loc == IdToInstSetMap.end())
+    return SPIRVEIS_Count;
   return Loc->second;
 }
 
--- a/lib/SPIRV/libSPIRV/SPIRVStream.h
+++ b/lib/SPIRV/libSPIRV/SPIRVStream.h
@@ -41,6 +41,7 @@
 #define SPIRV_LIBSPIRV_SPIRVSTREAM_H
 
 #include "SPIRVDebug.h"
+#include "SPIRVError.h"
 #include "SPIRVExtInst.h"
 #include "SPIRVModule.h"
 #include <cctype>
@@ -152,6 +153,11 @@ const SPIRVDecoder &operator>>(const SPIRVDecoder &I, T *&P) {
   SPIRVId Id;
   I >> Id;
   P = static_cast<T *>(I.M.getEntry(Id));
+  if (!P) {
+    I.M.getErrorLog().checkError(false, SPIRVEC_InvalidModule,
+                                 "operand references an undefined Id");
+    I.M.setInvalid();
+  }
   return I;
 }
 
--- a/lib/SPIRV/libSPIRV/SPIRVType.cpp
+++ b/lib/SPIRV/libSPIRV/SPIRVType.cpp
@@ -54,9 +54,15 @@ SPIRVType *SPIRVType::getArrayElementType() const {
 uint64_t SPIRVType::getArrayLength() const {
   assert(OpCode == OpTypeArray && "Not array type");
   const SPIRVTypeArray *AsArray = static_cast<const SPIRVTypeArray *>(this);
-  assert(AsArray->getLength()->getOpCode() == OpConstant &&
-         "getArrayLength can only be called with constant array lengths");
-  return static_cast<SPIRVConstant *>(AsArray->getLength())->getZExtIntValue();
+  SPIRVValue *Len = AsArray->getLength();
+  if (!Len || !isConstantOpCode(Len->getOpCode())) {
+    getErrorLog().checkError(false, SPIRVEC_InvalidModule,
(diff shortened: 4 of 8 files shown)


idubinov:
@aobolensk @maarquitos14 @MrSidims @vmaksimo

lib/SPIRV/SPIRVReader.cpp
```diff
@@ -356,6 +356,11 @@ Type *SPIRVToLLVM::transType(SPIRVType *T, bool UseTPT) {
     // The length might be an OpSpecConstantOp, that needs to be specialized
     // and evaluated before the LLVM ArrayType can be constructed.
     auto *LenExpr = static_cast<const SPIRVTypeArray *>(T)->getLength();
+    if (!BM->getErrorLog().checkError(
+            LenExpr && isConstantOpCode(LenExpr->getOpCode()),
+            SPIRVEC_InvalidModule,
+            "OpTypeArray length must reference an integer constant"))
```
MrSidims, line 362: this should be an assertion imho

lib/SPIRV/libSPIRV/SPIRVType.cpp
```diff
@@ -54,9 +54,15 @@ SPIRVType *SPIRVType::getArrayElementType() const {
 uint64_t SPIRVType::getArrayLength() const {
   assert(OpCode == OpTypeArray && "Not array type");
   const SPIRVTypeArray *AsArray = static_cast<const SPIRVTypeArray *>(this);
-  assert(AsArray->getLength()->getOpCode() == OpConstant &&
-         "getArrayLength can only be called with constant array lengths");
-  return static_cast<SPIRVConstant *>(AsArray->getLength())->getZExtIntValue();
+  SPIRVValue *Len = AsArray->getLength();
+  if (!Len || !isConstantOpCode(Len->getOpCode())) {
+    getErrorLog().checkError(false, SPIRVEC_InvalidModule,
+                             "OpTypeArray length must reference a constant");
```
MrSidims, line 60: this should be an assertion imho

lib/SPIRV/libSPIRV/SPIRVStream.h
```diff
@@ -152,6 +153,11 @@ const SPIRVDecoder &operator>>(const SPIRVDecoder &I, T *&P) {
   SPIRVId Id;
   I >> Id;
   P = static_cast<T *>(I.M.getEntry(Id));
+  if (!P) {
+    I.M.getErrorLog().checkError(false, SPIRVEC_InvalidModule,
+                                 "operand references an undefined Id");
+    I.M.setInvalid();
+  }
```
MrSidims, line 160: The check belongs to getEntry implementation

MrSidims (commented):
@idubinov I believe the PR should be re-opened.

Please humanize the description. May be it's just me. but I've re-read it several times and didn't understood it. Meanwhile guarding `getEntry` result seem to fix a potential and real problem.

lib/SPIRV/libSPIRV/SPIRVType.cpp
idubinov: this check should be runtime check because it prevents potential uncontrollable data read

lib/SPIRV/SPIRVReader.cpp
idubinov: this runtime check prevents OOB data read

lib/SPIRV/libSPIRV/SPIRVStream.h
idubinov: Disagree. The main idea of this code is to label the module invalid in case of try to get incorrect Entry. The getter should not change state of class. I also think that both asserts inside getEntry function were placed by mistake - it is normal behavior for getter to return nullptr in case of no item found

idubinov:
@MrSidims ,

> May be it's just me 

Me too

lib/SPIRV/SPIRVReader.cpp
MrSidims: How?

lib/SPIRV/libSPIRV/SPIRVType.cpp
MrSidims: How?

lib/SPIRV/libSPIRV/SPIRVStream.h
MrSidims: It's not a getter.

lib/SPIRV/libSPIRV/SPIRVModule.cpp
```diff
@@ -1001,7 +1001,8 @@ SPIRVEntry *SPIRVModuleImpl::getEntry(SPIRVId Id) const {
 
 SPIRVExtInstSetKind SPIRVModuleImpl::getBuiltinSet(SPIRVId SetId) const {
   auto Loc = IdToInstSetMap.find(SetId);
-  assert(Loc != IdToInstSetMap.end() && "Invalid builtin set id");
```
maarquitos14, line 1004: I'm happy to keep the assert too for debugging purposes if this is something that should not happen. To be clear, I mean keeping the assert additionally to the new changes you added. The new changes should be absolutely kept.

lib/SPIRV/libSPIRV/SPIRVStream.h
maarquitos14: @idubinov is it really meaningful to mark the module invalid? Are we not terminating the process with `I.M.getErrorLog().checkError(...)`?

lib/SPIRV/libSPIRV/SPIRVType.cpp
maarquitos14: @idubinov do you mean it could happen within the translator? I don't see any calls to `getArrayLength()` that could be dangerous. Am I missing anything?

lib/SPIRV/SPIRVReader.cpp
maarquitos14: I also fail to see how this can result into OOB data read. A test case showing it would be helpful.

lib/SPIRV/libSPIRV/SPIRVStream.h
MrSidims: > Disagree. The main idea of this code is to label the module invalid in case of try to get incorrect Entry. The getter should not change state of class. I also think that both asserts inside getEntry function were placed by mistake - it is normal behavior for getter to return nullptr in case of no item found

This contradicts your other patch in https://github.com/KhronosGroup/SPIRV-LLVM-Translator/pull/3798/changes#r3544542206 which adds getEntry error out.

idubinov: @maarquitos14 , it can be aborted, asserted, or ignored, so, in general, the app is not ends in case of check fail
SPIRVError.h:156

maarquitos14: Theoretically, yes, but what I've seen is that it's set to `SPIRVDbgErrorHandlingKinds::Exit`, and there is no mechanism to change that value. I'm not sure if any downstream tool would ever use that, otherwise we could even think of removing the other two options, since they're not used anywhere. Obviously, that would be a different patch.

idubinov: @maarquitos14 SPIRVDbgErrorHandlingKinds::Exit is set in this project. but was added for translator to be able to be used by "some custom clang toolchains" (#561). So, I expect that in some custom toolchains the SPIRVDbgError may have different value and checkError may have different behavior

maarquitos14: Thanks for the explanation and the reference.

PR #3821, opened 2026-06-30
Propagate intrinsic param attributes to generated helper functions

aobolensk:
Inspired by https://github.com/llvm/llvm-project/pull/206061


test/llvm-intrinsics/memset-param-attrs.ll
```diff
@@ -0,0 +1,26 @@
+; Verify that captures(none) on the @llvm.memset destination parameter is
```
MrSidims, line 1: We should probably also check, what LLVM IR is generated after the roundtrip, do we still have the attribute?

MrSidims (approved):

test/llvm-intrinsics/memset-param-attrs.ll
aobolensk: Added checks

MrSidims:
Restarting CI with new HEAD

PR #3823, opened 2026-06-30
Fix invalid constant composite of untyped pointer variables

vmaksimo:
A module-scope variable is not a constant and cannot be a constituent of a constant composite.
For untyped pointers, a constant array or struct of pointers-to-globals was emitted as an `OpConstantComposite` of the bare `OpUntypedVariableKHR` operands, which is invalid SPIR-V.

Wrap a variable constituent so the composite refers to a constant.

AI-assisted: Claude Opus 4.8 (commercial SaaS)

### DIFF
--- a/test/extensions/KHR/SPV_KHR_untyped_pointers/globals.ll
+++ b/test/extensions/KHR/SPV_KHR_untyped_pointers/globals.ll
@@ -41,7 +41,12 @@ target triple = "spir-unknown-unknown"
 ; CHECK-SPIRV: UntypedVariableKHR [[#PTRTY]] [[#VARB:]] 5 [[#I32]]
 ; CHECK-SPIRV: UntypedVariableKHR [[#PTRTY]] [[#VARC:]] 5 [[#PTRTY]] [[#VARA]]
 ; CHECK-SPIRV: UntypedVariableKHR [[#LOCALPTRTY]] [[#VARD:]] 4 [[#PTRTY]]
-; CHECK-SPIRV: Variable [[#ARRAYPTRTY]] [[#VARE:]] 5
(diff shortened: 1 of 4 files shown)


test/extensions/KHR/SPV_KHR_untyped_pointers/globals.ll
```diff
@@ -41,7 +41,12 @@ target triple = "spir-unknown-unknown"
 ; CHECK-SPIRV: UntypedVariableKHR [[#PTRTY]] [[#VARB:]] 5 [[#I32]]
 ; CHECK-SPIRV: UntypedVariableKHR [[#PTRTY]] [[#VARC:]] 5 [[#PTRTY]] [[#VARA]]
 ; CHECK-SPIRV: UntypedVariableKHR [[#LOCALPTRTY]] [[#VARD:]] 4 [[#PTRTY]]
-; CHECK-SPIRV: Variable [[#ARRAYPTRTY]] [[#VARE:]] 5
```
vmaksimo, line 44: Before this fix, `@e`'s array initializer was emitted as a constant composite of the bare variables (invalid per SPIR-V core spec):
```
ConstantComposite %arrTy %e_init %a %b
Variable %arrPtrTy %e 5 %e_init
```

MrSidims:
@AlexVlx FYI, if I recall correctly you wanted to do something similar

AlexVlx:
> @AlexVlx FYI, if I recall correctly you wanted to do something similar

It was one option, yes. For what it's worth, I'd argue that as currently defined `OpVariable` is actually "constant", since it's the address (pointer to) of a global, rather than a global value, and the address is constant for the duration of a program's execution, so we should probably fix the spec.

PR #3826, opened 2026-07-01
Add storage class memory semantics bit to LLVM IR atomic translation

aobolensk:
Deduce storage class memory semantic from the pointer accessed by atomic intrinsics

Inspired by https://github.com/llvm/llvm-project/pull/193696, https://github.com/llvm/llvm-project/pull/195049

### DIFF
--- a/lib/SPIRV/OCLToSPIRV.cpp
+++ b/lib/SPIRV/OCLToSPIRV.cpp
@@ -69,87 +69,6 @@ using namespace OCLUtil;
 
 namespace SPIRV {
 
-static unsigned getAddressSpaceFromType(const Type *Ty) {
-  assert(Ty && "Can't deduce pointer AS");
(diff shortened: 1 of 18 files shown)


MrSidims:
> Per SPIR-V spec section 3.32 (Memory Semantics), atomic operations must combine the ordering bits ... with the relevant storage class bit ...

The description is misleading. There is no such spec requirement. Instead we should tell: deduce storage class memory semantic from the pointer accessed by atomic intrinsics

MrSidims (approved):

aobolensk:
> > Per SPIR-V spec section 3.32 (Memory Semantics), atomic operations must combine the ordering bits ... with the relevant storage class bit ...
> 
> The description is misleading. There is no such spec requirement. Instead we should tell: deduce storage class memory semantic from the pointer accessed by atomic intrinsics

Modified description

PR #3829, opened 2026-07-02
Keep VectorAnyINTEL for VectorCompute modules with SPV_EXT_long_vector

vmaksimo:
#3777 added `SPV_EXT_long_vector` and gave it priority over `SPV_INTEL_vector_compute`: when both extensions are enabled and a vector has a non-standard component count, the translator now emits `CapabilityLongVectorEXT` instead of `CapabilityVectorAnyINTEL`.

This regresses `VectorCompute` modules. The offloading flow (producing both `VectorCompute` and non-`VectorCompute` modules) applies a single, fixed extension list uniformly to every device image, so a `VectorCompute` module is translated with both `SPV_INTEL_vector_compute` and `SPV_EXT_long_vector` enabled. Before #3777 such a module emitted `VectorAnyINTEL` for its non-standard vectors; after #3777 it emits `LongVectorEXT`. `VectorCompute` consumers require `VectorAnyINTEL`, so this is a functional regression.

Hence, the enabled extension list does not identify a module as `VectorCompute` - the `VectorCompute` metadata (`VCFunction` /`VCGlobalVariable`) does. `IsVectorComputeModule` records the presence of that metadata before type translation. `VectorCompute` modules keep using `VectorAnyINTEL`, restoring the pre-#3777 behavior; all other modules still prefer the multi-vendor `LongVectorEXT` when its extension is enabled.


Qi-Ye-079:
@YuriPlyakhin Rebased and ready, please take a look. (I don't have permission to mark someone else's draft PR as ready)

lib/SPIRV/libSPIRV/SPIRVModule.h
```diff
@@ -196,6 +196,8 @@ class SPIRVModule {
   virtual void setAutoAddCapability(bool E) { AutoAddCapability = E; }
   virtual void setValidateCapability(bool E) { ValidateCapability = E; }
   virtual void setAutoAddExtensions(bool E) { AutoAddExtensions = E; }
+  virtual void setVectorComputeModule(bool E) { IsVectorComputeModule = E; }
+  bool isVectorComputeModule() const { return IsVectorComputeModule; }
```
YuriPlyakhin: `isVectorComputeModule` is not a module changing function. It seems it needs to be placed somewhere else in the file.
I think: `// Module query functions` above

lib/SPIRV/libSPIRV/SPIRVModule.h
```diff
@@ -196,6 +196,8 @@ class SPIRVModule {
   virtual void setAutoAddCapability(bool E) { AutoAddCapability = E; }
   virtual void setValidateCapability(bool E) { ValidateCapability = E; }
   virtual void setAutoAddExtensions(bool E) { AutoAddExtensions = E; }
+  virtual void setVectorComputeModule(bool E) { IsVectorComputeModule = E; }
```
YuriPlyakhin: Not sure why we make it virtual. It is a simple setter. I have the same question for other similar setters, but that's a legacy code.

```suggestion
  void setVectorComputeModule(bool E) { IsVectorComputeModule = E; }
```

I think the good example to follow is this one (in terms of inheritance design):
```
void setInvalid() { IsValid = false; }
bool isModuleValid() { return IsValid; }
```

lib/SPIRV/libSPIRV/SPIRVModule.h
```diff
@@ -679,6 +681,7 @@ class SPIRVModule {
   bool AutoAddCapability;
   bool ValidateCapability;
   bool AutoAddExtensions = true;
+  bool IsVectorComputeModule = false;
```
YuriPlyakhin: not sure why it is protected. Let's move it to private close to `bool IsValid;`

lib/SPIRV/libSPIRV/SPIRVModule.h
YuriPlyakhin:
```suggestion
  virtual void setVectorCompute(bool E) { IsVectorCompute = E; }
```

lib/SPIRV/libSPIRV/SPIRVModule.h
YuriPlyakhin:
```suggestion
  bool isVectorCompute() const { return IsVectorCompute; }
```

lib/SPIRV/libSPIRV/SPIRVModule.h
YuriPlyakhin:
```suggestion
  bool IsVectorCompute = false;
```

lib/SPIRV/libSPIRV/SPIRVModule.h
MrSidims: While I agree, I'd prefer not to change pre-existing code style in a functional commit (aka move de-virtualization in a separate PR).

test/extensions/INTEL/SPV_INTEL_vector_compute/prioritize_over_LongVecEXT_func.ll
```diff
@@ -0,0 +1,20 @@
+; RUN: llvm-spirv %s --spirv-ext=+SPV_EXT_long_vector,+SPV_INTEL_vector_compute -o %t.spv
```
MrSidims, line 5: All of the tests are doing the same thing per PR purpose. Keep 1.

test/extensions/INTEL/SPV_INTEL_vector_compute/prioritize_over_LongVecEXT_func.ll
```diff
@@ -0,0 +1,20 @@
...
+; RUN: llvm-spirv %t.spv -to-text -o - | FileCheck %s --implicit-check-not="Capability LongVectorEXT"
+; TODO: re-enable spirv-val once it can recognize that TypeVector is allowed to have any number of
+; components when SPV_INTEL_vector_compute is enabled.
+; RUNx: spirv-val %t.spv
+
+; CHECK-DAG: Capability VectorAnyINTEL
+; CHECK-DAG: Extension "SPV_INTEL_vector_compute"
```
MrSidims, line 9: Should we still check for long vector ext/cap?

lib/SPIRV/libSPIRV/SPIRVModule.h
MrSidims: So, the logic is:

>   IF SPV_EXT_long_vector is enabled
>   AND IF SPV_INTEL_vector_compute is enabled
>   AND IF there is a strange vector
>   AND IF VectorCompute metadata present in the module
>   THEN it's vector compute module - add the appropriate capability along with long vector ext


May I ask you to provide in the description, why you need to keep it? Without insights/clarification I kinda feel, that what you really want to achieve is actually:

>   IF SPV_EXT_long_vector is enabled
>   AND IF SPV_INTEL_vector_compute is enabled
>   AND IF there is a strange vector
>   **~AND IF VectorCompute metadata present in the module~**
>   THEN it's vector compute module - add the appropriate capability long vector ext

So `IsVectorComputeModule` (and the logic that it adds) is not needed.

maarquitos14 (commented):
Could you please describe what is this trying to do and why is it needed? Honestly, it's not really clear to me. Also, I tried to find the extension in SPIR-V Registry and in intel/llvm, and the best I could find is a https://github.com/intel/llvm/pull/1612, but the extension doesn't seem to exist in the repo anymore, so I'm not sure if this is the current state of the extension.

lib/SPIRV/libSPIRV/SPIRVModule.h
YuriPlyakhin: yep. No need to change pre-existing code style. But since we already have a precedent in the existing code base with non-virtual getter/setter, I suggest following this style for new introduced getter/setter to not expand ill pattern.

test/extensions/INTEL/SPV_INTEL_vector_compute/prioritize_over_LongVecEXT_func.ll
YuriPlyakhin: I believe you mean keeping size 1 test, but keeping 2 tests total: one for function and one for global. Otherwise, coverage would not be complete. Please, confirm.

test/extensions/INTEL/SPV_INTEL_vector_compute/prioritize_over_LongVecEXT_func.ll
YuriPlyakhin: we have `--implicit-check-not="Capability LongVectorEXT"`. I guess we can extend it for extension as well.

lib/SPIRV/libSPIRV/SPIRVModule.h
YuriPlyakhin: I updated the description.

YuriPlyakhin:
> Could you please describe what is this trying to do and why is it needed? Honestly, it's not really clear to me.

Sure! I posted the answer in description. I hope that clarifies the patch.

> Also, I tried to find the extension in SPIR-V Registry and in intel/llvm, and the best I could find is a [intel/llvm#1612](https://github.com/intel/llvm/pull/1612), but the extension doesn't seem to exist in the repo anymore, so I'm not sure if this is the current state of the extension.

I'm afraid what you found is the only spec we have today and it doesn't look like current, but it is the only one we have.

test/extensions/INTEL/SPV_INTEL_vector_compute/prioritize_over_LongVecEXT_func.ll
```diff
@@ -0,0 +1,20 @@
+; RUN: llvm-spirv %s --spirv-ext=+SPV_EXT_long_vector,+SPV_INTEL_vector_compute -o %t.spv
+; RUN: llvm-spirv %t.spv -to-text -o - | FileCheck %s --implicit-check-not="Capability LongVectorEXT"
+; TODO: re-enable spirv-val once it can recognize that TypeVector is allowed to have any number of
```
maarquitos14: Is there an issue for this in SPIR-V Tools repo? If so, can we link/reference it?

test/extensions/INTEL/SPV_INTEL_vector_compute/prioritize_over_LongVecEXT_func.ll
```diff
@@ -0,0 +1,20 @@
...
+; components when SPV_INTEL_vector_compute is enabled.
+; RUNx: spirv-val %t.spv
+
```
maarquitos14, line 7: Can we have a brief description for each test?

maarquitos14 (approved):
LGTM, but I wonder if `VectorCompute consumers require VectorAnyINTEL, so this is a functional regression.` should be addressed in consumer's side instead. I understand that I can take time to adapt for consumers, so is there plan for consumers to adapt? If so, we might want to open an issue to remind us that this should be removed when consumers catch up.

YuriPlyakhin:
> I wonder if `VectorCompute consumers require VectorAnyINTEL, so this is a functional regression.` should be addressed in consumer's side instead. I understand that I can take time to adapt for consumers, so is there plan for consumers to adapt? If so, we might want to open an issue to remind us that this should be removed when consumers catch up.

Last time we checked it didn't look like consumers of `SPV_INTEL_vector_compute` are willing to change status quo.

MrSidims (approved):

lib/SPIRV/libSPIRV/SPIRVModule.h
MrSidims: > with both SPV_INTEL_vector_compute and SPV_EXT_long_vector enabled

Why do you want both extensions enabled simultaneously? A single extension is enough. Or is there a flow, that adds +all (if so I can assure you, that it's a bad practice to enable all the extensions)?

test/extensions/INTEL/SPV_INTEL_vector_compute/prioritize_over_LongVecEXT_func.ll
MrSidims: Oh, there are also globals. Yeah, 2 tests are enough.

lib/SPIRV/libSPIRV/SPIRVModule.h
Qi-Ye-079: Done

Qi-Ye-079: Done

lib/SPIRV/libSPIRV/SPIRVModule.h
Qi-Ye-079: Done

Qi-Ye-079: Done

lib/SPIRV/libSPIRV/SPIRVModule.h
Qi-Ye-079: Done

Qi-Ye-079: Done

test/extensions/INTEL/SPV_INTEL_vector_compute/prioritize_over_LongVecEXT_func.ll
Qi-Ye-079: Done. 1 for function and 1 for global.

test/extensions/INTEL/SPV_INTEL_vector_compute/prioritize_over_LongVecEXT_func.ll
Qi-Ye-079: Added `--implicit-check-not='Extension "SPV_EXT_long_vector"'`

test/extensions/INTEL/SPV_INTEL_vector_compute/prioritize_over_LongVecEXT_func.ll
Qi-Ye-079: Doesn't look like so. I and Opus searched the repo but couldn't find any.

test/extensions/INTEL/SPV_INTEL_vector_compute/prioritize_over_LongVecEXT_func.ll
Qi-Ye-079: Done

test/extensions/INTEL/SPV_INTEL_vector_compute/prioritize_over_LongVecEXT_func.ll
maarquitos14: Okay, guess we should open it then. Otherwise I don't think this will be addressed.

lib/SPIRV/libSPIRV/SPIRVModule.h
YuriPlyakhin: It's not `+all`, but in the current productized offloading model we can't enable a distinct set of extensions per module — a single subset is applied across all of them. At the same time, we are trying to get rid of `SPV_INTEL_vector_compute`, but it is still needed for SYCL ESIMD. So, we want to generate SPV_EXT_long_vector for everything, but ESIMD.

I agree the cleaner long-term direction is finer-grained control over which extensions are enabled per module, which would remove the need for this special case, and we are designing it. Meanwhile, the metadata check is a small, self-contained way to keep `VectorCompute` modules on VectorAnyINTEL while everything else still prefers `LongVectorEXT`, and I'd propose it as an interim measure.

YuriPlyakhin (approved):
LGTM

lib/SPIRV/libSPIRV/SPIRVModule.h
MrSidims: I see. I do remember there was a way to check for ESIMD pipeline in https://github.com/intel/llvm/blob/sycl/clang/lib/Driver/ToolChains/Clang.cpp (and guard it appropriately), but I no longer see such way.

BTW, since `SPV_INTEL_vector_compute` is now enabled by default, does it mean, that I can write variable-length vectors in SYCL and IGC will be happy about it? :) Worth trying for my pet project!

test/extensions/INTEL/SPV_INTEL_vector_compute/prioritize_over_LongVecEXT_func.ll
Qi-Ye-079: https://github.com/KhronosGroup/SPIRV-Tools/issues/6786

MrSidims: > https://github.com/KhronosGroup/SPIRV-Tools/issues/6786

If you are deprecating the extension, do you really want to register it in SPIR-V tools?

MrSidims: This TODO is redundant. spirv-val must never be run with SPV_INTEL_vector_compute. Only with SPV_EXT_long_vector.

lib/SPIRV/libSPIRV/SPIRVModule.h
YuriPlyakhin: Well, I don't think IGC will be happy about `SPV_INTEL_vector_compute` usage, but they are accepting it for now:
https://github.com/intel/intel-graphics-compiler/blob/48e8e97e3e194b29a65ac4653e50f1f8f9e0c1fd/IGC/AdaptorOCL/ocl_igc_interface/extensions/SPIRVExtensions.td#L253-L262
But we hope to replace it with `SPV_EXT_long_vector` for scalar path eventually.

Anyway, you can try variable-length vectors and see how it goes, no promises, but I'm also curious. :)

test/extensions/INTEL/SPV_INTEL_vector_compute/prioritize_over_LongVecEXT_func.ll
YuriPlyakhin: @Qi-Ye-079 , please, close the issue and remove the TODO.

lib/SPIRV/libSPIRV/SPIRVModule.h
YuriPlyakhin: I believe that this thread is resolved.

test/extensions/INTEL/SPV_INTEL_vector_compute/prioritize_over_LongVecEXT_func.ll
Qi-Ye-079: Done

MrSidims (commented):
fine with me

PR #3830, opened 2026-07-02
Fix SPV_INTEL_function_pointers translation with untyped pointers

Qi-Ye-079:
The `SPV_INTEL_function_pointers` extension enables indirect function calls with function pointer. However, when the `SPV_KHR_untyped_pointers` is also enabled, all function pointers become untyped too, and the Reader will crash with assertion error, because it still assumes the function pointer is always typed.

This PR adds proper handling for untyped function pointer  in the Reader in case of an indirect function call.

In addition, to preserve type information for `byval`/`sret` function arguments at call site, and also to fix the type mismatch between function params vs args for a direct function call, the Writer has been changed to always emit these type-attributed function arguments as typed, and the Reader assumes that such function arguments remain typed.

AI-assisted by: Claude Opus 4.8


Qi-Ye-079:
@vmaksimo Please take a look.

P.S. For the testcases, I only enabled the `SPV_KHR_untyped_pointers` for a subset of testcases with indirect function calls; others only have global variables of function pointers but no indirect calls, so they're out of scope for this PR (the only exception is `global-function-pointer.ll` which I think is enough in terms of test coverage). Let me know if you think it's actually necessary to enable the extension for all of them under test/extensions/INTEL/SPV_INTEL_function_pointers.

test/extensions/INTEL/SPV_INTEL_function_pointers/bitcast.ll
```diff
@@ -25,6 +31,19 @@
...
+; CHECK-SPIRV-UNTYPED: Extension "SPV_INTEL_function_pointers"
+; CHECK-SPIRV-UNTYPED: Extension "SPV_KHR_untyped_pointers"
+; CHECK-SPIRV-UNTYPED-DAG: TypeInt [[#I8:]] 8
+; CHECK-SPIRV-UNTYPED-DAG: TypeInt [[#I32:]] 32
+; CHECK-SPIRV-UNTYPED-DAG: TypeFunction [[#FOO_TY:]] [[#I8]] [[#I8]]
+; CHECK-SPIRV-UNTYPED-DAG: TypeFunction [[#DEST_TY:]] [[#I32]] [[#I32]]
+; CHECK-SPIRV-UNTYPED-DAG: TypeUntypedPointerKHR [[#FOO_TY_PTR:]] 7
+; CHECK-SPIRV-UNTYPED: ConstantFunctionPointerINTEL [[#FOO_TY_PTR]] [[#FOO_PTR:]] [[#FOO:]]
+; CHECK-SPIRV-UNTYPED: Function [[#]] [[#FOO]] [[#]] [[#FOO_TY]]
+; CHECK-SPIRV-UNTYPED: Bitcast [[#FOO_TY_PTR]] [[#]] [[#FOO_PTR]]
```
vmaksimo, line 44: These lines except the `CHECK-SPIRV-UNTYPED-DAG: TypeUntypedPointerKHR [[#FOO_TY_PTR:]] 7` fully duplicate the CHECKs for typed pointer case.
Please merge them so only the different lines are actually differ.
See [this test/PR](https://github.com/KhronosGroup/SPIRV-LLVM-Translator/pull/2827/changes#diff-906fb50fa1b10c20aef7ddbe9245e5fc2975d443e6d09f52426dc513cd588cd2) as an example.
Please apply in all the tests wherever possible/makes sense

lib/SPIRV/SPIRVReader.cpp
```diff
@@ -2737,12 +2748,20 @@ Value *SPIRVToLLVM::transValueWithoutDecoration(SPIRVValue *BV, Function *F,
...
         static_cast<SPIRVFunctionPointerCallINTEL *>(BV);
     auto *V = transValue(BC->getCalledValue(), F, BB);
+    std::vector<Value *> Args = transValue(BC->getArgumentValues(), F, BB);
     auto *SpirvFnTy = BC->getCalledValue()->getType()->getPointerElementType();
-    auto *FnTy = cast<FunctionType>(transType(SpirvFnTy));
-    auto *Call = CallInst::Create(
-        FnTy, V, transValue(BC->getArgumentValues(), F, BB), BC->getName(), BB);
-    transFunctionPointerCallArgumentAttributes(
-        BV, Call, static_cast<SPIRVTypeFunction *>(SpirvFnTy));
+    auto *FnTy = dyn_cast<FunctionType>(transType(SpirvFnTy));
```
vmaksimo: We should not use `dyn_cast` if it's not inside the condition - use `cast` or `static_cast`

lib/SPIRV/SPIRVReader.cpp
```diff
@@ -2737,12 +2748,20 @@ Value *SPIRVToLLVM::transValueWithoutDecoration(SPIRVValue *BV, Function *F,
...
+    if (!FnTy) {
+      // An untyped function pointer has no pointee type, so rebuild the
+      // signature from the call's return and argument types.
+      SmallVector<Type *, 8> ArgTys;
+      for (Value *Arg : Args)
+        ArgTys.push_back(Arg->getType());
+      FnTy = FunctionType::get(transType(BC->getType()), ArgTys,
+                               /*isVarArg=*/false);
```
vmaksimo, line 2792: Just curious, what will happen if we translate function with variable args number? Should we test this case?

lib/SPIRV/SPIRVReader.cpp
Qi-Ye-079: Done

lib/SPIRV/SPIRVReader.cpp
Qi-Ye-079: I don't think SPIRV supports functions with variadic args at all. The Writer will reject it with `SPIRVEC_UnsupportedVarArgFunction` error. The Reader, when translating `OpTypeFunction`, always sets `isVarArg=false`, so variadic-arg function can never appear in valid SPIRV.

test/extensions/INTEL/SPV_INTEL_function_pointers/bitcast.ll
Qi-Ye-079: Done

vmaksimo (commented):
Please next time avoid force-pushes - it breaks PR iterations history and complicates reviewing process

Qi-Ye-079:
> Please next time avoid force-pushes - it breaks PR iterations history and complicates reviewing process

The force-push is to rebase the code to latest main (since the PR was created a while ago). The new commit to fix the comments is preserved.

vmaksimo:
One can just do `git merge origin/main` to update the branch to the latest main - it would preserve earlier commit hashes. As we usually merge per squash strategy - no ugly history will be created.
But thanks for putting the change in a separate commit at least, that's easier to review :)

test/extensions/INTEL/SPV_INTEL_function_pointers/CodeSectionINTEL/bitcast.ll
```diff
@@ -9,23 +9,38 @@
...
 ; CHECK-SPIRV-DAG: TypeInt [[#I32:]] 32
 ; CHECK-SPIRV-DAG: TypeFunction [[#FOO_TY:]] [[#I8]] [[#I8]]
 ; CHECK-SPIRV-DAG: TypeFunction [[#DEST_TY:]] [[#I32]] [[#I32]]
-; CHECK-SPIRV-DAG: TypePointer [[#DEST_TY_PTR:]] [[#]] [[#DEST_TY]]
-; CHECK-SPIRV-DAG: TypePointer [[#FOO_TY_PTR:]] [[#]] [[#FOO_TY]]
-; CHECK-SPIRV: ConstantFunctionPointerINTEL [[#FOO_TY_PTR]] [[#FOO_PTR:]] [[#FOO:]]
+; CHECK-SPIRV-TYPED-DAG: TypePointer [[#DEST_TY_PTR:]] [[#]] [[#DEST_TY]]
+; CHECK-SPIRV-TYPED-DAG: TypePointer [[#FOO_TY_PTR:]] [[#]] [[#FOO_TY]]
+; CHECK-SPIRV-UNTYPED-DAG: TypeVoid [[#VOID:]]
+; CHECK-SPIRV-UNTYPED-DAG: TypeUntypedPointerKHR [[#PTR:]] [[#VOID]]
```
vmaksimo: The second arg is a Storage Class, no relation to void type.
Rename to [[#]], [[#STORAGE_CLASS]], or just place real storage class literal number instead

test/extensions/INTEL/SPV_INTEL_function_pointers/CodeSectionINTEL/decor-func-ptr-arg-attr.ll
```diff
@@ -1,21 +1,42 @@
...
-; CHECK-SPIRV-SAME: [[#TargetId]]
+; CHECK-SPIRV-TYPED: Decorate [[#TargetId:]] ArgumentAttributeINTEL 0 5
+; CHECK-SPIRV-TYPED: Decorate [[#TargetId]] ArgumentAttributeINTEL 0 4
+; CHECK-SPIRV-TYPED: Decorate [[#TargetId]] ArgumentAttributeINTEL 0 2
+; CHECK-SPIRV-TYPED: FunctionPointerCallINTEL
+; CHECK-SPIRV-TYPED-SAME: [[#TargetId]]
+; CHECK-SPIRV-UNTYPED-DAG: Decorate [[#ARG:]] ArgumentAttributeINTEL 0 5
+; CHECK-SPIRV-UNTYPED-DAG: Decorate [[#ARG]] ArgumentAttributeINTEL 0 4
+; CHECK-SPIRV-UNTYPED-DAG: Decorate [[#ARG]] ArgumentAttributeINTEL 0 2
+; CHECK-SPIRV-UNTYPED-DAG: TypeVoid [[#VOID:]]
```
vmaksimo: void type not used?

lib/SPIRV/SPIRVReader.cpp
```diff
@@ -1538,12 +1539,22 @@ void SPIRVToLLVM::transFunctionPointerCallArgumentAttributes(
...
+    Attribute LlvmAttr;
+    if (Attribute::isTypeAttrKind(LlvmAttrKind)) {
+      // byval/sret need the argument's pointee type. Take it from the argument
+      // value so it works for both typed and untyped pointers.
+      SPIRVValue *Arg = ArgValues[ArgNo];
+      SPIRVType *PointeeTy = Arg->getType()->getPointerElementType();
+      if (Arg->isUntypedVariable()) {
+        if (SPIRVType *DT =
+                static_cast<SPIRVUntypedVariableKHR *>(Arg)->getDataType())
+          PointeeTy = DT;
```
vmaksimo: Guess it makes sense to reuse existing `getTypedPtrFromUntypedOperand` method for pointee type (it handles more instructions than just untyped variable)

https://github.com/KhronosGroup/SPIRV-LLVM-Translator/blob/221ab532b310cccb2577b9e6606a6619e641f758/lib/SPIRV/SPIRVReader.cpp#L3909-L3911

lib/SPIRV/SPIRVReader.cpp
```diff
@@ -1538,12 +1539,22 @@ void SPIRVToLLVM::transFunctionPointerCallArgumentAttributes(
...
-        Attribute::isTypeAttrKind(LlvmAttrKind)
-            ? Attribute::get(CI->getContext(), LlvmAttrKind,
-                             transType(CalledFnTy->getParameterType(ArgNo)
-                                           ->getPointerElementType()))
-            : Attribute::get(CI->getContext(), LlvmAttrKind);
+    Attribute LlvmAttr;
+    if (Attribute::isTypeAttrKind(LlvmAttrKind)) {
+      // byval/sret need the argument's pointee type. Take it from the argument
+      // value so it works for both typed and untyped pointers.
+      SPIRVValue *Arg = ArgValues[ArgNo];
```
vmaksimo: Nit: Just as we're already here, may you add bound check? 
`ArgNo` comes directly from an `ArgumentAttributeINTEL` decoration literal, it's an independent metadata, not validated against the call's actual args number.

```suggestion
      SPIRVValue *Arg = ArgValues[ArgNo];
      if (ArgNo >= ArgValues.size())
        continue; // Ignore a malformed ArgumentAttributeINTEL decoration.
```

This is no-blocking, as no real producer emits an out-of-range `ArgNo`.

Qi-Ye-079: Done. I moved the check right after `ArgNo` is read, so that it guards all paths, not just for this byval/byref path.

lib/SPIRV/SPIRVReader.cpp
Qi-Ye-079: Done

test/extensions/INTEL/SPV_INTEL_function_pointers/CodeSectionINTEL/bitcast.ll
Qi-Ye-079: Nice catch, it's a coincidence that the `TypeVoid` above happen to have ID 7 which is the same as Function storage class, so it passed without issue. Changed to `[[#]]` and removed `TypeVoid`.

test/extensions/INTEL/SPV_INTEL_function_pointers/CodeSectionINTEL/decor-func-ptr-arg-attr.ll
Qi-Ye-079: Done. Removed it.

Qi-Ye-079:
@MrSidims @svenvh This PR involves an intel extension only, but any community feedback is appreciated if any. Thanks!

test/extensions/INTEL/SPV_INTEL_function_pointers/CodeSectionINTEL/bitcast.ll
```diff
@@ -9,23 +9,37 @@
 
 ; RUN: llvm-spirv %s -spirv-ext=+SPV_INTEL_function_pointers -o %t.spv
```
MrSidims: Would you mind adding llc (aka SPIR-V backend) producer call to the test for both TYPED and UNTYPED cases to check if it works?

lib/SPIRV/SPIRVReader.cpp
```diff
@@ -1520,14 +1520,17 @@ void SPIRVToLLVM::addMemAliasMetadata(Instruction *I, SPIRVId AliasListId,
...
   std::vector<SPIRVDecorate const *> ArgumentAttributes =
       BV->getDecorations(internal::DecorationArgumentAttributeINTEL);
 
+  std::vector<SPIRVValue *> ArgValues = Call->getArgumentValues();
   for (const auto *Dec : ArgumentAttributes) {
     std::vector<SPIRVWord> Literals = Dec->getVecLiteral();
     SPIRVWord ArgNo = Literals[0];
     SPIRVWord SpirvAttr = Literals[1];
+    if (ArgNo >= ArgValues.size())
+      continue; // Ignore a malformed ArgumentAttributeINTEL decoration.
```
MrSidims: Wait, why? Does it belong in this patch?

MrSidims, line 1533: Does it belong to this patch?

test/extensions/INTEL/SPV_INTEL_function_pointers/CodeSectionINTEL/bitcast.ll
MrSidims, line 10: Since llvm-spirv now should also consume modules produced from SPIR-V backend, could you please add llc runs to make sure reader doesn't crash? Applied to both TYPED and UNTYPED cases across the tests.

test/extensions/INTEL/SPV_INTEL_function_pointers/CodeSectionINTEL/function-pointer-as-function-arg.ll
```diff
@@ -80,6 +89,45 @@
...
+; CHECK-LLVM-UNTYPED: alloca ptr
+; CHECK-LLVM-UNTYPED: store ptr %
+; CHECK-LLVM-UNTYPED: load ptr, ptr %
+; CHECK-LLVM-UNTYPED: call spir_func i32 %
+; CHECK-LLVM-UNTYPED: define spir_kernel void @test
+; CHECK-LLVM-UNTYPED: %fp = alloca ptr
+; CHECK-LLVM-UNTYPED: store ptr addrspacecast (ptr addrspace(9) @foo to ptr), ptr %fp
+; CHECK-LLVM-UNTYPED: store ptr addrspacecast (ptr addrspace(9) @bar to ptr), ptr %fp
+; CHECK-LLVM-UNTYPED: load ptr, ptr %fp
+; CHECK-LLVM-UNTYPED: call spir_func i32 @helper(ptr %
```
MrSidims: Lets have more detailed checks for function types and actual function pointers. See CHECK-LLVM from above.

test/extensions/INTEL/SPV_INTEL_function_pointers/CodeSectionINTEL/function-pointer.ll
```diff
@@ -17,28 +24,46 @@
...
+; CHECK-SPIRV-UNTYPED: UntypedVariableKHR [[#PTR]] [[#ALLOCA:]] [[#]] [[#PTR]]
+; CHECK-SPIRV-UNTYPED: Store [[#ALLOCA]] [[#FP]]
+; CHECK-SPIRV-UNTYPED: Load [[#PTR]] [[#LOADED:]] [[#ALLOCA]]
+; CHECK-SPIRV-UNTYPED: FunctionPointerCallINTEL [[#]] [[#]] [[#LOADED]]
+
+; CHECK-LLVM-UNTYPED: define spir_kernel void @test
+; CHECK-LLVM-UNTYPED: %fp = alloca ptr
+; CHECK-LLVM-UNTYPED: store ptr addrspacecast (ptr addrspace(9) @foo to ptr), ptr %fp
+; CHECK-LLVM-UNTYPED: %{{.*}} = load ptr, ptr %fp
+; CHECK-LLVM-UNTYPED: %call = call spir_func i32 %{{.*}}(i32 %{{.*}})
```
MrSidims: Result of the load is important to check.

MrSidims: Same applicable to other tests below.

lib/SPIRV/SPIRVReader.cpp
```diff
@@ -1538,12 +1541,35 @@ void SPIRVToLLVM::transFunctionPointerCallArgumentAttributes(
...
+                "argument with type attribute of an indirect call"))
+          return;
+        PointeeTy = cast<TypedPointerType>(TPT)->getElementType();
+      } else {
+        PointeeTy = transType(ArgTy->getPointerElementType());
+      }
+      LlvmAttr = Attribute::get(CI->getContext(), LlvmAttrKind, PointeeTy);
+    } else {
+      LlvmAttr = Attribute::get(CI->getContext(), LlvmAttrKind);
+    }
```
MrSidims: Why do we need type deduction on **main** branch?

MrSidims (commented):
> This PR involves an intel extension only, but any community feedback is appreciated if any.

We use both extensions in SPIR-V backend compilation flow as well as in llvm-spirv flow. I wonder why I haven't seen any issues on the reader side during test runs. Will double check this.

> However, when the SPV_KHR_untyped_pointers is also enabled, all function pointers become untyped too, and the Reader will crash with assertion error, because it still assumes the function pointer is always typed.

FYI SPIR-V backend keep function pointers typed even with the untyped extension enabled. It's not necessarily the best approach, but it's also correct and t "just works".

lib/SPIRV/SPIRVReader.cpp
Qi-Ye-079: No necessarily, but it's requested by Viktoria: https://github.com/KhronosGroup/SPIRV-LLVM-Translator/pull/3830/changes#r3757892939

lib/SPIRV/SPIRVReader.cpp
Qi-Ye-079: I'm not sure if I understand your question. This part isn't for the function pointer itself, but for the function arguments of the caller. If any argument is an (untyped) pointer with type attributes (`byval`/`byref`/...), we need to infer its pointee type correctly, as it's needed by the type attributes, even in untyped case.
One test example that covers this scenario is in `decor-func-ptr-arg-attr.ll`, where you can see this line of indirect function call:
```
call spir_func void %cond.i.i(ptr nonnull byval(%"multi_ptr") align 8 noalias captures(none) %agg.tmp.i.i)
```
The argument `%agg.tmp.i.i` is a pointer value with `byval(%"multi_ptr")` attribtue. In untyped pointer scenario, during reverse translation, we must try to infer and restore the pointee type for `byval` in the resulting LLVM IR.

test/extensions/INTEL/SPV_INTEL_function_pointers/CodeSectionINTEL/bitcast.ll
Qi-Ye-079: So I tried adding llc runs followed by the reader for all the `SPV_INTEL_function_pointers` tests, and I find that the reader _does_ crash with the output from llc for some. Among these tests:
- for typed scenario (without `+SPV_KHR_untyped_pointer` for llc), the reader crashes on 3 tests.
- for untyped scenario, (adding `+SPV_KHR_untyped_pointer` to llc), the reader crashes on 10.

Their crashing spots seem to vary, not from a single source. Note that I've verified that all these crashes are pre-existing, even without this PR, so it's out of the scope of this patch to fix them. But I can create an issue for them later.

So to your request:
1. For those crashing tests, add RUNx "llc + reader" lines + a one-liner comment to say they currently crash.
2. For those successful tests: they don't crash so I can add "llc + reader" runs, but just a remark that, the llc seems to generate different output than the writer, which in turn leads to different output from the reader. So, I can't simply reuse the existing CHECK-LLVM lines to check the reader output from llc, but need to add new CHECKs for "llc + reader" runs specifically.

test/extensions/INTEL/SPV_INTEL_function_pointers/CodeSectionINTEL/function-pointer-as-function-arg.ll
Qi-Ye-079: Done

test/extensions/INTEL/SPV_INTEL_function_pointers/CodeSectionINTEL/function-pointer.ll
Qi-Ye-079: Done

lib/SPIRV/SPIRVReader.cpp
MrSidims: @vmaksimo do you remember, if some time ago we have decided to leave `byval/byref/sret` attributes alone and not translate their relative pointer args as `OpTypeUntypedPointerKHR` in the Writer as there is no way to preserve necessary type info, because https://registry.khronos.org/SPIR-V/specs/unified1/SPIRV.html#Function_Parameter_Attribute do not carry type info?

Qi-Ye-079: Right now, I believe that the Writer only translates `byval/byref/sret` attributed pointer for *function parameters* as typed pointer, not *function arguments* at call site:
https://github.com/KhronosGroup/SPIRV-LLVM-Translator/blob/de0bb2ad43bbf45d7da208b9ca0e1e8c3c6cba30/lib/SPIRV/SPIRVWriter.cpp#L873-L878
For attributed function arguments at call site, the Writer will emit them as `OpTypeUntypedPointerKHR`, plus a `ArgumentAttributeINTEL <ArgNo> <Attr>` decoration for the caller. This is what the `decor-func-ptr-arg-attr.ll` test covers. Therefore in the Reader, it has to try to infer the pointee type for function arguments at this point.

Qi-Ye-079:
@MrSidims Could you please take a another look to see if you have more comments?
@vmaksimo Please also give a final review when you have time. I think you already left comments for tests before and I fixed them.
Thank you both.

test/extensions/INTEL/SPV_INTEL_function_pointers/CodeSectionINTEL/bitcast.ll
MrSidims: I see, if they crash before the patch - no objections from me to leave it as is.

lib/SPIRV/SPIRVReader.cpp
MrSidims: I still don't understand, why we can't keep byval/sret parameters typed on both call and declaration/definition sites. And I still believe that this type deduction is not needed. But I probably understand, where is the problem and what should be done.

Consider this:
```
%struct.S = type { i32, float }
  define spir_func void @callee(ptr byval(%struct.S) align 8 %p) { ret void }
  define spir_kernel void @kern() {
    %tmp = alloca %struct.S, align 8
    call spir_func void @callee(ptr byval(%struct.S) align 8 %tmp)
    ret void
  }
```
note, no function pointers, just a normal function call.
after llvm-spirv -spirv-ext=+SPV_KHR_untyped_pointers:

```
  TypeStruct 3 4 5
  TypePointer 6 7 3                 <- callee's param: TYPED
  TypeFunction 7 2 6
  TypeUntypedPointerKHR 14 7
  ...
  UntypedVariableKHR 14 15 7 3      <- the alloca: UNTYPED
  FunctionCall 2 16 8 15            <- passes UNTYPED alloca
```
so we get an invalid SPIR-V for untyped pointers. So I believe we have to resolve this issue first and then, knowing the resolution, return to this comment. I feel like when the Writer bug is fixed - no type deduction will be required on the Reader side.

MrSidims (changes requested):

lib/SPIRV/SPIRVReader.cpp
Qi-Ye-079: Makes perfect sense, thank you. I'll make changes to emit byval/sret function arguments at caller as typed too in the Writer.
~~Also, I can take a step further to always emit _function pointers_ as typed as well, just to align with SPIRV-BE. In this case, all reader changes should be safely reverted and this will become a Writer-change only PR.~~
UPD: the Writer will now always translate byval/sret function arguments at caller to typed pointer, and the type inference part has been removed in the Reader. Please take a look.
For the function pointers itself though, I tried a bit but the changes required to emit them as typed doesn't seem trivial in the writer, since we need to cover every path where function pointers might be emitted. If we really want to do this to align with SPIRV-BE, I think it deserves a separate PR.

lib/SPIRV/SPIRVWriter.cpp
```diff
@@ -2603,9 +2603,9 @@ LLVMToSPIRVBase::transValueWithoutDecoration(Value *V, SPIRVBasicBlock *BB,
                                LoopControl, Parameters, SuccessorTrue);
       }
     }
-    // Starting with SPIR-V 1.6, the True Label and False Label of an
-    // "OpBranchConditional" must not be the same, so emit an unconditional
-    // branch - always valid, required for 1.6+
+    // Starting with SPIR-V 1.6, the True Label and False Label of an
+    // "OpBranchConditional" must not be the same, so emit an unconditional
+    // branch - always valid, required for 1.6+
```
MrSidims, line 2608: Seems unrelated

lib/SPIRV/SPIRVReader.cpp
```diff
@@ -1541,35 +1541,13 @@ void SPIRVToLLVM::transFunctionPointerCallArgumentAttributes(
...
-        PointeeTy = cast<TypedPointerType>(TPT)->getElementType();
-      } else {
-        PointeeTy = transType(ArgTy->getPointerElementType());
-      }
-      LlvmAttr = Attribute::get(CI->getContext(), LlvmAttrKind, PointeeTy);
-    } else {
-      LlvmAttr = Attribute::get(CI->getContext(), LlvmAttrKind);
-    }
+    SPIRVValue *Arg = ArgValues[ArgNo];
+    // assume all byval/sret args are always emitted as typed pointers
```
MrSidims, line 1545: Lets add assertion then

lib/SPIRV/SPIRVWriter.cpp
```diff
@@ -6182,24 +6182,38 @@ LLVMToSPIRVBase::transValue(const std::vector<Value *> &Args,
   return BArgs;
 }
 
-std::vector<SPIRVWord>
-LLVMToSPIRVBase::transValue(const std::vector<Value *> &Args,
-                            SPIRVBasicBlock *BB, SPIRVEntry *Entry) {
+std::vector<SPIRVWord> LLVMToSPIRVBase::transArguments(CallInst *CI,
```
MrSidims, line 6185: I'd slightly prefer to outline this into a separate PR, but won't insist. Regardless, PR description should be updated.

MrSidims (approved):
Thanks!

lib/SPIRV/SPIRVReader.cpp
MrSidims: I don't have a strong opinion about this. IMHO the BE could do better and also emit untyped pointers.

test/extensions/INTEL/SPV_INTEL_function_pointers/CodeSectionINTEL/bitcast.ll
```diff
@@ -9,23 +9,44 @@
...
 ; RUN: FileCheck < %t.r.ll %s --check-prefix=CHECK-LLVM
 
+; RUN: llvm-spirv %s -spirv-ext=+SPV_INTEL_function_pointers,+SPV_KHR_untyped_pointers -o %t.u.spv
+; RUN: llvm-spirv %t.u.spv -to-text -o %t.u.spt
+; RUN: FileCheck < %t.u.spt %s --check-prefixes=CHECK-SPIRV,CHECK-SPIRV-UNTYPED
+; RUN: llvm-spirv -r -spirv-emit-function-ptr-addr-space %t.u.spv -o %t.ru.bc
+; RUN: llvm-dis %t.ru.bc -o %t.ru.ll
+; RUN: FileCheck < %t.ru.ll %s --check-prefix=CHECK-LLVM-UNTYPED
+
+; TODO: reader currently crashes with output of llc for this test with function pointers (for both typed and untyped pointers)
```
vmaksimo: With the latest changes, does llc still crash?

test/extensions/INTEL/SPV_INTEL_function_pointers/CodeSectionINTEL/decor-func-ptr-arg-attr.ll
```diff
@@ -1,22 +1,55 @@
...
+; CHECK-SPIRV-UNTYPED: Select [[#PTR]] [[#SEL:]] [[#]] [[#FP]] [[#]]
+; CHECK-SPIRV-UNTYPED: Bitcast [[#]] [[#BARG:]] [[#VAR]]
+; CHECK-SPIRV-UNTYPED: FunctionPointerCallINTEL [[#]] [[#ARG]] [[#SEL]] [[#BARG]]
 
 ; CHECK-LLVM: call spir_func addrspace(9) void %cond.i.i(ptr noalias byval(%multi_ptr) captures(none) %agg.tmp.i.i)
 
+; CHECK-LLVM-UNTYPED: %[[AGG:.*]] = alloca %multi_ptr
+; CHECK-LLVM-UNTYPED: select i1 %_arg_, ptr addrspacecast (ptr addrspace(9) @inc_function to ptr), ptr null
+; CHECK-LLVM-UNTYPED: %[[BC:.*]] = bitcast ptr %[[AGG]] to ptr
+; CHECK-LLVM-UNTYPED: call spir_func void %cond.i.i(ptr noalias byval(%multi_ptr) captures(none) %[[BC]])
```
vmaksimo: I wonder why we don't get `addrspace(9)` in return type for untyped version

test/extensions/INTEL/SPV_INTEL_function_pointers/CodeSectionINTEL/decor-func-ptr-arg-attr.ll
```diff
@@ -1,22 +1,55 @@
...
+; CHECK-SPIRV-UNTYPED-DAG: Decorate [[#ARG]] ArgumentAttributeINTEL 0 4
+; CHECK-SPIRV-UNTYPED-DAG: Decorate [[#ARG]] ArgumentAttributeINTEL 0 2
+; CHECK-SPIRV-UNTYPED-DAG: TypeUntypedPointerKHR [[#PTR:]] [[#]]
+; CHECK-SPIRV-UNTYPED-DAG: TypeStruct [[#MULTI_PTR:]] [[#PTR]]
+; CHECK-SPIRV-UNTYPED-DAG: ConstantFunctionPointerINTEL [[#PTR]] [[#FP:]] [[#INC:]]
+; CHECK-SPIRV-UNTYPED: Function [[#]] [[#INC]]
+; CHECK-SPIRV-UNTYPED: UntypedVariableKHR [[#PTR]] [[#VAR:]] [[#]] [[#MULTI_PTR]]
+; CHECK-SPIRV-UNTYPED: Select [[#PTR]] [[#SEL:]] [[#]] [[#FP]] [[#]]
+; CHECK-SPIRV-UNTYPED: Bitcast [[#]] [[#BARG:]] [[#VAR]]
+; CHECK-SPIRV-UNTYPED: FunctionPointerCallINTEL [[#]] [[#ARG]] [[#SEL]] [[#BARG]]
```
vmaksimo: [TODO] is this a place for simplification?

test/extensions/INTEL/SPV_INTEL_function_pointers/CodeSectionINTEL/fp-from-host.ll
```diff
@@ -14,23 +31,37 @@
...
+; CHECK-SPIRV-TYPED: Function {{[0-9]+}} [[KERNEL_ID]]
+; CHECK-SPIRV-TYPED: FunctionParameter [[INT_PTR]] [[FP:[0-9]+]]
+; CHECK-SPIRV-TYPED: Load [[INT32_TYPE_ID]] [[FUNC_ADDR:[0-9]+]] [[FP]]
+; CHECK-SPIRV-TYPED: ConvertUToPtr [[FOO_TYPE_PTR_ID]] [[FOO_PTR:[0-9]+]] [[FUNC_ADDR]]
+; CHECK-SPIRV-TYPED: FunctionPointerCallINTEL [[INT32_TYPE_ID]] {{[0-9]+}} [[FOO_PTR]]
+; CHECK-SPIRV-UNTYPED: TypeInt [[#INT32:]] 32
+; CHECK-SPIRV-UNTYPED: TypeFunction [[#FOO_TY:]] [[#INT32]] [[#INT32]]
+; CHECK-SPIRV-UNTYPED: TypeUntypedPointerKHR [[#FOO_PTR:]] [[#]]
+; CHECK-SPIRV-UNTYPED: ConvertUToPtr [[#FOO_PTR]] [[#FP:]]
+; CHECK-SPIRV-UNTYPED: FunctionPointerCallINTEL [[#INT32]] [[#]] [[#FP]]
```
vmaksimo: [TODO] suggest improvement

test/extensions/INTEL/SPV_INTEL_function_pointers/CodeSectionINTEL/function-pointer-as-function-arg.ll
```diff
@@ -80,6 +98,49 @@
 ; CHECK-LLVM: %[[FP_LOADED:.*]] = load ptr addrspace(9), ptr %[[FP]]
 ; CHECK-LLVM: call spir_func i32 @helper(ptr addrspace(9) %[[FP_LOADED]]
 
+; CHECK-SPIRV-UNTYPED-DAG: EntryPoint [[#]] [[#KERNEL_ID:]] "test"
```
vmaksimo: [TODO] any way to reuse old values?

test/extensions/INTEL/SPV_INTEL_function_pointers/CodeSectionINTEL/function-pointer-dedicated-as.ll
```diff
@@ -40,20 +40,56 @@
 ; RUN: llvm-dis %t.r.bc -o %t.r.ll
 ; RUN: FileCheck < %t.r.ll %s --check-prefix=CHECK-LLVM-AS
 
+; RUN: llvm-as %s -o %t.bc
+; RUN: llvm-spirv %t.bc -spirv-text --spirv-ext=+SPV_INTEL_function_pointers,+SPV_KHR_untyped_pointers -spirv-emit-function-ptr-addr-space -o %t.u.spt
```
vmaksimo:
```suggestion
; RUN: llvm-spirv %t.bc -spirv-text --spirv-ext=+SPV_INTEL_function_pointers,+SPV_KHR_untyped_pointers -spirv-emit-function-ptr-addr-space -o %t.u.spt
```

test/extensions/INTEL/SPV_INTEL_function_pointers/CodeSectionINTEL/decor-func-ptr-arg-attr.ll
vmaksimo, line 48: I wonder why we don't get `addrspace(9)` in return type for untyped version?

UPD: the addrspace(9) tagging relies on checking the pointer's element type (`getPointerElementType()->getOpCode() == OpTypeFunction`) to spot function pointers. Obviously, it doesn't work for untyped pointers yet - worth being fixed (in a separate PR?)

lib/SPIRV/SPIRVReader.cpp
vmaksimo: Agree, but let's use `checkError` instead of `assert`

lib/SPIRV/SPIRVWriter.cpp
vmaksimo: Me too, but not sure if it worth the bother - just for the more consecutive git history

test/extensions/INTEL/SPV_INTEL_function_pointers/CodeSectionINTEL/function-pointer.ll
```diff
@@ -17,28 +34,58 @@
...
+; CHECK-SPIRV-UNTYPED: Capability UntypedPointersKHR
 ; CHECK-SPIRV: Capability FunctionPointersINTEL
 ; CHECK-SPIRV: Extension "SPV_INTEL_function_pointers"
-; CHECK-SPIRV: EntryPoint [[#]] [[KERNEL_ID:[0-9]+]] "test"
-; CHECK-SPIRV: TypeInt [[TYPE_INT_ID:[0-9]+]]
-; CHECK-SPIRV: TypeFunction [[FOO_TYPE_ID:[0-9]+]] [[TYPE_INT_ID]] [[TYPE_INT_ID]]
-; CHECK-SPIRV: TypePointer [[FOO_PTR_ID:[0-9]+]] {{[0-9]+}} [[FOO_TYPE_ID]]
-; CHECK-SPIRV: TypePointer [[FOO_PTR_ALLOCA_ID:[0-9]+]] 7 [[FOO_PTR_ID]]
-; CHECK-SPIRV: ConstantFunctionPointerINTEL [[FOO_PTR_ID]] [[FOO_PTR:[0-9]+]] [[FOO_ID:[0-9]+]]
+; CHECK-SPIRV-UNTYPED: Extension "SPV_KHR_untyped_pointers"
```
vmaksimo: `select.ll` (and `non-uniform-function-pointer.ll`/`referenced-indirectly.ll`) interleave the shared `CHECK-SPIRV`/`-DAG` lines with `CHECK-SPIRV-TYPED`/`-UNTYPED` only where they diverge. This file duplicates the whole block instead — worth aligning to that pattern.

test/extensions/INTEL/SPV_INTEL_function_pointers/CodeSectionINTEL/global-function-pointer.ll
```diff
@@ -1,21 +1,48 @@
...
 
 target datalayout = "e-i64:64-v16:16-v24:32-v32:32-v48:64-v96:128-v192:256-v256:256-v512:512-v1024:1024-n8:16:32:64"
 target triple = "spir64"
 
 
 ; CHECK-SPIRV: Capability FunctionPointersINTEL
 ; CHECK-SPIRV: Extension "SPV_INTEL_function_pointers"
-; CHECK-SPIRV: TypeFunction [[#FOO_TY:]] [[#]] [[#]]
-; CHECK-SPIRV: TypePointer [[#FOO_TY_PTR:]] [[#]] [[#FOO_TY]]
+; CHECK-SPIRV-UNTYPED: Extension "SPV_KHR_untyped_pointers"
```
vmaksimo: `select.ll` (and `non-uniform-function-pointer.ll`/`referenced-indirectly.ll`) interleave the shared `CHECK-SPIRV`/`-DAG` lines with `CHECK-SPIRV-TYPED`/`-UNTYPED` only where they diverge. This file duplicates the whole block instead — worth aligning to that pattern.

test/extensions/INTEL/SPV_INTEL_function_pointers/CodeSectionINTEL/decor-func-ptr-arg-attr.ll
vmaksimo: Reuse more common checks please, e.g.:
```suggestion
; CHECK-SPIRV-DAG: Decorate [[#TARGET:]] ArgumentAttributeINTEL 0 5
; CHECK-SPIRV-DAG: Decorate [[#TARGET]] ArgumentAttributeINTEL 0 4
; CHECK-SPIRV-DAG: Decorate [[#TARGET]] ArgumentAttributeINTEL 0 2
; CHECK-SPIRV-TYPED: FunctionPointerCallINTEL
; CHECK-SPIRV-TYPED-SAME: [[#TARGET]]
; CHECK-SPIRV-UNTYPED-DAG: TypeUntypedPointerKHR [[#PTR:]] [[#]]
; CHECK-SPIRV-UNTYPED-DAG: TypeStruct [[#MULTI_PTR:]] [[#PTR]]
; CHECK-SPIRV-UNTYPED-DAG: ConstantFunctionPointerINTEL [[#PTR]] [[#FP:]] [[#INC:]]
; CHECK-SPIRV-UNTYPED: Function [[#]] [[#INC]]
; CHECK-SPIRV-UNTYPED: UntypedVariableKHR [[#PTR]] [[#VAR:]] [[#]] [[#MULTI_PTR]]
; CHECK-SPIRV-UNTYPED: Select [[#PTR]] [[#SEL:]] [[#]] [[#FP]] [[#]]
; CHECK-SPIRV-UNTYPED: Bitcast [[#]] [[#BARG:]] [[#VAR]]
; CHECK-SPIRV-UNTYPED: FunctionPointerCallINTEL [[#]] [[#TARGET]] [[#SEL]] [[#BARG]]
```

test/extensions/INTEL/SPV_INTEL_function_pointers/CodeSectionINTEL/fp-from-host.ll
vmaksimo:
```suggestion
; CHECK-SPIRV: EntryPoint [[#]] [[#KERNEL_ID:]] "test"
; CHECK-SPIRV: TypeInt [[#INT32:]] 32
; CHECK-SPIRV-TYPED: TypePointer [[#INT_PTR:]] 5 [[#INT32]]
; CHECK-SPIRV-UNTYPED: TypeUntypedPointerKHR [[#INT_PTR:]] [[#]]
; CHECK-SPIRV: TypeFunction [[#FOO_TY:]] [[#INT32]] [[#INT32]]
; CHECK-SPIRV-TYPED: TypePointer [[#FOO_TY_PTR:]] [[#]] [[#FOO_TY]]
; CHECK-SPIRV-UNTYPED: TypeUntypedPointerKHR [[#FOO_TY_PTR:]] [[#]]
;
; CHECK-SPIRV: Function [[#]] [[#KERNEL_ID]]
; CHECK-SPIRV: FunctionParameter [[#INT_PTR]] [[#FP:]]
; CHECK-SPIRV: Load [[#INT32]] [[#FUNC_ADDR:]] [[#FP]]
; CHECK-SPIRV: ConvertUToPtr [[#FOO_TY_PTR]] [[#FOO_PTR:]] [[#FUNC_ADDR]]
; CHECK-SPIRV: FunctionPointerCallINTEL [[#INT32]] [[#]] [[#FOO_PTR]]
```

test/extensions/INTEL/SPV_INTEL_function_pointers/CodeSectionINTEL/function-pointer-as-function-arg.ll
```diff
@@ -30,40 +46,42 @@
...
-; CHECK-SPIRV: Store [[F_PTR_ALLOCA_ID]] [[FOO_PTR_ID]]
-; CHECK-SPIRV: Store [[F_PTR_ALLOCA_ID]] [[BAR_PTR_ID]]
-; CHECK-SPIRV: Load [[FOO_PTR_TYPE_ID]] [[LOADED_F_PTR:[0-9]+]] [[F_PTR_ALLOCA_ID]]
-; CHECK-SPIRV: FunctionCall {{[0-9]+}} {{[0-9]+}} [[HELPER_ID]] [[LOADED_F_PTR]]
+; CHECK-SPIRV-TYPED: Function {{[0-9]+}} [[KERNEL_ID]]
+; CHECK-SPIRV-TYPED: Variable [[FOO_PTR_ALLOCA_TYPE_ID]] [[F_PTR_ALLOCA_ID:[0-9]+]]
+; CHECK-SPIRV-TYPED: Store [[F_PTR_ALLOCA_ID]] [[FOO_PTR_ID]]
+; CHECK-SPIRV-TYPED: Store [[F_PTR_ALLOCA_ID]] [[BAR_PTR_ID]]
+; CHECK-SPIRV-TYPED: Load [[FOO_PTR_TYPE_ID]] [[LOADED_F_PTR:[0-9]+]] [[F_PTR_ALLOCA_ID]]
+; CHECK-SPIRV-TYPED: FunctionCall {{[0-9]+}} {{[0-9]+}} [[HELPER_ID]] [[LOADED_F_PTR]]
```
vmaksimo: Actual merge, AI-verified against real `llvm-spirv` output (both TYPED and UNTYPED FileCheck runs pass). 
The old UNTYPED-only block below (`CHECK-SPIRV-UNTYPED-DAG: EntryPoint`...) becomes redundant with this and should be deleted -- see next comment in the file.
```suggestion
; CHECK-SPIRV: EntryPoint [[#]] [[#KERNEL_ID:]] "test"
; CHECK-SPIRV: TypeInt [[#INT32:]] 32
; CHECK-SPIRV: TypeFunction [[#FOO_TY:]] [[#INT32]] [[#INT32]]
; CHECK-SPIRV-TYPED: TypePointer [[#PTR:]] [[#]] [[#FOO_TY]]
; CHECK-SPIRV-UNTYPED: TypeUntypedPointerKHR [[#PTR:]] [[#]]
; CHECK-SPIRV: TypeFunction [[#HELPER_TY:]] [[#INT32]] [[#PTR]] [[#INT32]]
; CHECK-SPIRV-TYPED: TypePointer [[#PTR_ALLOCA_TY:]] [[#]] [[#PTR]]
; CHECK-SPIRV-TYPED: TypePointer [[#INT32_ALLOCA_TY:]] [[#]] [[#INT32]]
; CHECK-SPIRV-TYPED: FunctionPointerINTEL [[#PTR]] [[#FOO_PTR:]] [[#FOO_FN:]]
; CHECK-SPIRV-TYPED: FunctionPointerINTEL [[#PTR]] [[#BAR_PTR:]] [[#BAR_FN:]]
; CHECK-SPIRV-UNTYPED: ConstantFunctionPointerINTEL [[#PTR]] [[#FOO_PTR:]] [[#FOO_FN:]]
; CHECK-SPIRV-UNTYPED: ConstantFunctionPointerINTEL [[#PTR]] [[#BAR_PTR:]] [[#BAR_FN:]]
;
; CHECK-SPIRV: Function [[#INT32]] [[#HELPER_ID:]] [[#]] [[#HELPER_TY]]
; CHECK-SPIRV: FunctionParameter [[#PTR]] [[#T_PTR_ARG:]]
; CHECK-SPIRV: FunctionParameter [[#INT32]] [[#INT_ARG:]]
; CHECK-SPIRV-TYPED: Variable [[#PTR_ALLOCA_TY]] [[#T_PTR_ALLOCA:]]
; CHECK-SPIRV-TYPED: Variable [[#INT32_ALLOCA_TY]] [[#INT_ALLOCA:]]
; CHECK-SPIRV-UNTYPED: UntypedVariableKHR [[#PTR]] [[#T_PTR_ALLOCA:]] [[#]] [[#PTR]]
; CHECK-SPIRV-UNTYPED: UntypedVariableKHR [[#PTR]] [[#INT_ALLOCA:]] [[#]] [[#INT32]]
; CHECK-SPIRV: Store [[#T_PTR_ALLOCA]] [[#T_PTR_ARG]]
; CHECK-SPIRV: Store [[#INT_ALLOCA]] [[#INT_ARG]]
; CHECK-SPIRV: Load [[#PTR]] [[#LOADED_T_PTR:]] [[#T_PTR_ALLOCA]]
; CHECK-SPIRV: Load [[#INT32]] [[#LOADED_INT:]] [[#INT_ALLOCA]]
; CHECK-SPIRV: FunctionPointerCallINTEL [[#INT32]] [[#RESULT:]] [[#LOADED_T_PTR]] [[#LOADED_INT]]
; CHECK-SPIRV: ReturnValue [[#RESULT]]
;
; CHECK-SPIRV: Function [[#INT32]] [[#FOO_FN]] [[#]] [[#FOO_TY]]
; CHECK-SPIRV: Function [[#INT32]] [[#BAR_FN]] [[#]] [[#FOO_TY]]
;
; CHECK-SPIRV: Function [[#]] [[#KERNEL_ID]]
; CHECK-SPIRV-TYPED: Variable [[#PTR_ALLOCA_TY]] [[#F_PTR_ALLOCA:]]
; CHECK-SPIRV-UNTYPED: UntypedVariableKHR [[#PTR]] [[#F_PTR_ALLOCA:]] [[#]] [[#PTR]]
; CHECK-SPIRV: Store [[#F_PTR_ALLOCA]] [[#FOO_PTR]]
; CHECK-SPIRV: Store [[#F_PTR_ALLOCA]] [[#BAR_PTR]]
; CHECK-SPIRV: Load [[#PTR]] [[#LOADED_F_PTR:]] [[#F_PTR_ALLOCA]]
; CHECK-SPIRV: FunctionCall [[#INT32]] [[#]] [[#HELPER_ID]] [[#LOADED_F_PTR]]
```

test/extensions/INTEL/SPV_INTEL_function_pointers/CodeSectionINTEL/function-pointer-as-function-arg.ll
```diff
@@ -80,6 +98,49 @@
...
+; CHECK-SPIRV-UNTYPED: FunctionPointerCallINTEL [[#INT]] [[#RESULT:]] [[#LOADED_T_PTR]] [[#LOADED_INT]]
+; CHECK-SPIRV-UNTYPED: ReturnValue [[#RESULT]]
+; CHECK-SPIRV-UNTYPED: Function [[#INT]] [[#FOO_FN]] [[#]] [[#FOO_TY]]
+; CHECK-SPIRV-UNTYPED: Function [[#INT]] [[#BAR_FN]] [[#]] [[#FOO_TY]]
+; CHECK-SPIRV-UNTYPED: Function [[#]] [[#KERNEL_ID]] [[#]] [[#]]
+; CHECK-SPIRV-UNTYPED: UntypedVariableKHR [[#PTR]] [[#F_PTR_ALLOCA:]] [[#]] [[#PTR]]
+; CHECK-SPIRV-UNTYPED: Store [[#F_PTR_ALLOCA]] [[#FOO]]
+; CHECK-SPIRV-UNTYPED: Store [[#F_PTR_ALLOCA]] [[#BAR]]
+; CHECK-SPIRV-UNTYPED: Load [[#PTR]] [[#LOADED_F_PTR:]] [[#F_PTR_ALLOCA]]
+; CHECK-SPIRV-UNTYPED: FunctionCall [[#INT]] [[#]] [[#HELPER_ID]] [[#LOADED_F_PTR]]
```
vmaksimo: Now redundant after the merge above - delete this whole block.
```suggestion
```

test/extensions/INTEL/SPV_INTEL_function_pointers/CodeSectionINTEL/function-pointer.ll
```diff
@@ -17,28 +34,58 @@
...
-; CHECK-SPIRV: Variable [[FOO_PTR_ALLOCA_ID]] [[FOO_PTR_ALLOCA:[0-9]+]]
-; CHECK-SPIRV: Store [[FOO_PTR_ALLOCA]] [[FOO_PTR]]
-; CHECK-SPIRV: Load [[FOO_PTR_ID]] [[LOADED_FOO_PTR:[0-9]+]] [[FOO_PTR_ALLOCA]]
-; CHECK-SPIRV: FunctionPointerCallINTEL 2 {{[0-9]+}} [[LOADED_FOO_PTR]]
+; CHECK-SPIRV-TYPED: Function {{[0-9]+}} [[FOO_ID]] {{[0-9]+}} [[FOO_TYPE_ID]]
+; CHECK-SPIRV-TYPED: Function {{[0-9]+}} [[KERNEL_ID]]
+; CHECK-SPIRV-TYPED: Variable [[FOO_PTR_ALLOCA_ID]] [[FOO_PTR_ALLOCA:[0-9]+]]
+; CHECK-SPIRV-TYPED: Store [[FOO_PTR_ALLOCA]] [[FOO_PTR]]
+; CHECK-SPIRV-TYPED: Load [[FOO_PTR_ID]] [[LOADED_FOO_PTR:[0-9]+]] [[FOO_PTR_ALLOCA]]
+; CHECK-SPIRV-TYPED: FunctionPointerCallINTEL 2 {{[0-9]+}} [[LOADED_FOO_PTR]]
```
vmaksimo: Checked the real output: `2` is `TypeInt`'s own result ID in both modes (`TypeInt 2 32 0` identically, and `FunctionPointerCallINTEL 2 ...` identically). Not a special/different value between TYPED and UNTYPED, fix please, e.g.:
```suggestion
; CHECK-SPIRV: EntryPoint [[#]] [[#KERNEL_ID:]] "test"
; CHECK-SPIRV: TypeInt [[#INT:]]
; CHECK-SPIRV: TypeFunction [[#FOO_TY:]] [[#INT]] [[#INT]]
; CHECK-SPIRV-TYPED: TypePointer [[#PTR:]] [[#]] [[#FOO_TY]]
; CHECK-SPIRV-TYPED: TypePointer [[#ALLOCA_TY:]] 7 [[#PTR]]
; CHECK-SPIRV-UNTYPED: TypeUntypedPointerKHR [[#PTR:]] [[#]]
; CHECK-SPIRV: ConstantFunctionPointerINTEL [[#PTR]] [[#FP:]] [[#FOO_FN:]]
;
; CHECK-SPIRV: Function [[#]] [[#FOO_FN]] [[#]] [[#FOO_TY]]
; CHECK-SPIRV: Function [[#]] [[#KERNEL_ID]]
; CHECK-SPIRV-TYPED: Variable [[#ALLOCA_TY]] [[#ALLOCA:]]
; CHECK-SPIRV-UNTYPED: UntypedVariableKHR [[#PTR]] [[#ALLOCA:]] [[#]] [[#PTR]]
; CHECK-SPIRV: Store [[#ALLOCA]] [[#FP]]
; CHECK-SPIRV: Load [[#PTR]] [[#LOADED:]] [[#ALLOCA]]
; CHECK-SPIRV: FunctionPointerCallINTEL [[#INT]] [[#]] [[#LOADED]]
```

test/extensions/INTEL/SPV_INTEL_function_pointers/CodeSectionINTEL/function-pointer.ll
```diff
@@ -17,28 +34,58 @@
...
 
+; CHECK-SPIRV-UNTYPED-DAG: TypeInt [[#INT:]]
+; CHECK-SPIRV-UNTYPED-DAG: TypeFunction [[#FOO_TY:]] [[#INT]] [[#INT]]
+; CHECK-SPIRV-UNTYPED-DAG: TypeUntypedPointerKHR [[#PTR:]] [[#]]
+; CHECK-SPIRV-UNTYPED-DAG: ConstantFunctionPointerINTEL [[#PTR]] [[#FP:]] [[#FOO_FN:]]
+; CHECK-SPIRV-UNTYPED: Function [[#]] [[#FOO_FN]] [[#]] [[#FOO_TY]]
+; CHECK-SPIRV-UNTYPED: UntypedVariableKHR [[#PTR]] [[#ALLOCA:]] [[#]] [[#PTR]]
+; CHECK-SPIRV-UNTYPED: Store [[#ALLOCA]] [[#FP]]
+; CHECK-SPIRV-UNTYPED: Load [[#PTR]] [[#LOADED:]] [[#ALLOCA]]
+; CHECK-SPIRV-UNTYPED: FunctionPointerCallINTEL [[#]] [[#]] [[#LOADED]]
```
vmaksimo: Now redundant after the merge above -- delete this whole block.
```suggestion
```

test/extensions/INTEL/SPV_INTEL_function_pointers/select.ll
```diff
@@ -1,25 +1,48 @@
...
+; RUN: %if spirv-backend %{ FileCheck %s --check-prefixes=CHECK-LLVM-LLC < %t.llc.rev.ll %}
+
+; TODO: reader currently crashes with output of llc with untyped pointers for this test
+; RUN: %if spirv-backend %{ llc -O0 -mtriple=spirv64-unknown-unknown --spirv-ext=+SPV_INTEL_function_pointers,+SPV_KHR_untyped_pointers -filetype=obj %s -o %t.llc.u.spv %}
+; RUNx: %if spirv-backend %{ llvm-spirv -r %t.llc.u.spv -o %t.llc.u.rev.bc %}
+
+; CHECK-SPIRV-UNTYPED: Capability UntypedPointersKHR
+; CHECK-SPIRV-UNTYPED: Capability FunctionPointersINTEL
+; CHECK-SPIRV-UNTYPED: Extension "SPV_INTEL_function_pointers"
+; CHECK-SPIRV-UNTYPED: Extension "SPV_KHR_untyped_pointers"
```
vmaksimo, line 26:
```suggestion
; CHECK-SPIRV-UNTYPED: Capability UntypedPointersKHR
; CHECK-SPIRV: Capability FunctionPointersINTEL
; CHECK-SPIRV: Extension "SPV_INTEL_function_pointers"
; CHECK-SPIRV-UNTYPED: Extension "SPV_KHR_untyped_pointers"
```

test/extensions/INTEL/SPV_INTEL_function_pointers/global-function-pointer.ll
```diff
@@ -2,6 +2,16 @@
 ; RUN: llvm-spirv %t.spv -spirv-ext=+SPV_INTEL_function_pointers -to-text -o - | FileCheck %s --check-prefix=CHECK-SPIRV
 ; RUN: llvm-spirv -r %t.spv -o - | llvm-dis | FileCheck %s --check-prefix=CHECK-LLVM
```
vmaksimo: This file never got an untyped-pointers run added

vmaksimo (commented):
Happy to see the issue fixed in the right place!
But I reviewed the tests and waiting for the fixups there

lib/SPIRV/SPIRVReader.cpp
Qi-Ye-079: Done

test/extensions/INTEL/SPV_INTEL_function_pointers/CodeSectionINTEL/decor-func-ptr-arg-attr.ll
Qi-Ye-079: Done

test/extensions/INTEL/SPV_INTEL_function_pointers/CodeSectionINTEL/decor-func-ptr-arg-attr.ll
Qi-Ye-079: Yeah I think we discussed this earlier, that this is a bug in the writer that we should fix later in a separate PR.

test/extensions/INTEL/SPV_INTEL_function_pointers/CodeSectionINTEL/fp-from-host.ll
Qi-Ye-079: Done

test/extensions/INTEL/SPV_INTEL_function_pointers/CodeSectionINTEL/function-pointer-as-function-arg.ll
Qi-Ye-079: done

test/extensions/INTEL/SPV_INTEL_function_pointers/CodeSectionINTEL/function-pointer-as-function-arg.ll
Qi-Ye-079: Done

test/extensions/INTEL/SPV_INTEL_function_pointers/CodeSectionINTEL/function-pointer-dedicated-as.ll
Qi-Ye-079: Done

test/extensions/INTEL/SPV_INTEL_function_pointers/CodeSectionINTEL/function-pointer.ll
Qi-Ye-079: Done

test/extensions/INTEL/SPV_INTEL_function_pointers/CodeSectionINTEL/function-pointer.ll
Qi-Ye-079: Done

test/extensions/INTEL/SPV_INTEL_function_pointers/select.ll
Qi-Ye-079: Done

test/extensions/INTEL/SPV_INTEL_function_pointers/global-function-pointer.ll
Qi-Ye-079: Done

lib/SPIRV/SPIRVWriter.cpp
Qi-Ye-079: I already reverted this unrelated change from my latest commit a827970bbfd5e58de7e642ec440ce0b6c8a697c0. These 3 lines had CRLF ending but I accidentally changed them to LF ending. Now it's been reverted.

vmaksimo (approved):

PR #3840, opened 2026-07-06
Optimize checkError

MrSidims:
Inspired by recent discussion whether we should validate each instruction and what is cost for it - surprisingly the checks themselves were cheap, the cost came from the way checkError was written.

The function constructed a std::stringstream unconditionally, *before* the early-out taken when the condition is satisfied, and the SPIRVCK / SPIRVCKRT macros built a std::string ("std::string() + (ErrMsg)") at every call site. Both costs were therefore paid even on the error check success path.

Moved "std::stringstream SS" past the early-out in checkError(). Added a const char* checkError overload that returns before constructing any std::string when the condition holds.
Changed SPIRVCK/SPIRVCKRT to pass the message through directly.

On OpVariable-heavy modules it improves translation time by 5-10%.

Assisted-By: Claude Fable 5

### DIFF
--- a/lib/SPIRV/libSPIRV/SPIRVError.h
+++ b/lib/SPIRV/libSPIRV/SPIRVError.h
@@ -53,27 +53,24 @@ namespace SPIRV {
 // Emit absolute path only in debug mode.
 #ifdef NDEBUG
 #define SPIRVCK(Condition, ErrCode, ErrMsg)                                    \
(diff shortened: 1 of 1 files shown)


MrSidims:
FYI @maarquitos14 @aobolensk @idubinov

idubinov (approved):

aobolensk (approved):

maarquitos14 (approved):

YuriPlyakhin (approved):
LGTM

PR #3850, opened 2026-07-10
Avoid per-instruction stream seeks in validateWordCount

MrSidims:
validateWordCount() seeked the input stream to its end and back on every module-scope instruction just to bounds-check the word count.

Cache the total stream size once in parseSPIRV() and bounds-check each instruction with a single tellg().

AI-assisted: Claude Fable 5

### DIFF
--- a/lib/SPIRV/libSPIRV/SPIRVModule.cpp
+++ b/lib/SPIRV/libSPIRV/SPIRVModule.cpp
@@ -110,6 +110,8 @@ public:
 
   // Error handling functions
   SPIRVErrorLog &getErrorLog() override { return ErrLog; }
+  // Total size of the input stream, cached once by parseSPIRV.
+  std::streamoff getInputStreamSize() const { return InputStreamSize; }
   SPIRVErrorCode getError(std::string &ErrMsg) override {
     return ErrLog.getError(ErrMsg);
   }
@@ -623,6 +625,8 @@ private:
   SPIRVIdToInstructionSetMap IdToInstSetMap;
   SPIRVIdToBuiltinSetMap IdBuiltinMap;
   SPIRVIdSet NamedId;
+  // Cached total byte size of the input stream.
+  std::streamoff InputStreamSize = 0;
   SPIRVStringVec StringVec;
   SPIRVMemberNameVec MemberNameVec;
   std::shared_ptr<const SPIRVLine> CurrentLine;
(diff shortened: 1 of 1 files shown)


MrSidims:
@aobolensk @idubinov please take a look

MrSidims:
It's continuation of reflection, how llvm-spirv parses the input module. I'm currently focusing on something,
that always looked sub-optimal in the translator, like translation of decorations and forward references. Current
effort affects only the reader part, as SPIR-V backend is somewhat product ready and we have switched to it by default
for HIP compiler. But I do believe some of the future patches might give some food to think about forward translation improvements if anybody will have interest.

For this patch - global variable and decorations heavy modules are now being translated significantly faster (on my toy benchmarks ~20% faster).

maarquitos14 (approved):

aobolensk (approved):
Approach LGTM, but I think comments you added could be omitted

lib/SPIRV/libSPIRV/SPIRVModule.cpp
```diff
@@ -2450,11 +2454,9 @@ void SPIRVModuleImpl::addUnknownStructField(SPIRVTypeStruct *Struct, unsigned I,
...
                               SPIRVWord WordCount) {
   if (!SPIRVUseTextFormat) {
-    std::streampos CurrentPos = IS.tellg();
-    IS.seekg(0, std::ios::end);
-    std::streamoff RemainingBytes = IS.tellg() - CurrentPos;
-    IS.clear();
-    IS.seekg(CurrentPos);
+    // Bounds-check against the total stream size.
+    std::streamoff RemainingBytes =
+        M.getInputStreamSize() - static_cast<std::streamoff>(IS.tellg());
```
idubinov: std::tellg is implicitly convertible to std::streamoff , i expect no static cast needed

idubinov (approved):
Cool! nice catch, nice optimization, 20% sounds amazing

have only one not significant comment, which should not be blocker for merge.
I am a bit confused that information about input stream is stored in a module and there is no binding between them, so validateWordCount can be potentially used with wrong pair of [module, istream]. but due to the fact validate function is used in only one place - no matter.

lib/SPIRV/libSPIRV/SPIRVModule.cpp
MrSidims: applied

YuriPlyakhin (approved):
LGTM (agreed with comment on a potential problem with a mismatched [module, istream] pair, but not a blocker)
overall great performance improvement!

svenvh (approved):

MrSidims:
restarting CI

PR #3851, opened 2026-07-10
Early-out transIntelFPGADecorations for undecorated values

MrSidims:
Return early when the value has no decorations of its own and is not an alloca of struct type.


MrSidims:
Same effort as [here](https://github.com/KhronosGroup/SPIRV-LLVM-Translator/pull/3850#issuecomment-4935674452). Probably something can be improved on the forward path. Perf improvements here are more modest tho, ~2% on decorations-heavy modules perf my benchmarks. 

@YuriPlyakhin may I ask someone from Intel to take a look?

maarquitos14 (approved):
LGTM. I assume there already tests for this, right?

MrSidims:
> I assume there already tests for this, right?

For correctness - there are already plenty of decorations-related tests in the repository.

lib/SPIRV/SPIRVReader.cpp
```diff
@@ -4432,6 +4432,14 @@ void SPIRVToLLVM::transIntelFPGADecorations(SPIRVValue *BV, Value *V) {
     auto *AL = dyn_cast<AllocaInst>(Inst);
     Type *AllocatedTy = AL ? AL->getAllocatedType() : Inst->getType();
 
+    // A value with no decorations of its own can contribute Intel FPGA
+    // annotations only through struct-member decorations on an alloca of
+    // struct type. Skip the boiler-plate work for the rest undecorated values.
```
YuriPlyakhin: nit
```suggestion
    // struct type. Skip the boilerplate work for all other undecorated values.
```

lib/SPIRV/SPIRVReader.cpp
```diff
@@ -4432,6 +4432,14 @@ void SPIRVToLLVM::transIntelFPGADecorations(SPIRVValue *BV, Value *V) {
...
+    const bool MaybeStructMemberAnnot =
+        AL && BV->getType()->getPointerElementType()->isTypeStruct();
```
YuriPlyakhin, line 4439: nit
The same check is used in line 4455
```
 if (AL && BV->getType()->getPointerElementType()->isTypeStruct()) {
```
We can reuse `MaybeStructMemberAnnot`

YuriPlyakhin (approved):
couple nits. LGTM.

PR #3852, opened 2026-07-10
Skip writer-only lowering path when reading SPIR-V

MrSidims:
NamedId is consumed only by the writer, so maintaining it while decoding an existing SPIR-V module is dead work. Add an explicit IsReading flag, set by parseSPIRV()/parseSPT().

Also gate addDecorate()'s required-capability collection on AutoAddCapability, mirroring addEntry().

AI-assisted: Claude Fable 5

### DIFF
--- a/lib/SPIRV/libSPIRV/SPIRVModule.cpp
+++ b/lib/SPIRV/libSPIRV/SPIRVModule.cpp
@@ -108,6 +108,8 @@ public:
     return CurrentLine.get() || !DebugInstVec.empty();
   }
 
+  void setReadingModule() override { IsReading = true; }
+
   // Error handling functions
   SPIRVErrorLog &getErrorLog() override { return ErrLog; }
   SPIRVErrorCode getError(std::string &ErrMsg) override {
@@ -2547,7 +2549,6 @@ std::istream &SPIRVModuleImpl::parseSPT(std::istream &I) {
   SPIRVModuleImpl &MI = *this;
   MI.setAutoAddCapability(false);
   MI.setAutoAddExtensions(false);
-  MI.IsReading = true;
   auto ReadSPIRVWord = [](std::istream &I) {
     uint32_t W;
     I >> skipcomment >> W;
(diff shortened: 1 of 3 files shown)


MrSidims:
Continuation of https://github.com/KhronosGroup/SPIRV-LLVM-Translator/pull/3850#issuecomment-4935674452 and the last patch of this effort.

MrSidims:
@aobolensk @idubinov please take a look

idubinov (approved):
LGTM

lib/SPIRV/libSPIRV/SPIRVModule.cpp
```diff
@@ -108,6 +108,8 @@ class SPIRVModuleImpl : public SPIRVModule {
     return CurrentLine.get() || !DebugInstVec.empty();
   }
 
+  void setReadingModule() override { IsReading = true; }
```
aobolensk, line 111: Q: Will we possibly need an opportunity to toggle this mode?

aobolensk (commented):
Please, remove comments, they are also unnecessary here

lib/SPIRV/libSPIRV/SPIRVModule.cpp
idubinov: as for me, by now - no. only to just arrange the interface with other functions like setAutoAddCapability and others

YuriPlyakhin (commented):
PR description needs update: `Add an explicit IsReading flag, set by parseSPIRV()/parseSPT()`
It is now set by `readSpirvModule()`

YuriPlyakhin (commented):
The PR breaks the `SPV_INTEL_function_variants` write-back path.
When `--fnvar-spv-out` is used, `readSpirv()` re-serializes the same `BM` via `OFSSpv << *BM` (SPIRVReader.cpp:6036).

Since the writer emits `OpNames` only by iterating `NamedId` (SPIRVModule.cpp:2347), and `NamedId` is not populated while `IsReading` is set, all original `OpNames` are dropped from the specialized output.

I verified it with this change to `test/extensions/INTEL/SPV_INTEL_function_variants/cl_dot4.spt`:

   ; CHECK-SPIRV-COMMON: EntryPoint
  +; OpName decorations must survive the read/specialize/write round-trip
  +; CHECK-SPIRV-COMMON: Name {{[0-9]+}} "C"
   ; CHECK-SPIRV-FP16: TypeFloat {{[0-9]+}} 16

The existing tests don't CHECK any Name lines, so CI doesn't catch this.

Ideas on how to solve without `IsReading` flag (there are likely performance trade-offs though):
Can writer derive `OpNames` directly from `IdEntryMap` at emit time instead of the pre-maintained `NamedId` set? We would need to sort them likely, before emitting.
Or maybe we could populate `NamedId` from the entries at the start of the write by iterating through `IdEntryMap`?

PR #3858, opened 2026-07-10
[Backport to 16] .clang-tidy: disable conflicting misc-use-anonymous-namespaces (#3286)

Qi-Ye-079:
Backport of #3286.

Remove checks `misc-use-anonymous-namespace`, otherwise we'll be forced to place static functions into anonymous namespace, which contridicts with current LLVM practice and leads to inconsistent style across release branches during backporting.


YuriPlyakhin (approved):
LGTM

PR #3861, opened 2026-07-14
Hook up in-tree spirv-tools build for testing if LLVM_INCLUDE_SPIRV_TOOLS_TESTS is ON

wenju-he:
Tests currently locate spirv-val/as/dis/link via a system-installed SPIRV-Tools package (pkg-config or find_package), so machines without SPIRV-Tools silently skipped SPIR-V validation. The llvm-project repo has an opt-in in-tree SPIRV-Tools build (llvm/tools/spirv-tools, gated by LLVM_INCLUDE_SPIRV_TOOLS_TESTS). SPIRV-LLVM-Translator can leverage it when built in-tree as part of llvm-project.

Skip the pkg-config/find_package probing for a system SPIRV-Tools in root CMakeLists.txt when LLVM_INCLUDE_SPIRV_TOOLS_TESTS is on and this project is built in-tree (NOT LLVM_SPIRV_BUILD_EXTERNAL). The guard keeps standalone (out-of-tree) builds unaffected even if LLVM_INCLUDE_SPIRV_TOOLS_TESTS is set in the CMake cache, since a standalone build has no in-tree spirv-as/dis/val/link targets to fall back on. In-tree, skipping the probe avoids a case where a machine that also happens to have a SPIRV-Tools install discoverable via PATH/CMAKE_PREFIX_PATH triggers find_package(SPIRV-Tools-tools), which imports all spirv-tools targets while the in-tree build has already defined spirv-as/dis/val/link, and CMake fatal-errors with "Some (but not all) targets in this export set were already defined."


test/CMakeLists.txt
```diff
@@ -129,6 +146,14 @@ if(NOT LLVM_SPIRV_BUILD_EXTERNAL)
 endif(NOT LLVM_SPIRV_BUILD_EXTERNAL)
 
 
+if(LLVM_INCLUDE_SPIRV_TOOLS_TESTS AND NOT SPIRV_TOOLS_FOUND AND NOT SPIRV-Tools-tools_FOUND)
+  foreach(tool spirv-as spirv-dis spirv-link spirv-val)
+    if(TARGET ${tool})
```
Qi-Ye-079: Correct me if I'm wrong, but I doubt if the checks here will work correctly for in-tree build. I think the issue is about the configuration order of `SPIRV-LLVM-Translator` vs `llvm-project/llvm/tools/spirv-tools` for in-tree build. From the `llvm-project/llvm/CMakeLists.txt`:

1431: add_subdirectory(projects)   # llvm/projects/SPIRV-LLVM-Translator
1434: add_subdirectory(tools)      # llvm/tools/spirv-tools  ← option + spirv-* targets defined here

So the translator is always configured _before_ `llvm/tools/spirv-tools`. This means, by the time cmake configures the translator, the `LLVM_INCLUDE_SPIRV_TOOLS_TESTS` option isn't defined yet, and all these `spirv-as/dis/link/val` targets aren't even available.

If `-DLLVM_INCLUDE_SPIRV_TOOLS_TESTS=ON` is set via command line, you're still faced with unavailable spirv-* targets at this point, so those `if (TARGET ...)` checks will always return false.

I appreciate the effort for a long-term solution to get past the old `spirv-val` issue in our environment, but given the current ordering of configuration for in-tree build, I don't know if there is an elegant and safe way to get around this.

Ping @YuriPlyakhin to follow-up as well.

wenju-he: thanks @Qi-Ye-079, it is a good catch of the configuration order and these targets in SPIR-Tools don't exist yet at this moment. I've removed the target existence check and assume they will be set-up later when LLVM_INCLUDE_SPIRV_TOOLS_TESTS is ON.

> by the time cmake configures the translator, the LLVM_INCLUDE_SPIRV_TOOLS_TESTS option isn't defined yet

This is actually safe. If the user passes -DLLVM_INCLUDE_SPIRV_TOOLS_TESTS=ON on the command line, CMake populates the cache with that value before it processes a single CMakeLists.txt and later `option(LLVM_INCLUDE_SPIRV_TOOLS_TESTS ...)` in `llvm/tools/spirv-tools` is a no-op as the variable is already in the cmake cache. Its default value in `llvm/tools/spirv-tools` only take effect if command line options doesn't specify value for the macro.

wenju-he: Right, the ordering issue is hard to resolve. There is similar w/a in polly/test/CMakeLists.txt

test/CMakeLists.txt
```diff
@@ -64,6 +64,28 @@ elseif(SPIRV-Tools-tools_FOUND)
   endif()
 
   set(SPIRV_TOOLS_BINDIR "${SPIRV_AS_PATH}")
+elseif(LLVM_INCLUDE_SPIRV_TOOLS_TESTS)
```
YuriPlyakhin: Should we add `NOT LLVM_SPIRV_BUILD_EXTERNAL` guard to both places here as well, to avoid problem with a standalone build with `-DLLVM_INCLUDE_SPIRV_TOOLS_TESTS=ON` flag passed and no system SPIRV-Tools found?

CMakeLists.txt
```diff
@@ -121,8 +121,10 @@ endif()
 set(LLVM_SPIRV_INCLUDE_DIRS ${CMAKE_CURRENT_SOURCE_DIR}/include)
 
 # first try locating SPIRV-Tools via pkgconfig (the old way)
-pkg_search_module(SPIRV_TOOLS SPIRV-Tools)
-if (NOT SPIRV_TOOLS_FOUND)
+if (NOT (LLVM_INCLUDE_SPIRV_TOOLS_TESTS AND NOT LLVM_SPIRV_BUILD_EXTERNAL))
```
YuriPlyakhin: This condition `NOT (LLVM_INCLUDE_SPIRV_TOOLS_TESTS AND NOT LLVM_SPIRV_BUILD_EXTERNAL)` is duplicated on two lines. I suggest making a new var: `LLVM_SPIRV_USE_INTREE_SPIRV_TOOLS` to keep those 2 checks in sync.

CMakeLists.txt
```diff
@@ -121,8 +121,10 @@ endif()
...
+  pkg_search_module(SPIRV_TOOLS SPIRV-Tools)
+endif()
+if (NOT SPIRV_TOOLS_FOUND AND NOT (LLVM_INCLUDE_SPIRV_TOOLS_TESTS AND NOT LLVM_SPIRV_BUILD_EXTERNAL))
```
YuriPlyakhin: For in-tree build with `LLVM_INCLUDE_SPIRV_TOOLS_TESTS=ON`, `SPIRV_TOOLS_FOUND` will be always false.
It later has this check: `if (NOT SPIRV_TOOLS_FOUND AND LLVM_SPIRV_ENABLE_LIBSPIRV_DIS)`
It seems --spirv-tools-dis will be always disabled in that scenario.

CMakeLists.txt
wenju-he: added LLVM_SPIRV_USE_INTREE_SPIRV_TOOLS

test/CMakeLists.txt
wenju-he: added `NOT LLVM_SPIRV_BUILD_EXTERNAL` guard in root CMakeLists.txt‎ to control LLVM_SPIRV_USE_INTREE_SPIRV_TOOLS

CMakeLists.txt
wenju-he: good catch, fixed SPIRV_TOOLS_FOUND

test/CMakeLists.txt
wenju-he: there is precedent that llvm-spirv/test/CMakeLists.txt adds dependency on llc from llvm/tools

PR #3863, opened 2026-07-14
Fix OpGroupNonUniformRotateKHR mangling

mateuszchudyk:
According to https://github.khronos.org/SPIRV-Registry/extensions/KHR/SPV_KHR_subgroup_rotate.html both Delta and ClusterSize must be a scalar of integer type, whose Signedness operand is 0.


svenvh:
LGTM, thanks!  ~Could you add a small test for this (or extend an existing test)?~ Actually some tests are failing, so they just need an update.

mateuszchudyk:
The remaining failing tests:
```
  LLVM_SPIRV :: DebugInfo/LocalAddressSpace.ll
  LLVM_SPIRV :: DebugInfo/NonSemantic/Shader200/DebugInfoStringType.ll
  LLVM_SPIRV :: DebugInfo/X86/DW_AT_specification.ll
  LLVM_SPIRV :: DebugInfo/X86/dimodule-external-fortran.ll
```
are unrelated to the change and fail also on the HEAD.

svenvh (approved):

bashbaug (approved):
LGTM, I'll go ahead and merge this.

There is a bit of weirdness here that's a bit unfortunate, recording for completeness:

* The SPIR-V [OpGroupNonUniformRotateKHR](https://github.khronos.org/SPIRV-Registry/extensions/KHR/SPV_KHR_subgroup_rotate.html#OpGroupNonUniformRotateKHR) operand is explicitly unsigned ("Delta is treated as unsigned...").
* But, the OpenCL C [sub_group_rotate](https://registry.khronos.org/OpenCL/specs/unified/html/OpenCL_C.html#sub-group-rotate-functions) parameter is a signed integer ("sub_group_rotate(gentype value, int delta)").

I think it makes sense for the SPIR-V mangling to match the SPIR-V semantics though, so switching to an unsigned integer is the right thing to do, even though it's a change from the current behavior and it differs from the analogous OpenCL C function.

PR #3864, opened 2026-07-14
Update LLVM version from 23 to 24

svenvh:


svenvh:
@MrSidims / @vmaksimo do you think we should merge this now even though the required llvm apt packages are not available yet, or should we wait a bit longer?

YuriPlyakhin (approved):
LGTM, but what is the benefit of merging this one before apt packages are available?

svenvh:
> LGTM, but what is the benefit of merging this one before apt packages are available?

Thanks for the review!  Merging of this PR is already overdue because of the apt package delay, so I prefer to merge this real soon.  That will mark a clear llvm version boundary in the git history, as we have various llvm 24 fixes piling up now.

After all fixes have been merged, I can try a temporary workaround for in-tree builds to work around the missing apt packages.

Issue #645, opened 2020-07-19
SPIR-V translator is failed to translate "unsupported" vectors.


bader:
Here is an example of OpenCL program SPIR-V translator fails to translate because `InstCombine` pass changes vector type to another vector type with the same size, but different # of elements.
```C
long3 y;

kernel void foo(global int *x) {
    x[0] = y.x;
}
```
https://godbolt.org/z/oMM9qv - LLVM IR contains 6 - element vectors. Changing data types in this example we can get a lot of vector types, which will be rejected by SPIR-V translator if `SPV_INTEL_vector_compute` extension is not supported.

PR #3866, opened 2026-07-14
phi node: filter multiple entries for same predecessor

### LINKED ISSUE #3865: phi node with multiple entries for same predecessor not handled correctly
This is kind of a weird edge case, but in LLVM IR, you can have multiple entries for the same predecessor block in a phi node if a switch or branch has duplicate destinations. This was reduced from a real failure in the OpenCL.jl test suite when running with the translator backend.

```llvm
target datalayout = "e-i64:64-v16:16-v24:32-v32:32-v48:64-v96:128-v192:256-v256:256-v512:512-v1024:1024-G1"

define spir_kernel void @f() {
entry:
  br i1 undef, label %L815, label %L815

L815:                                             ; preds = %entry, %entry
  %v = phi i64 [ 0, %entry ], [ 0, %entry ]
  ret void
}
```

SPIRV-LLVM-Translator translates this verbatim:

```spirv
; SPIR-V
; Version: 1.0
; Generator: Khronos LLVM/SPIR-V Translator; 14
; Bound: 12
; Schema: 0
               OpCapability Addresses
               OpCapability Kernel
               OpCapability Int64
          %1 = OpExtInstImport "OpenCL.std"
               OpEntryPoint Kernel %4 "f"
               OpSource Unknown 0
               OpName %entry "entry"
               OpName %L815 "L815"
               OpName %v "v"
      %ulong = OpTypeInt 64 0
    %ulong_0 = OpConstant %ulong 0
       %void = OpTypeVoid
          %3 = OpTypeFunction %void
       %bool = OpTypeBool
          %8 = OpUndef %bool
          %4 = OpFunction %void None %3
      %entry = OpLabel
               OpBranchConditional %8 %L815 %L815
       %L815 = OpLabel
          %v = OpPhi %ulong %ulong_0 %entry %ulong_0 %entry
               OpReturn
               OpFunctionEnd

```

which then fails validation with:
```
> spirv-val test.spv
error: line 21: OpPhi's number of incoming blocks (2) does not match block's predecessor count (2).
  %v = OpPhi %ulong %ulong_0 %entry %ulong_0 %entry
```

I am not 100% sure whether this is just a bug in spirv-val, but the LLVM backend at least produces `%v = OpPhi %ulong %ulong_0 %entry` in line 21 here (I actually had to make `%v` the return and pass `-O0` for `llc` to not just optimize away the phi node)

simeonschaub:
fixes #3865

### DIFF
--- a/lib/SPIRV/SPIRVWriter.cpp
+++ b/lib/SPIRV/SPIRVWriter.cpp
@@ -2647,8 +2647,16 @@ LLVMToSPIRVBase::transValueWithoutDecoration(Value *V, SPIRVBasicBlock *BB,
                                    FuncTransMode::Pointer);
       if (Val->getType() != Ty)
         Val = BM->addUnaryInst(OpBitcast, Ty, Val, BB);
-      IncomingPairs.push_back(Val);
-      IncomingPairs.push_back(transValue(Phi->getIncomingBlock(I), nullptr));
+      SPIRVValue *Block = transValue(Phi->getIncomingBlock(I), nullptr);
+      bool IsDuplicate = false;
+      for (size_t Idx = 1; Idx < IncomingPairs.size(); Idx += 2) {
+        if (IncomingPairs[Idx] == Block)
+          IsDuplicate = true;
+      }
+      if (!IsDuplicate) {
+        IncomingPairs.push_back(Val);
+        IncomingPairs.push_back(Block);
+      }
     }
     return mapValue(V, BM->addPhiInst(Ty, IncomingPairs, BB));
   }
--- a/lib/SPIRV/SPIRVWriter.cpp
+++ b/lib/SPIRV/SPIRVWriter.cpp
@@ -2593,9 +2593,16 @@ LLVMToSPIRVBase::transValueWithoutDecoration(Value *V, SPIRVBasicBlock *BB,
                                LoopControl, Parameters, SuccessorTrue);
       }
     }
-    return mapValue(
-        V, BM->addBranchConditionalInst(transValue(Branch->getCondition(), BB),
-                                        SuccessorTrue, SuccessorFalse, BB));
+    // Starting with SPIR-V 1.6, the True Label and False Label of an
+    // "OpBranchConditional" must not be the same, so emit an unconditional
+    // branch.
+    if (SuccessorTrue != SuccessorFalse) {
+      return mapValue(
+          V, BM->addBranchConditionalInst(transValue(Branch->getCondition(), BB),
+                                          SuccessorTrue, SuccessorFalse, BB));
+    } else {
+      return mapValue(V, BM->addBranchInst(SuccessorTrue, BB));
+    }
   }
 
   if (auto *Branch = dyn_cast<UncondBrInst>(V)) {
--- a/lib/SPIRV/SPIRVWriter.cpp
+++ b/lib/SPIRV/SPIRVWriter.cpp
@@ -2595,7 +2595,7 @@ LLVMToSPIRVBase::transValueWithoutDecoration(Value *V, SPIRVBasicBlock *BB,
     }
     // Starting with SPIR-V 1.6, the True Label and False Label of an
     // "OpBranchConditional" must not be the same, so emit an unconditional
-    // branch.
+    // branch - always valid, required for 1.6+
     if (SuccessorTrue == SuccessorFalse)
       return mapValue(V, BM->addBranchInst(SuccessorTrue, BB));
 
--- a/lib/SPIRV/SPIRVWriter.cpp
+++ b/lib/SPIRV/SPIRVWriter.cpp
@@ -2596,13 +2596,12 @@ LLVMToSPIRVBase::transValueWithoutDecoration(Value *V, SPIRVBasicBlock *BB,
     // Starting with SPIR-V 1.6, the True Label and False Label of an
     // "OpBranchConditional" must not be the same, so emit an unconditional
     // branch.
-    if (SuccessorTrue != SuccessorFalse) {
-      return mapValue(
-          V, BM->addBranchConditionalInst(transValue(Branch->getCondition(), BB),
-                                          SuccessorTrue, SuccessorFalse, BB));
-    } else {
+    if (SuccessorTrue == SuccessorFalse)
       return mapValue(V, BM->addBranchInst(SuccessorTrue, BB));
-    }
+
+    return mapValue(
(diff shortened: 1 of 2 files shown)


test/phi-duplicate-predecessors.ll
```diff
@@ -0,0 +1,22 @@
+; REQUIRES: spirv-dis
+; RUN: llvm-as %s -o %t.bc
+; RUN: llvm-spirv %t.bc -o %t.spv
+; RUN: spirv-val %t.spv
+; RUN: spirv-dis %t.spv | FileCheck %s
+
+; CHECK: %entry = OpLabel
+; CHECK: OpBranchConditional %[[NULL:[0-9]+]] %L815 %L815
+; CHECK: %L815 = OpLabel
+; CHECK: %v = OpPhi %ulong %ulong_0 %entry
```
MrSidims:
```suggestion
; CHECK: %[[Entry:[a-z0-9_]+]] = OpLabel
; CHECK: OpBranchConditional %[[#NULL:]] %[[Lable:[a-z0-9_]+]] %[[Lable]]
; CHECK: %[[Lable]] = OpLabel
; CHECK: %{{.*}} = OpPhi %{{.*}} %{{.*}} %[[Entry]]
```

test/phi-duplicate-predecessors.ll
```diff
@@ -0,0 +1,22 @@
+; REQUIRES: spirv-dis
+; RUN: llvm-as %s -o %t.bc
+; RUN: llvm-spirv %t.bc -o %t.spv
+; RUN: spirv-val %t.spv
+; RUN: spirv-dis %t.spv | FileCheck %s
```
MrSidims: Please also add llvm-spirv -r RUN and check, so we make sure, we don't have a bug on the reverse path.

MrSidims (approved):
LGTM, thank you!

test/phi-duplicate-predecessors.ll
MrSidims: (tho I never remember string variable declaration in LIT spelling, better to double check)

simeonschaub:
The In-tree failure looks real, I wonder if this assert might be too strict?

https://github.com/KhronosGroup/SPIRV-LLVM-Translator/blob/880cbb08c8d3c260800feb53a1c18537e9ecdbdb/lib/SPIRV/libSPIRV/SPIRVInstruction.h#L1017

simeonschaub:
Bump, any ideas what to do about the failing assertion? Should it just be deleted?

MrSidims:
> Bump, any ideas what to do about the failing assertion? Should it just be deleted?

Apologies, will try to take a look by EOW.

MrSidims:
@simeonschaub assert is legit, it reflects "#645: The True Label and False Label of an [OpBranchConditional](https://registry.khronos.org/SPIR-V/specs/unified1/SPIRV.html#OpBranchConditional) must not be the same, starting with version 1.6.", see https://registry.khronos.org/SPIR-V/specs/unified1/SPIRV.html#_changes_from_version_1_5 - not sure why spirv-val is passing, probably because generated module is 1.0 (but I'm always forgetting the exact mechanism the translator picks the version when none are specified ;-P) - would be nice to add to the test max version enforcement to 1.6.

Regarding what to do with it. IMHO - just drop? Such a branch is equivalent to an unconditional one - the condition is a plain value, evaluating it has no side effects - so emit OpBranch instead. WDYT?

simeonschaub:
Makes sense! I have added special handling for conditional branches where both successors are the same and adjusted the test

MrSidims:
Cool! I'm ramping down my contributions to llvm-spirv tool, so would be nice to get feedback from @svenvh and @vmaksimo

lib/SPIRV/SPIRVWriter.cpp
```diff
@@ -2593,6 +2593,12 @@ LLVMToSPIRVBase::transValueWithoutDecoration(Value *V, SPIRVBasicBlock *BB,
                                LoopControl, Parameters, SuccessorTrue);
       }
     }
+    // Starting with SPIR-V 1.6, the True Label and False Label of an
+    // "OpBranchConditional" must not be the same, so emit an unconditional
+    // branch.
+    if (SuccessorTrue == SuccessorFalse)
+      return mapValue(V, BM->addBranchInst(SuccessorTrue, BB));
+
```
MrSidims: May be the assertion should also be guarded with IF SPIR_V_1_6 check, 

**UPD**: though, with this replacement it should never fire (which is exactly the purpose of assertions). So I don't have a strong opinion about it actually. May leave the assertion as is - your call :)

simeonschaub: My thought is that this was probably just left under-specified before SPIR-V 1.6, so it's probably good to assert nonetheless to prevent any issues down the road?

test/phi-duplicate-predecessors.ll
```diff
@@ -0,0 +1,31 @@
+; REQUIRES: spirv-dis
+; RUN: llvm-as %s -o %t.bc
+; RUN: llvm-spirv %t.bc -o %t.spv
```
vmaksimo:
```suggestion
; RUN: llvm-spirv %s -o %t.spv
```
just a small suggestion

lib/SPIRV/SPIRVWriter.cpp
```diff
@@ -2593,6 +2593,12 @@ LLVMToSPIRVBase::transValueWithoutDecoration(Value *V, SPIRVBasicBlock *BB,
                                LoopControl, Parameters, SuccessorTrue);
       }
     }
+    // Starting with SPIR-V 1.6, the True Label and False Label of an
+    // "OpBranchConditional" must not be the same, so emit an unconditional
+    // branch.
```
vmaksimo: The transform here is applied with any SPIR-V version, maybe we can reword the comment to avoid the confusion, e.g. like:
```
... emit an unconditional branch - always valid, required for 1.6+
```

vmaksimo (commented):
LGTM, just few things to improve

lib/SPIRV/SPIRVWriter.cpp
simeonschaub:
```suggestion
    // Starting with SPIR-V 1.6, the True Label and False Label of an
    // "OpBranchConditional" must not be the same, so emit an unconditional
    // branch - always valid, required for 1.6+
```

vmaksimo (approved):

PR #3782, opened 2026-06-17
Update SPV_INTEL_fp_conversions to the latest spec revision

vmaksimo:
- Split FloatConversionsINTEL (6215) into FloatConversionsFtoFINTEL (6215) and FloatConversionsFtoSINTEL (6216).
- Drop OpClampConvertFToFINTEL and OpClampStochasticRoundFToFINTEL. Their behavior is now expressed as OpFConvert / OpStochasticRoundFToFINTEL decorated with SaturatedToLargestFloat8NormalConversionEXT (fp8) or unconditionally clamped (fp4).
- Drop ClampConvert\*ToE2M1INTEL builtins - fp4 saturation is unconditional, so they collapse into Convert*ToE2M1INTEL.
- Update LIT tests and OCPTypesRepresentationInLLVM.rst accordingly. Add an FtoS-only test verifying FloatConversionsFtoFINTEL is not declared when only FtoS instructions are used.

Same-release breakage of consumers using removed opcodes / capability name is acceptable -- no legacy decode path is retained.

Spec:
https://github.com/intel/llvm/blob/sycl/sycl/doc/design/spirv-extensions/SPV_INTEL_fp_conversions.asciidoc

AI-assisted: Claude Opus 4.7 (commercial SaaS)

PR #3868, opened 2026-07-15
Add 2 opcodes back from old SPV_INTEL_fp_conversions spec

Qi-Ye-079:
Temporarily add 2 opcodes back from old spec of `SPV_INTEL_fp_conversions`, for backward compatibility of SPIRV Reader only:
- `OpClampConvertFToFINTEL`: must translate to the same LLVM IR as `OpFConvert` + `SaturatedToLargestFloat8NormalConversionEXT`
- `OpClampStochasticRoundFToFINTEL`: must translate to the same LLVM IR as `OpStochasticRoundFToFINTEL` + `SaturatedToLargestFloat8NormalConversionEXT`.

The Writer keeps unchanged and emits the current ops from latest spec.

### DIFF
--- a/docs/OCPTypesRepresentationInLLVM.rst
+++ b/docs/OCPTypesRepresentationInLLVM.rst
@@ -145,26 +145,6 @@ decoration (clamp + stochastic rounding):**
   __builtin_spirv_ClampStochasticRoundFP16ToE5M2INTEL, __builtin_spirv_ClampStochasticRoundFP16ToE4M3INTEL,
   __builtin_spirv_ClampStochasticRoundBF16ToE5M2INTEL, __builtin_spirv_ClampStochasticRoundBF16ToE4M3INTEL
 
-The ``OpClampConvertFToFINTEL`` and ``OpClampStochasticRoundFToFINTEL`` instructions below belong to a
-previous revision of this extension. They are re-added **temporarily** for backward compatibility: the
-*Reader* still accepts them, while the *Writer* only emits the current revision above.
-
-**Translated to OpClampConvertFToFINTEL, the equivalent of OpFConvert decorated with
-SaturatedToLargestFloat8NormalConversionEXT (clamp rounding):**
-
-.. code-block:: C
-
-  __builtin_spirv_ClampConvertFP16ToE4M3INTEL, __builtin_spirv_ClampConvertBF16ToE4M3INTEL,
-  __builtin_spirv_ClampConvertFP16ToE5M2INTEL, __builtin_spirv_ClampConvertBF16ToE5M2INTEL
-
-**Translated to OpClampStochasticRoundFToFINTEL, the equivalent of OpStochasticRoundFToFINTEL decorated with
-SaturatedToLargestFloat8NormalConversionEXT (clamp + stochastic rounding):**
-
-.. code-block:: C
-
-  __builtin_spirv_ClampStochasticRoundFP16ToE5M2INTEL, __builtin_spirv_ClampStochasticRoundFP16ToE4M3INTEL,
-  __builtin_spirv_ClampStochasticRoundBF16ToE5M2INTEL, __builtin_spirv_ClampStochasticRoundBF16ToE4M3INTEL
-
 
 Example LLVM IR to SPIR-V translation:
 Input LLVM IR
--- a/docs/OCPTypesRepresentationInLLVM.rst
+++ b/docs/OCPTypesRepresentationInLLVM.rst
@@ -145,6 +145,26 @@ decoration (clamp + stochastic rounding):**
   __builtin_spirv_ClampStochasticRoundFP16ToE5M2INTEL, __builtin_spirv_ClampStochasticRoundFP16ToE4M3INTEL,
   __builtin_spirv_ClampStochasticRoundBF16ToE5M2INTEL, __builtin_spirv_ClampStochasticRoundBF16ToE4M3INTEL
 
+The ``OpClampConvertFToFINTEL`` and ``OpClampStochasticRoundFToFINTEL`` instructions below belong to a
+previous revision of this extension. They are re-added **temporarily** for backward compatibility: the
+*Reader* still accepts them, while the *Writer* only emits the current revision above.
+
+**Translated to OpClampConvertFToFINTEL, the equivalent of OpFConvert decorated with
+SaturatedToLargestFloat8NormalConversionEXT (clamp rounding):**
+
+.. code-block:: C
+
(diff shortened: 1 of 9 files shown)


Qi-Ye-079:
ping @YuriPlyakhin and @bashbaug

docs/OCPTypesRepresentationInLLVM.rst
bashbaug, line 1: The changes in this file can be reverted.  We don't need to support translation **to** these old opcodes.  We only need to be able to translate **from** them, to preserve compatibility for SPIR-V files in the wild that are still using them.

lib/SPIRV/libSPIRV/spirv_internal.hpp
```diff
@@ -68,7 +68,9 @@ enum InternalOp {
   IOpCooperativeMatrixStoreCheckedINTEL = 6194,
   IOpCooperativeMatrixConstructCheckedINTEL = 6195,
   IOpTypeTaskSequenceINTEL = 6199,
+  IOpClampConvertFToFINTEL = 6216,
   IOpStochasticRoundFToFINTEL = 6217,
+  IOpClampStochasticRoundFToFINTEL = 6218,
```
bashbaug: Could we add a comment here so we don't forget that these are old opcodes for backward compatibility only?

At some point (hopefully soon) we'll add support for SPV_INTEL_fp_conversions into the upstream headers.  When that happens, we can remove most of the support for SPV_INTEL_fp_conversions from this file and use the support from the upstream headers instead, but we'll need to keep these lines until we're confident that they're no longer needed for backwards compatibility.

```suggestion
  IOpClampConvertFToFINTEL = 6216, // old opcode, for backward compatibility
  IOpStochasticRoundFToFINTEL = 6217,
  IOpClampStochasticRoundFToFINTEL = 6218, // old opcode, for backward compatibility
```

lib/SPIRV/SPIRVReader.cpp
```diff
@@ -1095,12 +1103,17 @@ Value *SPIRVToLLVM::transConvertInst(SPIRVValue *BV, Function *F,
...
+              DecorationSaturatedToLargestFloat8NormalConversionEXT) ||
+          (IsOldConvertFToFOp &&
+           (DstEnc == FPEncodingWrap::E4M3 || DstEnc == FPEncodingWrap::E5M2));
       if (IsSaturatedFP8) {
         BM->getErrorLog().checkError(
             (OC == OpFConvert || OC == OpConvertSToF || OC == OpConvertUToF ||
-             OC == internal::OpStochasticRoundFToFINTEL) &&
+             OC == internal::OpStochasticRoundFToFINTEL ||
+             OC == internal::OpClampConvertFToFINTEL ||
+             OC == internal::OpClampStochasticRoundFToFINTEL) &&
```
YuriPlyakhin:
```suggestion
             IsOldConvertFToFOp) &&
```

lib/SPIRV/SPIRVReader.cpp
YuriPlyakhin, line 1123: This error message will be misleading for those who use `OpClampConvertFToFINTEL` and `OpClampStochasticRoundFToFINTEL` and not `SaturatedToLargestFloat8NormalConversionEXT`

YuriPlyakhin: should we also return back `internal::OpClampConvertFToFINTEL` and `internal::OpClampStochasticRoundFToFINTEL` to handle other types with this instruction? otherwise, that part is still broken, comparing with the state before https://github.com/KhronosGroup/SPIRV-LLVM-Translator/pull/3782. @bashbaug , what do you think?

YuriPlyakhin, line 4121: `transSPIRVBuiltinFromInst` also was handling `OpClampConvertFToFINTEL` and `OpClampStochasticRoundFToFINTEL`. Should it be rolled back as well? tagging also @bashbaug

YuriPlyakhin: figured this one is related to the question above, which about the code that is calling `transSPIRVBuiltinFromInst`

docs/OCPTypesRepresentationInLLVM.rst
Qi-Ye-079: Done

lib/SPIRV/libSPIRV/spirv_internal.hpp
Qi-Ye-079: Done

lib/SPIRV/SPIRVReader.cpp
Qi-Ye-079: Excluded the 2 old opcodes from checkError - see comment for error message below.

lib/SPIRV/SPIRVReader.cpp
Qi-Ye-079: I just realized that there's no need to check error for the 2 old opcodes, because:
1. For these 2 opcodes, `IsSaturatedFP8` solely depends on their destination type, and has nothing to do with the `SaturatedToLargestFloat8NormalConversionEXT` decoration (and the decoration could never be allowed to add to these 2 old opcodes anyway, before #3782 )
2. All the other ops keep the check unchanged, because `IsSaturatedFP8` is determined by the decoration, thus checking the valid opcodes and destination types are necessary.

So I refactored the code to skip the `checkError` for the 2 opcodes.

Qi-Ye-079: Done. Added it back.

Qi-Ye-079: Done. Added it back.

Qi-Ye-079: Actually, I think it's still better to add another `checkError` to explicitly check that the decoration can't be added on the 2 old opcodes, for completeness and explicitness to fully align with pre #3782 code.
I'd better also add a negative test that checks this error for the 2 opcodes carrying the decoration, since I don't think there's existing coverage for this particular scenario and opcodes before #3782 .

UPD: done.

bashbaug (approved):

bashbaug:
Thanks, latest version LGTM.

@svenvh @MrSidims, mind giving this a quick review and merge?  This reverts part of #3782 so we don't break any SPIR-V modules in the wild that are still targeting the prior spec.  Since it's for a vendor extension and it's largely reverting to prior behavior I don't think it will be complicated or contentious, but better safe than sorry.  Thanks!

MrSidims (approved):

svenvh (approved):
Looks reasonable; I'll merge this on the assumption that the author of #3782 (@vmaksimo) wouldn't object.

YuriPlyakhin (commented):
LGTM. Thanks!

PR #3869, opened 2026-07-16
Add missing coverage for SPV_EXT_long_vector

Qi-Ye-079:
This is a quick bug fix PR, as a follow-up of https://github.com/KhronosGroup/SPIRV-LLVM-Translator/pull/3777, which didn't cover the early-out path in  SPIRVLowerBitCastToNonStandardType.cpp. As a consequence, if the pass has vector with non-standard size as input but with only `SPV_EXT_long_vector` enabled, it will crash later.

Also added missing coverage in SPIRVUtils.cpp when handling intrinsics.

The new tests simply mirror the ones in `extensions/INTEL/SPV_INTEL_vector_compute`.

### DIFF
--- a/test/extensions/EXT/SPV_EXT_long_vector/lower-non-standard-vec-abs-with-ext.ll
+++ b/test/extensions/EXT/SPV_EXT_long_vector/lower-non-standard-vec-abs-with-ext.ll
@@ -1,0 +1,42 @@
+; RUN: llvm-as %s -o %t.bc
+; RUN: not llvm-spirv -s %t.bc
+; RUN: llvm-spirv --spirv-ext=+SPV_EXT_long_vector %t.bc
+; RUN: llvm-spirv %t.spv -to-text -o - | FileCheck %s
+
+; CHECK: ExtInstImport [[ExtInstSetId:[0-9]+]] "OpenCL.std"
+; CHECK: TypeInt [[Int:[0-9]+]] 32
+; CHECK: TypeVector [[Int5:[0-9]+]] [[Int]] 5
+
+; ModuleID = 'lower-non-standard-vec-with-ext'
+target datalayout = "e-i64:64-v16:16-v24:32-v32:32-v48:64-v96:128-v192:256-v256:256-v512:512-v1024:1024-n8:16:32:64"
+target triple = "spir64-unknown-unknown"
+
+@Id = external dso_local local_unnamed_addr addrspace(1) constant <3 x i64>, align 32
+
+declare <5 x i32> @llvm.abs.v5i32(<5 x i32> %x, i1 %is_int_min_poison)
+
+; Function Attrs: convergent norecurse
+define dso_local spir_func <5 x i32> @test_abs(<5 x i32> %src) local_unnamed_addr #0 !sycl_explicit_simd !4 !intel_reqd_sub_group_size !6 {
+entry:
+  %res = call <5 x i32> @llvm.abs.v5i32(<5 x i32> %src, i1 false)
+; CHECK: ExtInst [[Int5]] {{[0-9]+}} [[ExtInstSetId]] s_abs
+  ret <5 x i32> %res
+}
+
+attributes #0 = { convergent norecurse "frame-pointer"="all" "min-legal-vector-width"="256" "no-trapping-math"="true" "stack-protector-buffer-size"="8" "sycl-module-id"="lower-external-funcs-with-z.cpp" }
+
+!llvm.module.flags = !{!0, !1}
+!opencl.spir.version = !{!2}
+!spirv.Source = !{!3}
+!opencl.used.extensions = !{!4}
+!opencl.used.optional.core.features = !{!4}
+!opencl.compiler.options = !{!4}
+!llvm.ident = !{!5}
+
+!0 = !{i32 1, !"wchar_size", i32 4}
+!1 = !{i32 7, !"frame-pointer", i32 2}
+!2 = !{i32 1, i32 2}
+!3 = !{i32 0, i32 100000}
+!4 = !{}
+!5 = !{!"Compiler"}
+!6 = !{i32 1}
--- a/test/extensions/EXT/SPV_EXT_long_vector/lower-non-standard-vec-abs-with-ext.ll
+++ b/test/extensions/EXT/SPV_EXT_long_vector/lower-non-standard-vec-abs-with-ext.ll
@@ -1,7 +1,9 @@
 ; RUN: llvm-as %s -o %t.bc
-; RUN: not llvm-spirv -s %t.bc
+; RUN: not llvm-spirv -s %t.bc 2>&1 | FileCheck %s --check-prefix=CHECK-ERROR
 ; RUN: llvm-spirv --spirv-ext=+SPV_EXT_long_vector %t.bc
 ; RUN: llvm-spirv %t.spv -to-text -o - | FileCheck %s
+
+; CHECK-ERROR: Unsupported vector type with 5 elements
 
 ; CHECK: ExtInstImport [[ExtInstSetId:[0-9]+]] "OpenCL.std"
 ; CHECK: TypeInt [[Int:[0-9]+]] 32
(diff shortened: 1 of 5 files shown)


Qi-Ye-079:
@aratajew @bader @maarquitos14 @MrSidims @svenvh Do you mind having a quick review and merge this? Much thanks.

Also @YuriPlyakhin @vmaksimo to follow up later.

test/extensions/EXT/SPV_EXT_long_vector/lower-non-standard-vec-abs-with-ext.ll
```diff
@@ -0,0 +1,42 @@
+; RUN: llvm-as %s -o %t.bc
+; RUN: not llvm-spirv -s %t.bc
```
maarquitos14: Should we also check the error message? We could have other errors rather than the one we're looking for and this would pass anyway.

test/extensions/EXT/SPV_EXT_long_vector/lower-non-standard-vec-sqrt-with-ext.ll
```diff
@@ -0,0 +1,42 @@
+; RUN: llvm-as %s -o %t.bc
+; RUN: not llvm-spirv -s %t.bc
```
maarquitos14: Same as above.

test/extensions/EXT/SPV_EXT_long_vector/lower-non-standard-vec-with-ext.ll
```diff
@@ -0,0 +1,38 @@
+; RUN: llvm-as %s -o %t.bc
+; RUN: not llvm-spirv -s %t.bc
```
maarquitos14: Same as above.

test/extensions/EXT/SPV_EXT_long_vector/lower-non-standard-vec-abs-with-ext.ll
```diff
@@ -0,0 +1,42 @@
...
+!opencl.compiler.options = !{!4}
+!llvm.ident = !{!5}
+
+!0 = !{i32 1, !"wchar_size", i32 4}
+!1 = !{i32 7, !"frame-pointer", i32 2}
+!2 = !{i32 1, i32 2}
+!3 = !{i32 0, i32 100000}
+!4 = !{}
+!5 = !{!"Compiler"}
+!6 = !{i32 1}
```
maarquitos14: Are all these relevant for the test?

test/extensions/EXT/SPV_EXT_long_vector/lower-non-standard-vec-abs-with-ext.ll
```diff
@@ -0,0 +1,42 @@
...
+
+; CHECK: ExtInstImport [[ExtInstSetId:[0-9]+]] "OpenCL.std"
+; CHECK: TypeInt [[Int:[0-9]+]] 32
+; CHECK: TypeVector [[Int5:[0-9]+]] [[Int]] 5
+
+; ModuleID = 'lower-non-standard-vec-with-ext'
+target datalayout = "e-i64:64-v16:16-v24:32-v32:32-v48:64-v96:128-v192:256-v256:256-v512:512-v1024:1024-n8:16:32:64"
+target triple = "spir64-unknown-unknown"
+
+@Id = external dso_local local_unnamed_addr addrspace(1) constant <3 x i64>, align 32
```
maarquitos14: Is this relevant for the test?

maarquitos14: Applies to all the tests.

test/extensions/EXT/SPV_EXT_long_vector/lower-non-standard-vec-abs-with-ext.ll
maarquitos14: Applies to all the tests.

test/extensions/EXT/SPV_EXT_long_vector/lower-non-standard-vec-sqrt-with-ext.ll
```diff
@@ -0,0 +1,42 @@
...
+target triple = "spir64-unknown-unknown"
+
+@Id = external dso_local local_unnamed_addr addrspace(1) constant <3 x i64>, align 32
+
+declare <5 x float> @llvm.sqrt.f32(<5 x float> %x)
+
+; Function Attrs: convergent norecurse
+define dso_local spir_func <5 x float> @test_sqrt(<5 x float> %src) local_unnamed_addr #0 !sycl_explicit_simd !4 !intel_reqd_sub_group_size !6 {
+entry:
+  %res = call <5 x float> @llvm.sqrt.f32(<5 x float> %src)
```
maarquitos14, line 21: Can't this be added to the test above so we get rid of one file?

test/extensions/EXT/SPV_EXT_long_vector/lower-non-standard-vec-with-ext.ll
```diff
@@ -0,0 +1,38 @@
...
+; RUN: llvm-spirv --spirv-ext=+SPV_EXT_long_vector -s %t.bc
```
maarquitos14: No check at all? The test would pass if we generated an empty module as long as it doesn't crash.

test/extensions/EXT/SPV_EXT_long_vector/lower-non-standard-vec-with-ext.ll
```diff
@@ -0,0 +1,38 @@
...
+
+; Function Attrs: convergent norecurse
+define dso_local spir_func void @vmult2() local_unnamed_addr #0 !sycl_explicit_simd !4 !intel_reqd_sub_group_size !6 {
+entry:
+  %0 = bitcast <1 x i32> <i32 65793> to <4 x i8>
+  %1 = extractelement <4 x i8> %0, i32 0
+  %2 = bitcast <1 x i32> <i32 131586> to <4 x i8>
+  %3 = extractelement <4 x i8> %2, i32 0
+  %4 = bitcast <5 x i32> <i32 1, i32 1, i32 1, i32 1, i32 1> to <20 x i8>
+  ret void
```
maarquitos14, line 29: Is this checking anything other than what the other two tests do?

test/extensions/EXT/SPV_EXT_long_vector/lower-non-standard-vec-abs-with-ext.ll
Qi-Ye-079: Agree

test/extensions/EXT/SPV_EXT_long_vector/lower-non-standard-vec-abs-with-ext.ll
Qi-Ye-079: Not really. Will remove.

test/extensions/EXT/SPV_EXT_long_vector/lower-non-standard-vec-sqrt-with-ext.ll
Qi-Ye-079: They both have the `RUN: not llvm-spirv -s %t.bc` check as negative test that it should crash without the extension for this particular intrinsic. If we merge them into one file, we can't really verify that both can lead to the crash. Since this repo doesn't use `split-file`, I'll keep them as seperate.

test/extensions/EXT/SPV_EXT_long_vector/lower-non-standard-vec-with-ext.ll
Qi-Ye-079: Agree, we should check something.

test/extensions/EXT/SPV_EXT_long_vector/lower-non-standard-vec-with-ext.ll
Qi-Ye-079: This checks that whether the `SPIRVLowerBitCastToNonStandardType` pass will crash or not, depending on whether the extension is enabled. This pass tries to rewrite the particular pattern of `bitcast` + `extractelement` (as a result of LLVM's `InstCombine`, which may produce vector types with non-standard size) to vector types with standard sizes. If it can't, then it crashes with `report_fatal_error`. This is what the 2 check lines do: the 1st `not llvm-spirv` checks it should crash without the extension, the 2nd `llvm-spirv` checks it should NOT crash with the extension enabled (by bailing out early in this case because there's no need to run this pass if non-standard vector size is supported by the extension).

The other 2 tests the particular presence of `llvm.sqrt` and `llvm.abs` intrinsics with non-standard-size vector types specifically, covered in SPIRVUtil.cpp.

test/extensions/EXT/SPV_EXT_long_vector/lower-non-standard-vec-abs-with-ext.ll
Qi-Ye-079: Not really. Will remove.

test/extensions/EXT/SPV_EXT_long_vector/lower-non-standard-vec-sqrt-with-ext.ll
Qi-Ye-079: Agree

test/extensions/EXT/SPV_EXT_long_vector/lower-non-standard-vec-with-ext.ll
Qi-Ye-079: Agree

Qi-Ye-079 (commented):
Thanks for carefully review, I simply grabbed the same tests from `extensions/INTEL/SPV_INTEL_vector_compute` without thinking too much (just changed the extension)... they need improvement indeed.

YuriPlyakhin (approved):
LGTM

maarquitos14 (approved):

svenvh (approved):

PR #3828, opened 2026-07-02
Represent FP4 E2M1 with SPV_EXT_ocp_microscaling_types

vmaksimo:
Switch the FP4 (E2M1) type from `SPV_INTEL_float4` extension to the multi-vendor `SPV_EXT_ocp_microscaling_types` extension.

Plain FP4 conversion builtins get the EXT postfix (e.g. ConvertE2M1ToFP16EXT);
stochastic-rounding builtins keep the INTEL postfix as their operation belongs to `SPV_INTEL_fp_conversions`.

Specs: 
- https://github.khronos.org/SPIRV-Registry/extensions/EXT/SPV_EXT_ocp_microscaling_types.html
- https://github.com/intel/llvm/pull/22528 - related spec change for `SPV_INTEL_float4` - now it only allows FP4 type for cooperative matrix components.

Note:
In this patch we still accept both SPV_INTEL_float4 and SPV_EXT_ocp_microscaling_types encodings for E2M1; each encoding round-trips to its own builtin. The reason is not all producers/consumers are ready to switch to the new extension yet. At the same time, we want to introduce E2M1 through SPV_EXT_ocp_microscaling_types ASAP for those who are ready and willing to switch to it.

AI-assisted: Claude Opus 4.8 (commercial SaaS)

PR #3877, opened 2026-07-16
[Backport to 15] [NFC] Move mini-float tests to correct dir and add 2 close-to-real-life tests,Represent FP4 E2M1 with SPV_EXT_ocp_microscaling_types

Qi-Ye-079:
Backport of #3464, #3828


test/extensions/INTEL/SPV_INTEL_float4/cooperative_matrix.ll
```diff
@@ -0,0 +1,68 @@
+; Checks that Float4E2M1EXT encoding is preserved through a round-trip
+; translation when it's used as a cooperative-matrix component type.
+
+; Target extension types (target("spirv.CooperativeMatrixKHR", ...)) are not
+; supported by the LLVM 15 IR parser, so llvm-as cannot assemble this test.
+; XFAIL: *
```
YuriPlyakhin, line 6: @Qi-Ye-079 , I merged it, because I saw that resolution like that was already applied previously for backports of Cooperative Matrix tests to branches 14 and 15.
I think in a perfect world, it would be good to port those tests to the matrix representation used by these branches (example: https://github.com/KhronosGroup/SPIRV-LLVM-Translator/blob/5926441cba05aad179ae090b99680013a479c90e/test/extensions/KHR/SPV_KHR_cooperative_matrix/access_store.ll#L35)
Not sure how important it is though...

PR #3879, opened 2026-07-17
Fix test/transcoding/dbginfo-bug-on-bool-converts.ll test expectations

aobolensk:
https://github.com/llvm/llvm-project/pull/209232 fixed the SPIR-V backend selectIToF to preserve signedness (sitofp i1 true was miscompiled as 1.0 instead of -1.0), so the select now produces -1 rather than 1

PR #3898, opened 2026-07-21
[CI] Mark failing tests as expected-fail to keep CI green

idubinov:
Mark broken tests as expected fail to unblock CI
Created issue: https://github.com/KhronosGroup/SPIRV-LLVM-Translator/issues/3899

### DIFF
--- a/test/DebugInfo/LocalAddressSpace.ll
+++ b/test/DebugInfo/LocalAddressSpace.ll
@@ -1,5 +1,5 @@
 ; XFAIL: *
-; ISSUE #
+; ISSUE #3899
 ; Source:
 ;__kernel void foo(void) {
 ;  __local int a;
--- a/test/DebugInfo/LocalAddressSpace.ll
+++ b/test/DebugInfo/LocalAddressSpace.ll
@@ -1,3 +1,5 @@
+; XFAIL: *
+; ISSUE #
 ; Source:
 ;__kernel void foo(void) {
 ;  __local int a;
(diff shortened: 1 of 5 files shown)


idubinov:
@MrSidims @aobolensk @svenvh @vmaksimo @maarquitos14 @Qi-Ye-079 
As far as i understood, the [tests](https://github.com/KhronosGroup/SPIRV-LLVM-Translator/issues/3899) are blocked by https://github.com/KhronosGroup/SPIRV-LLVM-Translator/pull/3879 and https://github.com/KhronosGroup/SPIRV-LLVM-Translator/pull/3864, so, until that i think it will be good to xfail them to make the CI green which make contribution more comfortable

MrSidims:
We don't have that many of contributions and contributors and we don't have frequent releases. So my preference to not have "keep CI green" goal - instead if tests are failing - just understand, why they are failing and fix/adjust them when convenient (unless LLVM upstream change breaks build or exposes a real llvm-spirv bug).

Note, your patch won't keep CI green. Lets imagine we fixed the tests - then it would make CI back to red for in-tree CI. Then we have to revert your patch - it would make in-tree CI green again, but it would make out-of-tree CI temporary red.

idubinov:
@MrSidims , 

I think red CI may be the cause we skip new regression. for example, current red CI have two regressions: one regression of out-of-tree builds and another regression of in-tree builds which happened later, not sure is it caused by changes, but anyway - new regression happened while CI was red

> Lets imagine we fixed the tests

usually, developer removes XFAIL with test fix

I don't want to force the idea of green CI, if we have different opinions - this is topic for discussion

PR #3900, opened 2026-07-21
Error handling for malformed SPIRV input

maarquitos14:
Further hardening of the translator against malformed input to prevent crashes. Specifically, this PR:
- Reports an error and invalidates module in `SPIRVFunction::decode()` for opcodes that don't belong in function scope.
- Adds null check with module invalidation and early exit after `Decoder.getEntry()` in `SPIRVFunction::decodeBB()` because it can return `nullptr` and it was otherwise dereferenced down the line.
- Reports an error and invalidates module for unknown extended instruction set IDs.
- Reports an  error and invalidates module for truncated instructions.

5 new lit tests are added that were crashing before this patch. These new tests are a bit special, they need to feed deliberately malformed SPIR-V to the translator, including opcodes that don't exist, instructions in the wrong scope or truncated streams. Since spirv-as normally validates its input, the tests use its `!<integer>` syntax to inject raw 32-bit words directly into the binary output, bypassing validation. For example, `!75535` emits the raw word `(1<<16)|9999`, which encodes a single-word instruction with unknown opcode 9999. Care must be taken with placement: `!` words after variadic instructions (like OpTypeFunction) get absorbed as extra operands rather than emitted as standalone instructions.

### DIFF
--- a/lib/SPIRV/libSPIRV/SPIRVEntry.cpp
+++ b/lib/SPIRV/libSPIRV/SPIRVEntry.cpp
@@ -95,8 +95,7 @@ SPIRVEntry *SPIRVEntry::create(Op OpCode) {
     return Loc->second();
 
   SPIRVDBG(spvdbgs() << "No factory for OpCode " << (unsigned)OpCode << '\n';)
-  assert(0 && "Not implemented");
-  return 0;
+  return nullptr;
 }
 
 std::unique_ptr<SPIRV::SPIRVEntry> SPIRVEntry::createUnique(Op OC) {
--- a/lib/SPIRV/libSPIRV/SPIRVFunction.cpp
+++ b/lib/SPIRV/libSPIRV/SPIRVFunction.cpp
@@ -116,7 +116,12 @@ void SPIRVFunction::decode(std::istream &I) {
       break;
     }
     default:
-      assert(0 && "Invalid SPIRV format");
+      Module->getErrorLog().checkError(false, SPIRVEC_InvalidModule,
+                                       "unexpected opcode " +
+                                           std::to_string(Decoder.OpCode) +
+                                           " in function definition");
+      Module->setInvalid();
+      return;
     }
   }
 }
@@ -142,6 +147,10 @@ bool SPIRVFunction::decodeBB(SPIRVDecoder &Decoder) {
     }
 
     SPIRVEntry *Entry = Decoder.getEntry();
+    if (!Entry) {
+      Module->setInvalid();
+      return false;
+    }
 
     if (Decoder.OpCode == OpLine) {
       std::shared_ptr<const SPIRVLine> L(static_cast<SPIRVLine *>(Entry));
--- a/lib/SPIRV/libSPIRV/SPIRVFunction.cpp
+++ b/lib/SPIRV/libSPIRV/SPIRVFunction.cpp
@@ -148,6 +148,9 @@ bool SPIRVFunction::decodeBB(SPIRVDecoder &Decoder) {
 
     SPIRVEntry *Entry = Decoder.getEntry();
     if (!Entry) {
+      Module->getErrorLog().checkError(false, SPIRVEC_InvalidInstruction,
+                                       "failed to decode instruction in "
+                                       "basic block");
       Module->setInvalid();
       return false;
     }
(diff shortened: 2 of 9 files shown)


maarquitos14:
@MrSidims @svenvh friendly ping. As far as I can tell, CI failures are unrelated.

MrSidims (commented):
@idubinov @aobolensk please take a look

lib/SPIRV/libSPIRV/SPIRVEntry.cpp
```diff
@@ -95,8 +95,7 @@ SPIRVEntry *SPIRVEntry::create(Op OpCode) {
     return Loc->second();
 
   SPIRVDBG(spvdbgs() << "No factory for OpCode " << (unsigned)OpCode << '\n';)
-  assert(0 && "Not implemented");
```
idubinov, line 98: Should not we avoid removing assert here?

lib/SPIRV/libSPIRV/SPIRVModule.cpp
```diff
@@ -1007,7 +1007,14 @@ SPIRVEntry *SPIRVModuleImpl::getEntry(SPIRVId Id) const {
...
 SPIRVExtInstSetKind SPIRVModuleImpl::getBuiltinSet(SPIRVId SetId) const {
   auto Loc = IdToInstSetMap.find(SetId);
-  assert(Loc != IdToInstSetMap.end() && "Invalid builtin set id");
+  if (Loc == IdToInstSetMap.end()) {
+    const_cast<SPIRVModuleImpl *>(this)->getErrorLog().checkError(
+        false, SPIRVEC_InvalidModule,
+        "input SPIR-V module references an unknown extended instruction set "
+        "id " +
+            std::to_string(SetId));
+    return SPIRVEIS_Count;
```
idubinov, line 1016: Is this return being caught somewhere?
Also, it may be useful only for not general implementation - due to hardcoded behavior of checkError

lib/SPIRV/libSPIRV/SPIRVFunction.cpp
```diff
@@ -142,6 +147,10 @@ bool SPIRVFunction::decodeBB(SPIRVDecoder &Decoder) {
     }
 
     SPIRVEntry *Entry = Decoder.getEntry();
+    if (!Entry) {
+      Module->setInvalid();
```
aobolensk, line 154: Can we report an error of some sort at this point (for debugability)?

lib/SPIRV/libSPIRV/SPIRVEntry.cpp
maarquitos14: I can restore it if you think it's helpful, but the two places where this was crashing do have proper error reporting now, so I lean towards thinking it's not very helpful anymore.

idubinov: cool, thank you

idubinov (approved):

lib/SPIRV/libSPIRV/SPIRVModule.cpp
maarquitos14: This function is called from:

1. `lib/SPIRV/SPIRVReader.cpp:5564`: it's within an assert that checks if the returned value is `SPIRVEIS_OpenCL`.
2. `lib/SPIRV/libSPIRV/SPIRVInstruction.h:2034`: assigns to a value that is immediately checked within an assert to see if it's any of the valid values.

In both cases, this return value would make them fail.

lib/SPIRV/libSPIRV/SPIRVFunction.cpp
maarquitos14: We can! Done.

aobolensk (approved):

MrSidims (approved):

MrSidims (approved):

maarquitos14:
@svenvh @vmaksimo any concerns with this? If not, I think this is ready to merge, the CI failures seem unrelated.

svenvh:
LGTM; closing/reopening for a fresh CI run, as the CI issues should have been addressed now.

PR #3901, opened 2026-07-21
Read old SPV_INTEL_float4 type for StochasticRound*ToE2M1INTEL

fda0:
For backward compatibility of the SPIRV Reader only: older modules
emitted the `SPV_INTEL_float4` `Float4E2M1INTEL` type (6214) as the
`StochasticRoundFToFINTEL` result. Map it to the same builtin as the
current `SPV_EXT_ocp_microscaling_types` `Float4E2M1EXT` type (4225).

The Writer keeps unchanged and emits the EXT type.

### DIFF
--- a/lib/SPIRV/SPIRVReader.cpp
+++ b/lib/SPIRV/SPIRVReader.cpp
@@ -1136,7 +1136,11 @@ Value *SPIRVToLLVM::transConvertInst(SPIRVValue *BV, Function *F,
           LookupOC = OpFConvert;
         else if (OC == internal::OpClampStochasticRoundFToFINTEL)
           LookupOC = internal::OpStochasticRoundFToFINTEL;
-        FPConversionDesc FPDesc = {SrcEnc, DstEnc, LookupOC,
+        FPEncodingWrap LookupDstEnc = DstEnc;
+        if (LookupOC == internal::OpStochasticRoundFToFINTEL &&
+            DstEnc == FPEncodingWrap::E2M1INTEL)
+          LookupDstEnc = FPEncodingWrap::E2M1;
+        FPConversionDesc FPDesc = {SrcEnc, LookupDstEnc, LookupOC,
                                    /*Saturate=*/IsSaturatedFP8};
         auto Conv = SPIRV::FPConvertToEncodingMap::rmap(FPDesc);
         std::vector<Value *> Ops = {Src};
--- a/test/extensions/INTEL/SPV_INTEL_fp_conversions/spv_intel_fp_conversions_e2m1intel_reader.spt
+++ b/test/extensions/INTEL/SPV_INTEL_fp_conversions/spv_intel_fp_conversions_e2m1intel_reader.spt
@@ -1,0 +1,51 @@
+; RUN: llvm-spirv %s -to-binary -o %t.spv
+; RUN: llvm-spirv -r %t.spv -o %t.rev.bc
+; RUN: llvm-dis %t.rev.bc -o %t.rev.ll
+; RUN: FileCheck < %t.rev.ll %s
+
+; CHECK: call spir_func i4 @_Z46__builtin_spirv_StochasticRoundFP16ToE2M1INTELDhi(half 1.000000e+00, i32 1)
+; CHECK: call spir_func i4 @_Z46__builtin_spirv_StochasticRoundBF16ToE2M1INTELDF16bi(bfloat 1.000000e+00, i32 1)
+
+119734787 65536 393230 19 0
+2 Capability Addresses
+2 Capability Linkage
+2 Capability Kernel
+2 Capability Float16Buffer
+2 Capability Int4TypeINTEL
+2 Capability BFloat16TypeKHR
+2 Capability Float4E2M1INTEL
+2 Capability FloatConversionsFtoFINTEL
+6 Extension "SPV_INTEL_float4"
+8 Extension "SPV_INTEL_fp_conversions"
+5 Extension "SPV_INTEL_int4"
+6 Extension "SPV_KHR_bfloat16"
+5 ExtInstImport 1 "OpenCL.std"
+3 MemoryModel 2 2
+3 Source 0 0
+8 Name 4 "hf16_fp4e2m1_stochastic"
+4 Name 5 "entry"
+8 Name 13 "bf16_fp4e2m1_stochastic"
+4 Name 14 "entry"
+10 Decorate 4 LinkageAttributes "hf16_fp4e2m1_stochastic" Export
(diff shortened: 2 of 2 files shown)


test/extensions/INTEL/SPV_INTEL_fp_conversions/spv_intel_fp_conversions.ll
```diff
@@ -4,24 +4,26 @@
...
 ; for scalar
 
-; RUN: llvm-spirv %s -o %t.spv --spirv-ext=+SPV_EXT_float8,+SPV_EXT_ocp_microscaling_types,+SPV_INTEL_int4,+SPV_KHR_bfloat16,+SPV_INTEL_fp_conversions
+; RUN: llvm-spirv %s -o %t.spv --spirv-ext=+SPV_EXT_float8,+SPV_INTEL_int4,+SPV_KHR_bfloat16,+SPV_INTEL_fp_conversions,+SPV_INTEL_float4
 ; TODO: re-enable spirv-val once it recognizes the FloatConversionsFtoFINTEL capability (6215)
 ; RUNx: spirv-val %t.spv
 ; RUN: llvm-spirv %t.spv -o %t.spt --to-text
-; RUN: FileCheck < %t.spt %s --check-prefix=CHECK-SPIRV
+; The E2M1INTEL builtins must produce the SPV_INTEL_float4 type (encoding 6214),
+; not the SPV_EXT_ocp_microscaling_types Float4E2M1EXT type (encoding 4225).
```
Qi-Ye-079: Nit: personally I think this comment is redundant, because the produced `TypeFloat [[#E2M1Ty:]] 4 6214` says it all.

Qi-Ye-079:
LGTM. I can't say I fully know the context of this PR but I think it makes sense technically.

@bashbaug Maybe you know the context better for this one? Please also help review, approve and merge this one if it looks ok to you, as I don't have write permission to approve and merge.

bashbaug:
Note, the "INTEL" suffix should be considered part of the instruction name, not part of the type name, so I disagree that this change should be done for some sort of naming reasons.

I'd also like to move away from SPV_INTEL_float4 if at all possible in the SPIR-V writer, since we have no plans to productize this extension.  Ideally, the SPIR-V writer should be generating SPV_EXT_ocp_microscaling_types in all cases, instead.  We can continue to support SPV_INTEL_float4 in the SPIR-V reader for some time if needed for backwards compatibility, but eventually we should move away from it, also.

I suggest:

1. Keep the path that generates the SPV_EXT_ocp_microscaling_types Float4E2M1EXT float encoding as the preferred path when SPV_EXT_ocp_microscaling_types is enabled.
2. If needed, also support a path that generates the SPV_INTEL_float4 Float4E2M1INTEL float encoding as a backwards compatibility path when SPV_INTEL_float4 is enabled and SPV_EXT_ocp_microscaling_types is NOT enabled, but this won't be a preferred path, and it should eventually be removed.

fda0:
@bashbaug @Qi-Ye-079 thanks for the feedback. Pushed an update that is based on previous commit 6734efd9e659963788d84369065c064463c4d861

This should be scoped to SPIRV-Reader. The change now simply allows us to use `OpStochasticRoundFToFINTEL` with `E2M1INTEL` as was the case with older SPIRV-Reader versions.

test/extensions/INTEL/SPV_INTEL_fp_conversions/spv_intel_fp_conversions_e2m1intel_reader.spt
```diff
@@ -0,0 +1,51 @@
+; RUN: llvm-spirv %s -to-binary -o %t.spv
+; RUN: llvm-spirv -r %t.spv -o %t.rev.bc
+; RUN: llvm-dis %t.rev.bc -o %t.rev.ll
+; RUN: FileCheck < %t.rev.ll %s --check-prefix=CHECK-LLVM
+
+; CHECK-LLVM: call spir_func i4 @_Z46__builtin_spirv_StochasticRoundFP16ToE2M1INTELDhi(half 1.000000e+00, i32 1)
+; CHECK-LLVM: call spir_func i4 @_Z46__builtin_spirv_StochasticRoundBF16ToE2M1INTELDF16bi(bfloat 1.000000e+00, i32 1)
```
Qi-Ye-079: Nit: you don't need `--check-prefix=CHECK-LLVM` here; a plain `CHECK` will suffice.

lib/SPIRV/SPIRVReader.cpp
```diff
@@ -1136,7 +1136,11 @@ Value *SPIRVToLLVM::transConvertInst(SPIRVValue *BV, Function *F,
           LookupOC = OpFConvert;
         else if (OC == internal::OpClampStochasticRoundFToFINTEL)
           LookupOC = internal::OpStochasticRoundFToFINTEL;
-        FPConversionDesc FPDesc = {SrcEnc, DstEnc, LookupOC,
+        FPEncodingWrap LookupDstEnc = DstEnc;
+        if (LookupOC == internal::OpStochasticRoundFToFINTEL &&
+            DstEnc == FPEncodingWrap::E2M1INTEL)
+          LookupDstEnc = FPEncodingWrap::E2M1;
```
bashbaug, line 1142: This is done conditionally right now, only for OpStochasticRoundFToFINTEL, but might we want it to be done unconditionally to fix all usages of the Intel-specific encoding?

We might want to do the same for the source encoding, also.

Even if we decide to do this, I'm fine doing it in a separate PR.

bashbaug (approved):

lib/SPIRV/SPIRVReader.cpp
fda0: I don't know to be honest why only this case is an issue - but with this conditional fix for OpStochasticRoundFToFINTEL we have full backwards compatibility for SPV_INTEL_float4 right now (at least that's what testing shows). So broadening this is not strictly needed but I can prepare follow up a PR if you think it's a good idea.

Qi-Ye-079:
@svenvh The failing CIs are unrelated to this PR, and this PR is needed ASAP for intel. Is it ok if you can help approve and merge this PR without waiting for all CIs to be fixed? This small PR is intel-specific changes and it's already approved by one of our people. Thanks in advance.

YuriPlyakhin (approved):
LGTM

YuriPlyakhin:
The failing tests in CI are all unrelated to this PR, PR is very limited and specific to one vendor (and basically is just restoring backward compatibility) and approved by vendor. I'm merging. I hope that's ok with everyone.

PR #3913, opened 2026-07-25
Prefer RequiredPassInfoMixin

boomanaiden154:
In LLVM 23, {Optional,Required}PassInfoMixin became preferred for creating passes instead of raw PassInfoMixin with isRequired. In LLVM 24, PassInfoMixin will be moved to a detail namespace, causing compilation failures.


svenvh (approved):
LGTM; I believe we can ignore the CI failures.

MrSidims (approved):

PR #3930, opened 2026-07-28
Fix UniformId decoration round trip and SPIR-V version negotiation

aobolensk:
Emit UniformId as OpDecorateId with a Scope id operand instead of a plain OpDecorate literal, preserve id-based decorations through the -r round trip (SPIRVEntry only walked Decorates, not DecorateIds), and require SPIR-V 1.6 when UniformDecoration capability is selected

Inspired by related SPIR-V backend change https://github.com/llvm/llvm-project/pull/207958


aobolensk:

test/DecorateUniformIdRoundTrip.spvasm
```diff
@@ -0,0 +1,34 @@
+; REQUIRES: spirv-as
```
maarquitos14: We can use  `RUN: llvm-spirv -to-binary -o %t.spv %s` so that we don't need spirv-as.

test/DecorateUniformIdRoundTrip.spvasm
```diff
@@ -0,0 +1,34 @@
...
+; RUN: spirv-val %t.spv
+; RUN: llvm-spirv -r --spirv-target-env=SPV-IR %t.spv -o %t.rev.bc
+; RUN: llvm-dis %t.rev.bc -o - | FileCheck %s
+; RUN: llvm-spirv %t.rev.bc -o %t.fwd.spv
+; RUN: spirv-val %t.fwd.spv
+
+; Check that a UniformId decoration on an OpVariable survives a -r round
+; trip: it must show up as !spirv.Decorations metadata with the resolved
+; scope value (not the SPIR-V id of the scope constant), and translating
+; that metadata back to SPIR-V must produce a valid module again.
```
maarquitos14: Apart from producing a valid module again, should we check that the `Scope` operand is restored to the correct value?

aobolensk: That makes sense. Added, thanks

test/DecorateUniformIdRoundTrip.spvasm
aobolensk: Switched to .spt for that

maarquitos14 (approved):

Issue #3115, opened 2025-04-09
Adjust kernel wrapper after https://github.com/llvm/llvm-project/pull/115821


MrSidims:
After this change clang generates for this [OpenCL code](https://godbolt.org/z/Pz1f7sPE1):
{code}
kernel void Test() {}
{code}

following LLVM IR:
{code}
; Function Attrs: convergent noinline norecurse nounwind optnone
define dso_local spir_kernel void @Test() #0 !kernel_arg_addr_space !5 !kernel_arg_access_qual !5 !kernel_arg_type !5 !kernel_arg_base_type !5 !kernel_arg_type_qual !5 {
  call spir_func void @__clang_ocl_kern_imp_Test() #1
  ret void
}

; Function Attrs: convergent noinline norecurse nounwind optnone
define dso_local spir_func void @__clang_ocl_kern_imp_Test() #0 !kernel_arg_addr_space !5 !kernel_arg_access_qual !5 !kernel_arg_type !5 !kernel_arg_base_type !5 !kernel_arg_type_qual !5 {
  ret void
}
{code}
which means, that the translator will translate it to this:
{code}
; SPIR-V
; Version: 1.0
; Generator: Khronos LLVM/SPIR-V Translator; 14
; Bound: 12
; Schema: 0
               OpCapability Addresses
               OpCapability Linkage
               OpCapability Kernel
          %1 = OpExtInstImport "OpenCL.std"
               OpMemoryModel Physical32 OpenCL
               OpEntryPoint Kernel %9 "Test"
               OpSource OpenCL_C 300000
               OpName %Test "Test"
               OpName %__clang_ocl_kern_imp_Test "__clang_ocl_kern_imp_Test"
               OpDecorate %Test LinkageAttributes "Test" Export
               OpDecorate %__clang_ocl_kern_imp_Test LinkageAttributes "__clang_ocl_kern_imp_Test" Export
       %void = OpTypeVoid
          %3 = OpTypeFunction %void
       %Test = OpFunction %void DontInline %3
          %5 = OpLabel
          %7 = OpFunctionCall %void %__clang_ocl_kern_imp_Test
               OpReturn
               OpFunctionEnd
%__clang_ocl_kern_imp_Test = OpFunction %void DontInline %3
          %8 = OpLabel
               OpReturn
               OpFunctionEnd
          %9 = OpFunction %void DontInline %3
         %10 = OpLabel
         %11 = OpFunctionCall %void %Test
[...]

PR #3933, opened 2026-07-28
[NFC] Remove invalid entry_point_func.ll test

aobolensk:
Per SPIR-V spec it's invalid to call EntryPoint from another EntryPoint


MrSidims (approved):

MrSidims (commented):
Actually, hold on, this requires a bit of git log digging

aobolensk:
@svenvh @maarquitos14 please take a look

maarquitos14 (approved):

MrSidims (changes requested):
Per SPIR-V spec it's invalid to call EntryPoint from another EntryPoint.

Per OpenCL spec it's valid. Previously the translator has to deal with it, but then this job was moved to clang, see https://github.com/KhronosGroup/SPIRV-LLVM-Translator/issues/3115

So changing this:
```
define spir_kernel void @callerfunction() {
   call spir_kernel void @testfunction()
   call spir_kernel void @testdeclaration()
   ret void
}
```
to
```
define spir_kernel void @callerfunction() {
   call spir_func void @testhelper()
   call spir_func void @testdeclaration()
   ret void
}
```
doesn't make sense. Probably we should just remove the test.

aobolensk:
> Per SPIR-V spec it's invalid to call EntryPoint from another EntryPoint.
> 
> Per OpenCL spec it's valid. Previously the translator has to deal with it, but then this job was moved to clang, see #3115
> 
> So changing this:
> 
> ```
> define spir_kernel void @callerfunction() {
>    call spir_kernel void @testfunction()
>    call spir_kernel void @testdeclaration()
>    ret void
> }
> ```
> 
> to
> 
> ```
> define spir_kernel void @callerfunction() {
>    call spir_func void @testhelper()
>    call spir_func void @testdeclaration()
>    ret void
> }
> ```
> 
> doesn't make sense. Probably we should just remove the test.

Sounds reasonable to me. Removed

MrSidims (approved):

vmaksimo (approved):

aobolensk:

svenvh (approved):

PR #3941, opened 2026-07-30
Make the error handling kind selectable through TranslatorOpts

### LINKED ISSUE #3939: readSpirv() calls std::exit() on a rejected module, making its documented error contract unreachable for embedders
The public entry point `readSpirv()` is documented as reporting failure through its return value and an error string:

    /// \brief Load SPIR-V from istream and translate to LLVM module.
    /// \returns true if succeeds.
    bool readSpirv(LLVMContext &C, std::istream &IS, Module *&M, std::string &ErrMsg);

In practice it can terminate the calling process instead. `SPIRV::SPIRVDbgError` defaults to `SPIRVDbgErrorHandlingKinds::Exit` (`lib/SPIRV/libSPIRV/SPIRVDebug.cpp:52`), so `SPIRVErrorLog::checkError()` reaches `std::exit(ErrCode)` (`lib/SPIRV/libSPIRV/SPIRVError.h:163`) while parsing. The documented `false` return and `ErrMsg` are then unreachable, and so is any error handling the caller wrote.

Three things make this hard for an embedder to live with:

1. The library is written to recover. The parser sets `M.setInvalid()` around the same failures, and `readSpirv()` checks `M->getError(ErrMsg)` and returns `false`. `std::exit()` short-circuits that path rather than complementing it.
2. The knob is not reachable from an installed header. `SPIRVDebug.h` is not among the installed `LLVMSPIRVLib` headers; only `LLVMSPIRVLib.h`, `LLVMSPIRVOpts.h` and `LLVMSPIRVExtensions.inc` are. An embedder that wants non fatal behaviour has to redeclare `enum class SPIRVDbgErrorHandlingKinds` and `extern SPIRVDbgError` itself, which silently sets the wrong value if the enum is ever reordered.
3. `llvm-spirv` the tool is the only consumer that actually wants process termination, and it could opt into it explicitly.

Concrete effect downstream: PoCL calls `readSpirv()` from `clBuildProgram`. Handed a SPIR-V module using an extension it was not built to accept, the host process is killed rather than `CL_BUILD_PROGRAM_FAILURE` being returned, which is an OpenCL specification violation that PoCL cannot fix without redeclaring these internals. Reproduced with PoCL 2.15.0 and 2.16.0 (`exit(11)` and `exit(10)` respectively).

Minimal reproducer, about 25 lines of `clCreateProgramWithIL` plus `clBuildProgram`, plus a SPIR-V module using `SPV_INTEL_function_pointers`. Available on request, or reproducible with any module using an extension the build does not enable.

Suggestion: expose the selection through `SPIRV::TranslatorOpts` in `LLVMSPIRVOpts.h`, which is already public and already carries per translation options, and default library use to the non fatal behaviour while `llvm-spirv` opts into `Exit`. Changing the default alone would also work, since the recovery path already exists.

pvelesko:
Fixes #3939

Declares the error handling kind in the public `LLVMSPIRVOpts.h` and carries it on `TranslatorOpts`, so an embedder can select the non fatal behaviour without redeclaring library internals. The default stays `Exit`, so nothing changes for callers that do not ask.

Two further fixes were needed to make `Ignore` usable: `parseSPT()` held the error log by value, so `getError()` lost the reason, and `Ignore` wrote to stderr unconditionally. `llvm-spirv` gains `--spirv-error-handling`, which the new test uses.

Validated on the `llvm_release_220` branch against LLVM 22.1.0, where the test fails before the change and passes after, with the rest of the suite unaffected (1017 passed, 0 unexpected failures). The patch applies to `main` with identical added and removed lines, but `main` needs LLVM 24, which I cannot build locally, so I have not run it there.

@svenvh as you suggested in the issue.

### DIFF
--- a/include/LLVMSPIRVOpts.h
+++ b/include/LLVMSPIRVOpts.h
@@ -135,6 +135,14 @@ enum SPIRAddressSpace : uint32_t {
 
 using AddrSpaceMap = std::array<uint32_t, SPIRAS_Count>;
 
+/// \brief What the translator does with a module it cannot accept.
+///
+/// Ignore leaves the failure to be reported through the return value and the
+/// error string of the entry point, which is what an embedded use wants.
+/// Abort and Exit terminate the process and are only appropriate for a
+/// standalone tool.
+enum class SPIRVDbgErrorHandlingKinds { Abort, Exit, Ignore };
+
 /// \brief Helper class to manage SPIR-V translation
 class TranslatorOpts {
 public:
@@ -338,6 +346,13 @@ public:
   // success, false on failure.
   bool validateFnVarOpts() const;
 
+  SPIRVDbgErrorHandlingKinds getErrorHandlingKind() const noexcept {
+    return ErrorHandling;
+  }
+  void setErrorHandlingKind(SPIRVDbgErrorHandlingKinds Kind) noexcept {
+    ErrorHandling = Kind;
+  }
+
 private:
   // Common translation options
   VersionNumber MaxVersion = VersionNumber::MaximumVersion;
@@ -366,6 +381,11 @@ private:
   // Unknown LLVM intrinsics will be translated as external function calls in
   // SPIR-V
   std::optional<ArgList> SPIRVAllowUnknownIntrinsics{};
+
+  // What to do with a module the translator cannot accept. Kept at Exit, the
+  // historical behaviour of the SPIRVDbgError global, so that this is not a
+  // change for anyone who does not ask for it.
+  SPIRVDbgErrorHandlingKinds ErrorHandling = SPIRVDbgErrorHandlingKinds::Exit;
 
   // Enable support for extra DIExpression opcodes not listed in the SPIR-V
   // DebugInfo specification.
--- a/include/LLVMSPIRVOpts.h
+++ b/include/LLVMSPIRVOpts.h
@@ -138,9 +138,8 @@ using AddrSpaceMap = std::array<uint32_t, SPIRAS_Count>;
 /// \brief What the translator does with a module it cannot accept.
 ///
 /// Ignore leaves the failure to be reported through the return value and the
-/// error string of the entry point, which is what an embedded use wants.
-/// Abort and Exit terminate the process and are only appropriate for a
-/// standalone tool.
+/// error string of the entry point, typical for embedded uses. Abort and Exit
+/// terminate the process, typical for standalone tools.
 enum class SPIRVDbgErrorHandlingKinds { Abort, Exit, Ignore };
 
 /// \brief Helper class to manage SPIR-V translation
@@ -382,9 +381,7 @@ private:
   // SPIR-V
   std::optional<ArgList> SPIRVAllowUnknownIntrinsics{};
 
-  // What to do with a module the translator cannot accept. Kept at Exit, the
-  // historical behaviour of the SPIRVDbgError global, so that this is not a
-  // change for anyone who does not ask for it.
+  // What to do with a module the translator cannot accept.
   SPIRVDbgErrorHandlingKinds ErrorHandling = SPIRVDbgErrorHandlingKinds::Exit;
 
   // Enable support for extra DIExpression opcodes not listed in the SPIR-V
--- a/lib/SPIRV/libSPIRV/SPIRVDebug.h
+++ b/lib/SPIRV/libSPIRV/SPIRVDebug.h
@@ -40,6 +40,7 @@
 #ifndef SPIRV_LIBSPIRV_SPIRVDEBUG_H
 #define SPIRV_LIBSPIRV_SPIRVDEBUG_H
 
+#include "LLVMSPIRVOpts.h"
 #include "SPIRVUtil.h"
 
 #include <iostream>
@@ -54,8 +55,10 @@ namespace SPIRV {
 // Include source file and line number in error message.
 extern bool SPIRVDbgErrorMsgIncludesSourceInfo;
 
-// Enable assert or exit on error
-enum class SPIRVDbgErrorHandlingKinds { Abort, Exit, Ignore };
+// Enable assert or exit on error. The kind is declared in the public
+// LLVMSPIRVOpts.h so that embedders can select it through TranslatorOpts
+// without redeclaring it. This global remains the default for translation
+// that does not go through TranslatorOpts.
 extern SPIRVDbgErrorHandlingKinds SPIRVDbgError;
 
 // Enable debug output.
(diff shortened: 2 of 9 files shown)


include/LLVMSPIRVOpts.h
```diff
@@ -367,6 +382,11 @@ class TranslatorOpts {
   // SPIR-V
   std::optional<ArgList> SPIRVAllowUnknownIntrinsics{};
 
+  // What to do with a module the translator cannot accept. Kept at Exit, the
+  // historical behaviour of the SPIRVDbgError global, so that this is not a
+  // change for anyone who does not ask for it.
```
svenvh: Please only describe the current state; for historical background we have the commit message / git log.

lib/SPIRV/libSPIRV/SPIRVDebug.h
```diff
@@ -54,8 +55,10 @@ namespace SPIRV {
 // Include source file and line number in error message.
 extern bool SPIRVDbgErrorMsgIncludesSourceInfo;
 
-// Enable assert or exit on error
-enum class SPIRVDbgErrorHandlingKinds { Abort, Exit, Ignore };
+// Enable assert or exit on error. The kind is declared in the public
+// LLVMSPIRVOpts.h so that embedders can select it through TranslatorOpts
+// without redeclaring it. This global remains the default for translation
+// that does not go through TranslatorOpts.
```
svenvh: I think we can prune the comment down to:
```suggestion
// Enable assert or exit on error. This global holds the default for translation
// that does not go through TranslatorOpts.
```

pvelesko:
The two red checks are not related to this change. Linux (Debug) and Linux (Release) fail in the Install dependencies step, before anything is compiled:

```
E: Unable to locate package clang-24
E: Unable to locate package llvm-24-dev
```

The same two jobs fail identically on current main (run 30502695866, commit 706ad87a, the base of this branch), and I opened #3942, a comment only change on top of main, which reproduced the same two failures in the same step. Everything that does build this change is green: In-tree Linux, Windows and macOS, plus code style. The test suite goes from 890 to 891 passing against main, with unsupported and expected failures unchanged.

tools/llvm-spirv/llvm-spirv.cpp
```diff
@@ -145,6 +145,18 @@ static cl::opt<SPIRV::ExtInst> ExtInst(
...
+static cl::opt<SPIRV::SPIRVDbgErrorHandlingKinds> ErrorHandling(
+    "spirv-error-handling",
+    cl::desc("Specify what happens when a module is rejected"),
+    cl::values(clEnumValN(SPIRV::SPIRVDbgErrorHandlingKinds::Abort, "abort",
+                          "Print the error and abort"),
+               clEnumValN(SPIRV::SPIRVDbgErrorHandlingKinds::Exit, "exit",
+                          "Print the error and exit with its error code"),
+               clEnumValN(SPIRV::SPIRVDbgErrorHandlingKinds::Ignore, "ignore",
+                          "Report the error through the return value and let "
+                          "the caller print it")),
```
svenvh: Please rephrase this message.  This is a user facing help message for a standalone tool, so "return value" or "caller" have no meaning here.

include/LLVMSPIRVOpts.h
```diff
@@ -135,6 +135,14 @@ enum SPIRAddressSpace : uint32_t {
 
 using AddrSpaceMap = std::array<uint32_t, SPIRAS_Count>;
 
+/// \brief What the translator does with a module it cannot accept.
+///
+/// Ignore leaves the failure to be reported through the return value and the
+/// error string of the entry point, which is what an embedded use wants.
+/// Abort and Exit terminate the process and are only appropriate for a
+/// standalone tool.
```
svenvh: I'd phrase this a bit less compulsory: "..., typical for embedded uses", "..., typical for standalone tools".

svenvh (commented):
Thanks, mostly LGTM except for a few comments.

The out-of-tree CI failures are known, don't worry about those.

pvelesko:
@svenvh done

svenvh (approved):

lib/SPIRV/libSPIRV/SPIRVModule.cpp
```diff
@@ -2563,7 +2564,9 @@ std::istream &SPIRVModuleImpl::parseSPT(std::istream &I) {
     SPIRVDBG(spvdbgs() << "Read word: W = " << W << " V = 0\n");
     return W;
   };
-  SPIRVErrorLog ErrorLog = MI.getErrorLog();
+  // A reference, not a copy: setError() has to reach the module's own log so
+  // that getError() can report the reason to the caller.
```
MrSidims: this comment is redundant

MrSidims (approved):

lib/SPIRV/libSPIRV/SPIRVModule.cpp
pvelesko: @MrSidims done.

lib/SPIRV/libSPIRV/SPIRVDebug.h
```diff
@@ -54,8 +55,8 @@ namespace SPIRV {
 // Include source file and line number in error message.
 extern bool SPIRVDbgErrorMsgIncludesSourceInfo;
 
-// Enable assert or exit on error
-enum class SPIRVDbgErrorHandlingKinds { Abort, Exit, Ignore };
+// Enable assert or exit on error. This global holds the default for translation
+// that does not go through TranslatorOpts.
```
vmaksimo: What kinds of calls actually end up reading this global? 
I see that every entry point for translation constructs a `TranslatorOpts` with the default values and overrides this value.

pvelesko: @vmaksimo correct, no translation path reads the global. The latest commits finish the job: `getSpecConstInfo` (the one remaining entry point that could only be selected through the global) gained a TranslatorOpts overload, `SPIRVLowerLLVMIntrinsic`'s stack local log now takes the kind from the pass options, and the `SPIRVDbgError` comment now names its actual consumers.

vmaksimo (approved):

pvelesko:
@svenvh @MrSidims @vmaksimo please merge if all good. Thanks

pvelesko:
@svenvh LGTM?

vmaksimo:
Reopening to trigger CI run

pvelesko:
@vmaksimo same CI failures as before

vmaksimo:
Yeah, we're stuck with OOT build failure for some time unfortunately (not a blocker, merging your change)

PR #3943, opened 2026-08-03
Support typed pointers for OpTypeBufferSurfaceINTEL

michalpaszkowski:
Translate buffer surfaces to pointers to opaque structs when typed pointers are enabled, instead of using target extension types.


test/extensions/INTEL/SPV_INTEL_vector_compute/buffer_surface_intel_typed_pointers.ll
vmaksimo, line 1: I'd change the test to *.spt, because the fix is only for the reader

lib/SPIRV/SPIRVReader.cpp
```diff
@@ -467,6 +467,9 @@ Type *SPIRVToLLVM::transType(SPIRVType *T, bool UseTPT) {
     if (UseTPT) {
       Type *StructTy = getOrCreateOpaqueStructType(M, transVCTypeName(PST));
       Ty = TypedPointerType::get(StructTy, SPIRAS_Global);
+    } else if (Context->supportsTypedPointers()) {
```
vmaksimo: Let's better reuse [already calculated option](https://github.com/michalpaszkowski/SPIRV-LLVM-Translator/blob/ba313baaa09c83c6f8514ef8a250020ea99e6029/lib/SPIRV/SPIRVReader.cpp#L3749-L3750); this way we won't change the behavior for default (not SPIR-V friendly) path:

```suggestion
    } else if (!UseTargetTypes) {
```

...took a while to understand that `UseTPT` ("using typed pointer type") and `supportsTypedPointers()` (supports typed pointers) are different things here 😵‍💫

vmaksimo (commented):
Am I right the change is for LLVM-16 branch only? Or are there any plans to backport this to other branches?

test/extensions/INTEL/SPV_INTEL_vector_compute/buffer_surface_intel_typed_pointers.ll
michalpaszkowski: Done

michalpaszkowski:
> Am I right the change is for LLVM-16 branch only? Or are there any plans to backport this to other branches?

You are right. This is just for LLVM 16 since it is the last LLVM version to support typed pointers. Earlier versions do not have this issue.

vmaksimo (approved):

PR #3945, opened 2026-08-03
Translate {uinc/udec}_wrap as opaque intrinsics and preserve AMDGPU metadata through NonSemantic.AuxData

maarquitos14:
The PR includes an extension to `NonSemantic.AuxData` spec to add `NonSemanticAuxDataInstructionMetadata`, which enables preservation of metadata for instructions other than `GlobalVariable` and `Function` objects. At the same time, it includes the implementation of this mechanism for some AMDGPU metadata. Finally, it includes the translation of `uinc_wrap` and `udec_wrap` as if they were opaque llvm intrinsics, so that they can be restored in the reverse translation to the exact original code.

### DIFF
--- a/docs/NonSemantic.AuxData.asciidoc
+++ b/docs/NonSemantic.AuxData.asciidoc
@@ -33,8 +33,8 @@ Version
 
 [width="40%",cols="25,25"]
 |========================================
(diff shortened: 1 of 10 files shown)


maarquitos14:
Split into #3964 and #3965.

PR #3946, opened 2026-08-03
Reject OpSelect with scalar result and vector condition

aobolensk:
Inspired by https://github.com/llvm/llvm-project/pull/193745

### DIFF
--- a/lib/SPIRV/SPIRVWriter.cpp
+++ b/lib/SPIRV/SPIRVWriter.cpp
@@ -7568,6 +7568,13 @@ LLVMToSPIRVBase::transBuiltinToInstWithoutDecoration(Op OC, CallInst *CI,
       SPIRVType *SPRetTy = nullptr;
       Type *RetTy = CI->getType();
       auto *F = CI->getCalledFunction();
+      if (OC == OpSelect)
+        BM->getErrorLog().checkError(
+            RetTy->isVectorTy() ||
+                !CI->getArgOperand(0)->getType()->isVectorTy(),
+            SPIRVEC_InvalidInstruction, CI,
+            "OpSelect with a scalar result requires a scalar boolean "
+            "condition\n");
       if (!RetTy->isVoidTy()) {
         SPRetTy = transScavengedType(CI);
       } else if (Args.size() > 0 && F->arg_begin()->hasStructRetAttr()) {
--- a/test/negative/select-vector-cond-scalar-result.ll
+++ b/test/negative/select-vector-cond-scalar-result.ll
@@ -1,0 +1,16 @@
+; RUN: llvm-as < %s -o %t.bc
+; RUN: not llvm-spirv %t.bc -o %t.spv 2>&1 | FileCheck %s
+
+; CHECK: InvalidInstruction: Can't translate llvm instruction:
+; CHECK: OpSelect with a scalar result requires a scalar boolean condition
+
+target datalayout = "e-i64:64-v16:16-v24:32-v32:32-v48:64-v96:128-v192:256-v256:256-v512:512-v1024:1024-n8:16:32:64"
+target triple = "spir64-unknown-unknown"
+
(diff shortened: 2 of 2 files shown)


aobolensk:

idubinov:
looks like that discussion about asserts that we had before.
It feels like it never hits that error check for input that previously passed spirv-val.

aobolensk:
> looks like that discussion about asserts that we had before. It feels like it never hits that error check for input that previously passed spirv-val.

Please see the motivation discussion from the original patch in this thread: https://github.com/llvm/llvm-project/pull/193745#discussion_r3147133536

lib/SPIRV/SPIRVWriter.cpp
```diff
@@ -7568,6 +7568,13 @@ LLVMToSPIRVBase::transBuiltinToInstWithoutDecoration(Op OC, CallInst *CI,
       SPIRVType *SPRetTy = nullptr;
       Type *RetTy = CI->getType();
       auto *F = CI->getCalledFunction();
+      if (OC == OpSelect)
+        BM->getErrorLog().checkError(
+            RetTy->isVectorTy() ||
+                !CI->getArgOperand(0)->getType()->isVectorTy(),
```
idubinov, line 7670: The condition may be bool with return type vector.

lib/SPIRV/SPIRVWriter.cpp
```diff
@@ -7568,6 +7568,13 @@ LLVMToSPIRVBase::transBuiltinToInstWithoutDecoration(Op OC, CallInst *CI,
       SPIRVType *SPRetTy = nullptr;
       Type *RetTy = CI->getType();
       auto *F = CI->getCalledFunction();
+      if (OC == OpSelect)
```
idubinov, line 7667: From my point of view, such check should be in spirv-val

aobolensk: As you can see there is this check in spirv-val. But valid LLVM IR can be transformed to the invalid SPIR-V as of now

lib/SPIRV/SPIRVWriter.cpp
aobolensk: "if Condition is a vector, Result Type must be a vector..." according to the spec

aobolensk: https://registry.khronos.org/SPIR-V/specs/unified1/SPIRV.html#OpSelect

idubinov: my mistake, condition is correct

MrSidims:
If you don't mind, I'd abstain from review :) I really don't have a strong opinion, whether we should bother about __spirv_Select way of exposing OpSelect, tho we have added such check in the SPIR-V backend per discussion in https://github.com/llvm/llvm-project/pull/193745#discussion_r3147133536

lib/SPIRV/SPIRVWriter.cpp
```diff
@@ -7679,6 +7679,13 @@ LLVMToSPIRVBase::transBuiltinToInstWithoutDecoration(Op OC, CallInst *CI,
...
+        BM->getErrorLog().checkError(
+            RetTy->isVectorTy() ||
+                !CI->getArgOperand(0)->getType()->isVectorTy(),
+            SPIRVEC_InvalidInstruction, CI,
+            "OpSelect with a scalar result requires a scalar boolean "
```
maarquitos14: According to the spec, the implication is the opposite: 

> If Condition is a vector, Result Type must be a vector with the same number of components as Condition

I'd go with "OpSelect with a vector condition requires a vector result." or similar.

maarquitos14 (approved):

lib/SPIRV/SPIRVWriter.cpp
maarquitos14: Just a nit, btw.

aobolensk: Rephrased

PR #3949, opened 2026-08-05
Diagnose unsupported module level inline assembly instead of silently dropping it

aobolensk:
Inspired by https://github.com/llvm/llvm-project/pull/199992

### DIFF
--- a/test/negative/module-asm-unsupported.ll
+++ b/test/negative/module-asm-unsupported.ll
@@ -1,5 +1,4 @@
-; RUN: llvm-as < %s -o %t.bc
-; RUN: not llvm-spirv %t.bc -o %t.spv 2>&1 | FileCheck %s
+; RUN: not llvm-spirv %s -o %t.spv 2>&1 | FileCheck %s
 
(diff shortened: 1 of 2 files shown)


aobolensk:

maarquitos14:
I think the linked PR is not the correct one, at least I can't see anything related to this.

maarquitos14 (approved):

aobolensk:
> I think the linked PR is not the correct one, at least I can't see anything related to this.

Replaced, thanks

test/negative/module-asm-unsupported.ll
```diff
@@ -0,0 +1,13 @@
+; RUN: llvm-as < %s -o %t.bc
+; RUN: not llvm-spirv %t.bc -o %t.spv 2>&1 | FileCheck %s
```
svenvh: You should be able to feed `%s` into llvm-spirv directly, without llvm-as.

aobolensk: Done

svenvh (approved):

PR #3951, opened 2026-08-06
Support new OpTypeVectorIdEXT op with constant instructions in Reader

Qi-Ye-079:
Adds reverse-translation of the new `OpTypeVectorIdEXT` op from `SPV_EXT_long_vector` extension in the Reader. Its component size must be a constant instruction (`OpConstant`/`OpSpecConstant`/`OpSpecConstantOp`).

The goal is to cover `OpTypeVectorIdEXT` for all instructions that can use `OpTypeVector` type. However, there's a fundamental difference when it comes to **checking the component count** for these instructions:
- We can't check component count in `validate()` methods of these instructions, like they do for `OpTypeVector`-typed results/operands, because the component count is an instruction which must be translated/evaluated to know the values.
- We can't _fully_ rely on the `spirv-val` to validate component count for instructions either, because `spirv-val` can't correctly evaluate specialization constants which can be overriden by users via TranslatorOpts.

Given these, the component check of `OpTypeVectorIdEXT` for instructions are performed by the Reader during `transValue`. For the same reason, a new `SPIRVTypeVectorIdEXT` class is added, since its state and implementation differ a lot from `SPIRVTypeVector`.


lib/SPIRV/libSPIRV/SPIRVType.h
```diff
@@ -554,6 +554,43 @@ class SPIRVTypeArray : public SPIRVType {
...
+  }
+  // Incomplete constructor
+  SPIRVTypeVectorIdEXT()
+      : SPIRVType(OpTypeVectorIdEXT), CompType(nullptr),
+        ComponentCount(SPIRVID_INVALID) {}
+
+  SPIRVType *getComponentType() const { return CompType; }
+  SPIRVValue *getComponentCount() const { return getValue(ComponentCount); }
+
+  SPIRVCapVec getRequiredCapability() const override {
```
vmaksimo, line 575: +maybe add `getRequiredExtension()` as well

lib/SPIRV/libSPIRV/SPIRVType.h
```diff
@@ -554,6 +554,43 @@ class SPIRVTypeArray : public SPIRVType {
   SPIRVId Length;      // Array Length
 };
 
+class SPIRVTypeVectorIdEXT : public SPIRVType {
```
vmaksimo, line 558: any sense or possibility to reuse `SPIRVTypeVector`? e.g. refactoring with the common base class

lib/SPIRV/libSPIRV/SPIRVType.h
```diff
@@ -554,6 +554,43 @@ class SPIRVTypeArray : public SPIRVType {
...
+  SPIRVCapVec getRequiredCapability() const override {
+    SPIRVCapVec V(getComponentType()->getRequiredCapability());
+    V.push_back(CapabilityLongVectorEXT);
+    return V;
+  }
+
+  std::vector<SPIRVEntry *> getNonLiteralOperands() const override {
+    std::vector<SPIRVEntry *> Operands(2, CompType);
+    Operands[1] = (SPIRVEntry *)getComponentCount();
+    return Operands;
```
vmaksimo: Isn't it an over-complication?
won't it work just like that?

```suggestion
    return std::vector<SPIRVEntry *>(1, CompType);
```

I was taking a look at the same method for `TypeCooperativeMatrixKHR` and it also does not use direct literals as their values and we are getting only type as non-literal operand (maybe there is a bug there, need to check):
https://github.com/KhronosGroup/SPIRV-LLVM-Translator/blob/221ab532b310cccb2577b9e6606a6619e641f758/lib/SPIRV/libSPIRV/SPIRVType.h#L1259

lib/SPIRV/libSPIRV/SPIRVType.cpp
```diff
@@ -308,6 +308,37 @@ SPIRVValue *SPIRVTypeArray::getLength() const { return getValue(Length); }
...
+  SPVErrLog.checkError(CompType->isTypeBool() || CompType->isTypeInt() ||
+                           CompType->isTypeFloat(),
+                       SPIRVEC_InvalidInstruction,
+                       InstName + ": component type must be a scalar type\n");
+  SPIRVValue *Count = getComponentCount();
+  // Component Count must be a constant instruction with 32-bit int type
+  Op CountOC = Count->getOpCode();
+  SPVErrLog.checkError(
+      CountOC == OpConstant || CountOC == OpSpecConstant ||
+          CountOC == OpSpecConstantOp,
```
vmaksimo, line 335:
```suggestion
      isConstantOpCode(CountOC)
```
will this work?

vmaksimo (commented):
..and a bit more review from claude code:

>   1. Missing isTypeVector() recognition for OpTypeVectorIdEXT. The PR adds a new vector-category type but never updates this predicate.
>  2. Missing getVectorComponentType()/getVectorComponentCount() dispatch for OpTypeVectorIdEXT. Both are  hardcoded to OpTypeVector (plus OpTypeCooperativeMatrixKHR for the type accessor only). Not extended for the new opcode.

These two above may seem questionable as no real use is intended; still there is a fair failure found:
 > 3. Consequence of #1/#2 — reproducible crashes on valid, spirv-val-clean input:
     - OpCompositeExtract on an OpTypeVectorIdEXT value → assertion failure in SPIRVCompositeExtractBase::validate()
    - OpCompositeConstruct building an OpTypeVectorIdEXT value → assert(false && "Invalid type") in SPIRVCompositeConstruct::validate()
    - OpVectorExtractDynamic on an OpTypeVectorIdEXT value → assertion failure in SPIRVVectorExtractDynamic::validate()
  All three reproduced locally with hand-assembled, spirv-val-passing .spvasm inputs run through the PR's built llvm-spirv -r. These are real-world-reachable (any consumer doing basic element access on a long vector hits this), not artificial edge cases.
>  4. Test gap: SPV_EXT_long_vector/OpTypeVectorIdEXT_constants.spvasm only passes OpTypeVectorIdEXT opaquely as a function parameter/return; it never exercises OpCompositeExtract/OpCompositeConstruct/vector-access, so the crashes above are untested and would not be caught.

For me it seems like at least the #4 should be addressed.

lib/SPIRV/libSPIRV/SPIRVType.h
Qi-Ye-079: Technically it works, but personally I prefer not to reuse `SPIRVTypeVector`, because some states and implementation would differ between `OpTypeVector` and `OpTypeVectorIdEXT`:
1. `SPIRVTypeVector` models `OpTypeVector`, whose component size is a constant literal, so its constructor takes a `SPIRVWord` directly, while the component size of `OpTypeVectorIdEXT` is a constant instruction whose value is only known after it's translated/evaluated, thus we need to keep it as an ID or `SPIRVEntry` in the class.
2. The `SPIRVTypeVector::getRequiredCapability` method adds the `CapabilityLongVectorEXT` or `CapabilityVectorAnyINTEL` capabilities _conditionally_ - when it involves non-standard component size; while for `SPIRVTypeVectorIdEXT`, the `CapabilityLongVectorEXT` capability is added unconditionally.
3. The `getNonLiteralOperands` difference - see the comment below.
4. The validation logic is completely different also :)

Given these, I think using a separate class is a cleaner, more expressive option. Otherwise, we need different constructor, states and implementation for one class and a flag to differentiate them all.

lib/SPIRV/libSPIRV/SPIRVType.h
Qi-Ye-079: Done

lib/SPIRV/libSPIRV/SPIRVType.h
Qi-Ye-079: Maybe I'm missing something, but from what I understand, this `getNonLiteralOperands` simply returns what the name suggests: get all the operands that aren't literals. So for `OpTypeVectorIdEXT`, it has 2 non-literal operands: the component type and the component size, both are instructions. Am I wrong?

lib/SPIRV/libSPIRV/SPIRVType.cpp
Qi-Ye-079: The issue with `isConstantOpCode` is that its scope is way too broad: is has some ops that shouldn't work with `OpTypeVectorIdEXT`, such as `OpConstantTrue`, `OpConstantFalse` or `OpUndef`. Here I'm explicitly checking the allowed ops (at least for now) for narrower coverage.

Qi-Ye-079:
> ..and a bit more review from claude code:
> 
> > 1. Missing isTypeVector() recognition for OpTypeVectorIdEXT. The PR adds a new vector-category type but never updates this predicate.
> > 2. Missing getVectorComponentType()/getVectorComponentCount() dispatch for OpTypeVectorIdEXT. Both are  hardcoded to OpTypeVector (plus OpTypeCooperativeMatrixKHR for the type accessor only). Not extended for the new opcode.
> 
> These two above may seem questionable as no real use is intended; still there is a fair failure found:
> 
> > 3. Consequence of #1/#2 — reproducible crashes on valid, spirv-val-clean input:
> >    - OpCompositeExtract on an OpTypeVectorIdEXT value → assertion failure in SPIRVCompositeExtractBase::validate()
> >    - OpCompositeConstruct building an OpTypeVectorIdEXT value → assert(false && "Invalid type") in SPIRVCompositeConstruct::validate()
> >    - OpVectorExtractDynamic on an OpTypeVectorIdEXT value → assertion failure in SPIRVVectorExtractDynamic::validate()
> >    All three reproduced locally with hand-assembled, spirv-val-passing .spvasm inputs run through the PR's built llvm-spirv -r. These are real-world-reachable (any consumer doing basic element access on a long vector hits this), not artificial edge cases.
> > 4. Test gap: SPV_EXT_long_vector/OpTypeVectorIdEXT_constants.spvasm only passes OpTypeVectorIdEXT opaquely as a function parameter/return; it never exercises OpCompositeExtract/OpCompositeConstruct/vector-access, so the crashes above are untested and would not be caught.
> 
> For me it seems like at least the #4 should be addressed.

@vmaksimo I'm not sure how much of these feedback is applicable. For these ops: `OpConstantComposite`, `OpSpecConstantComposite`, and `OpCompositeConstruct`, I'm not sure if it really makes sense for `OpTypeVectorIdEXT` to be used as operand for these ops. First I'm not sure whether there is a producer that will generate `OpTypeVectorIdEXT` as operand for them; secondly:
- If the component size is `OpConstant`, then its equivalent as `OpTypeVector`, and the producer should generate `OpTypeVector` in the first place;
- If the component size is a specialization constant, then you can't override its value via opts, because the number of the constituents for these ops are always fixed; it doesn't make any sense to use `OpTypeVectorIdEXT` with specialization constants for these composite ops.

Given above, I think for these composite ops, there's no valid reason to use `OpTypeVectorIdEXT` in the first place; a producer should always generate `OpTypeVector` as operand for them, if the producer needs. @bashbaug Please, give some feedback as to whether we need to support `OpTypeVectorIdEXT` being used as operand for these composite ops.

bashbaug:
> Please, give some feedback as to whether we need to support `OpTypeVectorIdEXT` being used as operand for these composite ops.

As long as it's valid SPIR-V - and I think all of the cases describe above are valid? - we need to handle them even if we aren't aware of a SPIR-V producer that generates them today.  For bonus points, we should have tests for these also, even if they're hand-written, to make sure these scenarios don't break in the future.

lib/SPIRV/libSPIRV/SPIRVType.h
vmaksimo: Yeah, I agree the name suggests exactly that. Just I found out the example (`TypeCooperativeMatrixKHR`) that does not follow this convention - could you please investigate if there is a bug there/in other classes that are using `getNonLiteralOperands()`, or is it only the function name that confuses us?

Qi-Ye-079: I consulted other types and am almost sure the `TypeCooperativeMatrixKHR::getNonLiteralOperands()` is wrong, it should return the remaining non-literal <id> operands (Scope/Rows/Columns/Use).
The closest type to `OpTypeVectorIdEXT` is `OpTypeArray`, and you can see it does return 2 <id> operands: the element type and length. Other types include `OpTypeStruct` and `OpTypeFunction` does return all <id> operands as well.

Qi-Ye-079:
@vmaksimo Following what Ben suggested, to support all valid SPIRV, I've tried to identify all instructions that support `OpTypeVector` types and added `OpTypeVectorIdEXT` support for them (hopefully I don't miss any), beyond just those composite ops you mentioned. Please take a look when you have time; there are lots of changes.

@bashbaug FYI, you can take a look too if you have time :)

Qi-Ye-079:
@vmaksimo @bashbaug A summary of our discussion as a reference (for others), regarding the component check for applicable instructions done in `SPIRVToLLVM::checkTypeVectorIdEXTComponentCount()`:

More validation is preferable in general, but there's a concern about compile time and code size if we do too much validation in the Reader (this PR grows very large with these checks and new tests).

As discussed, I've removed the checks for some instructions (binary/compare/shift/bitwise/logical) and LLVM's `Builder` can assert the validity of `FixedVectorSize` when building these instructions. We all agree that this is a fair middle ground and acceptable tradeoff.

lib/SPIRV/SPIRVReader.cpp
```diff
@@ -995,6 +1010,130 @@ void SPIRVToLLVM::transLLVMLoopMetadata(const Function *F) {
...
+    // FillEmpty(3).
+    auto *MG = static_cast<SPIRVMaskedGatherINTELInst *>(BV);
+    SPIRVType *RetTy = MG->getType();
+    SPIRVType *PtrVecTy = MG->getOperand(0)->getType();
+    SPIRVType *MaskTy = MG->getOperand(2)->getType();
+    SPIRVType *FillEmptyTy = MG->getOperand(3)->getType();
+    CheckEqualVectorComponentCounts(
+        {RetTy, PtrVecTy, MaskTy, FillEmptyTy},
+        "MaskedGatherINTEL: result component count must match the number of "
+        "components selected\n");
```
```suggestion
        "MaskedGatherINTEL: Result, PtrVector, Mask and FillEmpty vectors must have the same size\n");
```

lib/SPIRV/libSPIRV/SPIRVType.h
```diff
@@ -554,6 +555,47 @@ class SPIRVTypeArray : public SPIRVType {
...
+    return V;
+  }
+
+  std::optional<ExtensionID> getRequiredExtension() const override {
+    return ExtensionID::SPV_EXT_long_vector;
+  }
+
+  std::vector<SPIRVEntry *> getNonLiteralOperands() const override {
+    std::vector<SPIRVEntry *> Operands(2, CompType);
+    Operands[1] = (SPIRVEntry *)getComponentCount();
```
vmaksimo: initialize 2nd element and rewrite on the next line looks strange, maybe this is better:
```suggestion
    std::vector<SPIRVEntry *> Operands ={ CompType, (SPIRVEntry *)getComponentCount()};
```

lib/SPIRV/SPIRVReader.cpp
```diff
@@ -386,6 +386,18 @@ Type *SPIRVToLLVM::transType(SPIRVType *T, bool UseTPT) {
     return mapType(T,
                    FixedVectorType::get(transType(T->getVectorComponentType()),
                                         T->getVectorComponentCount()));
+  case OpTypeVectorIdEXT: {
+    // The component size might be a specialization constant, that needs to be
+    // specialized and evaluated before the FixedVectorType can be constructed
+    auto *VT = static_cast<const SPIRVTypeVectorIdEXT *>(T);
+    auto *CountValue = cast<ConstantInt>(
+        transValue(VT->getComponentCount(), nullptr, nullptr));
```
```suggestion
    auto *CountValue = dyn_cast<ConstantInt>(
        transValue(VT->getComponentCount(), nullptr, nullptr));
    BM->getErrorLog().checkError(
        CountValue, SPIRVEC_InvalidInstruction,
        "TypeVectorIdEXT: component count must evaluate to a constant integer\n");

```

vmaksimo: Otherwise it'll fail for smth like:
```
%count_op  = OpSpecConstantOp %uint UDiv %uint_0 %uint_0
      %vec = OpTypeVectorIdEXT %float %count_op
```

lib/SPIRV/libSPIRV/SPIRVInstruction.h
vmaksimo, line 1: For now the `SPV_EXT_long_vector` extension does not describe the interaction with other extensions we implement here (`SPV_INTEL_bfloat16_conversion, SPV_INTEL_masked_gather_scatter, SPV_INTEL_ternary_bitwise_function, SPV_INTEL_tensor_float32_conversion`).
I see only `SPV_KHR_integer_dot_product` was actually incorporated in the core spec.

Do we really need to support id-sized vector for non-core extensions right now? I guess no, and suggest to remove such if possible.

lib/SPIRV/SPIRVReader.cpp
```diff
@@ -995,6 +1010,130 @@ void SPIRVToLLVM::transLLVMLoopMetadata(const Function *F) {
...
+         (OC >= OpIAddCarry && OC <= OpSMulExtended);
+}
+
+// Evaluate and check component counts of OpTypeVectorIdEXT, for instructions
+// that already check it for OpTypeVector types. Not all instructions are
+// checked: for binary/compare/shift/bitwise/logical ops, we can rely on LLVM's
+// Builder.CreateBinOp/CreateICmp/CreateFCmp that assert operand types for
+// vector length. We don't want the Reader to do too much validation as there's
+// a concern for code size and compile time, so omitting some instructions as
+// tradeoff.
```
vmaksimo: Let's avoid over-explanation, e.g.:
```suggestion
// Checks OpTypeVectorIdEXT component counts for instructions that mirror
// OpTypeVector's validation. Binary/compare/shift/bitwise/logical ops are
// skipped since LLVM's Builder already asserts on mismatched vector lengths.
```

vmaksimo (commented):
note: i haven't reviewed the tests yet

lib/SPIRV/libSPIRV/SPIRVInstruction.h
Qi-Ye-079: I'm ok either way. If you think supporting those non-core extensions is unnecessary and introduce more code and test changes, I can remove them and keep changes for core extensions for sure.
Please give a thumb-up to confirm so that I can remove those changes.

lib/SPIRV/libSPIRV/SPIRVType.h
Qi-Ye-079: It's just directly copied from `OpTypeArray`, but yeah this should be cleaner.

lib/SPIRV/SPIRVReader.cpp
Qi-Ye-079: Done

lib/SPIRV/SPIRVReader.cpp
Qi-Ye-079: Done

lib/SPIRV/SPIRVReader.cpp
Qi-Ye-079: Thanks for catching for this edge case (div by 0), which can't be validated by `SPIRVTypeVectorIdEXT::validate()`.

lib/SPIRV/libSPIRV/SPIRVInstruction.h
```diff
@@ -3989,7 +4020,6 @@ class SPIRVMaskedGatherINTELInst
     SPIRVType *ResTy = this->getType();
     SPVErrLog.checkError(ResTy->isTypeVector(), SPIRVEC_InvalidInstruction,
                          InstName + "\nResult must be a vector type\n");
-    SPIRVWord ResCompCount = ResTy->getVectorComponentCount();
```
vmaksimo, line 3992: Still seeing file modifications for masked gather/scatter in full PR diff.

lib/SPIRV/SPIRVReader.cpp
```diff
@@ -995,6 +1013,79 @@ void SPIRVToLLVM::transLLVMLoopMetadata(const Function *F) {
   }
 }
 
+static bool isVectorBinaryOpWithoutVectorResult(Op OC) {
+  // these opcodes can have 2 vector operands but the result isn't vector type
+  return (OC >= OpSDotKHR && OC <= OpSUDotAccSatKHR) || OC == OpDot ||
+         (OC >= OpIAddCarry && OC <= OpSMulExtended);
+}
```
vmaksimo: Nit: I'd refactor this to either lambda inside `checkTypeVectorIdEXTComponentCount` or as a function in SPIRVUtil

lib/SPIRV/SPIRVReader.cpp
```diff
@@ -1348,10 +1440,17 @@ Value *SPIRVToLLVM::transCmpInst(SPIRVValue *BV, BasicBlock *BB, Function *F) {
   if (OP == OpLessOrGreater)
     OP = OpFOrdNotEqual;
 
-  if (BT->isTypeVectorOrScalarInt() || BT->isTypeVectorOrScalarBool() ||
-      BT->isTypePointer())
+  bool IsIntLike = BT->isTypeVectorOrScalarInt() ||
+                   BT->isTypeVectorOrScalarBool() || BT->isTypePointer();
+  bool IsFloatLike = BT->isTypeVectorOrScalarFloat();
```
vmaksimo: To double-check, isn't these code changes were related to bfloat16 conversions? If so, we don't need this anymore

vmaksimo (commented):
I know we wanted to reduce the number of the tests wherever possible. 
But I wonder if we should add test case for mixed OpTypeVector + OpTypeVectorIdEXT operands? e.g. like:
```
     %uint_5 = OpConstant %uint 5
    %v5float = OpTypeVector %float 5
 %v5float_id = OpTypeVectorIdEXT %float %uint_5
      %fnty = OpTypeFunction %v5float %v5float %v5float_id

%test_mixed = OpFunction %v5float None %fnty
          %a = OpFunctionParameter %v5float
          %b = OpFunctionParameter %v5float_id
          %e = OpLabel
          %r = OpFAdd %v5float %a %b
               OpReturnValue %r
               OpFunctionEnd
```

lib/SPIRV/libSPIRV/SPIRVInstruction.h
Qi-Ye-079: Removed

lib/SPIRV/SPIRVReader.cpp
Qi-Ye-079: I simply inlined it, since there's no need to keep it as a function anymore.

lib/SPIRV/SPIRVReader.cpp
Qi-Ye-079: No these are for comparison instructions, not related to BF16 conversions.

Qi-Ye-079:
> I know we wanted to reduce the number of the tests wherever possible. But I wonder if we should add test case for mixed OpTypeVector + OpTypeVectorIdEXT operands? e.g. like:
> 
> ```
>      %uint_5 = OpConstant %uint 5
>     %v5float = OpTypeVector %float 5
>  %v5float_id = OpTypeVectorIdEXT %float %uint_5
>       %fnty = OpTypeFunction %v5float %v5float %v5float_id
> 
> %test_mixed = OpFunction %v5float None %fnty
>           %a = OpFunctionParameter %v5float
>           %b = OpFunctionParameter %v5float_id
>           %e = OpLabel
>           %r = OpFAdd %v5float %a %b
>                OpReturnValue %r
>                OpFunctionEnd
> ```

I couldn't because `spirv-val` rejects mixed `OpTypeVector` + `OpTypeVectorIdEXT` case for `OpFAdd` and many others, as it requires the rsult and operands to be identical types. I found `OpVectorShuffle` as a candidate and added the mixed case for it.

vmaksimo (approved):
LGTM, let's now gather the community feedback as well

Qi-Ye-079:
@MrSidims @svenvh Could any of you please help review this PR? Thanks

lib/SPIRV/SPIRVReader.cpp
```diff
@@ -386,6 +386,21 @@ Type *SPIRVToLLVM::transType(SPIRVType *T, bool UseTPT) {
...
                                         T->getVectorComponentCount()));
+  case OpTypeVectorIdEXT: {
+    // The component size might be a specialization constant, that needs to be
+    // specialized and evaluated before the FixedVectorType can be constructed
+    auto *VT = static_cast<const SPIRVTypeVectorIdEXT *>(T);
+    auto *CountValue = dyn_cast<ConstantInt>(
+        transValue(VT->getComponentCount(), nullptr, nullptr));
+    BM->getErrorLog().checkError(CountValue, SPIRVEC_InvalidInstruction,
+                                 "TypeVectorIdEXT: component count must "
+                                 "evaluate to a constant integer\n");
```
MrSidims: I always forget, does `BM->getErrorLog().checkError` immediately aborts the translation in case if the first condition is false? If no - then `!CountValue->isZero()` next line is NULL dereference.

MrSidims: Per my understanding checkError does not unwind - it only exits because SPIRVDbgError defaults to Exit. So an explicit return after each check is required, not optional.

lib/SPIRV/libSPIRV/SPIRVInstruction.h
```diff
@@ -3361,8 +3387,13 @@ class SPIRVDotKHRBase : public SPIRVInstTemplateBase {
     (void)Vec1;
     (void)Vec2;
 
-    assert(getValueType(Vec1) == getValueType(Vec2) &&
-           "Input vectors must have the same type");
+    // getValueType(Vec1) == getValueType(Vec2) is pointer identity; for a
+    // long-vector-id (SPV_EXT_long_vector) two distinct type <id>s of equal
+    // count are valid, so the count is verified in the reader instead.
```
MrSidims: Why? See https://registry.khronos.org/SPIR-V/specs/unified1/SPIRV.html#TypesVariables :

> It is invalid to declare multiple non-aggregate, non-pointer type <id>s having the same opcode and operands.

and https://github.com/KhronosGroup/SPIRV-Registry/blob/main/extensions/EXT/SPV_EXT_long_vector.asciidoc doesn't lift off this restriction.

lib/SPIRV/SPIRVReader.cpp
MrSidims, line 3825: Should it also be updated?

lib/SPIRV/SPIRVReader.cpp
```diff
@@ -565,6 +580,9 @@ std::string SPIRVToLLVM::transTypeToOCLTypeName(SPIRVType *T, bool IsSigned) {
   case OpTypeVector:
     return transTypeToOCLTypeName(T->getVectorComponentType()) +
            T->getVectorComponentCount();
+  case OpTypeVectorIdEXT:
+    return transTypeToOCLTypeName(T->getVectorComponentType()) +
+           cast<FixedVectorType>(transType(T))->getNumElements();
```
MrSidims, line 588: So, if we have a <5 x float>, we get `float5` in the mangling? I don't know, what should be here, better to consult with @bashbaug and see, what OpenCL env spec mandates here.

MrSidims (commented):
May be we should extend the count-independent predicates instead of patching call sites? 

isTypeVectorBool/Int/Float, isTypeVectorOrScalarBool/Int/Float, isTypeIEEE754Float, and isTypeVectorPointer all remain OpCode == OpTypeVector-only, even though none of them look at the count - they only need getVectorComponentType(), which this PR already taught about OpTypeVectorIdEXT.

I also spotted, that the patch doesn't touch gather/scatter, is it indented/in TODO?

test/extensions/EXT/SPV_EXT_long_vector/negative/OpTypeVectorIdEXT_dot_count_mismatch.spvasm
```diff
@@ -0,0 +1,35 @@
+; Checks the reader rejects OpSDotKHR when its OpTypeVectorIdEXT input vectors have
+; different component counts. Without the fix the reader aborts on a Debug
+; type-identity assert (Release: silent wrong IR); the count is verified in the
+; reader from the translated FixedVectorTypes.
```
MrSidims: Please humanize the comments in the tests.

lib/SPIRV/libSPIRV/SPIRVInstruction.h
Qi-Ye-079: Sure I agree, but you can have 2 different `OpTypeVectorIdEXT` with different count `<id>`s like the following:
```
%c0 = OpConstant %uint 5
%c1 = OpSpecConstant %uint 5
%t0 = OpTypeVectorIdEXT %float %c0
%t1 = OpTypeVectorIdEXT %float %c1
```
I don't think this case violates the spec, as their operands are different? The `spirv-val` doesn't throw error in this case (because it can't check `OpTypeVectorIdEXT` properly), and the `getValueType(Vec1) == getValueType(Vec2)` assert will trigger, so I think excluding `OpTypeVectorIdEXT` here and checking it in the Reader is necessary.

lib/SPIRV/SPIRVReader.cpp
Qi-Ye-079: You're right. Done.

lib/SPIRV/SPIRVReader.cpp
Qi-Ye-079: I can't say for sure about the OCL env spec, but this is simply to align with the mangling for `OpTypeVector` above. If this is wrong, then the mangling for `OpTypeVector` should also be wrong, because when `SPV_EXT_long_vector` is enabled, `OpTypeVector` is also allowed to have non-standard size such as 5.

lib/SPIRV/SPIRVReader.cpp
Qi-Ye-079: I took your suggestion to extend isTypeVectorBool/Int/Float to also cover `OpTypeVectorIdEXT`, so no need to update here; also I've reverted it for the `transOCLBuiltinFromInstPreproc` for the same reason.

test/extensions/EXT/SPV_EXT_long_vector/negative/OpTypeVectorIdEXT_dot_count_mismatch.spvasm
Qi-Ye-079: Done

Qi-Ye-079:
> I also spotted, that the patch doesn't touch gather/scatter, is it indented/in TODO?

This is intended and suggested by Viktoria: https://github.com/KhronosGroup/SPIRV-LLVM-Translator/pull/3951#discussion_r3881082172.

> May be we should extend the count-independent predicates instead of patching call sites?
> 
> isTypeVectorBool/Int/Float, isTypeVectorOrScalarBool/Int/Float, isTypeIEEE754Float, and isTypeVectorPointer all remain OpCode == OpTypeVector-only, even though none of them look at the count - they only need getVectorComponentType(), which this PR already taught about OpTypeVectorIdEXT.

I agree, I've extended `OpTypeVectorIdEXT` coverage for `isTypeVectorBool`/`Int`/`Float` (which transitively applies to `isTypeVectorOrScalarBool`/`Int`/`Float` as well).

I won't touch the rest though:
- `isTypeIEEE754Float`: only used in writer right now
- `isTypeVectorPointer`: not applicable ~~because I think [the spec](https://github.khronos.org/SPIRV-Registry/extensions/EXT/[SPV_EXT_long_vector](https://github.khronos.org/SPIRV-Registry/extensions/EXT/SPV_EXT_long_vector.html#_modifications_to_the_spir_v_specification_version_1_6).html#_modifications_to_the_spir_v_specification_version_1_6) only allows the component type to be a Scalar Type though~~:
  ```
  OpTypeVectorIdEXT

  Declare a new vector type with Component Count components of the requested scalar type.

  Component Type is the type of each component in the resulting type. It must be a scalar type.
  ```
  ~~Technically speaking this should also apply to `OpTypeVector`, but it currently allows pointer type as its component type. Isn't this a violation of spec?~~
  UPD: just found out that this is allowed only by `SPV_INTEL_masked_gather_scatter` extension; however, we've decided to only cover extensions/instruction from core spec for now as suggested by Viktoria, I still won't extend this predicate in this PR.

lib/SPIRV/SPIRVReader.cpp
MrSidims: > OpTypeVector is also allowed to have non-standard size such as 5.

If so - it contradicts the spec. [SPV_EXT_long_vector](https://github.com/KhronosGroup/SPIRV-Registry/blob/main/extensions/EXT/SPV_EXT_long_vector.asciidoc) only extends **OpTypeVector** definition by allowing [component count of 1](https://github.com/KhronosGroup/SPIRV-Registry/blob/main/extensions/EXT/SPV_EXT_long_vector.asciidoc#3566-type-declaration-instructions). Lets say 5-component vector _must_ be declared as **OpTypeVectorIdEXT**.

MrSidims: May be it comes from ESIMD extension?

lib/SPIRV/libSPIRV/SPIRVInstruction.h
MrSidims: I must admit, I like this interpretation as it allows many interesting things :) But unfortunately, that is invalid SPIR-V (assuming you feed these vectors to dot product). See `OpSDot` defintion:

> Vector 1 and Vector 2 **must have the same type**.

And "the same type" means literally the same type - sharing the same `<id>`.

lib/SPIRV/libSPIRV/SPIRVInstruction.h
```diff
@@ -3361,8 +3387,13 @@ class SPIRVDotKHRBase : public SPIRVInstTemplateBase {
...
+    if (!getValueType(Vec1)->isTypeVectorIdEXT() &&
+        !getValueType(Vec2)->isTypeVectorIdEXT())
```
MrSidims:
```suggestion
    if (!getValueType(Vec1)->isTypeVectorIdEXT() ||
        !getValueType(Vec2)->isTypeVectorIdEXT())
```
?

lib/SPIRV/SPIRVReader.cpp
bashbaug: Hmm, I'm not sure exactly what should go here, but I'm not sure that it matters much, either.  It looks like this function is mostly only used by `transOCLKernelArgTypeName` to setup kernel arg metadata. Is this correct? If so, something like `float5` is probably fine for now, until we can convince ourselves that something else is more appropriate.

FWIW, I don't think the OpenCL SPIR-V environment spec will say anything about this, so that wouldn't be the right document. Maybe if there's an OpenCL C spec for extended vectors that might provide some future guidance?

lib/SPIRV/libSPIRV/SPIRVInstruction.h
Qi-Ye-079: Ok, thanks

lib/SPIRV/libSPIRV/SPIRVInstruction.h
Qi-Ye-079: From you comment above, if their types must be the same `<id>`, then this guard is no longer needed and should be reverted back to original. The same should apply to opcodes in `isBinaryOpCode` inside `SPIRVBinary` class, and ops in `isGenericNegateOpCode` inside `SPIRVUnary` class as well.

lib/SPIRV/SPIRVReader.cpp
Qi-Ye-079: Thanks for Ben's comment, I'll just leave it as-is for now unless someone has a better idea than `float5`.

Qi-Ye-079:
@MrSidims I've fixed all your last round of comments. Could you please take a look and see if you're ok with them? Thanks.

lib/SPIRV/libSPIRV/SPIRVInstruction.h
MrSidims: The assertion in dot_product::validate still has to be updated to allow usage of OpTypeVectorIdEXT

lib/SPIRV/libSPIRV/SPIRVType.cpp
MrSidims, line 62: Should we update this as well?

lib/SPIRV/SPIRVReader.cpp
```diff
@@ -2700,6 +2725,16 @@ Value *SPIRVToLLVM::transValueWithoutDecoration(SPIRVValue *BV, Function *F,
...
     auto *VS = static_cast<SPIRVVectorShuffle *>(BV);
+    if (BV->getType()->isTypeVectorIdEXT()) {
+      unsigned ResCount =
+          cast<FixedVectorType>(transType(BV->getType()))->getNumElements();
+      if (!BM->getErrorLog().checkError(
+              ResCount == VS->getComponents().size(),
+              SPIRVEC_InvalidInstruction,
+              "VectorShuffle: result component count must match the number of "
+              "components selected\n"))
+        return nullptr;
```
MrSidims, line 2736: Should we also check input?

test/extensions/EXT/SPV_EXT_long_vector/OpTypeVectorIdEXT_vector_ops.spvasm
```diff
@@ -0,0 +1,80 @@
...
+               OpDecorate %fn_vid LinkageAttributes "test_vec_insert_dynamic" Export
+               OpDecorate %fn_vs LinkageAttributes "test_vec_shuffle" Export
+               OpDecorate %fn_ac LinkageAttributes "test_access_chain" Export
+       %uint = OpTypeInt 32 0
+      %float = OpTypeFloat 32
+     %uint_5 = OpConstant %uint 5
+   %count_sc = OpSpecConstant %uint 5
+    %v5float = OpTypeVectorIdEXT %float %uint_5
+ %v5float_sc = OpTypeVectorIdEXT %float %count_sc
+  %pv5float = OpTypeVector %float 5
```
MrSidims: This is invalid OpTypeVector .

lib/SPIRV/SPIRVReader.cpp
Qi-Ye-079: Done. Added bounds checking w.r.t input vectors.

lib/SPIRV/libSPIRV/SPIRVType.cpp
Qi-Ye-079: Done.

test/extensions/EXT/SPV_EXT_long_vector/OpTypeVectorIdEXT_vector_ops.spvasm
Qi-Ye-079: I've changed all testcases that use `OpTypeVector` to size 4.

However, from my understanding of the spec, when `LongVectorEXT` capability is enabled, the `OpTypeType` should be allowed to have a size 5:
```
LongVectorEXT
Enables vector types with more than four components, one component, and OpTypeVectorIdEXT.
```
Also the `spirv-val` accepts any positive size for `OpTypeVector` when `LongVectorEXT` is enabled, in source/val/validate_type.cpp:
```
auto num_components = inst->GetOperandAs<const uint32_t>(2);
  if (num_components == 2 || num_components == 3 || num_components == 4) {
    return SPV_SUCCESS;
  } else if (num_components > 0 &&
             _.HasCapability(spv::Capability::LongVectorEXT)) {
    return SPV_SUCCESS;                                // ← long-vector path
  } else if (num_components == 8 || num_components == 16) {
    if (_.HasCapability(spv::Capability::Vector16)) {
      return SPV_SUCCESS;
    }
    return _.diag(...) << "requires the Vector16 or LongVectorEXT capability";
  } else {
    return _.diag(...) << "Illegal number of components ...";
  }
```
So if it's really illegal, then both `SPIRVTypeVector::validate()` and the `spirv-val` are wrong and should be fixed.

lib/SPIRV/libSPIRV/SPIRVInstruction.h
Qi-Ye-079: It's already updated and support `OpTypeVectorIdEXT`, which is covered in OpTypeVectorIdEXT_value_operand_ops.spvasm testcase. As you mentioned, I reverted to the unconditional `assert(getValueType(Vec1) == getValueType(Vec2)` because the input types must be the same `<id>`. I also updated the last assert to also cover `OpTypeVectorIdEXT` too though.

test/extensions/EXT/SPV_EXT_long_vector/OpTypeVectorIdEXT_vector_ops.spvasm
MrSidims: Apparently I have misinterpreted the specification. I assumed, that it adds a "vector type" definition as a collection of OpTypeVector and OpTypeVectorIdEXT, but it's not. Also I thought, legal widths parametrization is a part of OpTypeVector description, therefore a line from the extension saying:

> Modify OpTypeVector to allow a Component Count of 1.

allows only one extra element size.

But apparently the rule

> Vector types must be parameterized only with 2, 3, or 4 components, plus any additional sizes enabled by [capabilities](https://registry.khronos.org/SPIR-V/specs/unified1/SPIRV.html#Capability).

is in verification section in the spec. Therefore just having a capability is enough to relax the rules.

Please accept my apologies and restore the original vector type in the test.

MrSidims (approved):

lib/SPIRV/libSPIRV/SPIRVInstruction.h
```diff
@@ -735,12 +735,16 @@ class SPIRVBinary : public SPIRVInstTemplateBase {
...
       Op2Ty = getValueType(Op2)->getVectorComponentType();
-      assert(getValueType(Op1)->getVectorComponentCount() ==
-                 getValueType(Op2)->getVectorComponentCount() &&
-             "Inconsistent Vector component width");
+      // getVectorComponentCount() won't work for OpTypeVectorIdEXT, as its
+      // count is not a literal and only known after being translated.
+      if (getValueType(Op1)->isTypeVector())
+        assert(getValueType(Op1)->getVectorComponentCount() ==
+                   getValueType(Op2)->getVectorComponentCount() &&
+               "Inconsistent Vector component width");
```
MrSidims, line 747: This actually has the same problem as dot product instructions had previously. The type **must** be the same for operands. See as an example:

> **OpFAdd**
> 
> Floating-point addition of Operand 1 and Operand 2.
> 
> Result Type must be a scalar or vector of [floating-point type](https://registry.khronos.org/SPIR-V/specs/unified1/SPIRV.html#Floating).
> 
> _The types of Operand 1 and Operand 2 both must be the same as Result Type._
> 
> Results are computed per component.

MrSidims (changes requested):
Overall LGTM, but validate methods should be double checked.

lib/SPIRV/libSPIRV/SPIRVInstruction.h
Qi-Ye-079: My change only involves guarding the original component count checks for `OpTypeVector`, because `getVectorComponentCount()` won't work for `OpTypeVectorIdEXT`.
The original validation code doesn't check the vector type `<id>` as a whole, but component type and count are checked separately. Here I'm just making the minimal change possible for my PR. Such inconsistency has always existed in the codebase and not introduced by me.
If we need to refactor the validation to check type `<id>` just like `SPIRVDotKHRBase::validate()`, then it's not just for this `SPIRVBinary` class, but also for many others too, to maintain consistency for all these instruction classes. I think this is out of scope and should deserve a separate PR.

MrSidims (approved):

PR #3961, opened 2026-08-10
Fix memory leak in isInternalSPIRVBuiltin

ustachow:
itaniumDemangle returns a malloc'd buffer that the caller must free, but
its result was assigned directly to a StringRef, so the buffer was never
released. This leaked on every internal FP4/FP8/int4 conversion builtin
processed by transFunctionDecl and transDirectCallInst.

The demangled name is simply the length-prefixed identifier already
present in the mangled name, so extract it as a substring of Name (the same
pattern used by the other demanglers above) instead of calling
itaniumDemangle. This removes the allocation entirely, so there is nothing
to leak.

### DIFF
--- a/lib/SPIRV/SPIRVUtil.cpp
+++ b/lib/SPIRV/SPIRVUtil.cpp
@@ -498,10 +498,13 @@ bool isInternalSPIRVBuiltin(StringRef Name, StringRef &DemangledName) {
     return false;
   constexpr unsigned DemangledNameLenStart = 2;
   size_t Start = Name.find_first_not_of("0123456789", DemangledNameLenStart);
-  if (!Name.substr(Start, Name.size() - 1)
-           .starts_with(kSPIRVName::InternalBuiltinPrefix))
+  if (!Name.substr(Start).starts_with(kSPIRVName::InternalBuiltinPrefix))
     return false;
-  DemangledName = llvm::itaniumDemangle(Name.data(), false);
+  size_t Len = 0;
+  if (Name.substr(DemangledNameLenStart, Start - DemangledNameLenStart)
+          .getAsInteger(10, Len))
+    return false;
+  DemangledName = Name.substr(Start, Len);
   DemangledName.consume_front(kSPIRVName::InternalBuiltinPrefix);
   return true;
 }


lib/SPIRV/SPIRVUtil.cpp
```diff
@@ -491,17 +492,23 @@ bool oclIsBuiltin(StringRef Name, StringRef &DemangledName, bool IsCpp) {
...
   if (!Name.substr(Start, Name.size() - 1)
            .starts_with(kSPIRVName::InternalBuiltinPrefix))
     return false;
-  DemangledName = llvm::itaniumDemangle(Name.data(), false);
+  char *Demangled = llvm::itaniumDemangle(Name.data(), /*ParseParams=*/false);
+  if (!Demangled)
+    return false;
+  Storage = Demangled;
+  std::free(Demangled);
+  DemangledName = Storage;
```
MrSidims: While this is correct - it looks redundant. We could do some RAII magic instead of introducing a new storage, naive version would be looking like this:
```
bool isInternalSPIRVBuiltin(StringRef Name, StringRef &DemangledName) {
    ...
    std::unique_ptr<char, decltype(&std::free)> D{
        llvm::itaniumDemangle(Name.data(), /*ParseParams=*/false), std::free};
    if (!D)
      return false;
    DemangledName = D.get();
    DemangledName.consume_front(kSPIRVName::InternalBuiltinPrefix);
    return true;
  }
```
But frankly, isInternalSPIRVBuiltin could be re-written into:
```
  bool isInternalSPIRVBuiltin(StringRef Name, StringRef &DemangledName) {
    if (!Name.starts_with("_Z"))
      return false;
    constexpr unsigned DemangledNameLenStart = 2;
    size_t Start = Name.find_first_not_of("0123456789", DemangledNameLenStart);
    if (!Name.substr(Start).starts_with(kSPIRVName::InternalBuiltinPrefix))
      return false;
    size_t Len = 0;
    if (Name.substr(DemangledNameLenStart, Start - DemangledNameLenStart)
            .getAsInteger(10, Len))
      return false;
    DemangledName.consume_front(kSPIRVName::InternalBuiltinPrefix);
    return true;
  }
```
and that would be it, no itanium demangling is needed.

Bottom line caller-supplied Storage out-param is the one shape that gives you neither RAII's safety guarantee nor std::string's simplicity.

MrSidims (commented):
This reminds me, that I had a branch with ASAN CI workflow, lets me trim it and push some time later this week.

lib/SPIRV/SPIRVUtil.cpp
ustachow: Thanks, agreed! I went for a minimal fix initially to avoid changing the existing behavior more than needed, but you're right that the second version is much cleaner. Updating the PR now. And great to hear you're adding the ASAN workflow . :)

MrSidims (approved):
lgtm

svenvh (approved):

vmaksimo (approved):

PR #3962, opened 2026-08-10
Translate MaximumRegisters MD through SPV_INTEL_maximum_registers

sarnex:
Currently we translate the `RegisterAllocMode` metadata through `SPV_INTEL_maximum_registers` if available, and the user semantic `num-thread-per-eu` otherwise.

If we always translate `RegisterAllocMode` through `SPV_INTEL_maximum_registers` if it is available, it will break users who are using a consumer that doesn't support this extension as it will error on seeing this SPIR-V.

Instead, we rework 92ea64b77207d28e0318950358276747220b6e22 and introduce new metadata,`MaximumRegisters`, that is always translated through `SPV_INTEL_maximum_registers`, and always translate `RegisterAllocMode` through `num-threads-per-eu`.

The format of the new metadata is closer to the `SPV_INTEL_maximum_registers` extension. If it's a numeric literal, that is the literal number of registers requested. If it's a string, it must be `AutoINTEL`, the only one allowed by the spec, to request the consumer choose the maximum number of registers.

This allows Intel's SYCL compiler (which will generate the IR with this metadata) to choose if it wants to require the `SPV_INTEL_maximum_registers` extension or not.

For some background, we plan to enable `SPV_INTEL_maximum_registers` in Intel's SYCL compiler by default, and we will introduce a new SYCL language mechanism that will end up generating the `MaximumRegisters` metadata, and keep using the `RegisterAllocMode` metadata if the user does not opt in. This allows us to not break existing users while still allowing those who want to use the extension to use it.

### DIFF
--- a/test/extensions/INTEL/SPV_INTEL_maximum_registers/maxreg_extension.ll
+++ b/test/extensions/INTEL/SPV_INTEL_maximum_registers/maxreg_extension.ll
@@ -11,45 +11,45 @@
 ; CHECK-SPIRV: EntryPoint [[#]] [[#FUNC3:]] "main_l13"
 ; CHECK-SPIRV: EntryPoint [[#]] [[#FUNC4:]] "main_l19"
 
-; CHECK-SPIRV: ExecutionMode [[#FUNC0]] 6461 2
-; CHECK-SPIRV: ExecutionMode [[#FUNC1]] 6461 1
+; CHECK-SPIRV: ExecutionMode [[#FUNC0]] 6461 256
+; CHECK-SPIRV: ExecutionMode [[#FUNC1]] 6461 128
 ; CHECK-SPIRV: ExecutionMode [[#FUNC2]] 6463 0
 ; CHECK-SPIRV: ExecutionModeId [[#FUNC3]] 6462 [[#Const3:]]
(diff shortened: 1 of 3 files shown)


vmaksimo:
Note, out-of-tree failure is not related to your PR

sarnex:
PR sniped, thanks :)

sarnex:
I can't add reviewers for some reason but I'd appreciate a review from @vmaksimo and @MrSidims, thanks!

MrSidims (commented):
no objections from our side

test/extensions/INTEL/SPV_INTEL_maximum_registers/maxreg_extension.ll
vmaksimo, line 1: just curious, the numbers of register are changed to reflect real-life IR?

vmaksimo (approved):
LGTM

test/extensions/INTEL/SPV_INTEL_maximum_registers/maxreg_extension.ll
sarnex: yep, just to reflect the actual use case, the actual number itself doesn't matter for the test and it was passing with the original numbers

sarnex:
thanks @vmaksimo! can you or @MrSidims please merge? thanks again

PR #3963, opened 2026-08-11
Mangle `OpUMulExtended` builtin arguments as unsigned

vmaksimo:
Reverse translation mangled `OpUMulExtended`'s operands with signed suffixes (e.g. `ll` for i64) instead of unsigned (`mm`), causing undefined-reference errors for consumers expecting the unsigned name.

### DIFF
--- a/lib/SPIRV/OCLUtil.cpp
+++ b/lib/SPIRV/OCLUtil.cpp
@@ -1220,6 +1220,8 @@ void OCLBuiltinFuncMangleInfo::init(StringRef UniqName) {
   } else if (NameRef.starts_with("u_")) {
     addUnsignedArg(-1);
     NameRef = NameRef.drop_front(2);
+  } else if (NameRef == "__spirv_UMulExtended") {
+    addUnsignedArg(-1);
   } else if (NameRef == "fclamp") {
     NameRef = NameRef.drop_front(1);
   }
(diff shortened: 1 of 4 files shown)


lib/SPIRV/OCLUtil.cpp
```diff
@@ -1220,6 +1220,8 @@ void OCLBuiltinFuncMangleInfo::init(StringRef UniqName) {
   } else if (NameRef.starts_with("u_")) {
     addUnsignedArg(-1);
     NameRef = NameRef.drop_front(2);
+  } else if (NameRef == "__spirv_UMulExtended") {
```
aratajew, line 1223: Why is it implemented in OCLUtil? It is SPIR-V Friendly IR, so I think it should be implemented in https://github.com/KhronosGroup/SPIRV-LLVM-Translator/blob/main/lib/SPIRV/SPIRVUtil.cpp#L2475-L2672.

vmaksimo: Because this "builtin" is translated via `transBuiltinFromInst()`, which follows OCLUtil mangling by default.
...and I just verified the fix should appear both in OCLUtil and SPIRVUtil if we want to work it for all the workflows - thanks for catching this!

vmaksimo:
Tagging @Qi-Ye-079 for review

YuriPlyakhin:
> Tagging @Qi-Ye-079 for review

@vmaksimo , can we add to @Qi-Ye-079 enough permissions to be able to assign him as reviewer? What is the process for that?

YuriPlyakhin (approved):
LGTM

svenvh (approved):

PR #3964, opened 2026-08-11
Translate {uinc/udec}_wrap as opaque intrinsics for AMD

maarquitos14:
Translation of `uinc_wrap` and `udec_wrap` as if they were opaque llvm intrinsics, so that they can be restored in the reverse translation to the exact original code. This only applies to AMD targets.

Part 1 of #3945 split.


lib/SPIRV/SPIRVRegularizeLLVM.cpp
```diff
@@ -793,6 +793,66 @@ bool SPIRVRegularizeLLVMBase::regularize() {
           Cmpxchg->replaceAllUsesWith(V2);
           ToErase.push_back(Cmpxchg);
         }
+        if (auto *ARMW = dyn_cast<AtomicRMWInst>(&II)) {
```
MrSidims, line 814: This _must_ be guarded by a new translator option. I believe our friends from ARM and Intel (and other contributors to the translator) wouldn't like to get some random link-time error (that depends on driver's version) of the unresolved symbol one day instead of frontend translator-side unresolved instruction/op error.

MrSidims (commented):
@svenvh @vmaksimo @sarnex not sure if your review is needed here, just FYI that we are doing something shady here, well, you know the drill :)

lib/SPIRV/SPIRVRegularizeLLVM.cpp
maarquitos14: Isn't the guard in line 803 enough?

MrSidims: You mean "M->getTargetTriple().getVendor() == Triple::AMD" (since the code has moved)? Makes sense. Yet some static analyzer tool will complain about `AOp` being trivially dead and unused if `M->getTargetTriple().getVendor() == Triple::AMD` is `false`. So may be worth some reordering regardless.

lib/SPIRV/SPIRVRegularizeLLVM.cpp
```diff
@@ -793,6 +816,69 @@ bool SPIRVRegularizeLLVMBase::regularize() {
...
+          AtomicRMWInst::BinOp AOp = ARMW->getOperation();
+          // Carrying these across the SPIR-V boundary as a call to an imported
```
MrSidims: This comment and the comment below are a bit way too verbose. May be make sense to shorten them, saying (in the top comment), that: "for AMD triple we translate AtomicRMWInst::UIncWrap and AtomicRMWInst::UDecWrap to __translate_spirv_atomic_uinc_wrap external function call" and in the following comment explain how mangling works? Just to make agents consume less tokens while parsing this.

lib/SPIRV/SPIRVRegularizeLLVM.cpp
```diff
@@ -793,6 +816,69 @@ bool SPIRVRegularizeLLVMBase::regularize() {
...
+            if (auto *Callee = dyn_cast<Function>(FC.getCallee()))
+              Callee->setCallingConv(CallingConv::SPIR_FUNC);
+
+            IRBuilder<> Builder(ARMW);
+            CallInst *Call =
+                Builder.CreateCall(FC, {Ptr, MemoryScope, Sem, Val});
+            Call->setCallingConv(CallingConv::SPIR_FUNC);
+            Call->takeName(ARMW);
+
+            ARMW->replaceAllUsesWith(Call);
```
MrSidims, line 866: Shall we also copy metadata/debug info?

lib/SPIRV/SPIRVRegularizeLLVM.cpp
```diff
@@ -793,6 +816,69 @@ bool SPIRVRegularizeLLVMBase::regularize() {
...
+            FuncName +=
+                "_p" +
+                std::to_string(Ptr->getType()->getPointerAddressSpace()) + "_" +
+                getAtomicWrapTypeSuffix(MemType);
+
+            Type *Int32Ty = Type::getInt32Ty(M->getContext());
+            FunctionType *FT = FunctionType::get(
+                MemType, {Ptr->getType(), Int32Ty, Int32Ty, MemType}, false);
+            FunctionCallee FC = M->getOrInsertFunction(FuncName, FT);
+            if (auto *Callee = dyn_cast<Function>(FC.getCallee()))
```
MrSidims: Can it ever be nullptr?

lib/SPIRV/SPIRVToOCL.cpp
```diff
@@ -1195,6 +1202,80 @@ SPIRVToOCLBase::getOCLPipeOpaqueType(SmallVector<std::string, 8> &Postfixes) {
...
+
+  Value *Ptr = CI->getArgOperand(0);
+  auto Scope = cast<ConstantInt>(CI->getArgOperand(1))->getZExtValue();
+  auto MemSem = cast<ConstantInt>(CI->getArgOperand(2))->getZExtValue();
+  Value *Val = CI->getArgOperand(3);
+
+  AtomicOrdering Ordering = mapSPIRVMemSemToAtomicOrdering(MemSem);
+  SyncScope::ID SSID = M->getTargetTriple().isAMDGCN()
+                           ? mapSPIRVScopeToAMDGPU(CI->getContext(), Scope)
+                           : mapSPIRVScopeToLLVM(CI->getContext(), Scope);
```
MrSidims: Do we expect it be used outside of AMD case?

test/atomicrmw_uinc_udec_wrap.ll
```diff
@@ -0,0 +1,157 @@
+; SPIR-V has no core atomic opcode for LLVM's uinc_wrap/udec_wrap, so they are
+; translated to a call to an imported __translate_spirv_atomic_uinc_wrap /
+; __translate_spirv_atomic_udec_wrap function, and the reverse translation maps
+; that call back to the original atomicrmw.
+;
+; The helper name carries a _p<addrspace>_i<width> suffix, because a module may
+; need several mutually incompatible signatures while SPIR-V resolves an
+; imported function by its linkage name alone.
+;
+; The helper is an AMD extension, so this requires an AMD target triple; other
```
MrSidims: It's not an extension, it's a ~workaround~ internal llvm-spirv builtin.

test/atomicrmw_uinc_udec_wrap.ll
```diff
@@ -0,0 +1,157 @@
...
+
+; RUN: llvm-spirv %s -o %t.spv
+; RUN: llvm-spirv -to-text %t.spv -o - | FileCheck %s --check-prefix=CHECK-SPIRV
+; RUN: llvm-spirv -r %t.spv -o %t.rev.bc
+; RUN: llvm-dis %t.rev.bc -o - | FileCheck %s --check-prefix=CHECK-LLVM
+; RUN: spirv-val %t.spv
+
+; None of the helpers is defined in the module; all of them are imported. Each
+; distinct (address space, value type) combination gets its own declaration,
+; and every call site of a given combination shares that one declaration.
```
MrSidims: This is redundant.

test/atomicrmw_uinc_udec_wrap.ll
```diff
@@ -0,0 +1,157 @@
...
+; CHECK-SPIRV-DAG: Decorate [[UIncFn:[0-9]+]] LinkageAttributes "__translate_spirv_atomic_uinc_wrap_p1_i32" Import
```
MrSidims: Please use [[#VarName:]] spelling

test/atomicrmw_uinc_udec_wrap.ll
```diff
@@ -0,0 +1,157 @@
...
+; CHECK-SPIRV: FunctionCall [[Int]] [[#]] [[UDecFn]] [[#]] [[Scope_Invocation]] [[MemSem_SequentiallyConsistent]] [[Value]]
+; CHECK-LLVM: atomicrmw udec_wrap ptr addrspace(1) @ui, i32 42 syncscope("singlethread") seq_cst
+define spir_func void @test_udec_wrap_singlethread_scope() {
+entry:
+  %r = atomicrmw udec_wrap ptr addrspace(1) @ui, i32 42 syncscope("singlethread") seq_cst
+  ret void
+}
+
+; Two atomics of the same value type in different address spaces need two
+; incompatible signatures, so they must resolve to two distinct declarations.
```
MrSidims: I believe test name `test_uinc_wrap_mixed_addrspace` is self-explanatory. Same for other comments.

test/atomicrmw_uinc_udec_wrap.ll
```diff
@@ -0,0 +1,157 @@
...
+; CHECK-SPIRV: FunctionCall [[Int]] [[#]] [[UDecFn]] [[#]] [[Scope_CrossDevice]] [[MemSem_SequentiallyConsistent]] [[Value]]
+; CHECK-LLVM: atomicrmw udec_wrap ptr addrspace(1) @ui, i32 42 seq_cst
+define spir_func void @test_udec_wrap_seq_cst() {
+entry:
+  %r = atomicrmw udec_wrap ptr addrspace(1) @ui, i32 42 seq_cst
+  ret void
+}
+
+; Sync scopes. The reverse translation picks the generic LLVM scope names for a
+; non-AMDGCN target triple.
```
MrSidims: Do we have AMDGCN targe triple test for reverse translation?

lib/SPIRV/SPIRVToOCL.cpp
MrSidims: Frankly, before this patch, the translator should be introduced with an ability to pick between target triples during reverse translation.

lib/SPIRV/SPIRVToOCL.cpp
```diff
@@ -1195,6 +1202,80 @@ SPIRVToOCLBase::getOCLPipeOpaqueType(SmallVector<std::string, 8> &Postfixes) {
   return PipeAccess ? kSPR2TypeName::PipeWO : kSPR2TypeName::PipeRO;
 }
 
+static AtomicOrdering mapSPIRVMemSemToAtomicOrdering(uint32_t MemSem) {
```
MrSidims: I believe this doesn't belong to SPIRVToOCL. It runs conditionally/optionally after the translation has already happened. Therefore for a potential case, where it's not run we end up with a module with a global symbol, which nobody will resolve.

I suggest to move the lowering to either Utils (and call it from the Reader) or to the Reader itself.

BTW, are these new maps needed? mapSPIRVMemSemToAtomicOrdering feels like mapSPIRVMemOrderToOCL. The rest maps also look familiar.

test/atomicrmw_uinc_udec_wrap_vector.ll
```diff
@@ -0,0 +1,59 @@
+; atomicrmw uinc_wrap/udec_wrap also accept a fixed vector of integers. Verify
+; that a vector operand is carried across as a call to the imported helper and
+; reconstituted by the reverse translation, exactly like a scalar one, under a
+; symbol of its own: the _p<addrspace>_<type> suffix spells the value type the
+; LLVM way, so <2 x i32> becomes v2i32. The element width is what varies
+; between these cases, so all three must resolve to distinct declarations. The
+; scalar cases are covered by atomicrmw_uinc_udec_wrap.ll, the over-limit ones
+; by atomicrmw_uinc_udec_wrap_oversized.ll.
```
MrSidims: lets save water by saving tokens!
```suggestion
; between these cases, so all three must resolve to distinct declarations. 
```

lib/SPIRV/SPIRVRegularizeLLVM.cpp
maarquitos14: It's added in #3965, I can bring it into this one if you like it better.

lib/SPIRV/SPIRVRegularizeLLVM.cpp
maarquitos14: Actually, it can if any other `GlobalValue` which is not a `Function` already holds the name we need. The name is reserved, and if this happens, our approach would break, so I'm adding a diagnostic to fail loudly if this occurs. Thanks for pointing this out!

lib/SPIRV/SPIRVToOCL.cpp
maarquitos14: https://github.com/KhronosGroup/SPIRV-LLVM-Translator/pull/3991 addresses this.

lib/SPIRV/SPIRVRegularizeLLVM.cpp
maarquitos14: Addressed in https://github.com/KhronosGroup/SPIRV-LLVM-Translator/pull/3964/commits/4489157b5c6497748d41ced103ad93b6f6b60ed5.

lib/SPIRV/SPIRVRegularizeLLVM.cpp
maarquitos14: Addressed in https://github.com/KhronosGroup/SPIRV-LLVM-Translator/pull/3964/commits/4489157b5c6497748d41ced103ad93b6f6b60ed5.

lib/SPIRV/SPIRVRegularizeLLVM.cpp
maarquitos14: Addressed in https://github.com/KhronosGroup/SPIRV-LLVM-Translator/pull/3964/commits/4489157b5c6497748d41ced103ad93b6f6b60ed5.

lib/SPIRV/SPIRVToOCL.cpp
maarquitos14: Addressed in #3991 and cherry-picked into here.

test/atomicrmw_uinc_udec_wrap.ll
maarquitos14: Agreed, I tried to rephrase all the occurrences of "extension" in https://github.com/KhronosGroup/SPIRV-LLVM-Translator/pull/3964/commits/4489157b5c6497748d41ced103ad93b6f6b60ed5.

test/atomicrmw_uinc_udec_wrap.ll
maarquitos14: Addressed in https://github.com/KhronosGroup/SPIRV-LLVM-Translator/pull/3964/commits/4489157b5c6497748d41ced103ad93b6f6b60ed5. Let me know if it's better or you would rather completely drop the comment.

test/atomicrmw_uinc_udec_wrap.ll
maarquitos14: Addressed in https://github.com/KhronosGroup/SPIRV-LLVM-Translator/pull/3964/commits/4489157b5c6497748d41ced103ad93b6f6b60ed5.

test/atomicrmw_uinc_udec_wrap.ll
maarquitos14: Addressed in https://github.com/KhronosGroup/SPIRV-LLVM-Translator/pull/3964/commits/4489157b5c6497748d41ced103ad93b6f6b60ed5.

test/atomicrmw_uinc_udec_wrap.ll
maarquitos14: Added in https://github.com/KhronosGroup/SPIRV-LLVM-Translator/pull/3964/commits/701d077c5ce59585f2d04288ab6042c601734370.

lib/SPIRV/SPIRVToOCL.cpp
maarquitos14: Agreed, addressed in https://github.com/KhronosGroup/SPIRV-LLVM-Translator/pull/3964/changes/72a43354a4b139ca6ea04b4b00b508df1250bfa7.

test/atomicrmw_uinc_udec_wrap_vector.ll
maarquitos14: Addressed in https://github.com/KhronosGroup/SPIRV-LLVM-Translator/commit/4489157b5c6497748d41ced103ad93b6f6b60ed5.

lib/SPIRV/SPIRVRegularizeLLVM.cpp
maarquitos14: Should be better now.

lib/SPIRV/SPIRVRegularizeLLVM.cpp
```diff
@@ -793,6 +811,62 @@ bool SPIRVRegularizeLLVMBase::regularize() {
...
+                                 "' is already used by another symbol");
+            Callee->setCallingConv(CallingConv::SPIR_FUNC);
+
+            IRBuilder<> Builder(ARMW);
+            CallInst *Call =
+                Builder.CreateCall(FC, {Ptr, MemoryScope, Sem, Val});
+            Call->setCallingConv(CallingConv::SPIR_FUNC);
+            Call->takeName(ARMW);
+            Call->copyMetadata(*ARMW);
+
```
MrSidims, line 866: Should we check ARMW->isVolatile() and ARMW->isElementwise() and preserve them?

lib/SPIRV/SPIRVReader.cpp
```diff
@@ -4241,6 +4241,9 @@ bool SPIRVToLLVM::translate() {
   transGeneratorMD();
   if (!lowerBuiltins(BM, M))
     return false;
+  // Unconditional; emptied helper decls are cleaned up by
+  // eraseUselessFunctions.
```
MrSidims: This is unconditional, but reverse is not. IMO they should be aligned in a defensive way - reject lowering if we didn't ask for it.

lib/SPIRV/SPIRVRegularizeLLVM.cpp
```diff
@@ -793,6 +811,62 @@ bool SPIRVRegularizeLLVMBase::regularize() {
...
+        if (auto *ARMW = dyn_cast<AtomicRMWInst>(&II)) {
+          // For an AMD triple, uinc_wrap/udec_wrap become a call to an imported
+          // __translate_spirv_atomic_u{inc,dec}_wrap helper.
+          if (M->getTargetTriple().getVendor() == Triple::AMD &&
+              (ARMW->getOperation() == AtomicRMWInst::UIncWrap ||
+               ARMW->getOperation() == AtomicRMWInst::UDecWrap) &&
+              isAtomicWrapSizeSupported(M, ARMW->getValOperand()->getType())) {
+            AtomicRMWInst::BinOp AOp = ARMW->getOperation();
+            // The name carries a _p<addrspace>_<type> suffix (e.g. _p1_i32,
+            // _p1_v2i32) since SPIR-V resolves an import by linkage name alone.
```
MrSidims, line 822: Does it make sense to carry the address spaces? Note: they are triple dependent. So suffix during forward will be having different meaning from the suffix on reverse. Either handle it or remove as we won't expect any driver to handle these "builtins" outside of llvm-spirv -r flow.

lib/SPIRV/SPIRVRegularizeLLVM.cpp
```diff
@@ -793,6 +811,62 @@ bool SPIRVRegularizeLLVMBase::regularize() {
...
+            FunctionType *FT = FunctionType::get(
+                MemType, {Ptr->getType(), Int32Ty, Int32Ty, MemType}, false);
+            FunctionCallee FC = M->getOrInsertFunction(FuncName, FT);
+            // The name is reserved, so it must resolve to our own function; a
+            // clash with any other symbol means we cannot lower correctly.
+            auto *Callee = dyn_cast<Function>(FC.getCallee());
+            if (!Callee)
+              report_fatal_error(Twine("Reserved atomic wrap helper name '") +
+                                 FuncName +
+                                 "' is already used by another symbol");
```
MrSidims: Lets not add new `report_fatal_error`.

test/atomicrmw_uinc_udec_wrap.ll
```diff
@@ -0,0 +1,139 @@
...
+
+target datalayout = "e-i64:64-v16:16-v24:32-v32:32-v48:64-v96:128-v192:256-v256:256-v512:512-v1024:1024"
+target triple = "spirv64-amd-amdhsa"
+
+@ui = common dso_local addrspace(1) global i32 0, align 4
+@lui = common dso_local addrspace(3) global i32 0, align 4
+@ul = common dso_local addrspace(1) global i64 0, align 8
+
+; CHECK-SPIRV: FunctionCall [[#Int]] [[#]] [[#UIncFn]] [[#]] [[#Scope_CrossDevice]] [[#MemSem_Relaxed]] [[#Value]]
+; CHECK-LLVM: atomicrmw uinc_wrap ptr addrspace(1) @ui, i32 42 monotonic
```
MrSidims: For all tests: add users of atomicrmw and check them

test/atomicrmw_uinc_udec_wrap.ll
```diff
@@ -0,0 +1,139 @@
+; uinc_wrap/udec_wrap have no SPIR-V opcode; on an AMD triple they are carried
```
MrSidims: for all tests: move these tests to transcoding.

for amd triple - add amd suffix to them.

test/atomicrmw_uinc_udec_wrap.ll
```diff
@@ -0,0 +1,139 @@
...
+; across as a call to an imported __translate_spirv_atomic_u{inc,dec}_wrap
+; helper and restored to the original atomicrmw on reverse translation. The name
+; carries a _p<addrspace>_i<width> suffix since SPIR-V resolves an import by
+; linkage name alone. This is an AMD internal spirv builtin; other vendors
+; report the atomicrmw as unsupported (atomicrmw_uinc_udec_wrap_non_amd.ll).
```
MrSidims: If error is not checked in the test - remove the lines

test/atomicrmw_uinc_udec_wrap.ll
```diff
@@ -0,0 +1,139 @@
+; uinc_wrap/udec_wrap have no SPIR-V opcode; on an AMD triple they are carried
+; across as a call to an imported __translate_spirv_atomic_u{inc,dec}_wrap
```
MrSidims: for all tests: please double check comments - some things here are meaningless; humanize comments

test/reverse-target-triple-amdgcn-addrspace.ll
```diff
@@ -0,0 +1,83 @@
...
+; RUN: llvm-spirv -r %t.spv --spirv-target-triple=nvptx64-nvidia-cuda \
+; RUN:   -o - | llvm-dis | FileCheck %s --check-prefix=CHECK-DEFAULT
+
+; Interaction: an explicit --spirv-addrspace-map wins over the triple-derived
+; map. 0:5 keeps the private alloca in the AMDGPU-required addrspace(5) (so the
+; module still verifies), while leaving the other classes at their identity
+; numbering.
+; RUN: llvm-spirv -r %t.spv --spirv-target-triple=amdgcn-amd-amdhsa \
+; RUN:   --spirv-addrspace-map=0:5 \
+; RUN:   -o - | llvm-dis | FileCheck %s --check-prefix=CHECK-OVERRIDE
```
MrSidims: Please remove triple-related patch from the PR and use stacked PR feature.

maarquitos14: As far as I know, I cannot stack on #3994 here because the branch is on my fork, and I can only use branches that exist in this very own repo. Alternatively, I can rebase, but that would mean a force push afterwards. Please, let me know if you would like me to do that or if I'm wrong in my understanding of stacked PRs.

MrSidims: Assuming questions/comments are answered - I don't mind a force push as a lesser evil.

lib/SPIRV/SPIRVRegularizeLLVM.cpp
maarquitos14: - *volatile*: The `Volatile` decoration must be applied only to memory object declarations or members of a structure type, so that means we cannot use it for an `OpFunctionCall`. Another option I considered is MemorySemantics `Volatile`, but it needs `VulkanMemoryModel`, so that discards it. The other `Volatile` option I see in the spec is in `Memory Operands`, but an `OpFunctionCall` doesn't have such operand. At this point, I see two possible options for this:

1. Add `_volatile` suffix to the function name: makes the function name even longer and could double the amount of functions declared.
2. Add an extra argument to the function to pass volatility: seems the way to go, IMO. The one thing I don't love about this approach is that it's not immediately obvious that volatility is preserved this way when you look at it.

- *ElementWise*: if I checked correctly, for the non-AMD path the translator produces exactly the same code regardless of elementwise. That's just an FYI, in case we want to follow up. For our case, I'm following the same approach as above, and adding a new argument for it.

lib/SPIRV/SPIRVReader.cpp
maarquitos14: I'm happy to add it, but if the module contains any of these, and they're not targeting AMD, they will get a linker error about import functions that nobody provides. I don't if this is the best error to find.

lib/SPIRV/SPIRVRegularizeLLVM.cpp
maarquitos14: The address spaces were originally added to avoid function name collision. If the module contains two different `uinc_wrap/udec_wrap` with the same type but pointers of different address spaces, they were colliding because function overload is not supported. Adding the address space prevents the collision. It is not really used for actually reconstructing the original atomic op in the reverse translation, so there is nothing to _handle_, and if we drop it, then we go back into name collision. I do see that it might be confusing because it's directly spelled in the module and doesn't always match the reverse target, so if need be, we can get the storage class names in the function name instead of the bare address spaces.

lib/SPIRV/SPIRVRegularizeLLVM.cpp
maarquitos14: Should we go with simple `cast<>` instead? That would assert cleanly in debug mode, and on release we'd see something like:
```
InvalidInstruction: Can't translate llvm instruction:
   Global variable cannot have Function storage class. Consider setting a proper address space.
   Original LLVM value:
  @__translate_spirv_atomic_uinc_wrap_p1_i32 = global i32 0
```

lib/SPIRV/SPIRVRegularizeLLVM.cpp
maarquitos14: Done in https://github.com/KhronosGroup/SPIRV-LLVM-Translator/pull/3964/commits/354aeb9e3f88656d180df1ee4bc0b601896faab8. I chose #2 for volatile.

lib/SPIRV/SPIRVReader.cpp
maarquitos14: Added in https://github.com/KhronosGroup/SPIRV-LLVM-Translator/pull/3964/commits/354aeb9e3f88656d180df1ee4bc0b601896faab8.

lib/SPIRV/SPIRVRegularizeLLVM.cpp
maarquitos14: Done in https://github.com/KhronosGroup/SPIRV-LLVM-Translator/pull/3964/commits/354aeb9e3f88656d180df1ee4bc0b601896faab8.

test/atomicrmw_uinc_udec_wrap.ll
maarquitos14: Done in https://github.com/KhronosGroup/SPIRV-LLVM-Translator/pull/3964/commits/354aeb9e3f88656d180df1ee4bc0b601896faab8.

test/atomicrmw_uinc_udec_wrap.ll
maarquitos14: Done in https://github.com/KhronosGroup/SPIRV-LLVM-Translator/pull/3964/commits/354aeb9e3f88656d180df1ee4bc0b601896faab8.

test/atomicrmw_uinc_udec_wrap.ll
maarquitos14: Done in https://github.com/KhronosGroup/SPIRV-LLVM-Translator/pull/3964/commits/354aeb9e3f88656d180df1ee4bc0b601896faab8.

test/atomicrmw_uinc_udec_wrap.ll
maarquitos14: Done in https://github.com/KhronosGroup/SPIRV-LLVM-Translator/pull/3964/commits/354aeb9e3f88656d180df1ee4bc0b601896faab8.

test/reverse-target-triple-amdgcn-addrspace.ll
maarquitos14: Force-push made.

lib/SPIRV/SPIRVRegularizeLLVM.cpp
```diff
@@ -793,6 +810,64 @@ bool SPIRVRegularizeLLVMBase::regularize() {
...
+              isAtomicWrapSizeSupported(M, ARMW->getValOperand()->getType())) {
+            AtomicRMWInst::BinOp AOp = ARMW->getOperation();
+            // The name carries a _p<addrspace>_<type> suffix (e.g. _p1_i32,
+            // _p1_v2i32) since SPIR-V resolves an import by linkage name alone.
+            Value *Ptr = ARMW->getPointerOperand();
+            Value *Val = ARMW->getValOperand();
+            Type *MemType = Val->getType();
+
+            spv::Scope S =
+                toSPIRVScope(ARMW->getContext(), ARMW->getSyncScopeID());
```
MrSidims, line 828: Won't this silently default to ScopeCrossDevice for AMD-specific scopes?

lib/SPIRV/SPIRVRegularizeLLVM.cpp
```diff
@@ -610,6 +610,23 @@ void prepareCacheControlsTranslation(Metadata *MD, Instruction *Inst) {
 }
 } // namespace
 
+/// Spell an integer or vector-of-integer type the LLVM way for the helper name
+/// suffix: i32, v2i32.
+static std::string getAtomicWrapTypeSuffix(Type *Ty) {
```
MrSidims, line 615: Shall we use mangleBuiltin machinery?

maarquitos14: We could, but I'd rather not, for several reasons:

1. Other helpers are emitted unmangled (e.g. __translate_spirv_memory_scope). It's fair to argue they don't need any type or addresspace suffix, so it's not exactly the same thing, but it's a precedent.
2. I don't think we really need it. We only do this to avoid name collision, and for that, with type and address space is enough. Mangling would apply to all the function args too, which we don't need. 
3. Unmangled is more readable. Looking at the unmangled you immediately can understand what the LLVM IR counterpart is.

lib/SPIRV/SPIRVRegularizeLLVM.cpp
maarquitos14: Yes, but according to the comment in `toSPIRVScope` that seems to be by design, and this function has other callers. I'm happy to fix that, but I think we should do it in a separate PR. However, if you agree, I'm going to extend the test checking scopes to include all the known scopes for AMD and SPIR triples, at least so that we are aware which ones can be round-tripped to the original value and those that can't. I think we should cover the following scopes: `system`, `singlethread`, `workgroup`, `device`, `subgroup`, `work_item`, `all_svm_devices`, `agent`, `wavefront`, and `cluster`. Let me know if I forgot any. Apart from that, given that it can be any string, I think for unrecognized names it's fair to fall back to the most conservative option: `ScopeCrossDevice`.

MrSidims: >  I think we should do it in a separate PR

Sure, but please fix it ASAP - I expect samples you are fixing would still run slow enough to be considered as hang on CI.

> I'm going to extend the test checking scopes to include all the known scopes for AMD and SPIR triples

Yes please.

MrSidims: Though, you are already adding `mapSPIRVScopeToLLVM` . So may be still worth to do in this PR?

MrSidims (approved):
lgtm

lib/SPIRV/SPIRVRegularizeLLVM.cpp
maarquitos14: Okay, that's a good point. I'll address here.

maarquitos14: Done in https://github.com/KhronosGroup/SPIRV-LLVM-Translator/pull/3964/commits/df27cc7d17375ca3742bb479ed07d2f3cb383bb4.

MrSidims: Decided to move into a dedicated PR

PR #3966, opened 2026-08-13
Emit SPV_EXT_bfloat16_arithmetic instead of SPV_INTEL_bfloat16_arithmetic

vmaksimo:
BFloat16ArithmeticEXT is the multi-vendor successor to BFloat16ArithmeticINTEL, reusing the same capability enumerant (6226).
We replace the usage of INTEL extension with the published EXT one, but continue to accept modules produced with INTEL names for backward compatibility.

Spec: https://github.com/KhronosGroup/SPIRV-Registry/pull/444

TODO: replace the internal values with the published once the Registry and Headers PRs are merged.


vmaksimo:
FYI @Qi-Ye-079

test/extensions/INTEL/SPV_INTEL_bfloat16_arithmetic/reader_compat.spt
```diff
@@ -0,0 +1,36 @@
...
+; capability names.
+
+; CHECK: %r = fadd bfloat %a, %b
+
+119734787 65536 393230 9 0
+2 Capability Addresses
+2 Capability Linkage
+2 Capability Kernel
+2 Capability BFloat16TypeKHR
+2 Capability BFloat16ArithmeticINTEL
```
bashbaug: I think we could switch this to...

```suggestion
2 Capability BFloat16ArithmeticEXT
```

... and then remove the alias.  I don't expect to see older SPIR-V modules using the text form of the capability, so this shouldn't be needed for backward compatibility.  The only reason we need the name of the extension for backward compatibility is because it gets encoded into the binary SPIR-V module via OpExtension.

lib/SPIRV/libSPIRV/SPIRVNameMapEnum.h
```diff
@@ -698,7 +698,8 @@ template <> inline void SPIRVMap<Capability, std::string>::init() {
   add(CapabilityInt4CooperativeMatrixINTEL, "Int4CooperativeMatrixINTEL");
   add(CapabilityFunctionVariantsINTEL, "FunctionVariantsINTEL");
   add(CapabilitySpecConditionalINTEL, "SpecConditionalINTEL");
-  add(internal::CapabilityBFloat16ArithmeticINTEL, "BFloat16ArithmeticINTEL");
+  add(internal::CapabilityBFloat16ArithmeticEXT, "BFloat16ArithmeticINTEL");
+  add(internal::CapabilityBFloat16ArithmeticEXT, "BFloat16ArithmeticEXT");
```
bashbaug: This is the alias I was referring to.  If we decide that we don't care about backward compatibility for SPIR-V text modules, we can remove it and just use:

```suggestion
  add(internal::CapabilityBFloat16ArithmeticEXT, "BFloat16ArithmeticEXT");
```

test/extensions/INTEL/SPV_INTEL_bfloat16_arithmetic/reader_compat.spt
vmaksimo: okay, agree!

bashbaug (approved):

lib/SPIRV/libSPIRV/spirv_internal.hpp
```diff
@@ -108,7 +108,7 @@ enum InternalCapability {
   ICapabilityFloat4E2M1CooperativeMatrixINTEL = 6213,
   ICapabilityFloatConversionsFtoFINTEL = 6215,
   ICapabilityFloatConversionsFtoSINTEL = 6216,
-  ICapabilityBFloat16ArithmeticINTEL = 6226,
+  ICapabilityBFloat16ArithmeticEXT = 6226,
```
MrSidims, line 111: Lets just update SPIR-V Header tag instead.

Hm, unless there is no SPIR-V Headers patch.

MrSidims (approved):

lib/SPIRV/libSPIRV/spirv_internal.hpp
bashbaug: I'm all for this, but it requires merging this PR first: https://github.com/KhronosGroup/SPIRV-Headers/pull/604

Are we ready to merge it? I'm not anticipating any changes to the extension, personally.

PR #3970, opened 2026-08-13
[CI] Use llvm-toolchain-plucky instead of noble for out-of-tree nightly

vmaksimo:
The noble (Ubuntu 24.04) snapshot on apt.llvm.org has been stuck at a commit from 2026-08-10 since 2026-08-11, causing the out-of-tree nightly to fail with CASPluginTest export errors even though the underlying issue was fixed upstream in llvm/llvm-project#215292

Switch to plucky (Ubuntu 25.04) packages, which are publishing fresh snapshots daily and include the fix.

Note: This installs 25.04 packages on the 24.04 runner, which may cause dependency issues if the ABI has changed. 
If this fails, we can either revert to noble once the snapshot resumes or switch the runner to 22.04 and use llvm-toolchain-jammy instead.

W/A for https://github.com/llvm/llvm-project/issues/216102

### DIFF
--- a/.github/workflows/check-out-of-tree-build.yml
+++ b/.github/workflows/check-out-of-tree-build.yml
@@ -56,7 +56,7 @@ jobs:
       matrix:
         build_type: [Release, Debug]
       fail-fast: false
-    runs-on: ubuntu-24.04
+    runs-on: ubuntu-22.04
     steps:
       - name: Install dependencies
         run: |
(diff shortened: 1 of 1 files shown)


vmaksimo:
@svenvh what's your call here? 
should we wait for the packages update, or maybe we can upgrade to 26.04 (though it's only in preview state right now)?

svenvh:
> @svenvh what's your call here? should we wait for the packages update, or maybe we can upgrade to 26.04 (though it's only in preview state right now)?

I think for now we will have wait until the Ubuntu 24 apt.llvm.org packages are updated again, and accept that OOB builds will be broken until then.

Longer term, we should perhaps look into doing an out-of-tree build ourselves with caching enabled, to reduce our dependence on the apt packages.  We could follow the approach we have for spirv-tools, as that seems to work quite well?

vmaksimo:
The snapshot builder picked back up - OOB is back to green. Closing

PR #3976, opened 2026-08-17
Replace removed GlobalVariable::getAlignment() with getAlign()

lamb-j:
llvm/llvm-project@b71ce8f has removed `GlobalVariable::getAlignment()` (long marked *"FIXME: Remove this function once transition to Align is over"*), which breaks the build of `transAlign()` in `SPIRVWriter.cpp`.

This replicates the old behavior via the stable `getAlign()` API, returning 0 when the global has no explicit alignment — matching the removed helper exactly. `getAlign()`/`MaybeAlign` has existed for years, so it compiles against both older and newer LLVM.


vmaksimo (approved):

vmaksimo:
Windows regression was reverted in llvm/llvm-project@d37fa5a

PR #3977, opened 2026-08-18
Lower `llvm.umul.with.overflow` via `OpUMulExtended`

vmaksimo:
The previous emulation could divide by zero and checked the wrong operand, so it sometimes reported incorrect overflow.
Use `OpUMulExtended`'s high word instead, the same way `uadd`/`usub` use `OpIAddCarry`/`OpISubBorrow`.

Also fixes the rebuilt result struct not matching the intrinsic's actual return type.

### DIFF
--- a/test/llvm-intrinsics/umul.with.overflow.ll
+++ b/test/llvm-intrinsics/umul.with.overflow.ll
@@ -21,9 +21,13 @@ target triple = "spir"
 define dso_local spir_func void @_Z4foo8hhPh(i8 zeroext %a, i8 zeroext %b, ptr captures(none) %c) local_unnamed_addr #0 {
 entry:
   ; CHECK-LLVM: call spir_func void @_Z20__spirv_UMulExtendedhh(ptr sret(%[[STRUCT8]]) %{{.*}}, i8 %a, i8 %b)
-  ; CHECK-SPIRV: UMulExtended [[#]] [[#]] [[#]] [[#]]
-  ; CHECK-SPIRV: CompositeExtract [[#]] [[#HI:]] [[#]] 1
-  ; CHECK-SPIRV: INotEqual [[#]] [[#]] [[#HI]] [[#]]
+  ; CHECK-SPIRV: Constant [[#]] [[#ZERO8:]] 0
+  ; CHECK-SPIRV: Function
+  ; CHECK-SPIRV: UMulExtended [[#]] [[#MUL:]] [[#]] [[#]]
+  ; CHECK-SPIRV: Store [[#PTR:]] [[#MUL]]
+  ; CHECK-SPIRV: Load [[#]] [[#LOADED:]] [[#PTR]]
+  ; CHECK-SPIRV: CompositeExtract [[#]] [[#HI:]] [[#LOADED]] 1
+  ; CHECK-SPIRV: INotEqual [[#]] [[#]] [[#HI]] [[#ZERO8]]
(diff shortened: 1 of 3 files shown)


test/llvm-intrinsics/umul.with.overflow.ll
```diff
@@ -10,18 +10,20 @@
...
 
 target datalayout = "e-p:32:32-i64:64-v16:16-v24:32-v32:32-v48:64-v96:128-v192:256-v256:256-v512:512-v1024:1024"
 target triple = "spir"
 
 ; Function Attrs: nofree nounwind writeonly
 define dso_local spir_func void @_Z4foo8hhPh(i8 zeroext %a, i8 zeroext %b, ptr captures(none) %c) local_unnamed_addr #0 {
 entry:
-  ; CHECK-LLVM: call { i8, i1 } @llvm.umul.with.overflow.i8
-  ; CHECK-SPIRV: FunctionCall [[#]] [[#]] [[NAME_UMUL_FUNC_8]]
+  ; CHECK-LLVM: call spir_func void @_Z20__spirv_UMulExtendedhh(ptr sret(%[[STRUCT8]]) %{{.*}}, i8 %a, i8 %b)
```
Qi-Ye-079, line 23: I think for `OpIAddCarry` and `OpISubBorrow`, the Reader will translate them back to original LLVM intrinsics (llvm.uadd/usub.with.overflow). Should we do the same and maintain the old behavior for `OpUMulExtended`? Or is it out of scope for this PR?

test/llvm-intrinsics/umul.with.overflow.ll
```diff
@@ -10,18 +10,20 @@
...
+  ; CHECK-SPIRV: UMulExtended [[#]] [[#]] [[#]] [[#]]
+  ; CHECK-SPIRV: CompositeExtract [[#]] [[#HI:]] [[#]] 1
+  ; CHECK-SPIRV: INotEqual [[#]] [[#]] [[#HI]] [[#]]
```
Qi-Ye-079: A nit: some operands aren't matched here; it should show more dataflow connection among these instructions. Maybe we should also check, that `CompositeExtract` takes the result generated by `UMulExtended`, and `INotEqual` is actually comparing against a zero-constant?

```suggestion
  ; CHECK-SPIRV: Constant [[#]] [[#ZERO:]] 0
  ; CHECK-SPIRV: UMulExtended [[#]] [[#MUL:]] [[#]] [[#]]
  ; CHECK-SPIRV: CompositeExtract [[#]] [[#HI:]] [[#MUL]] 1
  ; CHECK-SPIRV: INotEqual [[#]] [[#]] [[#HI]] [[#ZERO]]
```

test/llvm-intrinsics/umul.with.overflow.ll
vmaksimo: Though it's out of scope of this PR, I'd think we should not translate these back to llvm intrinsics.
`OpUMulExtended` contains "full" result, while llvm intrinsic supports only the indication of overflow. Semantically this works only LLVM->SPIR-V, but not back - we can't tell for sure if `OpUMulExtended` was translated from "reduced" llvm intrinsic, or from SPIR-V friendly IR call.

Qi-Ye-079 (approved):
LGTM

vmaksimo:
Re-trigger CI

svenvh (approved):

PR #3978, opened 2026-08-18
Use poison instead of null for DebugInfoNone declares

karolzwolak:
DebugInfoNone in DebugDeclare's Variable operand means the variable has no location. Translating it to a null pointer describes the variable as living at address 0 instead: it is not a kill location per DbgVariableRecord::isKillLocation, so consumers materialize it as a real constant.

### DIFF
--- a/test/DebugInfo/mem2reged_local_var.ll
+++ b/test/DebugInfo/mem2reged_local_var.ll
@@ -25,7 +25,7 @@
 ; CHECK-SPIRV: ExtInst [[#]] [[#LocalVar:]] [[#]] DebugLocalVariable
 ; CHECK-SPIRV: ExtInst [[#]] [[#]] [[#]] DebugDeclare [[#LocalVar]] [[#None]] [[#]]
 
-; CHECK-LLVM: #dbg_declare(ptr null, ![[#LocalVar:]], !DIExpression(DW_OP_constu, 4, DW_OP_swap, DW_OP_xderef), ![[#Loc:]])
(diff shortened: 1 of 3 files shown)


test/DebugInfo/mem2reged_local_var.ll
```diff
@@ -25,7 +25,7 @@
 ; CHECK-SPIRV: ExtInst [[#]] [[#LocalVar:]] [[#]] DebugLocalVariable
 ; CHECK-SPIRV: ExtInst [[#]] [[#]] [[#]] DebugDeclare [[#LocalVar]] [[#None]] [[#]]
 
-; CHECK-LLVM: #dbg_declare(ptr null, ![[#LocalVar:]], !DIExpression(DW_OP_constu, 4, DW_OP_swap, DW_OP_xderef), ![[#Loc:]])
+; CHECK-LLVM: #dbg_declare(ptr poison, ![[#LocalVar:]], !DIExpression(DW_OP_constu, 4, DW_OP_swap, DW_OP_xderef), ![[#Loc:]])
```
MrSidims, line 28: But we had

> #dbg_declare(**ptr null**, !8, !DIExpression(DW_OP_constu, 4, DW_OP_swap, DW_OP_xderef), !10)

on **[L41](https://github.com/KhronosGroup/SPIRV-LLVM-Translator/pull/3978/changes#diff-91f8f4e0baf2bb1b2c7326caff42cf8516577a463181d8dce67690f561277383R41)**

PR #3987, opened 2026-08-20
Translate llvm.llround, llvm.lrint and llvm.llrint intrinsics

pvelesko:
llvm.lround gained translation support in #2904 as an OpenCL.std round followed by OpConvertFToS. Its three siblings were left out, so a module containing any of them still stops with `InvalidFunctionCall: Unexpected llvm intrinsic`. This completes that set.

llround reuses the existing shape unchanged. The destination type is taken from the intrinsic result type, so the wider return type needs no special handling and the vector overloads fall out for free. lrint and llrint use the same shape with OpenCLLIB::Rint in place of OpenCLLIB::Round.

One caveat is worth stating plainly, because the two specifications word rounding differently. OpenCL rint is round to nearest even unconditionally, whereas LangRef defines llvm.lrint and llvm.llrint as returning what the libm functions return, and libm lrint follows the current rounding mode. They coincide here because the OpenCL extended instruction set fixes the mode for its math instructions regardless of the caller. From the Math extended instructions section of the OpenCL extended instruction set specification:

> The math instructions are not affected by the prevailing rounding mode in the calling environment, and always return the same value as they would if called with the round to nearest even rounding mode.

LLVM in turn assumes the default floating point environment unless the constrained intrinsics are used. This is the same reasoning already baked into the tree: in transIntrinsicInst, Intrinsic::rint and Intrinsic::roundeven both map to OpenCLLIB::Rint, as does Intrinsic::nearbyint, which is likewise a current rounding mode libm function.

All four intrinsics are also registered in isKnownIntrinsic. lround was missing there, which looks like an oversight in #2904. Under -spirv-allow-unknown-intrinsics, transFunctionDecl still emitted an imported OpFunction declaration for an intrinsic whose calls transIntrinsicInst had already translated inline, so the output carried a stray

    OpDecorate %5 LinkageAttributes "llvm.lround.i32.f32" Import

alongside the inline OpenCL.std round. Nothing in the suite exercised that mode, so test/llvm-intrinsics/lround.ll now translates a second time with -spirv-allow-unknown-intrinsics and asserts that no such decoration survives. The two new tests assert the same for the three added intrinsics. Keeping the translation case but removing the isKnownIntrinsic entries reproduces all ten spurious Import decorations across the three files, so the assertions are not vacuous.

Known follow up, deliberately not addressed here. The lround family does not call checkTypeForSPIRVExtendedInstLowering, so an unsupported floating point argument such as x86_fp80 reaches transType and trips the assert at lib/SPIRV/libSPIRV/SPIRVType.h:302 instead of producing a diagnostic. That is existing behaviour of the lround case and this change does not alter it. It is left alone because the obvious one line guard is not a correct fix. The shared math case block in checkTypeForSPIRVExtendedInstLowering begins by requiring the argument type to equal the result type, which is never true for these intrinsics, so adding them to its case list makes it return false for every call, and since it returns false without calling SPIRVCK there is no diagnostic either: the translator drops the call and exits 0 having emitted a function body with no rounding in it at all. A correct fix needs a new case in that function that validates the argument type rather than the result type, with its own tests, which belongs in a separate change. Worth noting too that when x86_fp80 appears in a function signature the same assert fires for llvm.rint.f80, which does have the guard, so the guard is not the whole story.

The branch was built and tested against a local LLVM 23.1.0-rc2, not the LLVM 24 trunk that upstream CI builds against, because no LLVM 24 was available on the machine used. CI is therefore the first real check on the intended toolchain. Locally the three rounding tests pass, and the full lit suite goes from 1088 passed with 24 failed at the base commit to 1090 passed with 24 failed on this branch, with an identical failure list. Those 24 are environmental and reproduce unchanged on the base commit, mostly a local spirv-val that rejects newer capability operands such as 5156 and 6265, plus two OpTypeInt bit width rejections.

### DIFF
--- a/lib/SPIRV/SPIRVWriter.cpp
+++ b/lib/SPIRV/SPIRVWriter.cpp
@@ -4175,9 +4175,13 @@ bool LLVMToSPIRVBase::isKnownIntrinsic(Intrinsic::ID Id) {
   case Intrinsic::floor:
   case Intrinsic::fma:
   case Intrinsic::frexp:
+  case Intrinsic::llrint:
+  case Intrinsic::llround:
   case Intrinsic::log:
   case Intrinsic::log10:
   case Intrinsic::log2:
+  case Intrinsic::lrint:
+  case Intrinsic::lround:
   case Intrinsic::maximumnum:
   case Intrinsic::maximum:
   case Intrinsic::maxnum:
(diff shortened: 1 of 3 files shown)


maarquitos14:
@pvelesko I see this is marked as draft, so I'm not sure if it's ready for review. Could you please confirm?

PR #3988, opened 2026-08-22
[NFC] Fix backend checks in spirv_param_decorations_quals.ll

lamb-j:
llvm/llvm-project#215790 removed the backend's Volatile parameter decoration
derived from `!kernel_arg_type_qual`, so the `llc` legs can no longer share
`CHECK-SPV-IR` with the translator legs, which still expect it.

Give them a `CHECK-BACKEND` prefix matching current output, keeping the
cross-compilation coverage from #3564. Fixes the nightly `main` failure.

### DIFF
--- a/test/spirv_param_decorations_quals.ll
+++ b/test/spirv_param_decorations_quals.ll
@@ -49,9 +49,6 @@ entry:
 ; CHECK-SPV-IR-DAG: ![[VolatileDecoId]] = !{i32 21}
 ; CHECK-SPV-IR-DAG: ![[KernelArgTypeQual]] = !{!"volatile restrict"}
 
-; The SPIR-V backend stopped deriving a Volatile decoration from
-; !kernel_arg_type_qual in llvm/llvm-project#215790, so its output carries only
-; the NoAlias decoration spelled out in !spirv.ParameterDecorations.
 ; CHECK-BACKEND: define spir_kernel void @k(ptr addrspace(1) noalias %a)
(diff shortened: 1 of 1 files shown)


test/spirv_param_decorations_quals.ll
```diff
@@ -49,6 +49,17 @@ entry:
 ; CHECK-SPV-IR-DAG: ![[VolatileDecoId]] = !{i32 21}
 ; CHECK-SPV-IR-DAG: ![[KernelArgTypeQual]] = !{!"volatile restrict"}
 
+; The SPIR-V backend stopped deriving a Volatile decoration from
+; !kernel_arg_type_qual in llvm/llvm-project#215790, so its output carries only
+; the NoAlias decoration spelled out in !spirv.ParameterDecorations.
```
lamb-j:
```suggestion
```

lamb-j:
@aobolensk

aobolensk (approved):
LGTM

lamb-j:
How do we handle the Out-of-tree builds for fixes like these? As they still carry the old LLVM.

Do fixes need to be backward compatible? Or do we need to wait for those to update LLVM? Or do we merge with them red?

svenvh (approved):

svenvh:
> How do we handle the Out-of-tree builds for fixes like these? As they still carry the old LLVM.
> 
> Do fixes need to be backward compatible? Or do we need to wait for those to update LLVM? Or do we merge with them red?

We tend to merge them with red jobs (as I've just done), otherwise development would be held up too much.

PR #3990, opened 2026-08-25
Translate the signed overflow intrinsics, memcpy.inline, ptrmask and objectsize instead of rejecting them

pvelesko:
Six intrinsics that have an exact expression in core SPIR-V are rejected by the writer, so ordinary hardened C++ (`__builtin_sub_overflow`, `__builtin_mul_overflow`) and `_FORTIFY_SOURCE` code turns into a translation failure with no source location:

```
InvalidFunctionCall: Unexpected llvm intrinsic:
 llvm.ssub.with.overflow.i32 [Src: lib/SPIRV/SPIRVWriter.cpp:5684 false ]
```

`llvm.sadd.with.overflow` and `llvm.ssub.with.overflow` now become `OpIAdd` / `OpISub` plus the sign comparison that detects the overflow. An addition overflows when both operands share a sign that the result does not have; a subtraction overflows when the operands have different signs and the result does not have the sign of the minuend. Both are `((X ^ A) & (Y ^ B)) < 0` over the appropriate operand pairs. The scalar `sadd` overloads still go through the hand written helper functions in `SPIRVLowerLLVMIntrinsic`, whose table is keyed on the exact intrinsic name; only the vector overloads, which that table does not name, reach the writer. Moving the scalar overloads onto the inline expansion as well would be a separate change.

`llvm.smul.with.overflow` becomes `OpSMulExtended`. From the SPIR-V specification, `specs/binary.adoc`, OpSMulExtended:

```
Member 0 of the result gets the low-order bits of the multiplication.

Member 1 of the result gets the high-order bits of the multiplication.
```

so the product is representable in the operand width exactly when the high half is the sign extension of the low half, which is `OpShiftRightArithmetic` of member 0 by `bits - 1` compared against member 1 with `OpINotEqual`. `OpSMulExtended` carries no capability requirement, and the reader already maps it to `__spirv_SMulExtended`, so the round trip works. All three predicates were checked exhaustively over every `i8` operand pair against `__builtin_{add,sub,mul}_overflow`, with zero mismatches.

`llvm.memcpy.inline` joins the `llvm.memcpy` path. LangRef says the intrinsics "copy a block of memory from the source location to the destination location and guarantees that no external functions are called", and `OpCopyMemorySized` is an instruction rather than a call, so the existing translation already satisfies that guarantee. The zero length case keeps dropping the intrinsic entirely, since `OpCopyMemorySized` with a Size of 0 is invalid SPIR-V.

`llvm.ptrmask` is expanded in `SPIRVRegularizeLLVM` into the sequence LangRef specifies, with `iPtrIdx` the index type size of the pointer:

```
%intptr = ptrtoint ptr %ptr to iPtrIdx ; this may truncate
%masked = and iPtrIdx %intptr, %mask
%diff = sub iPtrIdx %masked, %intptr
%result = getelementptr i8, ptr %ptr, iPtrIdx %diff
```

The existing GEP translation turns that into `OpConvertPtrToU` / `OpBitwiseAnd` / `OpISub` / `OpPtrAccessChain`, so the result keeps the storage class of the pointer operand. An `inttoptr` of the masked value would not preserve it. Doing this in the regularization pass rather than in the writer also means vectors of pointers and the untyped pointer mode need no special handling.

`llvm.objectsize` is folded to a constant. The static form goes to `lowerObjectSizeCall`, which recovers the real size for objects the module can see. For the dynamic form (fourth operand set) that function would splice an instruction sequence in front of the call, which is not possible while the enclosing function is already being translated, so it gets the conservative answer LangRef prescribes: "If the size cannot be determined, `llvm.objectsize` returns `i32/i64 -1 or 0` (depending on the `min` argument)."

Two commits, tests first. The six new tests under `test/llvm-intrinsics/` fail on the parent commit with the error above and pass after the fix; each one runs `llvm-spirv`, `spirv-dis`, `spirv-val` and an `llvm-spirv -r` round trip.

Caveat on local validation. This was built and tested against LLVM 23.1.0-rc2, not the LLVM 24 trunk that `main` targets, because no LLVM 24 install was available on the machine. The only version sensitive API used is `lowerObjectSizeCall(IntrinsicInst *, const DataLayout &, const TargetLibraryInfo *, bool)`, whose declaration in `llvm/Analysis/MemoryBuiltins.h` is unchanged on trunk, so CI should be the deciding check for the LLVM 24 configuration. In the LLVM 23 configuration the full suite goes from 1087 passed and 31 failed to 1093 passed and 25 failed, so exactly the six new tests flip and nothing else changes. The 25 that remain are pre-existing on that box and unrelated: a SPIRV-Tools v2025.2 that predates the capabilities and integer widths those tests use, plus backend comparisons against LLVM 23. Nothing was executed on a device.

### DIFF
--- a/lib/SPIRV/SPIRVWriter.cpp
+++ b/lib/SPIRV/SPIRVWriter.cpp
@@ -70,6 +70,7 @@
 #include "llvm/ADT/StringSwitch.h"
 #include "llvm/Analysis/LoopAnalysisManager.h"
 #include "llvm/Analysis/LoopInfo.h"
+#include "llvm/Analysis/MemoryBuiltins.h"
 #include "llvm/Analysis/ValueTracking.h"
 #include "llvm/IR/Constants.h"
 #include "llvm/IR/DerivedTypes.h"
@@ -4225,6 +4226,7 @@ bool LLVMToSPIRVBase::isKnownIntrinsic(Intrinsic::ID Id) {
   case Intrinsic::fmuladd:
   case Intrinsic::memset:
   case Intrinsic::memcpy:
+  case Intrinsic::memcpy_inline:
   case Intrinsic::lifetime_start:
   case Intrinsic::lifetime_end:
   case Intrinsic::dbg_declare:
@@ -4240,6 +4242,10 @@ bool LLVMToSPIRVBase::isKnownIntrinsic(Intrinsic::ID Id) {
   case Intrinsic::debugtrap:
   case Intrinsic::uadd_with_overflow:
   case Intrinsic::usub_with_overflow:
+  case Intrinsic::sadd_with_overflow:
+  case Intrinsic::ssub_with_overflow:
+  case Intrinsic::smul_with_overflow:
+  case Intrinsic::objectsize:
   case Intrinsic::arithmetic_fence:
   case Intrinsic::masked_gather:
   case Intrinsic::masked_scatter:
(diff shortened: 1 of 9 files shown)


lib/SPIRV/SPIRVWriter.cpp
```diff
@@ -5025,6 +5031,78 @@ SPIRVValue *LLVMToSPIRVBase::transIntrinsicInst(IntrinsicInst *II,
...
+                                         {Low->getId(), Overflow->getId()}, BB);
+  }
+  case Intrinsic::objectsize: {
+    // llvm.objectsize never reads memory: it must fold to a constant. The
+    // dynamic form (fourth operand set) lets lowerObjectSizeCall() splice an
+    // instruction sequence in front of the call, which is not possible while
+    // the enclosing function is already being translated, so only the static
+    // form is handed to it. For the dynamic form fall back to the answer
+    // LangRef prescribes for an unknown object: -1 when min is false, 0 when
+    // min is true.
```
MrSidims, line 5094: General comment: LLMs tend to overexplain and reference things that are understandable only in the context of its session. For example, all of `lowerObjectSizeCall` sentence only makes sense when knowing, that somewhere in the context of the session (not necessary the original prompt) an idea of porting Analysis/MemoryBuiltins has appeared.

I'm not sure if such comments are useful for a human or LLM reader, so I suggest to shorten these comments.

test/llvm-intrinsics/memcpy.inline.ll
```diff
@@ -0,0 +1,48 @@
...
+; RUN: llvm-dis -o - %t.rev.bc | FileCheck --check-prefix CHECK-LLVM %s
+
+; llvm.memcpy.inline differs from llvm.memcpy only in guaranteeing that the
+; copy is not turned into a call to an external function. OpCopyMemorySized is
+; an instruction rather than a call, so the plain memcpy translation already
+; satisfies that guarantee.
+
+target triple = "spir64-unknown-unknown"
+
+; CHECK-SPIRV-DAG:  [[ulong:%[a-z0-9_]+]] = OpTypeInt 64 0
```
MrSidims, line 15: General comment: lets use [[#VAR_NAME:]] style of regex number-like variables for new tests.

MrSidims (commented):
Please split PR per intrinsic with an exception of *_with_overflow family - those can be submitted together.

PR #3991, opened 2026-08-27
Add option to pick reverse translation target triple

maarquitos14:

PR #3994, opened 2026-09-01
Derive addrspace map out of explicit target triple

maarquitos14:
Stacked on top of #3991.

When an explicit target triple is specified through `--spirv-target-triple`, we need to make sure that the addrspace map is aligned with the specified triple. This PR only adds the addrspace map for AMDGCN triples, but new ones can be easily added if needed. If there is no triple specified, or the triple is unknown, we fallback to the default addrspace map.

### DIFF
--- a/lib/SPIRV/LLVMSPIRVOpts.cpp
+++ b/lib/SPIRV/LLVMSPIRVOpts.cpp
@@ -74,7 +74,7 @@ constexpr TargetAddrSpaceMapping BuiltinAddrSpaceMaps[] = {
          AMDGPUAS::BUFFER_FAT_POINTER, // SPIRAS_Input
          AMDGPUAS::BUFFER_RESOURCE,    // SPIRAS_Output
          // TODO: FLAT clashes with BUFFER_STRIDED_POINTER; may need revisiting.
-         AMDGPUAS::FLAT_ADDRESS,       // SPIRAS_CodeSectionINTEL
+         AMDGPUAS::FLAT_ADDRESS, // SPIRAS_CodeSectionINTEL
      },
      AMDGPUAS::FLAT_ADDRESS},
 };
--- a/lib/SPIRV/LLVMSPIRVOpts.cpp
+++ b/lib/SPIRV/LLVMSPIRVOpts.cpp
@@ -44,10 +44,55 @@
 #include <llvm/ADT/SmallVector.h>
 #include <llvm/ADT/StringRef.h>
 #include <llvm/IR/IntrinsicInst.h>
+#include <llvm/Support/AMDGPUAddrSpace.h>
+#include <llvm/TargetParser/Triple.h>
 #include <optional>
 
 using namespace llvm;
 using namespace SPIRV;
+
+namespace {
+// AddrSpaceMap is a positional std::array; Map is indexed by SPIRAS_*.
+struct TargetAddrSpaceMapping {
+  Triple::ArchType Arch;
+  AddrSpaceMap Map;
+  // Pinned so getFunctionProgramAddrSpace() does not fall back to the private
+  // AS and emit a spurious -P<n> in the datalayout.
+  uint32_t ProgramAS;
+};
+
+constexpr TargetAddrSpaceMapping BuiltinAddrSpaceMaps[] = {
+    {Triple::amdgpu,
+     {
+         AMDGPUAS::PRIVATE_ADDRESS,    // SPIRAS_Private
+         AMDGPUAS::GLOBAL_ADDRESS,     // SPIRAS_Global
+         AMDGPUAS::CONSTANT_ADDRESS,   // SPIRAS_Constant
+         AMDGPUAS::LOCAL_ADDRESS,      // SPIRAS_Local
+         AMDGPUAS::FLAT_ADDRESS,       // SPIRAS_Generic
+         AMDGPUAS::GLOBAL_ADDRESS,     // SPIRAS_GlobalDevice
+         AMDGPUAS::GLOBAL_ADDRESS,     // SPIRAS_GlobalHost
+         AMDGPUAS::BUFFER_FAT_POINTER, // SPIRAS_Input
+         AMDGPUAS::BUFFER_RESOURCE,    // SPIRAS_Output
+         // TODO: FLAT clashes with BUFFER_STRIDED_POINTER; may need revisiting.
+         AMDGPUAS::FLAT_ADDRESS,       // SPIRAS_CodeSectionINTEL
+     },
+     AMDGPUAS::FLAT_ADDRESS},
+};
+} // namespace
+
+void TranslatorOpts::deriveTargetAddrSpaces() {
+  // An explicit --spirv-addrspace-map wins.
+  if (getAddrSpaceMap())
+    return;
+  Triple TT(Triple::normalize(getSPIRVTargetTriple()));
+  for (const auto &Entry : BuiltinAddrSpaceMaps) {
+    if (TT.getArch() != Entry.Arch)
+      continue;
+    setAddrSpaceMap(Entry.Map);
+    setFunctionProgramAddrSpace(Entry.ProgramAS);
+    return;
+  }
+}
 
 void TranslatorOpts::enableAllExtensions() {
 #define EXT(X) ExtStatusMap[ExtensionID::X] = true;
--- a/lib/SPIRV/LLVMSPIRVOpts.cpp
+++ b/lib/SPIRV/LLVMSPIRVOpts.cpp
@@ -80,10 +80,10 @@ constexpr TargetAddrSpaceMapping BuiltinAddrSpaceMaps[] = {
 };
 } // namespace
 
-void TranslatorOpts::deriveTargetAddrSpaces() {
+bool TranslatorOpts::deriveTargetAddrSpaces() {
   // An explicit --spirv-addrspace-map wins.
   if (getAddrSpaceMap())
-    return;
+    return true;
   Triple TT(Triple::normalize(getSPIRVTargetTriple()));
   for (const auto &Entry : BuiltinAddrSpaceMaps) {
     if (TT.getArch() != Entry.Arch)
@@ -92,8 +92,12 @@ void TranslatorOpts::deriveTargetAddrSpaces() {
     // An explicit --spirv-function-program-addrspace wins.
     if (!FunctionProgramAS.has_value())
       setFunctionProgramAddrSpace(Entry.ProgramAS);
-    return;
+    return true;
   }
+  // SPIR/SPIRV: default is correct. Any other valid target lacking a map: error
+  // rather than emit SPIR numbering under it. UnknownArch is explicitly allowed
+  // here because it will error out later with a better, more explicit error.
+  return TT.isSPIR() || TT.isSPIRV() || TT.getArch() == Triple::UnknownArch;
 }
 
 void TranslatorOpts::enableAllExtensions() {
--- a/lib/SPIRV/LLVMSPIRVOpts.cpp
+++ b/lib/SPIRV/LLVMSPIRVOpts.cpp
@@ -89,7 +89,9 @@ void TranslatorOpts::deriveTargetAddrSpaces() {
     if (TT.getArch() != Entry.Arch)
       continue;
     setAddrSpaceMap(Entry.Map);
-    setFunctionProgramAddrSpace(Entry.ProgramAS);
+    // An explicit --spirv-function-program-addrspace wins.
+    if (!FunctionProgramAS.has_value())
+      setFunctionProgramAddrSpace(Entry.ProgramAS);
     return;
   }
 }
--- a/lib/SPIRV/LLVMSPIRVOpts.cpp
+++ b/lib/SPIRV/LLVMSPIRVOpts.cpp
@@ -64,16 +64,25 @@ struct TargetAddrSpaceMapping {
 constexpr TargetAddrSpaceMapping BuiltinAddrSpaceMaps[] = {
     {Triple::amdgpu,
      {
-         AMDGPUAS::PRIVATE_ADDRESS,    // SPIRAS_Private
-         AMDGPUAS::GLOBAL_ADDRESS,     // SPIRAS_Global
-         AMDGPUAS::CONSTANT_ADDRESS,   // SPIRAS_Constant
-         AMDGPUAS::LOCAL_ADDRESS,      // SPIRAS_Local
-         AMDGPUAS::FLAT_ADDRESS,       // SPIRAS_Generic
-         AMDGPUAS::GLOBAL_ADDRESS,     // SPIRAS_GlobalDevice
-         AMDGPUAS::GLOBAL_ADDRESS,     // SPIRAS_GlobalHost
+         AMDGPUAS::PRIVATE_ADDRESS,  // SPIRAS_Private
+         AMDGPUAS::GLOBAL_ADDRESS,   // SPIRAS_Global
+         AMDGPUAS::CONSTANT_ADDRESS, // SPIRAS_Constant
+         AMDGPUAS::LOCAL_ADDRESS,    // SPIRAS_Local
+         AMDGPUAS::FLAT_ADDRESS,     // SPIRAS_Generic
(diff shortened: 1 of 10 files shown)


maarquitos14:
@MrSidims FYI

lib/SPIRV/LLVMSPIRVOpts.cpp
```diff
@@ -44,11 +44,56 @@
...
+     {
+         AMDGPUAS::PRIVATE_ADDRESS,    // SPIRAS_Private
+         AMDGPUAS::GLOBAL_ADDRESS,     // SPIRAS_Global
+         AMDGPUAS::CONSTANT_ADDRESS,   // SPIRAS_Constant
+         AMDGPUAS::LOCAL_ADDRESS,      // SPIRAS_Local
+         AMDGPUAS::FLAT_ADDRESS,       // SPIRAS_Generic
+         AMDGPUAS::GLOBAL_ADDRESS,     // SPIRAS_GlobalDevice
+         AMDGPUAS::GLOBAL_ADDRESS,     // SPIRAS_GlobalHost
+         AMDGPUAS::BUFFER_FAT_POINTER, // SPIRAS_Input
+         AMDGPUAS::BUFFER_RESOURCE,    // SPIRAS_Output
```
MrSidims, line 75: This look wrong, and it should be causing regressions as __spirv_BuiltIn* uses Input. I'd consult with @AlexVlx regarding which mapping to use here.

lib/SPIRV/LLVMSPIRVOpts.cpp
```diff
@@ -44,11 +44,56 @@
...
+void TranslatorOpts::deriveTargetAddrSpaces() {
+  // An explicit --spirv-addrspace-map wins.
+  if (getAddrSpaceMap())
+    return;
+  Triple TT(Triple::normalize(getSPIRVTargetTriple()));
+  for (const auto &Entry : BuiltinAddrSpaceMaps) {
+    if (TT.getArch() != Entry.Arch)
+      continue;
+    setAddrSpaceMap(Entry.Map);
+    setFunctionProgramAddrSpace(Entry.ProgramAS);
```
MrSidims: This seem to override --spirv-function-program-addrspace . Is it expected?

lib/SPIRV/SPIRVReader.cpp
```diff
@@ -6032,7 +6032,15 @@ SPIRVModuleTextReport formatSpirvReport(const SPIRVModuleReport &Report) {
...
                                              const SPIRV::TranslatorOpts &Opts,
                                              std::string &ErrMsg) {
-  std::unique_ptr<SPIRVModule> BM(SPIRVModule::createSPIRVModule(Opts));
+  const SPIRV::TranslatorOpts *EffectiveOpts = &Opts;
+  SPIRV::TranslatorOpts AdjustedOpts;
+  if (!Opts.getSPIRVTargetTriple().empty()) {
+    AdjustedOpts = Opts;
+    AdjustedOpts.deriveTargetAddrSpaces();
+    EffectiveOpts = &AdjustedOpts;
+  }
```
MrSidims, line 6046: Should it also be updating Opts?

test/reverse-target-triple-amdgcn-addrspace.ll
```diff
@@ -0,0 +1,73 @@
...
+
+; RUN: llvm-spirv -r %t.spv --spirv-target-triple=amdgcn-amd-amdhsa \
+; RUN:   -o - | llvm-dis | FileCheck %s --check-prefix=CHECK-AMDGCN
+
+; RUN: llvm-spirv -r %t.spv \
+; RUN:   -o - | llvm-dis | FileCheck %s --check-prefix=CHECK-DEFAULT
+
+; Untabled (non-AMDGCN) triple: falls through to default.
+; RUN: llvm-spirv -r %t.spv --spirv-target-triple=nvptx64-nvidia-cuda \
+; RUN:   -o - | llvm-dis | FileCheck %s --check-prefix=CHECK-DEFAULT
```
MrSidims: This should be an error imho, not a silent fallback

test/reverse-target-triple-amdgcn-addrspace.ll
```diff
@@ -0,0 +1,73 @@
...
+
+; Global (1) -> AMDGPU GLOBAL (1); addrspace(1) every case.
+; CHECK-AMDGCN: @gv = {{.*}}addrspace(1){{.*}}global i32
+; CHECK-DEFAULT: @gv = {{.*}}addrspace(1){{.*}}global i32
+; CHECK-OVERRIDE: @gv = {{.*}}addrspace(1){{.*}}global i32
+@gv = addrspace(1) global i32 0, align 4
+
+; Global (1), local (3) stable. Generic (4) -> AMDGPU flat (0) under amdgcn;
+; stays (4) under default and 0:5.
+; CHECK-AMDGCN: define{{.*}} @test_stable_and_generic({{.*}}ptr addrspace(1){{.*}}ptr addrspace(3){{.*}}ptr{{( addrspace\(0\))?}}
```
MrSidims, line 63: I don't get, which AS we have for generic?

lib/SPIRV/LLVMSPIRVOpts.cpp
```diff
@@ -44,11 +44,56 @@
...
+         AMDGPUAS::FLAT_ADDRESS, // SPIRAS_CodeSectionINTEL
+     },
+     AMDGPUAS::FLAT_ADDRESS},
+};
+} // namespace
+
+void TranslatorOpts::deriveTargetAddrSpaces() {
+  // An explicit --spirv-addrspace-map wins.
+  if (getAddrSpaceMap())
+    return;
```
MrSidims: Is this option documented?

lib/SPIRV/LLVMSPIRVOpts.cpp
maarquitos14: This is taken from https://github.com/ROCm/SPIRV-LLVM-Translator/blob/amd-staging/lib/SPIRV/OCLUtil.h#L508.

lib/SPIRV/SPIRVReader.cpp
maarquitos14: I don't think we can do that, `Opts` is const.

lib/SPIRV/LLVMSPIRVOpts.cpp
maarquitos14: It is not! I wasn't aware of this flag. Addressed by https://github.com/KhronosGroup/SPIRV-LLVM-Translator/pull/3994/commits/d0aa37f6005c834ac34dfc5a3f3a4c903da3eb86. Thanks for the catch!

test/reverse-target-triple-amdgcn-addrspace.ll
maarquitos14: Done in https://github.com/KhronosGroup/SPIRV-LLVM-Translator/pull/3994/commits/adfdab2f12c3cab7df2d7906a2cbe0d8b2a91ffa.

lib/SPIRV/LLVMSPIRVOpts.cpp
maarquitos14: It is, at least in the help output of the translator. Is that what you meant?

https://github.com/KhronosGroup/SPIRV-LLVM-Translator/blob/d916d6b34ee50930e92d51db7b9a69a87b864be4/tools/llvm-spirv/llvm-spirv.cpp#L184-L191

test/reverse-target-triple-amdgcn-addrspace.ll
maarquitos14: `generic` (which is 4 in SPIR) maps to 0 for AMDGPU. 

Clarified in https://github.com/KhronosGroup/SPIRV-LLVM-Translator/pull/3994/commits/70de89d371585023d1ac3843db8febaab87bb5db. Let me know if that is okay now or it's too verbose.

lib/SPIRV/LLVMSPIRVOpts.cpp
MrSidims: Lets leave it for later than.

lib/SPIRV/SPIRVReader.cpp
MrSidims: `readSpirv` calls `readSpirvModule` and `convertSpirvToLLVM(...opts...)` with original, not updated `opts`. As of right now it's not a problem, because it only needs it for getting desired IR representation (OCL vs SPIR-V friendly). But keeping it as is is a recipe for potential bugs in the future.

test/reverse-target-triple-amdgcn-addrspace.ll
MrSidims: Got it, thanks! One day I'll start remembering this.

test/reverse-target-triple-amdgcn-addrspace.ll
```diff
@@ -0,0 +1,102 @@
...
+;   generic          4       0   (bare `ptr`)
+;   GlobalDevice     5       1
+;   GlobalHost       6       1
+
+; SPV_INTEL_usm_storage_classes: keeps GlobalDevice/GlobalHost forward, so the
+; reverse map is exercised for them.
+; RUN: llvm-spirv %s --spirv-ext=+SPV_INTEL_usm_storage_classes -o %t.spv
+
+; RUN: llvm-spirv -r %t.spv --spirv-target-triple=amdgcn-amd-amdhsa \
+; RUN:   -o - | llvm-dis | FileCheck %s --check-prefix=CHECK-AMDGCN
```
MrSidims, line 21: Please check datalayout as well

lib/SPIRV/LLVMSPIRVOpts.cpp
MrSidims: My bad. I forgot about this newly added option

MrSidims:
Overall LGTM. Would be nice to hear from @svenvh @vmaksimo and @mbelicki

lib/SPIRV/SPIRVReader.cpp
maarquitos14: Okay, so the problem I see with moving it to `readSpirv` is that `readSpirvModule` is also a public API, so if the code is not there too, a case where `readSpirvModule` is directly called with `Opts` setting the triple but not the address space map, wouldn't get the correct address spaces. Therefore, we need to duplicate the code (and then we might end up with a divergence again). Do you think duplicating (likely with the code outlined into a function) is the lesser evil? In the case where `readSpirv` already derives the map, `readSpirvModule` would return early in the first branch of `deriveTargetAddrSpaces()`.

lib/SPIRV/LLVMSPIRVOpts.cpp
vmaksimo: I think the concern @MrSidims raised is still open. 
Should we add a short comment/TODO noting this is unvalidated for the OpenCL-builtin case, so it's discoverable later?

lib/SPIRV/LLVMSPIRVOpts.cpp
```diff
@@ -44,11 +44,62 @@
...
+         AMDGPUAS::PRIVATE_ADDRESS,    // SPIRAS_Private
+         AMDGPUAS::GLOBAL_ADDRESS,     // SPIRAS_Global
+         AMDGPUAS::CONSTANT_ADDRESS,   // SPIRAS_Constant
+         AMDGPUAS::LOCAL_ADDRESS,      // SPIRAS_Local
+         AMDGPUAS::FLAT_ADDRESS,       // SPIRAS_Generic
+         AMDGPUAS::GLOBAL_ADDRESS,     // SPIRAS_GlobalDevice
+         AMDGPUAS::GLOBAL_ADDRESS,     // SPIRAS_GlobalHost
+         AMDGPUAS::BUFFER_FAT_POINTER, // SPIRAS_Input
+         AMDGPUAS::BUFFER_RESOURCE,    // SPIRAS_Output
+         // TODO: FLAT clashes with BUFFER_STRIDED_POINTER; may need revisiting.
```
vmaksimo: Nit: maybe to clarify a bit, as from the first reading it's not clear how `0` clashes with `9`
```suggestion
         // TODO: Generic and CodeSectionINTEL share FLAT; consider using a distinct BUFFER_STRIDED_POINTER
```

lib/SPIRV/SPIRVReader.cpp
MrSidims: We can make readSpirv take its opts from the module it just built, instead of from its own parameter, smthing like that:

```
  // SPIRVModule.h
  const SPIRV::TranslatorOpts &getTranslationOpts() const noexcept {
    return TranslationOpts;
  }
```

```
  // SPIRVReader.cpp, llvm::readSpirv
  std::unique_ptr<SPIRVModule> BM(readSpirvModule(IS, Opts, ErrMsg));
  ...
  M = convertSpirvToLLVM(C, *BM, BM->getTranslationOpts(), ErrMsg).release();
```
readSpirvModule stays the single normalization point, readSpirv can no longer diverge from it by construction, and any future opt that gets normalized on the way in is picked up for free.

test/reverse-target-triple-amdgcn-addrspace.ll
```diff
@@ -0,0 +1,118 @@
...
+
+target datalayout = "e-i64:64-v16:16-v24:32-v32:32-v48:64-v96:128-v192:256-v256:256-v512:512-v1024:1024-G1"
+target triple = "spir64-unknown-unknown"
+
+; Datalayout follows the triple: -A is the alloca AS, -P the program
+; (function) AS.
+
+; -A5: derived map put Private -> 5. No -P: the triple pins the program AS to
+; flat (0), the default, so it is elided.
+; CHECK-AMDGCN: target datalayout = "m:e-e-p:64:64:64{{.*}}-A5-G1-ni:
```
MrSidims: This looks odd actually. Is it our deed or auto-upgraders? AFAIK AMDGPU has no pointers 64 bit width.

lib/SPIRV/LLVMSPIRVOpts.cpp
maarquitos14: Addressed by https://github.com/KhronosGroup/SPIRV-LLVM-Translator/pull/3994/commits/d92c8426d121624ed1e140e286a6b138f5bd421a.

lib/SPIRV/SPIRVReader.cpp
maarquitos14: Done, thanks for the suggestion!

test/reverse-target-triple-amdgcn-addrspace.ll
maarquitos14: Done in https://github.com/KhronosGroup/SPIRV-LLVM-Translator/pull/3994/commits/d92c8426d121624ed1e140e286a6b138f5bd421a

lib/SPIRV/LLVMSPIRVOpts.cpp
maarquitos14: I clarified it, let me know if it reads better now.

test/reverse-target-triple-amdgcn-addrspace.ll
maarquitos14: Good eye! It was a little bit of both. Fixed by https://github.com/KhronosGroup/SPIRV-LLVM-Translator/pull/3994/commits/d92c8426d121624ed1e140e286a6b138f5bd421a.

MrSidims (approved):

MrSidims:
Friendly ping @svenvh @vmaksimo and @mbelicki . Please take a look if the approach is fine for you. It _might_ break something due to https://github.com/KhronosGroup/SPIRV-LLVM-Translator/pull/3994#discussion_r3913703191 but definitely not on your end :)

svenvh:
> Friendly ping @svenvh [...]. Please take a look if the approach is fine for you. It _might_ break something due to [#3994 (comment)](https://github.com/KhronosGroup/SPIRV-LLVM-Translator/pull/3994#discussion_r3913703191) but definitely not on your end :)

No concerns from our side.

maarquitos14:
Merging this, @vmaksimo @mbelicki feel free to provide post-merge feedback, and I'll address it in a follow-up PR.

PR #4003, opened 2026-09-04
additional tests and other improvements for SPV_INTEL_16bit_atomics

bashbaug:
This PR adds some additional tests for SPV_INTEL_16bit_atomics and some improvements to pass the new tests.

### DIFF
--- a/lib/SPIRV/libSPIRV/SPIRVInstruction.h
+++ b/lib/SPIRV/libSPIRV/SPIRVInstruction.h
@@ -3053,6 +3053,36 @@ _SPIRV_OP(PowN, true, 11)
 
 class SPIRVAtomicInstBase : public SPIRVInstTemplateBase {
 public:
+  // SPV_INTEL_16bit_atomics allows bfloat16 only for this subset of the atomic
+  // instructions; notably, compare-exchange is not included.
+  static bool allowsBFloat16(Op OC) {
+    switch (OC) {
+    case OpAtomicLoad:
+    case OpAtomicStore:
+    case OpAtomicExchange:
+    case OpAtomicFAddEXT:
+    case OpAtomicFMinEXT:
+    case OpAtomicFMaxEXT:
+      return true;
+    default:
+      return false;
+    }
+  }
+
+  bool isBFloat16Atomic() const {
+    return hasType() && getType()->isTypeFloat(16, FPEncodingBFloat16KHR);
+  }
+
+  bool isInt16Atomic() const { return hasType() && getType()->isTypeInt(16); }
+
+  std::optional<ExtensionID> getRequiredExtension() const override {
+    // The OpenCL SPIR-V environment permits only 32-bit atomics, widened to
+    // 64-bit by Int64Atomics, so any 16-bit atomic requires the extension.
+    if (isInt16Atomic() || (isBFloat16Atomic() && allowsBFloat16(OpCode)))
+      return ExtensionID::SPV_INTEL_16bit_atomics;
+    return SPIRVInstTemplateBase::getRequiredExtension();
+  }
+
   SPIRVCapVec getRequiredCapability() const override {
     // Most of the atomic instructions require a specific capability when
     // operating on 64-bit integers.
@@ -3065,12 +3095,8 @@ public:
     if (hasType()) {
       if (getType()->isTypeInt(64))
         return {CapabilityInt64Atomics};
-      if (getType()->isTypeInt(16) &&
-          Module->isAllowedToUseExtension(
-              ExtensionID::SPV_INTEL_16bit_atomics)) {
-        Module->addExtension(ExtensionID::SPV_INTEL_16bit_atomics);
+      if (isInt16Atomic())
         return {internal::CapabilityInt16AtomicsINTEL};
-      }
     }
     return {};
   }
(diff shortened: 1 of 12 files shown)


bashbaug:
I should have written a more detailed PR description - sorry about that!

Here is some more information about these changes.  Note, for easier reference, the current spec for this extension is here: https://github.com/intel/llvm/blob/sycl/sycl/doc/design/spirv-extensions/SPV_INTEL_16bit_atomics.asciidoc

1. AtomicLoadStoreExchangeBFloat16.ll - Adds a test for bfloat16 OpAtomicLoad, OpAtomicStore, and OpAtomicExchange instructions, which was missing. These instructions should generate the AtomicBFloat16LoadStoreINTEL capability (plus BFloat16TypeKHR).
2. AtomicCompareExchangeWeakInt16.ll - Adds a test for OpAtomicCompareExchangeWeak with 16-bit integer atomics, which was missing.
3. AtomicStoreInt16.ll‎ - Adds a dedicated test for OpAtomicStore with a 16-bit integer source.  This is a special case, since there is no result (and hence no result type) for this instruction.
4. Int16AtomicsINTELOps.ll‎ - Add tests for all other 16-bit integer atomic operations - add, sub, signed and unsigned min and max, and, or, increment, and decrement.  Maybe this should be added to the existing Int16AtomicsINTEL.ll insetad?
5. negative-cmpxchg-bfloat16.ll‎ - Adds a test to ensure that bfloat16 atomic compare-and-exchange is NOT supported by SPV_INTEL_16bit_atomics.  Because atomic compare-and-exchange is bitwise, there is no need to support a specific bfloat16 atomic compare-and-exchange instruction.
6. negative-loadstore.ll - Adds a test to check that bfloat16 atomic load requires support for both SPV_KHR_bfloat16 and SPV_INTEL_16bit_atomics.
7. negative-minmax.ll‎ - Adds a test to check that bfloat16 atomic min requires support for both SPV_KHR_bfloat16 and SPV_INTEL_16bit_atomics.
8. Various - Adds checks to ensure a proper error is generated if 16-bit atomics are used without enabling the SPV_INTEL_16bit_atomics extension.

Most of these tests are currently written to use `__spirv` intrinsics. Should they use LLVM atomics instead?

Then, make changes so these tests pass:

1. Record that the new SPV_INTEL_16bit_atomic capabilities require the SPV_INTEL_16bit_atomic extension.  This is something we should look at improving in the future (in a separate PR).
2. For SPIRVAtomicInstBase, add checks for SPV_INTEL_16bit_atomics to getRequiredExtensions, vs. adding the extension as part of getRequiredCapability.  Note, if the capability is required and we aren't allowed to use the extension, we should generate an error.
3. Additionally, for SPIRVAtomicInstBase, validate that only the correct subset of bfloat16 atomic operations is supported.
4. For SPIRVAtomicCompareExchangeInstructions, return the proper required capability for both bfloat16 and 16-bit integer atomics.  Note, the extension requirement is added by getRequiredExtensions.
5. For SPIRVAtomicStoreInst, add handling for bfloat16 atomic stores and 16-bit integer atomic stores.

test/extensions/INTEL/SPV_INTEL_16bit_atomics/AtomicInt16CompareExchangeCapability.ll
```diff
@@ -6,22 +6,17 @@
...
 ; RUN: FileCheck < %t.rev.ll %s --check-prefixes=CHECK-LLVM
 
-; Check that without extension we don't use its capabilities - there is no
-; limitation on using i16 with atomic instruction in the core specification.
-; RUN: llvm-spirv %s -o %t.noext.spv
-; RUN: spirv-val %t.noext.spv
-; RUN: llvm-spirv -to-text %t.noext.spv -o %t.noext.spt
-; RUN: FileCheck < %t.noext.spt %s --check-prefix=CHECK-SPIRV-NOEXT
+; The OpenCL SPIR-V environment supports only 32-bit atomics (64-bit with
+; Int64Atomics), so 16-bit atomics require the extension.
```
vmaksimo, line 10: Why we should be guided by OCL SPIR-V env spec instead of core spec? I see that the restriction on types was removed in SPIR-V 1.2 Rev 4:
> \#\254: Remove validation rules about the types atomic instructions can operate on. These rules belong instead to the client API.

See [earlier discussion](https://github.com/KhronosGroup/SPIRV-LLVM-Translator/pull/3424#discussion_r2510525807) for some context where we decided to proceed with not limiting 16-bit atomic operations.

(have we maybe filed the issue in Khronos about this that i don't remember by now?)

lib/SPIRV/libSPIRV/SPIRVInstruction.h
```diff
@@ -3116,25 +3149,34 @@ class SPIRVAtomicInstBase : public SPIRVInstTemplateBase {
...
+    SPIRVType *ValueTy = getOperand(ValueOperandIndex)->getType();
+    if (ValueTy->isTypeInt(64)) {
       Module->addCapability(CapabilityInt64Atomics);
+    } else if (ValueTy->isTypeFloat(16, FPEncodingBFloat16KHR)) {
+      Module->addExtension(ExtensionID::SPV_INTEL_16bit_atomics);
+      Module->addCapability(internal::CapabilityAtomicBFloat16LoadStoreINTEL);
+    } else if (ValueTy->isTypeInt(16)) {
+      Module->addExtension(ExtensionID::SPV_INTEL_16bit_atomics);
+      Module->addCapability(
+          internal::CapabilityAtomicInt16CompareExchangeINTEL);
```
vmaksimo, line 3178: Should not this be enough? You've just updated `getRequiredExtension` to map these capabilities to the extension.
```suggestion
      Module->addCapability(internal::CapabilityAtomicBFloat16LoadStoreINTEL);
    } else if (ValueTy->isTypeInt(16)) {
      Module->addCapability(
          internal::CapabilityAtomicInt16CompareExchangeINTEL);
```

test/extensions/INTEL/SPV_INTEL_16bit_atomics/negative-minmax.ll
```diff
@@ -0,0 +1,23 @@
+; Check that bfloat16 atomic min/max and load/store/exchange require both
```
vmaksimo, line 1:
```suggestion
; Check that bfloat16 atomic min/max require both
```

PR #4009, opened 2026-09-11
Fix handling of checkError calls on null pointer check

Qi-Ye-079:
Some `checkError` calls checks for non-null pointers, without checking its return value and no early return if it fails. This is fine if `SPIRVErrorLog` defaults to abort/exit; however, when it's set to "ignore" and the pointers are null, it will still continue execution and end up dereferencing null pointers.

This PR adds the check of the return value and an early return for `checkError` calls on null pointer checks, for a more graceful failure instead of a segfault by dereferencing null pointers in "ignore" mode. These are flagged by our internal static analysis tool and should be fixed.

NOTE: this PR fixes `checkError` on pointers only, to avoid segfault when dereferencing null pointers. `checkError` on non-pointer checks aren't included in this PR, as there are still _many_ inproper `checkError` calls across the entire repo; fixing them is out of scope for this PR, and at this point, they aren't as harmful as dereferencing null pointers.

### DIFF
--- a/lib/SPIRV/SPIRVWriter.cpp
+++ b/lib/SPIRV/SPIRVWriter.cpp
@@ -2959,8 +2959,9 @@ void addFuncPointerCallArgumentAttributes(CallInst *CI,
     ErrLog.checkError(NumOperands == 2, SPIRVEC_InvalidLlvmModule,             \
                       #NAME " requires exactly 1 extra operand");              \
     auto *StrDecoEO = dyn_cast<MDString>(DecoMD->getOperand(1));               \
-    ErrLog.checkError(StrDecoEO, SPIRVEC_InvalidLlvmModule,                    \
-                      #NAME " requires extra operand to be a string");         \
+    if (!ErrLog.checkError(StrDecoEO, SPIRVEC_InvalidLlvmModule,               \
+                           #NAME " requires extra operand to be a string"))    \
+      return;                                                                  \
     Target->addDecorate(                                                       \
         new SPIRVDecorate##NAME##Attr(Target, StrDecoEO->getString().str()));  \
     break;                                                                     \
@@ -2972,8 +2973,9 @@ void addFuncPointerCallArgumentAttributes(CallInst *CI,
                       #NAME " requires exactly 1 extra operand");              \
     auto *IntDecoEO =                                                          \
         mdconst::dyn_extract<ConstantInt>(DecoMD->getOperand(1));              \
-    ErrLog.checkError(IntDecoEO, SPIRVEC_InvalidLlvmModule,                    \
-                      #NAME " requires extra operand to be an integer");       \
+    if (!ErrLog.checkError(IntDecoEO, SPIRVEC_InvalidLlvmModule,               \
+                           #NAME " requires extra operand to be an integer"))  \
+      return;                                                                  \
     Target->addDecorate(new SPIRVDecorate##NAME(                               \
         Target, static_cast<TYPE>(IntDecoEO->getZExtValue())));                \
     break;                                                                     \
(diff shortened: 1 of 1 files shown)


lib/SPIRV/SPIRVWriter.cpp
```diff
@@ -2959,8 +2959,9 @@ void addFuncPointerCallArgumentAttributes(CallInst *CI,
     ErrLog.checkError(NumOperands == 2, SPIRVEC_InvalidLlvmModule,             \
```
vmaksimo, line 2959: This also needs early return, right?
(same for other macros)

lib/SPIRV/SPIRVWriter.cpp
```diff
@@ -3018,14 +3023,16 @@ static void transMetadataDecorations(Metadata *MD, SPIRVValue *Target) {
   assert(ArgDecoMD && "Decoration list must be a metadata node");
   for (unsigned I = 0, E = ArgDecoMD->getNumOperands(); I != E; ++I) {
     auto *DecoMD = dyn_cast<MDNode>(ArgDecoMD->getOperand(I));
-    ErrLog.checkError(DecoMD, SPIRVEC_InvalidLlvmModule,
-                      "Decoration does not name metadata");
+    if (!ErrLog.checkError(DecoMD, SPIRVEC_InvalidLlvmModule,
+                           "Decoration does not name metadata"))
+      return;
     ErrLog.checkError(DecoMD->getNumOperands() > 0, SPIRVEC_InvalidLlvmModule,
```
vmaksimo, line 3029: What about this one?

lib/SPIRV/SPIRVWriter.cpp
```diff
@@ -3069,8 +3077,10 @@ static void transMetadataDecorations(Metadata *MD, SPIRVValue *Target) {
       ErrLog.checkError(NumOperands == 2, SPIRVEC_InvalidLlvmModule,
```
vmaksimo, line 3077: This?

lib/SPIRV/SPIRVWriter.cpp
vmaksimo, line 3010: Here?

vmaksimo (commented):
Please revisit the checks in this file again, as I understood the purpose of the fix - more places are to be fixed here

Qi-Ye-079:
> Please revisit the checks in this file again, as I understood the purpose of the fix - more places are to be fixed here

Maybe I should emphasize, that the scope of this PR is about fixing `checkError` on **pointers only**, to avoid segfault when dereferencing it in Ignore mode. This is what's flagged by Coverity, and I should update the description to make this clearer.
And you're absolutely correct - that there are _much more_ usage of `checkError` (other than checking pointers) that needs fixing, maybe ~100 more places in the repo. But there are too many of them, are out of scope of this PR, and aren't so harmful as dereferencing null pointers, so I won't include fixes for these non-pointer checks for this PR.

lib/SPIRV/SPIRVWriter.cpp
Qi-Ye-079: Please see my reply to your comment.

lib/SPIRV/SPIRVWriter.cpp
Qi-Ye-079: Same, see my reply.

lib/SPIRV/SPIRVWriter.cpp
Qi-Ye-079: Same, see my reply.

lib/SPIRV/SPIRVWriter.cpp
Qi-Ye-079: Same, see my reply.

vmaksimo:
Hm, I see. Do we track this as an issue anywhere?

Qi-Ye-079:
> Hm, I see. Do we track this as an issue anywhere?

Not that I can find. We can open new tracker for other `checkError` checks.

vmaksimo (approved):

PR #4010, opened 2026-09-14
Translate branch weight metadata to and from `OpBranchConditional`

vmaksimo:
Conditional branches carrying `!prof branch_weights` metadata (from `__builtin_expect`, `[[likely]]/[[unlikely]]`, or profile-guided optimization) now preserve those weights through the optional `Branch Weights` literal operands of `OpBranchConditional`.
Weights wider than 32 bits, or whose sum would overflow a 32-bit unsigned integer are downscaled to fit the SPIR-V literal format while preserving their ratio.

Translating such a SPIR-V module back to LLVM IR reconstructs the `!prof` metadata from the operands.
Branches without profiling metadata are unaffected and continue to emit the 4-word form of the instruction.

### DIFF
--- a/lib/SPIRV/libSPIRV/SPIRVModule.cpp
+++ b/lib/SPIRV/libSPIRV/SPIRVModule.cpp
@@ -353,9 +353,9 @@ public:
 
   // Constant creation functions
   SPIRVInstruction *addBranchInst(SPIRVLabel *, SPIRVBasicBlock *) override;
-  SPIRVInstruction *addBranchConditionalInst(SPIRVValue *, SPIRVLabel *,
-                                             SPIRVLabel *,
-                                             SPIRVBasicBlock *) override;
+  SPIRVInstruction *addBranchConditionalInst(
+      SPIRVValue *, SPIRVLabel *, SPIRVLabel *, SPIRVBasicBlock *,
+      const std::vector<SPIRVWord> &BranchWeights = {}) override;
   SPIRVValue *addCompositeConstant(SPIRVType *,
                                    const std::vector<SPIRVValue *> &) override;
   SPIRVEntry *addCompositeConstantContinuedINTEL(
@@ -1832,7 +1832,12 @@ SPIRVInstruction *SPIRVModuleImpl::addBranchInst(SPIRVLabel *TargetLabel,
 
 SPIRVInstruction *SPIRVModuleImpl::addBranchConditionalInst(
     SPIRVValue *Condition, SPIRVLabel *TrueLabel, SPIRVLabel *FalseLabel,
-    SPIRVBasicBlock *BB) {
+    SPIRVBasicBlock *BB, const std::vector<SPIRVWord> &BranchWeights) {
+  if (BranchWeights.size() == 2)
+    return addInstruction(
+        new SPIRVBranchConditional(Condition, TrueLabel, FalseLabel, BB,
+                                   BranchWeights[0], BranchWeights[1]),
+        BB);
   return addInstruction(
       new SPIRVBranchConditional(Condition, TrueLabel, FalseLabel, BB), BB);
 }
--- a/test/branching/OpBranchConditionalWeights.ll
+++ b/test/branching/OpBranchConditionalWeights.ll
@@ -1,0 +1,52 @@
+; REQUIRES: spirv-dis
+; RUN: llvm-spirv %s -o %t.spv
+; RUN: spirv-val %t.spv
+; RUN: spirv-dis %t.spv | FileCheck %s --check-prefix=CHECK-SPIRV
+; RUN: llvm-spirv %t.spv -o %t.rev.bc -r --spirv-target-env=SPV-IR
+; RUN: llvm-dis %t.rev.bc -o %t.rev.ll
+; RUN: FileCheck --input-file=%t.rev.ll %s --check-prefix=CHECK-LLVM
+
+; Branch weight metadata (as produced by __builtin_expect, [[likely]]/
+; [[unlikely]], or PGO) is translated to and from the optional Branch
+; Weights operands of OpBranchConditional.
+
+target datalayout = "e-i64:64-v16:16-v24:32-v32:32-v48:64-v96:128-v192:256-v256:256-v512:512-v1024:1024-G1"
+target triple = "spir64-unknown-unknown"
+
+; CHECK-SPIRV: %weighted_branch = OpFunction
+; CHECK-SPIRV: OpBranchConditional %cmp %if_then %if_else 2000 1
+define spir_func i32 @weighted_branch(i32 %x) {
+entry:
+  %cmp = icmp sgt i32 %x, 10
+  br i1 %cmp, label %if.then, label %if.else, !prof !0
+
+if.then:
+  ret i32 %x
+
+if.else:
+  ret i32 %x
+}
+
+; CHECK-SPIRV: %unweighted_branch = OpFunction
+; CHECK-SPIRV: OpBranchConditional %cmp_0 %if_then_0 %if_else_0{{$}}
+define spir_func i32 @unweighted_branch(i32 %x) {
+entry:
+  %cmp = icmp sgt i32 %x, 10
+  br i1 %cmp, label %if.then, label %if.else
+
+if.then:
+  ret i32 %x
+
+if.else:
+  ret i32 %x
+}
+
+; CHECK-LLVM-LABEL: define spir_func i32 @weighted_branch
+; CHECK-LLVM: br i1 %{{.*}}, label %{{.*}}, label %{{.*}}, !prof ![[#PROF:]]
+
+; CHECK-LLVM-LABEL: define spir_func i32 @unweighted_branch
+; CHECK-LLVM-NOT: !prof
+
+; CHECK-LLVM: ![[#PROF]] = !{!"branch_weights", i32 2000, i32 1}
+
+!0 = !{!"branch_weights", i32 2000, i32 1}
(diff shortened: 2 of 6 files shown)


lib/SPIRV/libSPIRV/SPIRVModule.cpp
```diff
@@ -1832,7 +1832,12 @@ SPIRVInstruction *SPIRVModuleImpl::addBranchInst(SPIRVLabel *TargetLabel,
 
 SPIRVInstruction *SPIRVModuleImpl::addBranchConditionalInst(
     SPIRVValue *Condition, SPIRVLabel *TrueLabel, SPIRVLabel *FalseLabel,
-    SPIRVBasicBlock *BB) {
+    SPIRVBasicBlock *BB, const std::vector<SPIRVWord> &BranchWeights) {
+  if (BranchWeights.size() == 2)
```
maarquitos14, line 1836: In case we have weights, but for whatever reason (e.g. a bug) the vector has another size, we would silently drop them. Should we at least assert or warn?

test/branching/OpBranchConditionalWeights.ll
```diff
@@ -0,0 +1,111 @@
+; REQUIRES: spirv-dis
+; RUN: llvm-spirv %s -o %t.spv
+; RUN: spirv-val %t.spv
+; RUN: spirv-dis %t.spv | FileCheck %s --check-prefix=CHECK-SPIRV
```
maarquitos14, line 4: Can we use `llvm-spirv --to-text` instead? That way this test always runs, even if no spirv-dis is available.

test/branching/OpBranchConditionalWeights.ll
```diff
@@ -0,0 +1,111 @@
...
+; CHECK-LLVM: br i1 %{{.*}}, label %{{.*}}, label %{{.*}}, !prof ![[#PROF3:]]
+
+; CHECK-LLVM-LABEL: define spir_func i32 @zero_weights_branch
+; CHECK-LLVM-NOT: !prof
+
+; CHECK-LLVM: ![[#PROF]] = !{!"branch_weights", i32 2000, i32 1}
+; CHECK-LLVM: ![[#PROF2]] = !{!"branch_weights", i32 1500000000, i32 1500000000}
+; CHECK-LLVM: ![[#PROF3]] = !{!"branch_weights", i32 0, i32 5}
+
+!0 = !{!"branch_weights", i32 2000, i32 1}
```
maarquitos14, line 108: Is it legal LLVM IR to have only one weight, or more than two weights? If so, I'd like to have a test for those cases.

test/branching/OpBranchConditionalWeights.ll
vmaksimo: I believe we're trying to use spirv-tools more wherever possible, so I'd lean towards having spirv-dis as a req

test/branching/OpBranchConditionalWeights.ll
vmaksimo: Yes, but not for `CondBrInst` - there only 2 args are expected.

maarquitos14: So would that fail loudly on LLVM IR side, or how is that enforced?

test/branching/OpBranchConditionalWeights.ll
maarquitos14: Well, I think it's a risk because it can regress without anybody noticing in an environment where spirv-tools is not available, but I won't block the PR on this. If that's the decision that was made, I'm fine with it.

vmaksimo: But CI is always run with the spirv-tools, so we'd catch such a regression via PR's CI runs. Do you think it may be enough? 
I'm not sure there was a specific "decision", just the direction, maybe @svenvh can share his thoughts on this

test/branching/OpBranchConditionalWeights.ll
vmaksimo: Yep, LLVM assembly would fail, e.g.:
```
llvm-as: assembly parsed, but does not verify as correct!
Wrong number of operands
!0 = !{!"branch_weights", i32 2000}
Wrong number of operands
!1= !{!"branch_weights", i32 2000, i32 1, i32 4}
```

lib/SPIRV/libSPIRV/SPIRVModule.cpp
vmaksimo: We already do this silently - `llvm-spirv` just don't handle weighted branches at all.
So this won't be a regressing behavior for sure, and asserting would be too strict change. I'd add a warning, but as I remember we don't have such mechanism in `llvm-spirv` yet

test/branching/OpBranchConditionalWeights.ll
maarquitos14: Right, it won't fail upstream, but could fail downstream. Anyway, not a blocker for me.

lib/SPIRV/libSPIRV/SPIRVModule.cpp
maarquitos14: Definitely wouldn't be a regression now, but could be in the future. I don't know if asserting is too strict, having exactly 2 branch weights it's a pre-condition, and assertions are a way to check exactly that.

test/branching/OpBranchConditionalWeights.ll
svenvh: I don't have a strong opinion as long as we'd catch it in CI.

test/branching/OpBranchConditionalWeights.ll
```diff
@@ -0,0 +1,111 @@
...
+; CHECK-LLVM: br i1 %{{.*}}, label %{{.*}}, label %{{.*}}, !prof ![[#PROF2:]]
+
+; CHECK-LLVM-LABEL: define spir_func i32 @one_zero_weight_branch
+; CHECK-LLVM: br i1 %{{.*}}, label %{{.*}}, label %{{.*}}, !prof ![[#PROF3:]]
+
+; CHECK-LLVM-LABEL: define spir_func i32 @zero_weights_branch
+; CHECK-LLVM-NOT: !prof
+
+; CHECK-LLVM: ![[#PROF]] = !{!"branch_weights", i32 2000, i32 1}
+; CHECK-LLVM: ![[#PROF2]] = !{!"branch_weights", i32 1500000000, i32 1500000000}
```
MrSidims, line 105: ~Looks like round-trip failure.~

test/branching/OpBranchConditionalWeights.ll
```diff
@@ -0,0 +1,111 @@
...
+if.then:
+  ret i32 %x
+
+if.else:
+  ret i32 %x
+}
+
+; Sum overflows 32 bits, so weights are scaled down.
+; CHECK-SPIRV: %large_weights_branch = OpFunction
+; CHECK-SPIRV: OpBranchConditional %cmp_1 %if_then_1 %if_else_1 1500000000 1500000000
```
MrSidims, line 46: ~Or even forward translation failure. Note: input LLVM IR had:~
`!1 = !{!"branch_weights", i32 3000000000, i32 3000000000}`

MrSidims (commented):
the approach makes sense

I didn't take a detailed look into the code, but test's behaviour looks odd, added a comment  there.

**UPD:** I see, it came from:
> Weights wider than 32 bits, or whose sum would overflow a 32-bit unsigned integer are downscaled to fit the SPIR-V literal format while preserving their ratio.

let me think if we have figure out something better.

**UPD2:**
per spec: `The sum of the two weights must not overflow a 32-bit unsigned integer.`

I would suggest to drop hints in this case instead of doing a wrap. IMHO this wrap could harm more, than losing prof information.

vmaksimo:
>  I would suggest to drop hints in this case instead of doing a wrap. IMHO this wrap could harm more, than losing prof information.

@MrSidims Isn't the ratio (`weight / sum`) the only meaningful thing here? 
SPIR-V says so explicitly, and on the LLVM side, `BranchProbabilityInfo::calcMetadataWeights` does the same scaling itself before reducing everything to a ratio:
https://github.com/llvm/llvm-project/blob/5301245b7d20fc4871a3253fd235a31e69f1bdfb/llvm/lib/Analysis/BranchProbabilityInfo.cpp#L318-L321

I believe the use of the `downscaleWeight`s/`setFittedBranchWeights` LLVM API solves the issue you're worrying about - dividing both weights by the same scale won't make the smaller weight end up bigger than the one that started out bigger, so we can't accidentally tell a consumer the wrong branch is hot.

Does it make sense?

MrSidims:
Agreed, I misread it as a wrap.

Yet few things I'd like to point out:
1. per https://llvm.org/docs/BranchWeightMetadata.html branch_weights is always i32, so I'm surprised to see LLVM bitcode writer to accept the test in the first place, may be it downscaled implicitly already?
2. (if you decide to keep the current logic) please add a test exposing, that the ratio is preserved - the test added doesn't show it.
3. (if you decide to keep the current logic) please add a test for exactly UINT32_MAX branch_weights , I feel like we might see a bug there.

vmaksimo:
> 1. per https://llvm.org/docs/BranchWeightMetadata.html branch_weights is always i32, so I'm surprised to see LLVM bitcode writer to accept the test in the first place, may be it downscaled implicitly already?

@MrSidims  Not sure I fully get your concern, but no - no downscaling happens there, there's no restriction on the sum to fit into i32 on the LLVM side. 3000000000 fits fine into unsigned int

> 2. (if you decide to keep the current logic) please add a test exposing, that the ratio is preserved - the test added doesn't show it.

Added the test with ratio 2:1 (3000000000;1500000000 -> 1500000000;750000000)

> 3. (if you decide to keep the current logic) please add a test for exactly UINT32_MAX branch_weights , I feel like we might see a bug there. 

You were right that something can be wrong - "luckily" it's in upstream LLVM.
`calculateCountScale` downscales a sum of exactly UINT32_MAX even though it already fits (checks `< UINT32_MAX` instead of `<= UINT32_MAX`). 
I added a test documenting the current behavior with a note. Will file/fix the issue separately in upstream and update the test once it's fixed.

PR #4012, opened 2026-09-14
[Backport to 22] Support new OpTypeVectorIdEXT op with constant instructions in Reader (#3951)

Qi-Ye-079:
Backport of #3951.

Discrepancies from original PR:
1. All `.spvasm` tests are rewritten to translator-internal `.spt` format to avoid usage of `spirv-as`, which can't recognize `LongVectorEXT` capability and `OpTypeVectorIdEXT`.
2. Change in `builtin_vars_different_type.ll` is excluded, since there's no LLC lines on this branch at all, thus it remains unaffected.


test/extensions/EXT/SPV_EXT_long_vector/negative/OpTypeVectorIdEXT_non32bit_count.spt
```diff
@@ -0,0 +1,34 @@
...
+
+; CHECK: InvalidInstruction
+; CHECK: TypeVectorIdEXT: component count must have scalar 32-bit integer type
+
+119734787 66304 458752 10 0
+2 Capability Addresses
+2 Capability Linkage
+2 Capability Kernel
+2 Capability Int64
+6 Extension "SPV_EXT_long_vector"
```
vmaksimo, line 16: Looks like we're missing `OpCapability LongVectorEXT` in all the tests (unfortunately, in the original PR as well)
Do you think we might need to fix this up?

Qi-Ye-079: I think you misread it, in the original PR all tests do have `OpCapability LongVectorEXT`. Maybe you're reading the diff from this PR, where the old `.spvasm` lost it? This is because the `spirv-as` complaint about `Invalid capability: LongVectorEXT` at first, so I removed them in the backport PRs to see if I can bypass the `spirv-as` issue. That's why you didn't see the capability from the diff.
For the rewritten `.spt` tests, maybe I _should_ fix them? Leaving out capability doesn't affect the Reader since it doesn't check required extensions and capabilities at all; it's not a blocker for what we want to test. I'll just push a new commit to add it for validity and consistency though.

vmaksimo: Ah, okay. Then please restore the capability lines back as we won't use `spirv-as` here and on older branches

Qi-Ye-079: Done. Pushed to remaining branches as well.

vmaksimo (approved):

PR #4021, opened 2026-09-15
Fix FPRoundingMode decoration handling on mini-float conversions

vmaksimo:
The writer applies decoration on a call after its translation, to whatever value it ended up mapped to.
For mini-float conversions (`Float4E2M1EXT`, `Float8E4M3EXT`, `Float8E5M2EXT`) that's the bitcast to their packed-integer LLVM
representation added afterwards, not the conversion.
Decorate the conversion directly and clear the metadata, so there's nothing left for the bitcast to pick up.

Once correctly placed, the reader must also route these decorated conversions through the existing mini-float-aware path instead of the generic builtin-name mangler. 


lib/SPIRV/SPIRVReader.cpp
```diff
@@ -3240,7 +3240,19 @@ Value *SPIRVToLLVM::transValueWithoutDecoration(SPIRVValue *BV, Function *F,
...
+               Ty->isTypeFloat(8, FPEncodingFloat8E5M2EXT) ||
+               Ty->isTypeFloat(4, FPEncodingFloat4E2M1EXT) ||
+               Ty->isTypeFloat(4, internal::FPEncodingFloat4E2M1INTEL) ||
+               Ty->isTypeInt(4);
+      };
+      // Check both sides: the encoding may be on the source (e.g. an upcast
+      // out of Float4E2M1) rather than the result.
+      if ((BI->hasFPRoundingMode() || BI->isSaturatedConversion()) &&
+          !IsMiniFloatOrInt4(BI->getType()) &&
+          !IsMiniFloatOrInt4(
```
bashbaug, line 3254: I think this is due to (from the PR description):

> the reader must also route these decorated conversions through the existing mini-float-aware path instead of the generic builtin-name mangler, which has no case for a 4-bit result.

If so, might be worth a comment, since this is awfully subtle.

The PR description only describes a 4-bit result. If that's the case, do we really need to check the source operand also?

lib/SPIRV/SPIRVWriter.cpp
```diff
@@ -6014,6 +6014,12 @@ SPIRVValue *LLVMToSPIRVBase::transDirectCallInst(CallInst *CI,
         Conv->addDecorate(new SPIRVDecorate(
             DecorationSaturatedToLargestFloat8NormalConversionEXT, Conv));
 
+      // Target the conversion itself, not the bitcast possibly added below.
+      if (auto *IDecoMD = CI->getMetadata(SPIRV_MD_DECORATIONS)) {
+        transMetadataDecorations(IDecoMD, Conv);
+        CI->setMetadata(SPIRV_MD_DECORATIONS, nullptr);
+      }
```
bashbaug, line 6021: I think this is correct, but it's also a little scary. Are there other conditions where we might need to process decorations immediately on the first instruction in a sequence vs. following the usual path and applying the decoration to the last instruction in the sequence (which might be the only instruction)?  Can we make this more robust?

test/extensions/EXT/SPV_EXT_ocp_microscaling_types/fp_rounding_mode_rte.ll
```diff
@@ -0,0 +1,39 @@
...
+  %inPtr = getelementptr half, half addrspace(1)* %input, i64 %gid
+  %val = load half, half addrspace(1)* %inPtr, align 2
+  %conv = call i4 @_Z36__builtin_spirv_ConvertFP16ToE2M1EXTDh(half %val), !spirv.Decorations !1
+  %v0 = insertelement <2 x i4> undef, i4 %conv, i32 0
+  %v1 = insertelement <2 x i4> %v0, i4 0, i32 1
+  %outPtr = getelementptr i8, i8 addrspace(1)* %output, i64 %gid
+  %cast = bitcast i8 addrspace(1)* %outPtr to <2 x i4> addrspace(1)*
+  store <2 x i4> %v1, <2 x i4> addrspace(1)* %cast, align 1
+  ret void
+}
```
bashbaug: Not sure that it matters much, since this test is already pretty short, but I do think we could simplify and shorten this test considerably. For the purposes of the test, we really just need a half value (somehow) and the call to `__builtin_spirv_ConvertFP16ToE2M1EXT`, correct?

bashbaug (approved):
LGTM, but I do have a couple of possible improvements to consider.

Out of curiosity, could we have caught this issue sooner if we ran the SPIR-V validator more aggressively? Does the SPIR-V validator flag an FPRoundingMode on a bitcast as an error?

test/extensions/EXT/SPV_EXT_ocp_microscaling_types/fp_rounding_mode_rte.ll
vmaksimo: Sure, I'll trim the test + add the upcast case as well

vmaksimo:
> Out of curiosity, could we have caught this issue sooner if we ran the SPIR-V validator more aggressively? Does the SPIR-V validator flag an FPRoundingMode on a bitcast as an error?

Unfortunately yes, we could've caught this earlier - running it on the forward-translation output would've flagged the original bug at build time instead of needing the actual bug report.  Example of the failure:
```
error: line 24: FPRoundingMode decoration can be applied only to a conversion instruction to or from a floating-point type.
  %10 = OpBitcast %u4 %9
```

lib/SPIRV/SPIRVReader.cpp
vmaksimo: Without the source check we get incorrect IR in reverse like `_Z26__spirv_FConvert_Rhalf_rtei` instead of expected `_Z36__builtin_spirv_ConvertE2M1ToFP16EXTi`.
I'll add the test case and reword the PR description

lib/SPIRV/SPIRVWriter.cpp
vmaksimo: It's safe today because `transDecoration` has one call site, reached only after `transDirectCallInst` returns, and `transValue` translates each value once. But there was no real precedent for clearing this metadata before.

More robust and reusable thing might be:
```
  // Add new argument to set where the decoration should be applied to.
  SPIRVValue *DecoTarget = nullptr;
  auto *BV = transValueWithoutDecoration(V, BB, CreateForward, FuncTrans, &DecoTarget);
  ...
  if (!BV->isForward() && !transDecoration(V, DecoTarget ? DecoTarget : BV))
    return nullptr;


 SPIRVValue *LLVMToSPIRVBase::transDirectCallInst(CallInst *CI, SPIRVBasicBlock *BB,
                                                    SPIRVValue *DecoTarget) {
    SPIRVValue *Conv = BM->addInstTemplate(OC, BM->getIds(Ops), BB, DstTy);
    DecoTarget = Conv;

```
On one hand, I'm not totally sure if the mechanism will be used much; on other - I can imagine some other instructions that require further bitcast + similar fix.

What are your thoughts on this?

bashbaug: That's an interesting idea, and simpler than some of the other solutions I was considering.

I don't think we need to solve this right now. Even though it's a little scary, we seem to have gotten away without a mechanism like this until now :smile:.

If we run into a similar issue again, though, let's consider developing a more robust solution.

bashbaug (approved):

MrSidims (approved):
lgtm

PR #4022, opened 2026-09-15
add support for SPV_INTEL_subgroup_bitcast_shuffle

bashbaug:
Adds support for the SPV_INTEL_subgroup_bitcast_shuffle extension, see:

https://github.com/KhronosGroup/SPIRV-Registry/pull/454

This currently uses "internal" support for the new enums added by this capability.  If preferred, we could wait for the headers PR for this extension to land first and then switch to the "official" support, but it might be better to land the "internal" support first and then switch over to the "official" support at some TBD date:

https://github.com/KhronosGroup/SPIRV-Headers/pull/634


test/extensions/INTEL/SPV_INTEL_subgroup_bitcast_shuffle/bitcast_shuffle.ll
```diff
@@ -0,0 +1,276 @@
...
+; CHECK-SPIRV-DAG: TypeVector [[#TV8I16:]] [[#TI16]] 8
+; CHECK-SPIRV-DAG: TypeVector [[#TV4I32:]] [[#TI32]] 4
+; CHECK-SPIRV-DAG: TypeVector [[#TV2I64:]] [[#TI64]] 2
+; CHECK-SPIRV-DAG: TypeVector [[#TV16I16:]] [[#TI16]] 16
+; CHECK-SPIRV-DAG: TypeVector [[#TV8I32:]] [[#TI32]] 8
+; CHECK-SPIRV-DAG: TypeVector [[#TV4I64:]] [[#TI64]] 4
+; CHECK-SPIRV-DAG: TypeVector [[#TV16I32:]] [[#TI32]] 16
+; CHECK-SPIRV-DAG: TypeVector [[#TV8I64:]] [[#TI64]] 8
+
+; CHECK-SPIRV: SubgroupBitcastShuffleINTEL [[#TI16]] {{.*}} {{[0-9]+}}
```
vmaksimo: If we check only return type, maybe we can simplify the checks
```suggestion
; CHECK-SPIRV: SubgroupBitcastShuffleINTEL [[#TI16]]
```

test/extensions/INTEL/SPV_INTEL_subgroup_bitcast_shuffle/bitcast_shuffle.ll
```diff
@@ -0,0 +1,276 @@
...
+; CHECK-SPIRV: SubgroupBitcastShuffleINTEL [[#TV16I16]] {{.*}} {{[0-9]+}}
+; CHECK-SPIRV: SubgroupBitcastShuffleINTEL [[#TV4I64]] {{.*}} {{[0-9]+}}
+; CHECK-SPIRV: SubgroupBitcastShuffleINTEL [[#TV16I16]] {{.*}} {{[0-9]+}}
+; CHECK-SPIRV: SubgroupBitcastShuffleINTEL [[#TV4I64]] {{.*}} {{[0-9]+}}
+; CHECK-SPIRV: SubgroupBitcastShuffleINTEL [[#TV8I32]] {{.*}} {{[0-9]+}}
+; CHECK-SPIRV: SubgroupBitcastShuffleINTEL [[#TV8I64]] {{.*}} {{[0-9]+}}
+; CHECK-SPIRV: SubgroupBitcastShuffleINTEL [[#TV16I32]] {{.*}} {{[0-9]+}}
+
+; CHECK-LLVM-LABEL: define spir_kernel void @test_16bit()
+; CHECK-LLVM: call spir_func i16 @{{.*SubgroupBitcastShuffleINTEL_Rushort.*}}(<2 x i8>
```
vmaksimo: We mostly use full mangled instruction name in the tests - would you mind updating the checks?
This might be helpful in cases we somehow change mangle behavior to catch the regressions
```suggestion
; CHECK-LLVM: call spir_func i16 @_Z43__spirv_SubgroupBitcastShuffleINTEL_RushortDv2_c(<2 x i8>
```

lib/SPIRV/SPIRVReader.cpp
```diff
@@ -4207,6 +4207,7 @@ Instruction *SPIRVToLLVM::transSPIRVBuiltinFromInst(SPIRVInstruction *BI,
   // Old opcodes, for backward compatibility.
   case internal::OpClampConvertFToFINTEL:
   case internal::OpClampStochasticRoundFToFINTEL:
+  case internal::OpSubgroupBitcastShuffleINTEL:
```
vmaksimo: does not seem to be an old opcode 🤔

test/extensions/INTEL/SPV_INTEL_subgroup_bitcast_shuffle/bitcast_shuffle.ll
bashbaug: Sounds good, fixed!

lib/SPIRV/SPIRVReader.cpp
bashbaug: Hah! Fixed.

test/extensions/INTEL/SPV_INTEL_subgroup_bitcast_shuffle/bitcast_shuffle.ll
bashbaug: Sure, I changed to the fully mangled names.

I admittedly kind of liked leaving off some of the mangling information, but I can see why having the fully mangled names in the test is valuable, and they're already in the LLVM IR source so may as well have them in the checks also...

vmaksimo (approved):
I'd postpone the backports till the Headers are actually updated, but don't mind merging this to main as is now

bashbaug:
> I'd postpone the backports till the Headers are actually updated

OK, interesting. This will mean that the backports rely on the latest headers. That should be fine, but I thought it could be preferable to keep the internal support for backports and therefore work with older headers.

If we'd prefer to wait for the headers support for the backports instead, let's just hold off merging this to the main branch until the header changes have been merged, then we won't need to worry about cleaning up the internal support later.

vmaksimo:
> I thought it could be preferable to keep the internal support for backports and therefore work with older headers

I believe it doesn't matter much. Just if we wait, we'd be able to move the headers tag on every release branch with the backport PRs and avoid later cherry-picking of "switch to official headers instead of internal one" work.

bashbaug:
> Just if we wait, we'd be able to move the headers tag on every release branch with the backport PRs and avoid later cherry-picking of "switch to official headers instead of internal one" work

Gotcha, works for me. I'm hoping that the header support will land soon, in which case I'm fine waiting and switching to the official support. I'll mark this as "draft" so it doesn't get accidentally merged before then.

bashbaug:
I bumped the tag and switched to the support in the upstream headers, so I think this is ready to go now.

PR #4026, opened 2026-09-16
Fix OpTypeBufferSurfaceINTEL translation with target extension types

michalpaszkowski:
transType() is called twice for the same SPIR-V type, with UseTPT=true to get a TypedPointerType used only for mangling the builtin name, and with UseTPT=false to get the type the translated value actually receives. Types routed through BuiltinCallHelper::getSPIRVType() test UseTargetTypes before UseRealType, so both calls agree.

OpTypeBufferSurfaceINTEL tested UseTPT first, so with target extension types enabled the two disagreed. For OpConvertPtrToU on a buffer surface operand, transBuiltinFromInst() built a __spirv_ConvertPtrToU signature taking ptr addrspace(1) while the argument was a
target("spirv.BufferSurfaceINTEL", ...), tripping
"Calling a function with a bad signature!" in CallInst::Create().

Test UseTargetTypes first, and add OpTypeBufferSurfaceINTEL to getOCLOpaqueTypeAddrSpace() so the mangled name keeps the global address space the typed pointer representation used (otherwise "No address space is determined for some OCL type").


michalpaszkowski:
@vmaksimo @YuriPlyakhin This will most likely also need to go to branches for newer LLVM revisions but I need to do additional tests.

lib/SPIRV/SPIRVReader.cpp
```diff
@@ -465,7 +447,7 @@ Type *SPIRVToLLVM::transType(SPIRVType *T, bool UseTPT) {
   case OpTypeBufferSurfaceINTEL: {
     auto *PST = static_cast<SPIRVTypeBufferSurfaceINTEL *>(T);
     Type *Ty = nullptr;
-    if (UseTPT) {
+    if (UseTPT && !UseTargetTypes) {
```
vmaksimo, line 450: Oh, this is a direct follow-up to #3943 - we should've verified all the code paths back then

test/extensions/INTEL/SPV_INTEL_vector_compute/buffer_surface_intel_convert_ptr_to_u.ll
```diff
@@ -0,0 +1,49 @@
...
+
+target triple = "spir64"
+
+%intel.buffer_rw_t = type opaque
+
+declare void @llvm.some.unknown.intrinsic.i32(i32)
+
+define spir_kernel void @test(%intel.buffer_rw_t addrspace(1)* %buf) #0 {
+entry:
+; LLVM: call spir_func i32 @_Z26__spirv_ConvertPtrToU_RintPU3AS1{{[0-9]+}}__spirv_BufferSurfaceINTEL{{[_0-9]*}}(target("spirv.BufferSurfaceINTEL", 2) %buf)
```
vmaksimo, line 42: I'd include full mangling instead

vmaksimo (approved):
As this follow-ups the earlier fix for 16-branch only - we can proceed.
But I'd still like to see the related fix on main and newer release branches as well.

On main (but actually any release starting from 17 and newer), the reader fix will be simpler than the reorder here - just drop the `if (UseTPT)/else` and always build the `TargetExtType`.

test/extensions/INTEL/SPV_INTEL_vector_compute/buffer_surface_intel_convert_ptr_to_u.ll
michalpaszkowski: Changed in https://github.com/KhronosGroup/SPIRV-LLVM-Translator/pull/4026/commits/3a8a5489335c7d8f6d0b5627d441f3f004e5422d

PR #4027, opened 2026-09-22
Legalize unsupported-length vectors into two legal halves

alexander-shaposhnikov:
SROA can promote an aggregate into a vector whose component count SPIR-V                                                                                                                                    
does not support, e.g. bitcast <8 x i32> to <32 x i8> read back by                                                                                                                                          extractelement and by slicing shufflevectors. Those previously reached                                                                                                                                     report_fatal_error("Unsupported vector type with N elements").                                                                                                                                              
                                                                                                                                                                                                                
 Split such values into two halves of a supported length and rewrite the                                                                                                                                     
consumers. Shapes that are still not handled keep their uses and are                                                                                                                                      reported as before.


alexander-shaposhnikov:
cc: @Qi-Ye-079

test/lower-non-standard-vector-sroa.ll
```diff
@@ -0,0 +1,40 @@
+; RUN: llvm-as %s -o %t.bc
+; RUN: llvm-spirv -s %t.bc -o - | llvm-dis -o - | FileCheck %s \
```
MrSidims, line 2: Please add a test showing, that `llvm-spirv --spirv-ext=+SPV_EXT_long_vector` prevents this legalization

MrSidims (commented):
Frankly, I'm not the greatest fan of SPIRVLowerBitCastToNonStandardType pass at all as legalizing types was never in llvm-spirv design and was never a goal of the tool.

I'd like to suggest few alternatives to consider, rather than extending a pass that should never be here:
1. If you need to have type legalization when converting LLVM IR to SPIR-V I recommend to try out SPIR-V target in LLVM, which is doing it already and is in a some-what product-ready quality state;
2. ~You may also (try to) stop LLVM middleend from creating such vectors by not listing them in target datalayout (if I remember correctly, this trick helped in SYCL compiler to prevent SROA from collapsing things like <3 x i16> to i48~ (UPD: this won't work actually in this case);
3. FYI: there is [SPV_EXT_long_vector](https://github.com/KhronosGroup/SPIRV-Registry/blob/main/extensions/EXT/SPV_EXT_long_vector.asciidoc) extension that is supported in llvm-spirv, so you may preserve the vector you are concerned about in IR and let a proper compiler backend to deal with it.

alexander-shaposhnikov:
Many thanks,
yeah, in fact, I started with exploring the idea of getting rid of this translation in favor of the spir-v target, but this turned out to be a bigger undertaking for us (still hoping to try to push on this direction, but not in the near term, unfortunately, the compilation pipeline I'm dealing with involves some proprietary steps + there are more complications).

MrSidims:
> Many thanks, yeah, in fact, I started with exploring the idea of getting rid of this translation in favor of the spir-v target, but this turned out to be a bigger undertaking for us (still hoping to try to push on this direction, but not in the near term, unfortunately, the compilation pipeline I'm dealing with involves some proprietary steps + there are more complications).

Understood. Feel free to submit issues against it in llvm-project. I’ll take a closer look, when reach my laptop, but for now, what I can see there are few (potential) issues:
1. Debug info placed on the bitcast seem to be not preserved;
2. (corrected) I suspect there might be some issues with i1 vectors - ~they~ bitcasts involving them are not supported in SPIR-V, but ~they~ these bitcasts might slip through the lowering with the patch.

alexander-shaposhnikov:
Thanks, added more tests.

PR #4028, opened 2026-09-22
Update test for SPIR-V llvm target NoCapture emission

svenvh:
Update after llvm-project commit eebfba998a5f ("[SPIR-V] Emit NoCapture function parameter decoration for captures(none) args (#223694)", 2026-09-18).

PR #4029, opened 2026-09-22
Split fence_inst.ll backend check into its own CHECK-LLC prefix

vmaksimo:
llvm/llvm-project#223023 fixes the backend's fence lowering to name every reachable storage class, so `fence_inst.ll`'s backend path now recovers `mem_fence` flags as `7` instead of `0`. The `llvm-spirv` translator path is unaffected and keeps its own `CHECK-LLVM` expectation of `0`.

(Rebased after #4028 landed the identical fix for `builtin_function_readnone_attr.ll`'s `CHECK-LLC`, which this branch originally also included.)


test/transcoding/builtin_function_readnone_attr.ll
```diff
@@ -16,7 +16,7 @@
 ; CHECK-SPIRV: Decorate [[#B]] FuncParamAttr 7
 
 ; CHECK-LLVM: {{.*}}void @test_builtin_readnone(ptr readonly captures(none) %{{.*}}, ptr readnone captures(none) %{{.*}})
-; CHECK-LLC: {{.*}}void @test_builtin_readnone(ptr readonly %{{.*}}, ptr readnone %{{.*}})
```
svenvh, line 19: heh, I just merged a fix for this in #4028 :)

test/transcoding/fence_inst.ll
```diff
@@ -10,7 +10,7 @@
 ; RUN: %if spirv-backend %{ llc -O0 -mtriple=spirv64-unknown-unknown -filetype=obj %s -o %t.llc.spv %}
 ; RUN: %if spirv-backend %{ llvm-spirv -r %t.llc.spv -o %t.llc.rev.bc %}
 ; RUN: %if spirv-backend %{ llvm-dis %t.llc.rev.bc -o %t.llc.rev.ll %}
-; RUN: %if spirv-backend %{ FileCheck %s --check-prefix=CHECK-LLVM < %t.llc.rev.ll %}
+; RUN: %if spirv-backend %{ FileCheck %s --check-prefix=CHECK-LLC < %t.llc.rev.ll %}
```
svenvh, line 13: I wonder if we should instead update the SPIR-V writer to produce the same as the SPIR-V backend?

vmaksimo: I thought of maybe just drop the qualifiers checks from the tests, but your idea seems more reasonable :)

vmaksimo:
Not needed after #4028

test/transcoding/fence_inst.ll
svenvh: Cool, I've opened #4030 to do that.

